PR #334, opened 2023-05-11
feat: Include knative webhooks in readiness probes

ellistarn:
Fixes aws/karpenter/issues/3867

**Description**
Webhook readiness was not considered by probes, leading to brief downtime of the webhook during rolling updates.

Knative hardcodes port 8080, so I moved the metrics port to container port 8000 in a buddy PR. Metrics are still scrapable at the karpenter service on port 8080.

**How was this change tested?**
* Update Provisioner during single pod rolling update

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #416, opened 2023-07-19
fix: Close response to fix file descriptor leak

jonathan-innis:
Fixes https://github.com/aws/karpenter/issues/4296

**Description**

#4296 caught an issue where Karpenter was leaking file descriptors in its webhook checks. This can be observed from the following graph that shows the Karpenter `container_file_descriptors` metric that tracks the number of open file descriptors on a container.

Screenshot 2023-07-19 at 10 26 58 AM

The root cause of this issue is that Golang's default HTTP client opens a connection and holds that connection open when receiving a valid response from the server, which causes a file descriptor to remain open until the connection is closed (some details described [here](https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body)). 

This issue with not closing the response body was technically introduced in the following PR: #334 (introduced in Karpenter `v0.28.0`) but did not become an issue until `v0.29.1` since the response returned by the server used in `v0.28.0-v0.29.0` did not send a response body. Adding a response body to the response sent from the server called out to in these version also results in a file descriptor leak.

This change closes the response body and and ensures that the response body is read so that the http client can re-use the same connection from the connection pool instead of having to create another one. The file descriptor leak can seen to be resolved through the `container_file_descriptors` metric now:

Screenshot 2023-07-19 at 10 33 50 AM

`make presubmit`
- Deployed and viewed prometheus metrics
- Confirmation in the issue from the user who reported it

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -190,10 +190,15 @@ func webhookChecker(ctx context.Context) healthz.Checker {
 		if err != nil {
 			return err
 		}
+		// Close the body to avoid leaking file descriptors
+		// Always read the body so we can re-use the connection: https://stackoverflow.com/questions/17948827/reusing-http-connections-in-go
+		_, _ = io.ReadAll(res.Body)
+		res.Body.Close()
+
 		// If there is a server-side error or path not found,
 		// consider liveness to have failed
 		if res.StatusCode >= 500 || res.StatusCode == 404 {
-			return fmt.Errorf("webhook probe failed, %s", lo.Must(io.ReadAll(res.Body)))
+			return fmt.Errorf("webhook probe failed with status code %d", res.StatusCode)
 		}
 		return nil
 	}


pkg/operator/operator.go
```diff
@@ -190,10 +190,15 @@ func webhookChecker(ctx context.Context) healthz.Checker {
 		if err != nil {
 			return err
 		}
+		// Close the body to avoid leaking file descriptors
+		// Always read the body so we can re-use the connection: https://stackoverflow.com/questions/17948827/reusing-http-connections-in-go
+		body := lo.Must(io.ReadAll(res.Body))
+		res.Body.Close()
```
njtran, line 196: Let's do something like this so that there's no problem with lo.Must()
```
    // Close the connection to reuse it
    defer response.Body.Close()

    body, err := ioutil.ReadAll(response.Body)
    if err != nil {
        log.Fatalf("Couldn't parse response body. %+v", err)
    }

    return body
```

tzneal: I don't think you want to fatal here. Wouldn't any transient network error here cause Karpenter to restart?

njtran (approved):

pkg/operator/operator.go
njtran: Yeah sorry I just copied from the stack overflow conversation, just meant to not do `lo.Must()`.

PR #418, opened 2023-07-19
perf: improve scheduling speed

tzneal:
Allocatable() is called a lot. Memoize this using a sync.Map so we can avoid the expensive calculation.

name               old time/op   new time/op   delta
Scheduling1-12       566µs ± 9%    382µs ± 5%  -32.54%  (p=0.004 n=5+6)
Scheduling50-12     26.2ms ± 4%   20.0ms ± 6%  -23.53%  (p=0.004 n=5+6)
Scheduling100-12    53.1ms ± 8%   39.4ms ± 6%  -25.70%  (p=0.004 n=5+6)
Scheduling500-12     272ms ± 5%    202ms ±10%  -25.68%  (p=0.004 n=5+6)
Scheduling1000-12    596ms ±11%    457ms ±12%  -23.38%  (p=0.004 n=5+6)
Scheduling2000-12    1.08s ± 4%    0.89s ±10%  -17.46%  (p=0.004 n=5+6)
Scheduling5000-12    2.83s ± 8%    2.21s ± 6%  -21.91%  (p=0.004 n=5+6)

Fixes #N/A 

**Description**

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"math"
 	"sort"
+	"sync"
 
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
@@ -88,10 +89,20 @@ type InstanceType struct {
 	// Overhead is the amount of resource overhead expected to be used by kubelet and any other system daemons outside
 	// of Kubernetes.
 	Overhead *InstanceTypeOverhead
+
+	once        sync.Once
+	allocatable v1.ResourceList
+}
+
+// precompute is used to ensure we only compute the allocatable resources onces as its called many times
+// and the operation is fairly expensive.
+func (i *InstanceType) precompute() {
+	i.allocatable = resources.Subtract(i.Capacity, i.Overhead.Total())
 }
 
 func (i *InstanceType) Allocatable() v1.ResourceList {
-	return resources.Subtract(i.Capacity, i.Overhead.Total())
+	i.once.Do(i.precompute)
+	return i.allocatable.DeepCopy()
 }
 
 type InstanceTypeOverhead struct {
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -296,7 +296,11 @@ func (in *StateNode) VolumeLimits() scheduling.VolumeCount {
 }
 
 func (in *StateNode) PodRequests() v1.ResourceList {
-	return resources.Merge(lo.Values(in.podRequests)...)
+	var totalRequests v1.ResourceList
+	for _, requests := range in.podRequests {
+		totalRequests = resources.MergeInto(totalRequests, requests)
+	}
+	return totalRequests
 }
 
 func (in *StateNode) PodLimits() v1.ResourceList {
--- a/pkg/scheduling/requirement.go
+++ b/pkg/scheduling/requirement.go
@@ -40,6 +40,20 @@ func NewRequirement(key string, operator v1.NodeSelectorOperator, values ...stri
 	if normalized, ok := v1alpha5.NormalizedLabels[key]; ok {
 		key = normalized
 	}
+
+	// This is a super-common case, so optimize for it an inline everything.
+	if operator == v1.NodeSelectorOpIn {
+		s := make(sets.Set[string], len(values))
+		for _, value := range values {
+			s[value] = sets.Empty{}
+		}
+		return &Requirement{
+			Key:        key,
+			values:     s,
+			complement: false,
+		}
+	}
+
 	r := &Requirement{
 		Key:        key,
 		values:     sets.New[string](),
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"time"
 
 	"github.com/samber/lo"
 	csitranslation "k8s.io/csi-translation-lib"
@@ -29,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/aws/karpenter-core/pkg/utils/atomic"
 	"github.com/aws/karpenter-core/pkg/utils/pretty"
 )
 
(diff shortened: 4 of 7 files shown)


pkg/cloudprovider/types.go
```diff
@@ -90,8 +91,18 @@ type InstanceType struct {
 	Overhead *InstanceTypeOverhead
 }
 
+// allocatableCache caches the allocatable resources for each instance type.  It makes the assumption that once an
+// instance type reports allocatable resources, it won't change in the future. This is important for correctness. This
+// change improved the scheduling performance ~20-30% in testing.
+var allocatableCache sync.Map
```
ellistarn, line 97: IIUC, instance types vary depending on NodeTemplate, so you can get name collision.

ellistarn (changes requested):

pkg/utils/resources/resources.go
```diff
@@ -59,6 +59,23 @@ func Merge(resources ...v1.ResourceList) v1.ResourceList {
 	return result
 }
 
+// MergeInto sums the resources from src into dest, modifying dest
+func MergeInto(dest v1.ResourceList, src v1.ResourceList) v1.ResourceList {
```
jonathan-innis, line 63: Can we make a comment that this is more performant than the alternative? I can see someone looking at this later and wondering why we have both this and `Merge`

pkg/scheduling/volumeusage.go
```diff
@@ -221,6 +235,7 @@ func (v *VolumeUsage) discoverDefaultStorageClassName(ctx context.Context, kubeC
 	sort.Slice(defaults, func(i, j int) bool {
 		return defaults[i].CreationTimestamp.After(defaults[j].CreationTimestamp.Time)
 	})
+	defaultStoragClass.Set(defaults[0].Name)
```
jonathan-innis, line 238:
```suggestion
	defaultStorageClass.Set(defaults[0].Name)
```

pkg/scheduling/volumeusage.go
```diff
@@ -50,6 +52,14 @@ type VolumeUsage struct {
 	podVolumes map[types.NamespacedName]volumes
 }
 
+// Cache the lookup of our default storage class, global atomic cache since the volume usage is per node
+var defaultStoragClass = atomic.NewCachedVariable[string](1 * time.Minute)
```
jonathan-innis, line 56:
```suggestion
var defaultStorageClass = atomic.NewCachedVariable[string](1 * time.Minute)
```

pkg/cloudprovider/types.go
```diff
@@ -88,10 +89,18 @@ type InstanceType struct {
...
+}
+
+func (i *InstanceType) precompute() {
+	i.allocatable = resources.Subtract(i.Capacity, i.Overhead.Total())
 }
 
 func (i *InstanceType) Allocatable() v1.ResourceList {
-	return resources.Subtract(i.Capacity, i.Overhead.Total())
+	i.once.Do(i.precompute)
+	return i.allocatable.DeepCopy()
```
jonathan-innis, line 103: Do we need this `DeepCopy` or can we assume that we shouldn't be modifying this readonly value?

pkg/cloudprovider/types.go
```diff
@@ -88,10 +89,18 @@ type InstanceType struct {
 	// Overhead is the amount of resource overhead expected to be used by kubelet and any other system daemons outside
 	// of Kubernetes.
 	Overhead *InstanceTypeOverhead
+
+	once        sync.Once
+	allocatable v1.ResourceList
+}
+
+func (i *InstanceType) precompute() {
```
jonathan-innis, line 97: Can we include some detail on why we are precomputing so someone doesn't come along later and think this code can be removed?

pkg/controllers/state/statenode.go
```diff
@@ -296,7 +296,11 @@ func (in *StateNode) VolumeLimits() scheduling.VolumeCount {
 }
 
 func (in *StateNode) PodRequests() v1.ResourceList {
-	return resources.Merge(lo.Values(in.podRequests)...)
+	var totalRequests v1.ResourceList
+	for _, requests := range in.podRequests {
```
jonathan-innis, line 300: `lo.Reduce`?

pkg/scheduling/requirement.go
```diff
@@ -40,6 +40,20 @@ func NewRequirement(key string, operator v1.NodeSelectorOperator, values ...stri
 	if normalized, ok := v1alpha5.NormalizedLabels[key]; ok {
 		key = normalized
 	}
+
+	// This is a super-common case, so optimize for it an inline everything.
+	if operator == v1.NodeSelectorOpIn {
+		s := make(sets.Set[string], len(values))
+		for _, value := range values {
```
jonathan-innis, line 47: Is this tangibly different than calling `Insert`?

jonathan-innis (commented):
Awesome finds!

pkg/utils/resources/resources.go
tzneal: Yes, added a comment.

pkg/cloudprovider/types.go
tzneal: I just err'd on the side of safety here, it's still faster than before.

pkg/cloudprovider/types.go
tzneal: Done

pkg/controllers/state/statenode.go
tzneal: I'm not a huge fan of lo and it would require more benchmarking to see if that hurts performance.

pkg/scheduling/requirement.go
tzneal: Yes, a bit faster:
```
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkManual-12    	12869192	       122.4 ns/op
BenchmarkNew-12       	14114200	       154.7 ns/op
BenchmarkInsert-12    	13081554	       136.3 ns/op
PASS
```
```
func BenchmarkManual(b *testing.B) {
	values := make([]string, b.N)
	for i := 0; i < b.N; i++ {
		values[i] = strconv.Itoa(i)
	}

	b.ResetTimer()
	s := make(sets.Set[string], len(values))
	for _, value := range values {
		s[value] = sets.Empty{}
	}
	_ = s
}

func BenchmarkNew(b *testing.B) {
	values := make([]string, b.N)
	for i := 0; i < b.N; i++ {
		values[i] = strconv.Itoa(i)
	}

	b.ResetTimer()
	s := sets.New(values...)
	_ = s
}
func BenchmarkInsert(b *testing.B) {
	values := make([]string, b.N)
	for i := 0; i < b.N; i++ {
		values[i] = strconv.Itoa(i)
	}

	b.ResetTimer()
	s := make(sets.Set[string], len(values))
	s.Insert(values...)
	_ = s
}
```

pkg/cloudprovider/types.go
tzneal: Modified to cache in the instance type itself.

jonathan-innis (approved):
LGTM 🚀

PR #423, opened 2023-07-27
feat: Add Node Requirement Drift

engedaam:
Fixes #N/A 

**Description**
- Karpenter will be able to detect machines that have been drifted from the Node Requirements of its provisioner
- Implements drift for Node requirements.

- Manually tested
- Added e2e testing (https://github.com/aws/karpenter/pull/3933)


pkg/controllers/machine/disruption/drift.go
```diff
@@ -100,3 +102,13 @@ func areStaticFieldsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5
 	}
 	return provisionerHash != machineHash
 }
+
+func areNodeRequirementsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5.Machine) bool {
+	return !lo.EveryBy(provisioner.Spec.Requirements, func(provisionerReq v1.NodeSelectorRequirement) bool {
+		return lo.ContainsBy(machine.Spec.Requirements, func(machineReq v1.NodeSelectorRequirement) bool {
+			return provisionerReq.Key == machineReq.Key &&
+				provisionerReq.Operator == machineReq.Operator &&
+				lo.Every(provisionerReq.Values, machineReq.Values)
```
njtran, line 111: I don't think this is right. Wouldn't this check if there's a node requirement that matches each provisioner requirement? For instance, If the provisioner requirement is `Key Exists` and the Machine requirement is `Key In Values =[a,b,c]`, this logic would say that's not compatible, when it is.

This also assumes incorrectly that each requirement will have a set of values.

njtran: I think the work is more complicated than this, and can use the requirements logic too. 

As a thought exercise, you should be able to check the cartesian product of all the possible Operators for a Provisioner and for a Machine

https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core#NodeSelectorOperator

pkg/controllers/machine/disruption/drift_test.go
```diff
@@ -127,7 +127,142 @@ var _ = Describe("Drift", func() {
...
+			}
+			ExpectApplied(ctx, env.Client, provisioner, machine, node)
+			ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKeyFromObject(machine))
+			machine = ExpectExists(ctx, env.Client, machine)
+			Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted)).To(BeNil())
+
+			provisioner.Spec.Requirements = []v1.NodeSelectorRequirement{
+				{Key: v1alpha5.LabelCapacityType, Operator: v1.NodeSelectorOpIn, Values: []string{v1alpha5.CapacityTypeOnDemand}},
+				{Key: v1.LabelOSStable, Operator: v1.NodeSelectorOpIn, Values: []string{string(v1.Linux)}},
+				{Key: v1.LabelArchStable, Operator: v1.NodeSelectorOpIn, Values: []string{v1alpha5.ArchitectureAmd64}},
```
njtran, line 177: What if this was the chosen architecture anyway for a Machine? Should this Drift?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -100,3 +103,43 @@ func areStaticFieldsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5
...
+}
+
+func driftedNRLessThen(provisionerValue string, machineValue string) bool {
+	valueInt, errMach := strconv.Atoi(machineValue)
+	provValueInt, errProv := strconv.Atoi(provisionerValue)
+	if errMach != nil && errProv != nil {
+		return false
+	}
+	return provValueInt > valueInt
+}
```
njtran, line 145: These are essentially the same function, consider parameterizing the difference into one function:
```suggestion
func nodeRequirementsCompare(provisionerValue string, machineValue string, lt bool) bool {
	valueInt, errMach := strconv.Atoi(machineValue)
	provValueInt, errProv := strconv.Atoi(provisionerValue)
	if errMach != nil && errProv != nil {
		return false
	}
	if lt {
	    return provValueInt > valueInt
	}
	return provValueInt < valueInt
}
```

pkg/controllers/machine/disruption/drift_test.go
```diff
@@ -127,7 +127,212 @@ var _ = Describe("Drift", func() {
...
+			),
+			Entry(
+				"should not return drifted if a machine is grater then node requirement",
+				[]v1.NodeSelectorRequirement{
+					{Key: v1alpha5.LabelCapacityType, Operator: v1.NodeSelectorOpIn, Values: []string{v1alpha5.CapacityTypeOnDemand}},
+					{Key: v1.LabelInstanceType, Operator: v1.NodeSelectorOpGt, Values: []string{"2"}},
+				},
+				[]v1.NodeSelectorRequirement{
+					{Key: v1alpha5.LabelCapacityType, Operator: v1.NodeSelectorOpIn, Values: []string{v1alpha5.CapacityTypeOnDemand}},
+					{Key: v1.LabelInstanceType, Operator: v1.NodeSelectorOpGt, Values: []string{"10"}},
```
njtran, line 257: nit: probably doesn't make a difference since it's a unit test, but consider using a non-stringed label. Like use LabelInstanceCPU instead of instanceType.

njtran (approved):
LGTM, just two nits.

pkg/controllers/machine/disruption/drift.go
```diff
@@ -100,3 +103,43 @@ func areStaticFieldsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5
 	}
 	return provisionerHash != machineHash
 }
+
+func areNodeRequirementsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5.Machine) bool {
+	return !lo.EveryBy(provisioner.Spec.Requirements, func(provisionerReq v1.NodeSelectorRequirement) bool {
```
jonathan-innis, line 108: I think there's a way to build the requirements using `scheduling.NewLabelReqruirements` and comparing compatability rather than having to do this manually.

pkg/controllers/machine/disruption/drift_test.go
engedaam: Non-stringed isn't available to use in core.

pkg/controllers/machine/disruption/drift.go
```diff
@@ -100,3 +101,12 @@ func areStaticFieldsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5
...
 	return provisionerHash != machineHash
 }
+
+func areNodeRequirementsDrifted(provisioner *v1alpha5.Provisioner, machine *v1alpha5.Machine) bool {
+	provisionerReq := scheduling.NewNodeSelectorRequirements(provisioner.Spec.Requirements...)
+	machineReq := scheduling.NewLabelRequirements(machine.Labels)
+
+	err := machineReq.StrictlyCompatible(provisionerReq)
+
+	return err != nil
```
jonathan-innis, line 111:
```suggestion
        // Every provisioner requirement is compatible with the Machine label set
	return machineReq.StrictlyCompatible(provisionerReq) != nil
```

pkg/scheduling/requirements.go
```diff
@@ -132,6 +132,19 @@ func (r Requirements) Compatible(requirements Requirements) (errs error) {
 	return multierr.Append(errs, r.Intersects(requirements))
 }
 
+// Compatible ensures the provided requirements can be met.
```
jonathan-innis, line 135: Can we update this comment?

pkg/scheduling/requirements.go
```diff
@@ -132,6 +132,19 @@ func (r Requirements) Compatible(requirements Requirements) (errs error) {
...
+func (r Requirements) StrictlyCompatible(requirements Requirements) (errs error) {
+	// All requirements must be defined in the requirements
+	for key := range requirements.Keys() {
+		if operator := requirements.Get(key).Operator(); r.Has(key) || operator == v1.NodeSelectorOpNotIn || operator == v1.NodeSelectorOpDoesNotExist {
```
njtran, line 139: What about keys that are in the Machine requirements, but aren't in the Provisioner requirements? For instance, my Machine Requirements have `Key In Values [a,b,c]` but my Provisioner requirements doesn't have Key. Wouldn't that be incompatible?

pkg/scheduling/requirements.go
```diff
@@ -132,6 +132,19 @@ func (r Requirements) Compatible(requirements Requirements) (errs error) {
 	return multierr.Append(errs, r.Intersects(requirements))
 }
 
+// Compatible ensures the provided requirements can be met.
+func (r Requirements) StrictlyCompatible(requirements Requirements) (errs error) {
```
jonathan-innis, line 136: Can we make sure that we have test cases for this in the `suite_test.go`

pkg/scheduling/requirements.go
```diff
@@ -132,6 +132,19 @@ func (r Requirements) Compatible(requirements Requirements) (errs error) {
...
+	// All requirements must be defined in the requirements
```
jonathan-innis, line 137:
```suggestion
	// All requirements in the passed through requirements should be defined in "r"
```

pkg/scheduling/requirements_test.go
```diff
@@ -288,6 +288,248 @@ var _ = Describe("Requirements", func() {
 			Expect(lessThan9.Compatible(lessThan1)).To(Succeed())
 			Expect(lessThan9.Compatible(lessThan9)).To(Succeed())
 		})
+		It("should be strictly compatible", func() {
```
jonathan-innis, line 291: It's hard to understand what's different here. It would be good to make a comment around what is _actually_ different between this and the above checks on `Compatible()` since (from what I can tell) this set of tests is basically just a copy from above with a few exceptions.

jonathan-innis (approved):
LGTM 🚀

engedaam:
Related to https://github.com/aws/karpenter/issues/1738

PR #424, opened 2023-07-27
[fix] Adding LeaderElectionNamespace to init of controllerruntime Manager

abeer-stripe:
Fixes - https://github.com/aws/karpenter/issues/4216

**Description**

Add leader election namespace during initialization so that Kubernetes clusters where default service account tokens are not automatically mounted can run Karpenter with leader election. Issue -  https://github.com/aws/karpenter/issues/4216

Tested in our setup where we don't mount default namespace service account.


jonathan-innis (approved):
LGTM 🚀

PR #426, opened 2023-07-28
chore: Add v1beta1 APIs

jonathan-innis:
Fixes #N/A 

**Description**

- This change adds the first iteration of the `v1beta1` APIs for `NodePool` and `NodeClaim`
- These APIs carry similar validation mechanisms to the `v1alpha5` APIs with similar fields
- These APIs are still subject to change due to approval of https://github.com/aws/karpenter/pull/4274

NOTE: This does not _release_ the APIs. This only preps other work that will lead up to the eventual formal release of these APIs into the project. Even after this PR is merged, these APIs _may_ change until the APIs are released through a formal release process of Karpenter.

`make presubmit`


pkg/apis/v1beta1/register.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	Group              = "karpenter.sh"
+	ExtensionsGroup    = "extensions." + Group
```
ellistarn, line 25: Still in use?

pkg/apis/v1beta1/register.go
```diff
@@ -0,0 +1,42 @@
...
+	CompatabilityGroup = "compatibility." + Group
+	TestingGroup       = "testing." + Group // Exclusively used for labeling/discovery in testing
```
ellistarn, line 27: Can we replace with a non-karpenter-specific testing group? Maybe a framework specific group?

pkg/apis/v1beta1/nodeclaim_status.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+var (
+	NodeLaunched      apis.ConditionType = "NodeLaunched"
+	NodeRegistered    apis.ConditionType = "NodeRegistered"
+	NodeInitialized   apis.ConditionType = "NodeInitialized"
+	NodeEmpty         apis.ConditionType = "NodeEmpty"
+	NodeDrifted       apis.ConditionType = "NodeDrifted"
+	NodeExpired       apis.ConditionType = "NodeExpired"
+	NodeUnderutilized apis.ConditionType = "NodeUnderutilized"
```
njtran, line 62: I don't think this is necessary to include here, but just a question if you think it impacts. 

Some users want a custom deprovisioning nomination mechanism, and I imagine it would fit cleanly here as `NodeUserDisrupted` condition. Would you agree? I'm sure it could just as easily be automatically done with a node controller and the machine disruption controller.

pkg/apis/v1beta1/nodepool.go
```diff
@@ -0,0 +1,141 @@
...
+	// +kubebuilder:validation:Maximum:=100
+	// +optional
+	Weight *int32 `json:"weight,omitempty"`
+}
+
+type Deprovisioning struct {
+	// ConsolidationTTL is the duration the controller will wait
+	// before attempting to terminate nodes that are empty
+	// Termination due to under-utilization is disabled if this field is not set.
+	// This field is mutually exclusive to consolidation.enabled
```
njtran, line 56: Do we need this line? There isn't any consolidation.enabled anymore.

pkg/apis/v1beta1/nodepool_validation.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		if err := IsRestrictedLabel(key); err != nil {
+			errs = errs.Also(apis.ErrInvalidKeyName(key, "labels", err.Error()))
+		}
+	}
+	return errs
+}
+
+func (in *Deprovisioning) validate() (errs *apis.FieldError) {
+	if in.ExpirationTTL.Duration < 0 {
```
njtran, line 73: Any point in also validating Emptiness TTL here too?

pkg/apis/v1beta1/labels.go
```diff
@@ -0,0 +1,133 @@
...
+const (
+	NodePoolLabelKey        = Group + "/nodepool"
+	NodeInitializedLabelKey = Group + "/initialized"
+	NodeRegisteredLabelKey  = Group + "/registered"
+	CapacityTypeLabelKey    = Group + "/capacity-type"
+)
+
+// Karpenter specific annotations
+const (
+	DoNotDisruptAnnotationKey          = Group + "/do-not-disrupt"
```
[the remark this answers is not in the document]
jonathan-innis: We may change this in a subsequent version. I don't want to block on our conversation with the SIG for the v1beta1 API.

pkg/apis/v1beta1/nodeclaim.go
```diff
@@ -0,0 +1,188 @@
...
+	// +optional
+	Resources ResourceRequirements `json:"resources,omitempty"`
+	// KubeletConfiguration are options passed to the kubelet when provisioning nodes
+	// +optional
+	KubeletConfiguration *KubeletConfiguration `json:"kubeletConfiguration,omitempty"`
+	// NodeClass is a reference to an object that defines provider specific configuration
+	// +required
+	NodeClass *NodeClassRef `json:"nodeClass"`
+	// Provider stores CloudProvider-specific details from a conversion from a v1alpha5.Provisioner
+	// TODO @joinnis: Remove this field when v1alpha5 is unsupported in a future version of Karpenter
```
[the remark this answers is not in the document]
jonathan-innis: It's not exposed as with other fields that are marked with json:"-". These fields all exist for internal conversion so all the business logic revolves around the v1beta1 versions.

pkg/apis/v1beta1/nodepool_validation.go
jonathan-innis: It's not exposed externally, so I don't see a reason to validate it in the webhook

pkg/apis/v1beta1/nodepool.go
jonathan-innis: Good catch!

pkg/apis/v1beta1/nodeclaim_status.go
jonathan-innis: Yeah, I think you're right. Definitely something that we can add when we fully support that feature.

njtran (dismissed):

njtran (approved):

njtran (approved):

PR #434, opened 2023-08-01
fix: Ensure `VolumeUsage` and `HostPortUsage` state is consistent

ellistarn:
Fixes https://github.com/aws/karpenter/issues/4349
Fixes https://github.com/aws/karpenter/issues/4254

**Description**

Propagate errors that occur in volume tracking. This is possible if a pod configures a PVC with a storage class that doesn't exist, and other clientgo failures. This PR re-orders the calls to `HostPortUsage` and `VolumeUsage` to ensure there are no errors when mutating state (either when scheduling nodes or updated the `cluster.State`). Calls will (in order):

1. Attempt to grab the `HostPorts` or the `Volumes` for a pod
2. Validate that we can add the `HostPorts` or `Volumes` to an existing node without violating scheduling constraints
3. Add the `HostPorts` and `Volumes` to update the state

```
make presubmit
```


pkg/scheduling/volumes.go
```diff
@@ -0,0 +1,218 @@
...
+func (u Volumes) Copy() Volumes {
+	cp := Volumes{}
+	for k, v := range u {
+		cp[k] = sets.New(sets.List(v)...)
+	}
+	return cp
+}
+
+//nolint:gocyclo
+func GetVolumes(ctx context.Context, kubeClient client.Client, pod *v1.Pod) (Volumes, error) {
```
ellistarn, line 98: Should this be tracked in Cluster State? Perhaps on UpdatePod?

jonathan-innis: I think this is fine where it is. There are a few different places where we are getting the pod volumes and I can see the utility of this being a utils function that can grab the volumes generically for a given pod.

ellistarn (commented):
@jonathan-innis and I collaborated on this PR. We'd like a neutral third party reviewer.

pkg/controllers/state/cluster.go
```diff
@@ -354,21 +353,13 @@ func (c *Cluster) DeleteDaemonSet(key types.NamespacedName) {
...
-		volumeUsage:       oldNode.volumeUsage,
-		daemonSetRequests: oldNode.daemonSetRequests,
-		daemonSetLimits:   oldNode.daemonSetLimits,
-		podRequests:       oldNode.podRequests,
-		podLimits:         oldNode.podLimits,
-		markedForDeletion: oldNode.markedForDeletion,
-		nominatedUntil:    oldNode.nominatedUntil,
+		n = NewNode()
+	} else {
+		n = oldNode.DeepCopy()
```
tzneal, line 360: This adds an extra copy of multiple maps that didn't used to be there.  The old code took advantage of the fact that the old node was not going to be used any longer and borrowed its internals.  Its cleaner to represent in language with move semantics.

pkg/controllers/state/cluster.go
```diff
@@ -392,25 +383,14 @@ func (c *Cluster) cleanupMachine(name string) {
...
-		podRequests:         map[types.NamespacedName]v1.ResourceList{},
-		podLimits:           map[types.NamespacedName]v1.ResourceList{},
-		hostPortUsage:       scheduling.NewHostPortUsage(),
-		volumeUsage:         scheduling.NewVolumeUsage(),
-		volumeLimits:        scheduling.VolumeCount{},
-		markedForDeletion:   oldNode.markedForDeletion,
-		nominatedUntil:      oldNode.nominatedUntil,
+		n = NewNode()
+	} else {
+		n = oldNode.DeepCopy()
```
tzneal, line 390: Similar to above, but now it copies multiple maps it doesn't need to and then clears them.

tzneal (commented):
This will be slower, not sure by how much. It would require profiling to determine.

tzneal (approved):
lgtm

tzneal (approved):
lgtm

PR #437, opened 2023-08-02
fix: Don't treat consolidation validation failures as errors, and emit a debug log instead

ellistarn:
Fixes #https://github.com/aws/karpenter/issues/4364

**Description**

Currently, different consolidation methods have different behavior.

* emptiness: returns an error, bailing from consolidation entirely and starting back at the beginning
* single machine: return an error, bailing from consolidation entirely and starting back at the beginning
* multi machine: log a DEBUG, moving to the next consolidation method

This change aligns all behavior to log and continue. e.g., if pod churn has invalidated an empty consolidation, we'll try multi. If multi has become invalidated, we'll try single. 

`make presubmit`

### DIFF
--- a/pkg/controllers/deprovisioning/emptymachineconsolidation.go
+++ b/pkg/controllers/deprovisioning/emptymachineconsolidation.go
@@ -82,7 +82,8 @@ func (c *EmptyMachineConsolidation) ComputeCommand(ctx context.Context, candidat
 	// the machine isn't a target of a recent scheduling simulation
 	for _, n := range candidatesToDelete {
 		if len(n.pods) != 0 || c.cluster.IsNodeNominated(n.Name()) {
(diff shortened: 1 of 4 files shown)


pkg/controllers/deprovisioning/emptymachineconsolidation.go
```diff
@@ -82,7 +82,8 @@ func (c *EmptyMachineConsolidation) ComputeCommand(ctx context.Context, candidat
 	// the machine isn't a target of a recent scheduling simulation
```
jonathan-innis, line 82: I wonder if it makes more sense to create an `ErrValidationFailed` error type and then log about this validation failure generally

jonathan-innis (approved):
LGTM 🚀

PR #438, opened 2023-08-02
fix: Translate all in-tree providers to CSI

jonathan-innis:
Fixes https://github.com/aws/karpenter/issues/4003

**Description**

Currently, we ignore in-tree drivers that we don't recognize since the volume limit for a node cannot be looked up from the CSINode that we use to maintain our other volume limits. This PR enables the translation mechanism from the in-tree provisioner name (if specified) to the out-of-tree CSI provisioner name (if one exists). This translation will happen both from the driver name in the volume and the provisioner name in the StorageClass.

`make presubmit`

### DIFF
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/samber/lo"
 	csitranslation "k8s.io/csi-translation-lib"
+	"k8s.io/csi-translation-lib/plugins"
 	"knative.dev/pkg/logging"
 
 	v1 "k8s.io/api/core/v1"
@@ -31,7 +32,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aws/karpenter-core/pkg/utils/atomic"
-	"github.com/aws/karpenter-core/pkg/utils/pretty"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/scheduling/volumeusage.go
```diff
@@ -171,99 +170,93 @@ func (v *VolumeUsage) Validate(ctx context.Context, kubeClient client.Client, po
 func (v *VolumeUsage) validate(ctx context.Context, kubeClient client.Client, pod *v1.Pod) (volumes, error) {
```
ellistarn, line 170: I have a PR WIP that calls this getPodVolumes. Thoughts?

pkg/scheduling/volumeusage.go
```diff
@@ -171,99 +170,93 @@ func (v *VolumeUsage) Validate(ctx context.Context, kubeClient client.Client, po
...
 		if driverName != "" {
-			podPVCs.Add(driverName, pvcID)
+			podPVCs.Add(driverName, client.ObjectKeyFromObject(pvc).String())
 		}
 	}
 	return podPVCs, nil
 }
 
-func (v *VolumeUsage) discoverDefaultStorageClassName(ctx context.Context, kubeClient client.Client) (*string, error) {
+func (v *VolumeUsage) discoverDefaultStorageClassName(ctx context.Context, kubeClient client.Client) (string, error) {
```
ellistarn, line 207: lol I did all the same here.

pkg/scheduling/volumeusage.go
jonathan-innis: Yeah, makes sense

pkg/scheduling/volumeusage.go
jonathan-innis: 😆

ellistarn (approved):

PR #441, opened 2023-08-04
fix: Handle ephemeral volume topology

jonathan-innis:
Fixes #N/A 

**Description**

Prior to this change, Karpenter would only handle volume topology for PVCs that were directly reference by pods through the `.spec.volumes[*].persistentVolumeClaim` field; however, pods can also reference [dynamically-generated generic ephemeral volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes), referenced through `.spec.volumes[*].ephemeral`. We did not respect PVCs that were bound to pods with volume topology. This change allows us to handle ephemeral volumes and scale-out if the current nodes do not satisfy the volume topology.

- `make presubmit`
- Deploying bound PVC/PV to cluster, removing nodes and expecting Karpenter to assign volume topologies correctly

### DIFF
--- a/pkg/controllers/provisioning/scheduling/suite_test.go
+++ b/pkg/controllers/provisioning/scheduling/suite_test.go
@@ -2392,6 +2392,96 @@ var _ = Describe("Existing Nodes", func() {
 		// Expect that the scheduled node is equal to the ready node since it's initialized
 		Expect(scheduledNode.Name).To(Equal(nodes[elem].Name))
 	})
+	It("should consider a pod incompatible with an existing node but compatible with Provisioner", func() {
+		machine, node := test.MachineAndNode(v1alpha5.Machine{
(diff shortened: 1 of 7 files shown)


ellistarn (commented):
Can we write an integ test for this?

jonathan-innis:
> Can we write an integ test for this?

Yep, we absolutely need tests for these scenarios with volume and storage class topologies. Right now, we aren't doing integration testing against _any_ ephemeral volume storage support either.

jonathan-innis:
Two tracking issues for adding E2E volume testing for ephemeral volumes and respecting volume topologies:
1. https://github.com/aws/karpenter/issues/4381
2. https://github.com/aws/karpenter/issues/4382

tzneal (approved):
lgtm

PR #442, opened 2023-08-04
feat: Disruption Controller Metrics

engedaam:
Fixes #N/A 

**Description**
- Added disruption controller metrics to provider customer with a better visibility into drift and expiration disruption

- Manually tested


pkg/controllers/machine/disruption/drift.go
```diff
@@ -83,12 +85,30 @@ func (d *Drift) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisioner
 // isDrifted will check if a machine is drifted from the fields in the provisioner.Spec and
 // the cloudprovider
 func (d *Drift) isDrifted(ctx context.Context, provisioner *v1alpha5.Provisioner, machine *v1alpha5.Machine) (bool, error) {
```
jonathan-innis, line 87: Is there a way to generalize this so we don't have to code this in so specifically?

jonathan-innis: We should also be thinking about how much the user cares about the kind of drift that caused the disruption. Do we think (at scale) this is relevant information to the user? Is there a way that we could codify the type of drift through the disruption reason or message in some way?

jonathan-innis (commented):
@ellistarn and I had talked a little bit about a status conditions library that monitors the amount of time a NodeClaim is in a certain status condition. I'm wondering if that might ber a more useful metric here since the user is probably most interested in the number of machines that are disrupted at any given time. I could also _potentially_ see an argument for firing a counter metric when the condition is added but this should be generalized in a way that every disruption type doesn't have to know about it.

pkg/metrics/metrics.go
```diff
@@ -165,10 +165,34 @@ var (
...
+		[]string{
+			ReasonLabel,
+			ProvisionerLabel,
+		},
+	)
+	MachinesDriftedReasonCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: Namespace,
+			Subsystem: machineSubsystem,
+			Name:      "drifted_reason",
```
jonathan-innis, line 184:
```suggestion
			Name:      "drifted",
```

pkg/metrics/metrics.go
```diff
@@ -165,10 +165,34 @@ var (
...
+			Subsystem: machineSubsystem,
+			Name:      "disrupted",
+			Help:      "Number of machines disrupted in total by Karpenter. Labeled by reason the machine was disrupted.",
+		},
+		[]string{
+			ReasonLabel,
+			ProvisionerLabel,
+		},
+	)
+	MachinesDriftedReasonCounter = prometheus.NewCounterVec(
```
jonathan-innis, line 180:
```suggestion
	MachinesDriftedCounter = prometheus.NewCounterVec(
```

jonathan-innis: Probably better to just have a separate counter that's called `karpenter_machines_drifted` and contains the reason

pkg/metrics/metrics.go
```diff
@@ -165,10 +165,34 @@ var (
...
 	)
+	MachinesDisruptedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: Namespace,
+			Subsystem: machineSubsystem,
+			Name:      "disrupted",
+			Help:      "Number of machines disrupted in total by Karpenter. Labeled by reason the machine was disrupted.",
+		},
+		[]string{
+			ReasonLabel,
```
jonathan-innis, line 176:
```suggestion
			TypeLabel,
```

jonathan-innis: This now becomes disruption type right?

pkg/metrics/constants.go
```diff
@@ -27,6 +27,7 @@ const (
 	ProvisionerLabel = "provisioner"
 	NodePoolLabel    = "nodepool"
 	ReasonLabel      = "reason"
+	TypeLabel        = "type"
```
jonathan-innis, line 30: This is specific to disruption so does this need to be in the generic metrics package?

engedaam: If we move this into the disruption controller, we would have a dependency cycle

jonathan-innis (approved):
LGTM 🚀

pkg/controllers/machine/disruption/drift.go
```diff
@@ -26,12 +26,20 @@ import (
...
+
 	"github.com/aws/karpenter-core/pkg/apis/settings"
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
+	"github.com/aws/karpenter-core/pkg/metrics"
 	"github.com/aws/karpenter-core/pkg/scheduling"
 )
 
+const (
+	ProvisionerSpecDrifted cloudprovider.DriftReason = "ProvisionerSpecDrifted"
```
njtran, line 39: Feel like you can just say Provisioner here. We don't do any Status drift

pkg/controllers/machine/disruption/expiration.go
```diff
@@ -83,7 +86,10 @@ func (e *Expiration) Reconcile(ctx context.Context, provisioner *v1alpha5.Provis
 	})
 	if !hasExpiredCondition {
 		logging.FromContext(ctx).Debugf("marking machine as expired")
+		metrics.MachinesDisruptedCounter.With(prometheus.Labels{
+			metrics.TypeLabel:        "expiration",
```
njtran, line 90: I think you should set this as an enum, same with the other machine ones

pkg/controllers/machine/disruption/drift.go
```diff
@@ -26,12 +26,20 @@ import (
...
 	"github.com/aws/karpenter-core/pkg/apis/settings"
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
+	"github.com/aws/karpenter-core/pkg/metrics"
 	"github.com/aws/karpenter-core/pkg/scheduling"
 )
 
+const (
+	ProvisionerStaticallyDrifted cloudprovider.DriftReason = "ProvisionerStaticallyDrifted"
+	NodeRequirementDrift         cloudprovider.DriftReason = "NodeRequirementDrifted"
```
njtran, line 40: nit: good to keep the variable name the same
```suggestion
	NodeRequirementDrifted         cloudprovider.DriftReason = "NodeRequirementDrifted"
```

njtran (approved):

pkg/controllers/machine/disruption/drift.go
```diff
@@ -78,6 +86,14 @@ func (d *Drift) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisioner
 	})
 	if !hasDriftedCondition {
 		logging.FromContext(ctx).Debugf("marking machine as drifted")
+		metrics.MachinesDisruptedCounter.With(prometheus.Labels{
+			metrics.TypeLabel:        metrics.DriftReason,
+			metrics.ProvisionerLabel: machine.Labels[v1alpha5.ProvisionerNameLabelKey],
+		}).Inc()
+		metrics.MachinesDriftedCounter.With(prometheus.Labels{
+			metrics.TypeLabel:        string(driftedReason),
```
njtran, line 94: Shouldn't this be the same as in line 90?

jonathan-innis (approved):
LGTM 🚀

PR #443, opened 2023-08-04
chore: `v1beta1` Conversion Support for Provisioning/Deprovisioning/State

jonathan-innis:
Fixes #N/A 

**Description**

This PR starts staging out the changes for `v1beta1` to be supported across all controllers. It adds conversion for `v1alpha5.Provisioner`s into `v1beta1.NodePool`s so that all Provisioners and NodePools are treated the same in-memory. It stores whether the NodePool is a true NodePool or whether it is an in-memory representation of a NodePool through the `IsMachine` flag. 

This PR also edits the deprovisioning filtering logic to appropriately check the `karpenter.sh/do-not-consolidate` key to ensure the value of this annotation key on a node is set to `true` and can't pass the deprovisioning logic just because this value is set to some value other than `true`.

`make presubmit`
`/karpenter snapshot`


pkg/controllers/deprovisioning/singlemachineconsolidation.go
```diff
@@ -62,6 +65,7 @@ func (c *SingleMachineConsolidation) ComputeCommand(ctx context.Context, candida
 			continue
 		}
 
+		v := NewValidation(consolidationTTL(cmd.candidates), s.clock, s.cluster, s.kubeClient, s.provisioner, s.cloudProvider, s.recorder)
```
jonathan-innis, line 68: We have to create a new validation for each new set of candidates since that set of candidates may have a different max consolidationTTL

pkg/controllers/deprovisioning/consolidation.go
```diff
@@ -93,17 +90,17 @@ func (c *consolidation) markConsolidated() {
 }
 
 // ShouldDeprovision is a predicate used to filter deprovisionable nodes
-func (c *consolidation) ShouldDeprovision(_ context.Context, cn *Candidate) bool {
-	if val, ok := cn.Annotations()[v1alpha5.DoNotConsolidateNodeAnnotationKey]; ok {
-		c.recorder.Publish(deprovisioningevents.Unconsolidatable(cn.Node, cn.Machine, fmt.Sprintf("%s annotation exists", v1alpha5.DoNotConsolidateNodeAnnotationKey))...)
-		return val != "true"
+func (c *consolidation) shouldDeprovision(_ context.Context, cn *Candidate, isEmptyConsolidation bool) bool {
+	if cn.Annotations()[v1alpha5.DoNotConsolidateNodeAnnotationKey] == "true" {
```
njtran, line 94: Just musing: it would be nice to rename this to `safe-to-consolidate` similar to `safe-to-disrupt` when we move it over.

pkg/controllers/deprovisioning/consolidation.go
```diff
@@ -93,17 +90,17 @@ func (c *consolidation) markConsolidated() {
...
+		c.recorder.Publish(deprovisioningevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("%s annotation exists", v1alpha5.DoNotConsolidateNodeAnnotationKey))...)
+		return false
 	}
-	if cn.provisioner == nil {
-		c.recorder.Publish(deprovisioningevents.Unconsolidatable(cn.Node, cn.Machine, "provisioner is unknown")...)
+	if cn.nodePool.Spec.Deprovisioning.ConsolidationPolicy == v1beta1.ConsolidationPolicyNever {
+		c.recorder.Publish(deprovisioningevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("nodepool %s has all consolidation disabled by consolidation policy: %s", cn.nodePool.Name, cn.nodePool.Spec.Deprovisioning.ConsolidationPolicy))...)
```
njtran, line 99: nit: I feel like you probably just need the name here of the nodepool. IIUC, this will just read 
"node pool default has all consolidation disabled by consolidation policy: never", which I feel is a little redundant. Up to you though.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -485,3 +483,10 @@ func IsPastEmptinessTTL(nodeClaim *v1beta1.NodeClaim, clock clock.Clock, nodePoo
 	return nodeClaim.StatusConditions().GetCondition(v1beta1.NodeEmpty).IsTrue() &&
 		!clock.Now().Before(nodeClaim.StatusConditions().GetCondition(v1beta1.NodeEmpty).LastTransitionTime.Inner.Add(nodePool.Spec.Deprovisioning.EmptinessTTL.Duration))
 }
+
+func GetDriftedTime(nodeClaim *v1beta1.NodeClaim) time.Time {
```
njtran, line 487: You can correct me if I'm wrong, but this shouldn't be needed anymore, since we just use it for deprovisioning.

pkg/controllers/state/statenode.go
```diff
@@ -142,32 +154,37 @@ func (in *StateNode) Annotations() map[string]string {
...
 	// TODO @joinnis: The !in.Initialized() check can be removed when we can assume that all nodes have the v1alpha5.NodeRegisteredLabel on them
 	// We can assume that all nodes will have this label and no back-propagation will be required once we hit v1
 	if !in.Registered() && !in.Initialized() {
-		return in.Machine.Annotations
+		return in.NodeClaim.Annotations
 	}
 	return in.Node.Annotations
 }
 
+// GetLabels exists to conform to the GetLabels() function on the metav1.Object interface
```
njtran, line 170: Where are we using this as a metav1.Object out of curiosity? Hard to find where, if you can just point me to it.

pkg/controllers/state/statenode.go
```diff
@@ -108,17 +110,27 @@ func NewNode() *StateNode {
 	}
 }
 
+func (in *StateNode) OwnerKey() nodepoolutil.Key {
```
njtran, line 113: Does it make sense for a NodeClaim to have two of these labels? Would that break the logic at all? Seems like it wouldn't, and the NodePool Label would just be ignored, which if we can, we should make it the other way around.

pkg/controllers/state/cluster.go
```diff
@@ -522,7 +526,7 @@ func (c *Cluster) updateNodeUsageFromPod(ctx context.Context, pod *v1.Pod) error
 	n, ok := c.nodes[c.nameToProviderID[pod.Spec.NodeName]]
 	if !ok {
 		// the node must exist for us to update the resource requests on the node
-		return errors.NewNotFound(schema.GroupResource{Resource: "Machine"}, pod.Spec.NodeName)
+		return errors.NewNotFound(schema.GroupResource{Resource: "Node"}, pod.Spec.NodeName)
```
njtran, line 529: Shouldn't this be NodeClaim since that's what we're renaming it to be?

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -121,9 +122,9 @@ func benchmarkScheduler(b *testing.B, instanceCount, podCount int) {
 	instanceTypes := fake.InstanceTypes(instanceCount)
 	cloudProvider = fake.NewCloudProvider()
 	cloudProvider.InstanceTypes = instanceTypes
-	scheduler := scheduling.NewScheduler(ctx, nil, []*scheduling.MachineTemplate{scheduling.NewMachineTemplate(provisioner)},
+	scheduler := scheduling.NewScheduler(ctx, nil, []*scheduling.NodeClaimTemplate{scheduling.NewNodeClaimTemplate(nodepool.New(provisioner))},
 		nil, state.NewCluster(&clock.RealClock{}, nil, cloudProvider), nil, &scheduling.Topology{},
-		map[string][]*cloudprovider.InstanceType{provisioner.Name: instanceTypes}, nil,
+		map[nodepool.Key][]*cloudprovider.InstanceType{nodepool.Key{Name: provisioner.Name, IsProvisioner: true}: instanceTypes}, nil,
```
njtran, line 127: Just so I understand, when we migrate, this should be changed to False, right? Do you need a TODO here as well?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -309,21 +308,21 @@ func (s *Scheduler) calculateExistingMachines(stateNodes []*state.StateNode, dae
...
-	overhead := map[*MachineTemplate]v1.ResourceList{}
+func getDaemonOverhead(nodeClaimTemplates []*NodeClaimTemplate, daemonSetPods []*v1.Pod) map[*NodeClaimTemplate]v1.ResourceList {
+	overhead := map[*NodeClaimTemplate]v1.ResourceList{}
 
-	for _, nodeTemplate := range nodeTemplates {
+	for _, nodeClaimTemplate := range nodeClaimTemplates {
 		var daemons []*v1.Pod
 		for _, p := range daemonSetPods {
-			if err := nodeTemplate.Taints.Tolerates(p); err != nil {
+			if err := scheduling.Taints(nodeClaimTemplate.Spec.Taints).Tolerates(p); err != nil {
```
njtran, line 317: Do we need to make a `scheduling.Taints()` call here? Isn't the nodeClaimTemplate.Spec.Taints already supposed to be this type?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -61,33 +61,31 @@ func NewScheduler(ctx context.Context, kubeClient client.Client, machines []*Mac
...
 	ctx                context.Context
-	newMachines        []*Machine
+	newNodeClaims      []*NodeClaim
 	existingNodes      []*ExistingNode
-	machineTemplates   []*MachineTemplate
-	remainingResources map[string]v1.ResourceList // provisioner name -> remaining resources for that provisioner
-	instanceTypes      map[string][]*cloudprovider.InstanceType
-	daemonOverhead     map[*MachineTemplate]v1.ResourceList
+	nodeClaimTemplates []*NodeClaimTemplate
+	remainingResources map[nodepoolutil.Key]v1.ResourceList               // (nodePool name, isProvisioner) -> remaining resources for that provisioner
```
njtran, line 86: Probably can change both comments if you're gonna change one.
```suggestion
	remainingResources map[nodepoolutil.Key]v1.ResourceList               // (nodePool name, isProvisioner) -> remaining resources for that nodepool
```

pkg/controllers/deprovisioning/types.go
```diff
@@ -68,33 +72,42 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 	if !node.Initialized() {
 		return nil, fmt.Errorf("state node isn't initialized")
 	}
+	if _, ok := node.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok {
+		recorder.Publish(deprovisioningevents.Blocked(node.Node, node.NodeClaim, fmt.Sprintf("Disruption is blocked with the %q annotation", v1beta1.DoNotDisruptAnnotationKey))...)
+		return nil, fmt.Errorf("disruption is blocked thorugh the %q annotation", v1beta1.DoNotDisruptAnnotationKey)
```
njtran, line 77:
```suggestion
		return nil, fmt.Errorf("disruption is blocked through the %q annotation", v1beta1.DoNotDisruptAnnotationKey)
```

pkg/controllers/deprovisioning/singlemachineconsolidation.go
njtran: Will be nice to have this identified before this even gets to the deprovisioning controller so that we don't have to do this.

pkg/controllers/deprovisioning/helpers.go
```diff
@@ -177,42 +180,43 @@ func disruptionCost(ctx context.Context, pods []*v1.Pod) float64 {
...
 }
 
-// buildProvisionerMap builds a provName -> provisioner map and a provName -> instanceName -> instance type map
-func buildProvisionerMap(ctx context.Context, kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) (map[string]*v1alpha5.Provisioner, map[string]map[string]*cloudprovider.InstanceType, error) {
-	provisioners := map[string]*v1alpha5.Provisioner{}
-	var provList v1alpha5.ProvisionerList
-	if err := kubeClient.List(ctx, &provList); err != nil {
-		return nil, nil, fmt.Errorf("listing provisioners, %w", err)
+// buildNodePoolMap builds a provName -> nodePool map and a provName -> instanceName -> instance type map
+func buildNodePoolMap(ctx context.Context, kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) (map[nodepoolutil.Key]*v1beta1.NodePool, map[nodepoolutil.Key]map[string]*cloudprovider.InstanceType, error) {
```
njtran, line 198: how does this impact our memory footprint? I realize it'll be only one version, but IIRC, building the instance type map is pretty big.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -201,41 +205,45 @@ var ErrProvisionersNotFound = errors.New("no provisioners found")
...
 	// since they are stored within a slice and scheduling
 	// will always attempt to schedule on the first nodeTemplate
-	provisionerList.OrderByWeight()
-	for i := range provisionerList.Items {
-		provisioner := &provisionerList.Items[i]
+	nodePoolList.OrderByWeight()
+
+	for i := range nodePoolList.Items {
+		nodePool := &nodePoolList.Items[i]
+		if !nodePool.DeletionTimestamp.IsZero() {
```
njtran, line 230: Don't we already filter this out on line 216?

pkg/controllers/deprovisioning/types.go
```diff
@@ -119,9 +132,9 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 // disruption cost is highest, and it approaches zero as the node ages towards its expiration time.
 func (c *Candidate) lifetimeRemaining(clock clock.Clock) float64 {
 	remaining := 1.0
-	if c.provisioner.Spec.TTLSecondsUntilExpired != nil {
+	if c.nodePool.Spec.Deprovisioning.ExpirationTTL.Duration > 0 {
```
njtran, line 135: What if the policy is Never? Is that enforced in validation?

njtran (commented):
Looks great, don't have any complaints at a high level. Just some minor comments here and there.

pkg/controllers/deprovisioning/singlemachineconsolidation.go
jonathan-innis: How do you mean? This value is dependent on the subset of candidates that we are trying to consolidate so it could change for each individual bespoke set.

pkg/controllers/deprovisioning/consolidation.go
jonathan-innis: We aren't planning to maintain this label long-term. The idea was to remove this label in favor of `do-not-disrupt` to stop all voluntarily disruption (since that's the use-case). This is there for compat and should be pulled out long-term.

pkg/controllers/state/statenode.go
jonathan-innis: It's called in [`OwnerKey`](https://github.com/jonathan-innis/karpenter-core/blob/provisioner-v1beta1-support/pkg/utils/nodeclaim/nodeclaim.go#L460) so that we can collapse getting the `OwnerKey` from both `runtime.Objects` (e.g. `Machines` and `NodeClaims`) and the `state.StateNode`

pkg/controllers/state/statenode.go
jonathan-innis: Good point. I'll flip the ordering here.

pkg/controllers/state/cluster.go
jonathan-innis: No, this should have been `Node` originally but this was incorrectly put as `Machine` when this was originally merged.

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
jonathan-innis: I think this will naturally come when we add in the tests to validate the `NodePool`/`NodeClaim` path so I'm not sure that a TODO make sense here right now.

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: Yes, the `Tolerates` call only works with a `scheduling.Taint`. Prior to this change, we were storing taints in this state as a separate object inside of the `ProvisionerNodeTemplate`. Now, we are using the `NodeClaimTemplate` (wrapped) to store all the details, so taints are stored as `v1.Taint` which doesn't support this function and requires a cast.

pkg/controllers/deprovisioning/helpers.go
jonathan-innis: It doesn't change our footprint. We are currently building a `ProvisionerToInstanceType` map right now, which means that this is only a name change that doesn't tangibly change the current memory footprint.

pkg/controllers/provisioning/provisioner.go
jonathan-innis: Rebase miss. Thanks for the catch

pkg/controllers/deprovisioning/types.go
jonathan-innis: It's filtered in the `shouldDeprovision` function. AFAIK, the `lifetimeRemaining` function has no bearing on whether the candidate gets consolidated.

pkg/apis/v1beta1/nodepool.go
```diff
@@ -35,18 +37,18 @@ type NodePoolSpec struct {
 	// Deprovisioning contains the parameters that relate to Karpenter's deprovisioning logic
 	// +kubebuilder:default={"consolidationTTL": "15s","consolidationPolicy": "WhenUnderutilized", "expirationTTL": "90d"}
 	// +optional
-	Deprovisioning Deprovisioning `json:"deprovisioning"`
+	Deprovisioning Deprovisioning `json:"deprovisioning" hash:"ignore"`
```
ellistarn, line 40: Do we want to hash the full spec, or just hash the template, so you don't need all of these fields.

pkg/controllers/deprovisioning/consolidation.go
```diff
@@ -93,17 +91,17 @@ func (c *consolidation) markConsolidated() {
 }
 
 // ShouldDeprovision is a predicate used to filter deprovisionable nodes
-func (c *consolidation) ShouldDeprovision(_ context.Context, cn *Candidate) bool {
-	if val, ok := cn.Annotations()[v1alpha5.DoNotConsolidateNodeAnnotationKey]; ok {
-		c.recorder.Publish(deprovisioningevents.Unconsolidatable(cn.Node, cn.Machine, fmt.Sprintf("%s annotation exists", v1alpha5.DoNotConsolidateNodeAnnotationKey))...)
-		return val != "true"
+func (c *consolidation) shouldDeprovision(_ context.Context, cn *Candidate, isEmptyConsolidation bool) bool {
+	if cn.Annotations()[v1alpha5.DoNotConsolidateNodeAnnotationKey] == "true" {
```
ellistarn, line 95: We've changed this behavior, and will need a new test. Consider punting on this change for a future PR.

pkg/controllers/deprovisioning/controller.go
```diff
@@ -221,25 +216,25 @@ func (c *Controller) launchReplacementMachines(ctx context.Context, action Comma
 		return fmt.Errorf("cordoning nodes, %w", err)
 	}
 
-	machineNames, err := c.provisioner.LaunchMachines(ctx, action.replacements, provisioning.WithReason(reason))
+	nodeClaimKeys, err := c.provisioner.LaunchNodeClaims(ctx, action.replacements, provisioning.WithReason(reason))
```
ellistarn, line 219: Wonder if we should consider new language with the new semantic.
```suggestion
	nodeClaimKeys, err := c.provisioner.CreateNodeClaims(ctx, action.replacements, provisioning.WithReason(reason))
```

pkg/controllers/deprovisioning/emptymachineconsolidation.go
```diff
@@ -40,6 +40,10 @@ func NewEmptyMachineConsolidation(clk clock.Clock, cluster *state.Cluster, kubeC
 	return &EmptyMachineConsolidation{consolidation: makeConsolidation(clk, cluster, kubeClient, provisioner, cp, recorder)}
 }
 
+func (c *EmptyMachineConsolidation) ShouldDeprovision(ctx context.Context, cn *Candidate) bool {
+	return c.shouldDeprovision(ctx, cn, true)
+}
```
ellistarn, line 45: ?

pkg/utils/nodepool/nodepool.go
```diff
@@ -55,20 +55,18 @@ func New(provisioner *v1alpha5.Provisioner) *v1beta1.NodePool {
...
-	}
 	if provisioner.Spec.TTLSecondsUntilExpired != nil {
 		np.Spec.Deprovisioning.ExpirationTTL = metav1.Duration{Duration: lo.Must(time.ParseDuration(fmt.Sprintf("%ds", lo.FromPtr[int64](provisioner.Spec.TTLSecondsUntilExpired))))}
 	} else {
 		np.Spec.Deprovisioning.ExpirationTTL = metav1.Duration{Duration: -1}
 	}
-	if provisioner.Spec.Consolidation != nil {
+	if provisioner.Spec.Consolidation != nil && lo.FromPtr(provisioner.Spec.Consolidation.Enabled) {
 		np.Spec.Deprovisioning.ConsolidationPolicy = v1beta1.ConsolidationPolicyWhenUnderutilized
-		np.Spec.Deprovisioning.ConsolidationTTL = metav1.Duration{Duration: lo.Must(time.ParseDuration("15s"))}
```
njtran, line 68: I think we still need to discuss this right?

pkg/controllers/deprovisioning/suite_test.go
```diff
@@ -1838,6 +1838,230 @@ var _ = Describe("Topology Consideration", func() {
 	})
 })
 
+var _ = Describe("Empty Nodes (Consolidation)", func() {
```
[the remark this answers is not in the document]
jonathan-innis: Ah smart, I didn't notice that. I'll fix that and do a proper rebase.

pkg/utils/nodepool/nodepool.go
jonathan-innis: We do, but we can make this change after this PR if the thinking changes. For now, it's just conversion logic which means that it can be handled any way that we want.

pkg/controllers/deprovisioning/consolidation.go
jonathan-innis: I'm opting to call it out in the PR description. Frankly, this is the correct behavior to begin with so this change seems reasonable to me

jonathan-innis: I'll call it out in the PR description since this is technically the way that this behavior should have been previously.

njtran (approved):
LGTM

PR #444, opened 2023-08-07
fix: Add volume limit to StateNode from Machine

engedaam:
Fixes https://github.com/aws/karpenter/issues/4254

**Description**
- Currently karpenter is not taking into account the node volume limit when recoiling machines into the state node. 
- This results in consolidation not being able to determine volume limits that can block nodes from running on other pods.

- `make presubmit`
- Deployed a stateful into a cluster and scaled to a 100 replicas to find consolidation 
- karpenter e2e tests 
- Need to include more integration testing https://github.com/aws/karpenter/issues/4382

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -358,16 +358,20 @@ func (c *Cluster) newStateFromMachine(machine *v1alpha5.Machine, oldNode *StateN
 		oldNode = NewNode()
 	}
 	n := &StateNode{
-		Node:              oldNode.Node,
-		Machine:           machine,
-		hostPortUsage:     oldNode.hostPortUsage,
-		volumeUsage:       oldNode.volumeUsage,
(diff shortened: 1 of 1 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -362,6 +362,7 @@ func (c *Cluster) newStateFromMachine(machine *v1alpha5.Machine, oldNode *StateN
 		Machine:           machine,
 		hostPortUsage:     oldNode.hostPortUsage,
 		volumeUsage:       oldNode.volumeUsage,
+		volumeLimits:      oldNode.volumeLimits,
```
jonathan-innis, line 365: Can we make sure that we copy over the inflight details as well? There's a few details here that weren't correctly mapped over.

jonathan-innis (commented):
Can we make sure that we add a test case for the bug that you caught so we make sure that it doesn't happen again?

jonathan-innis:
Closing in favor of #434

PR #445, opened 2023-08-07
chore: Add `v1beta1/NodeClaim` controllers that convert from `v1alpha5/Machine`

jonathan-innis:
Fixes #N/A 

**Description**

This PR starts staging out the changes for `v1beta1` to be supported across all controllers. It adds conversion for `v1alpha5.Provisioner`s into `v1beta1.NodePool`s so that all Provisioners and NodePools are treated the same in-memory. It stores whether the NodePool is a true NodePool or whether it is an in-memory representation of a NodePool through the `IsMachine` flag. 

`make presubmit`
`/karpenter snapshot`


pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -58,47 +57,39 @@ type Controller struct {
...
 }
 
 // Reconcile executes a control loop for the resource
-func (c *Controller) Reconcile(ctx context.Context, machine *v1alpha5.Machine) (reconcile.Result, error) {
-	stored := machine.DeepCopy()
-	if _, ok := machine.Labels[v1alpha5.ProvisionerNameLabelKey]; !ok {
-		return reconcile.Result{}, nil
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	stored := nodeClaim.DeepCopy()
+	nodePool, err := nodeclaimutil.Owner(ctx, c.kubeClient, nodeClaim)
```
ellistarn, line 72: Does it make sense for us to isolate the compatibility logic into a package named `compat`, to help us remove things in the future.

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -108,6 +99,84 @@ func (c *Controller) Reconcile(ctx context.Context, machine *v1alpha5.Machine) (
...
+			nodeclaimutil.NodeEventHandler(ctx, c.kubeClient),
+		).
+		Watches(
+			&source.Kind{Type: &v1.Pod{}},
+			nodeclaimutil.PodEventHandler(ctx, c.kubeClient),
+		),
+	)
+}
+
+type MachineController struct {
```
ellistarn, line 162: Optional, parameterize the controller, in a single constructor e.g.

`NewController(v1alpha5.Machine{}, ...)`

Generics can remove a lot of this code, I think.

Looking a bit deeper, maybe this is the right approach, since your EventHandlers would then need to be parameterized.

pkg/controllers/nodeclaim/disruption/controller.go
jonathan-innis: This is a good point. I can make a separate package in a separate PR but changing that naming is probably beyond the scope of this PR

pkg/controllers/machine/disruption/controller.go
```diff
@@ -108,6 +99,84 @@ func (c *Controller) Reconcile(ctx context.Context, machine *v1alpha5.Machine) (
...
+func (c *NodeClaimController) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	return c.Controller.Reconcile(ctx, nodeClaim)
+}
+
+func (c *NodeClaimController) Builder(ctx context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.Adapt(controllerruntime.
+		NewControllerManagedBy(m).
+		For(&v1beta1.NodeClaim{}, builder.WithPredicates(
+			predicate.Or(
+				predicate.GenerationChangedPredicate{},
```
ellistarn, line 125: ?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,69 +28,71 @@ import (
...
-			logging.FromContext(ctx).Debugf("removing drift status condition from machine as drift has been disabled")
+			logging.FromContext(ctx).Debugf("removing drift status condition since drift has been disabled")
 		}
 		return reconcile.Result{}, nil
 	}
-	// 2. If Machine is not launched, remove the drift status condition
-	if launchCond := machine.StatusConditions().GetCondition(v1alpha5.MachineLaunched); launchCond == nil || launchCond.IsFalse() {
-		_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineDrifted)
+	// 2. If NodeClaim is not launched, remove the drift status condition
+	if !nodeClaim.StatusConditions().GetCondition(v1beta1.NodeLaunched).IsTrue() {
```
ellistarn, line 55: Was this just a simplify from line 53?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,69 +28,71 @@ import (
...
+		_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.NodeDrifted)
 		if hasDriftedCondition {
-			logging.FromContext(ctx).Debugf("removing drift status condition from machine as machine isn't launched")
+			logging.FromContext(ctx).Debugf("removing drift status condition since isn't launched")
```
ellistarn, line 58:
```suggestion
			logging.FromContext(ctx).Debugf("removing drift status condition since node isn't launched")
```

pkg/controllers/machine/disruption/emptiness.go
```diff
@@ -40,34 +40,34 @@ type Emptiness struct {
...
 	if err != nil {
-		// 3. If Machine -> Node mapping doesn't exist, remove the emptiness status condition
-		if machineutil.IsDuplicateNodeError(err) || machineutil.IsNodeNotFoundError(err) {
-			_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineEmpty)
+		// 3. If Node mapping doesn't exist, remove the emptiness status condition
+		if nodeclaimutil.IsDuplicateNodeError(err) || nodeclaimutil.IsNodeNotFoundError(err) {
+			_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.NodeEmpty)
 			if hasEmptyCondition {
-				logging.FromContext(ctx).Debugf("removing emptiness status condition from machine since machine doesn't have a single node mapping")
+				logging.FromContext(ctx).Debugf("removing emptiness status condition since doesn't have a single node mapping")
```
ellistarn, line 70: What does this log mean?

pkg/controllers/machine/disruption/emptiness.go
```diff
@@ -90,20 +90,20 @@ func (e *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisi
 	}
 	// 5. If there are pods that are actively scheduled to the Node, remove the emptiness status condition
 	if len(pods) > 0 {
-		_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineEmpty)
+		_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.NodeEmpty)
 		if hasEmptyCondition {
-			logging.FromContext(ctx).Debugf("removing emptiness status condition from node")
+			logging.FromContext(ctx).Debugf("removing emptiness status condition since not empty")
```
ellistarn, line 95: Maybe commas lead to more natural language than `since` everywhere
```suggestion
			logging.FromContext(ctx).Debugf("removing emptiness status condition, not empty")
```

pkg/controllers/machine/disruption/controller.go
jonathan-innis: This is there to avoid re-queueing on the machine/nodeClaim disruption status condition patch. So, we will only re-queue when a machine/nodeClaim status condition that affects its readiness changes OR the spec changes (which could affect drift). This is more of a performance concern than anything and doesn't tangibly affect the logic of the code.

pkg/controllers/machine/disruption/drift.go
jonathan-innis: Yeah, realized that the nil check with `IsFalse()` is unneccesary

pkg/controllers/machine/disruption/drift.go
jonathan-innis: Yeah, I think you're right that we should probably just not consider any events if we aren't past the launch stage.

pkg/controllers/machine/disruption/emptiness.go
jonathan-innis: Updated by making this a pre-requisite for the reconcile for emptiness.

pkg/controllers/machine/disruption/controller.go
ellistarn: Kinda wish this was the controllerruntime default. I should use it more :)

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,81 +28,83 @@ import (
...
-// Drift is a machine sub-controller that adds or removes status conditions on drifted machines
+// Drift is a nodeClaim sub-controller that adds or removes status conditions on drifted nodeClaims
 type Drift struct {
 	cloudProvider cloudprovider.CloudProvider
 }
 
-func (d *Drift) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisioner, machine *v1alpha5.Machine) (reconcile.Result, error) {
-	hasDriftedCondition := machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted) != nil
+func (d *Drift) Reconcile(ctx context.Context, nodePool *v1beta1.NodePool, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	if !nodeClaim.StatusConditions().GetCondition(v1beta1.NodeLaunched).IsTrue() {
```
ellistarn, line 43: Why change this logic now?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,81 +28,83 @@ import (
...
 	if !settings.FromContext(ctx).DriftEnabled {
-		_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineDrifted)
-		if hasDriftedCondition {
-			logging.FromContext(ctx).Debugf("removing drift status condition from machine as drift has been disabled")
-		}
-		return reconcile.Result{}, nil
-	}
-	// 2. If Machine is not launched, remove the drift status condition
-	if launchCond := machine.StatusConditions().GetCondition(v1alpha5.MachineLaunched); launchCond == nil || launchCond.IsFalse() {
-		_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineDrifted)
```
ellistarn, line 54: Can we punt to another PR?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,81 +28,83 @@ import (
...
+		_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.NodeDrifted)
 		if hasDriftedCondition {
-			logging.FromContext(ctx).Debugf("removing drift status condition from machine as machine isn't launched")
+			logging.FromContext(ctx).Debugf("removing drift status condition, drift has been disabled")
```
ellistarn, line 54: Don't we shortcircuit if drift is disabled? How can we get here?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,81 +28,83 @@ import (
...
 		return reconcile.Result{}, nil
 	}
-	driftedReason, err := d.isDrifted(ctx, provisioner, machine)
+	driftedReason, err := d.isDrifted(ctx, nodePool, nodeClaim)
 	if err != nil {
-		return reconcile.Result{}, cloudprovider.IgnoreMachineNotFoundError(fmt.Errorf("getting drift for machine, %w", err))
+		return reconcile.Result{}, cloudprovider.IgnoreMachineNotFoundError(fmt.Errorf("getting drift for nodeClaim, %w", err))
 	}
-	// 3. Otherwise, if the machine isn't drifted, but has the status condition, remove it.
+	// 2. Otherwise, if the nodeClaim isn't drifted, but has the status condition, remove it.
```
ellistarn, line 62: I used to number my comments. I stopped doing it because they kept getting out of sync.

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,81 +28,83 @@ import (
...
+	driftedReason, err := d.cloudProvider.IsMachineDrifted(ctx, machineutil.NewFromNodeClaim(nodeClaim))
 	if err != nil {
 		return "", err
 	}
-
-	reason := lo.FindOrElse([]cloudprovider.DriftReason{driftedReason, areStaticFieldsDrifted(provisioner, machine), areNodeRequirementsDrifted(provisioner, machine)}, "", func(i cloudprovider.DriftReason) bool {
-		return i != ""
-	})
-
+	reason := lo.FindOrElse([]cloudprovider.DriftReason{
```
ellistarn, line 91: Avoid these touchups if you can.

pkg/controllers/machine/disruption/drift_test.go
```diff
@@ -83,9 +83,9 @@ var _ = Describe("Drift", func() {
 		machine = ExpectExists(ctx, env.Client, machine)
 		Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted)).To(BeNil())
 	})
-	It("should remove the status condition from the machine when the machine launch condition is false", func() {
+	It("should not reconcile when the machine launch condition is false", func() {
+		// Machine would normally be marked as Drifted but isn't here since it isn't Launched yet
```
ellistarn, line 87: Does this comment help future readers, or just this PR review? :)

pkg/controllers/machine/disruption/emptiness.go
```diff
@@ -90,20 +77,20 @@ func (e *Emptiness) Reconcile(ctx context.Context, provisioner *v1alpha5.Provisi
 	}
 	// 5. If there are pods that are actively scheduled to the Node, remove the emptiness status condition
 	if len(pods) > 0 {
-		_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineEmpty)
+		_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.NodeEmpty)
 		if hasEmptyCondition {
-			logging.FromContext(ctx).Debugf("removing emptiness status condition from node")
+			logging.FromContext(ctx).Debugf("removing emptiness status condition, not empty")
```
ellistarn, line 82: wonder if these log lines should be `empty`, `not empty`, and let everything else just be context logging.

pkg/controllers/machine/disruption/emptiness.go
```diff
@@ -40,47 +40,34 @@ type Emptiness struct {
...
-	if provisioner.Spec.TTLSecondsAfterEmpty == nil {
-		if hasEmptyCondition {
-			_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineEmpty)
-			logging.FromContext(ctx).Debugf("removing emptiness status condition from machine since emptiness is disabled")
-		}
-		return reconcile.Result{}, nil
-	}
-	// 2. If Machine is not initialized, remove the emptiness status condition
-	if initCond := machine.StatusConditions().GetCondition(v1alpha5.MachineInitialized); initCond == nil || initCond.IsFalse() {
+	if nodePool.Spec.Deprovisioning.EmptinessTTL == nil {
```
ellistarn, line 56: Let's take another look at this `Deprovisioning` subkey before we launch.

pkg/controllers/machine/disruption/emptiness.go
```diff
@@ -40,47 +40,34 @@ type Emptiness struct {
...
+func (e *Emptiness) Reconcile(ctx context.Context, nodePool *v1beta1.NodePool, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	if !nodeClaim.StatusConditions().GetCondition(v1beta1.NodeInitialized).IsTrue() {
+		return reconcile.Result{}, nil
+	}
+	n, err := nodeclaimutil.NodeForNodeClaim(ctx, e.kubeClient, nodeClaim)
+	if err != nil {
+		return reconcile.Result{}, nodeclaimutil.IgnoreNodeNotFoundError(err)
+	}
+
+	hasEmptyCondition := nodeClaim.StatusConditions().GetCondition(v1beta1.NodeEmpty) != nil
```
ellistarn, line 52: Is there a way to make this a straight rename? Hard for me to tell if this is still right.

pkg/controllers/machine/disruption/emptiness_test.go
```diff
@@ -71,8 +71,8 @@ var _ = Describe("Emptiness", func() {
 		machine = ExpectExists(ctx, env.Client, machine)
 		Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineEmpty)).To(BeNil())
 	})
-	It("should remove the status condition from the machine when the machine initialization condition is false", func() {
-		machine.StatusConditions().MarkTrue(v1alpha5.MachineEmpty)
+	It("should not reconcile the machine when the machine initialization condition is false", func() {
+		// Machine would normally be marked as Empty but isn't here since it isn't initialized
```
ellistarn, line 75: These comments are super confusing

pkg/controllers/machine/disruption/emptiness.go
ellistarn: Ah -- this is one of the unavoidable semantic differences where we have to support both.

ellistarn (changes requested):
Ok -- I made it through. Firstly, this is a lot of work, so thanks for the effort on this so far. 

I'm a bit nervous about this PR, and the approach in general. Migrations like this are a hard problem, and you do have a track record of getting them right. As I was reviewing, I felt like I was running a fuzzy match algorithm, trying to see what was just a rename, and what was just a semantic difference. I'm confident that if you missed something, I did too. This is exacerbated by the fact that the tests now only test the node claim code path, and no longer test the machine code path.

I wonder if it would be safer for us to do a hard fork of the Machine controller code, and maintain parallel implementations for machine and node claim. You could choose to make any semantic changes in the NodeClaim code, or do them in a followon. 

This has a few benefits:
* We remove any risk that this change impacts existing customers using the v1alpha5 APIs. Impact is contained to the new API, and rollback from NodePool -> Provisioner without re-releasing karpenter is possible.
* The testing problem you have goes away, you simply maintain parallel tests
* You no longer need to maintain the complexity of the compatibility library you've built
* It's much easier to reason about the controller code as we build new features, and the implementations diverge

And a few drawbacks:
* It gets twice as expensive to maintain any changes to the v1alpha5 Machines -- it essentially becomes a backporting effort, including unit/e2e/manual tests. I suspect we might be able to get away with a minimal amount of backporting.
* We will have to be vigilant to explain the codebase to newcomers, who may find the state of the codebase confusing during the migration.

pkg/controllers/machine/disruption/drift.go
jonathan-innis: We do not short-circuit in this case. Drift being disabled causes us to remove the conditions. I believe this is the correct handling here.

pkg/controllers/machine/disruption/drift.go
jonathan-innis: Undone this change

pkg/controllers/machine/disruption/drift.go
jonathan-innis: Making a note, will take care of it in another PR

pkg/controllers/machine/disruption/controller.go
```diff
@@ -58,47 +57,42 @@ type Controller struct {
...
-	if _, ok := machine.Labels[v1alpha5.ProvisionerNameLabelKey]; !ok {
-		return reconcile.Result{}, nil
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	stored := nodeClaim.DeepCopy()
+	nodePool, err := nodeclaimutil.Owner(ctx, c.kubeClient, nodeClaim)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
-	if !machine.DeletionTimestamp.IsZero() {
+	if nodePool == nil {
```
ellistarn, line 76: How can owner be nil? Shouldn't the above method have 404ed?

pkg/controllers/machine/disruption/controller.go
```diff
@@ -58,47 +57,42 @@ type Controller struct {
...
 		return reconcile.Result{}, nil
 	}
-	provisioner := &v1alpha5.Provisioner{}
-	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: machine.Labels[v1alpha5.ProvisionerNameLabelKey]}, provisioner); err != nil {
-		return reconcile.Result{}, client.IgnoreNotFound(err)
+	if !nodeClaim.DeletionTimestamp.IsZero() {
```
ellistarn, line 79: Should we push this above retrieving the owner?

pkg/controllers/machine/disruption/drift.go
```diff
@@ -28,70 +28,72 @@ import (
...
 	}
-	// 2. If Machine is not launched, remove the drift status condition
-	if launchCond := machine.StatusConditions().GetCondition(v1alpha5.MachineLaunched); launchCond == nil || launchCond.IsFalse() {
-		_ = machine.StatusConditions().ClearCondition(v1alpha5.MachineDrifted)
+	// 2. If NodeClaim is not launched, remove the drift status condition
+	if launchCond := nodeClaim.StatusConditions().GetCondition(v1beta1.NodeLaunched); launchCond == nil || launchCond.IsFalse() {
+		_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.NodeDrifted)
 		if hasDriftedCondition {
-			logging.FromContext(ctx).Debugf("removing drift status condition from machine as machine isn't launched")
+			logging.FromContext(ctx).Debugf("removing drift status condition, isn't launched")
```
ellistarn, line 58: Can you take a note to look at this one? Seems like this case should never happen, and is potentially just defensive programming

pkg/controllers/machine/termination/controller.go
```diff
@@ -82,26 +83,101 @@ func (c *Controller) Finalize(ctx context.Context, machine *v1alpha5.Machine) (r
...
-	if !equality.Semantic.DeepEqual(stored, machine) {
-		if err := c.kubeClient.Patch(ctx, machine, client.MergeFrom(stored)); err != nil {
-			return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("removing machine termination finalizer, %w", err))
+	controllerutil.RemoveFinalizer(nodeClaim, v1beta1.TerminationFinalizer)
+	if !equality.Semantic.DeepEqual(stored, nodeClaim) {
+		if err = nodeclaimutil.Patch(ctx, c.kubeClient, stored, nodeClaim); err != nil {
+			return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("removing termination finalizer, %w", err))
 		}
-		logging.FromContext(ctx).Infof("deleted machine")
+		logging.FromContext(ctx).Infof("deleted %s", lo.Ternary(nodeClaim.IsMachine, "machine", "nodeclaim"))
```
ellistarn, line 100:
```suggestion
		logging.FromContext(ctx).Infof("deleted"))
```

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -474,7 +478,7 @@ func IsExpired(obj client.Object, clock clock.Clock, nodePool *v1beta1.NodePool)
 }
 
 func GetExpirationTime(obj client.Object, nodePool *v1beta1.NodePool) time.Time {
-	if nodePool == nil || nodePool.Spec.Deprovisioning.ExpirationTTL.Duration >= 0 || obj == nil {
+	if nodePool == nil || nodePool.Spec.Deprovisioning.ExpirationTTL.Duration < 0 || obj == nil {
```
ellistarn, line 481: ?

jonathan-innis: Removed this function and are doing the check in the the expiration disruption directly

ellistarn (approved):

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -468,20 +492,3 @@ func OwnerName(obj interface{ GetLabels() map[string]string }) string {
...
-	if nodePool == nil || nodePool.Spec.Deprovisioning.ExpirationTTL.Duration >= 0 || obj == nil {
-		// If not defined, return some much larger time.
-		return time.Date(5000, 0, 0, 0, 0, 0, 0, time.UTC)
-	}
-	return obj.GetCreationTimestamp().Add(nodePool.Spec.Deprovisioning.ExpirationTTL.Duration)
-}
-
-func IsPastEmptinessTTL(nodeClaim *v1beta1.NodeClaim, clock clock.Clock, nodePool *v1beta1.NodePool) bool {
-	return nodeClaim.StatusConditions().GetCondition(v1beta1.NodeEmpty).IsTrue() &&
-		!clock.Now().Before(nodeClaim.StatusConditions().GetCondition(v1beta1.NodeEmpty).LastTransitionTime.Inner.Add(nodePool.Spec.Deprovisioning.EmptinessTTL.Duration))
```
njtran, line 486: If it's removed from the nodeclaim and machine utils are we fully removing? Or is it somewhere else?

njtran (commented):
Looks good, just some questions and nits.

pkg/controllers/machine/disruption/drift.go
```diff
@@ -45,71 +46,65 @@ type Drift struct {
...
 	// First check for static drift or node requirements have drifted to save on API calls.
-	if reason := lo.FindOrElse([]cloudprovider.DriftReason{areStaticFieldsDrifted(provisioner, machine), areNodeRequirementsDrifted(provisioner, machine)}, "", func(i cloudprovider.DriftReason) bool {
+	if reason := lo.FindOrElse([]cloudprovider.DriftReason{areStaticFieldsDrifted(nodePool, nodeClaim), areNodeRequirementsDrifted(nodePool, nodeClaim)}, "", func(i cloudprovider.DriftReason) bool {
 		return i != ""
 	}); reason != "" {
 		return reason, nil
 	}
 
-	driftedReason, err := d.cloudProvider.IsMachineDrifted(ctx, machine)
+	driftedReason, err := d.cloudProvider.IsMachineDrifted(ctx, machineutil.NewFromNodeClaim(nodeClaim))
```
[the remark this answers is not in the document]
jonathan-innis: I don't want to touch the CloudProvider interface in this PR. Planning to change that in a follow-up PR where I make the relevant changes in this package as well as in upstream `aws/karpenter`

pkg/utils/nodeclaim/nodeclaim.go
jonathan-innis: Yes, I'll fully remove this.

pkg/controllers/machine/garbagecollection/controller.go
```diff
@@ -68,26 +68,30 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
...
-			metrics.ReasonLabel:      "garbage_collected",
-			metrics.ProvisionerLabel: machines[i].Labels[v1alpha5.ProvisionerNameLabelKey],
-		}).Inc()
+			With(
+				lo.Ternary(nodeClaims[i].IsMachine, "machine", "nodeclaim"),
+				nodeClaims[i].Name,
+				"provider-id",
+				nodeClaims[i].Status.ProviderID,
+				lo.Ternary(nodeClaims[i].IsMachine, "provisioner", "nodepool"),
+				nodeclaimutil.OwnerName(nodeClaims[i]),
```
njtran, line 91: nit: personally think it'd be more readable to have this in 3 lines for the 3 context values, but up to you.
```suggestion
				lo.Ternary(nodeClaims[i].IsMachine, "machine", "nodeclaim"), nodeClaims[i].Name,
				"provider-id", nodeClaims[i].Status.ProviderID,
				lo.Ternary(nodeClaims[i].IsMachine, "provisioner", "nodepool"), nodeclaimutil.OwnerName(nodeClaims[i]),
```

njtran (dismissed):
LGTM, one nit

pkg/controllers/machine/garbagecollection/controller.go
jonathan-innis: Agreed, good point.

njtran (approved):

PR #447, opened 2023-08-07
chore: Enable `v1beta1/NodeClaim` conversion in consistency controller

jonathan-innis:
Fixes #N/A 

**Description**

This PR starts staging out the changes for `v1beta1` to be supported across all controllers. It adds conversion for `v1alpha5.Machine`s into `v1beta1.NodeClaim`s so that all Machines and NodeClaims are treated the same in-memory for the Consistency Controller to update the status for the `NodeClaim` and the `Machine`

`make presubmit`
`make apply`

### DIFF
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -70,7 +70,7 @@ func NewControllers(
 		metricspod.NewController(kubeClient),
 		metricsprovisioner.NewController(kubeClient),
 		counter.NewProvisionerController(kubeClient, cluster),
-		consistency.NewController(clock, kubeClient, recorder, cloudProvider),
+		consistency.NewMachineController(clock, kubeClient, recorder, cloudProvider),
 		nodeclaimlifecycle.NewMachineController(clock, kubeClient, cloudProvider, recorder),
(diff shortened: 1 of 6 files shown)


pkg/controllers/controllers.go
```diff
@@ -70,7 +70,7 @@ func NewControllers(
 		metricspod.NewController(kubeClient),
 		metricsprovisioner.NewController(kubeClient),
 		counter.NewController(kubeClient, cluster),
-		consistency.NewController(clock, kubeClient, recorder, cloudProvider),
+		consistency.NewMachineController(clock, kubeClient, recorder, cloudProvider),
```
engedaam, line 73: Why are we not using `NewNodeClaimController`?

jonathan-innis: This would enforce that we have `NodeClaims` that are deployed and are shipped as part of the chart. That'll happen when we are ready to release the changes all together. For now, we are just staging out the changes and having it work with the older API version.

engedaam (dismissed):

engedaam (approved):

PR #448, opened 2023-08-08
chore: Enable `v1beta1/NodePool` conversion in counter controller

jonathan-innis:
Fixes #N/A 

**Description**

This PR starts staging out the changes for `v1beta1` to be supported across all controllers. It adds conversion for `v1alpha5.Provisioner`s into `v1beta1.NodePool`s so that all Provisioners and NodePools are treated the same in-memory for the Counter controller to update the status for the `NodePool` and the `Provisioner`.

`make presubmit`
`make apply`

### DIFF
--- a/pkg/controllers/counter/controller.go
+++ b/pkg/controllers/counter/controller.go
@@ -18,13 +18,16 @@ import (
 	"context"
 	"time"
 
+	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/apis/v1beta1"
 	"github.com/aws/karpenter-core/pkg/controllers/state"
 	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
 	"github.com/aws/karpenter-core/pkg/utils/functional"
+	nodepoolutil "github.com/aws/karpenter-core/pkg/utils/nodepool"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/controllers/counter/controller.go
```diff
@@ -39,46 +42,40 @@ import (
...
-var _ corecontroller.TypedController[*v1alpha5.Provisioner] = (*Controller)(nil)
-
 // Controller for the resource
 type Controller struct {
 	kubeClient client.Client
 	cluster    *state.Cluster
 }
 
 // NewController is a constructor
-func NewController(kubeClient client.Client, cluster *state.Cluster) corecontroller.Controller {
```
engedaam, line 51: What do you think about combining this controller with the provisioner controller?

pkg/controllers/counter/controller.go
```diff
@@ -89,15 +86,67 @@ func (c *Controller) resourceCountsFor(provisionerName string) v1.ResourceList {
...
 	})
 	return functional.FilterMap(res, func(_ v1.ResourceName, v resource.Quantity) bool { return !v.IsZero() })
 }
 
-func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+type NodePoolController struct {
+	*Controller
+}
+
+func NewNodePoolController(kubeClient client.Client, cluster *state.Cluster) corecontroller.Controller {
```
engedaam, line 101: Do we need to update the static drift annotation for nodepools?

pkg/controllers/counter/controller.go
jonathan-innis: We _should_ do that but I think that change should be in a separate PR

pkg/controllers/counter/controller.go
jonathan-innis: Yep, we need to update the drift annotation. We'll have that change in a separate PR

engedaam (approved):

PR #449, opened 2023-08-08
fix: add pod overhead into pod resources calculation

tzneal:
Fixes https://github.com/aws/karpenter/issues/4392

**Description**

add pod overhead into pod resources calculation

Unit testing and with the example pod in the issue.

Before change

```
karpenter-58d6968796-6mnj6 controller 2023-08-08T13:38:36.847Z	INFO	controller.provisioner	created machine	{"commit": "3f1bce4", "provisioner": "default", "requests": {"cpu":"185m","memory":"240Mi","pods":"5"}, "instance-types": "c3.2xlarge, c3.large, c3.xlarge, c4.2xlarge, c4.large and 95 other(s)"}
```

After change

```

karpenter-9d8787b9c-5pbpl controller 2023-08-08T13:43:24.686Z	ERROR	controller.provisioner	Could not schedule pod, incompatible with provisioner "default", daemonset overhead={"cpu":"185m","memory":"240Mi","pods":"4"}, no instance type satisfied resources {"cpu":"24185m","memory":"240Mi","pods":"5"} and requirements karpenter.k8s.aws/instance-category In [c m r], karpenter.k8s.aws/instance-generation Exists >2, karpenter.k8s.aws/instance-size In [2xlarge 4xlarge large medium xlarge], karpenter.sh/capacity-type In [on-demand], karpenter.sh/provisioner-name In [default], kubernetes.io/arch In [amd64] (no instance type which had enough resources and the required offering met the scheduling requirements)	{"commit": "3f1bce4", "pod": "default/trouble-5bd94c6d68-qktjc"}

```

### DIFF
--- a/pkg/utils/resources/resources.go
+++ b/pkg/utils/resources/resources.go
@@ -97,12 +97,15 @@ func Subtract(lhs, rhs v1.ResourceList) v1.ResourceList {
 func Ceiling(pod *v1.Pod) v1.ResourceRequirements {
 	var resources v1.ResourceRequirements
 	for _, container := range pod.Spec.Containers {
-		resources.Requests = Merge(resources.Requests, MergeResourceLimitsIntoRequests(container))
-		resources.Limits = Merge(resources.Limits, container.Resources.Limits)
+		resources.Requests = MergeInto(resources.Requests, MergeResourceLimitsIntoRequests(container))
+		resources.Limits = MergeInto(resources.Limits, container.Resources.Limits)
 	}
 	for _, container := range pod.Spec.InitContainers {
 		resources.Requests = MaxResources(resources.Requests, MergeResourceLimitsIntoRequests(container))
 		resources.Limits = MaxResources(resources.Limits, container.Resources.Limits)
+	}
+	if pod.Spec.Overhead != nil {
+		resources.Requests = MergeInto(resources.Requests, pod.Spec.Overhead)
 	}
 	return resources
 }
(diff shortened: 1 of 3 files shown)


pkg/utils/resources/resources.go
```diff
@@ -97,13 +97,16 @@ func Subtract(lhs, rhs v1.ResourceList) v1.ResourceList {
...
-		resources.Limits = Merge(resources.Limits, container.Resources.Limits)
+		resources.Requests = MergeInto(resources.Requests, MergeResourceLimitsIntoRequests(container))
+		resources.Limits = MergeInto(resources.Limits, container.Resources.Limits)
 	}
 	for _, container := range pod.Spec.InitContainers {
 		resources.Requests = MaxResources(resources.Requests, MergeResourceLimitsIntoRequests(container))
 		resources.Limits = MaxResources(resources.Limits, container.Resources.Limits)
 	}
+	if pod.Spec.Overhead != nil {
+		resources.Requests = MergeInto(resources.Requests, pod.Spec.Overhead)
```
jonathan-innis, line 108: Do we need the `nil` check or can we always merge the overhead into the requests even if it's `nil`?

jonathan-innis (approved):
LGTM, one question/nit

pkg/utils/resources/resources.go
tzneal: Yeah, we could skip it and it would still work. I like the explicit check, you don't ever need to worry about MergeInto changing in some way that makes it no longer valid to pass a nil, but it could go either way.

PR #450, opened 2023-08-08
fix: resolve issue interpreting preferred node affinities

tzneal:
Fixes #N/A 

**Description**

We were treating preferred node affinities as required for topology spread purposes. This would cause us to incorrectly restrict the pod's domains to only ones that were acceptable per the preference.  In some cases, this can lead to pods remaining pending as we assume a pod will schedule to an existing node, but it won't due to a required topology spread and our incorrect restriction.

Unit test


ellistarn (approved):

jonathan-innis (approved):
LGTM 🚀

PR #456, opened 2023-08-09
chore: debug log when we add requirements due to volumes

tzneal:
Fixes #N/A 

**Description**

These can affect scheduling decisions and make it difficult to debug why pods aren't scheduling.  With this change, the new requirements are logged and it should become easier to diagnose incompatibility between volumes and pod requirements. Currently this creates an empty set requirement and the pod just fails to schedule.

Unit testing and observed the new log.

```
DEBUG	scheduling/volumetopology.go:74	adding requirements derived from pod volumes, [{topology.kubernetes.io/zone In [test-zone-2 test-zone-3]}]	{"pod": "default/facegarnet-49-9jvxmqcv0v"}
```


thomaschaaf:
Wow great that you added this so quickly!

PR #458, opened 2023-08-09
fix: ignore permanently pending pods during consolidation

### LINKED ISSUE #457: Karpenter is unable to consolidate when a pod is permanently stuck in Pending
### Description

**Observed Behavior**:
When a pod is stuck permanently in Pending, (pod will never match any node due to some error), Karpenter is unable to consolidate any nodes. This seems to be because during simulation https://github.com/aws/karpenter-core/blob/dca87d6ed6e8b4ed2459458508c2efa9287aa183/pkg/controllers/deprovisioning/helpers.go#L69C20-L69C20 when it performs getPendingPods, https://github.com/aws/karpenter-core/blob/dca87d6ed6e8b4ed2459458508c2efa9287aa183/pkg/controllers/provisioning/provisioner.go#L154, to determine if all pods will be able land in all nodes minus the nodes being consolidated the permanently stuck pod will cause the simulation to fail 

**Expected Behavior**:
I don't believe we should get stuck on consolidation if there is an erroneous pending pod.

Introduce a pod with node selector that doesn't match any nodes. Attempt to drain pods from cluster such that Karpenter would consolidate. 

**Versions**:
Karpenter v0.29.2

tzneal:
Fixes #457

**Description**

We still consider pending pods for scheduling calculation purposes, but if a pod was pending before scheduling, we don't let the fact that we can't schedule it prevent consolidation. This ensures that permanently pending pods won't block consolidation.

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -32,6 +32,7 @@ import (
 	"github.com/aws/karpenter-core/pkg/controllers/state"
 	"github.com/aws/karpenter-core/pkg/events"
 	"github.com/aws/karpenter-core/pkg/scheduling"
+	"github.com/aws/karpenter-core/pkg/utils/pod"
 	"github.com/aws/karpenter-core/pkg/utils/resources"
 )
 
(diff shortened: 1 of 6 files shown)


njtran (approved):

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -102,8 +103,16 @@ type Results struct {
...
 }
 
-func (r Results) AllPodsScheduled() bool {
-	return len(r.PodErrors) == 0
+// AllNonPendingPodsScheduled returns true if all of the non-pending pods scheduled.  This is useful in consolidation as
+// we don't care if a pod was pending before consolidation and will still be pending after. It may be a pod that we
+// can't schedule at all and don't want it to block consolidation.
+func (r Results) AllNonPendingPodsScheduled() bool {
+	for p := range r.PodErrors {
+		if !pod.IsProvisionable(p) {
```
ellistarn, line 111: Does this go far enough? What if a pod can't be scheduled due to an impossible affinity rule?

PR #460, opened 2023-08-10
test: Deflake tests that validate deprovisioning ordering

jonathan-innis:
Fixes #N/A 

**Description**

This PR deflakes two test cases that deal with the ordering of nodes by specifically setting a status condition to a later time rather than relying on the `.StatusConditions().MarkTrue()` mechanism which relies on `time.Now()` and was prone to race conditions.

`make deflake`

### DIFF
--- a/pkg/controllers/deprovisioning/drift.go
+++ b/pkg/controllers/deprovisioning/drift.go
@@ -32,7 +32,6 @@ import (
 	"github.com/aws/karpenter-core/pkg/controllers/state"
 	"github.com/aws/karpenter-core/pkg/events"
 	"github.com/aws/karpenter-core/pkg/metrics"
-	machineutil "github.com/aws/karpenter-core/pkg/utils/machine"
 )
 
 // Drift is a subreconciler that deletes drifted machines.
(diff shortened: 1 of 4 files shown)


pkg/controllers/deprovisioning/drift.go
```diff
@@ -66,7 +65,8 @@ func (d *Drift) filterAndSortCandidates(ctx context.Context, nodes []*Candidate)
 		return nil, fmt.Errorf("filtering candidates, %w", err)
 	}
 	sort.Slice(candidates, func(i int, j int) bool {
-		return machineutil.GetDriftedTime(candidates[i].Machine).Before(machineutil.GetDriftedTime(candidates[j].Machine))
+		return candidates[i].Machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).LastTransitionTime.Inner.Time.Before(
```
njtran, line 68: Isn't this the same as what it currently is? 

https://github.com/aws/karpenter-core/blob/main/pkg/utils/machine/machine.go#L321-L326

njtran (approved):
Nice work. One comment

ellistarn (approved):

pkg/controllers/deprovisioning/drift.go
jonathan-innis: It is, I just didn't see the need for the separate utils func since we know that the status condition is set at this point.

PR #462, opened 2023-08-10
chore: Stop firing `DeprovisioningBlocked` until initialized

jonathan-innis:
Fixes #N/A 

**Description**

Deprovisioning events were spamming event messages by firing `DeprovisioningBlocked` (and some `Unconsolidatable`) events right around that a Machine/Node is starting up. These events are not particularly useful at this point in time (since these events are mainly there for debugging and there to inform the user why Karpenter has not made a decision against a specific Node).  Karpenter should not begin firing these events until the Machine/Node has been initialized. At this point, we can assume that the Node is in a steady state and begin adding debug events if Karpenter cannot take an action against it.

### Before PR

```console
➜  karpenter git:(main) k get events | grep DeprovisioningBlocked | wc -l           
    2000
```

### After PR

```console
➜  karpenter git:(main) k get events | grep DeprovisioningBlocked | wc -l           
    0
```

`make presubmit`
`make apply`

### DIFF
--- a/pkg/controllers/deprovisioning/events/events.go
+++ b/pkg/controllers/deprovisioning/events/events.go
@@ -75,6 +75,8 @@ func Terminating(node *v1.Node, machine *v1alpha5.Machine, reason string) []even
 	}
 }
 
+// Unconsolidatable is an event that informs the user that a Machine/Node combination cannot be consolidated
+// due to the state of the Machine/Node or due to some state of the pods that are scheduled to the Machine/Node
 func Unconsolidatable(node *v1.Node, machine *v1alpha5.Machine, reason string) []events.Event {
 	return []events.Event{
 		{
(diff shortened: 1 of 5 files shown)


pkg/controllers/deprovisioning/events/events.go
```diff
@@ -73,7 +73,15 @@ func Terminating(node *v1.Node, machine *v1alpha5.Machine, reason string) []even
 	}
 }
 
-func Unconsolidatable(node *v1.Node, machine *v1alpha5.Machine, reason string) []events.Event {
+// Unconsolidatable is an event that informs the user that a Machine/Node combination cannot be consolidated
+// due to the state of the Machine/Node or due to some state of the pods that are scheduled to the Machine/Node
+// Unconsolidatable does not begin respecting events that are requested for it until either the Machine or Node has existed
+// for more than 30 minutes. This is to prevent the Machine/Node from spamming "Unconsolidatable" messages during startup
```
ellistarn, line 79: Can we use node initialization, instead? 30m seems like an odd choice, since it doesn't relate to our algorithm at all.

jonathan-innis: The reason we _weren't_ doing this was surfacing the fact that a lack of initialization could also block deprovisioning from occurring. I wonder if we use a value similar to our consistency timeout to track the timeout when we start to consider blocked events. The alternative is that we don't show Deprovisioning blocked when the node isn't initialized or we create a very specific timeout for that specific scenario.

jonathan-innis: Discussed offline: came to the conclusion that since initialization is now a surfaced condition through the Machine status that we are fine to wait for initialization before firing these events.

ellistarn (approved):

PR #464, opened 2023-08-10
fix: reorder drift checks to save on computation

njtran:
Fixes #N/A 

**Description**
We checked cloud provider drift before provisioner neutral drift, resulting in always making an cloudprovider.IsMachineDrifted() and subequent API calls for each reconciliation loop. We shouldn't need to make those calls if the Provisioner is drifted. 

- make presubmit

### DIFF
--- a/pkg/controllers/machine/disruption/drift_test.go
+++ b/pkg/controllers/machine/disruption/drift_test.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/aws/karpenter-core/pkg/apis/settings"
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/controllers/machine/disruption"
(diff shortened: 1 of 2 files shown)


pkg/controllers/machine/disruption/drift_test.go
```diff
@@ -63,6 +64,33 @@ var _ = Describe("Drift", func() {
 		machine = ExpectExists(ctx, env.Client, machine)
 		Expect(machine.StatusConditions().GetCondition(v1alpha5.MachineDrifted).IsTrue()).To(BeTrue())
 	})
+	It("should detect static drift before cloud provider drift", func() {
+		cp.Drifted = "drifted"
+		provisioner.Annotations = lo.Assign(provisioner.Annotations, map[string]string{
```
engedaam, line 69: Can we use the `provisionerController` for setting the hash?

jonathan-innis (approved):
LGTM 🚀

engedaam (approved):

PR #465, opened 2023-08-12
feat: Respect DeletionTimestamp and DeletionGracePeriodSeconds for Node, Machine, and Pods

ellistarn:
Fixes #NA

**Description**

I started this PR by experimenting with cascading delete of CRDs. For example, if I uninstall Karpenter CRDs before uninstalling Karpenter, will Karpenter successfully clean up all underlying instances before allowing the CRD to be deleted? I expected something along the following lines:

1. `k delete customresourcedefinitions.apiextensions.k8s.io machines.karpenter.sh`
2. CRD hangs in the CRD finalizer, since machine CRs exist
3. Machine deletion triggers cordon/drain on the node
4. Pods are successfully drained from the node
5. The machine is deleted, and thus the CRD is removed 

This *almost* happened, but I found that it's possible to hang indefinitely if pods cannot be drained. This is fairly common if PDBs exist. Even worse, if we're removing the machine CRDs, we're removing Karpenter's ability to make new nodes, so even fairly permissive PDBs will likely get stuck, since the pods cannot roll out successfully. 

I want to provide the pods with the opportunity to exit gracefully (e.g. pod.spec.TerminationGracePeriodSeconds) close their connections, etc, but I don't want to respect PDBs, since this explicit deletion is necessarily a forcing action. Kubernetes has a concept for this in `DeletionTimestamp`, where the semantic states that the resource must be deleted at that time. DeletionTimestamp is can be set in the future, when a grace period is set.

This essentially creates an equivalent experience for `kubectl delete machine` => `kubectl delete machine --grace-period 5` and  `kubectl delete pod` => `kubectl delete pod --grace-period 5`.

After this change, my desired workflow above will always complete, barring pods that declare unreasonable `TerminationGracePeriodSeconds`. I see this as an acceptable tradeoff, and the installer can intervene in this rare case.

Note: there is no change to the current behavior of the deprovisioning controller, which triggers deletion with an indefinite grace period. We may choose to enable customers to configure a Machine's `TerminationGracePeriodSeconds` in the future, which we've heard a few times: https://github.com/aws/karpenter/issues/2481, https://github.com/aws/karpenter/issues/1526

**Open Question**: I've chosen to respect the `TerminationGracePeriodSeconds` of the most conservative pod, and transition from evicting to deleting with enough time for the pod to take the full period. e.g., if a pod requires 30 seconds to terminate, and the node has a 1 minute grace period. It'll spend 30 seconds draining, and then add the NoExecute taint to trigger pod shutdown for the remaining 30 seconds. Instead, I could wait the full node's graceterminationseconds before adding the NoExecute taint.

* Unit
* Manually 
* E2e

### Scenario
* Deployment, 1 replica
* PDB, maxUnavailable=0

```
k delete machines --all
machine.karpenter.sh "default-w9vxb" deleted
```

```
karpenter-7bdc98d4d8-xdvhp controller 2023-08-12T00:23:40.717Z	INFO	controller.termination	tainting no schedule	{"commit": "8547ba1-dirty", "node": "ip-192-168-181-149.us-west-2.compute.internal"}
karpenter-7bdc98d4d8-xdvhp controller 2023-08-12T00:23:41.133Z	INFO	controller.termination	deleted node	{"commit": "8547ba1-dirty", "node": "ip-192-168-181-149.us-west-2.compute.internal"}
karpenter-7bdc98d4d8-xdvhp controller 2023-08-12T00:23:41.382Z	INFO	controller.machine.termination	deleted machine	{"commit": "8547ba1-dirty", "machine": "default-w9vxb", "node": "ip-192-168-181-149.us-west-2.compute.internal", "provisioner": "default", "provider-id": "aws:///us-west-2a/i-0322e7bfc3b203d4f"}
```

### DIFF
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -38,6 +38,11 @@ const (
 	CapacityTypeLabelKey    = Group + "/capacity-type"
 )
 
+// Karpenter specific taints
+const (
+	TaintKeyTerminating = Group + "/terminating"
+)
+
 // Karpenter specific annotations
 const (
 	DoNotDisruptAnnotationKey          = Group + "/do-not-disrupt"
--- a/pkg/controllers/machine/termination/controller.go
+++ b/pkg/controllers/machine/termination/controller.go
@@ -39,6 +39,7 @@ import (
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
 	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
 	machineutil "github.com/aws/karpenter-core/pkg/utils/machine"
+	"github.com/samber/lo"
 )
 
 var _ corecontroller.FinalizingTypedController[*v1alpha5.Machine] = (*Controller)(nil)
@@ -77,8 +78,8 @@ func (c *Controller) Finalize(ctx context.Context, machine *v1alpha5.Machine) (r
 		return reconcile.Result{}, err
 	}
 	for _, node := range nodes {
-		// We delete nodes to trigger the node finalization and deletion flow
-		if err = c.kubeClient.Delete(ctx, node); client.IgnoreNotFound(err) != nil {
+		// Trigger finalization to gracefully terminate the node, propagating the deletion grace period
+		if err = c.kubeClient.Delete(ctx, node, client.GracePeriodSeconds(lo.FromPtr(machine.DeletionGracePeriodSeconds))); client.IgnoreNotFound(err) != nil {
 			return reconcile.Result{}, err
 		}
 	}
--- a/pkg/controllers/termination/terminator/terminator.go
+++ b/pkg/controllers/termination/terminator/terminator.go
@@ -26,6 +26,8 @@ import (
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/aws/karpenter-core/pkg/apis/v1beta1"
+	"github.com/aws/karpenter-core/pkg/scheduling"
 	podutil "github.com/aws/karpenter-core/pkg/utils/pod"
 )
 
@@ -59,7 +61,7 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
 		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
 			return err
 		}
-		logging.FromContext(ctx).Infof("cordoned node")
+		logging.FromContext(ctx).Infof("cordoning")
 	}
 	return nil
 }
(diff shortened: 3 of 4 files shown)


pkg/controllers/termination/terminator/terminator.go
```diff
@@ -59,60 +61,61 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
...
+		return NewNodeDrainError(fmt.Errorf("%d pods are gracefully shutting down", len(pods)))
+	}
+	// Evict remaining pods
+	t.evict(evictable)
+	return NewNodeDrainError(fmt.Errorf("%d pods are waiting to be evicted", len(evictable)))
+}
+func (t *Terminator) withinGracefulShutdownPeriod(node *v1.Node, pods []*v1.Pod) bool {
+	gracePeriodSeconds := lo.Max(lo.Map(pods, func(pod *v1.Pod, _ int) int64 { return lo.FromPtr(pod.Spec.TerminationGracePeriodSeconds) }))
+	gracePeriod := time.Duration(gracePeriodSeconds) * time.Second
+	return t.clock.Now().Add(gracePeriod).After(node.DeletionTimestamp.Time)
```
njtran, line 103: Just posting here to continue this in the review:

I think we still need to align on the philosophy here. The two philosophies are:
1. **[Your solution]** A Machine's DeletionTimestamp dictates the time it should no longer be present in the APIServer, so it should also include the GracefulTerminationPeriodSeconds of the pods that are scheduled to it. This means that we need a pod's deletion timestamp to line up with the node's deletion timestamp, where pods that haven't begun scaling down due to Eviction Errors with PDBs will receive a SIGINT at time: `MachineDeletionTimestamp.Time() - Pod.TerminationGracePeriodSeconds`.  
- If a pod's GracePeriod is >= machine GracePeriod, that means that we'll immediately add the NoExecute taint, beginning force deletion of all pods on the node (since TaintEviction doesn't respect PDBs). The further the maxPodGracePeriodSeconds is from the machine Grace Period, the longer it'll take for those pods to be forcibly removed.

2. A Machine's GracePeriodSeconds (which is used to calculate the DeletionTimestamp) dictates how long the object can be gracefully deleted before forceful actions are taken. In this case, we wouldn't try to sync up the pod's and machine's deletion timestamps, as they're not linked by any OwnerReferences/CascadeDeletion. Karpenter could take the Machine's GracePeriodSeconds to gracefully evict as many pods as possible, then initiate force deletion of all pods, respecting each of their GracefulTerminationPeriodSeconds, rather than initiating force deletion earlier for pods due to the GracefulTerminationPeriodSeconds of other pods, more likely violating PDBs.

I'm a fan of option 2, but curious what yours and other's thoughts are.

pkg/controllers/termination/terminator/terminator.go
```diff
@@ -59,60 +61,61 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
...
-	for _, p := range podList.Items {
-		// Ignore if the pod is complete and doesn't need to be evicted
-		if podutil.IsTerminal(lo.ToPtr(p)) {
-			continue
+	pods := lo.Map(podList.Items, func(pod v1.Pod, _ int) *v1.Pod { return lo.ToPtr(pod) })
+	evictable := lo.Reject(pods, func(p *v1.Pod, _ int) bool {
+		return podutil.IsTerminal(p) ||
+			(podutil.IsTerminating(p) && t.clock.Now().After(p.DeletionTimestamp.Time.Add(1*time.Minute))) ||
+			podutil.ToleratesUnschedulableTaint(p) ||
+			podutil.IsOwnedByNode(p)
```
njtran, line 86: would be nice to add this into a util like `podutil.IsEvictable()` for readibility.

pkg/controllers/termination/terminator/terminator.go
```diff
@@ -59,60 +61,61 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
...
 	}
-	var pods []*v1.Pod
-	for _, p := range podList.Items {
-		// Ignore if the pod is complete and doesn't need to be evicted
-		if podutil.IsTerminal(lo.ToPtr(p)) {
-			continue
+	pods := lo.Map(podList.Items, func(pod v1.Pod, _ int) *v1.Pod { return lo.ToPtr(pod) })
+	evictable := lo.Reject(pods, func(p *v1.Pod, _ int) bool {
+		return podutil.IsTerminal(p) ||
+			(podutil.IsTerminating(p) && t.clock.Now().After(p.DeletionTimestamp.Time.Add(1*time.Minute))) ||
```
njtran, line 84: Why do we need to add a minute?
```suggestion
			(podutil.IsTerminating(p) && t.clock.Now().After(p.DeletionTimestamp.Time))) ||
```

njtran (commented):
Nice work! Glad you brought this up!

pkg/controllers/machine/termination/controller.go
```diff
@@ -77,8 +78,8 @@ func (c *Controller) Finalize(ctx context.Context, machine *v1alpha5.Machine) (r
 		return reconcile.Result{}, err
 	}
 	for _, node := range nodes {
-		// We delete nodes to trigger the node finalization and deletion flow
-		if err = c.kubeClient.Delete(ctx, node); client.IgnoreNotFound(err) != nil {
+		// Trigger finalization to gracefully terminate the node, propagating the deletion grace period
+		if err = c.kubeClient.Delete(ctx, node, client.GracePeriodSeconds(lo.FromPtr(machine.DeletionGracePeriodSeconds))); client.IgnoreNotFound(err) != nil {
```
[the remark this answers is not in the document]
ellistarn: As implemented, Karpenter will only ever terminate machines by setting `client.GracePeriodSeconds(math.MaxInt64)`. This will cascade down to Nodes.

In the manual use case, if a customer executes `kubectl delete machine`, that's intended as a force delete. In the future, we can consider making this configurable, similar to how pods have a configurable grace period.

pkg/apis/v1beta1/labels.go
```diff
@@ -38,6 +38,11 @@ const (
 	CapacityTypeLabelKey    = Group + "/capacity-type"
 )
 
+// Karpenter specific taints
+const (
+	TaintKeyTerminating = Group + "/terminating"
```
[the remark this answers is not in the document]
ellistarn: Was trying to reuse existing concepts, but yours is more accurate. Naming over slack?

pkg/controllers/termination/terminator/terminator.go
```diff
@@ -59,60 +61,61 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
 		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
 			return err
 		}
-		logging.FromContext(ctx).Infof("cordoned node")
+		logging.FromContext(ctx).Infof("cordoning")
```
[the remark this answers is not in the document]
ellistarn: Oh I actually didn't mean to change this. Thanks!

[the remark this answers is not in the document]
ellistarn: Gerunds should only be used before the action is taken.

pkg/controllers/termination/terminator/terminator.go
ellistarn: Solution 1 violates PDBs. Solution 2 violates GraceTerminationPeriodSeconds. In my mind, it's more important to give the process time to clean up, than it is to wait for a replacement to come online. Obviously, we want to respect both, but if we have to choose, I think GraceTerminationPeriodSeconds is more essential. Think about the batch use case, where the ML training job needs to checkpoint and save its work, but there may not be a PDB.

pkg/controllers/termination/terminator/terminator.go
ellistarn: I thought about it, but we mock the time, which makes this a bit tricky to wire the clock through. We could rewrite the tests, but 🤷

pkg/controllers/termination/terminator/terminator.go
ellistarn: This is "stuck terminating". I just kept the minute that was already implemented. I assume it's to catch race conditions, but not totally sure.

ellistarn:
No wonder I was having a hard time making this work: https://github.com/kubernetes/kubernetes/issues/84310#issuecomment-546101771

sftim:
> No wonder I was having a hard time making this work: [kubernetes/kubernetes#84310 (comment)](https://github.com/kubernetes/kubernetes/issues/84310#issuecomment-546101771). Graceful deletion is only supported for pods, it looks like.
> 
> Seems like we may need to go with a tainting approach and shift draining to be before the delete call.

What I'd do (untested / might need work):
- use a custom finalizer for Machines and for Nodes
- if a Machine has a Node and someone marks the node for deletion, taint the Node. Wait up to the `terminationGracePeriodSeconds`<sup>†</sup> of the Machine for graceful shutdown, then remove the node finalizer. Finally, once the Node deletes, remove the finalizer from the associated machine.
- if a Machine has a Node and someone marks the Machine for deletion, trigger deletion for the Node. Then follow the process in the previous point.

I'd apply a `NoSchedule` taint to stop new Pods arriving. Then, allow another component such as a the [descheduler](https://github.com/kubernetes-sigs/descheduler) to try to remove pods from the tainted node, as gracefully as is feasible. The descheduler strategy that supports this is [`RemovePodsViolatingNodeTaints`](https://github.com/kubernetes-sigs/descheduler#removepodsviolatingnodetaints).

Relying on a separate Kubernetes component for graceful Pod removal lets Karpenter focus on its core aim: managing _nodes_. If it turns out we / people using Karpenter want finer-grained control over that descheduling - and I expect that we would - then that improvement could come as PRs against descheduler.

`terminationGracePeriodSeconds` could allow a special value of `Automatic`, meaning that Karpenter infers it based on the Pods that are bound to the node at the time the taint is applied (and those Pods' grace periods, disruption budgets, etc). That could even be the default.

`terminationGracePeriodSeconds` could also live only as a field in the Provisioner and not be copied to each Machine; I don't mind


Finally, a note about the descheduler: I think Karpenter could create a Job to run the descheduler, and manage the lifecycle of that Job. When there is nothing left to deschedule, no need to run a descheduler.

---

**†** new imaginary field

pkg/controllers/termination/terminator/terminator.go
njtran: In the case of an ML training job without a PDB, it should start being cleaned up as soon as the DeletionTimestamp gets set, as Karpenter will evict all pods that don't have PDBs, predicting here won't start the cleanup any sooner. 

Like you said, being predictive helps out on enforcing the Machine's GraceTerminationPeriodSeconds, yet I think users really care more about enforcing the proper cleanup of their pods, and enforcement of availability requirements (not violating PDBs). I'd prefer we violate the Machine's GraceTerminationPeriodSeconds in this case since the implication here is paying more for an instance, but respecting the user's requirements on availability/graceful termination for their pods, which is what they're paying for.

pkg/controllers/machine/termination/controller.go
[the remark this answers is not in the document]
njtran: > In the manual use case, if a customer executes kubectl delete machine, that's intended as a force delete. 

Prior to this change, the "force delete" is meant in respect to `do-not-evict` pods and any other deprovisioning blocking conditions. I think we should make Karpenter's logic infer the same GracePeriodSeconds as the kubectl CLI would, which defaults to a non-zero DeletionGracePeriodSeconds. Making it a default --grace-period-seconds=0 sounds like a weird sharp edge that users have to be aware of any time they want to delete nodes/machines, as I know some users use the `kubectl delete machine` as a way to pick a Machine to be deleted, but not necessarily nuke the machine itself ASAP.

ellistarn:
closing in favor of https://github.com/aws/karpenter-core/pull/466

ellistarn:
Thanks @sftim for the comments. Worth a longer discussion -- gonna play with a new rev on another branch.

PR #466, opened 2023-08-14
feat: Support cascading delete from CRD deletion

ellistarn:
Fixes #N/A 

**Description**

If I uninstall Karpenter CRDs before uninstalling Karpenter, Karpenter should successfully clean up all underlying instances before allowing the CRD to be deleted.

1. `k delete customresourcedefinitions.apiextensions.k8s.io machines.karpenter.sh`
2. CRD hangs in the CRD finalizer, since machine CRs exist
3. Machine deletion triggers cordon/drain on the node
4. Pods are successfully drained from the node
5. The machine is deleted, and thus the CRD is removed

This almost happens today, but I found that it's possible to hang indefinitely if pods cannot be drained. This is fairly common if PDBs exist. Even worse, if we're removing the machine CRDs, we're removing Karpenter's ability to make new nodes, so even fairly permissive PDBs will likely get stuck, since the pods cannot roll out successfully.

My [initial approach](https://github.com/aws/karpenter-core/pull/465) was to leverage Kubernetes Graceful Deletion, but unfortunately, the API Server only [respects that metadata field for pods](https://github.com/kubernetes/kubernetes/blob/6e0cb243d57592c917fe449dde20b0e246bc66be/staging/src/k8s.io/apiserver/pkg/registry/rest/delete.go#L63C2-L63C21). 

This approach decouples graceful from forceful termination, by shifting drain logic to happen before the CR is deleted. 

### Forceful flow
* Machine is deleted
* Cascade to delete node
* Node Finalizer deletes CP instance, and removes finalizer

OR 

* Node is deleted
* Cascade to machine
* Node Finalizer deletes CP instance, and removes finalizer

### Graceful flow
* Add a "Terminating Taint", when the instance is nominated for termination
* Add Terminating:NoSchedule to the node, and start draining it
* Once no more pods can be drained, add Terminating:NoExecute, which will use taint based eviction to remove (credit to @njtran for this idea) remaining pods (e.g. stuck, . 
* It's possible pods will not be drainable (e.g. PDB), in keeping with current behavior, we don't change this. We may choose to implement a drain timeout here in the future (https://github.com/aws/karpenter/issues/2481, https://github.com/aws/karpenter/issues/1526)
* Once all pods are removed from the node, we will delete the node
* The node will then finalize, call cloudprovider delete, and remove the finalizer

* Unit
* E2E TBD

### DIFF
--- a/pkg/controllers/termination/terminator/terminator.go
+++ b/pkg/controllers/termination/terminator/terminator.go
@@ -26,6 +26,8 @@ import (
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/aws/karpenter-core/pkg/apis/v1beta1"
+	"github.com/aws/karpenter-core/pkg/scheduling"
 	podutil "github.com/aws/karpenter-core/pkg/utils/pod"
 )
 
@@ -59,7 +61,7 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
 		if err := t.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
 			return err
 		}
-		logging.FromContext(ctx).Infof("cordoned node")
+		logging.FromContext(ctx).Infof("cordoned")
 	}
 	return nil
 }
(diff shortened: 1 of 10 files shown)


sftim:
> It's possible pods will not be drainable (e.g. PDB), in keeping with current behavior, we don't change this.

- we could fire an Event when we detect this, maybe against the Machine or the Provisioner?
- we could annotate affected nodes with some more details, eg `karpenter.sh/node-drain-notice`
- we could label affected nodes with `karpenter.sh/drain-required: Manual`

all those ideas aimed at making the problem visible to a cluster operator

pkg/controllers/termination/terminator/terminator.go
```diff
@@ -59,60 +61,53 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
...
-	t.evict(podsToEvict)
-
-	if len(podsToEvict) > 0 {
-		return NewNodeDrainError(fmt.Errorf("%d pods are waiting to be evicted", len(podsToEvict)))
+	pods := lo.Reject(lo.ToSlicePtr(podList.Items), func(p *v1.Pod, _ int) bool { return podutil.IsNotEvictable(p) }) // Ignore pods that can't be drained
+	if len(pods) == 0 {
+		return nil
 	}
-	return nil
+	// Evict, respecting PDBs
```
mallow111, line 80: what if PDB has broken pods, are we going to force delete those broken pods somewhere?

pkg/controllers/termination/terminator/terminator.go
```diff
@@ -59,60 +61,53 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
...
-			continue
-		}
-		// Ignore if kubelet is partitioned and pods are beyond graceful termination window
-		if t.isStuckTerminating(lo.ToPtr(p)) {
-			continue
+	pods := lo.Reject(lo.ToSlicePtr(podList.Items), func(p *v1.Pod, _ int) bool { return podutil.IsNotForceEvictable(p) }) // Ignore pods that can't be drained
+	if len(pods) == 0 {
+		return nil
+	}
+	// Taint NoExecute, ignoring PDBs, and triggering graceful deletion for remaining pods
```
mallow111, line 94: similar questions, if we ignore PDB, how can we make sure nodes that are stuck by pdb broken pods will be deleted, do we just reply on timeout to delete the nodes?

PR #471, opened 2023-08-15
chore: Garbage Collect Leaked Node Lease

engedaam:
Fixes #N/A 

**Description**
- As part of our mitigation strategy for https://github.com/aws/karpenter/issues/4363, we are implementing a controller to delete node leases that do not contain any OwnerReference. 
- This controller should be removed once https://github.com/kubernetes/kubernetes/issues/109777 is fixed upstream

- Manually tested

### DIFF
--- a/pkg/controllers/leasegarbagecollection/controller.go
+++ b/pkg/controllers/leasegarbagecollection/controller.go
@@ -1,0 +1,69 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package leasegarbagecollection
+
+import (
+	"context"
+
+	v1 "k8s.io/api/coordination/v1"
+	"knative.dev/pkg/logging"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	corecontroller "github.com/aws/karpenter-core/pkg/operator/controller"
+)
+
+// Controller for the resource
+type Controller struct {
+	kubeClient client.Client
+}
+
+// NewController is a constructor
+func NewController(kubeClient client.Client) corecontroller.Controller {
+	return corecontroller.Typed[*v1.Lease](kubeClient, &Controller{
+		kubeClient: kubeClient,
+	})
+}
+
+func (c *Controller) Name() string {
+	return "lease.garbagecollection"
+}
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil || l.Namespace != "kube-node-lease" {
+		return reconcile.Result{}, nil
+	}
+	err := c.kubeClient.Delete(ctx, l)
+	if err == nil {
+		logging.FromContext(ctx).Debug("found and delete leaked lease")
+		NodeLeaseDeletedCounter.Inc()
+	}
+
+	return reconcile.Result{}, client.IgnoreNotFound(err)
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.Adapt(controllerruntime.
+		NewControllerManagedBy(m).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
+		For(&v1.Lease{}).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}),
+	)
+}
--- a/pkg/controllers/leasegarbagecollection/metrics.go
+++ b/pkg/controllers/leasegarbagecollection/metrics.go
@@ -1,0 +1,37 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package leasegarbagecollection
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/aws/karpenter-core/pkg/metrics"
+)
+
+var (
+	NodeLeaseDeletedCounter = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "karpenter",
+			Subsystem: metrics.NodeSubsystem,
+			Name:      "leases_deleted",
+			Help:      "Number of deleted leaked leases.",
+		},
+	)
+)
+
+func init() {
+	crmetrics.Registry.MustRegister(NodeLeaseDeletedCounter)
+}
--- a/pkg/controllers/leasegarbagecollection/suite_test.go
+++ b/pkg/controllers/leasegarbagecollection/suite_test.go
@@ -1,0 +1,127 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package leasegarbagecollection_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	coordinationsv1 "k8s.io/api/coordination/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/samber/lo"
+
+	"github.com/aws/karpenter-core/pkg/apis"
+	"github.com/aws/karpenter-core/pkg/apis/settings"
+	"github.com/aws/karpenter-core/pkg/controllers/leasegarbagecollection"
+	"github.com/aws/karpenter-core/pkg/operator/controller"
+	"github.com/aws/karpenter-core/pkg/operator/scheme"
+	"github.com/aws/karpenter-core/pkg/test"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "knative.dev/pkg/logging/testing"
+
+	. "github.com/aws/karpenter-core/pkg/test/expectations"
+)
+
+var ctx context.Context
+var env *test.Environment
+var garbageCollectionController controller.Controller
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "LeaseGarbageCollection")
+}
+
+var _ = BeforeSuite(func() {
+	env = test.NewEnvironment(scheme.Scheme, test.WithCRDs(apis.CRDs...))
+	ctx = settings.ToContext(ctx, test.Settings())
+
+	garbageCollectionController = leasegarbagecollection.NewController(env.Client)
+})
+
+var _ = AfterSuite(func() {
+	Expect(env.Stop()).To(Succeed(), "Failed to stop environment")
+})
+
+var _ = AfterEach(func() {
+	ExpectCleanedUp(ctx, env.Client)
+})
+
+var _ = Describe("GarbageCollection", func() {
+	var goodLease *coordinationsv1.Lease
+	var badLease *coordinationsv1.Lease
+	BeforeEach(func() {
+		node := test.Node(test.NodeOptions{})
+		ExpectApplied(ctx, env.Client, node)
+		node = ExpectExists(ctx, env.Client, node)
+		goodLease = &coordinationsv1.Lease{
+			ObjectMeta: v1.ObjectMeta{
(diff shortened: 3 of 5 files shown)


pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,67 @@
...
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.Adapt(controllerruntime.
+		NewControllerManagedBy(m).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
njtran, line 63: you should pre-filter out any leases that aren't the kube-node-lease so that we don't accidentally mess with any other leases in the cluster.

njtran (commented):
Nice. We'll need to add this to the RBAC permissions in the chart in karpenter, right?

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,67 @@
...
+func (c *Controller) Name() string {
+	return "lease.garbagecollection"
+}
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil || l.Namespace != "kube-node-lease" {
+		return reconcile.Result{}, nil
+	}
+	err := c.kubeClient.Delete(ctx, l)
```
jonathan-innis, line 52: We can collapse all of this to

```go
return reconcile.Result{}, client.IgnoreNotFound(c.kubeClient.Delete(ctx, l))
```

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,67 @@
...
+	})
+}
+
+func (c *Controller) Name() string {
+	return "lease.garbagecollection"
+}
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil || l.Namespace != "kube-node-lease" {
```
jonathan-innis, line 49: Are there any other traits of the lease that ensure that we aren't messing with other leases that have no owner references in this namespace?

pkg/controllers/leasegarbagecollection/controller.go
jonathan-innis: This saves on performance but we should still apply these same filtering characteristics in the Reconcile as well

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+var _ = AfterSuite(func() {
+	Expect(env.Stop()).To(Succeed(), "Failed to stop environment")
+})
+
+var _ = AfterEach(func() {
+	ExpectCleanedUp(ctx, env.Client)
+})
+
+var _ = Describe("GarbageCollection", func() {
+	It("should not delete node lease that does contains an OwnerReference", func() {
```
jonathan-innis, line 66:
```suggestion
	It("should not delete node lease that contains an OwnerReference", func() {
```

nit: more concise

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+						UID:        node.UID,
+					},
+				},
+			},
+		}
+		ExpectApplied(ctx, env.Client, badLease)
+		ExpectReconcileSucceeded(ctx, garbageCollectionController, client.ObjectKeyFromObject(badLease))
+		ExpectExists(ctx, env.Client, badLease)
+	})
+	It("should delete node lease that does not contain OwnerReference", func() {
```
jonathan-innis, line 89:
```suggestion
	It("should delete node lease that contains OwnerReference", func() {
```

nit: more concise

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+	Expect(env.Stop()).To(Succeed(), "Failed to stop environment")
+})
+
+var _ = AfterEach(func() {
+	ExpectCleanedUp(ctx, env.Client)
+})
+
+var _ = Describe("GarbageCollection", func() {
+	It("should not delete node lease that does contains an OwnerReference", func() {
+		_, node := test.MachineAndNode(v1alpha5.Machine{})
```
jonathan-innis, line 67: Do we need the `Machine` and the `Node` or just the `Node`?

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+var ctx context.Context
+var env *test.Environment
+var garbageCollectionController controller.Controller
+
+func TestAPIs(t *testing.T) {
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Machine")
```
jonathan-innis, line 47:
```suggestion
	RunSpecs(t, "LeaseGarbageCollection")
```

pkg/controllers/leasegarbagecollection/suite_test.go
engedaam: ?

pkg/controllers/leasegarbagecollection/controller.go
engedaam: We could also limit it also nodes karpenter has launched

njtran: How are we able to detect if it's from a node that Karpenter launched if it has no owner references? I don't think that's possible.

njtran (commented):
LGTM, just waiting to approve to resolve the last comments

pkg/controllers/leasegarbagecollection/controller.go
engedaam: I was thinking we could get the instance from EC2 and look at the tags. Since the name of the lease is the same as the node

njtran: Maybe if there's a creationTimestamp we can ensure that it's at least an hour old so we don't run into race conditions?

engedaam: It also look like the ownerReference is add on the lease creation

https://github.com/kubernetes/kubernetes/blob/b1e35d5616e066c4b0b3f550c33bbf67bb3f876e/staging/src/k8s.io/component-helpers/apimachinery/lease/controller.go#L210

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (c *Controller) Name() string {
+	return "lease.garbagecollection"
+}
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil ||
+		l.Namespace != "kube-node-lease" ||
+		l.CreationTimestamp.Time.Before(time.Now().Add(-time.Hour)) {
```
jonathan-innis, line 52: Based on looking at the kubelet code (https://github.com/kubernetes/kubernetes/blob/dad95e1be65d4e74d33c810f4395596a22115717/pkg/kubelet/util/nodelease.go#L32 and https://github.com/kubernetes/kubernetes/blob/dad95e1be65d4e74d33c810f4395596a22115717/staging/src/k8s.io/component-helpers/apimachinery/lease/controller.go#L210), we know that the lease always adds its ownerReference at the same time that it is created so we don't need this delay for "eventual consistency"

jonathan-innis (commented):
Can we validate what happens if I manually make a change to the ownerReference on the least? Like, does the lease recover if the kubelet still exists? Does our garbage collector delete it and then the kubelet will re-create it?

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,66 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	return reconcile.Result{}, client.IgnoreNotFound(c.kubeClient.Delete(ctx, l))
```
jonathan-innis, line 56: Realizing that we should probably log a debug log when we garbage collect these leases as well as fire metrics that we deleted the lease

engedaam:
> Can we validate what happens if I manually make a change to the ownerReference on the least? Like, does the lease recover if the kubelet still exists? Does our garbage collector delete it and then the kubelet will re-create it?

Manually validate if a user changes the lease, karpenter will not delete the lease. If the user removes the ownerReference karpenter will delete the lease, and Kubelet will replace the lease.

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil || l.Namespace != "kube-node-lease" {
+		return reconcile.Result{}, nil
+	}
+	err := c.kubeClient.Delete(ctx, l)
+	if err == nil {
+		logging.FromContext(ctx).Infof("found and delete leaked lease: %s", l.Name)
+		NodeDeletedLeakedLeaseCounter.With(prometheus.Labels{})
```
jonathan-innis, line 58: This needs an `Inc()`. Also, you don't need the With(), since this metric doesn't need to be a vector if it doesn't have any labels

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil || l.Namespace != "kube-node-lease" {
+		return reconcile.Result{}, nil
+	}
+	err := c.kubeClient.Delete(ctx, l)
+	if err == nil {
+		logging.FromContext(ctx).Infof("found and delete leaked lease: %s", l.Name)
```
jonathan-innis, line 57:
```suggestion
		logging.FromContext(ctx).Debug("found and delete leaked lease")
```

I think `Info` may be too high here. Also, the lease name is already injected in the logger by default. Have we tested this and ensured that the log looks right when this happens?

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+var _ = AfterSuite(func() {
+	Expect(env.Stop()).To(Succeed(), "Failed to stop environment")
+})
+
+var _ = AfterEach(func() {
+	ExpectCleanedUp(ctx, env.Client)
+})
+
+var _ = Describe("GarbageCollection", func() {
```
jonathan-innis, line 65: Test for the metric firing?

pkg/controllers/leasegarbagecollection/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	NodeDeletedLeakedLeaseCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "karpenter",
+			Subsystem: "nodes",
+			Name:      "deleted_leaked_lease",
```
jonathan-innis, line 27:
```suggestion
			Name:      "leases_deleted",
```

pkg/controllers/leasegarbagecollection/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	NodeDeletedLeakedLeaseCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "karpenter",
+			Subsystem: "nodes",
```
jonathan-innis, line 26: You can reference the global `nodes` subsystem, right?

pkg/controllers/leasegarbagecollection/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+
+package leasegarbagecollection
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	NodeDeletedLeakedLeaseCounter = prometheus.NewCounterVec(
```
jonathan-innis, line 23:
```suggestion
	NodeDeletedLeakedLeaseCounter = prometheus.NewCounter(
```

If you don't have any labels, this just needs to be a counter and not a counterVec

pkg/controllers/leasegarbagecollection/metrics.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/aws/karpenter-core/pkg/metrics"
+)
+
+var (
+	NodeLeaseDeletedCounter = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: "karpenter",
+			Subsystem: metrics.NodeSubsystem,
+			Name:      "leases_deleted",
+			Help:      "Number of deleted leaked leases. Labeled by owning provisioner.",
```
jonathan-innis, line 30:
```suggestion
			Help:      "Number of deleted leaked leases.",
```

This isn't labeled by the owning provisioner

pkg/controllers/leasegarbagecollection/controller.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Reconcile the resource
+func (c *Controller) Reconcile(ctx context.Context, l *v1.Lease) (reconcile.Result, error) {
+	if l.OwnerReferences != nil || l.Namespace != "kube-node-lease" {
+		return reconcile.Result{}, nil
+	}
+	err := c.kubeClient.Delete(ctx, l)
+	if err == nil {
+		logging.FromContext(ctx).Debug("found and delete leaked lease")
```
tzneal, line 55: Should log the lease that we are deleting for debugging purposes.

engedaam: The lease is injected into the logger

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,150 @@
...
+	Context("Metrics", func() {
+		It("should fire the terminationSummary metric when deleting nodes", func() {
```
jonathan-innis, line 69:
```suggestion
		It("should fire the leaseDeletedCounter metric when deleting leases", func() {
```

pkg/controllers/leasegarbagecollection/suite_test.go
```diff
@@ -0,0 +1,150 @@
...
+					},
+				},
+			},
+		}
+		ExpectApplied(ctx, env.Client, goodLease)
+		ExpectReconcileSucceeded(ctx, garbageCollectionController, client.ObjectKeyFromObject(goodLease))
+		ExpectExists(ctx, env.Client, goodLease)
+	})
+	It("should delete node lease that does not contain an OwnerReference", func() {
+		badLease := &coordinationsv1.Lease{
```
jonathan-innis, line 113: Are any of these declarations repetitive and can be put into a `BeforeEach` statement?

jonathan-innis (approved):
Nice 🎉

PR #472, opened 2023-08-15
perf: Implement timeouts and Max-batch size for Consolidation

njtran:
Fixes #N/A 

**Description**
I ran some tests with [KwoK](https://github.com/kubernetes-sigs/kwok) and a cost monitoring tool to scale up clusters to 2000, 1500, 1000, 500, and 100 nodes. I tracked metrics like cumulative cost, cumulative pending pod seconds, and overall test time. 

The test scaled down the clusters incrementally by 25% (100% -> 75% -> 50% -> 25%), waiting for Consolidation to fully scale down to the next level and exiting when the number of nodes got to 25% of the peak. With a timeout of **1 minute for Multi Machine Consolidation**, **3 minutes for Single Machine Consolidation**, and a **max batch size of 100 machines for Multi**, Karpenter was able to take more actions quicker, speeding up the whole test to 33 minutes.

- Includes a new metric counter that tracks how many time a timeout has been reached for both Multi-Machine and Single-Machine Consolidation.
- This also adds in higher bucket sizes for Prometheus histogram metrics since Deprovisioning has some histogram items that can reach higher than 60 seconds.

- Tested with KWOK, and `make presubmit`


pkg/controllers/deprovisioning/multimachineconsolidation.go
```diff
@@ -89,10 +96,20 @@ func (m *MultiMachineConsolidation) firstNMachineConsolidationOption(ctx context
 	}
 
 	lastSavedCommand := Command{}
+	// Set a timeout
+	timeout := m.clock.Now().Add(MultiMachineConsolidationTimeoutDuration)
 	// binary search to find the maximum number of machines we can terminate
 	for min <= max {
+		if m.clock.Now().After(timeout) {
+			deprovisioningConsolidationTimeoutsCounter.WithLabelValues("multi-machine").Inc()
```
engedaam, line 104: I might be nice to have typed string for the different consolidation methods?

pkg/controllers/deprovisioning/multimachineconsolidation.go
```diff
@@ -18,19 +18,24 @@ import (
...
+	"github.com/samber/lo"
+
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
 	"github.com/aws/karpenter-core/pkg/controllers/provisioning"
 	"github.com/aws/karpenter-core/pkg/controllers/provisioning/scheduling"
 	"github.com/aws/karpenter-core/pkg/controllers/state"
 	"github.com/aws/karpenter-core/pkg/events"
 )
 
+const MultiMachineConsolidationTimeoutDuration = 1 * time.Minute
```
engedaam, line 37: nit: Have you considerd having all the timeout in one file? Maybe in the consolidation controller?

pkg/controllers/provisioning/scheduling/metrics.go
```diff
@@ -0,0 +1,36 @@
...
+func init() {
+	crmetrics.Registry.MustRegister(schedulingSimulationDuration)
+}
+
+var schedulingSimulationDuration = prometheus.NewHistogram(
+	prometheus.HistogramOpts{
+		Namespace: metrics.Namespace,
+		Subsystem: "provisioner",
+		Name:      "scheduling_simulation_duration_seconds",
+		Help:      "Duration of scheduling process in seconds.",
```
engedaam, line 33: Can we change this description? `schedulingSimulationDuration` seems to have the same description as `schedulingDuration`. Might be confusing to customers

pkg/controllers/deprovisioning/multimachineconsolidation.go
njtran: I'm gonna opt to keep it here, don't think it should really need to be propagated elsewhere.

tzneal (approved):
lgtm

tzneal:
Merging so we can get some test runs.

PR #475, opened 2023-08-17
feat: add a metric for replacement nodes that fail to launch

njtran:
Fixes #N/A 

**Description**
Adds in a counter that increments any time there is a launch failure for replacement nodes within the deprovisioning loop.

- make presubmit


pkg/controllers/deprovisioning/controller.go
```diff
@@ -239,7 +239,11 @@ func (c *Controller) launchReplacementMachines(ctx context.Context, action Comma
 	workqueue.ParallelizeUntil(ctx, len(machineNames), len(machineNames), func(i int) {
 		// machine never became ready or the machines that we tried to launch got Insufficient Capacity or some
 		// other transient error
-		errs[i] = c.waitForReadiness(ctx, machineNames[i], reason)
+		err := c.waitForReadiness(ctx, machineNames[i], reason)
```
jonathan-innis, line 242: Can't we put this in one line? Like 

```go
if err := c.waitForReadiness(ctx, machineNames[i], reason); err != nil {
...
}
```

jonathan-innis (commented):
Looks good to me. Just have one nit. Do we need any other metrics in deprovisioning for things that fail like this? Like failed to terminate in time, etc.?

engedaam (approved):

PR #481, opened 2023-08-21
feat: move blocked conditions to the machine status conditions

njtran:
Fixes #N/A 

**Description**
Adds in a Machine Disruption Controller to check if a machine is deprovisionable and adds it as a status condition that can be referenced in the deprovisioning controller. 

This should allow users to query if a Machine is deprovisionable by checking its status, rather than having to find events, which have a de-dupe interval. 

### DIFF
--- a/pkg/apis/v1beta1/nodeclaim_status.go
+++ b/pkg/apis/v1beta1/nodeclaim_status.go
@@ -60,6 +60,7 @@ var (
 	NodeDrifted       apis.ConditionType = "NodeDrifted"
 	NodeExpired       apis.ConditionType = "NodeExpired"
 	NodeUnderutilized apis.ConditionType = "NodeUnderutilized"
+	NodeDeprovisioningBlocked apis.ConditionType = "NodeDeprovisioningBlocked"
 )
 
 func (in *NodeClaim) GetConditions() apis.Conditions {
--- a/pkg/apis/v1beta1/nodeclaim_status.go
+++ b/pkg/apis/v1beta1/nodeclaim_status.go
@@ -53,13 +53,13 @@ var LivingConditions = []apis.ConditionType{
 }
 
 var (
-	NodeLaunched      apis.ConditionType = "NodeLaunched"
-	NodeRegistered    apis.ConditionType = "NodeRegistered"
-	NodeInitialized   apis.ConditionType = "NodeInitialized"
-	NodeEmpty         apis.ConditionType = "NodeEmpty"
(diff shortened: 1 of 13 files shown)


pkg/apis/v1beta1/nodeclaim_status.go
```diff
@@ -60,6 +60,7 @@ var (
 	NodeDrifted       apis.ConditionType = "NodeDrifted"
 	NodeExpired       apis.ConditionType = "NodeExpired"
 	NodeUnderutilized apis.ConditionType = "NodeUnderutilized"
+	NodeDeprovisioningBlocked apis.ConditionType = "NodeDeprovisioningBlocked"
```
ellistarn, line 63: Should all of these be prefixed by node? Seems redundant, since the condition is on a node!

jonathan-innis: I could go either way on this. The statusCondition currently exists on the `NodeClaim` so these status conditions are pointing to the fact that the `Node` that it owns is experiencing this status. We can look at status conditions that exist on `Node` to see what the prior art looks for something like this:

```yaml
 conditions:
    - lastHeartbeatTime: "2023-08-21T18:43:20Z"
      lastTransitionTime: "2023-07-03T17:07:56Z"
      message: kubelet has sufficient memory available
      reason: KubeletHasSufficientMemory
      status: "False"
      type: MemoryPressure
    - lastHeartbeatTime: "2023-08-21T18:43:20Z"
      lastTransitionTime: "2023-07-03T17:07:56Z"
      message: kubelet has no disk pressure
      reason: KubeletHasNoDiskPressure
      status: "False"
      type: DiskPressure
    - lastHeartbeatTime: "2023-08-21T18:43:20Z"
      lastTransitionTime: "2023-07-03T17:07:56Z"
      message: kubelet has sufficient PID available
      reason: KubeletHasSufficientPID
      status: "False"
      type: PIDPressure
    - lastHeartbeatTime: "2023-08-21T18:43:20Z"
      lastTransitionTime: "2023-07-03T17:08:10Z"
      message: kubelet is posting ready status
      reason: KubeletReady
      status: "True"
      type: Ready
```

Generally, I've seen status conditions prefixed with the "thing" that is experiencing the condition and then the actual type "Expired", "Drifted", etc.

ellistarn: Even here, the types don't have a prefix. The Reasons have Kubelet in them, since it's a major component of the node.

PR #483, opened 2023-08-23
chore: Change duration type for settings

engedaam:
Fixes #N/A 

**Description**
- `metav1.Duration` was used originally for json serialization/deserialization. Since we don't need to string serialize, removing in favor of `time.Duration` 

- `make presubmit`


pkg/apis/settings/settings.go
```diff
@@ -52,8 +51,8 @@ func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context,
 	s := defaultSettings.DeepCopy()
 
 	if err := configmap.Parse(cm.Data,
-		AsMetaDuration("batchMaxDuration", &s.BatchMaxDuration),
-		AsMetaDuration("batchIdleDuration", &s.BatchIdleDuration),
+		AsDuration("batchMaxDuration", &s.BatchMaxDuration),
```
jonathan-innis, line 54: Isn't there a `configmap.AsDuration` function?

pkg/apis/settings/settings.go
```diff
@@ -65,37 +64,19 @@ func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context,
 }
 
 func (in *Settings) Validate() (err error) {
-	if in.BatchMaxDuration == nil {
+	if in.BatchMaxDuration == 0 {
```
jonathan-innis, line 67: Not sure if we need the required check now that this value should always be set. Maybe we should just require users to set this value to at least a second, since I don't think a value lower than that makes a ton of sense.

jonathan-innis (commented):
Nice, a couple comments

pkg/apis/settings/settings.go
```diff
@@ -65,37 +64,15 @@ func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context,
 }
 
 func (in *Settings) Validate() (err error) {
-	if in.BatchMaxDuration == nil {
-		err = multierr.Append(err, fmt.Errorf("batchMaxDuration is required"))
-	} else if in.BatchMaxDuration.Duration <= 0 {
-		err = multierr.Append(err, fmt.Errorf("batchMaxDuration cannot be negative"))
+	if in.BatchMaxDuration < 1 {
```
jonathan-innis, line 67: this should be `time.Second`. IIRC, this actually means 1 microsecond, which isn't what we want here.

pkg/apis/settings/suite_test.go
```diff
@@ -80,7 +80,7 @@ var _ = Describe("Validation", func() {
 		_, err := (&settings.Settings{}).Inject(ctx, cm)
 		Expect(err).To(HaveOccurred())
 	})
-	It("should fail validation when batchIdleDuration is negative", func() {
+	It("should fail validation when batchIdleDuration is less then 1s", func() {
```
jonathan-innis, line 83: We should actually validate that we are maintaining this minimum by testing a value that is just below that value. Neither of these tests ensure that. They both test negative values (which are good test cases, to be clear) but we also need other test cases for the min case.

pkg/apis/settings/suite_test.go
```diff
@@ -71,6 +71,15 @@ var _ = Describe("Validation", func() {
 		_, err := (&settings.Settings{}).Inject(ctx, cm)
 		Expect(err).To(HaveOccurred())
 	})
+	It("should fail validation when batchMaxDuration is less then 1s", func() {
+		cm := &v1.ConfigMap{
+			Data: map[string]string{
+				"batchMaxDuration": "0.8s",
```
njtran, line 77: Can you use `ms` just so we know that we're not running into any odd parsing issues with decimals? 
```suggestion
				"batchMaxDuration": "800ms",
```

pkg/test/settings.go
```diff
@@ -31,11 +30,11 @@ func Settings(overrides ...settings.Settings) *settings.Settings {
 			panic(fmt.Sprintf("Failed to merge pod options: %s", err))
 		}
 	}
-	if options.BatchMaxDuration == nil {
-		options.BatchMaxDuration = &metav1.Duration{}
+	if options.BatchMaxDuration == 0 {
+		options.BatchMaxDuration = time.Duration(0)
```
njtran, line 34: Shouldn't we default this to 1 since that's what the expected behavior is?

pkg/test/settings.go
```diff
@@ -31,11 +30,11 @@ func Settings(overrides ...settings.Settings) *settings.Settings {
...
 	}
-	if options.BatchIdleDuration == nil {
-		options.BatchIdleDuration = &metav1.Duration{}
+	if options.BatchIdleDuration == 0 {
+		options.BatchIdleDuration = time.Duration(0)
```
njtran, line 37: Same here with defaulting to 1

pkg/test/settings.go
```diff
@@ -31,11 +30,11 @@ func Settings(overrides ...settings.Settings) *settings.Settings {
 			panic(fmt.Sprintf("Failed to merge pod options: %s", err))
 		}
 	}
-	if options.BatchMaxDuration == nil {
-		options.BatchMaxDuration = &metav1.Duration{}
+	if options.BatchMaxDuration == 0 {
+		options.BatchMaxDuration = time.Second
```
njtran, line 34: Sorry, I said 1 second, but I think it should be 10: https://github.com/aws/karpenter-core/pull/483/files#diff-5da54946f6369916fac94035e642087425fd64d2ce6c20f495743f86f4dbe7ccR32

njtran (approved):

pkg/test/settings.go
```diff
@@ -31,11 +30,11 @@ func Settings(overrides ...settings.Settings) *settings.Settings {
 			panic(fmt.Sprintf("Failed to merge pod options: %s", err))
 		}
 	}
-	if options.BatchMaxDuration == nil {
-		options.BatchMaxDuration = &metav1.Duration{}
+	if options.BatchMaxDuration == 0 {
+		options.BatchMaxDuration = 10 * time.Second
```
jonathan-innis, line 34: This is going to slow testing down IIRC since these are used when doing the reconcile calls inside of provisioner

njtran: I feel like having the max duration at 10 won't affect things. It's really the idleDuration that matters here. If we make our max too small, we may not grab all the pods that we want. Whereas the idleDuration will let the scheduling simulation begin after 1s if there's no more pods (which usually we don't add in pods in a stream in unit tests).

pkg/controllers/state/suite_test.go
```diff
@@ -1517,9 +1517,9 @@ var _ = Describe("Node Resource Level", func() {
 
 		// Expect that the node is now nominated
 		Expect(ExpectStateNodeExists(node).Nominated()).To(BeTrue())
-		time.Sleep(time.Second * 5) // nomination window is 10s so it should still be nominated
+		time.Sleep(time.Second * 10) // nomination window is 10s so it should still be nominated
```
njtran, line 1520: Nomination window is twice the max batch I believe. https://github.com/aws/karpenter-core/blob/a4c05cede32f6210fb3df7437f9468258936c8a8/pkg/controllers/state/statenode.go#L391

I think code is right but the comment is misleading.

pkg/test/settings.go
jonathan-innis: Discussed offline. We no longer wait for the batcher in our testing code. At one point, we were calling the `provisioner.Reconcile` method directly but have since changed the code to just make a call directly to `provisioner.Schedule`. This means that these values shouldn't tangibly affect test time anymore.

jonathan-innis (approved):
LGTM 🚀 Nice work!

PR #487, opened 2023-08-24
chore: Handle shared state for Machine/NodeClaim

jonathan-innis:
Fixes #N/A 

**Description**

This change converts all of the state functions to take in a provider id as a string rather than a name for `NominateForPod`, `IsNodeNominated`, `MarkForDeletion`, and `UnmarkForDeletion`. This change also changes the name tracking to providerID for deletion to happen per-kind so no naming conflicts cause inconsistent state or leaks.

`make presubmit`
`/karpenter snapshot`

### DIFF
--- a/pkg/controllers/deprovisioning/controller.go
+++ b/pkg/controllers/deprovisioning/controller.go
@@ -209,17 +209,16 @@ func (c *Controller) executeCommand(ctx context.Context, d Deprovisioner, comman
 // nolint:gocyclo
 func (c *Controller) launchReplacementMachines(ctx context.Context, action Command, reason string) error {
 	defer metrics.Measure(deprovisioningReplacementNodeInitializedHistogram)()
-	candidateNodeNames := lo.Map(action.candidates, func(c *Candidate, _ int) string { return c.Node.Name })
 
 	// cordon the old nodes before we launch the replacements to prevent new pods from scheduling to the old nodes
-	if err := c.setNodesUnschedulable(ctx, true, candidateNodeNames...); err != nil {
+	if err := c.setNodesUnschedulable(ctx, true, action.candidates...); err != nil {
 		return fmt.Errorf("cordoning nodes, %w", err)
 	}
 
 	nodeClaimKeys, err := c.provisioner.CreateNodeClaims(ctx, action.replacements, provisioning.WithReason(reason))
 	if err != nil {
 		// uncordon the nodes as the launch may fail (e.g. ICE)
-		err = multierr.Append(err, c.setNodesUnschedulable(ctx, false, candidateNodeNames...))
+		err = multierr.Append(err, c.setNodesUnschedulable(ctx, false, action.candidates...))
 		return err
 	}
 	if len(nodeClaimKeys) != len(action.replacements) {
@@ -227,8 +226,10 @@ func (c *Controller) launchReplacementMachines(ctx context.Context, action Comma
 		return fmt.Errorf("expected %d nodes, got %d", len(action.replacements), len(nodeClaimKeys))
 	}
 
+	candidateProviderIDs := lo.Map(action.candidates, func(c *Candidate, _ int) string { return c.ProviderID() })
+
 	// We have the new machines created at the API server so mark the old machines for deletion
-	c.cluster.MarkForDeletion(candidateNodeNames...)
+	c.cluster.MarkForDeletion(candidateProviderIDs...)
 
 	errs := make([]error, len(nodeClaimKeys))
 	workqueue.ParallelizeUntil(ctx, len(nodeClaimKeys), len(nodeClaimKeys), func(i int) {
@@ -240,8 +241,8 @@ func (c *Controller) launchReplacementMachines(ctx context.Context, action Comma
 		}
 	})
 	if err = multierr.Combine(errs...); err != nil {
-		c.cluster.UnmarkForDeletion(candidateNodeNames...)
-		return multierr.Combine(c.setNodesUnschedulable(ctx, false, candidateNodeNames...),
+		c.cluster.UnmarkForDeletion(candidateProviderIDs...)
+		return multierr.Combine(c.setNodesUnschedulable(ctx, false, action.candidates...),
 			fmt.Errorf("timed out checking machine readiness, %w", err))
 	}
 	return nil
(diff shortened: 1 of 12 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -51,11 +51,12 @@ type Cluster struct {
...
-	mu               sync.RWMutex
-	nodes            map[string]*StateNode           // provider id -> cached node
-	bindings         map[types.NamespacedName]string // pod namespaced named -> node name
-	nameToProviderID map[string]string               // node name -> provider id
-	daemonSetPods    sync.Map                        // daemonSet -> existing pod
+	mu                       sync.RWMutex
+	nodes                    map[string]*StateNode           // provider id -> cached node
+	bindings                 map[types.NamespacedName]string // pod namespaced named -> node name
+	nodeNameToProviderID     map[string]string               // node name -> provider id
+	nodeClaimKeyToProviderID map[nodeclaimutil.Key]string    // node claim key -> provider id
```
bwagner5, line 58: IIUC the NodeClaim key is just a transitionary representation when we're moving from Machine to NodeClaims. After the transition, NodeClaims should just be a unique string name?

pkg/controllers/deprovisioning/controller.go
```diff
@@ -209,26 +209,27 @@ func (c *Controller) executeCommand(ctx context.Context, d Deprovisioner, comman
...
 	if len(nodeClaimKeys) != len(action.replacements) {
 		// shouldn't ever occur since a partially failed CreateNodeClaims should return an error
 		return fmt.Errorf("expected %d nodes, got %d", len(action.replacements), len(nodeClaimKeys))
 	}
 
+	candidateProviderIDs := lo.Map(action.candidates, func(c *Candidate, _ int) string { return c.ProviderID() })
+
 	// We have the new machines created at the API server so mark the old machines for deletion
-	c.cluster.MarkForDeletion(candidateNodeNames...)
+	c.cluster.MarkForDeletion(candidateProviderIDs...)
```
bwagner5, line 232: does changing to provider ID affect a node that never gets a provider ID for some reason?

bwagner5 (dismissed):
lgtm

pkg/controllers/provisioning/provisioner.go
```diff
@@ -349,7 +349,6 @@ func (p *Provisioner) launchMachine(ctx context.Context, n *scheduler.NodeClaim,
 	}
 	instanceTypeRequirement, _ := lo.Find(machine.Spec.Requirements, func(req v1.NodeSelectorRequirement) bool { return req.Key == v1.LabelInstanceTypeStable })
 	logging.FromContext(ctx).With("machine", machine.Name, "requests", machine.Spec.Resources.Requests, "instance-types", instanceTypeList(instanceTypeRequirement.Values)).Infof("created machine")
-	p.cluster.NominateNodeForPod(ctx, machine.Name)
```
njtran, line 352: Are we losing anything here by removing this? I know we have the same nomination call in the scheduling simulation, but I wonder why we did double nomination here anyway.

pkg/controllers/state/cluster.go
```diff
@@ -51,11 +51,12 @@ type Cluster struct {
...
 
-	mu               sync.RWMutex
-	nodes            map[string]*StateNode           // provider id -> cached node
-	bindings         map[types.NamespacedName]string // pod namespaced named -> node name
-	nameToProviderID map[string]string               // node name -> provider id
-	daemonSetPods    sync.Map                        // daemonSet -> existing pod
+	mu                       sync.RWMutex
+	nodes                    map[string]*StateNode           // provider id -> cached node
+	bindings                 map[types.NamespacedName]string // pod namespaced named -> node name
+	nodeNameToProviderID     map[string]string               // node name -> provider id
```
njtran, line 57: I see we only really need this for the podAntiAffinity function now. It'd be nice if we could just remove this eventually.

pkg/controllers/state/cluster.go
```diff
@@ -96,25 +98,31 @@ func (c *Cluster) Synced(ctx context.Context) bool {
...
+		}
+	}
+	stateNodeNames := sets.New(lo.Keys(c.nodeNameToProviderID)...)
 	c.mu.RUnlock()
 
-	names := sets.New[string]()
+	machineNames := sets.New[string]()
 	for _, machine := range machineList.Items {
-		// If the machine hasn't resolved its provider id, then it hasn't resolved its status
-		if machine.Status.ProviderID == "" {
```
njtran, line 105: Don't we need this? Since this function does a list call on the active machines/nodes, this should ensure that we wait for any machines/nodes not in cluster state that don't have a providerID, and consider the cluster state not synced.

pkg/controllers/state/statenode.go
```diff
@@ -135,6 +135,16 @@ func (in *StateNode) Name() string {
 	return in.Node.Name
 }
 
+// ProviderID is the key that is used to map this StateNode
+// If the Node and NodeClaim have a providerID, this should map to a real providerID
+// If the Node does not have a providerID, this will map to the node name
```
njtran, line 140: How does this map to the node name? Is this a typo, because I don't see a reference to the  node name in the code.

pkg/controllers/provisioning/provisioner.go
jonathan-innis: It was actually useless, looking at it now. Since the Machine wouldn't exist in state when we made this call (the providerID would have had to have been resolved by the time that we got here for this to work and the machine reconciler would have had to reconcile the Machine with that providerID). We should have another scheduling simulation with the resolved machine after this call where we should nominate a pod to this machine. Realistically, having this mapping immediately is a little difficult since the state updates are decoupled.

pkg/controllers/state/cluster.go
jonathan-innis: Sorry, it highlighted a bunch of stuff here. Which part do you think that we should remove?

njtran: just the nodeNameToProviderID

pkg/controllers/state/cluster.go
njtran: Ah that's fair, but wouldn't keeping this in here short circuit all the computation we do here anyway? If we know a machine without a providerRef is going to make L125 fail anyway, seems like we can just return false here anyway.

pkg/controllers/state/statenode.go
jonathan-innis: Here's the code pointer: https://github.com/aws/karpenter-core/blob/main/pkg/controllers/state/cluster.go#L239

pkg/controllers/state/cluster.go
jonathan-innis: I'm not sure that that's feasible. The real issue here is:
1. With node bindings, we need to know the node mapping to get the state node
2. When we do a delete, we only get the node name (and we don't get the providerID) so we need some way to map this name back to providerID for cleanup

pkg/controllers/state/cluster.go
jonathan-innis: It was more about making simplifying assumptions to not overcomplicate. I could go either way. It's technically redundant, but I'm fine keeping it if you feel like we should

njtran: I think it's more clear to the reader that this is a way that this can fail, rather than having to infer and debug later.

jonathan-innis: Updated

njtran (approved):

PR #494, opened 2023-08-29
chore: Rename `.spec.disruption` fields in `v1beta1/NodePool`

jonathan-innis:
Fixes #N/A 

**Description**

This PR renames the disruption fields in the `v1beta1/NodePool` and removes the `consolidationPolicy==Never`. Disabling expiration and consolidation are now driven through setting `expireAfter: Never` and `consolidateAfter: Never`

`make presubmit`
`make apply`


pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -51,37 +51,37 @@ spec:
...
+                  expireAfter: 720h
+                description: Disruption contains the parameters that relate to Karpenter's
+                  disruption logic
                 properties:
+                  consolidateAfter:
+                    default: 15s
+                    description: ConsolidateAfter is the duration the controller will
+                      wait before attempting to terminate nodes that are underutilized.
+                      Refer to ConsolidationPolicy for how underutilization is considered.
+                    pattern: ^(([0-9]+(s|m|h))+)|(Never)$
```
njtran, line 67: Calling this out: since duration allows multi-types (like `10m5s`), this will fail for those durations. I think it's an acceptable stance, but a weird gotcha. Any thoughts to allowing this? It would be a tougher regex string for sure.

pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -51,37 +51,37 @@ spec:
...
-                      is created. This is useful to implement features like eventually
-                      consistent node upgrade, memory leak protection, and disruption
-                      testing.
+                  expireAfter:
+                    default: 720h
+                    description: ExpireAfter is the duration the controller will wait
+                      before terminating a node, measured from when the node is created.
+                      This is useful to implement features like eventually consistent
+                      node upgrade, memory leak protection, and disruption testing.
+                    pattern: ^(([0-9]+(s|m|h))+)|(Never)$
```
njtran, line 84: Same here

pkg/apis/crds/karpenter.sh_nodepools.yaml
njtran: Additionally, I see that the validation logic below just uses the`time.ParseDuration` which allows for multi types, so may be worth aligning here.

pkg/controllers/deprovisioning/events/events.go
```diff
@@ -93,7 +93,7 @@ func Terminating(node *v1.Node, nodeClaim *v1beta1.NodeClaim, reason string) []e
 			InvolvedObject: node,
 			Type:           v1.EventTypeNormal,
 			Reason:         "DeprovisioningTerminating",
-			Message:        fmt.Sprintf("Deprovisioning Node: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
+			Message:        fmt.Sprintf("Disruption Node: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
```
njtran, line 96:
```suggestion
			Message:        fmt.Sprintf("Disrupting Node: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
```

pkg/controllers/deprovisioning/events/events.go
```diff
@@ -103,15 +103,15 @@ func Terminating(node *v1.Node, nodeClaim *v1beta1.NodeClaim, reason string) []e
 			InvolvedObject: machine,
 			Type:           v1.EventTypeNormal,
 			Reason:         "DeprovisioningTerminating",
-			Message:        fmt.Sprintf("Deprovisioning Machine: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
+			Message:        fmt.Sprintf("Disruption Machine: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
```
njtran, line 106:
```suggestion
			Message:        fmt.Sprintf("Disrupting Machine: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
```

pkg/controllers/deprovisioning/events/events.go
```diff
@@ -103,15 +103,15 @@ func Terminating(node *v1.Node, nodeClaim *v1beta1.NodeClaim, reason string) []e
...
 			DedupeValues:   []string{string(machine.UID), reason},
 		})
 	} else {
 		evts = append(evts, events.Event{
 			InvolvedObject: nodeClaim,
 			Type:           v1.EventTypeNormal,
 			Reason:         "DeprovisioningTerminating",
-			Message:        fmt.Sprintf("Deprovisioning NodeClaim: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
+			Message:        fmt.Sprintf("Disruption NodeClaim: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
```
njtran, line 114:
```suggestion
			Message:        fmt.Sprintf("Disrupting NodeClaim: %s", cases.Title(language.Und, cases.NoLower).String(reason)),
```

pkg/apis/crds/karpenter.sh_nodepools.yaml
jonathan-innis: > this will fail for those durations

It won't fail due to the `+` at the end. That allow you to have 1 or more of the inner group, so you _could_ do something like `10m5s` like you are proposing above.

pkg/controllers/deprovisioning/events/events.go
jonathan-innis: Whoops, looks like the refactor monster got me again.

njtran (approved):

PR #495, opened 2023-08-30
chore: single init inflight capacity / startup taints

jmdeal:
Fixes #N/A 

**Description**
Adds an initialization flag to the inflight capacity and startup taints for `StateNode` ensuring that the values are only set once. Additionally, this change helps clean up spurious logging due to inflight resource checks during test environment tear down.

`make presubmit`

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -375,19 +375,21 @@ func (c *Cluster) newStateFromNodeClaim(nodeClaim *v1beta1.NodeClaim, oldNode *S
 		oldNode = NewNode()
 	}
 	n := &StateNode{
-		Node:                oldNode.Node,
-		NodeClaim:           nodeClaim,
-		inflightAllocatable: oldNode.inflightAllocatable,
-		inflightCapacity:    oldNode.inflightCapacity,
-		startupTaints:       oldNode.startupTaints,
-		daemonSetRequests:   oldNode.daemonSetRequests,
-		daemonSetLimits:     oldNode.daemonSetLimits,
-		podRequests:         oldNode.podRequests,
-		podLimits:           oldNode.podLimits,
-		hostPortUsage:       oldNode.hostPortUsage,
-		volumeUsage:         oldNode.volumeUsage,
-		markedForDeletion:   oldNode.markedForDeletion,
-		nominatedUntil:      oldNode.nominatedUntil,
+		Node:                     oldNode.Node,
+		NodeClaim:                nodeClaim,
+		inflightInitialized:      oldNode.inflightInitialized,
+		inflightAllocatable:      oldNode.inflightAllocatable,
+		inflightCapacity:         oldNode.inflightCapacity,
+		startupTaintsInitialized: oldNode.startupTaintsInitialized,
+		startupTaints:            oldNode.startupTaints,
+		daemonSetRequests:        oldNode.daemonSetRequests,
+		daemonSetLimits:          oldNode.daemonSetLimits,
+		podRequests:              oldNode.podRequests,
+		podLimits:                oldNode.podLimits,
+		hostPortUsage:            oldNode.hostPortUsage,
+		volumeUsage:              oldNode.volumeUsage,
+		markedForDeletion:        oldNode.markedForDeletion,
+		nominatedUntil:           oldNode.nominatedUntil,
 	}
 	// Cleanup the old nodeClaim with its old providerID if its providerID changes
 	// This can happen since nodes don't get created with providerIDs. Rather, CCM picks up the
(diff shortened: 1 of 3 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -461,18 +465,27 @@ func (c *Cluster) cleanupNode(name string) {
...
 	if err != nil {
 		return client.IgnoreNotFound(err)
 	}
+	n.startupTaintsInitialized = true
 	n.startupTaints = owner.Spec.Template.Spec.StartupTaints
 	return nil
 }
 
 func (c *Cluster) populateInflight(ctx context.Context, n *StateNode) error {
+	if n.inflightInitialized {
```
njtran, line 485: Can you add a comment here why we do this? 
```suggestion
        // We only need to populate the in-flight details once. This prevents us from firing many un-needed errors when the cloud provider Instance types change.
	if n.inflightInitialized {
```

pkg/controllers/state/cluster.go
```diff
@@ -461,18 +465,27 @@ func (c *Cluster) cleanupNode(name string) {
 }
 
 func (c *Cluster) populateStartupTaints(ctx context.Context, n *StateNode) error {
+	if n.startupTaintsInitialized {
```
njtran, line 468: nit: just a comment here        
```suggestion
        // We only need to populate the startup taints once.
	if n.startupTaintsInitialized {
```

pkg/controllers/state/cluster.go
njtran, line 1: Could you add two quick unit tests for this?

pkg/controllers/state/statenode.go
```diff
@@ -77,9 +77,12 @@ type StateNode struct {
 	Node      *v1.Node
 	NodeClaim *v1beta1.NodeClaim
 
+	inflightInitialized bool
```
jonathan-innis, line 80: Can we create a TODO here to remove this as well when machine is added

pkg/controllers/state/suite_test.go
```diff
@@ -779,6 +801,109 @@ var _ = Describe("Inflight Nodes", func() {
 			},
 		}))
 	})
+	It("should only set startup taints once", func() {
```
jonathan-innis, line 804: I think we only need the node to populate the startup taints, right? This code to populate the startup taints is meant to cover the case where there is no machine?

pkg/controllers/state/suite_test.go
```diff
@@ -779,6 +801,109 @@ var _ = Describe("Inflight Nodes", func() {
...
+		machine := test.Machine(v1alpha5.Machine{
```
njtran, line 805: nit: You may be able to do MachineAndNode()

pkg/controllers/state/suite_test.go
```diff
@@ -333,6 +334,27 @@ var _ = Describe("Inflight Nodes", func() {
...
+		ExpectReconcileSucceeded(ctx, nodeController, client.ObjectKeyFromObject(node))
+
+		ExpectStateNodeCount("==", 1)
+		ExpectResources(instanceType.Allocatable(), ExpectStateNodeExists(node).Allocatable())
+		ExpectResources(instanceType.Capacity, ExpectStateNodeExists(node).Capacity())
+
+		cloudProvider.InstanceTypes = cloudProvider.InstanceTypes[1:]
+		// Still expect to succeed w/o instance type since inflight details are set once
+		ExpectReconcileSucceeded(ctx, nodeController, client.ObjectKeyFromObject(node))
+		cloudProvider.InstanceTypes = append([]*cloudprovider.InstanceType{instanceType}, cloudProvider.InstanceTypes...)
```
njtran, line 356: You don't need to add this back in since we re-fill the instance types on the BeforeEach call here https://github.com/aws/karpenter-core/pull/495/files#diff-8eac1a1e4069ba52700c1850c05457d7ff0edf9169df468beb9438918202a5ecR88

pkg/controllers/state/suite_test.go
```diff
@@ -779,6 +801,45 @@ var _ = Describe("Inflight Nodes", func() {
...
+
+		// Update the provisioner to the default suite provisioner w/ no StartupTaints. StartupTaints should still be
+		// populated by custom provisioner.
+		node.Labels[v1alpha5.ProvisionerNameLabelKey] = provisioner.Name
+		ExpectApplied(ctx, env.Client, node)
+		ExpectReconcileSucceeded(ctx, nodeController, client.ObjectKeyFromObject(node))
+
+		ExpectStateNodeCount("==", 1)
+		stateNode := ExpectStateNodeExists(node)
+		Expect(stateNode.Taints()).To(HaveLen(0))
```
njtran, line 841: Shouldn't this be len(1)? The taints are added, and on the second pass, it should just propagate what the old state node got.

jmdeal: `Taints()` returns an empty slice because it's using the list of startup taints from the `customProvisioner`. Since the node is not yet initialized it filters the startup taints out of the taint list. If the node were initialized we would see the taints.

njtran (approved):

PR #497, opened 2023-08-31
test: static provisioner hash test

jmdeal:
Fixes #N/A 

**Description**
Adds a test to compare the hash of different provisioners to previously computed values. This ensures that equivalent provisioners hash to the same value between versions.

`make presubmit`

### DIFF
--- a/pkg/apis/v1alpha5/suite_test.go
+++ b/pkg/apis/v1alpha5/suite_test.go
@@ -567,38 +567,40 @@ var _ = Describe("Provisioner Annotation", func() {
 		}
 		provisioner = test.Provisioner(testProvisionerOptions)
 	})
-	It("should match previous static hashes", func() {
-		provisioners := []struct {
-			provisioner *Provisioner
-			hash        string
-		}{
-			{
-				provisioner: provisioner,
-				hash:        "14114424411830460479",
-			},
-			{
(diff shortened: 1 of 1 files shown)


pkg/apis/v1alpha5/suite_test.go
```diff
@@ -567,6 +567,40 @@ var _ = Describe("Provisioner Annotation", func() {
 		}
 		provisioner = test.Provisioner(testProvisionerOptions)
 	})
+	DescribeTable(
+		"Static Hash Values",
```
jonathan-innis, line 571: Just for sanity checking and code coverage, can we ensure that changing the rest of the fields in the spec get a specific hash value as well?

pkg/apis/v1alpha5/suite_test.go
```diff
@@ -567,6 +570,84 @@ var _ = Describe("Provisioner Annotation", func() {
...
```
jonathan-innis, line 574: nit: I think the test gets prefixed with each container that wraps it so if you find yourself repeating `"should match static hash values"`, you may just want to push that up to the container-level or remove it entirely

jonathan-innis (approved):
Nice work. 🎉

pkg/apis/v1alpha5/suite_test.go
jmdeal: I've changed it in the latest commit to just be `"should match with ..."`. I didn't want to move that up a level since it would then read `static hash values should match with [It] ...` in the `go test` logs, rather than `Static Hash Values [It] should match with ...`.

PR #502, opened 2023-09-06
chore: Support `v1beta1` in CloudProvider interface

jonathan-innis:
Fixes #N/A 

**Description**

This PR migrates the CloudProvider interface to use the new `v1beta1` APIs. A change will have to be made in `aws/karpenter` to support this breaking change. This change is another one of the changes on the path to fully migrating the repository over to the new v1beta1 APIs.

`make presubmit`
`/karpenter snapshot`

### DIFF
--- a/pkg/cloudprovider/metrics/cloudprovider.go
+++ b/pkg/cloudprovider/metrics/cloudprovider.go
@@ -20,7 +20,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
-	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
+	"github.com/aws/karpenter-core/pkg/apis/v1beta1"
 	"github.com/aws/karpenter-core/pkg/operator/injection"
 
 	"github.com/aws/karpenter-core/pkg/cloudprovider"
@@ -35,7 +35,7 @@ const (
 	// MetricLabelErrorDefaultVal is the default string value that represents "error type unknown"
 	MetricLabelErrorDefaultVal = ""
 	// Well-known metricLabelError values
-	MachineNotFoundError      = "MachineNotFoundError"
+	NodeClaimNotFoundError    = "NodeClaimNotFoundError"
 	InsufficientCapacityError = "InsufficientCapacityError"
 )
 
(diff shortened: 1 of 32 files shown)


pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -92,60 +92,60 @@ func Decorate(cloudProvider cloudprovider.CloudProvider) cloudprovider.CloudProv
...
+	instanceType, err := d.CloudProvider.GetInstanceTypes(ctx, nodePool)
 	if err != nil {
 		errorsTotalCounter.With(getLabelsMapForError(ctx, d, method, err)).Inc()
 	}
 	return instanceType, err
 }
 
-func (d *decorator) IsMachineDrifted(ctx context.Context, machine *v1alpha5.Machine) (cloudprovider.DriftReason, error) {
-	method := "IsMachineDrifted"
+func (d *decorator) IsDrifted(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (cloudprovider.DriftReason, error) {
```
njtran, line 145: Is there a reason you don't want to do `IsNodeClaimDrifted`? The karpenter-core cloud provider still does IsMachineDrifted, which I imagine we would have renamed it to IsNodeClaimDrifted.

pkg/apis/v1alpha5/register.go
```diff
@@ -25,7 +25,6 @@ const (
 	Group              = "karpenter.sh"
 	ExtensionsGroup    = "extensions." + Group
 	CompatabilityGroup = "compatibility." + Group
-	TestingGroup       = "testing." + Group // Exclusively used for labeling/discovery in testing
```
njtran, line 28: Is there a reason why we're removing this variable, only to use the raw value in the files it was already used in?

pkg/cloudprovider/types.go
```diff
@@ -36,21 +36,21 @@ type DriftReason string
...
 	// Availability of types or zone may vary by provisioner or over time.  Regardless of
 	// availability, the GetInstanceTypes method should always return all instance types,
 	// even those with no offerings available.
-	GetInstanceTypes(context.Context, *v1alpha5.Provisioner) ([]*InstanceType, error)
-	// IsMachineDrifted returns whether a machine has drifted from the provisioning requirements
+	GetInstanceTypes(context.Context, *v1beta1.NodePool) ([]*InstanceType, error)
+	// IsDrifted returns whether a machine has drifted from the provisioning requirements
 	// it is tied to.
-	IsMachineDrifted(context.Context, *v1alpha5.Machine) (DriftReason, error)
+	IsDrifted(context.Context, *v1beta1.NodeClaim) (DriftReason, error)
```
njtran, line 53: Did we discuss making this IsDrifted() rather than IsNodeClaimDrifted?

pkg/apis/v1alpha5/register.go
jonathan-innis: Actually, good point. We shouldn't remove it from the previous API package (since someone could be relying on it) but we don't want to support this as a native concept moving forward.

pkg/cloudprovider/types.go
jonathan-innis: Discussed offline: Agreed that `IsDrifted` is a reasonable name for the interface method

njtran (dismissed):

njtran (approved):

PR #503, opened 2023-09-06
chore: remove unprovisionable pods from scheduling results

njtran:
Fixes #N/A 

**Description**
This change essentially moves the check to see if podSchedulingErrors has unprovisionable pods into the scheduling Solve() function itself, since we only need to log the PodSchedulingErrors for unprovisionable pods when we're in scheduling. Only deprovisioning needs to be aware of the PodSchedulingErrors, where unprovisionable pods are ignored.

- make presubmit 

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -102,16 +102,11 @@ type Results struct {
 	PodErrors     map[*v1.Pod]error
 }
 
-// AllNonPendingPodsScheduled returns true if all of the non-pending pods scheduled.  This is useful in consolidation as
-// we don't care if a pod was pending before consolidation and will still be pending after. It may be a pod that we
-// can't schedule at all and don't want it to block consolidation.
-func (r Results) AllNonPendingPodsScheduled() bool {
-	for p := range r.PodErrors {
-		if !pod.IsProvisionable(p) {
-			return false
-		}
-	}
-	return true
+// AllPodsScheduled returns true if all pods scheduled. SchedulingResults will not have errors of pods that aren't provisionable
+// since we don't care if a pod was pending before consolidation and will still be pending after. It may be a pod that we can't
+// schedule at all and don't want it to block consolidation.
+func (r Results) AllPodsScheduled() bool {
+	return len(r.PodErrors) == 0
 }
 
 // PodSchedulingErrors creates a string that describes why pods wouldn't schedule that is suitable for presentation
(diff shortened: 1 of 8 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -172,16 +167,19 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*v1.Pod) (*Results, error)
 		s.recordSchedulingResults(ctx, pods, q.List(), errors)
 	}
 	// clear any nil errors so we can know that len(PodErrors) == 0 => all pods scheduled
+	// we can also clear any pod errors for pods that aren't provisionable since
+	// 1. Provisioning only cares about the new nodes we need to provisionable
+	// 2. Deprovisioning only cares about the pods on the candidates we're trying to delete.
 	for k, v := range errors {
-		if v == nil {
+		if v == nil || !pod.IsProvisionable(k) {
```
jonathan-innis, line 174: Is there a case where we even handle Provisionable pods? I know that the PendingPods call already filters these out. Are we just handling pods on Deleting nodes here?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -102,39 +102,39 @@ type Results struct {
 	PodErrors     map[*v1.Pod]error
 }
 
-// AllNonPendingPodsScheduled returns true if all of the non-pending pods scheduled.  This is useful in consolidation as
-// we don't care if a pod was pending before consolidation and will still be pending after. It may be a pod that we
-// can't schedule at all and don't want it to block consolidation.
+// AllNonPendingPodsScheduled returns true if all pods scheduled.
+// We don't care if a pod was pending before consolidation and will still be pending after. It may be a pod that we can't
+// schedule at all and don't want it to block consolidation.
 func (r Results) AllNonPendingPodsScheduled() bool {
```
jonathan-innis, line 108: This looks good to me. Just a thought: I wonder if we could just have a filter on the results so that we could do something like

```go
results.NonPendingPods().Errors()
results.PendingPods().Errors()
results.Errors()
```

jonathan-innis (approved):
Just one thought

pkg/controllers/provisioning/scheduling/scheduler.go
njtran: Since we only care about the NonPendingPods in code, I don't think there's a use in adding these, but could definitely see the use in the future.

PR #504, opened 2023-09-07
chore: Add `v1beta1/NodeClaim` informer for cluster state

jonathan-innis:
Fixes #N/A 

**Description**

This change adds the informer for the `v1beta1/NodeClaim` to the `cluster.State`. This change is another one of the changes on the path to fully migrating the repository over to the new v1beta1 APIs.

`make presubmit`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,10 +5,13 @@ help: ## Display help
 
 presubmit: verify test licenses vulncheck ## Run all steps required for code to be checked in
 
+## TODO @joinnis: Reduce the gingko timeout back to 10m once the v1alpha5 testing is removed
 test: ## Run tests
 	go test ./... \
 		-race \
+		-timeout 20m \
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -8,14 +8,17 @@ presubmit: verify test licenses vulncheck ## Run all steps required for code to
 test: ## Run tests
 	go test ./... \
 		-race \
+		-timeout 20m \
```
njtran, line 11: I'm guessing that this is because we've essentially doubled our cluster state suite tests, right? Can we add a TODO here when we remove all the machine suite tests here? Or should we keep this here in general to alert us when our tests are taking too long?

jonathan-innis: Makes sense. I can add a TODO here

njtran (approved):

PR #506, opened 2023-09-08
chore: Remove reference to `machine` in deprovisioning

jonathan-innis:
Fixes #N/A 

**Description**

Remove references to the term "machine" in deprovisioning

`make presubmit`

### DIFF
--- a/pkg/controllers/deprovisioning/controller.go
+++ b/pkg/controllers/deprovisioning/controller.go
@@ -62,7 +62,7 @@ const immediately = time.Millisecond
 
 var errCandidateDeleting = fmt.Errorf("candidate is deleting")
 
-// waitRetryOptions are the retry options used when waiting on a machine to become ready or to be deleted
+// waitRetryOptions are the retry options used when waiting on a node to become ready or to be deleted
 // readiness can take some time as the node needs to come up, have any daemonset extended resoruce plugins register, etc.
 // deletion can take some time in the case of restrictive PDBs that throttle the rate at which the node is drained
 func waitRetryOptions(ctx context.Context) []retry.Option {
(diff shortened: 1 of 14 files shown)


pkg/controllers/deprovisioning/metrics.go
```diff
@@ -32,8 +32,8 @@ const (
 	actionLabel             = "action"
 	consolidationType       = "consolidation_type"
 
-	multiMachineConsolidationLabelValue  = "multi-machine"
-	singleMachineConsolidationLabelValue = "single-machine"
+	multiNodeConsolidationLabelValue  = "multi-node"
+	singleNodeConsolidationLabelValue = "single-node"
```
njtran, line 36: This makes the PR breaking, no?

pkg/controllers/deprovisioning/metrics.go
```diff
@@ -50,8 +50,8 @@ var (
 		prometheus.HistogramOpts{
 			Namespace: metrics.Namespace,
 			Subsystem: deprovisioningSubsystem,
-			Name:      "replacement_machine_initialized_seconds",
-			Help:      "Amount of time required for a replacement machine to become initialized.",
+			Name:      "replacement_node_initialized_seconds",
```
njtran, line 53: Along with this and all other metrics changes.

pkg/controllers/deprovisioning/metrics.go
```diff
@@ -63,12 +63,12 @@ var (
...
 		[]string{actionLabel, deprovisionerLabel},
 	)
-	deprovisioningEligibleMachinesGauge = prometheus.NewGaugeVec(
+	deprovisioningEligibleNodesGauge = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
 			Namespace: metrics.Namespace,
 			Subsystem: deprovisioningSubsystem,
-			Name:      "eligible_machines",
-			Help:      "Number of machines eligible for deprovisioning by Karpenter. Labeled by deprovisioner",
+			Name:      "eligible_nodes",
```
njtran, line 70: should we be calling it `eligible_node_claims` here and likewise in other places?

pkg/controllers/deprovisioning/multinodeconsolidation.go
```diff
@@ -140,23 +140,23 @@ func (m *MultiMachineConsolidation) firstNMachineConsolidationOption(ctx context
...
-// machines=[t3a.2xlarge, t3a.2xlarge, t3a.small] -> 1 of t3a.nano, t3a.small, t3a.xlarge, t3a.2xlarge
+// nodes=[t3a.2xlarge, t3a.2xlarge, t3a.small] -> 1 of t3a.nano, t3a.small, t3a.xlarge, t3a.2xlarge
 //
 // This code sees that t3a.small is the cheapest type in both lists and filters it and anything more expensive out
 // leaving the valid consolidation:
-// machines=[t3a.2xlarge, t3a.2xlarge, t3a.small] -> 1 of t3a.nano
-func filterOutSameType(newMachine *scheduling.NodeClaim, consolidate []*Candidate) []*cloudprovider.InstanceType {
-	existingInstanceTypes := sets.NewString()
+// nodes=[t3a.2xlarge, t3a.2xlarge, t3a.small] -> 1 of t3a.nano
+func filterOutSameType(newNode *scheduling.NodeClaim, consolidate []*Candidate) []*cloudprovider.InstanceType {
```
njtran, line 155: nit: match the variable name
```suggestion
func filterOutSameType(newNodeClaim *scheduling.NodeClaim, consolidate []*Candidate) []*cloudprovider.InstanceType {
```

pkg/controllers/deprovisioning/metrics.go
jonathan-innis: Yes it does. We need to add this with your taint changes so that we package all the breaking changes together.

pkg/controllers/deprovisioning/controller.go
```diff
@@ -259,9 +259,9 @@ func (c *Controller) waitForReadiness(ctx context.Context, key nodeclaimutil.Key
 			// The NodeClaim got deleted after an initial eventual consistency delay
 			// This means that there was an ICE error or the Node initializationTTL expired
 			if errors.IsNotFound(err) && c.clock.Since(pollStart) > time.Second*5 {
-				return retry.Unrecoverable(fmt.Errorf("getting machine, %w", err))
+				return retry.Unrecoverable(fmt.Errorf("getting %s, %w", lo.Ternary(key.IsMachine, "machine", "nodeclaim"), err))
```
jackfrancis, line 262: I see we're still referring to "machine" here, is that intentional?

jonathan-innis: Yes, we're referring to Machine here for nodes that are still represented as Machines. For those that are represented as NodeClaims, we use the "nodeclaim" string value

ellistarn:
Thoughts on single-node and multi-node? 

Also. I wonder if we even should have a compatibility guarantee on the internals of consolidation. Should we think about including "internal" on some of these metrics?

jonathan-innis:
> Thoughts on single-node and multi-node

I went back on forth on this with @njtran. I'm back on the side of doing node here. I think it makes the most sense.

njtran (approved):

PR #508, opened 2023-09-11
BREAKING CHANGE: replace node.kubernetes.io/unschedulable with a Karpenter-specific taint

njtran:
Fixes #N/A 

**Description**
Design Doc: https://github.com/aws/karpenter-core/pull/585

Karpenter will now use the `karpenter.sh/disrupting` taint instead of `node.kubernetes.io/unschedulable` taint when deprovisioning nodes. Karpenter only adds this taint once it has begun spinning up replacement nodes for deprovisioning actions, and subsequently deleting the candidate node. 

- Karpenter would add/remove this taint, assuming no other agent had been managing this taint.
- This fixes an issue where in rare cases where Karpenter restarted while executing a deprovisioning action, Karpenter can leave nodes that were meant to be deprovisioned as cordoned. 
- This breaks some users with pods that tolerate the `node.kubernetes.io/unschedulable` taint or tolerate all keys: a `*` toleration. If users want to ensure their daemonsets and other pods that had this toleration are not evicted as part of termination, they'll need to add this toleration to their pods. 
- This also modifies the cluster state logic to remove the `stateNode.MarkedForDeletion` variable. Karpenter can now rely on the existence of the `karpenter.sh/disrupting` taint to know when a node has been marked for deletion.

- make presubmit 
- make apply && manual test

### DIFF
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -46,12 +46,23 @@ const (
 	NodePoolHashAnnotationKey          = Group + "/nodepool-hash"
 )
 
+// Karpenter specific taints
+const (
+	DisruptingTaintKey = Group + "/disrupting"
+)
+
 // Karpenter specific finalizers
 const (
 	TerminationFinalizer = Group + "/termination"
 )
 
 var (
+	// DisruptingNoScheduleTaint is used by the deprovisioning controller to ensure no pods
+	// are scheduled to a node that Karpenter is actively disrupting.
+	DisruptingNoScheduleTaint = v1.Taint{
+		Key:    DisruptingTaintKey,
+		Effect: v1.TaintEffectNoSchedule,
+	}
 	// RestrictedLabelDomains are either prohibited by the kubelet or reserved by karpenter
 	RestrictedLabelDomains = sets.New(
 		"kubernetes.io",
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -46,23 +46,12 @@ const (
 	NodePoolHashAnnotationKey          = Group + "/nodepool-hash"
 )
 
-// Karpenter specific taints
-const (
-	DisruptingTaintKey = Group + "/disrupting"
-)
-
 // Karpenter specific finalizers
 const (
 	TerminationFinalizer = Group + "/termination"
 )
 
 var (
-	// DisruptingNoScheduleTaint is used by the deprovisioning controller to ensure no pods
-	// are scheduled to a node that Karpenter is actively disrupting.
-	DisruptingNoScheduleTaint = v1.Taint{
-		Key:    DisruptingTaintKey,
-		Effect: v1.TaintEffectNoSchedule,
-	}
 	// RestrictedLabelDomains are either prohibited by the kubelet or reserved by karpenter
 	RestrictedLabelDomains = sets.New(
 		"kubernetes.io",
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -31,3 +31,7 @@ var (
 		Value:  DisruptingNoScheduleTaintValue,
 	}
 )
+
+func IsDisruptingTaint(taint v1.Taint) bool {
+	return taint.MatchTaint(&DisruptionNoScheduleTaint) && taint.Value == DisruptingNoScheduleTaintValue
+}
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -19,7 +19,7 @@ import v1 "k8s.io/api/core/v1"
 // Karpenter specific taints
 const (
 	DisruptionTaintKey   = Group + "/disruption"
-	DisruptionTaintValue = "disrupting"
+	DisruptingNoScheduleTaintValue = "disrupting"
 )
 
 var (
@@ -28,6 +28,6 @@ var (
 	DisruptionNoScheduleTaint = v1.Taint{
 		Key:    DisruptionTaintKey,
 		Effect: v1.TaintEffectNoSchedule,
-		Value:  DisruptionTaintValue,
+		Value:  DisruptingNoScheduleTaintValue,
 	}
 )
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -18,7 +18,7 @@ import v1 "k8s.io/api/core/v1"
 
 // Karpenter specific taints
 const (
-	DisruptionTaintKey   = Group + "/disruption"
+	DisruptionTaintKey             = Group + "/disruption"
 	DisruptingNoScheduleTaintValue = "disrupting"
 )
 
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -1,0 +1,33 @@
+/*
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import v1 "k8s.io/api/core/v1"
+
+// Karpenter specific taints
+const (
+	DisruptionTaintKey   = Group + "/disruption"
+	DisruptionTaintValue = "disrupting"
+)
+
+var (
+	// DisruptionNoScheduleTaint is used by the deprovisioning controller to ensure no pods
+	// are scheduled to a node that Karpenter is actively disrupting.
+	DisruptionNoScheduleTaint = v1.Taint{
+		Key:    DisruptionTaintKey,
+		Effect: v1.TaintEffectNoSchedule,
+		Value:  DisruptionTaintValue,
+	}
+)
--- a/pkg/controllers/deprovisioning/controller.go
+++ b/pkg/controllers/deprovisioning/controller.go
@@ -221,14 +221,14 @@ func (c *Controller) launchReplacementNodeClaims(ctx context.Context, action Com
 	stateNodes := lo.Map(action.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode })
 
 	// cordon the old nodes before we launch the replacements to prevent new pods from scheduling to the old nodes
-	if err := c.taintCandidates(ctx, stateNodes...); err != nil {
+	if err := c.taintCandidates(ctx, true, stateNodes...); err != nil {
 		return fmt.Errorf("cordoning nodes, %w", err)
 	}
 
 	nodeClaimKeys, err := c.provisioner.CreateNodeClaims(ctx, action.replacements, provisioning.WithReason(reason))
 	if err != nil {
 		// uncordon the nodes as the launch may fail (e.g. ICE)
-		err = multierr.Append(err, c.unTaintCandidates(ctx, stateNodes...))
+		err = multierr.Append(err, c.taintCandidates(ctx, false, stateNodes...))
 		return err
 	}
 	if len(nodeClaimKeys) != len(action.replacements) {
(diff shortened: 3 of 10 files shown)


pkg/controllers/deprovisioning/controller.go
```diff
@@ -298,33 +303,65 @@ func (c *Controller) waitForDeletion(ctx context.Context, nodeClaim *v1beta1.Nod
 	}
 }
 
-func (c *Controller) setNodesUnschedulable(ctx context.Context, isUnschedulable bool, candidates ...*Candidate) error {
+func (c *Controller) taintCandidates(ctx context.Context, addTaint bool, nodes ...*state.StateNode) error {
```
jackfrancis, line 305: 1. Can we add a comment to the method that documents that the `addTaint` bool parameter is overloaded to "add if true" or "remove if false"? Maybe `requireTaint` is a better name?
2. Do we want to make this a general func that takes a taint as input, or do we want this to be *just* for enforcing the `v1beta1.DisruptingNoScheduleTaint` taint config specifically? If the latter, should we maybe rename the method to something like `disruptingNoScheduleTaintCandidates` (I know that's a horribly long name).

njtran:
Adding the blocked label so we can first vet if changing this taint right now is what we want.

pkg/apis/v1beta1/labels.go
```diff
@@ -46,12 +46,23 @@ const (
 	NodePoolHashAnnotationKey          = Group + "/nodepool-hash"
 )
 
+// Karpenter specific taints
```
jonathan-innis, line 49: musing: But this is kind-of ceasing to become just a "labels" file. Should we separate taints into "taints.go" or consider renaming this from "labels.go"

pkg/controllers/deprovisioning/controller.go
```diff
@@ -297,26 +301,38 @@ func (c *Controller) waitForDeletion(ctx context.Context, nodeClaim *v1beta1.Nod
 	}
 }
 
-func (c *Controller) setNodesUnschedulable(ctx context.Context, isUnschedulable bool, candidates ...*Candidate) error {
+// requireDisruptingNoScheduleTaint will either add/remove the karpenter.sh/disrupting taint from the candidates.
+// This is used to enforce no taints at the beginning of deprovisioning, and
+// to add/remove taints while executing a deprovisioning action.
+func (c *Controller) requireDisruptingNoScheduleTaint(ctx context.Context, addTaint bool, nodes ...*state.StateNode) error {
```
jonathan-innis, line 307: I'm personally not a big fan of a single function that conveys vastly different behavior dictated by a flag. In this case we are using an `addTaint` flag to determine whether we remove or add taints from the nodes. Does it make more sense to just break this into two separate functions to make this more explicit?

pkg/controllers/state/statenode.go
```diff
@@ -220,7 +219,7 @@ func (in *StateNode) Taints() []v1.Taint {
 	// We can assume that all nodes will have this label and no back-propagation will be required once we hit v1
 	if !in.Registered() && !in.Initialized() && in.NodeClaim != nil {
```
jonathan-innis, line 220:
```suggestion
	if (!in.Registered() && !in.Initialized() && in.NodeClaim != nil) || in.Node == nil {
```

Should this be the handling? If it's not initialized or registered and the nodeclaim is nil or if the node doesn't exist yet.

jonathan-innis (commented):
Looks good! Just a few thoughts.

pkg/apis/v1beta1/labels.go
```diff
@@ -46,12 +46,23 @@ const (
...
+const (
+	DisruptingTaintKey = Group + "/disrupting"
```
ellistarn, line 51: I'd suggest a single disruption taint with multiple values, rather than multiple taints. This enables users to take advantage of taint values to write more expressive tolerations, instead of writing explicit tolerations for each case. 

We have another chance to change this before v1, and I'd like to see how far we get with the more minimal approach.

njtran: Synced offline, and changed to one taint with multiple values.

pkg/controllers/state/statenode.go
```diff
@@ -339,11 +338,14 @@ func (in *StateNode) PodLimits() v1.ResourceList {
...
 func (in *StateNode) MarkedForDeletion() bool {
-	// The Node is marked for the Deletion if:
-	//  1. The Node has explicitly MarkedForDeletion
+	// The Node is marked for deletion if:
+	//  1. The Node has the karpenter.sh/disruption:NoSchedule=disrupting taint
 	//  2. The Node has a NodeClaim counterpart and is actively deleting
 	//  3. The Node has no NodeClaim counterpart and is actively deleting
-	return in.markedForDeletion ||
+	_, ok := lo.Find(in.Taints(), func(taint v1.Taint) bool {
+		return v1beta1.DisruptionNoScheduleTaint.MatchTaint(&taint)
```
jonathan-innis, line 346: Does match taint match on values and effect? From what I remember, it might just be two of the three?

pkg/controllers/deprovisioning/controller.go
```diff
@@ -297,26 +301,40 @@ func (c *Controller) waitForDeletion(ctx context.Context, nodeClaim *v1beta1.Nod
...
+// requireDisruptionNoScheduleTaint will either add/remove the karpenter.sh/disruption taint from the candidates.
+// This is used to enforce no taints at the beginning of deprovisioning, and
+// to add/remove taints while executing a deprovisioning action.
+func (c *Controller) requireDisruptionNoScheduleTaint(ctx context.Context, addTaint bool, nodes ...*state.StateNode) error {
 	var multiErr error
-	for _, cn := range candidates {
+	for _, n := range nodes {
 		node := &v1.Node{}
-		if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: cn.Node.Name}, node); err != nil {
+		if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: n.Name()}, node); client.IgnoreNotFound(err) != nil {
```
jonathan-innis, line 311:
```suggestion
		if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: n.Node.Name}, node); client.IgnoreNotFound(err) != nil {
```

You should probably just directly grab the node name here just to ensure that you are getting the right name here. `n.Name()` _should_ resolve to the node name, but it has the potential to not to.

pkg/controllers/deprovisioning/controller.go
```diff
@@ -297,26 +301,40 @@ func (c *Controller) waitForDeletion(ctx context.Context, nodeClaim *v1beta1.Nod
...
 			multiErr = multierr.Append(multiErr, fmt.Errorf("getting node, %w", err))
 		}
 
-		// node is being deleted already, so no need to un-cordon
-		if !isUnschedulable && !node.DeletionTimestamp.IsZero() {
+		// If the node already has the taint, continue to the next
+		_, ok := lo.Find(node.Spec.Taints, func(taint v1.Taint) bool {
+			return v1beta1.DisruptionNoScheduleTaint.MatchTaint(&taint)
```
jonathan-innis, line 317: Let's double-check that MatchTaint has the correct logic for handling, key, value, effect. IIRC, it might not be correct

pkg/controllers/deprovisioning/controller.go
```diff
@@ -297,26 +301,40 @@ func (c *Controller) waitForDeletion(ctx context.Context, nodeClaim *v1beta1.Nod
...
+		})
+
+		// node is being deleted, so no need to remove taint as the node will be gone soon
+		if ok && !node.DeletionTimestamp.IsZero() {
```
jonathan-innis, line 321: If the taint doesn't exist and we're deleting, should we add it?

pkg/apis/v1beta1/taints.go
```diff
@@ -0,0 +1,33 @@
...
+*/
+
+package v1beta1
+
+import v1 "k8s.io/api/core/v1"
+
+// Karpenter specific taints
+const (
+	DisruptionTaintKey             = Group + "/disruption"
+	DisruptingNoScheduleTaintValue = "disrupting"
```
sftim, line 22: So, I'd call the taint “node-consolidation-planned”. It's tricky though because if we have lots of taints and people want to tolerate them, they have to learn what those taints are (or tolerate `*`, which has its own problems).

The actual name isn't the point here; it's the idea of having a taint that's specific to each different intent.
If future Karpenter drains nodes due to an elevated spot termination risk, then IMO the taint could and should be different.

ellistarn, line 22: Draining?

pkg/controllers/state/statenode.go
njtran: Good catch

pkg/apis/v1beta1/taints.go
jonathan-innis: @sftim Can you detail what the problems are with tolerating `*`? The original intent around making this a single taint key with multiple values was to make it easier for users to tolerate all of the Karpenter taints across the board without having to think about it too much. It seems like the node SIG takes the opposite approach where each part of the lifecycle has its own taint. I'm curious what the strengths of drawbacks of each of these approaches are in your opinion.

sftim: The trouble with tolerations is that you can only tolerate `*`. So if there's a taint `example.com/node-actually-on-fire` and you tolerate `*`, your Pod schedules there.

What you probably want to do is to tolerate families of taints. You might have a DaemonSet pod that provides packet filtering (and fails closed).

What you want is roughly to deploy first, drain last, and tolerate at least:
- `node.kubernetes.io/memory-pressure`
- `node.kubernetes.io/disk-pressure`
- `node.kubernetes.io/pid-pressure`
- `node.kubernetes.io/network-unavailable`

(which, fortunately, the DaemonSet controller adds automatically).

You've never heard of `spot-compute.example/preemption-imminent` and you'd like to tolerate that. The problem is, you don't want to tolerate _every_ taint.

Anyway, the risk of having lots of fine grained taints is grumpy sysadmins moaning that there's a growing list of taints they need to explicitly tolerate. The answer may be to expect a level of moaning and have clear documentation to steer people towards.

sftim: If we plan to add future taints, we can reserve them and document the reservation. That helps people plan ahead and tolerate the taints now, even before we ever use them.

pkg/controllers/deprovisioning/suite_test.go
```diff
@@ -130,6 +131,259 @@ var _ = AfterEach(func() {
...
+		})
+		ExpectDeleted(ctx, env.Client, createdMachine[0])
+		ExpectMachinesCascadeDeletion(ctx, env.Client, createdMachine[0])
+		ExpectNotFound(ctx, env.Client, createdMachine[0])
+		wg.Wait()
+		machineNode = ExpectNodeExists(ctx, env.Client, machineNode.Name)
+		Expect(machineNode.Spec.Unschedulable).To(BeFalse())
+	})
+})
+
```
jonathan-innis, line 386: Do we have a test case to validate that we remove the taints on re-reconcile if they are left behind after the original reconcile loop? You can probably just mock this behavior by adding the taint to a NodeClaim node and then just executing a reconcile to ensure that it gets removed

pkg/controllers/deprovisioning/suite_test.go
```diff
@@ -130,6 +131,259 @@ var _ = AfterEach(func() {
...
+		// inform cluster state about nodes and machines
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{nodeClaimNode}, []*v1beta1.NodeClaim{nodeClaim})
+
+		// Trigger the reconcile loop to start but don't trigger the verify action
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			ExpectTriggerVerifyAction(&wg)
+			_, _ = deprovisioningController.Reconcile(ctx, reconcile.Request{})
```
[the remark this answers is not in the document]
njtran: good call. I decided to remove these two tests since the latter two tests test the same exact thing.

pkg/apis/v1beta1/taints.go
njtran: > The trouble with tolerations is that you can only tolerate *.

I just tried this out on my cluster and saw tolerations apply with specific non * values. 

I can understand the grumpy sysadmins case. We plan on adding this taint in, but I think the only other two taints would be mutually exclusive with this taint and definitely fully documented when we progress on those.

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (approved):
LGTM 🚀

pkg/apis/v1beta1/taints.go
sftim: Oh, yeah, that was unclear. I meant that `*` is the only wildcard available.

PR #516, opened 2023-09-15
docs: add disruption controls design

njtran:
Fixes 

**Description**
This is a request for comments on the disruption controls design.

Once this is implemented, it should solve the following:
- https://github.com/aws/karpenter/issues/1738
- https://github.com/aws/karpenter/issues/1302
- https://github.com/aws/karpenter/issues/4529
- https://github.com/aws/karpenter/issues/3059

Please leave your comments! 


designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
+# Disruption Controls
+
+Karpenter is including new Disruption Controls in its `v1beta1` APIs. These APIs include a new `Disruption` block in the `NodePool` that will contain all scale-down behavioral fields. Initially, this will include `Budgets`, which define a (1) parallelism of how many nodes can be deprovisioned at a time and a (2) cron schedule that determines when the parallelism applies. Disruption Controls will also include a `ConsolidateAfter` field that will allow users to affect the speed at which Karpenter will scale down underutilized nodes.
+
+## Proposed Spec
+
+This only includes the `Budgets` and `ConsolidateAfter` portion of the disruption spec.
+
+```{yaml}
```
jonathan-innis, line 9: Huh, I've never done brackets like this. Are these needed or is it just a style choice whether you have them or don't?

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec:
+  consolidationPolicy: WhenUnderutilized || WhenEmpty
```
jonathan-innis, line 15: These fields should be in the disruption block

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+    Crontab *string `json:"crontab,omitempty" hash:"ignore"`
+    // Duration determines how long a Budget is active since each Crontab hit.
+    // This is required if Crontab is set.
+    Duration *metav1.Duration `json:"duration,omitempty" hash:"ignore"`
+}
+```
+
+## Validation/Defaults
+
+For each `Budget`, users must set a non-negative `MaxUnavailable`. Users can disable scale down for a NodePool by setting this to `0`. Users must either omit both `Crontab` and `Duration` or set both of them, since `Crontab` and `Duration` are inherently linked. Omitting these two fields will be equivalent to an always active `Budget`. If a user defines seconds for `Duration`, Karpenter will ignore it, effectively rounding down to the nearest minute, since the smallest denomination of time in upstream Crontabs are minutes.
```
jonathan-innis, line 65: If we're not going to respect seconds, we should just not allow them to be specified within our regex

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+
+- An omitted `Budgets` will be defaulted to one `Budget` with `MaxUnavailable: 10%` if left undefined.
```
jonathan-innis, line 67:
```suggestion
- Omitting the field `Budgets` will cause the field to be defaulted to one `Budget` with `MaxUnavailable: 10%`.
```

nit: wording

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+- `ConsolidationPolicy` will be defaulted to `WhenUnderutilized`, with a `consolidateAfter` value of `15s`, which is the same value for Consolidation in v1alpha5.
```
jonathan-innis, line 68: Does `15s` makes sense here? Once this is async per node, does it make more sense for this value to be longer like `1m` to give more time for pods to potentially schedule?

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+    consolidateAfter: 15s
+    budgets:
+    - maxUnavailable: 10%
+```
+
+### Defaulting Alternatives Explored
+
+If Karpenter were to persist a default for the `Crontab` and `Duration` fields, it would need to pick sane values for an “always enabled” Budget. The proposal was to default `Crontab` to `* * * * *` and `Duration` to `1m`, but since `Duration` has no impact outside of the `Crontab`, users who want to set their `Crontab` but forget to set `Duration` may find this a sharp edge.
+
+## Pros and Cons For In-line with the NodePool
```
jonathan-innis, line 87: Could you collapse this section into the next section and call that section something like `API Options`. That way all the solutions are enumerated within the same section.

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
+# Disruption Controls
```
jonathan-innis, line 1: Should this doc cover detail on how we got to refining our disruption controls down to these API fields as well as the use-cases that these controls are going to cover explicitly.

designs/disruption-controls.md
njtran: I did it wrong here, but doing 
```
```go 
```
works at formats/highlights it for you!

designs/disruption-controls.md
njtran: Great idea.

designs/disruption-controls.md
njtran: Good callout. Added some bit about this here.

designs/disruption-controls.md
njtran: Synced offline, the default will be 15s right now, but can be backported if implementation changes this.

blakepettersson (approved):
LGTM

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+    - crontab: "0 9 * * mon-fri"
+      duration: 8h
+      maxUnavailable: 0
+    # Every other time, only allow 10 nodes to be deprovisioned simultaneously
+    - maxUnavailable: 10
+```
+
+## Code Definition
+
+```go
```
engedaam, line 31: Where are we defining `expireAfter` in the struct?

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+    ConsolidationPolicy string `json:"consolidationPolicy" hash:"ignore"`
+}
+// Budget specifies periods of times where Karpenter will restrict the
+// number of Node Claims that can be terminated at a time.
+// Unless specified, a budget is always active.
+type Budget struct {
+    // MaxUnavailable dictates how many Node Claims owned by this NodePool
+    // can be terminating at once.
+    // This only respects and considers nodes with a DeletionTimestamp set.
+    MaxUnavailable intstr.IntOrString `json:"maxUnavailable" hash:"ignore"`
```
charliedmcb, line 53: Wonder if making this resource based instead of node bases makes more sense.

With karpenter nodes having varied resources treating nodes as the same seems potentially risky/counter intuitive. 

I wondering about making this field a % based only which would be a % of the NodePool's current resources.

Would be slightly more difficult logic, but it seems to match karpenters patterning of having different nodes with different compute/resourcing.

charliedmcb, line 53: Currently drift will only rotate one node at a time if it has a workload. 

Will this end up affecting drift and the number of nodes containing a workload which it can rotate in parallel?

charliedmcb (commented):
Great to see the progress on disruption controls! 

Have just a few questions/thoughts on `MaxUnavailable`

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+metadata:
+  name: default
+spec: # This is not a complete NodePool Spec.
+  disruption:
+    consolidationPolicy: WhenUnderutilized || WhenEmpty
+    consolidateAfter: 10m || Never # metav1.Duration
+    expireAfter: 10m || Never # Equivalent to v1alpha5 TTLSecondsUntilExpired
+    budgets:
+    # On Weekdays during business hours, don't do any deprovisioning.
+    - crontab: "0 9 * * mon-fri"
```
sidewinder12s, line 22: Would drift be controlled by this disruption budget?

We've found another pain point for management is that we have some provisioners that scale up/down regularly, so we'd like to try and let them drain out naturally (in this case, they have expensive to rebuild caches that we'd like to not force a rebuild of).

At the same time, we're ok with forcing drifting if we can control when it happens.

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+      duration: 8h
+      maxUnavailable: 0
+    # Every other time, only allow 10 nodes to be deprovisioned simultaneously
+    - maxUnavailable: 10
```
sidewinder12s, line 26: Can a schedule like this optionally force consolidation/drift on nodes with `do-not-evict` pods?

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+    Crontab *string `json:"crontab,omitempty" hash:"ignore"`
+    // Duration determines how long a Budget is active since each Crontab hit.
+    // This is required if Crontab is set.
+    Duration *metav1.Duration `json:"duration,omitempty" hash:"ignore"`
+}
+```
+
+## Validation/Defaults
+
+For each `Budget`, users must set a non-negative `MaxUnavailable`. Users can disable scale down for a NodePool by setting this to `0`. Users must either omit both `Crontab` and `Duration` or set both of them, since `Crontab` and `Duration` are inherently linked. Omitting these two fields will be equivalent to an always active `Budget`. Users cannot define a seconds value in `Duration`, since the smallest denomination of time in upstream Crontabs are minutes.
```
johngmyers, line 67: The document does not define how either the numerator or denominator of the budget are calculated.

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
+# Disruption Controls - (in v1beta1+)
+
+## Motivation
+Users have been asking for more control around the speed of Consolidation and the ability to restrict disruptions to their nodes. The demand for this is reflected by the +1s in [#1738](https://github.com/aws/karpenter/issues/1738). This is increasingly important as clusters can run at a highly varied scales, where some users have reported scale-down behaving too slow at higher scales, or too quick at smaller scales. Additionally, users that want to control when nodes can be disrupted must add PDBs to block node disruptions, add `do-not-evict` annotations on pods for sensitive workloads, or add `do-not-consolidate` annotations on nodes that shouldn't be consolidated. Natively integrating a better way to ensure nodes aren't disrupted will ease the burden that users have to take on with disruption.
```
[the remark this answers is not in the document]
johngmyers: From experience with a controller that did rolling updates under kOps:

We have time-restricted automated node drift/expiration remediation to business hours so that there are personnel around to deal with any issues caused by the disruption.

We currently limit concurrent node replacements to 30. That value is somewhat arbitrary and has not been tuned. With a value much higher than that we encountered problematic behavior stemming from excessive load on the Kubernetes API.

designs/disruption-controls.md
njtran: Great questions. I'll try to tackle them one at a time. 

> Wonder if making this resource based instead of node bases makes more sense.

We've definitely considered it, and this can definitely be done as a follow-up in the future. Ultimately Karpenter considers disrupting nodes on a per-node basis, where currently most disruption mechanisms (expiration, drift, single consolidation) will only disrupt one node at a time. Since Karpenter thinks about this in terms of number of nodes, and not the capacity, it always allows Karpenter to make disruption progress regardless of the instance sizes. 

Using resource-based disruption budgets can get users into funky situations where they might define a hard CPU value (10cpu), where all their instances in the cluster have at least 16cpu of capacity. For application developers who only care if their applications are disrupted or not, they are only guaranteed no disruptions if they have `maxUnavailable` set to 0, opting out all instance types.

> I wondering about making this field a % based only which would be a % of the NodePool's current resources.

For a set of nodes with varied instance sizes, using a percent shouldn't be used to drive a minimum percentage of application disruptions. This is meant to be supplemental to PDBs, where users define the level of downtime their applications can tolerate. It's important to note that this won't be incurring any more downtime than PDBs/do-not-evict/ other blocking mechanisms would allow.

njtran: Yep! I'm working on pulling out some of the synchronous waits we do in code to enable Karpenter to drift/expire/consolidate multiple nodes at a time.

designs/disruption-controls.md
njtran: Yep! I'm working on pulling out some of the synchronous waits we do in code to enable Karpenter to drift/expire/consolidate multiple nodes at a time. https://github.com/aws/karpenter-core/issues/255 is where I'm tracking this. 

The maxUnavailable will restrict all disruption mechanisms and include any other terminating nodes that may have been done manually in calculations.

designs/disruption-controls.md
njtran: There's a some issues in karpenter/#1738 that are intentionally not included in this design doc that would probably handle force-delete cases with `do-not-evict` pods. This Budget is meant to define periods where Karpenter cannot disrupt, rather than periods where Karpenter can more forcefully disrupt. 

Can you share some info about your use-case here?

designs/disruption-controls.md
njtran: Ah good point! You're talking about the case of using a percent with `MaxUnavailable`, right? 

Let's say if a user had 10 nodes, but used a `maxUnavailable` of 52%, that would mean Karpenter could disrupt at most 5.2 nodes at a time. Since `.2` nodes doesn't make sense, `6 nodes` would be violating that budget, but `5` wouldn't be. So it would effectively round down the value. 

Is this what you're talking about?

johngmyers: I'm saying you're not defining what is counted to arrive at the number `10` and the number that can't be more than `5`.

designs/disruption-controls.md
chlunde, line 22: kubernetes cronjobs has added support for time zones for CronJobs. I think you should consider the same here, to adjust the business hours based on daylight savings time.

If the same format is accepted as for CronJobs, I think it would be good to use the same names (`schedule` and `timeZone`)

https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#time-zones

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+## Proposed Spec
+
+```yaml
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec: # This is not a complete NodePool Spec.
+  disruption:
+    consolidationPolicy: WhenUnderutilized || WhenEmpty
```
chlunde, line 17: Did you consider moving this into `budgets` so empty nodes can be deleted after a while during business hours, and full consolidation is done the rest of the time?

chlunde:
Are you planning to be able to extend this to more mechanisms/controls with parameters, like "deprovision nodes if there are more than 20% over-allocation" or "more than 2 nodes of spare capacity"?

designs/disruption-controls.md
```diff
@@ -0,0 +1,118 @@
...
+
+These controls will include a new `Disruption` block in the `v1beta1` `NodePool` that will contain all scale-down behavioral fields. Initially, this will include `Budgets`, which define a (1) parallelism of how many nodes can be deprovisioned at a time and a (2) cron schedule that determines when the parallelism applies, and a `ConsolidateAfter` field that will allow users to affect the speed at which Karpenter will scale down underutilized nodes. Future scale-down behavioral fields should be colocated with these fields.
```
chlunde, line 6: > a `ConsolidateAfter` field that will allow users to affect the speed at which Karpenter will scale down underutilized nodes

can you add more information on this? Is it an interval where consolidation is done every Nth minute?

designs/disruption-controls.md
sidewinder12s: We're effectively trying to automate all normal cluster maintenance like node updates, TTLs, etc but a lot of our customers don't want this triggering at any moment. 

One consolation we've made is that we can do whatever during normal scheduled periods similar to RDS maintenance windows. 

So right now the biggest issues for us is that there is no ability to control drift beyond enable disable globally and the do not evict annotations have no way to be dealt with automatically. Our 2 primary users of the annotation are batch type jobs which will finish so are less of an issue and long running singletons that don't want/can't deal with disruption.

designs/disruption-controls.md
sidewinder12s: I guess we were more interested in the ability to control when drift happens, either through this budget mechanism and/or by Provisioner, rather than speeding up the process.

designs/disruption-controls.md
paulgmiller: @njtran  how would you feel about maxUnavailable pertaining to pods rather than nodes? 
The peculiar thing about using nodes here is they're theoretically different sizes aren't they? So I might need to replace ones that have drifted but move more than the majority of my workload because the first N nodes I replace happen to be large ones as opposed to small ones? I think pod percenatge would slow this down. 

May be off base here though.

njtran: You should use the Pod Disruption Budgets to control the level of `maxUnavailable` that you want to enforce for your applications. You won't need to manually replace any of your pods/nodes as that's all programmatically done by Karpenter.

designs/disruption-controls.md
njtran: Drift will happen any time you change the resolved resources in the cloud provider or change the CRDs. Is Karpenter drifting too often for you? If you set the Budget Crontab to restrict disruptions at times where you don't want drift to happen, you can still set a disruption budget for any other time where the previous crontab isn't active.

designs/disruption-controls.md
njtran: We're using this setting to drive what type of cost-consolidation method's can be done for nodes owned by this provisioner. We've heard that users may want to set `actions` that determine which types of disruption are restricted, which I think is what you're asking for. We're opting to explore that as a follow-up to this.

njtran:
@chlunde 
> Are you planning to be able to extend this to more mechanisms/controls with parameters, like "deprovision nodes if there are more than 20% over-allocation" or "more than 2 nodes of spare capacity"?

The hope was that defining a `disruption` block in the NodePool would make future features possibly like these to have a sensible place to live. We should evaluate what each of these features mean and how/if we should implement them, but this is definitely where they would live.

designs/disruption-controls.md
njtran: I've intentionally left some information out of this because the exact implementation of this is in-flux. The TLDR is that when Karpenter detects node(s) as underutilized, which is done one batch at a time, Karpenter will wait for the `consolidateAfter` before consolidating those node(s).

designs/disruption-controls.md
sidewinder12s: Unfortunately, we have quite a few workload types that are singletons or cannot handle any disruption. Or have expensive caches that we'd like to not disrupt during the day. Being able to use the budget crontab would thus be a nice operational improvement. 

Generally, we roll changes to the AMI or Provider and just wanted more control over how it gets rolled out.

ellistarn: We'd discussed scoping specific budgets for specific types of actions. For example, you could constrain drift on weekends to 0, but allow consolidation of 1 at a time.

sidewinder12s: At least for the discussions we've had, just being able to fit drift into a schedule was fine. Our current issue was just that there was no control for drift.

ellistarn: And to be clear, you'd only want to constrain drift, not other kinds of deprovisioning at that time.

designs/disruption-controls.md
njtran: It should be in the same place as the other fields now 👍

designs/disruption-controls.md
njtran: Yep! In the go struct spec, It's called out that     `// Crontab uses the same syntax as a Cronjob`. But I can definitely update it so that's more clear.

designs/disruption-controls.md
paulgmiller: @njtran  Already in the Pod Disruption Budgets fanclub but most clusters only have a few for their most imporant applications. There is still often an expectation of cluster operators that they move workloads to new nodes gradually (may be jobs or other transient workload) . Eventually you will also get an ask to pause between chunks of movement to give alerts time to detect funniness. 

Number or percent of Nodes was a good proxy for this in the past when usually pools of nodes were fairly homogenous but with karpenter they can be quite different and there's not a great way of saying replacing these 3 HUGE machines causes much more churn than 

My hope is that pod count is a a good replacement proxy in the karpenter world. if you move 5% of pods at random even without the dilligence of having everything protecte with PDBs you stand a good chance of letting new pods come up and detecting issues when they exist. 

My 2 cents hope that was more clear.

 - Miller

designs/disruption-controls.md
njtran: Karpenter will only be aware of nodes that have the node pool name label on and are created by Karpenter. So 10 in this case would be the number of nodes already created by Karpenter for this node pool. The number that we'll check in reference is how many of those nodes have a non-nil deletion timestamp, meaning they're actively being deleted.

johngmyers: That information should be put in the document.

I would suggest cordoning as a better signal than the deletion timestamp.

johngmyers: By "nodes" do you mean Kubernetes `Node` objects or does it include instances that have been created by Karpenter but which have not yet joined the cluster?

johngmyers: If there are 6 nodes of 10 without a deletion timestamp and 3 of those 6 are in NotReady state, that allows an additional potentially-Ready node to be drift-remediated?

designs/disruption-controls.md
sidewinder12s: I think we'd like the ability, Drift is just top of mind as currently there is no way to constrain it beyond global disable/enable where as a lot of the other options you can control it through things like multiple provisioners.

designs/disruption-controls.md
sidewinder12s: 👍 on what @paulgmiller is describing. 

On a binary level, a lot of alerts have threshold time periods before they'll fire so many times we like making any changes at a controlled rate to ensure anything unexpected has time to hit those thresholds. 

As far as the ask for pod count, as a cluster operator I do not understand (and my users don't really either) all of the interdependencies of their applications. While a PDB can work for the health of an individual deployment, something like controlling total churn in the cluster can help gracefully degrade instead of outright breaking an application when one of these interdependencies causes issues. 

I'd also point out things like kube-proxy where if you have a sudden, large change you can cause issues with cluster stability because of the time it takes for reconciliation (say, 2 large nodes with 100+ pods each vs 2 small nodes with 30 pods each).

njtran: Thanks for the feedback. 

You could imagine extending this to custom resources, ephemeral storage, pods, cpu, memory, or whatever else is used for bin-packing in the cluster.

If we're able to zero the chances that users can get themselves in tough situations with a budget and unfortunate instance type selections, Karpenter will always be able to converge towards an optimal cluster state to save on cost, which is the biggest pro in my mind. 

I'm 100% open to explore expanding this to more resource types in the future, but want to keep the initial implementation as simple as possible, leaving it open for expansion later.

designs/disruption-controls.md
njtran: Will do. The "cordoned" signal is a little tougher to reason about. It's possible that a node could be cordoned by another controller within the cluster, that we shouldn't consider. Right now, Karpenter uses the upstream `kubernetes.io/unschedulable` taint, which many other controllers and pods are aware of. Yet, just because a node is cordoned doesn't mean it's being disrupted. We're exploring changing this to another `disrupting` taint, which I think could be a better way for Karpenter to be aware of this. 

If we migrate to a new Karpenter-specific taint, you're right. I think that would probably the better way to drive it. 


> By "nodes" do you mean Kubernetes Node resources or does it include instances that have been created by Karpenter but which have not yet joined the cluster?

We will only be considering nodes in the cluster that have an associated NodeClaim, and vice versa. We only continue with disruptions at the moment if nodes/nodeClaims are initialized and have been registered with Karpenter.

> If there are 6 nodes of 10 without a deletion timestamp and 3 of those 6 are in NotReady state, that allows an additional potentially-Ready node to be drift-remediated?

Yeah, we should be including ways to handle nodes that are NotReady for an extended period of time, but I think regardless of if a node is Ready or not, if it's drifted or expired, I don't see why we shouldn't try to replace it if the budget allows it. Are you thinking that the 3 NotReady nodes in this case should also contribute to the budget?

johngmyers: I'm thinking that disruption should not progress until sufficient new nodes are at least in the Ready state; nodes in NotReady state probably should not count as satisfying the budget.

There is a gap between the "Ready" state and able to take workloads, but PodDisruptionBudgets should suffice to cover that. I see the node disruption budget as covering churn in supporting systems, such as the Kubernetes API, CNIs, monitoring systems, etc.

You might want to consider the gap between a new node becoming Ready and the end of elevated activity due to workloads piling onto it.

johngmyers: If it's `NodeClaims` that are being counted for the `10`, that should be what is documented. Or if it's the intersection of `Nodes` and `NodeClaims`, likewise.

designs/disruption-controls.md
garvinp-stripe: nit: Should it just be "Budget"? Can you set multiple budgets within a nodepool?

designs/disruption-controls.md
garvinp-stripe: I think the naming of `consolidationPolicy` and  `WhenUnderutilized`  is kind of poor. The reason is node pools doesn't talk about underutilization and through out the docs consolidation is talked about as a method of deprovisioning similar to emptiness and drift and so on. 

I also don't fully understand this new policy since there are 5 different deprovisioning mechanisms but this only touches 2. 

Could the control be more like:
```
disruption:
  deprovisionWhenEmpty: enabled || disabled
  deprovisionWhenExpired: enabled || disabled
  deprovisionedWhenConsolidated: enabled || disabled
  # interruption + drift
```
or 

```
disruption:
  emptiness: 
    enabled: true
    deprovisionAfter: 10m || never
  # the other deprovision methods
```

designs/disruption-controls.md
```diff
@@ -0,0 +1,122 @@
...
+spec:
+  disruption:
+    consolidationPolicy: WhenUnderutilized
+    consolidateAfter: 15s
+    expireAfter: 30d
+    budgets:
+    - maxUnavailable: 10%
+```
+Karpenter will not persist a default for the `Crontab` and `Duration` fields.
+
```
garvinp-stripe, line 90: I think this isn't cover but would love to see is how are we exposing these interactions either via metrics or events or both. I think Karpenter is already difficult enough to understand why it is making certain choices and this would make it even more difficult before any more complicated resource level or finer disruption budgets. In an ideal world, operators should be able to built a dashboard where at a glance they can understand why Node A isn't consolidated because of reasons X,Y,Z or why Node B will be consolidate in N minutes.

njtran: Totally, I was hoping to keep this as a high level doc about the feature itself. We'll definitely be including metrics/events/statuses as we're always trying to make this better. As a start we could include how many disruptions are allowed at a given time for a NodePool into the status. Anything you had in mind here?

designs/disruption-controls.md
njtran: Yeah, you can use multiple budgets.

designs/disruption-controls.md
garvinp-stripe: I think if we can tied each deprovisioning method as a status to a nodeclaim and maybe mentioned that this machine isn't deprovisionable because of NDP would be very helpful

designs/disruption-controls.md
njtran: We haven't made the changes to the docs yet, as this design relates to v1beta1 APIs, and the v1beta1 APIs haven't been released. We'll make the doc changes before those are released. We'll definitely reorganize the docs to match what we're talking about here. 

The idea is that underutilization encapsulates the current emptiness and consolidation, but consolidation also does emptiness checks. Calling it `consolidationPolicy` will allow users to choose how they want their nodes Consolidated, mapping to the two existing v1alpha5 APIs. `WhenUnderutilized` is a more general way to say "if it saves on cost" whereas `WhenEmpty` is very specific in that we'll only delete empty nodes. 

We're also trying to move away from the `deprovision` language, as we only used that because the API was called the Provisioner, where "deprovisioning" is the opposite. Using `disruption` is more akin to Kubernetes concepts (e.g. PDBs, [other disruptions](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/))

Some fields in the past had the semantic of omitting the field meant disabled, but that meant we couldn't persist defaults we wanted, requiring another field for toggling. Using the `10m || Never` helps us do all of these with one field. The idea is that a new user for Karpenter shouldn't have to think about these concepts and should be able to rely on Karpenter picking sane defaults, but allow users who don't want them to disable/modify them.

garvinp-stripe: Make sense thanks

designs/disruption-controls.md
njtran: I've added in that it's tracked based off nodeClaims and that it only refers to nodes with the `karpenter.sh/disruption` taint.

ellistarn:
Is this ready to merge?

ellistarn (approved):

designs/disruption-controls.md
njtran: I think https://github.com/aws/karpenter/issues/1526 would be the best for you here, as you ensure Karpenter doesn't disrupt these important applications, but can ensure that this do-not-disrupt pod will be evicted once it's reached an expected grace period.

Generally, we want drift to be the default experience on how Karpenter works, but enable users to determine when they want their nodes disrupted through enough pod/node annotations or conditions. 

More granular pod/node level controls is something we're looking to tackle as a second round of disruption controls after this first round.

PR #521, opened 2023-09-18
chore: Require `spec.template.spec.requirements` to be set and change `spec.template.spec.kubelet` field name in memory

jonathan-innis:
Fixes #N/A 

**Description**

This PR updates the `v1beta1` APIs in the following ways:
1. Requires the `spec.template.spec.requirements` value in the `v1beta1/NodePool` and the `spec.requirements` value in the `v1beta1/NodeClaim` to get set always. These values can still be set to "wide-open" by directly specifying an empty array like `requirements: []`
2. Updates the in-memory name for the `kubelet` to match the field name at the apiserver

`make presubmit`
`k apply -f pkg/apis/crds` and testing `NodePool` configuration


engedaam (approved):
LGTM!

PR #530, opened 2023-09-20
ci: Set the toolchain kubebuilder install path to be the default one

jonathan-innis:
Fixes #N/A 

**Description**

This PR updates the installation path for kubebuilder to change it from a path in your home directory to the default expected installation path `/usr/local/kubebuilder/bin`. This means that you no longer need to set the `KUBEBUILDER_ASSETS` variable in testing

`make presubmit`

### DIFF
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 K8S_VERSION="${K8S_VERSION:="1.27.x"}"
-KUBEBUILDER_ASSETS="${KUBEBUILDER_ASSETS:="${HOME}/.kubebuilder/bin"}"
+KUBEBUILDER_ASSETS="/usr/local/kubebuilder/bin"
 
 main() {
     tools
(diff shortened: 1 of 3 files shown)


hack/toolchain.sh
```diff
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 K8S_VERSION="${K8S_VERSION:="1.27.x"}"
-KUBEBUILDER_ASSETS="${KUBEBUILDER_ASSETS:="${HOME}/.kubebuilder/bin"}"
+KUBEBUILDER_ASSETS="${KUBEBUILDER_ASSETS:="/usr/local/kubebuilder/bin"}"
```
ellistarn, line 5: Do you even need to set this, if it's the internal default for kubebuilder?

jonathan-innis: We do a make directory and a find call in this script, so I think this _does_ need to be set. I could go either way on whether the variable should be statically defined or not though. Thoughts?

ellistarn (approved):

PR #533, opened 2023-09-21
fix getDaemonSet to preserve pod affinity in daemonset template pod spec.

### LINKED ISSUE #532: Daemonsets that would not be provisioned on a provisioner are included in the scheduler calculation
### Description

**Observed Behavior**:
Daemonsets that would not be provisioned on a provisioner due to pod template affinity rules ( **not selector labels** ) are also included in the scheduler calculation in [getDaemonOverhead](https://github.com/aws/karpenter-core/blob/c6fb5c1c53220289fd77cdf07fb74f0250245c77/pkg/controllers/provisioning/scheduling/scheduler.go#L313) 

This issue occurs due to multiple reasons:

1. Daemonset pods are not generated from template but instead retrieved from cache of existing pods due to #3199 
2. Daemonset admission controller would replace the Pod nodeAffinity on creation to one that match to only 1 node:
> spec:
>   affinity:
>     nodeAffinity:
>       requiredDuringSchedulingIgnoredDuringExecution:
>         nodeSelectorTerms:
>         - matchFields:
>           - key: metadata.name
>             operator: In
>             values:
>             - ip-10-110-217-205.ap-southeast-1.compute.internal
3. Multiple daemonsets targeting different provisioners using node affinities and not selectors are running in the same namespace.

Thus, since no node selectors are defined, and all Pod affinity rules are rewritten by the admission controller, losing the information of nodeAffinity matchExpressions, all the daemonsets resources are considered in the overhead calculation process.

We can get around this with using proper Node Selectors, but we also wanted to utilise the NotIn operator for matchExpressions in nodeAffinity, and could not do the same with Node Selectors

**Expected Behavior**:
NodeAffinity defined in daemonset Pod templates should be respected even after admission controllers change them. 

1. Create 1 provisioner(A) with instance types for 4cpu and labels `foo : bar`
2. Create another provisioner (B) with instance types for 2cpu and labels `foo: voo`
3. Create Daemonset(A) with requirements for 3cpu and select ProvisionerA using nodeAffinity rules within the Pod template:
> apiVersion: apps/v1
> kind: DaemonSet
> spec:
>   template:
>       spec:
>         affinity:
>           nodeAffinity:
>             requiredDuringSchedulingIgnoredDuringExecution:
>               nodeSelectorTerms:
>               - matchExpressions:
>                 - key: foo
>                   operator: In
>                   values:
>                   - bar
> 
4. Create Pod which targets provisioner B using nodeAffinity rules. 
5. Pod cannot be scheduled on provisioner B even though Daemonset A does not target provisioner B.

**Versions**:
- Chart Version: v0.30.0
- Kubernetes Version (`kubectl version`): 1.23

shikai93:
Fixes #532  

**Description**
Change getDaemonSetPods method to merge in daemon set template pod spec affinities with existing pod affinities to retain this information during scheduling. This would allow scheduler to not include daemonsets that would not be scheduled on a provisioner within their calculations of daemonset overhead

Unit tests included in the PR

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -416,17 +416,15 @@ func (p *Provisioner) getDaemonSetPods(ctx context.Context) ([]*v1.Pod, error) {
 		pod := p.cluster.GetDaemonSetPod(&d)
 		if pod == nil {
 			pod = &v1.Pod{Spec: d.Spec.Template.Spec}
-		} else if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil &&
-			d.Spec.Template.Spec.Affinity != nil &&
-			d.Spec.Template.Spec.Affinity.NodeAffinity != nil &&
-			d.Spec.Template.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
-
-			nodeAffinity := pod.Spec.Affinity.NodeAffinity
-			if nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
-				nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms = d.Spec.Template.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
-			} else {
-				nodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = d.Spec.Template.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
+		}
+		if d.Spec.Template.Spec.Affinity != nil && d.Spec.Template.Spec.Affinity.NodeAffinity != nil && d.Spec.Template.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
+			if pod.Spec.Affinity == nil {
+				pod.Spec.Affinity = &v1.Affinity{}
 			}
+			if pod.Spec.Affinity.NodeAffinity == nil {
+				pod.Spec.Affinity.NodeAffinity = &v1.NodeAffinity{}
+			}
+			pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution = d.Spec.Template.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution
 		}
 		return pod
 	}), nil
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -417,6 +417,9 @@ func (p *Provisioner) getDaemonSetPods(ctx context.Context) ([]*v1.Pod, error) {
 		if pod == nil {
 			pod = &v1.Pod{Spec: d.Spec.Template.Spec}
 		}
+		// Replacing retrieved pod affinity with daemonset pod template required node affinity since this is overridden
+		// by the daemonset controller during pod creation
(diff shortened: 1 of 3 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -416,11 +416,144 @@ func (p *Provisioner) getDaemonSetPods(ctx context.Context) ([]*v1.Pod, error) {
 		pod := p.cluster.GetDaemonSetPod(&d)
 		if pod == nil {
 			pod = &v1.Pod{Spec: d.Spec.Template.Spec}
+		} else {
+			mergedAffinity := p.MergeAffinities(d.Spec.Template.Spec.Affinity, pod.Spec.Affinity)
```
jonathan-innis, line 420: Does it make sense to merge the affinities or just use the DaemonSet node affinity instead of the node affinity that comes from the pod, since that one will always get overridden by the DaemonSet controller as shown here: https://github.com/kubernetes/kubernetes/blob/c5cf0ac1889f55ab51749798bec684aed876709d/pkg/controller/daemon/util/daemonset_util.go#L176.

I think that could help simplify some things. You basically take every detail of the pod, except for the Required Node Affinities which you expect to be overridden by the DaemonSet controller.

shikai93: I was attempting to do a thorough merge in the event that kubernetes behaviour ever changes again in the future, but i do agree that it could simplify some stuff too, let me make the change

jonathan-innis: It seems unlikely to me that they will make a change there and (even if they did) we would just update our logic then when we see the logic changes in the new K8s version changelog

jonathan-innis (commented):
Just one question

pkg/controllers/provisioning/provisioner.go
```diff
@@ -416,6 +416,17 @@ func (p *Provisioner) getDaemonSetPods(ctx context.Context) ([]*v1.Pod, error) {
 		pod := p.cluster.GetDaemonSetPod(&d)
 		if pod == nil {
 			pod = &v1.Pod{Spec: d.Spec.Template.Spec}
+		} else if pod.Spec.Affinity != nil && pod.Spec.Affinity.NodeAffinity != nil &&
```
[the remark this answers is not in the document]
shikai93: agreed changed in the newer version 🙏

pkg/controllers/provisioning/provisioner_test.go
```diff
@@ -537,6 +536,78 @@ var _ = Describe("Provisioning", func() {
 			Expect(*allocatable.Cpu()).To(Equal(resource.MustParse("4")))
 			Expect(*allocatable.Memory()).To(Equal(resource.MustParse("4Gi")))
 		})
+		It("should account for daemonset spec affinity ", func() {
+			ExpectCleanedUp(ctx, env.Client)
+			cluster.Reset()
```
jonathan-innis, line 541: This call and L540 get called during the AfterEach section for each Ginkgo test so we shouldn't need to execute them here.

pkg/controllers/provisioning/provisioner_test.go
```diff
@@ -537,6 +536,78 @@ var _ = Describe("Provisioning", func() {
 			Expect(*allocatable.Cpu()).To(Equal(resource.MustParse("4")))
 			Expect(*allocatable.Memory()).To(Equal(resource.MustParse("4Gi")))
 		})
+		It("should account for daemonset spec affinity ", func() {
```
jonathan-innis, line 539:
```suggestion
		It("should account for daemonset spec affinity", func() {
```

jonathan-innis (commented):
LGTM, just a couple things in the testing file

pkg/controllers/provisioning/provisioner_test.go
shikai93: yes i forgot to clean them up as i was debugging the unit test setup. thanks for pointing out

jonathan-innis (approved):
LGTM 🚀

tzneal (approved):
lgtm

PR #541, opened 2023-09-26
chore: karpenter-global-settings migration

jmdeal:
Fixes #N/A 

**Description**
This PR is the first step in removing the karpenter-global-config configuration map in favor of CLI flags and environment variables as part of v1beta1. It currently respects both values set by the cli flags / env and the configmap, but cli flags / env take precedent.

Manual testing with Karpenter deployment and `make test`

### DIFF
--- a/pkg/apis/settings/settings.go
+++ b/pkg/apis/settings/settings.go
@@ -30,7 +30,7 @@ var ContextKey = settingsKeyType{}
 
 var defaultSettings = &Settings{
 	BatchMaxDuration:  time.Second * 10,
-	BatchIdleDuration: time.Second * 1,
+	BatchIdleDuration: time.Second,
 	DriftEnabled:      false,
 }
 
@@ -49,7 +49,6 @@ func (*Settings) ConfigMap() string {
 // Inject creates a Settings from the supplied ConfigMap
 func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context, error) {
 	s := defaultSettings.DeepCopy()
-
 	if err := configmap.Parse(cm.Data,
 		configmap.AsDuration("batchMaxDuration", &s.BatchMaxDuration),
 		configmap.AsDuration("batchIdleDuration", &s.BatchIdleDuration),
@@ -80,8 +79,7 @@ func ToContext(ctx context.Context, s *Settings) context.Context {
 func FromContext(ctx context.Context) *Settings {
 	data := ctx.Value(ContextKey)
 	if data == nil {
-		// This is developer error if this happens, so we should panic
-		panic("settings doesn't exist in context")
+		return nil
 	}
 	return data.(*Settings)
 }
--- a/pkg/operator/injection/injection.go
+++ b/pkg/operator/injection/injection.go
@@ -30,22 +30,7 @@ import (
 	"knative.dev/pkg/system"
 
 	"github.com/aws/karpenter-core/pkg/apis/settings"
-	"github.com/aws/karpenter-core/pkg/operator/options"
 )
-
-type optionsKey struct{}
-
-func WithOptions(ctx context.Context, opts options.Options) context.Context {
-	return context.WithValue(ctx, optionsKey{}, opts)
-}
-
-func GetOptions(ctx context.Context) options.Options {
-	retval := ctx.Value(optionsKey{})
-	if retval == nil {
-		return options.Options{}
-	}
-	return retval.(options.Options)
-}
 
 type controllerNameKeyType struct{}
 
@@ -68,7 +53,7 @@ func GetControllerName(ctx context.Context) string {
 // dynamically at runtime due to the necessity of having to build logic around re-queueing to ensure that settings are
 // properly reloaded for things like feature gates
 func WithSettingsOrDie(ctx context.Context, kubernetesInterface kubernetes.Interface, settings ...settings.Injectable) context.Context {
-	cancelCtx, cancel := context.WithCancel(ctx)
+	cancelCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
 	defer cancel()
 
 	factory := informers.NewSharedInformerFactoryWithOptions(kubernetesInterface, time.Second*30, informers.WithNamespace(system.Namespace()))
@@ -76,7 +61,14 @@ func WithSettingsOrDie(ctx context.Context, kubernetesInterface kubernetes.Inter
 	factory.Start(cancelCtx.Done())
 
 	for _, setting := range settings {
-		cm := lo.Must(WaitForConfigMap(ctx, setting.ConfigMap(), informer))
+		cm, err := WaitForConfigMap(ctx, setting.ConfigMap(), informer)
+		if err != nil {
+			if !errors.IsNotFound(err) {
+				panic(fmt.Errorf("failed to get configmap %s, %w", setting.ConfigMap(), err))
+			}
+			continue
+		}
+
 		ctx = lo.Must(setting.Inject(ctx, cm))
 	}
 	return ctx
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -78,8 +78,9 @@ func NewOperator() (context.Context, *Operator) {
 
 	// Options
 	opts := options.New().MustParse(os.Args[1:]...)
-	ctx = options.ToContext(ctx, opts)
 
+	// Make the Karpenter binary aware of the container memory limit
+	// https://pkg.go.dev/runtime/debug#SetMemoryLimit
 	if opts.MemoryLimit > 0 {
 		newLimit := int64(float64(opts.MemoryLimit) * 0.9)
 		debug.SetMemoryLimit(newLimit)
@@ -108,7 +109,8 @@ func NewOperator() (context.Context, *Operator) {
 
 	// Inject settings from the ConfigMap(s) into the context
 	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
-	opts.MergeSettings(settings.FromContext(ctx))
+	opts = opts.MergeSettings(settings.FromContext(ctx))
+	ctx = options.ToContext(ctx, opts)
 
 	// Manager
 	mgr, err := controllerruntime.NewManager(config, controllerruntime.Options{
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -77,7 +77,7 @@ func NewOperator() (context.Context, *Operator) {
 	ctx = knativeinjection.WithNamespaceScope(ctx, system.Namespace())
 
 	// Options
-	opts := options.New().MustParse(os.Args[1:]...)
+	opts := lo.Must(options.New().Parse(os.Args[1:]...))
 
 	// Make the Karpenter binary aware of the container memory limit
 	// https://pkg.go.dev/runtime/debug#SetMemoryLimit
@@ -102,15 +102,15 @@ func NewOperator() (context.Context, *Operator) {
 	// Client
 	kubernetesInterface := kubernetes.NewForConfigOrDie(config)
 
-	// Logging
-	logger := logging.NewLogger(ctx, component, kubernetesInterface)
-	ctx = knativelogging.WithLogger(ctx, logger)
-	logging.ConfigureGlobalLoggers(ctx)
-
 	// Inject settings from the ConfigMap(s) into the context
 	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
 	opts = opts.MergeSettings(settings.FromContext(ctx))
 	ctx = options.ToContext(ctx, opts)
+
+	// Logging
+	logger := logging.NewLogger(ctx, component, kubernetesInterface)
+	ctx = knativelogging.WithLogger(ctx, logger)
+	logging.ConfigureGlobalLoggers(ctx)
 
 	// Manager
 	mgr, err := controllerruntime.NewManager(config, controllerruntime.Options{
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -18,6 +18,8 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"os"
+	"runtime/debug"
 	"sync"
 	"time"
 
@@ -42,6 +44,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	"github.com/aws/karpenter-core/pkg/apis"
+	"github.com/aws/karpenter-core/pkg/apis/settings"
 	"github.com/aws/karpenter-core/pkg/apis/v1alpha5"
 	"github.com/aws/karpenter-core/pkg/events"
 	"github.com/aws/karpenter-core/pkg/operator/controller"
@@ -74,8 +77,13 @@ func NewOperator() (context.Context, *Operator) {
 	ctx = knativeinjection.WithNamespaceScope(ctx, system.Namespace())
 
 	// Options
-	opts := options.New().MustParse()
-	ctx = injection.WithOptions(ctx, *opts)
+	opts := options.New().MustParse(os.Args[1:]...)
+	ctx = options.ToContext(ctx, opts)
+
+	if opts.MemoryLimit > 0 {
+		newLimit := int64(float64(opts.MemoryLimit) * 0.9)
+		debug.SetMemoryLimit(newLimit)
+	}
 
 	// Webhook
 	ctx = webhook.WithOptions(ctx, webhook.Options{
@@ -100,6 +108,7 @@ func NewOperator() (context.Context, *Operator) {
 
 	// Inject settings from the ConfigMap(s) into the context
 	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
+	opts.MergeSettings(settings.FromContext(ctx))
 
 	// Manager
 	mgr, err := controllerruntime.NewManager(config, controllerruntime.Options{
@@ -114,8 +123,8 @@ func NewOperator() (context.Context, *Operator) {
 		BaseContext: func() context.Context {
 			ctx := context.Background()
 			ctx = knativelogging.WithLogger(ctx, logger)
+			ctx = options.ToContext(ctx, opts)
 			ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
-			ctx = injection.WithOptions(ctx, *opts)
 			return ctx
 		},
 		NewCache: cache.BuilderWithOptions(cache.Options{
@@ -163,7 +172,7 @@ func (o *Operator) WithControllers(ctx context.Context, controllers ...controlle
 }
 
 func (o *Operator) WithWebhooks(ctx context.Context, ctors ...knativeinjection.ControllerConstructor) *Operator {
-	if !injection.GetOptions(ctx).DisableWebhook {
+	if !options.FromContext(ctx).DisableWebhook {
 		o.webhooks = append(o.webhooks, ctors...)
 		lo.Must0(o.Manager.AddReadyzCheck("webhooks", webhooks.HealthProbe(ctx)))
 		lo.Must0(o.Manager.AddHealthzCheck("webhooks", webhooks.HealthProbe(ctx)))
@@ -178,7 +187,7 @@ func (o *Operator) Start(ctx context.Context) {
 		defer wg.Done()
 		lo.Must0(o.Manager.Start(ctx))
 	}()
-	if injection.GetOptions(ctx).DisableWebhook {
+	if options.FromContext(ctx).DisableWebhook {
 		knativelogging.FromContext(ctx).Infof("webhook disabled")
 	} else {
 		wg.Add(1)
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -139,9 +139,9 @@ func (o *Options) MustParse(args ...string) *Options {
 
 // MergeSettings applies settings specified in the v1alpha5 configmap to options. If the value was already specified by
 // a CLI argument or environment variable, that value will be used.
-func (o *Options) MergeSettings(s *settings.Settings) {
+func (o *Options) MergeSettings(s *settings.Settings) *Options {
 	if s == nil {
-		return
+		return o
 	}
 
 	// Note: settings also has default values applied to it. If the option is specified by neither Settings nor Options,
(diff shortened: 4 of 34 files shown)


pkg/operator/injection/injection.go
```diff
@@ -76,36 +61,49 @@ func GetControllerName(ctx context.Context) string {
...
-		if configMap != nil && exists && err == nil {
-			return configMap.(*v1.ConfigMap), nil
-		}
-		select {
-		case <-ctx.Done():
-			return nil, fmt.Errorf("context canceled")
-		case <-time.After(time.Millisecond * 500):
-		}
+// getConfigMap waits until all registered configMaps in the settingsStore are created
+func getConfigMap(name string, informer cache.SharedIndexInformer) (*v1.ConfigMap, error) {
```
jonathan-innis, line 100: We discussed this but I don't think that we need to change this logic at all. We can just create a context that has a timeout and then based on if we don't resolve the ConfigMap, we can continue forward. Specifically, we should check if we are getting NotFound errors and not just transient errors from the kube client API calls. See: https://github.com/aws/karpenter-core/pull/544/files#diff-49fe45488dc421bbc55f13e760b3b06b8755016ae17c70915eb0b44752b5f476R96

pkg/operator/operator.go
```diff
@@ -105,7 +105,7 @@ func NewOperator() (context.Context, *Operator) {
 	ConfigureGlobalLoggers(ctx)
 
 	// Inject settings from the ConfigMap(s) into the context
-	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
+	ctx = injection.WithSettings(ctx, kubernetesInterface, apis.Options, apis.Settings...)
```
jonathan-innis, line 108: After this call is where I'd like to see a call like `MergeSettingsIntoOptions`

pkg/operator/options/injectable.go
```diff
@@ -0,0 +1,26 @@
+/*
```
jonathan-innis, line 1: I think we can drop this file

pkg/operator/options/options.go
```diff
@@ -57,23 +83,113 @@ func New() *Options {
 	f.BoolVar(&opts.EnableLeaderElection, "leader-elect", env.WithDefaultBool("LEADER_ELECT", true), "Start leader election client and gain leadership before executing the main loop. Enable this when running replicated components for high availability.")
 	f.Int64Var(&opts.MemoryLimit, "memory-limit", env.WithDefaultInt64("MEMORY_LIMIT", -1), "Memory limit on the container running the controller. The GC soft memory limit is set to 90% of this value.")
 
+	// Vars that must be merged with settings
+	f.DurationVar(&opts.BatchMaxDuration, "batch-max-duration", env.WithDefaultDuration("BATCH_MAX_DURATION", 10*time.Second), "TODO")
```
jonathan-innis, line 87: I see it's marked as a TODO, just a reminder to restore these descriptions

pkg/operator/options/options.go
```diff
@@ -57,23 +83,113 @@ func New() *Options {
...
 	return opts
 }
 
-// MustParse reads the user passed flags, environment variables, and default values.
-// Options are valided and panics if an error is returned
-func (o *Options) MustParse() *Options {
-	err := o.Parse(os.Args[1:])
-
-	if errors.Is(err, flag.ErrHelp) {
+func (*Options) Inject(ctx context.Context) (context.Context, error) {
```
jonathan-innis, line 99: Any reason to change this from `MustParse`

pkg/operator/options/options.go
```diff
@@ -57,23 +83,113 @@ func New() *Options {
...
+	opts := New()
+	if err := opts.Parse(os.Args[1:]); errors.Is(err, flag.ErrHelp) {
 		os.Exit(0)
+	} else if err != nil {
+		return ctx, err
+	}
+
+	// Check if BatchMaxDuration and BatchIdleDuration have been set. If they have not been set, they will be overwritten by
+	// values in the configmap, if present.
+	opts.Visit(func(f *flag.Flag) {
```
jonathan-innis, line 109: Looks like you found a good way to handle this natively with opts.Visit 👏

pkg/operator/options/options.go
```diff
@@ -57,23 +83,113 @@ func New() *Options {
...
+func FromContext(ctx context.Context) *Options {
+	retval := ctx.Value(optionsKey{})
+	if retval == nil {
+		// This is a developer error if this happens, so we should panic
+		panic("options doesn't exist in context")
+	}
+	return retval.(*Options)
+}
+
+func parseFeatureGates(val string) map[string]bool {
```
jonathan-innis, line 176: How does the upstream apiserver parse these args? Is it the same method? Can we leverage their parsing logic?

pkg/apis/apis.go
```diff
@@ -35,6 +36,7 @@ var (
 	)
 	// AddToScheme may be used to add all resources defined in the project to a Scheme
 	AddToScheme = Builder.AddToScheme
+	Options     = []options.Injectable{&options.Options{}}
```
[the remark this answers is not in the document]
jmdeal: Discussed offline, going to drop the `Injectable` abstraction 👍

pkg/apis/settings/settings.go
```diff
@@ -63,6 +56,10 @@ func (*Settings) Inject(ctx context.Context, cm *v1.ConfigMap) (context.Context,
 	return ToContext(ctx, s), nil
 }
 
+func (*Settings) FromContext(ctx context.Context) Injectable {
```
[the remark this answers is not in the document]
jmdeal: This was to enable merging settings into all options from Karpenter core, this will be dropped with the Injectable abstraction.

pkg/operator/options/options.go
jmdeal: Following the `Injectable` pattern, will change back since we're dropping that.

pkg/operator/options/options.go
```diff
@@ -59,23 +85,101 @@ func New() *Options {
...
+
+// MergeSettings applies settings specified in the v1alpha5 configmap to options. If the value was already specified by
+// a CLI argument or environment variable, that value will be used.
+func MergeSettings(ctx context.Context, s *settings.Settings) {
+	o := FromContext(ctx)
+	if !o.BatchMaxDurationSet {
+		o.BatchMaxDuration = s.BatchMaxDuration
+		o.BatchMaxDurationSet = true
+	}
+	if !o.BatchIdleDurationSet {
```
jonathan-innis, line 143: Same comment here. We should check each setting is overridable individually, or consider specifying the defaults at the settings struct (or both places potentially) and only override the options if the option wasn't explicitly set by the user.

pkg/operator/injection/injection.go
```diff
@@ -68,15 +53,23 @@ func GetControllerName(ctx context.Context) string {
...
 	defer cancel()
 
 	factory := informers.NewSharedInformerFactoryWithOptions(kubernetesInterface, time.Second*30, informers.WithNamespace(system.Namespace()))
 	informer := factory.Core().V1().ConfigMaps().Informer()
 	factory.Start(cancelCtx.Done())
 
 	for _, setting := range settings {
-		cm := lo.Must(WaitForConfigMap(ctx, setting.ConfigMap(), informer))
+		cm, err := WaitForConfigMap(ctx, setting.ConfigMap(), informer)
+		if err != nil {
```
jonathan-innis, line 65:
```suggestion
		if client.IgnoreNotFound(err) != nil {
```

Can shorten this and drop the second if

pkg/operator/options/options.go
jmdeal: Restoring the defaults to the settings struct makes sense to me, along with keeping the individual field check. I'll go ahead and do that.

pkg/operator/injection/injection.go
jmdeal: Not sure if this is what you had in mind but this is what I think it can be simplified to:
```go
    panic(err)
} else if err != nil {
    continue
}
```
It still needs the two if statements but gets rid of the nesting.

jonathan-innis: Oop. My fault, I missed L80 and I see why you can't drop the if check completely. I actually like the old way that you had it better now that I see L80 😛

pkg/operator/injection/injection.go
```diff
@@ -68,15 +54,21 @@ func GetControllerName(ctx context.Context) string {
 // dynamically at runtime due to the necessity of having to build logic around re-queueing to ensure that settings are
 // properly reloaded for things like feature gates
 func WithSettingsOrDie(ctx context.Context, kubernetesInterface kubernetes.Interface, settings ...settings.Injectable) context.Context {
-	cancelCtx, cancel := context.WithCancel(ctx)
+	cancelCtx, cancel := context.WithTimeout(ctx, 10*time.Second)
```
jonathan-innis, line 57:
```suggestion
	cancelCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
```

I did 5 seconds in my other PR and that should be more than enough time for the ConfigMap to be added and the change to be seen

pkg/operator/injection/injection.go
```diff
@@ -68,15 +54,21 @@ func GetControllerName(ctx context.Context) string {
...
 	defer cancel()
 
 	factory := informers.NewSharedInformerFactoryWithOptions(kubernetesInterface, time.Second*30, informers.WithNamespace(system.Namespace()))
 	informer := factory.Core().V1().ConfigMaps().Informer()
 	factory.Start(cancelCtx.Done())
 
 	for _, setting := range settings {
-		cm := lo.Must(WaitForConfigMap(ctx, setting.ConfigMap(), informer))
+		cm, err := WaitForConfigMap(ctx, setting.ConfigMap(), informer)
```
jonathan-innis, line 65: Made a comment on this from my original comment. Given L80, I think I like the original way that you did this better

pkg/operator/options/options.go
```diff
@@ -59,23 +86,85 @@ func New() *Options {
...
+		o.BatchIdleDurationSet = true
+	}
+
 	return o
 }
+
+// MergeSettings applies settings specified in the v1alpha5 configmap to options. If the value was already specified by
+// a CLI argument or environment variable, that value will be used.
+func (o *Options) MergeSettings(s *settings.Settings) {
+	mergeField(&o.BatchMaxDuration, s.BatchMaxDuration, o.BatchMaxDurationSet)
```
jonathan-innis, line 134: Does this still retain default values? I think the answer is still no if the CLI argument isn't set and the settings value isn't set. You still get a scenario where you have a default value that's stored from the CLI options and you override it with the setting every time, whether or not the setting value is set

jmdeal: Since settings also has default values applied to it, if neither are set it will apply the default from settings. The downside of this approach is that the defaults are defined in two places but the upside is there's no additional logic that needs to be torn out with settings once support is dropped.

pkg/utils/env/env.go
```diff
@@ -84,3 +85,17 @@ func WithDefaultBool(key string, def bool) bool {
 	}
 	return parsedVal
 }
+
+// WithDefaultBool returns the duration value of the supplied environment variable or, if not present,
```
jonathan-innis, line 89:
```suggestion
// WithDefaultDuration returns the duration value of the supplied environment variable or, if not present,
```

pkg/operator/options/options.go
```diff
@@ -44,6 +49,27 @@ type Options struct {
 	EnableLeaderElection bool
 	MemoryLimit          int64
 	LogLevel             string
+	BatchMaxDuration     time.Duration
+	BatchIdleDuration    time.Duration
+	DriftEnabled         bool
```
jonathan-innis, line 54: Does it make sense at all to have a `Gates` field here that stores all the feature gates and to pre-populate these gates with defaults at the time that we initialize the `NewOptions()`?

pkg/operator/options/options.go
```diff
@@ -44,6 +49,27 @@ type Options struct {
...
+
+type Options struct {
+	OptionFields
+	optionsFlags
+
+	featureGates string
+}
+
+var defaultFeatureGates = map[string]bool{
+	"Drift": true,
```
jonathan-innis, line 72: Make this a struct?

pkg/operator/options/options.go
```diff
@@ -66,26 +92,90 @@ func New() *Options {
...
 }
 
-// MustParse reads the user passed flags, environment variables, and default values.
-// Options are valided and panics if an error is returned
-func (o *Options) MustParse() *Options {
-	err := o.Parse(os.Args[1:])
-
-	if errors.Is(err, flag.ErrHelp) {
+func (o *Options) MustParse(args []string) *Options {
+	if err := o.Parse(args); errors.Is(err, flag.ErrHelp) {
```
jonathan-innis, line 112: Tend to prefer the style of
```go
if err := o.Parse(args); err != nil {
  if errors.Is(err, flag.ErrHelp) {
    os.Exit(0)
  }
  panic(err)
}
```

pkg/operator/injection/injection.go
```diff
@@ -68,15 +53,23 @@ func GetControllerName(ctx context.Context) string {
...
+		if err != nil {
+			if errors.IsNotFound(err) {
+				panic(err)
```
jonathan-innis, line 67: Should we provide any additional context to this error?

jonathan-innis: Also, this should be `if !errors.IsNotFound(err)` since that's the only error type that we want to ignore here

pkg/operator/injection/injection.go
```diff
@@ -68,15 +53,23 @@ func GetControllerName(ctx context.Context) string {
...
+			} else {
```
jonathan-innis, line 68: Since the previous `if` is a panic, there's no need for an else here

pkg/operator/injection/injection.go
jmdeal: Yep, that's a miss. It doesn't look like `WaitForConfigMap` provides much context so I think it makes sense to add some here (e.g. `panic(fmt.Errorf("failed to get configmap, %w", err)`).

pkg/operator/options/options.go
jmdeal: Sure, I think that would be a good improvement 👍

pkg/operator/operator.go
```diff
@@ -100,6 +102,9 @@ func NewOperator() (context.Context, *Operator) {
 
 	// Inject settings from the ConfigMap(s) into the context
 	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
+	if s := settings.FromContext(ctx); s != nil {
+		opts.MergeSettings(s)
```
jonathan-innis, line 106: nit: should we just always merge the settings and let the `opts.MergeSettings` function just handle the nil settings?

pkg/operator/options/options.go
```diff
@@ -44,6 +53,23 @@ type Options struct {
...
+	BatchMaxDurationSet  bool
+	BatchIdleDurationSet bool
+	FeatureGatesSet      bool
+}
+
+type Options struct {
+	OptionFields
+	optionsFlags
+
+	featureGateStr string
```
jonathan-innis, line 72: Does it make sense to store featureGatesStr as an unexported element under the featureGates in the struct?

pkg/test/options.go
```diff
@@ -0,0 +1,73 @@
...
+	HealthProbePort      *int
+	KubeClientQPS        *int
+	KubeClientBurst      *int
+	EnableProfiling      *bool
+	EnableLeaderElection *bool
+	MemoryLimit          *int64
+	LogLevel             *string
+	BatchMaxDuration     *time.Duration
+	BatchIdleDuration    *time.Duration
+	FeatureGates         *options.FeatureGates
```
jonathan-innis, line 43: I think we may need to create `test.FeatureGates` here that contains fields that are all pointers. I don't think we can determine if something is set like this since this will override the full featureGates struct even if there are multiple fields there and one of them wasn't set.

pkg/operator/options/suite_test.go
```diff
@@ -0,0 +1,89 @@
...
+	ctx = TestContextWithLogger(t)
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Settings")
+}
+
+var _ = Describe("Options", func() {
+	Context("Parse", func() {
+		DescribeTable(
+			"should successfully parse feature gates",
+			func(flagStr string, expectFunc func(*options.Options)) {
```
jonathan-innis, line 42: If you're having to create an expectation function like this, I think that's a good signal that these should just be It() statements and shouldn't be captured in a table. I personally find the way that this is written difficult to read than if they were their own It statements

pkg/operator/options/suite_test.go
```diff
@@ -0,0 +1,89 @@
...
+				Expect(o.FeatureGates.Drift).To(BeTrue())
+			}),
+			Entry("Drift = false", "Drift=false", func(o *options.Options) {
+				Expect(o.FeatureGates.Drift).To(BeFalse())
+			}),
+		)
+	})
+
+	Context("Merge", func() {
+		DescribeTable(
```
jonathan-innis, line 56: Same comment here. I'd just create separate Its

pkg/operator/options/suite_test.go
```diff
@@ -0,0 +1,89 @@
...
+			Entry("Drift = true", "Drift=true", func(o *options.Options) {
+				Expect(o.FeatureGates.Drift).To(BeTrue())
+			}),
+			Entry("Drift = false", "Drift=false", func(o *options.Options) {
+				Expect(o.FeatureGates.Drift).To(BeFalse())
+			}),
+		)
+	})
+
+	Context("Merge", func() {
```
jonathan-innis, line 55: Should we have some testing for defaults with parsing?

pkg/operator/operator.go
jmdeal: I don't have a strong opinion either way, that's fine with me.

pkg/test/options.go
jmdeal: That's true, I didn't really think about it since there's currently only a single field.

pkg/operator/options/suite_test.go
jmdeal: Ya, I had thought there would be more shared behavior and never separated it out after. Happy to make that change.

pkg/operator/options/suite_test.go
jmdeal: Seems reasonable 👍

pkg/operator/options/options.go
```diff
@@ -66,26 +92,101 @@ func New() *Options {
...
+
+func (o *Options) MustParse(args ...string) *Options {
+	if err := o.Parse(args); err != nil {
+		if errors.Is(err, flag.ErrHelp) {
+			os.Exit(0)
+		}
+		panic(err)
+	}
+
+	if o.MemoryLimit > 0 {
```
jonathan-innis, line 111: This has always felt awkward inside of the parsing logic. Realistically, I think this belongs outside of the `MustParse` at the top-level in `NewOperator` when we are doing setup for everything

jonathan-innis (commented):
Really close! Looking good! 🎉

pkg/operator/operator.go
```diff
@@ -74,8 +77,13 @@ func NewOperator() (context.Context, *Operator) {
 	ctx = knativeinjection.WithNamespaceScope(ctx, system.Namespace())
 
 	// Options
-	opts := options.New().MustParse()
-	ctx = injection.WithOptions(ctx, *opts)
+	opts := options.New().MustParse(os.Args[1:]...)
+	ctx = options.ToContext(ctx, opts)
+
+	if opts.MemoryLimit > 0 {
```
jonathan-innis, line 83: Consider adding a comment like 

```go
// Make the Karpenter binary aware of the container memory limit
// https://pkg.go.dev/runtime/debug#SetMemoryLimit
```

pkg/operator/options/options.go
```diff
@@ -66,26 +91,96 @@ func New() *Options {
...
+	if _, ok := os.LookupEnv("FEATURE_GATES"); ok {
+		o.FeatureGatesSet = true
+	}
+
 	return o
 }
+
+// MergeSettings applies settings specified in the v1alpha5 configmap to options. If the value was already specified by
+// a CLI argument or environment variable, that value will be used.
+func (o *Options) MergeSettings(s *settings.Settings) {
```
jonathan-innis, line 142: Thinking about this a little more: I'm a little concerned about how someone calling this function from the outside would reason about this function mutating the options in place. What are your thoughts about doing something like

```go
func MergeSettingsInto(s *settings.Settings, o *Options) {
  // same logic here
}
```

This makes it clear that you've made a change

pkg/operator/operator.go
```diff
@@ -100,6 +108,7 @@ func NewOperator() (context.Context, *Operator) {
 
 	// Inject settings from the ConfigMap(s) into the context
 	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
+	opts.MergeSettings(settings.FromContext(ctx))
```
jonathan-innis, line 111: The fact that you aren't re-injecting this new options back into the context is _technically_ correct since options is a pointer, but I find it confusing. I'd rather see us just make the call to inject the settings after this function call so that we don't have weird indirection here

pkg/operator/options/options.go
jmdeal: I agree that this could be made more clear. Thoughts on something like this:
```go
func (o *Options) MergeSettings(s *settings.Settings) *options.Options {
  // ...
}

options.ToContext(ctx, options.MergeSettings(settings.FromContext(ctx))
```
I think I like this a little more since it's a consistent signature style with `MustParse` but I'm not against your suggestion either.

jonathan-innis: Looks perfect to me! 🎉

jonathan-innis (dismissed):
LGTM 🚀

pkg/operator/operator.go
```diff
@@ -74,8 +78,17 @@ func NewOperator() (context.Context, *Operator) {
 	ctx = knativeinjection.WithNamespaceScope(ctx, system.Namespace())
 
 	// Options
-	opts := options.New().MustParse()
-	ctx = injection.WithOptions(ctx, *opts)
+	opts, err := options.New().Parse(os.Args[1:]...)
```
jonathan-innis, line 81: `lo.Must()`?

jonathan-innis (approved):
Nice! 🎉 One optional comment

jonathan-innis (approved):
LGTM 🚀

PR #546, opened 2023-09-27
feat: add support for local static pvs

### LINKED ISSUE #545: Support Limits for Static/Local PersistentVolumes
### Description

Karpenter isn't able to detect when a [local volume](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner) won't mount to an existing node due to insufficient PersistentVolumes. Because of this, workarounds like `antiAffinity` on `kubernetes.io/hostname` are needed in order to force scheduling:
```YAML
      affinity:
        podAntiAffinity:
          - labelSelector:
              matchExpressions:
                - key: app
                  operator: In
                  values:
                  - my-app
            topologyKey: "kubernetes.io/hostname"
```

It'd be great if Karpenter could detect the availability of local PVs for existing nodes. 

This gap creates significant scheduling & coordination challenges across applications.

---

alec-rabold:
Fixes #545

**Description**

Detects availability of local PersistentVolumes on existing nodes by matching on its node affinity/selector; the volume limit is determined by how many PVs currently exist on a node.

Deployed [`local-static-provisioner`](https://github.com/kubernetes-sigs/sig-storage-local-static-provisioner/tree/master), ran Karpenter controller locally, applied Karpenter `Provisioner` to target instances with NVMe drives & mount them, and created a `StatefulSet` requesting a local volume:
```YAML
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: dummy
spec:
  serviceName: "dummy"
  replicas: 3
  selector:
    matchLabels:
      app: dummy
  template:
    metadata:
      labels:
        app: dummy
    spec:
      containers:
        - name: dummy
          image: public.ecr.aws/docker/library/alpine:latest
          args:
            - /bin/sh
            - -c
            - "sleep infinity"
          volumeMounts:
            - mountPath: "/scratch"
              name: scratch-volume
      volumes:
        - name: scratch-volume
          ephemeral:
            volumeClaimTemplate:
              spec:
                accessModes: ["ReadWriteOnce"]
                resources:
                  requests:
                    storage: 1Gi
                storageClassName: "nvme-disk"
---
kind: Provisioner
metadata:
  name: nvme
spec:
  consolidation:
    enabled: true
  providerRef:
    name: nvme
  requirements:
      operator: In
      values:
        - c
        - m
        - r
    - key: kubernetes.io/os
      operator: In
      values:
        - linux
    - key: karpenter.k8s.aws/instance-local-nvme
      operator: Gt
      values:
        - "250"
      operator: Gt
      values:
        - "2"
      operator: In
      values:
        - spot
        - on-demand
    - key: kubernetes.io/os
      operator: In
      values:
        - linux
    - key: kubernetes.io/arch
      operator: In
      values:
        - amd64
  ttlSecondsUntilExpired: 604800
---
apiVersion: karpenter.k8s.aws/v1alpha1
kind: AWSNodeTemplate
metadata:
  name: nvme
spec:
  blockDeviceMappings:
    - deviceName: /dev/xvda
      ebs:
        deleteOnTermination: true
        encrypted: true
        iops: 3000
        volumeSize: 100Gi
        volumeType: gp3
  securityGroupSelector:
    karpenter.sh/discovery/cluster/<cluster>: "true"
  subnetSelector:
    kubernetes.io/cluster/<cluster>: <selector>
  userData: |
    # Install NVMe CLI
    yum install nvme-cli -y
    
    # Get list of NVMe Drives
    nvme_drives=$(nvme list | grep "Amazon EC2 NVMe Instance Storage" | cut -d " " -f 1 || true)
    readarray -t nvme_drives <<< "$nvme_drives"
    num_drives=${#nvme_drives[@]}
    
    # Install software RAID utility
    yum install mdadm -y
    
    # Create RAID-0 array across the instance store NVMe SSDs
    mdadm --create /dev/md0 --level=0 --name=md0 --raid-devices=$num_drives "${nvme_drives[@]}"

    # Format drive with Ext4
    mkfs.ext4 /dev/md0
    
    # Get RAID array's UUID
    uuid=$(blkid -o value -s UUID /dev/md0)

    # Create a filesystem path to mount the disk
    mount_location="/mnt/nvme-disks/${uuid}"
    mkdir -p $mount_location
    
    # Mount RAID device
    mount /dev/md0 $mount_location
    
    # Have disk be mounted on reboot
    mdadm --detail --scan >> /etc/mdadm.conf 
    echo /dev/md0 $mount_location ext4 defaults,noatime 0 2 >> /etc/fstab

```

--- 


jonathan-innis (commented):
This is really cool! Looking over this on a first-pass, I think we also need tests added to `pkg/provisioning/nodepool_test.go` to ensure that scheduling works out correctly given this new handling. I'll take a deeper look in my evening time.

alec-rabold:
Thanks for taking a look @jonathan-innis! I've added a few tests to `pkg/controllers/provisioning/scheduling/nodepool_test.go` alongside the others related to volume usage

jonathan-innis:
Apologize for the slowness on this one. We've been pushing to release the beta APIs, donate this repo to `kubernetes-sigs` and I have been at kubecon. Planning to get back to the review of this one soon

arabold-dd:
No worries! Appreciate the update

alec-rabold:
No worries! Appreciate the update

pkg/controllers/state/cluster.go
```diff
@@ -523,6 +525,44 @@ func (c *Cluster) populateInflight(ctx context.Context, n *StateNode) error {
...
+	if err := c.addStaticVolumeLimits(ctx, n); err != nil {
+		return fmt.Errorf("adding static volume limits, %w", err)
+	}
+	if err := c.addDynamicVolumeLimits(ctx, n); err != nil {
+		return fmt.Errorf("adding dynamic volume limits, %w", err)
+	}
+	return nil
+}
+
+func (c *Cluster) addStaticVolumeLimits(ctx context.Context, n *StateNode) error {
```
jonathan-innis, line 537: Do we get ourselves into an infinite scale-up here if the local static provisioner stops doing its job and we have pods that are requesting from a storage class that never succeeds to provisions PVs? From looking at the code here, I think we will probably be fine, because we just won't assume a limit without the presence of any PVs. Should we assume the limit value is 0 here? I'm assuming this really depends on how quickly the local static provisioner brings PVs up with respect to when the node joins.

alec-rabold: That’s a great point.. Looking back through, it does seem like this is held together by explicitly not setting a limit if there’s no PVs. So if a new node is spun up and the pod is waiting for PVs to get created, then Karpenter would continue to nominate that node because it has no limit (yet). 

In that case I think we’re safe from an infinite scale-up, but it does mean that a node where PVs fail to create would be continuously nominated

arabold-dd: True -- either the workload or NodePool would need to specify that requirement. I don't think it's _too_ dissimilar from things like external devices (e.g. if your GPU device plugin fails).

In my opinion, the missing piece is that `kube-scheduler` knows when a local-PV workload won't fit on a given set of nodes; whereas Karpenter doesn't know since they're currently ignored. 

Neither `kube-scheduler` nor Karpenter can be sure of whether a given node will eventually host local PVs, but they can semi-reasonably rule out the ones that are "full". In that case, a node that Karpenter continuously nominates for a local-PV workload could be considered "not-ready". 

It'd be best-effort, though how that's actually implemented in this PR definitely isn't very clear 😄

alec-rabold: True -- either the workload or NodePool would need to specify that requirement. I don't think it's too dissimilar from things like external devices (e.g. if your GPU device plugin fails).

In my opinion, the missing piece is that kube-scheduler knows when a local-PV workload won't fit on a given set of nodes; whereas Karpenter doesn't know since they're currently ignored.

Neither kube-scheduler nor Karpenter can be sure of whether a given node will eventually host local PVs, but they can semi-reasonably rule out the ones that are "full". In that case, a node that Karpenter continuously nominates for a local-PV workload could be considered "not-ready".

It'd be best-effort, though how that's actually implemented in this PR definitely isn't very clear 😄

alec-rabold: For total transparency, I think https://github.com/aws/karpenter-provider-aws/pull/4735 is the direction our use-case will ultimately want to head in anyway  but I have seen issue this brought up in other forums; understood if the hacky-ish implementation isn't justified here

jonathan-innis:
@jmdeal and I were discussing this PR and actually came up with an entirely different concern with this logic (which is actually even more concerning to me) which is that due to the initialization flow here, is it possible for PVs to trickle onto the cluster from the local-static-provisioner. If that's the case, it's also possible that we overshoot here as well since we'll then launch another node since it wouldn't fit into the current limits, then that one trickles, and we continue. A lot of it depends on how quickly we expect the local static provisioner to respond.

@alec-rabold Do you have a feeling how quickly the local static provisioner registers the PVs and attaches them to the storage class?

PR #547, opened 2023-09-28
fix: try other provisioners if one fails

tzneal:
Part of fixing https://github.com/aws/karpenter/issues/4698

**Description**

This allows cloud providers to return errors for GetInstanceTypes and not block all provisioning.

New error message:

```
karpenter-d7887b4d7-wx64q controller 2023-09-28T13:04:14.064Z	ERROR	controller.provisioner	skipping, unable to resolve instance types, resolving node class, resolving node template, getting providerRef, AWSNodeTemplate.karpenter.k8s.aws "defaultFake" not found	{"commit": "ff540d0", "provisioner": "default"}
karpenter-d7887b4d7-wx64q controller 2023-09-28T13:04:14.066Z	ERROR	controller.provisioner	Could not schedule pod, all available instance types exceed limits for provisioner: "default"	{"commit": "ff540d0", "pod": "default/inflate-54b8bcf56d-fq8kk"}
```

Deployed to EKS

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -230,7 +230,10 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
 		// Get instance type options
 		instanceTypeOptions, err := p.cloudProvider.GetInstanceTypes(ctx, nodePool)
 		if err != nil {
-			return nil, fmt.Errorf("getting instance types, %w", err)
+			// we just log an error and skip the provisioner to prevent a single mis-configured provisioner from stopping
+			// all scheduling
+			logging.FromContext(ctx).With(lo.Ternary(nodePool.IsProvisioner, "provisioner", "nodepool"), nodePool.Name).Errorf("skipping, unable to resolve instance types, %s", err)
+			continue
 		}
 		if len(instanceTypeOptions) == 0 {
 			logging.FromContext(ctx).With(lo.Ternary(nodePool.IsProvisioner, "provisioner", "nodepool"), nodePool.Name).Info("skipping, no resolved instance types found")


pkg/controllers/provisioning/provisioner.go
```diff
@@ -230,7 +230,10 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
 		// Get instance type options
 		instanceTypeOptions, err := p.cloudProvider.GetInstanceTypes(ctx, nodePool)
 		if err != nil {
-			return nil, fmt.Errorf("getting instance types, %w", err)
```
ellistarn, line 233: What about transient network issues? Don't we want to bail and retry on those?

ellistarn: It might be worth modeling an error type, I'm quite nervous of swallowing transient errors, as it could cause us to unintentionally fall back and confuse customers. 

I saw this is in response to `all available instance types exceed limits for provisioner`. Isn't that in our neutral code? We could model that error exactly once without plumbing through the cloudprovider.

tzneal: That problem was caused by `GetInstanceTypes` returning no error when the AWS node template didn't exist.  Making that return an error in that case then exposes an issue here where an error for a single provisioner blocks all provisioning.  

We've ran into this issue a few times before, its safer to log and try another provisioner than allow a single provisioner to block all node provisioning.

jonathan-innis: Could we model this as a `NodePoolNotReady` error? Right now, we have a `NodeClaimNotReady` error that we are using to communicate that the `NodeClaim` doesn't have all of its details resolved yet so that the cloud-neutral code can understand how to handle that case differently than other transient errors. I could see a similar logic here with handling cases where we want to communicate that we never recover. 

It's essentially modeling the status condition that we have been talking about for NodePools and NodeClaims (e.g. model a readiness condition to determine whether we can use the provisioner at this time, but is a little easier to implement than using the status condition outright right now.

tzneal: It's just higher risk to allow a cloud provider error on a single provisioner to stop all provisioning. I don't think a transient error leading to a non-preferred node launch is enough of a concern to risk stopping provisioning. 

We just got https://github.com/aws/karpenter/issues/4723 where all deprovisioning is stopped when a node template is removed.  This fix doesn't solve that problem, but its the same root cause.  Cloud provider errors will happen for us and other cloud provider implementers for a single provisioner, that shouldn't lead to a catastrophic failure in Karpenter where it just stops working.

ellistarn: Let's get this fix in. 

I think at some point, we might want to maintain an in memory representation of InstanceType per NodeClass, such that we can heal any transient network errors via async reconciliation, and just check the types on the launch path. As @jonathan-innis mentions, it would be great to model this via NodePool readiness, rolled up from NodeClass readiness.

jonathan-innis (approved):

PR #548, opened 2023-09-28
fix: don't allow a single bad provisioner to stop deprovisioning

tzneal:
Fixes https://github.com/aws/karpenter/issues/4723

**Description**

If you misconfigure a provisioner (e.g. delete its node template), we
- should log and let you know (fixed earlier)
- still be able to deprovision any nodes that were not created by that provisioner


jonathan-innis (approved):
LGTM 🚀

Pixis-Akshay-Gopani:
@tzneal  is it possilbe for karpenter to handle the deprovisioning for bad provisioner ??  is there any possibility to prevent the deletion of nodetemplate if its getting used by provisioner ??

tzneal:
Currently no, can you submit a feature request?  There is no current way to stop deletion of the node template, we could potentially put a finalizer on it to prevent deletion while its in use.

PR #549, opened 2023-09-29
fix: Topology domains should be constrained on labels

jonathan-innis:
Fixes #N/A 

**Description**

Topology Domains [currently observe the constraints of the Provisioner they are under](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/provisioning/provisioner.go#L249) so that the scheduler uses as much spread as is possible within the full set of topology options that are available on the cluster and within the set of options that Provisioners offer. 

Prior to this change, we were not respecting labels for constraining the domain set. This meant that if there was a label on a NodePool that also had a topologyConstraint on a pod and you didn't have Provisioners that covered the full set of domains that were returned by the instance types, scheduling would hang. 

This change ensures that the domain values are constrained by NodePool labels as well as NodePool requirements.

`make presubmit`


tzneal (approved):
lgtm, nice catch!

PR #553, opened 2023-09-29
chore: Enable configuring logger through volume mount and environment variable

jonathan-innis:
Fixes #N/A 

**Description**

This PR deprecates the `karpenter/config-logging` ConfigMap for configuring the logger, in favor of an opinionated approach to how the logger is configured. Now, Karpenter will configure the logger by default and will surface the `LOG_LEVEL` environment variable and `--log-level` CLI argument to allow changing the log level for the controller. This is the recommended way to configure Karpenter logging if log level changes are needed.

Alternatively, Karpenter will support a soon-to-be-deprecated form of configuring the logger through the filesystem. This method is introduced as an alternative to Karpenter discovering the ConfigMap through API discovery and, instead, uses the container-native filesystem mechanism of loading in the configuration.

### Filesystem Paths
1. Global Zap Logger Config: `/etc/karpenter/logging/zap-logger-config`
2. Controller Component Log Level Override: `/etc/karpenter/logging/loglevel.controller`
3. Webhook Component Log Level Override: `/etc/karpenter/logging/loglevel/webhook`

### Configuring the Volume Mount with the ConfigMap

```yaml
apiVersion: apps/v1
kind: Deployment
spec:
  template:
    spec:
    ...
      containers:
      - name: controller
        volumeMounts:
        - name: config-logging
          mountPath: /etc/karpenter/logging
      volumes:
      - name: config-logging
        configMap:
          name: config-logging
```

`make presubmit`

### DIFF
--- a/pkg/operator/logging/logging.go
+++ b/pkg/operator/logging/logging.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"log"
+	"os"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -44,6 +45,8 @@ import (
 const (
 	loggerCfgConfigMapName = "config-logging"
 	loggerCfgConfigMapKey  = "zap-logger-config"
+	loggerCfgDir           = "/etc/karpenter/logging"
+	loggerCfgFilePath      = loggerCfgDir + "/zap-logger-config"
 )
 
 func DefaultZapConfig() zap.Config {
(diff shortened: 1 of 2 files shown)


pkg/operator/logging/logging.go
```diff
@@ -94,8 +105,44 @@ func WithCommit(logger *zap.SugaredLogger) *zap.SugaredLogger {
...
+			return nil
+		}
+		log.Fatalf("retrieving logging configuration file from %q", loggerCfgFilePath)
+	}
+	cfg := DefaultZapConfig()
+	lo.Must0(json.Unmarshal(raw, &cfg))
+
+	raw, err = os.ReadFile(loggerCfgDir + fmt.Sprintf("/loglevel.%s", component))
+	if err != nil && !os.IsNotExist(err) {
+		log.Fatalf("retrieving logging controller log level file from %q", loggerCfgDir+fmt.Sprintf("/loglevel.%s", component))
```
njtran, line 133: If the log level doesn't exist, should we just return and fall back to the default?

jonathan-innis: Yes, I think so. A user is allowed not to have this file and (if they don't have it) then we should just use the default. I _do_ still want to catch errors that aren't related to notFound like transient errors with the filesystem.

njtran (approved):

PR #560, opened 2023-10-02
Status register empty nodepool

shikai93:
reconcile nodepool controller via updating instead of patching to allow nil map status to be reflected correctly

Fixes [#4628 ](https://github.com/aws/karpenter/issues/4628)

**Description**
Since nodepool controller reconciles with patch, when the last node is deleted, after [resourceCountsFor](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/nodepool/counter/controller.go#L78) returns nil map, the patch would merge this with the existing map and the status cannot be updated to have no CPU and Mem associated with a provisioner. 

Tested using the included unit test


jonathan-innis:
Hi @shikai93, looked into this issue. This issue seems to be unrelated to the Patch/Update operation and instead related to a miss in the conversion logic (see #561)

PR #562, opened 2023-10-02
feat: v1beta1

jonathan-innis:
Fixes #N/A 

**Description**


pkg/apis/v1beta1/nodeclaim_validation.go
```diff
@@ -124,9 +124,6 @@ func (in *NodeClaimSpec) validateTaintsField(taints []v1.Taint, existing map[tai
 // NodeClaim requirements only support well known labels.
 func (in *NodeClaimSpec) validateRequirements() (errs *apis.FieldError) {
 	for i, requirement := range in.Requirements {
-		if requirement.Key == NodePoolLabelKey {
-			errs = errs.Also(apis.ErrInvalidArrayValue(fmt.Sprintf("%s is restricted", requirement.Key), "requirements", i))
-		}
```
njtran, line 129: Should we be removing this? Are users now allowed to say that they want their pods to use a certain nodepool?

pkg/controllers/controllers.go
```diff
@@ -58,21 +58,28 @@ func NewControllers(
 		disruption.NewController(clock, kubeClient, p, cloudProvider, recorder, cluster),
 		provisioning.NewController(kubeClient, p, recorder),
 		nodepoolhash.NewProvisionerController(kubeClient),
+		nodepoolhash.NewNodePoolController(kubeClient),
 		informer.NewDaemonSetController(kubeClient, cluster),
 		informer.NewNodeController(kubeClient, cluster),
 		informer.NewPodController(kubeClient, cluster),
 		informer.NewProvisionerController(kubeClient, cluster),
```
njtran, line 65: Need a informer.NodePoolController?

pkg/controllers/controllers.go
```diff
@@ -58,21 +58,28 @@ func NewControllers(
...
 		informer.NewMachineController(kubeClient, cluster),
+		informer.NewNodeClaimController(kubeClient, cluster),
 		termination.NewController(kubeClient, cloudProvider, terminator.NewTerminator(clock, kubeClient, evictionQueue), recorder),
 		metricspod.NewController(kubeClient),
 		metricsprovisioner.NewController(kubeClient),
```
njtran, line 70: need a metricsnodepool.NewController()?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -198,7 +198,7 @@ func (p *Provisioner) consolidationWarnings(ctx context.Context, po v1.Pod) {
 	}
 }
 
-var ErrProvisionersNotFound = errors.New("no provisioners found")
+var ErrNodePoolsNotFound = errors.New("no nodepools found")
```
njtran, line 201: Doesn't this need to be a ternary?

pkg/apis/v1beta1/nodeclaim_validation.go
jonathan-innis: Responded offline. We can't validate this on NodeClaims since NodeClaims will _always_ be assigned to a NodePool when they are being provisioned. It also makes sense to have a NodePool requirement on NodeClaims whereas it makes no sense to have this on NodePools because the worst you could do there is shoot yourself in the foot.

pkg/controllers/controllers.go
jonathan-innis: Good catch

pkg/controllers/provisioning/provisioner.go
jonathan-innis: Resolved by updating the error message

njtran (dismissed):

njtran (approved):

PR #571, opened 2023-10-04
test: Move `env` to be created at the beginning of suite for `v1beta1`

engedaam:
Fixes #N/A 

**Description**
- To improve testing speed, the environment set-up should be create an the beginning of the suite for the v1bata1 test suite.

- N/A

### DIFF
--- a/pkg/apis/v1beta1/nodepool_validation_cel_test.go
+++ b/pkg/apis/v1beta1/nodepool_validation_cel_test.go
@@ -25,18 +25,13 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	"github.com/aws/karpenter-core/pkg/apis"
 	. "github.com/aws/karpenter-core/pkg/apis/v1beta1"
-	"github.com/aws/karpenter-core/pkg/operator/scheme"
-	"github.com/aws/karpenter-core/pkg/test"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/apis/v1beta1/suite_test.go
```diff
@@ -21,12 +21,21 @@ import (
...
 var ctx context.Context
+var env *test.Environment
 
 func TestAPIs(t *testing.T) {
 	ctx = TestContextWithLogger(t)
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "v1beta1")
 }
+
+var _ = BeforeSuite(func() {
```
jonathan-innis, line 39: I believe there is an `env.Stop()` call that we need to make sure that we call at the end. We should probably also call `ExpectCleanedUp` in the `AfterEach` just by convention

pkg/apis/v1beta1/nodepool_validation_cel_test.go
```diff
@@ -57,6 +53,9 @@ var _ = Describe("CEL/Validation", func() {
 			},
 		}
 	})
+	BeforeEach(func() {
```
jonathan-innis, line 56:
```suggestion
	AfterEach(func() {
```

This is `AfterEach` in all the other `suite_test.go`s and this top-level cleanup should generally sit in the `suite_test.go` file

jonathan-innis (approved):
Nice 🎉

PR #572, opened 2023-10-04
chore: move termination queue to singleton controller

njtran:
Fixes #N/A 

**Description**
This makes the termination controller into a singleton controller. This removes the goroutine from the constructor and allow the underlying controller manager to start/cancel the context so that this goroutine cleans up with Karpenter is shutting down.

- make presubmit
- applied to local cluster

### DIFF
--- a/pkg/controllers/termination/terminator/eviction.go
+++ b/pkg/controllers/termination/terminator/eviction.go
@@ -96,24 +96,37 @@ func (q *Queue) Add(pods ...*v1.Pod) {
 	}
 }
 
-func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (result reconcile.Result, err error) {
+// Pop returns a namespaced name from the queue and if the queue has shutdown.
+func (q *Queue) Pop() (types.NamespacedName, bool) {
 	// Get pod from queue. This waits until queue is non-empty.
 	item, shutdown := q.RateLimitingInterface.Get()
 	if shutdown {
-		return reconcile.Result{RequeueAfter: controller.Immediately}, nil
+		return types.NamespacedName{}, true
 	}
 	nn := item.(types.NamespacedName)
+	return nn, false
+}
+
+func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	nn, shutdown := q.Pop()
+	if shutdown {
+		return reconcile.Result{}, fmt.Errorf("queue has shutdown")
+	}
+	q.ProcessItem(ctx, nn)
+	return reconcile.Result{RequeueAfter: controller.Immediately}, nil
+}
+
+func (q *Queue) ProcessItem(ctx context.Context, nn types.NamespacedName) {
 	// Evict pod
 	if q.Evict(ctx, nn) {
 		q.RateLimitingInterface.Forget(nn)
 		q.Set.Remove(nn)
 		q.RateLimitingInterface.Done(nn)
-		return reconcile.Result{RequeueAfter: controller.Immediately}, nil
+		return
 	}
 	q.RateLimitingInterface.Done(nn)
 	// Requeue pod if eviction failed
 	q.RateLimitingInterface.AddRateLimited(nn)
-	return reconcile.Result{RequeueAfter: controller.Immediately}, nil
 }
 
 // Evict returns true if successful eviction call, and false if not an eviction-related error
--- a/pkg/controllers/termination/terminator/eviction.go
+++ b/pkg/controllers/termination/terminator/eviction.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/aws/karpenter-core/pkg/operator/controller"
 	set "github.com/deckarep/golang-set"
 	v1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
@@ -34,6 +33,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/aws/karpenter-core/pkg/operator/controller"
+
 	terminatorevents "github.com/aws/karpenter-core/pkg/controllers/termination/terminator/events"
 	"github.com/aws/karpenter-core/pkg/events"
 )
@@ -41,7 +42,6 @@ import (
 const (
 	evictionQueueBaseDelay = 100 * time.Millisecond
 	evictionQueueMaxDelay  = 10 * time.Second
-	immediately            = time.Millisecond
 )
 
 type NodeDrainError struct {
@@ -60,7 +60,7 @@ func IsNodeDrainError(err error) bool {
 	return errors.As(err, &nodeDrainErr)
 }
 
-type EvictionQueue struct {
+type Queue struct {
 	workqueue.RateLimitingInterface
 	set.Set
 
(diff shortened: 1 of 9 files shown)


pkg/controllers/termination/terminator/eviction.go
```diff
@@ -85,26 +96,26 @@ func (e *EvictionQueue) Add(pods ...*v1.Pod) {
...
-		nn := item.(types.NamespacedName)
-		// Evict pod
-		if e.evict(ctx, nn) {
-			e.RateLimitingInterface.Forget(nn)
-			e.Set.Remove(nn)
-			e.RateLimitingInterface.Done(nn)
-			continue
-		}
+func (e *EvictionQueue) Reconcile(ctx context.Context, _ reconcile.Request) (result reconcile.Result, err error) {
+	// Set the result to requeue immediately.
```
jonathan-innis, line 100: Don't have a huge problem with it but why reset it to a variable?

jonathan-innis (commented):
Nice! 🎉

pkg/controllers/termination/terminator/eviction.go
njtran: I had it just to make sure we're doing immediately and doing returned values, but I'll just copy it to the three lines

pkg/operator/controller/singleton.go
```diff
@@ -30,6 +30,10 @@ import (
 	"github.com/aws/karpenter-core/pkg/metrics"
 )
 
+const (
+	Immediately = time.Millisecond
```
jonathan-innis, line 34: This should be the number `1` or `time.NanoSecond`. Duration is in nanoseconds

pkg/controllers/termination/terminator/terminator.go
```diff
@@ -86,7 +86,7 @@ func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
 		podsToEvict = append(podsToEvict, p)
 	}
 	// Enqueue for eviction
-	t.evict(podsToEvict)
+	t.Evict(podsToEvict)
```
jonathan-innis, line 89: Why make it public?

pkg/controllers/termination/terminator/suite_test.go
```diff
@@ -0,0 +1,123 @@
+/*
```
jonathan-innis, line 1: Nice 🎉

pkg/controllers/termination/terminator/eviction.go
```diff
@@ -56,61 +60,79 @@ func IsNodeDrainError(err error) bool {
...
-			e.RateLimitingInterface.Forget(nn)
-			e.Set.Remove(nn)
-			e.RateLimitingInterface.Done(nn)
-			continue
-		}
-		e.RateLimitingInterface.Done(nn)
-		// Requeue pod if eviction failed
-		e.RateLimitingInterface.AddRateLimited(nn)
+// Pop returns a namespaced name from the queue and if the queue has shutdown.
+func (q *Queue) Pop() (types.NamespacedName, bool) {
```
jonathan-innis, line 100: This is just doing the interface conversion. I could go back and forth on whether you actually need a separate method for this. Thoughts?

pkg/controllers/termination/terminator/eviction.go
```diff
@@ -56,61 +60,79 @@ func IsNodeDrainError(err error) bool {
...
+
+func (q *Queue) ProcessItem(ctx context.Context, nn types.NamespacedName) {
+	// Evict pod
+	if q.Evict(ctx, nn) {
+		q.RateLimitingInterface.Forget(nn)
+		q.Set.Remove(nn)
+		q.RateLimitingInterface.Done(nn)
+		return
+	}
+	q.RateLimitingInterface.Done(nn)
```
jonathan-innis, line 127: I'm forgetting how the rate limiting interface handles this, but you call Done() to finish pulling an element out of the queue and then you add it back to the back of the queue? Is that correct? Does it make sense to call Done() after the initial Get() if it's a Pop() operation?

pkg/controllers/termination/terminator/terminator.go
njtran: it's called in the queue suite tests

pkg/controllers/termination/terminator/eviction.go
njtran: If we're not doing the interface conversion here, we'll have to do it in every mocked test (or just the one expectation function. I like it like this more, since any changes to the pop logic will be propagated to the tests and the real package without having to change it elsewhere.

pkg/controllers/termination/terminator/eviction.go
njtran: It's used as a way to mark the item as done processing. When you do Get() for the item, it pulls it out of the queue and removes it from the dirty (needing processing) set. Using done will make sure if the item gets re-enqueued while we're processing that it doesn't get requeued earlier than it should, putting it at the back of the queue. 


// Done marks item as done processing, and if it has been marked as dirty again
// while it was being processed, it will be re-added to the queue for
// re-processing.

jonathan-innis: Ah, got it. Then looks correct. I see that controller-runtime uses a defer statement for the `Done()` call. Is that something that we should consider doing?

pkg/controllers/termination/terminator/terminator.go
jonathan-innis: Got it 👍

pkg/controllers/termination/terminator/eviction.go
jonathan-innis: SGTM

jonathan-innis (approved):
One open comment. You can take it or leave it 👍

pkg/controllers/termination/terminator/eviction.go
njtran: Yeah, defer Done() sounds right will do.

pkg/controllers/termination/terminator/eviction.go
```diff
@@ -56,61 +60,78 @@ func IsNodeDrainError(err error) bool {
...
+func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+	nn, shutdown := q.Pop()
+	if shutdown {
+		return reconcile.Result{}, fmt.Errorf("queue has shutdown")
+	}
+	q.ProcessItem(ctx, nn)
+	return reconcile.Result{RequeueAfter: controller.Immediately}, nil
+}
+
+func (q *Queue) ProcessItem(ctx context.Context, nn types.NamespacedName) {
```
jonathan-innis, line 119: Does it make sense to just collapse this into the reconcile since it's so short anyways. And then you have all the `RateLimitingInterface` interactions happening in the same function

pkg/operator/controller/singleton.go
```diff
@@ -30,6 +30,10 @@ import (
 	"github.com/aws/karpenter-core/pkg/metrics"
 )
 
+const (
+	Immediately = 1 * time.Millisecond
```
jonathan-innis, line 34:
```suggestion
	Immediately = 1 * time.Nanosecond
```

pkg/controllers/termination/terminator/eviction.go
njtran: I was doing this for mocking purposes, but realizing I actually don't need this anymore. Thanks.

pkg/controllers/termination/terminator/eviction.go
```diff
@@ -56,61 +60,65 @@ func IsNodeDrainError(err error) bool {
...
+	if shutdown {
+		return reconcile.Result{}, fmt.Errorf("EvictionQueue is broken and has shutdown")
+	}
+	nn := item.(types.NamespacedName)
+	defer q.RateLimitingInterface.Done(nn)
+	// Evict pod
+	if q.Evict(ctx, nn) {
+		q.RateLimitingInterface.Forget(nn)
+		q.Set.Remove(nn)
+		return reconcile.Result{RequeueAfter: controller.Immediately}, nil
```
jonathan-innis, line 111: optional: The if/else just returns the same result so you _could_ convert this from a short-circuit to an if/else statement

jonathan-innis (approved):
LGTM with one question/nit 🚀

PR #576, opened 2023-10-06
test: fix termination test flake

njtran:
Fixes #N/A 

**Description**
Fixes a flake in termination tests. The ordering of pods aren't consistent across critical pods. This makes it so that we check if both critical pods are deleted after the queue processes both of them.

- make presubmit


jonathan-innis (approved):
LGTM 🚀

PR #578, opened 2023-10-07
test: replaced offsets with GinkgoHelper()

### LINKED ISSUE #395: Convert all `Expect...` test expectations to use `GingkoHelper`
### Description

Ginkgo offers the function called [`GinkgoHelper`](https://onsi.github.io/ginkgo/#:~:text=Instead%20of%20managing%20offsets%20you%20can%20use%20GinkgoHelper()) which means that you don't have to use the provided `...WithOffset` functions and can instead use this function to get the correct stack trace. We should convert all of our test expectation functions that use this function to use `GingkoHelper()`.

### Relevant Codepaths

1. https://github.com/aws/karpenter/blob/main/test/pkg/environment/aws/expectations.go
2. https://github.com/aws/karpenter/blob/main/test/pkg/environment/common/expectations.go

sadath-12:
Fixes #395  

**Description**

### DIFF
--- a/pkg/controllers/deprovisioning/suite_test.go
+++ b/pkg/controllers/deprovisioning/suite_test.go
@@ -1128,6 +1128,7 @@ func ExpectNewMachinesDeleted(ctx context.Context, c client.Client, wg *sync.Wai
 
 	wg.Add(1)
 	go func() {
+		GinkgoHelper()
 		machinesDeleted := 0
 		ctx, cancel := context.WithTimeout(ctx, time.Second*30) // give up after 30s
 		defer GinkgoRecover()
@@ -1145,7 +1146,7 @@ func ExpectNewMachinesDeleted(ctx context.Context, c client.Client, wg *sync.Wai
 					if existingMachineNames.Has(m.Name) {
 						continue
 					}
-					ExpectWithOffset(1, client.IgnoreNotFound(c.Delete(ctx, m))).To(Succeed())
+					Expect(client.IgnoreNotFound(c.Delete(ctx, m))).To(Succeed())
 					machinesDeleted++
 					if machinesDeleted == numNewMachines {
 						return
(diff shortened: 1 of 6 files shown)


pkg/controllers/deprovisioning/suite_test.go
```diff
@@ -1128,6 +1128,7 @@ func ExpectNewMachinesDeleted(ctx context.Context, c client.Client, wg *sync.Wai
 
 	wg.Add(1)
 	go func() {
+		GinkgoHelper()
```
jonathan-innis, line 1131: Not actually sure here but do you need to do two GinkgoHelpers()? One at the top of this function `ExpectNewMachinesDeleted` and the other the goroutine? I think the stack trace won't print correctly if this failed (to be clear I don't think this works correctly today either from looking at it, but this is a good time to fix it 😄 )

pkg/test/randomdelay.go
```diff
@@ -58,29 +58,32 @@ func (g *gomegaWrapper) Expect(actual interface{}, extra ...interface{}) types.A
 	return g.inner.Expect(actual, extra...)
 }
 
-func (g *gomegaWrapper) ExpectWithOffset(offset int, actual interface{}, extra ...interface{}) types.Assertion {
+func (g *gomegaWrapper) Expect(actual interface{}, extra ...interface{}) types.Assertion {
```
jonathan-innis, line 61: I think you should leave this interface alone since I believe that this wrapper interface needs to implement _every_ function that the top-level Gomega interface defines.

jonathan-innis (commented):
Looks good! Thanks for knocking this out! Just a few comments!

pkg/controllers/deprovisioning/suite_test.go
sadath-12: ok i just thought since the goroutines are decoupled from the original function they might need separate attention since they work separately

sadath-12: ya similar thought here

pkg/controllers/deprovisioning/suite_test.go
```diff
@@ -1214,6 +1216,7 @@ func ExpectNewNodeClaimsDeleted(ctx context.Context, c client.Client, wg *sync.W
 
 	wg.Add(1)
```
jonathan-innis, line 1217: I think you need a new `GinkgoHelper` at the top of this function as well

jonathan-innis (commented):
One blocking comment

sadath-12:
Changes applied @jonathan-innis

jonathan-innis (approved):
LGTM 🚀

PR #583, opened 2023-10-09
fix: Fix using the wrong pods when validating `do-not-evict` pods

jonathan-innis:
Fixes https://github.com/aws/karpenter/issues/4310, https://github.com/aws/karpenter/issues/4758

**Description**

This PR correctly validates that the candidates that we are attempting to deprovision are still eligible to be deprovisioned. It does this by mapping the old candidates onto the new candidates that we pull from state and checking that all of those candidates are still valid for deprovisioning based on PDB criteria, `do-not-evict` annotations, and scheduling criteria.

This PR also removes the caching mechanism for `validationCandidates` since we weren't calling `IsValid` multiple times in code, meaning that storing the candidates like this wasn't necessary for our usage.

https://github.com/aws/karpenter-core/pull/247 had originally intended to fix this issue, but uses the wrong pods for validation within the `filterCandidates` function. This causes validation to pass for candidate nodes that receive pods with blocking PDBs or `do-not-evict` pods within the validation window.

`make presubmit`

### DIFF
--- a/pkg/controllers/deprovisioning/nodeclaim_consolidation_test.go
+++ b/pkg/controllers/deprovisioning/nodeclaim_consolidation_test.go
@@ -2291,7 +2291,7 @@ var _ = Describe("NodeClaim/Consolidation", func() {
 			// Iterate in a loop until we get to the validation action
 			// Then, apply the pods to the cluster and bind them to the nodes
 			for {
-				time.Sleep(250 * time.Millisecond)
+				time.Sleep(100 * time.Millisecond)
 				if fakeClock.HasWaiters() {
 					break
 				}
@@ -2307,7 +2307,7 @@ var _ = Describe("NodeClaim/Consolidation", func() {
 			ExpectManualBinding(ctx, env.Client, doNotEvictPod, node)
 
 			// Step forward to satisfy the validation timeout and wait for the reconcile to finish
-			fakeClock.Step(45 * time.Second)
+			ExpectTriggerVerifyAction(&wg)
 			wg.Wait()
 
 			// we would normally be able to replace a node, but we are blocked by the do-not-evict pods during validation
@@ -2339,7 +2339,7 @@ var _ = Describe("NodeClaim/Consolidation", func() {
 			// Iterate in a loop until we get to the validation action
 			// Then, apply the pods to the cluster and bind them to the nodes
 			for {
-				time.Sleep(250 * time.Millisecond)
+				time.Sleep(100 * time.Millisecond)
 				if fakeClock.HasWaiters() {
 					break
 				}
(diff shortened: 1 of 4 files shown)


pkg/controllers/deprovisioning/nodeclaim_consolidation_test.go
```diff
@@ -2264,6 +2267,327 @@ var _ = Describe("NodeClaim/Consolidation", func() {
...
+			}()
+
+			// Iterate in a loop until we get to the validation action
+			// Then, apply the pods to the cluster and bind them to the nodes
+			for {
+				time.Sleep(250 * time.Millisecond)
+				if fakeClock.HasWaiters() {
+					break
+				}
+			}
```
njtran, line 2506: nit: In my experience, tests are pretty quick, thoughts on lowering this to 100ms? I think it's fine to just check a little quicker and save some time on average. 

also: Maybe nice to pull this out to a different function too since it looks like you're using this in the other tests.

pkg/controllers/deprovisioning/nodeclaim_consolidation_test.go
```diff
@@ -2264,6 +2267,327 @@ var _ = Describe("NodeClaim/Consolidation", func() {
...
+						v1beta1.DoNotDisruptAnnotationKey: "true",
+					},
+				},
+			})
+			ExpectApplied(ctx, env.Client, doNotEvictPods[0], doNotEvictPods[1])
+			ExpectManualBinding(ctx, env.Client, doNotEvictPods[0], node)
+			ExpectManualBinding(ctx, env.Client, doNotEvictPods[1], node2)
+
+			// Step forward to satisfy the validation timeout and wait for the reconcile to finish
+			fakeClock.Step(45 * time.Second)
```
njtran, line 2519: Can we use `ExpectTriggerVerifyAction()` for each time you make this call? Will make sure that we're stepping forward as many times as necessary so that it works and we don't have to adjust this when we make this validation period configurable.

pkg/controllers/deprovisioning/nodeclaim_consolidation_test.go
```diff
@@ -2264,6 +2267,327 @@ var _ = Describe("NodeClaim/Consolidation", func() {
...
+			pdb := test.PodDisruptionBudget(test.PDBOptions{
+				Labels:         labels,
+				MaxUnavailable: fromInt(0),
+				Status: &policyv1.PodDisruptionBudgetStatus{
+					ObservedGeneration: 1,
+					DisruptionsAllowed: 0,
+					CurrentHealthy:     1,
+					DesiredHealthy:     1,
+					ExpectedPods:       1,
+				},
```
njtran, line 2575: Do we need the PDB? In my experience, you just need to create one with the proper maxUnavailable, where this status will get updated by itself anyway.

njtran (commented):
good catch! Just some nits on the tests.

njtran (dismissed):
approved with comments nits

njtran (approved):

PR #589, opened 2023-10-11
release: Patch `v0.31.x` release for `v0.31.1`

jonathan-innis:
Fixes #N/A 

**Description**

This PR takes in cherry-picked changes for the `v0.31.1` release

`make presubmit`

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -505,6 +505,11 @@ func (c *Cluster) populateInflight(ctx context.Context, n *StateNode) error {
 	if n.inflightInitialized {
 		return nil
 	}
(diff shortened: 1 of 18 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -505,6 +505,11 @@ func (c *Cluster) populateInflight(ctx context.Context, n *StateNode) error {
 	if n.inflightInitialized {
 		return nil
 	}
+	// If the node ies already initialized, we don't need to populate its inflight capacity
```
njtran, line 508: nit on commit but you don't need to fix it. 
```suggestion
	// If the node is already initialized, we don't need to populate its inflight capacity
```

njtran (approved):

PR #593, opened 2023-10-12
chore: make options injectable

jmdeal:
Fixes #N/A 

**Description**
Introduces the Injectable pattern currently used for `Settings` for `Options`. This enables injecting `Options` from a cloud provider into the manager.

`make presubmit` and `/karpenter snapshot` (via [aws/karpenter@pr4789](https://github.com/aws/karpenter/pull/4789))


pkg/apis/apis.go
```diff
@@ -36,6 +37,7 @@ var (
 	// AddToScheme may be used to add all resources defined in the project to a Scheme
 	AddToScheme = Builder.AddToScheme
 	Settings    = []settings.Injectable{&settings.Settings{}}
+	Options     = []options.Injectable{&options.Options{}}
```
jonathan-innis, line 40: Does it make sense for this to be in `apis` or can we store the options in options like we do for the `scheme`? I'm thinking `aws/karpenter` will just do `coreoptions.Options = append(coreoptions.Options, options.Options)`

pkg/operator/operator.go
```diff
@@ -80,7 +78,8 @@ func NewOperator() (context.Context, *Operator) {
 	ctx = knativeinjection.WithNamespaceScope(ctx, system.Namespace())
 
 	// Options
-	opts := lo.Must(options.New().Parse(os.Args[1:]...))
+	ctx = injection.WithOptionsOrDie(ctx, apis.Options...)
+	opts := options.FromContext(ctx)
```
jonathan-innis, line 82: Do we need to store the opts like this if they are already injected into the context? We can just pull them out of the context starting from here, right?

pkg/operator/options/options.go
```diff
@@ -134,25 +135,32 @@ func (o *Options) Parse(args ...string) (*Options, error) {
 		o.FeatureGatesSet = true
 	}
 
-	return o, nil
+	ctx = ToContext(ctx, o)
```
jonathan-innis, line 138: Collapse into a single line?

pkg/operator/options/options.go
```diff
@@ -134,25 +135,32 @@ func (o *Options) Parse(args ...string) (*Options, error) {
...
 	// the default value is used from Settings.
-	mergeField(&o.BatchMaxDuration, s.BatchMaxDuration, o.BatchMaxDurationSet)
-	mergeField(&o.BatchIdleDuration, s.BatchIdleDuration, o.BatchIdleDurationSet)
-	mergeField(&o.FeatureGates.Drift, s.DriftEnabled, o.FeatureGatesSet)
-	return o
+	return ctx
 }
 
-func MustParseFeatureGates(gateStr string) FeatureGates {
+func FeatureGatesFromStr(gateStr string) FeatureGates {
```
jonathan-innis, line 163: Why change this name? I think the name before was fine since you have a must statement in here.

pkg/operator/options/options.go
```diff
@@ -134,25 +135,32 @@ func (o *Options) Parse(args ...string) (*Options, error) {
...
+		_, ok := in.(*settings.Settings)
+		if !ok {
+			continue
+		}
+		s := in.FromContext(ctx).(*settings.Settings)
+		o := FromContext(ctx)
+		mergeField(&o.BatchMaxDuration, s.BatchMaxDuration, o.BatchMaxDurationSet)
+		mergeField(&o.BatchIdleDuration, s.BatchIdleDuration, o.BatchIdleDurationSet)
+		mergeField(&o.FeatureGates.Drift, s.DriftEnabled, o.FeatureGatesSet)
+		ctx = ToContext(ctx, o)
```
jonathan-innis, line 155: You should break after you have done this in order to short-circuit

jonathan-innis (commented):
Looks great! Just a few things

pkg/operator/operator.go
jmdeal: We can, its used a lot though so there's some value in taking it out here for brevity elsewhere. Definitely a style preference though, I can go either way.

pkg/operator/options/options.go
jmdeal: I realized I really don't need this loop at all, rather than going over all settings injectables, it can just use the concrete settings directly...not sure why I got it stuck in my head that the abstraction was needed here.

pkg/operator/options/options.go
jmdeal: Honestly not sure, I think I was making some other changes that go scrapped but this name change slipped through. I'm happy to change it back.

pkg/apis/apis.go
jmdeal: Probably not, thats a good callout. I had put it there because I was copying the injectable pattern for settings but that's not an appropriate location. The only problem is the name conflict with the `Options` struct but I think it would be reasonable to rename it to `options.Injectables`.

pkg/operator/operator.go
jonathan-innis: I don't care either way either. Just a question. I'm fine either way.

jonathan-innis (approved):
Nice 🎉

pkg/operator/options/options.go
```diff
@@ -59,111 +61,89 @@ type OptionFields struct {
...
-	f.StringVar(&opts.LogLevel, "log-level", env.WithDefaultString("LOG_LEVEL", ""), "Log verbosity level. Can be one of 'debug', 'info', or 'error'")
-
-	// Vars that must be merged with settings
-	f.DurationVar(&opts.BatchMaxDuration, "batch-max-duration", env.WithDefaultDuration("BATCH_MAX_DURATION", 10*time.Second), "The maximum length of a batch window. The longer this is, the more pods we can consider for provisioning at one time which usually results in fewer but larger nodes.")
-	f.DurationVar(&opts.BatchIdleDuration, "batch-idle-duration", env.WithDefaultDuration("BATCH_IDLE_DURATION", time.Second), "The maximum amount of time with no new pending pods that if exceeded ends the current batching window. If pods arrive faster than this time, the batching window will be extended up to the maxDuration. If they arrive slower, the pods will be batched separately.")
-	f.StringVar(&opts.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "Drift=false"), "Optional features can be enabled / disabled using feature gates. Current options are: Drift")
-
-	return opts
+	fs := FlagSetFromContext(ctx)
```
jonathan-innis, line 67: Why store the flagset in the context. Why not just store a global flagset and have users add to it? e.g. `options.FlagSet`

pkg/operator/options/options.go
```diff
@@ -59,111 +61,89 @@ type OptionFields struct {
...
-func (o *Options) Parse(args ...string) (*Options, error) {
-	if err := o.FlagSet.Parse(args); err != nil {
-		if errors.Is(err, flag.ErrHelp) {
-			os.Exit(0)
-		}
-		return nil, err
-	}
-
+func (*Options) Validate(ctx context.Context) error {
+	o := FromContext(ctx)
```
jonathan-innis, line 87: I'm not sure I understand the use-case for the `Validate` function? Why not just do everything in the `Inject`?

pkg/operator/options/options.go
jonathan-innis: Or, maybe better. Create an interface method called AddFlags(fs *Flagest). That passes in the flagset and just registers the flags

jmdeal: Design decisions resolved offline

pkg/operator/options/options.go
jmdeal: Design decisions resolved offline

pkg/operator/injection/injection.go
```diff
@@ -16,12 +16,14 @@ package injection
...
 	"context"
+	"flag"
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	kubeerrors "k8s.io/apimachinery/pkg/api/errors"
```
jonathan-innis, line 26: Why do we need an alias for this?

pkg/operator/operator.go
```diff
@@ -80,35 +78,39 @@ func NewOperator() (context.Context, *Operator) {
...
 
 	// Inject settings from the ConfigMap(s) into the context
 	ctx = injection.WithSettingsOrDie(ctx, kubernetesInterface, apis.Settings...)
-	opts = opts.MergeSettings(settings.FromContext(ctx))
-	ctx = options.ToContext(ctx, opts)
+
+	// Temporarily merge settings into options until configmap is removed
+	// Note: injectables are pointer to those already in context
+	for _, o := range options.Injectables {
+		o.MergeSettings(ctx)
```
jonathan-innis, line 112: This doesn't look like it updates the context, should the interface show that it updates the context by outputting a new context. I get that this is a pointer but not returning a new context doesn't make it super clear that something is being updated here

pkg/operator/options/injectable.go
```diff
@@ -0,0 +1,41 @@
...
+// 	Validate(context.Context) error
+//
+// 	MergeSettings(context.Context) context.Context
+// }
+
+type Injectable interface {
+	AddFlags(*flag.FlagSet)
+	Parse(*flag.FlagSet, ...string) error
+	ToContext(context.Context) context.Context
+	MergeSettings(context.Context)
```
jonathan-innis, line 40:
```suggestion
	MergeSettings(context.Context) context.Context
```

As mentioned above, I think this interface method should return a new context

pkg/operator/options/injectable.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	"flag"
+)
+
+// Injectable defines a set of flag based options to be parsed and injected
+// into Karpenter's contexts
+// type Injectable interface {
+// 	// Inject is used to inject the Options into the context and set up flags.
```
jonathan-innis, line 25: If we have a set of comments like this, shouldn't these comments be directly over the function rather than at the top like this. General go convention is that you put your function comment directly above the function and the struct comment directly above the struct

pkg/operator/options/options.go
```diff
@@ -29,19 +30,20 @@ import (
 	"github.com/aws/karpenter-core/pkg/utils/env"
 )
 
-var validLogLevels = []string{"", "debug", "info", "error"}
+var (
```
jonathan-innis, line 33: Maybe I'm missing it but where does the global flagset get defined?

pkg/operator/options/options.go
```diff
@@ -57,113 +59,98 @@ type OptionFields struct {
...
-	o.FlagSet.Visit(func(f *flag.Flag) {
-		switch f.Name {
-		case "batch-max-duration":
-			o.BatchMaxDurationSet = true
-		case "batch-idle-duration":
-			o.BatchIdleDurationSet = true
-		case "feature-gates":
-			o.FeatureGatesSet = true
-		}
+	o.setFlags = make(map[string]bool)
```
jonathan-innis, line 102:
```suggestion
	o.setFlags = make(map[string]bool)
```

This is style but we've decided to do `map[string]bool{}` instead of `make` in this project. I don't personally care or think that it matters one way or another, but we should at least try to be consistent throughout

pkg/operator/options/options.go
```diff
@@ -57,113 +59,98 @@ type OptionFields struct {
...
+	fs.VisitAll(func(f *flag.Flag) {
+		envName := strings.ReplaceAll(strings.ToUpper(f.Name), "-", "_")
+		_, ok := os.LookupEnv(envName)
```
jonathan-innis, line 105: This seems reasonable to me. We should probably just make a comment that this makes an assumption about the format between environment variables and flags matching 👍

jonathan-innis: Nice simplification btw 🎉

pkg/operator/options/options.go
```diff
@@ -57,113 +59,98 @@ type OptionFields struct {
...
+	if !o.setFlags["batch-idle-duration"] {
+		o.BatchIdleDuration = s.BatchIdleDuration
+	}
+	if !o.setFlags["feature-gates"] {
+		o.FeatureGates.Drift = s.DriftEnabled
+	}
 }
 
-func MustParseFeatureGates(gateStr string) FeatureGates {
+func FeatureGatesFromStr(gateStr string) (FeatureGates, error) {
```
jonathan-innis, line 140:
```suggestion
func ParseFeatureGates(gateStr string) (FeatureGates, error) {
```

nit: just a naming suggestion

pkg/operator/injection/injection.go
jmdeal: A previous iteration imported `errors`, this can go away now for sure.

pkg/operator/operator.go
jmdeal: Ya I'm a bit mixed on that, that's why I left the comment on line 110. Maybe that can be more clear. I'm hesitant to make it return a context because I think it just changes the location where something isn't clear, since it wouldn't be clear why the function returns a context in the first place. That's why I felt a comment was more appropriate. Or is that pattern just common enough in the go / k8s world that its clear regardless?

pkg/operator/options/options.go
jmdeal: I didn't actually use a global flagset since it's scope could be limited to `injection.WithOptionsOrDie`. Do you still think there's value in a global flagset?

pkg/operator/options/injectable.go
jmdeal: That's actually the old Injectable struct that I commented out and forgot about. I'll move the comments.

pkg/operator/options/options.go
```diff
@@ -57,113 +59,98 @@ type OptionFields struct {
...
-	FeatureGatesSet      bool
-}
-
-type Options struct {
-	OptionFields
-	optionFlags
+	setFlags map[string]bool
 }
 
 // New creates an Options struct and registers CLI flags and environment variables to fill-in the Options struct fields
```
jonathan-innis, line 66:
```suggestion
// AddFlags creates an Options struct and registers CLI flags and environment variables to fill-in the Options struct fields
```

pkg/operator/operator.go
jmdeal: Resolved offline.

pkg/operator/options/injectable.go
jmdeal: Resolved offline

pkg/operator/options/options.go
```diff
@@ -57,113 +60,95 @@ type OptionFields struct {
...
-		case "feature-gates":
-			o.FeatureGatesSet = true
-		}
-	})
-	if _, ok := os.LookupEnv("BATCH_MAX_DURATION"); ok {
-		o.BatchMaxDurationSet = true
-	}
-	if _, ok := os.LookupEnv("BATCH_IDLE_DURATION"); ok {
-		o.BatchIdleDurationSet = true
+		return fmt.Errorf("failed to validate cli flags / env vars, invalid log level %q", o.LogLevel)
```
jonathan-innis, line 94:
```suggestion
		return fmt.Errorf("validating cli flags / env vars, invalid log level %q", o.LogLevel)
```

nit: style with wording

pkg/operator/options/options.go
```diff
@@ -57,113 +60,95 @@ type OptionFields struct {
...
-
-func (o *Options) Parse(args ...string) (*Options, error) {
-	if err := o.FlagSet.Parse(args); err != nil {
+func (o *Options) Parse(fs *flag.FlagSet, args ...string) error {
+	if err := fs.Parse(args); err != nil {
 		if errors.Is(err, flag.ErrHelp) {
 			os.Exit(0)
 		}
-		return nil, err
+		return err
```
jonathan-innis, line 90:
```suggestion
		return fmt.Errorf("parsing flags, %w")
```

pkg/operator/options/options.go
```diff
@@ -57,113 +60,95 @@ type OptionFields struct {
...
-	}
-	if _, ok := os.LookupEnv("BATCH_IDLE_DURATION"); ok {
-		o.BatchIdleDurationSet = true
+		return fmt.Errorf("failed to validate cli flags / env vars, invalid log level %q", o.LogLevel)
 	}
-	if _, ok := os.LookupEnv("FEATURE_GATES"); ok {
-		o.FeatureGatesSet = true
+	gates, err := ParseFeatureGates(o.FeatureGates.inputStr)
+	if err != nil {
+		return fmt.Errorf("failed to parse feature gates, %w", err)
```
jonathan-innis, line 98:
```suggestion
		return fmt.Errorf("parsing feature gates, %w", err)
```

pkg/operator/options/suite_test.go
```diff
@@ -36,17 +39,47 @@ func TestOptions(t *testing.T) {
...
+			Entry("Basic false", "Drift=false", false),
+			Entry("With whitespace", "Driftt\t= false", false),
+			Entry("With additional gates", "Hello=true,Drift=false,World=true", false),
+		)
+		It("should successfully parse well formed feature gate str", func() {
+			for _, entry := range []struct {
+				gateStr  string
+				driftVal bool
+			}{
+				{"Drift = true", true},
```
jonathan-innis, line 66: `DescribeTable`?

jonathan-innis (approved):
Approved with some minor suggestions

pkg/operator/options/suite_test.go
jmdeal: It was originally to take advantage of BeforeEach but I don't think it's necessary after the interface changes. It probably does make sense to switch to a for loop.

jonathan-innis: I pretty strongly disagree against this syntax. This is the syntax of `go test` tests but is not the recommended way to create tests that have structure like this. If you find yourself creating a struct to store test parameters/expected results, that's a great use-case for `DescribeTable`.

jonathan-innis: See: https://onsi.github.io/ginkgo/#table-specs

jonathan-innis: Also, the other good reason to do the `DescribeTable` methodology is that when one of the test cases fails, you will have a description built-in to the failure vs. no detail with what you have right now

jmdeal: So are you saying that should be a `DescribeTable`? That's what it was when you left the original comment, I changed it after the fact. I'm happy to change it back to a `DescribeTable`, whatever works best stylistically.

jonathan-innis (approved):
LGTM 🚀

PR #601, opened 2023-10-13
chore: Update controller names for v1beta1

jonathan-innis:
Fixes #N/A 

**Description**

This change updates the controller names for the following controllers:
1. metric_scraper -> metrics.node
2. pod_metrics -> metrics.pod
3. provisioner_metrics -> metrics.provisioner
4. machine.garbagecollection -> nodeclaim.garbagecollection
5. provisioner_trigger -> provisioner.trigger
6. provisioner -> provisioner
7. daemonset -> state.daemonset
8. machine-state -> state.machine
9. node_state -> state.node
10. pod_state -> state.pod
11. provisioner_state -> state.provisioner
12. termination -> node.termination

`make presubmit`


pkg/controllers/provisioning/provisioner.go
```diff
@@ -100,7 +100,7 @@ func NewProvisioner(kubeClient client.Client, coreV1Client corev1.CoreV1Interfac
 }
 
 func (p *Provisioner) Name() string {
-	return "provisioner"
+	return "provisioner.scheduler"
```
njtran, line 103: Why would we call this `provisioner.scheduler`? Personally, I think keeping it as provisioner is sufficient, especially if we're following the pattern of referencing the file directory names.

pkg/controllers/termination/controller.go
```diff
@@ -64,7 +64,7 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
 }
 
 func (c *Controller) Name() string {
-	return "termination"
+	return "node.termination"
```
njtran, line 67: Same here, I think this should just be termination.

jonathan-innis: This one I'm actually convinced _should_ be `node.termination`. Realistically, I could see us putting this in the node directory. I think we should take the convention of prefixing with whatever object that we are acting on.

pkg/controllers/provisioning/provisioner.go
jonathan-innis: I could go either way on this. I'm fine leaving it if you feel strongly about it.

ellistarn: Pref provisioner.

ellistarn (dismissed):

njtran (approved):

PR #604, opened 2023-10-13
fix: Handle NodeClaims existing but being unknown in earlier versions

jonathan-innis:
Fixes #N/A 

**Description**

This PR makes a change so that reconciliations do not fail if you create some NodePools and NodeClaims with v1beta1 and then choose to rollback to an earlier version of Karpenter where these concepts are not understood.

`make presubmit`
`make apply`

### DIFF
--- a/pkg/controllers/state/suite_test.go
+++ b/pkg/controllers/state/suite_test.go
@@ -38,6 +38,8 @@ import (
 	"github.com/aws/karpenter-core/pkg/operator/options"
 	"github.com/aws/karpenter-core/pkg/operator/scheme"
 	"github.com/aws/karpenter-core/pkg/scheduling"
+	nodeclaimutil "github.com/aws/karpenter-core/pkg/utils/nodeclaim"
+	nodepoolutil "github.com/aws/karpenter-core/pkg/utils/nodepool"
 
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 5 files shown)


pkg/controllers/state/suite_test.go
```diff
@@ -683,6 +687,28 @@ var _ = Describe("Inflight Nodes", func() {
 			ExpectReconcileSucceeded(ctx, machineController, client.ObjectKeyFromObject(machine))
 			ExpectStateNodeNotFoundForMachine(machine)
 		})
+		It("should ignore node updates if nodes don't have provider id and are owned", func() {
```
njtran, line 690: shouldn't it be something like this as both machine nodes and nodeclaim nodes will have providerIDs?

I'd also add the machine part on the end so it's easier to know why it's necessary (although I just had to expand the file and see it was in the Machine section)
```suggestion
		It("should ignore node updates if nodes have old labels and are owned for machines", func() {
```

pkg/controllers/state/suite_test.go
```diff
@@ -1280,6 +1306,104 @@ var _ = Describe("Inflight Nodes", func() {
 			ExpectReconcileSucceeded(ctx, nodeClaimController, client.ObjectKeyFromObject(nodeClaim))
 			ExpectStateNodeNotFoundForNodeClaim(nodeClaim)
 		})
+		It("should ignore node updates if nodes don't have provider id and are owned", func() {
```
njtran, line 1309: same question as the previous
```suggestion
		It("should ignore node updates if nodes don't have provider id and are owned for nodeclaims", func() {
```

njtran (approved):

pkg/controllers/state/suite_test.go
jonathan-innis: The case here is that we don't respect node updates if we don't have a provider ID yet since we would effectively have double the capacity for a bit when we expect that node to be matched up with a Machine/NodeClaim very soon. I'm inclined to leave this description as-is.

pkg/controllers/state/suite_test.go
jonathan-innis: Same comment as above

njtran (approved):

PR #605, opened 2023-10-16
chore: Fix a couple misses where the entire disruption method was printed

jonathan-innis:
Fixes #N/A 

**Description**

This PR fixes a couple of places in the disruption code where the entire method struct was printed out rather than just the type of the method

`make presubmit`


jmdeal (approved):
LGTM 🚀

PR #610, opened 2023-10-17
chore: Inject the events.Recorder into the context.Context

### LINKED ISSUE #304: Inject the `events.Recorder` into the `context.Context`

It would be nice if we didn't have to pass the event recorder through every time that we wanted to use it in any of the standard contexts. Instead, we could do something like `recorder.FromContext(ctx).Publish()`

N/A

Passing the event recorder around through functions anytime that it's needed

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

sadath-12:
Fixes #304  

**Description**

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -137,7 +137,8 @@ type Offerings []Offering
 // passed zone and capacity type
 func (ofs Offerings) Get(ct, zone string) (Offering, bool) {
 	return lo.Find(ofs, func(of Offering) bool {
-		return of.CapacityType == ct && of.Zone == zone
+		return of.CapacityType == ct && of.Zone ==
+		 zone
 	})
 }
 
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -137,8 +137,7 @@ type Offerings []Offering
 // passed zone and capacity type
 func (ofs Offerings) Get(ct, zone string) (Offering, bool) {
 	return lo.Find(ofs, func(of Offering) bool {
-		return of.CapacityType == ct && of.Zone ==
-		 zone
+		return of.CapacityType == ct && of.Zone == zone
 	})
 }
 
--- a/pkg/controllers/provisioning/controller.go
+++ b/pkg/controllers/provisioning/controller.go
@@ -52,7 +52,7 @@ func (c *Controller) Name() string {
 }
 
 // Reconcile the resource
-func (c *Controller) Reconcile(ctx context.Context, p *v1.Pod) (reconcile.Result, error) {
+func (c *Controller) Reconcile(_ context.Context, p *v1.Pod) (reconcile.Result, error) {
 	if !pod.IsProvisionable(p) {
 		return reconcile.Result{}, nil
 	}
--- a/pkg/controllers/provisioning/controller.go
+++ b/pkg/controllers/provisioning/controller.go
@@ -40,11 +40,10 @@ type Controller struct {
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, provisioner *Provisioner, recorder events.Recorder) corecontroller.Controller {
+func NewController(kubeClient client.Client, provisioner *Provisioner) corecontroller.Controller {
 	return corecontroller.Typed[*v1.Pod](kubeClient, &Controller{
 		kubeClient:  kubeClient,
 		provisioner: provisioner,
-		recorder:    recorder,
 	})
 }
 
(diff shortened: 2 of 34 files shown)


sadath-12:
Not completed !! for the review

tzneal:
I don't see that this is performance related. The docs for context state:

> Package context defines the Context type, which carries deadlines, cancellation signals, and other request-scoped values across API boundaries and between processes. 

The event recorder isn't request-scoped, so I don't think it really belongs in the context.

sadath-12:
Any suggestions on what approach to be taken to close issue #304

pkg/operator/operator.go
```diff
@@ -183,10 +183,11 @@ func NewOperator() (context.Context, *Operator) {
 	lo.Must0(mgr.AddHealthzCheck("healthz", healthz.Ping))
 	lo.Must0(mgr.AddReadyzCheck("readyz", healthz.Ping))
 
+	ctx = events.ToContext(ctx,events.NewRecorder(mgr.GetEventRecorderFor(appName)))
```
jonathan-innis, line 186: Nice 🎉

pkg/controllers/provisioning/suite_test.go
```diff
@@ -71,7 +69,7 @@ var _ = BeforeSuite(func() {
 	fakeClock = clock.NewFakeClock(time.Now())
 	cluster = state.NewCluster(fakeClock, env.Client, cloudProvider)
 	nodeController = informer.NewNodeController(env.Client, cluster)
-	prov = provisioning.NewProvisioner(env.Client, corev1.NewForConfigOrDie(env.Config), events.NewRecorder(&record.FakeRecorder{}), cloudProvider, cluster)
+	prov = provisioning.NewProvisioner(env.Client, corev1.NewForConfigOrDie(env.Config), cloudProvider, cluster)
```
jonathan-innis, line 72: I think we may need to inject a recorder into the context generally for all of our test environments. If we don't pass it through, I think these tests will panic when we try to fire errors

pkg/controllers/provisioning/controller.go
```diff
@@ -53,7 +52,7 @@ func (c *Controller) Name() string {
 }
 
 // Reconcile the resource
-func (c *Controller) Reconcile(_ context.Context, p *v1.Pod) (reconcile.Result, error) {
+func (c *Controller) Reconcile(ctx context.Context, p *v1.Pod) (reconcile.Result, error) {
```
jonathan-innis, line 55: Why does this need to be changed?

pkg/controllers/provisioning/controller.go
```diff
@@ -40,11 +40,10 @@ type Controller struct {
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, provisioner *Provisioner, recorder events.Recorder) corecontroller.Controller {
+func NewController(kubeClient client.Client, provisioner *Provisioner) corecontroller.Controller {
 	return corecontroller.Typed[*v1.Pod](kubeClient, &Controller{
 		kubeClient:  kubeClient,
 		provisioner: provisioner,
-		recorder:    recorder,
```
jonathan-innis, line 47: I'm imagining that this recorder can just be dropped outright without any other changes to this file

pkg/cloudprovider/types.go
```diff
@@ -137,7 +137,8 @@ type Offerings []Offering
 // passed zone and capacity type
 func (ofs Offerings) Get(ct, zone string) (Offering, bool) {
 	return lo.Find(ofs, func(of Offering) bool {
-		return of.CapacityType == ct && of.Zone == zone
+		return of.CapacityType == ct && of.Zone ==
+		 zone
```
jonathan-innis, line 141: Why did we need to add a newline here?

sadath-12:
here there is problem , we have to pass the recorder to the baseContext but recorder depends on mgr which comes after baseContext

pkg/operator/operator.go
sadath-12: here there is problem , we have to pass the recorder to the baseContext but recorder depends on mgr which comes after baseContext

pkg/controllers/provisioning/suite_test.go
sadath-12: ya thats what thinkered me

pkg/controllers/provisioning/controller.go
sadath-12: By miss

jonathan-innis:
> also what is the hack that we can do for

I'm not sure that there is a good hack here. Considering what you are saying about the baseContext, I'm not convinced that we should do the injection. I'm more convinced that we should just leave the implementation as-is. Thoughts?

sadath-12:
> > also what is the hack that we can do for
> 
> I'm not sure that there is a good hack here. Considering what you are saying about the baseContext, I'm not convinced that we should do the injection. I'm more convinced that we should just leave the implementation as-is. Thoughts?

Ya it was my concern that avoiding that injection might break something . Ya ill leave that decision on you

jonathan-innis:
> Ya ill leave that decision on you

I think given the bit that you've discovered about the ordering of when the recording gets instantiated. I'm inclined to close this PR and close the issue as a `wont-fix`. I'm going to close both for now. Please feel free to re-open if you think one of them was closed by mistake.

PR #614, opened 2023-10-18
chore: correctly parse boolean CLI flags

jmdeal:
Fixes #N/A 

**Description**
This PR adds a wrapper to `BoolFunc` which enables merging CLI flags with environment variables correctly. With the previous implementation, the following is possible:
```
DISABLE_WEBHOOK=true ./karpenter --disable-webhook
# disable-webhook resolves to false in options
```
This is because of the way that `flag` handles default values for booleans. 

`make test`

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -83,15 +83,15 @@ func (fs *FlagSet) BoolVarWithEnv(p *bool, name string, envVar string, val bool,
 
 func (o *Options) AddFlags(fs *FlagSet) {
 	fs.StringVar(&o.ServiceName, "karpenter-service", env.WithDefaultString("KARPENTER_SERVICE", ""), "The Karpenter Service name for the dynamic webhook certificate")
-	fs.BoolVarWithEnv(&o.DisableWebhook, "disable-webhook", false, "DISABLE_WEBHOOK", "Disable the admission and validation webhooks")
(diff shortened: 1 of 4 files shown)


pkg/operator/options/options.go
```diff
@@ -64,9 +64,26 @@ type Options struct {
...
 }
 
-func (o *Options) AddFlags(fs *flag.FlagSet) {
+type FlagSet struct {
+	*flag.FlagSet
+}
+
+// BoolVarWithEnv defines a bool flag with a specified name, default value, usage string, and fallback environment
+// variable.
+func (fs *FlagSet) BoolVarWithEnv(p *bool, name string, val bool, envVar string, usage string) {
```
jonathan-innis, line 73:
```suggestion
func (fs *FlagSet) BoolVarWithEnv(p *bool, name string, envVar string, val bool, usage string) {
```

IMO, this should be the argument ordering

jonathan-innis (commented):
Are we going to add a new test for this new behavior?

pkg/operator/options/options.go
jmdeal: We probably should add testing, the reason I didn't is the interesting part to test is the merge semantic with the env var IMO and without a refactor to enable mocking it out idk what a good methodology to test that would be. I didn't think a refactor of that would really be in scope here but maybe it should be?

jonathan-innis (approved):
LGTM 🚀

PR #619, opened 2023-10-20
chore: change log level of initialization

njtran:
Fixes #N/A 

**Description**
Initialization is the last stage of machine/node claim creation. It should be at info level, since `created` and `registered` are also at that level.

make presubmit


jonathan-innis (approved):
LGTM 🚀

Issue #621, opened 2022-10-05
Taint nodes before deletion


hamishforbes:
### Tell us about your request

When removing nodes due to consolidation I would like to be able to apply a taint to the node before it is removed.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Reason for this is to be able to gracefully stop DaemonSet pods, see related issues below

I have [consul](https://consul.io) agents running on nodes via DaemonSet, these agents join the cluster.
If they are just killed then they sit around in the cluster as `failed`, if the pod is given a stop signal then it will gracefully leave the cluster and then exit.

When a node is just deleted it leaves a bunch of hanging agents in my Consul cluster.
Applying a `NoExecute` taint prior to deletion will evict those pods.

System DaemonSets (e.g. Kube-proxy) tolerate all taints and so this won't evict those.

### Are you currently working around this issue?

Without Karpenter nodes are generally only removed 
a) Manually, in which case I manually taint the node with a `noExecute` taint
b) By the [node-termination-handler](https://github.com/aws/aws-node-termination-handler) which is configured to add a taint as well

With Karpenter... well the workaround is to manually clear out failed nodes from my Consul cluster or get this feature added!

### Additional Context

https://github.com/aws/aws-node-termination-handler/issues/273
https://github.com/kubernetes/kubernetes/issues/75482

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

sftim:
This didn't really get triaged one way or the other.

/remove-lifecycle rotten

jonathan-innis:
> thinks those nodes have become eligible for consolidation

I know there was some discussion on this, but I think the difficult bit is determining what "eligible" means -- the other thing is that if you base it on something like "consolidateAfter", then you end up potentially PreferNoSchedule tainting you entire cluster

jonathan-innis:
I think given that #624 tracks the "taint on disruption" case and looking at the original context of this issue, which should be addressed now with `karpenter.sh/disrupted:NoSchedule`, I'm inclined to close this

PR #626, opened 2023-10-21
feat: added NoExecute taint to terminating Nodes

### LINKED ISSUE #624: Mega Issue: Node Disruption Lifecycle Taints
### Description

Karpenter has driven disruption of nodes through annotations and processes maintained in memory. 

Karpenter should drive disruption by through its own taint mechanism(s) while it discovers and executes disruption actions.

This issue proposes that each node owned by Karpenter will be in one of four states:

1. Not Disrupting (No Taints) - Karpenter doesn't want to disrupt this node, and neither does the user.
2. Candidate (PreferNoSchedule Taint) - Karpenter identifies a node as a possible option for disruption for any of the programmatic disruption mechanisms that Karpenter does - expiration, drift, consolidation. A node that's chosen as a candidate can always be removed from candidacy.
3. Disrupting (NoSchedule Taint) - Karpenter has validated and executed the disruption action for the node, and has begun the [standard flow](https://karpenter.sh/docs/preview/disruption/#disruption-controller).
Karpenter can fail to disrupt a node. If it does, the node will go back to Not Disrupting, where it may be picked up as a Candidate again later.
4. Terminating (NoExecute Taint) - Karpenter has deleted the node, triggering the [finalization logic](https://karpenter.sh/docs/concepts/deprovisioning/#termination-controller), where the last of the pods (e.g. Daemonsets) need to be evicted before terminating the underlying instance, then removing the node.
Once a node has begun terminating, there's no turning back. Karpenter will eventually terminate it.

sadath-12:
Fixes #624 Part 4 (Terminating:NoExecute Taint)

**Description**

This pr attempts to solve issue of #621 as mentioned #624 . Tainting the nodes as ```NoExecute``` before attempting to terminate and also along side tries to make the taints operation code structure a bit more scalable and reusable 

```make presubmit```

### DIFF
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -14,13 +14,24 @@ limitations under the License.
 
 package v1beta1
 
-import v1 "k8s.io/api/core/v1"
+import (
+	"fmt"
+
+	v1 "k8s.io/api/core/v1"
+)
 
 // Karpenter specific taints
 const (
 	DisruptionTaintKey             = Group + "/disruption"
 	DisruptingNoScheduleTaintValue = "disrupting"
+	TerminationTaintKey            = Group + "/termination"
+	TerminationNoExecuteTaintValue = "Termination"
 )
+
+var TaintFuncs = map[v1.Taint]func(taint v1.Taint) bool{
+	DisruptionNoScheduleTaint: IsDisruptingTaint,
+	TerminationNoExecuteTaint: IsTerminatingTaint,
+}
 
 var (
 	// DisruptionNoScheduleTaint is used by the deprovisioning controller to ensure no pods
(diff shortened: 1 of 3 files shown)


jonathan-innis:
Hey @sadath-12. I'd discuss more on the issue with @njtran about this one. He is driving the design around this issue and is planning to put together an RFC around the implications of this design. This change has a number of implications, once of which is that adding a `NoExecute` taint necessarily means that pods that don't tolerate the taint will sit on the node up to their `tolerationSeconds` (default is `5m`) up until they are removed. That means that nodes in general will sit on the cluster for `5m` longer on average, which is untenable on its own. I think that @njtran has some thoughts around this, but it would be good to discuss this over the issue.

sadath-12:
Sure,Thanks @jonathan-innis

njtran:
Added some information on the #629. I think this PR/Design should be explored and iterated on after that is completed, which I believe is the opposite as what you've described.

sadath-12:
Sure @njtran

PR #627, opened 2023-10-22
chore: Wait for instance type label to exist for managed nodes

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds a change that causes nodes that we are watching not to be reconciled into the state until the node has its providerID and has the instance type label applied to it. This function will pass-through still if the instance type label doesn't exist but the node is considered to be initialized with the `karpenter.sh/initialized` label.

`make presubmit`
`make apply`/`make e2etests`

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -262,12 +262,19 @@ func (c *Cluster) UpdateNode(ctx context.Context, node *v1.Node) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	managed := node.Labels[v1alpha5.ProvisionerNameLabelKey] != "" || node.Labels[v1beta1.NodePoolLabelKey] != ""
+	initialized := node.Labels[v1beta1.NodeInitializedLabelKey] != ""
 	if node.Spec.ProviderID == "" {
 		// If we know that we own this node, we shouldn't allow the providerID to be empty
-		if node.Labels[v1alpha5.ProvisionerNameLabelKey] != "" || (node.Labels[v1beta1.NodePoolLabelKey] != "") {
+		if managed {
 			return nil
 		}
 		node.Spec.ProviderID = node.Name
+	}
+	// If we have a managed node with no instance type label that hasn't been initialized,
+	// we need to wait until the instance type label gets propagated on it
+	if managed && node.Labels[v1.LabelInstanceTypeStable] == "" && !initialized {
+		return nil
 	}
 	n, err := c.newStateFromNode(ctx, node, c.nodes[node.Spec.ProviderID])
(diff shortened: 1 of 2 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -262,13 +262,20 @@ func (c *Cluster) UpdateNode(ctx context.Context, node *v1.Node) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	managed := node.Labels[v1alpha5.ProvisionerNameLabelKey] != "" || node.Labels[v1beta1.NodePoolLabelKey] != ""
+	initialized := node.Labels[v1beta1.NodeInitializedLabelKey] != ""
 	if node.Spec.ProviderID == "" {
 		// If we know that we own this node, we shouldn't allow the providerID to be empty
-		if node.Labels[v1alpha5.ProvisionerNameLabelKey] != "" || (node.Labels[v1beta1.NodePoolLabelKey] != "") {
+		if managed {
```
njtran, line 269: Shouldn't this be `!managed`?

pkg/controllers/state/cluster.go
```diff
@@ -262,13 +262,20 @@ func (c *Cluster) UpdateNode(ctx context.Context, node *v1.Node) error {
...
 			return nil
 		}
 		node.Spec.ProviderID = node.Name
 	}
+	// If we have a managed node with no instance type label that hasn't been initialized,
+	// we need to wait until the instance type label gets propagated on it
+	if managed && node.Labels[v1.LabelInstanceTypeStable] == "" && !initialized {
```
njtran, line 276: nit: any reason not to inline initialized since it doesn't look like you're using it otherwise?

njtran: Also, looking at the code, doesn't this instance type label get propagated on machine registration ?

pkg/controllers/state/cluster.go
jonathan-innis: No; this means that if we don't have the providerID and are managing the node, that we should reject updates right now since we expect there to be a providerID and if we allowed the update to go through, we could get a "ghost node" that should otherwise be matched up with a Machine.

pkg/controllers/state/cluster.go
jonathan-innis: It does. This is covering the case where we get a Node creation update for a node that we manage and it hasn't gotten its instance type label update yet from Karpenter.

This also is here for the machine migration case. We are about to be able to drop all of this logic so that we can just rely on the NodeClaim allocatable/capacity and not have to discover this information from the cloudprovider GetInstanceTypes()

njtran (approved):

PR #628, opened 2023-10-23
chore: Allow the `karpenter.sh/nodepool` labelSelector and validate nodeSelector on pods

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds the ability to use the `karpenter.sh/nodepool` on nodeSelectors and nodeAffinities. This change also validates the `nodeSelector` alongside the node affinities that we already validate

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -445,19 +445,27 @@ func (p *Provisioner) getDaemonSetPods(ctx context.Context) ([]*v1.Pod, error) {
 
 func (p *Provisioner) Validate(ctx context.Context, pod *v1.Pod) error {
 	return multierr.Combine(
-		validateProvisionerNameCanExist(pod),
+		validateKarpenterManagedLabelCanExist(pod),
+		validateNodeSelector(pod),
 		validateAffinity(pod),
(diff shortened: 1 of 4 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -446,6 +446,8 @@ func (p *Provisioner) getDaemonSetPods(ctx context.Context) ([]*v1.Pod, error) {
 func (p *Provisioner) Validate(ctx context.Context, pod *v1.Pod) error {
 	return multierr.Combine(
 		validateProvisionerNameCanExist(pod),
```
ellistarn, line 448: Is this supported for the v1beta1 apis?

ellistarn (dismissed):

pkg/controllers/provisioning/provisioner.go
jonathan-innis: That's a good point. I think we need to validate that both the labels have a `DoesNotExist` requirement; else, we are going to get into a situation where some pods _can_ actually schedule to Karpenter nodes that are managed by a NodePool.

I also need to add this detail to the Upgrade Guide. I'm not sure that there is a good way around this one.

ellistarn (approved):

PR #634, opened 2023-10-24
feat: added disrupting candidates for nodepool status

### LINKED ISSUE #493: Surface health conditions of Karpenter resources as status conditions
### Description

Karpenter uses logs and events to alert users about both fleeting and unresolvable errors about their configurations. Logs may be spammy due to the fast-nature of re-queues, and events can give delayed information on when a health issue is actually resolved (waiting until the de-dupe interval is complete). 

On top of this concern, it may be useful for a user to monitor their NodePools or NodeClasses to determine if they are all in a healthy state and fire alarms when certain status conditions are not maintained or, more generally, when the Readiness condition for either the NodePools or the NodeClasses isn't satisfied. 

High-level, I think we are thinking the following status conditions would be nice to have:
- [ ] NodePools
  - [ ] `NodeClassReady` - Rolled-up status conditions from the `nodeClassRef` which checks the generic readiness of this object to determine if the NodeClass is usable
  - [ ] `InstanceTypesExist` - Validates whether there are provisionable instance types with available offerings to use
  - [ ] `NodesDisrupting` - Describes whether there are any nodes on the cluster that are actively being disrupted through replacement or deletion. This condition _would not_ have an affect on the NodePool readiness but would be purely informational for alerting and tracking

We have logic that is sitting in the code that [ignores certain NodePools for scheduling](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/provisioning/provisioner.go#L236) and have some logic in the code for [handling cases where the NodeClass isn't ready](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/nodeclaim/lifecycle/launch.go#L116). This logic could strictly rely on these new status conditions to ensure that we are always attempting to schedule and provision nodes that are sitting in a ready state.

Additional Info
It's possible that we could then export metrics from these status conditions so that users can alarm/track these metrics.

*

sadath-12:
Fixes #493 ( ```NodesDisrupting``` part )

**Description**

Whenever karpenter has decided to disrupt some nodes it will be updated in status field ```NodesDisrupting``` of nodepool and when that's done ```NodesDisrupting``` will become empty again

I ran few tests and saw in the console if the status were updating


jonathan-innis:
This is another PR that adds API surface and I think needs a design before we can consider taking this into the project. There's some things that we need to consider here like the value of the API surface to users, are there other API avenues that we could surface the same information, how many times will this cause us to update the Provisioner and will that cause too many APIServer calls?

This is just a subset of questions that I have, but given these are questions that have large impact, I think a short design that's proposed with the changes could be helpful here.

sadath-12:
sure @jonathan-innis

PR #641, opened 2023-10-26
BREAKING CHANGE: do not terminate pods that tolerate disruption taint for nodeclaims

njtran:
Fixes #N/A 

**Description**
This ensures that pods on nodes with NodeClaims that tolerate the `karpenter.sh/disruption` taint aren't deleted, since they would reschedule.

If users want their pods to not be terminated, they can add a toleration for the `karpenter.sh/disruption:NoSchedule=disrupting` taint.

This also adds in test duplication for NodeClaims and Machines for termination.

make presubmit


pkg/controllers/node/termination/controller.go
```diff
@@ -83,7 +83,7 @@ func (c *Controller) Finalize(ctx context.Context, node *v1.Node) (reconcile.Res
 		return reconcile.Result{}, fmt.Errorf("deleting nodeclaims, %w", err)
 	}
 	if err := c.terminator.Cordon(ctx, node); err != nil {
```
jonathan-innis, line 85: I forget what the word "cordon" means here but is it still appropriate to call this cordoning if it's not using the standard upstream K8s taint?

pkg/controllers/node/termination/controller.go
```diff
@@ -83,7 +83,7 @@ func (c *Controller) Finalize(ctx context.Context, node *v1.Node) (reconcile.Res
 		return reconcile.Result{}, fmt.Errorf("deleting nodeclaims, %w", err)
```
jonathan-innis, line 83: Is it worth adding a printer-column over the disruption taint in this PR?

jonathan-innis: It potentially could just be printed as part of the `-o wide` output

pkg/controllers/node/termination/controller.go
```diff
@@ -136,7 +136,7 @@ func (c *Controller) deleteAllNodeClaims(ctx context.Context, node *v1.Node) err
 
 func (c *Controller) removeFinalizer(ctx context.Context, n *v1.Node) error {
 	stored := n.DeepCopy()
-	controllerutil.RemoveFinalizer(n, v1alpha5.TerminationFinalizer)
+	controllerutil.RemoveFinalizer(n, v1beta1.TerminationFinalizer)
```
jonathan-innis, line 139: Nice catch! 🎉

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -43,10 +45,19 @@ func NewTerminator(clk clock.Clock, kubeClient client.Client, eq *Queue) *Termin
...
+// If the node is linked to a machine, it will use the node.kubernetes.io/unschedulable taint.
 func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
 	stored := node.DeepCopy()
-	node.Spec.Unschedulable = true
+
+	if _, isMachine := node.Labels[v1alpha5.ProvisionerNameLabelKey]; isMachine {
+		node.Spec.Unschedulable = true
+	} else {
+		node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t v1.Taint, _ int) bool {
+			return t.Key == v1beta1.DisruptionTaintKey
```
jonathan-innis, line 57: Should we leverage our `v1beta1.IsDisruptingTaint` function that you created here?

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -68,23 +79,27 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
 // https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
 func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
 	// Get evictable pods
-	pods, err := t.getPods(ctx, node)
-	if err != nil {
-		return fmt.Errorf("listing pods for node, %w", err)
+	pods := &v1.PodList{}
```
jonathan-innis, line 82: Listing pods doesn't work because it excludes DaemonSet pods? Should we create a function that includes the daemonset pods (or even an option) so that we don't include terminating pods in this calculation here?

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -68,23 +79,27 @@ func (t *Terminator) Cordon(ctx context.Context, node *v1.Node) error {
...
 	}
-	var podsToEvict []*v1.Pod
+
+	_, isMachine := node.Labels[v1alpha5.ProvisionerNameLabelKey]
 	// Skip node due to pods that are not able to be evicted
-	for _, p := range pods {
-		// Ignore if unschedulable is tolerated, since they will reschedule
-		if podutil.ToleratesUnschedulableTaint(p) {
-			continue
+	podsToEvict := lo.FilterMap(pods.Items, func(po v1.Pod, _ int) (*v1.Pod, bool) {
```
jonathan-innis, line 89: Ah, I see that you are filtering out here but I wonder if this is best delegated to a helper function

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -94,46 +109,43 @@ func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
...
+	criticalDaemon := []*v1.Pod{}
+	nonCriticalNonDaemon := []*v1.Pod{}
+	nonCriticalDaemon := []*v1.Pod{}
 	for _, pod := range pods {
 		if !pod.DeletionTimestamp.IsZero() {
 			continue
 		}
 		if pod.Spec.PriorityClassName == "system-cluster-critical" || pod.Spec.PriorityClassName == "system-node-critical" {
-			critical = append(critical, pod)
+			if podutil.IsOwnedByDaemonSet(pod) {
```
jonathan-innis, line 123: A lot of if/else conditionals here. I'm not sure if there's a way to simplify or delegate this to a function, but maybe there's a way to simplify?

pkg/controllers/node/termination/node_claim_test.go
```diff
@@ -0,0 +1,537 @@
+/*
```
jonathan-innis, line 1: We've been doing `nodeclaim` as one word for file naming so rename this to `nodeclaim_test.go`

pkg/controllers/node/termination/node_claim_test.go
```diff
@@ -0,0 +1,537 @@
...
+			ExpectEvicted(env.Client, pod)
+
+			// After grace period, node should delete. The deletion timestamps are from etcd which we can't control, so
+			// to eliminate test-flakiness we reset the time to current time + 90 seconds instead of just advancing
+			// the clock by 90 seconds.
+			fakeClock.SetTime(time.Now().Add(90 * time.Second))
+			ExpectReconcileSucceeded(ctx, terminationController, client.ObjectKeyFromObject(node))
+			ExpectNotFound(ctx, env.Client, node)
+		})
+	})
```
jonathan-innis, line 514: "Should evict pods that tolerate the unschedulable taint" to make sure that this edge case is covered for NodeClaims

pkg/controllers/node/termination/machine_test.go
```diff
@@ -0,0 +1,512 @@
...
+			})
+
+			ExpectApplied(ctx, env.Client, node, podNoEvict)
+
+			Expect(env.Client.Delete(ctx, node)).To(Succeed())
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			ExpectReconcileSucceeded(ctx, terminationController, client.ObjectKeyFromObject(node))
+			node = ExpectNodeExists(ctx, env.Client, node.Name)
+			Expect(node.Labels[v1.LabelNodeExcludeBalancers]).Should(Equal("karpenter"))
+		})
```
jonathan-innis, line 124: "Should evict pods that tolerate the karpenter disruption taint" to make sure that this edge case is covered for Machines

pkg/controllers/node/termination/controller.go
njtran: I went back and forth on this one, and wanted to keep it the same as is, since we're still cordoning nodes. But I also changed the log, so I'm just going to change this to `Taint` now since Unschedulable is still a taint.

pkg/controllers/node/termination/controller.go
njtran: I'd prefer to look into it as a follow-up if that's cool 👍

pkg/controllers/node/termination/terminator/terminator.go
njtran: Yeah I had just done the key here since it should apply to the future taints as well, but since we haven't added that in, I'll revert to the function call 👍

pkg/controllers/node/termination/terminator/terminator.go
njtran: Listing pods does list DaemonSet pods.

pkg/controllers/node/termination/terminator/terminator.go
njtran: Another way we could do this is by doing 4 lo.Filter() functions, but that'd be doing 4N operations as opposed to just iterating through the array once. 

Minimal performance impact, but I think it's better this way.

pkg/controllers/node/termination/terminator/terminator.go
njtran: I think it's easier to see the filters that we apply if we do it in-line.

jonathan-innis: Decided to punt on this and open a chore issue to look into this later

jonathan-innis (approved):
LGTM 🚀

PR #652, opened 2023-10-31
Add GHA linting

jonathan-innis:
Fixes #N/A 

**Description**

Add linting for GHA workflows
`actionlint` does not currently support composite action checking but this feature appears to be in the works (https://github.com/rhysd/actionlint/pull/366). When that feature is implemented, we could also add that feature to our CI

`actionlint -oneline`

### DIFF
--- a/.github/workflows/presubmit.yaml
+++ b/.github/workflows/presubmit.yaml
@@ -15,6 +15,8 @@ jobs:
     - uses: ./.github/actions/install-deps
       with:
         k8sVersion: ${{ matrix.k8sVersion }}
+    - name: Enable the actionlint matcher
+      run: echo "::add-matcher::.github/actionlint-matcher.json"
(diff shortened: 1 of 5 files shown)


.github/workflows/presubmit.yaml
```diff
@@ -15,6 +15,8 @@ jobs:
     - uses: ./.github/actions/install-deps
       with:
         k8sVersion: ${{ matrix.k8sVersion }}
+    - name: Enable the actionlint matcher
+      run: echo "::add-matcher::.github/actionlint-matcher.json"
```
ellistarn, line 19: Can we not capture this inside `make verify`?

jonathan-innis: Not really necessary. This is _super_ specific to GHA so having this generalized feels unnecessary to me

engedaam (approved):
LGTM 🚀

PR #762, opened 2023-11-06
chore: remove machine taint code

njtran:
Fixes #N/A 

**Description**
Removes logic that uses the unschedulable taint for Machines

make presubmit && make apply


pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -46,18 +46,12 @@ func NewTerminator(clk clock.Clock, kubeClient client.Client, eq *Queue) *Termin
...
-
-	if _, isMachine := node.Labels[v1alpha5.ProvisionerNameLabelKey]; isMachine {
-		node.Spec.Unschedulable = true
-	} else {
-		node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t v1.Taint, _ int) bool {
-			return v1beta1.IsDisruptingTaint(t)
-		})
-		node.Spec.Taints = append(node.Spec.Taints, v1beta1.DisruptionNoScheduleTaint)
-	}
+	node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t v1.Taint, _ int) bool {
```
jonathan-innis, line 51: Maybe a comment since this logic might be a little confusing to someone just looking at it. One concern looking at this code: Does it have the potential to re-order taints if someone appended their own taint to the end after we taint it? Should we just handle the add with a single range loop so we don't have the potential to affect the order?

pkg/controllers/disruption/controller.go
```diff
@@ -230,14 +230,14 @@ func (c *Controller) launchReplacementNodeClaims(ctx context.Context, m Method,
 	stateNodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode })
 
 	// taint the candidate nodes before we launch the replacements to prevent new pods from scheduling to the candidate nodes
-	if err := c.requireNoScheduleTaints(ctx, true, stateNodes...); err != nil {
+	if err := c.requireNodeClaimNoScheduleTaint(ctx, true, stateNodes...); err != nil {
```
jonathan-innis, line 233:
```suggestion
	if err := c.requireNoScheduleTaint(ctx, true, stateNodes...); err != nil {
```

Do we need to be specific about NodeClaim/Machine now that we just have one concept?

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -45,17 +45,12 @@ func NewTerminator(clk clock.Clock, kubeClient client.Client, eq *Queue) *Termin
...
 func (t *Terminator) Taint(ctx context.Context, node *v1.Node) error {
 	stored := node.DeepCopy()
-
-	if _, isMachine := node.Labels[v1alpha5.ProvisionerNameLabelKey]; isMachine {
-		node.Spec.Unschedulable = true
-	} else {
-		node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t v1.Taint, _ int) bool {
-			return v1beta1.IsDisruptingTaint(t)
-		})
+	if _, ok := lo.Find(node.Spec.Taints, func(t v1.Taint) bool {
```
jonathan-innis, line 51: Nice 🎉

jonathan-innis (dismissed):
LGTM 🚀

jonathan-innis (approved):

PR #768, opened 2023-11-07
docs: RFC on design for spot->spot consolidation

### LINKED ISSUE #763: Spot -> Spot Consolidation

originally discussed in aws/karpenter#1091, but when the conversation started to be fruitful "aws locked as resolved and limited conversation to collaborators"

using karpenter costs a lot of money after a scale up when nodes are left running behind

various approached discussed in #

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

please don't close this and keep conversation going.

ellistarn:
Design for: #763 

**Description**

RFC for spot->spot consolidation.


designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
+# Spot Consolidation
```
jonathan-innis, line 1: Should we say that this PR is going to fix #763? Since that will close the issue, we probably just want this design to relate to it, but not close the original issue when this PR gets merged in.

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+Karpenter’s consolidation algorithm minimizes the price of EC2 instances given pod scheduling constraints. However, price minimization presents challenges for EC2 Spot, which offers a tradeoff between price and interruption rate. Karpenter uses EC2 Spot’s PriceCapacityOptimized (PCO) allocation strategy to provide a balanced approach to this tradeoff, and doesn't always select the cheapest instance type, since it may result in immediate interruption. If Karpenter were to consolidate spot capacity, it would repeatedly interrupt the same instance, walking down the PCO decision ladder, until only the lowest price remained. Because of these complications, Karpenter does not currently support "spot replacement" consolidation.
```
jonathan-innis, line 7:
```suggestion
Karpenter’s consolidation algorithm minimizes the price of EC2 instances given pod scheduling constraints. However, price minimization presents challenges for EC2 Spot, which offers a tradeoff between price and interruption rate. Karpenter uses EC2 Spot’s PriceCapacityOptimized (PCO) allocation strategy to provide a balanced approach to this tradeoff, and doesn't always select the cheapest instance type, since it may result in immediate interruption. If Karpenter were to consolidate spot capacity based purely on cost, it would repeatedly interrupt the same instance, walking down the PCO decision ladder, until only the lowest price remained. Because of these complications, Karpenter does not currently support "spot replacement" consolidation.
```

Just a clarifying point here

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+All consolidation actions contain an inherent tradeoff between price and availability. Consolidation actions are equivalent to a spot interruption, but without the deadline. This tradeoff applies to on-demand capacity, as well. This document explores approaches to enable consolidation for spot instances without regressing to lowest price. 
+
+## Design Options
+
+### 1: Minimum Flexibility [Recommended] 
+
+Karpenter could skip consolidation if the number of potential replacement instance types were below a threshold (e.g. 20). To avoid repeat interruptions, e.g. "walking down the ladder", Karpenter would have to limit its flexibility to this threshold. Further, Karpenter would have to limit its flexibility during the launch scenario as well, else face an immediate consolidation if the launched instance type’s price were more than the price of any instance type under the flexibility threshold. Since on-demand capacity uses the lowest price strategy, its behavior is unaffected if these rules are applied equally to both spot and on-demand launches.
```
jonathan-innis, line 15:
```suggestion
Karpenter could skip consolidation if the number of potential replacement instance types were below a threshold (e.g. 20). To avoid repeat interruptions, e.g. "walking down the ladder", Karpenter would have to limit its flexibility to this threshold. By ensuring that we have a minimum flexibility in our launch decisions, we can be assured that we have enough instance diversity that it's very unlikely that all instances that we select will have high interruption. Further, Karpenter would have to limit its flexibility during the launch scenario as well, else face an immediate consolidation if the launched instance type’s price were more than the price of any instance type under the flexibility threshold. Since on-demand capacity uses the lowest price strategy, its behavior is unaffected if these rules are applied equally to both spot and on-demand launches.
```

I think it might help here if we describe exactly _why_ this minimum flexibility requirement works for our consolidation algorithm

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+Conceptually, this approach is straightforward to explain to customers and directly aligned with Spot best practices. Some customers may be disappointed by the flexibility requirement, but given the current lack of support for spot consolidation, it’s net positive from the status quo. While there may be valid use cases for customers to trade interruption rate for lower prices, we should delay enabling customers to configure this value until the use cases are better understood. 
+
+* 👍👍👍 Prevents consolidation from regressing to a lowest price allocation strategy
+* 👍 Easy to explain and implements spot best practices
+* 👍 Flexibility threshold is tunable, as best practices evolve. Cloud providers could choose to set it to 1 to disable this behavior.
+* 👎 Limits potential flexibility from status quo for both instance launch and replacement. Today, Karpenter can send unbounded flexibility to PCO (e.g. 100s of instance types). 
+* 👎 Fails to take advantage of opportunities to migrate to healthier pools due to limited instance type flexibility.
```
jonathan-innis, line 23: This may be captured within this item, but it also doesn't allow us to consolidate to the "best" overall pool in all cases. This effectively causes us to get to a state of good enough that accounts for us getting probably 90% of the way there in cost reduction

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
+# Spot Consolidation
+
+## Problem Statement
```
jonathan-innis, line 3: Is it worth touching on the fact that most cloud providers don't surface capacity pool information for spot (for obvious reasons) which presents unique challenges that account for why we have to go with many of the solutions and their heuristics below

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+Karpenter’s consolidation algorithm minimizes the price of EC2 instances given pod scheduling constraints. However, price minimization presents challenges for EC2 Spot, which offers a tradeoff between price and interruption rate. Karpenter uses EC2 Spot’s PriceCapacityOptimized (PCO) allocation strategy to provide a balanced approach to this tradeoff, and doesn't always select the cheapest instance type, since it may result in immediate interruption. If Karpenter were to consolidate spot capacity, it would repeatedly interrupt the same instance, walking down the PCO decision ladder, until only the lowest price remained. Because of these complications, Karpenter does not currently support "spot replacement" consolidation.
+
+All consolidation actions contain an inherent tradeoff between price and availability. Consolidation actions are equivalent to a spot interruption, but without the deadline. This tradeoff applies to on-demand capacity, as well. This document explores approaches to enable consolidation for spot instances without regressing to lowest price. 
+
+## Design Options
+
+### 1: Minimum Flexibility [Recommended]
```
jonathan-innis, line 13: Not sure that this needs to be touched on in this design doc (just due to scoping) but we should think about how to handle cases where the user hasn't prescribed enough flexibility on their NodePool to begin with, which would effectively completely disable consolidation

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+This approach is predicated on customers configuring instance family flexibility, and relatively similar spot prices within a given size. For a customer flexible to all instance families within c,m,r categories, PCO would have a broad range of instance types to choose from, and would be unlikely to be forced to choose an unhealthy pool. This assumption follows EC2 spot best practices, and is relied upon by customers who use attribute-based selection with price protection.
+
+For customers who rely on instance size flexibility within a single family, or lack sufficient flexibility, they may face increased interruption rates compared to options 1 and 2. Due to the natural price differences between different instance sizes within the same family, this approach would limit flexibility to one or two instances sizes, depending on current spot markets. However, having limited flexibility goes against spot best practices, and the customer should expect increased interruptions.
```
jonathan-innis, line 33: This same problem exists for the first option, I think. Assuming reduced flexibility on instance types, this may actually be better than option 1 since it would at least allow you to consolidate past a given price threshold

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+### 3. Optimistically Launch Replacements
+
+Karpenter could optimistically launch new EC2 instances to probe PCO decision making. If PCO launches a cheaper instance type from a CreateFleet request that include the current instance type, we know that the instance is both lower price and lower interruption rate. If the PCO algorithm selects the current instance type of the running instance, Karpenter would terminate the new instance and leave the current instance running. This approach would need to include an optimistic launch cache to minimize the number of optimistic launches.
+
+* 👍👍👍 Prevents consolidation from regressing to a lowest price allocation strategy
+* 👎 May not make sense for all cloud providers
+* 👎 May consolidate to make marginal improvements, resulting in unnecessary disruption
+* 👎 Optimistic launches may confuse customers looking deeply at their bill or audit logs
+* 👎 Bugs with this mechanism are higher risk, since real instances are being launched.
```
jonathan-innis, line 53: Should we mention the cost component of this?

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+* 👎 During a spot capacity crunch, if many capacity pools within a price range are unhealthy, the customer may get interrupted soon after the consolidation. This is exacerbated if the customer cannot tolerate instance type diversity. 
+* 👎 Tuning this value to migitage both marginal-improvement-churn and lowest-price-regression is challenging, and desired values may conflict.
+
+Note: Regardless of the decision made to solve the spot consolidation problem, we’d likely want to implement a price improvement in the future to prevent consolidation from interrupting nodes to make marginal improvements.
+
+### 3. Optimistically Launch Replacements
+
+Karpenter could optimistically launch new EC2 instances to probe PCO decision making. If PCO launches a cheaper instance type from a CreateFleet request that include the current instance type, we know that the instance is both lower price and lower interruption rate. If the PCO algorithm selects the current instance type of the running instance, Karpenter would terminate the new instance and leave the current instance running. This approach would need to include an optimistic launch cache to minimize the number of optimistic launches.
```
jonathan-innis, line 47: Should we have `CreateFleet` in here? This is a generic design document and this has some AWS-specific lingo in here

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+There are multiple ways to model this time based constraint. Minimum node lifetime would prevent any consolidation of a node under a given age. Alternatively, this could be modeled as minimum pod lifetime, where consolidation would not terminate a node until all pods are above a specified age. This is the most direct approach to mitigate the problem statement: if there’s too much interruption, reduce frequency. We should delay (if ever) enabling customers from being able to configure this. 
+
+Alternatively, we could rely on ttlSecondsAfterUnderutilized, where Karpenter would detect a potential consolidation and wait for the ttl before taking action. This is somewhat unnatural, since we’ve seen customers configure the similar ttlSecondsAfterEmpty in the range of 30-180 seconds, which isn’t enough time to avoid regression to lowest price. It’s likely that customers would want one value (e.g. 30s) to handle empty nodes and another value to prevent repeated interruption (e.g. 4h). Further, given that we’re building ttlSecondsAfterUnderutilized to meet other cases, customers would be required to reason about both minimum node lifetime and ttlSecondsAfterUnderutilized.
```
jonathan-innis, line 61:
```suggestion
Alternatively, we could rely on ttlSecondsAfterUnderutilized, where Karpenter would detect a potential consolidation and wait for the ttl before taking action. This is somewhat unnatural, since we’ve seen customers configure the similar ttlSecondsAfterEmpty in the range of 30-180 seconds, which isn’t enough time to avoid regression to lowest price. It’s likely that customers would want one value (e.g. 30s) to handle empty nodes and another value to prevent repeated interruption (e.g. 4h). Further, given that we’re building consolidateAfter to meet other cases, customers would be required to reason about both minimum node lifetime and consolidateAfter. 
```

designs/spot-consolidation.md
```diff
@@ -0,0 +1,105 @@
...
+
+Alternatively, we could rely on ttlSecondsAfterUnderutilized, where Karpenter would detect a potential consolidation and wait for the ttl before taking action. This is somewhat unnatural, since we’ve seen customers configure the similar ttlSecondsAfterEmpty in the range of 30-180 seconds, which isn’t enough time to avoid regression to lowest price. It’s likely that customers would want one value (e.g. 30s) to handle empty nodes and another value to prevent repeated interruption (e.g. 4h). Further, given that we’re building ttlSecondsAfterUnderutilized to meet other cases, customers would be required to reason about both minimum node lifetime and ttlSecondsAfterUnderutilized. 
+
+* 👍 Avoids thrashing and delays regression to a lowest price allocations strategy
+* 👎 Eventually converges to a lowest price allocation strategy, which could increase interruptions
```
jonathan-innis, line 64: Just a musing/question: Does this strategy (combined with expiration) avoid this regression?

designs/spot-consolidation.md
ellistarn: I have no plans to close the issue when this design is merged.

designs/spot-consolidation.md
ellistarn: That's exactly the intent of this con. Is there something about the language I can make more clear?

designs/spot-consolidation.md
ellistarn: Yes, but it requires expiration to be configured to a very low value. I see this as a nonstarter to expect your average customer couple these. FWIW, you can currently achieve spot consolidation w/ expiration.

designs/spot-consolidation.md
ellistarn: In general, I'm trying to balance a readable narrative with a concrete cloudprovider example (e.g. EC2), while ensuring the design is flexible to multiple providers.

designs/spot-consolidation.md
ellistarn: 🤔 , I wonder if we could follow this line of reasoning to make the consolidation logic agnostic to this spot problem, and contain it within the AWS provider. e.g.

1. AWS Cloud Provider could reject a spot request w/ <20 flexibility (configurable?)
2. During a consolidation pre-spin, this would appear as a `InsufficientCapacityError`
3. Consolidation could remove the candidate from the pool. 
4. OD launches would be unaffected.

designs/spot-consolidation.md
jonathan-innis: Right, but your description has the word "fixes" that will tell GH to close it when this is merged

designs/spot-consolidation.md
jonathan-innis: I didn't see cost mentioned here. From "healthy" I was understanding availability, but I think cost improvements are also constrained here too

designs/spot-consolidation.md
jonathan-innis: Good point. SGTM

designs/spot-consolidation.md
ellistarn: Thinking on this, I think this is a bad idea -- you really need a way to communicate the minimum flexibility to the binpacker, since it's ultimately the actor constraining the space of instance types.

I wonder if we should consider modeling this as `offering.requiredFlexibility`. e.g., an offering is only valid if the request is sufficiently flexible. The cloud provider could choose how to model flexibility for any of its offerings (e.g. spot/od, or specific instance types).

ellistarn: FWIW, this could even be seen as an implementation detail of option 1, to answer the following question:

"How is minimum flexibility extensible to different cloud providers"

designs/spot-consolidation.md
ellistarn: I'm not sure cost is a major factor, unless there's a bug that causes them to not terminate for some reason.

designs/spot-consolidation.md
```diff
@@ -0,0 +1,106 @@
+# Spot Consolidation
+
+## Problem Statement
+
+Customers face increased costs due to underutilized spot instances. A customer with 4 x 24cpu pods would launch a 24xl (96cpu) instance type. If three of the four pods complete, the instance will indefinitely run at 25% utilization until the pod completes, is able to fit elsewhere in the cluster, or additional work is scheduled to the node. Customers have raised this inefficiency as a critical pain point with Karpenter: https://github.com/aws/karpenter-core/issues/763.
```
jackfrancis, line 5: Why are there only three options:

1. Wait for remaining pod to complete in order to GC spot node.
2. Rely on another compute slot on another, existing spot node to accommodate a gracefuol pod migration, then delete the original, larger spot node.
3. Wait for more spot workloads to arrive to remove overhead inefficiencies on larger, currently underutilized spot node.

And not this one as well:

4. Build a new, smaller spot node to run the remaining workload, migrate workload, and delete the original, larger spot node.

?

ellistarn: We refer to 4 as "replacement", and is the only thing in scope of this document, discussed in the second paragraph.

> If Karpenter were to consolidate spot capacity based purely on cost, it would repeatedly interrupt the same instance, walking down the PCO decision ladder, until only the lowest price remained. Because of these complications, Karpenter does not currently support "spot replacement" consolidation.

I can clarify this in the language.

designs/spot-consolidation.md
jonathan-innis: I'm not sure that you want to model this as a minimum flexibility. I think you probably want to model it as a maximum flexibility in some cases and a minimum flexibility in others. For launches, you want it to be a maximum flexibility since you don't want to walk down the ladder after you launch, so you want to constrain the options that you launch with to some maximal flexibility (like 20). 

The other thing to keep in mind here is I think that you still want to _launch_ instances that are under the flexibility threshold, you just know that you won't be able to consolidate them after they are under the threshold. 

Obviously, for consolidation, we are talking about a minimum flexibility. That is -- the minimum required instance types that need to come back from the scheduling call in order for the consolidation operation to be able to proceed. 

It almost seems like we need a general "flexibility" concept on instance types at a minimum that can capture both of these angles. We might even be able to get away with hard-coding it on the outset but we need to hear back from other cloudproviders if there is a good starting point that we can work from that we could then add as a parameter later.

jonathan-innis: There's also a bit of a question here on how this affects things and your ability to consolidate down the line if you pack things into a larger spot instance, which means that you would necessarily be under the flexibility metric on launch. My assumption is that this is still alright since you just won't consolidate until the instance gets small enough that the instance flexibility gets large enough to consolidate. The only down-side of this approach is that you may be interrupted at a higher rate depending on the availability of the instance types that you launch with under the fleixibility.

jonathan-innis: This might be a better route to go, though, then outright rejecting users that are working under this flexibility threshold today for launches.

designs/spot-consolidation.md
```diff
@@ -0,0 +1,106 @@
...
+
+Karpenter’s consolidation algorithm minimizes the price of EC2 instances given pod scheduling constraints. However, price minimization presents challenges for EC2 Spot, which offers a tradeoff between price and interruption rate. Karpenter uses EC2 Spot’s PriceCapacityOptimized (PCO) allocation strategy to provide a balanced approach to this tradeoff, and doesn't always select the cheapest instance type, since it may result in immediate interruption. If Karpenter were to consolidate spot capacity based purely on cost, it would repeatedly interrupt the same instance, walking down the PCO decision ladder, until only the lowest price remained. Because of these complications, Karpenter does not currently support "spot replacement" consolidation.
```
jackfrancis, line 7: Should we consider as an option maintaining a "consolidation state machine" that keeps track of the lifecycle of consolidated nodes. For example:

1. Node is created as a result of net new Pending pods (i.e., a vanilla scale out scenario)
2. We create a lifecycle representation of this new infra:
  - "Infra XYZ" created at time T0
 3. At some point in the future, a consolidation criteria triggers a replacement of this node with one or more replacement nodes. For each new node launched as a result of this consolidation event:
  - "Infra XYZ" replaced at time T1

Let's use the simplest possible example of a node that is continually consolidated via a single node replacement event:

```
Name: Infra XYZ
Node: <a reference to the node that represents the current node on the cluster that has this "infra" history>
Events:
- created at time T0
- replaced at time T1
- replaced at time T2
- replaced at time T3
Lifetime: <computed property of the difference between T3 and T0>
EligibleForConsolidation: <bool that represents whether or not the rate of events is happening too frequently given the lifetime of the infra; if the rate is above a threshold, the value is false, if the rate is below, the value is true>
```

Using the above example, we could expose a single configurable property that expressed something like "node replacement tolerance", i.e., how much node thrashing is allowed assuming other scheduling constraints permit graceful rebalancing of workloads across nodes.

Is something like this worth considering?

jonathan-innis:
This shouldn't have been closed out. We're working through the implementation of this design right now and should have this RFC merged soon.

ellistarn:
Handing off to Nikhil, who will take it from here in another PR

designs/spot-consolidation.md
jonathan-innis: TLDR on the discussion here: We have talked about `minNodeLifetime` as a parameter that we might want to support as well as price thresholding. Both would allow us to achieve some of the tweaking that is being suggested here. I could see us adding these things to _all_ of consolidation as a fast-follow to this base solution that unlocks the ability for us to perform consolidations with spot.

designs/spot-consolidation.md
```diff
@@ -0,0 +1,112 @@
...
...e with a cheaper spot node for both single and multi-node consolidation but with an additional condition for single-node consolidation. Cheaper spot instance types are selected with the `price-capacity-optimized` strategy and often the cheapest spot instance type is not launched due to the likelihood of interruption. Karpenter would replace spot node for single-node consolidation only if there are more than 15 potential replacement spot instance types. Karpenter would have to limit its flexibility to this threshold to avoid repeat interruptions, e.g. "walking down the ladder". By ensuring that we have a minimum flexibility in our launch decisions, we can be assured that we have enough instance diversity that it's very unlikely that all instances that we select will have high interruption. Further, Karpenter would have to limit its flexibility during the launch scenario as well, else face an immediate consolidation if the launched instance type’s price were more than the price of any instance type under the flexibility threshold. Since on-demand capacity uses the lowest price strategy, its behavior is unaffected if these rules are applied equally to both spot and on-demand launches.
```
jonathan-innis, line 17:
```suggestion
Karpenter will replace a spot node with a cheaper spot node for both single and multi-node consolidation but with an additional condition for single-node consolidation. Cheaper spot instance types are selected with the `price-capacity-optimized` strategy and often the cheapest spot instance type is not launched due to the likelihood of interruption. Karpenter would replace spot node for single-node consolidation only if there are more than 15 potential replacement spot instance types. Karpenter would have to limit its flexibility to this threshold to avoid repeat interruptions, e.g. "walking down the ladder". By ensuring that we have a minimum flexibility in our launch decisions, we can be assured that we have enough instance diversity that it's very unlikely that all instances that we select will have high interruption. Further, Karpenter would have to limit its flexibility while launching replacements with consolidation, else face an immediate consolidation if the launched instance type’s price were more than the price of any instance type under the flexibility threshold. Since on-demand capacity uses the lowest price strategy, its behavior is unaffected if these rules are applied equally to both spot and on-demand launches.
```

nit: just to clarify with wording here

designs/spot-consolidation.md
nikmohan123: Added a paragraph stating the reason for the requirement of heuristics in various solutions that we have suggested.

designs/spot-consolidation.md
jonathan-innis, line 17: It might be nice to callout why we made this decision _only_ for single node consolidation. Basically, the issue here is that we basically think that multi-node consolidation decisions that required the minimum instance type flexibility wouldn't occur frequently enough if we enforced it. On top of this, there's some considerations that we should take into account with our current handling for mutli-node spot such as the mixed capacity type scenario that it might be good to callout here. 

In either case, I'd expand on the rationale with why we went this direction so that it's a bit more fleshed out.

designs/spot-consolidation.md
jonathan-innis: @nikmohan123 Can you expound on this item here?

nikmohan123: Addressed.

designs/spot-consolidation.md
nikmohan123: Added more details regarding why we chose this flexibility only for single-node and why did we decide to do it only in the consolidation path.

nikmohan123: Addressed.

designs/spot-consolidation.md
```diff
@@ -0,0 +1,114 @@
...
+
+All consolidation actions contain an inherent tradeoff between price and availability. Consolidation actions are equivalent to a spot interruption, but without the deadline. This tradeoff applies to on-demand capacity, as well. This document explores approaches to enable consolidation for spot instances without regressing to lowest price. 
+
+Each of the below solution works on a certain heuristics because most cloud providers don't surface capacity pool information for spot which presents unique challenges to efficiently select a certain spot instance type that would cater to better price and availability with respect to the current node. Hence, the idea is to enable the consolidation based on different heuristics so that it makes sense to re-size the nodes only when there is significant cost benefit.
```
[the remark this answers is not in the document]
nikmohan123: Updated and re-wrote few other parts of the doc to make it more concise.

designs/spot-consolidation.md
```diff
@@ -0,0 +1,112 @@
...
+Minimum flexibility of 15 spot instance types is decided after an analysis done on the flexiblity of AWS customers request in the launch path. Karpenter does not have the same instance type flexibility requirement for multi-node spot-to-spot consolidations (many nodes to 1 node) since doing so without requiring flexibility won't lead to "race to the bottom" scenarios. It is also to prevent inconsistent behavior between consolidation of mixed capacity type [od,spot] to [spot] and spot capacity type [spot,spot] to [spot] since we already support mixed capacity type multi-node consolidation to spot today without any additional conditions. And finally, the limits is placed only for launching replacements with consolidation instead of the entire launch path inorder to unblock scenarios where a user doesn't currently have enough flexibility with spot. As a result, they will get their intial launch but then they just won't continually consolidate after that launch and the trade-off would be that the user with enough flexibility in their initial requests might get an immediate consolidation after the intial launch due to the way that we are constraining ourselves during the consolidation.
```
jonathan-innis, line 17:
```suggestion
Minimum flexibility of 15 spot instance types is decided after an analysis done on the flexiblity of AWS customers request in the launch path. Karpenter does not have the same instance type flexibility requirement for multi-node spot-to-spot consolidations (many nodes to 1 node) since doing so without requiring flexibility won't lead to "race to the bottom" scenarios. It is also to prevent inconsistent behavior between consolidation of mixed capacity type [od,spot] to [spot] and spot capacity type [spot,spot] to [spot] since we already support mixed capacity type multi-node consolidation to spot today without any additional conditions.

Finally, a limit for the number of instance types in the launch path is placed only on consolidation instead of the entire launch path in the provisioning loop in order to unblock scenarios where a user doesn't have enough flexibility with spot specified in their NodePool. As a result, their initial launch would succeed but would get no consolidations after their initial launch.
```

nit: Just trying to clarify the point that you are trying to callout here

nikmohan123: Updated.

jonathan-innis (approved):
/lgtm
/approve

PR #778, opened 2023-11-11
fix: issue#723 -- Do-not-evict only for running pods

### LINKED ISSUE #723: "Do not evict" only for running pods

We configure all of our helm hooks and other important pods with the do not evict flag. Some pods might not move to the running state because of having errors like CreateContainerError, CreateContainerConfigError or others.

Therefore we would like to have a better karpenter behaviour, only consider the do not evict flag if a given pod is in state creating or running...  

Right now a single pod can keep a single node running for unlimited time. After some time this is very costly... Imagine having dozens of nodes in a left behind state. 

Manually kill these pods. Which does not work if you have dozens of teams and projects running on a given cluster...

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

mallow111:
Fixes #723 

**Description**
Check if the pod has do-not-evict annotation also in ready status, then we treat this pod as non-evictable. Otherwise, when pods are not in ready status, even if it has do-not-evict annotation, we can still evict the pods.

I tested this in my local cluster.


pkg/utils/pod/scheduling.go
```diff
@@ -69,6 +69,15 @@ func IsOwnedByNode(pod *v1.Pod) bool {
 	})
 }
 
+func IsPodReady(pod *v1.Pod) bool {
+	for _, condition := range pod.Status.Conditions {
+		if condition.Type != v1.PodReady {
```
jonathan-innis, line 74: I think you may have intended to check that the `condition.Type` is the `v1.PodRedy` type and then you want to check if this is true rather than just checking if there is _any_ condition that isn't the PodReadiness condition

mallow111: updated

jmdeal:
There are a few concerns that were laid out in the original issue that I don't think are addressed here. The main concern with the original feature request is that it isn't clear which container states are non-recoverable. We don't want to ignore the `do-not-evict` annotation if the pod could eventually become ready. Otherwise, the pod may become ready between the time Karpenter chooses to disrupt the node and it is removed. By waiting for the pod to become ready before respecting the `do-not-evict` annotation this could happen frequently during normal operation, such as pulling the container image. Do you have any thoughts on how to address these concerns?

mallow111:
@jmdeal @jonathan-innis  what do you guys recommend the fix here, the initial issue has been discussed back and forth, but there is no solid solution there. What I am proposing is to make sure do-not-evict annotation only applies on running pods, for pods not ready, it contains various reasons, if we want to set a waiting period, how long it should wait, also is that worthy the effort to wait for those un ready pods?

jmdeal:
Our stance was that Karpenter should only ignore the `do-not-evict` annotation in cases where it is certain that the pod will not become ready. Currently Karpenter does this when a pod is in a terminal state, i.e. Succeeded / Failed, or if it is terminating ([ref](https://github.com/kubernetes-sigs/karpenter/blob/095d037295eed869d33bd190a827e8311756bc3f/pkg/controllers/disruption/helpers.go#L308-L315)).

We don't want to only consider the annotation for running pods because it is possible that a pending pod becomes ready between the time of the disruption decision and the time that the pod is removed. IMO what this PR would have to do is determine a set of container states that are truly unrecoverable. Then, if a pod is pending but one of it's containers is in an unrecoverable state we know we could disrupt it. I haven't looked into this deeply myself but from [@njtran's comment](https://github.com/kubernetes-sigs/karpenter/issues/723#issuecomment-1790109668) on the original issue it doesn't look like there's a standard set of enum values for this. Without this standard set of 'unrecoverable values', anything Karpenter does here would be pretty opinionated and potentially dangerous.

mallow111:
If there is no standard set of enum values, is there a way to move forward the change. Shall we put the issue as hold until there is such a standard coming out, otherwise I cannot see how shall we proceed the fix.

jmdeal:
Ya, I don't think we can move forward with this unless this core concern is addressed. Like I said, I haven't looked to deeply into it myself so it may be possible with the information available today but it's not immediately obvious.

mallow111:
@jonathan-innis do you agree with the above comments, I am going to close this PR if you agree on the above.

PR #783, opened 2023-11-14
fix: correctly handle non-existent / empty StorageClasses

### LINKED ISSUE #664: Karpenter stuck on "Waiting On Cluster Sync"
### Description

**Observed Behavior**:

Hi everyone, I've deployed Karpenter using the corresponding Helm chart for version v0.31.0 but i encounter a very strange problem: the main replica keeps logging "Waiting on cluster sync" and basically doesn't spawn any nodes. I've checked the AWS Auth configmap of my Cluster and the role is correctly present as well as the Karpenter service account and basic Karpenter configmap. In the past months I've succesfully installed Karpenter v.0.27.3 in more than 15 clusters at my company using Helm chart v0.27.3 and we didn't encounter problems at all. 

(The Machine CRD was added to the workflow as described in the docs, since it was added in v0.28)

These are the logs

```
2023-10-27T17:01:47.700Z        DEBUG   controller.provisioner  waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:48.689Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:49.690Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:50.691Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:51.692Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:52.693Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:53.693Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:54.694Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:55.695Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:56.695Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:57.696Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:57.701Z        DEBUG   controller.provisioner  waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:58.697Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:01:59.697Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:00.698Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:01.699Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:02.699Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:03.700Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:04.700Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:05.701Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:06.701Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:07.702Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:07.702Z        DEBUG   controller.provisioner  waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:08.702Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:09.703Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:10.704Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:11.705Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:12.705Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:13.706Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:14.707Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:15.708Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:16.708Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:17.703Z        DEBUG   controller.provisioner  waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:17.709Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:18.710Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:19.711Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:20.712Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:21.712Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:22.713Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:23.714Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:24.715Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:25.715Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:26.716Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:27.704Z        DEBUG   controller.provisioner  waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:27.717Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:28.717Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:29.718Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:30.719Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:31.719Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:32.720Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:33.720Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:34.721Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:35.722Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:36.722Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:37.704Z        DEBUG   controller.provisioner  waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:37.723Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:38.724Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:39.724Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
2023-10-27T17:02:40.725Z        DEBUG   controller.deprovisioning       waiting on cluster sync {"commit": "61b3e1e-dirty"}
```

Another thing i noticed is that Machine resources are not getting created at all after deploying my AWSNodeTemplate and Provisioner. I fear that I'm missing some AWS-related resources, maybe were introduced after 0.27.3, but from the docs it doesn't seem anything AWS-related was added. The nodes are not even started (I can't see them in EC2 Console)

**Expected Behavior**:

I should see nodes getting started from AWS EC2 Console first and then I should see them getting attached to the Cluster

```
apiVersion: v1
kind: Namespace
metadata:
  name: karpetenter-test-ns
---
kind: Provisioner
metadata:
  name: provisioner-test
spec:
  consolidation:
    enabled: true
  labels:
    intent: apps
  requirements:
      operator: In
      values:
        - c
        - m
        - r
        - t
      operator: In
      values:
        - ondemand
        - spot
    - key: kubernetes.io/arch
      operator: In
      values:
        - amd64
    - key: karpenter.k8s.aws/instance-size
      operator: NotIn
      values:
        - nano
        - micro
        - small
        - medium
        - large
  limits:
    resources:
      cpu: 1000
      memory: 1000Gi
  ttlSecondsUntilExpired: 2592000
  providerRef:
    name: node-template-test

---
apiVersion: karpenter.k8s.aws/v1alpha1
kind: AWSNodeTemplate
metadata:
  name: node-template-test
spec:
  securityGroupSelector:
    karpenter.sh/discovery: mycluster
  subnetSelector:
    karpenter.sh/discovery: mycluster
  tags:
    KarpenerProvisionerName: provisioner-test
    NodeType: karpenter-workshop
    IntentLabel: apps
    Gias_ID: test
    Env: test
    Name: test
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: inflate
  namespace: karpetenter-test-ns
spec:
  replicas: 5
  selector:
    matchLabels:
      app: inflate
  template:
    metadata:
      labels:
        app: inflate
    spec:
      nodeSelector:
        intent: apps
      containers:
        - name: inflate
          image: public.ecr.aws/eks-distro/kubernetes/pause:3.2
          resources:
            requests:
              cpu: 1
              memory: 1.5Gi
```

This is the Cloudformation used to create Karpenter resources on AWS

```
#Template
AWSTemplateFormatVersion: '2010-09-09'

Description: EKS infrastructure for Karpenter

Parameters:
  EksClusterName:
    Type: String
    Description: Name of existing EKS Cluster
  QueueName:
    Type: String
    Description: Name of SQS Queue to create where interruption events will be notified
  RoleNamePrefix:
    Type: String
  EventBridgeRulePrefix:
    Type: String
  OidcEndpoint:
    Type: String
    Description: EKS OIDC Endpoint without https://

Resources:
  NodeRole:
    Type: 'AWS::IAM::Role'
    Properties:
      RoleName: !Join [ "-", [!Ref RoleNamePrefix, "KarpenterNodeRole"] ]
      AssumeRolePolicyDocument:
        Version: "2012-10-17"
        Statement:
          - Effect: Allow
            Principal:
              Service:
                - ec2.amazonaws.com
            Action:
              - 'sts:AssumeRole'
      Path: /
      ManagedPolicyArns:
        - arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy
        - arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy
        - arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly
        - arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
      Policies:
        - PolicyName: CWLogsPolicy
          PolicyDocument:
            Version: "2012-10-17"
            Statement:
              - Effect: Allow
                Action:
                  - logs:DescribeLogGroups
                  - logs:DescribeLogStreams
                  - logs:CreateLogGroup
                  - logs:CreateLogStream
                  - logs:PutLogEvent
                  - logs:PutRetentionPolicy
                Resource:
                    - "*"
        - PolicyName: ECRPullPushPolicy
          PolicyDocument:
            Version: "2012-10-17"
            Statement:
              - Effect: Allow
                Action:
                  - ecr:BatchCheckLayerAvailability
                  - ecr:GetDownloadUrlForLayer
                  - ecr:GetRepositoryPolicy
                  - ecr:DescribeRepositories
                  - ecr:ListImages
                  - ecr:DescribeImages
                  - ecr:BatchGetImage
                  - ecr:GetAuthorizationToken
                Resource:
                  - "*"
        - PolicyName: LambdaInVpcPolicy
          PolicyDocument:
            Version: "2012-10-17"
            Statement:
              - Effect: Allow
                Action:
                  - ec2:DescribeAvailabilityZones
                  - ec2:DescribeNetworkInterfaceAttribute
                  - ec2:DescribeNetworkInterfaces
                  - ec2:DescribeSecurityGroups
                  - ec2:DescribeSubnets
                  - autoscaling:DescribeVpcAttribute
                  - elasticfilesystem:Describe*
                  - kms:ListAliases
                Resource:
                  - "*"
        - PolicyName: ClusterAutoscalerPolicy
          PolicyDocument:
            Version: "2012-10-17"
            Statement:
              - Effect: Allow
                Action:
                  - autoscaling:DescribeAutoScalingGroups
                  - autoscaling:DescribeAutoScalingInstances
                  - autoscaling:DescribeLaunchConfigurations
                  - autoscaling:DescribeTags
                  - autoscaling:SetDesiredCapacity
                  - autoscaling:TerminateInstanceInAutoScalingGroup
                Resource:
                  - "*"                  
  NodeInstanceProfile:
    Type: 'AWS::IAM::InstanceProfile'
    Properties:
      Path: /
      InstanceProfileName:  !Join [ "-", [!Ref RoleNamePrefix, "KarpenterInstanceProfile"] ]
      Roles:
        - !Ref NodeRole
  ControllerRole:
    Type: 'AWS::IAM::Role'
    Properties:
      RoleName: !Join [ "-", [!Ref RoleNamePrefix, "KarpenterCtrlRole"] ]
      AssumeRolePolicyDocument: !Sub
        - |
          {
              "Version": "2012-10-17",
              "Statement": [
                  {
                      "Effect": "Allow",
                      "Principal": {
                          "Federated": "arn:${AWS::Partition}:iam::${AWS::AccountId}:oidc-provider/${OidcEndpoint}"
                      },
                      "Action": "sts:AssumeRoleWithWebIdentity",
                      "Condition": {
                          "StringEquals": {
                              "${OidcEndpoint}:aud": "sts.amazonaws.com",
                              "${OidcEndpoint}:sub": "system:serviceaccount:karpenter:karpenter"
                          }
                      }
                  }
              ]
          }
        - OidcEndpoint: !Ref OidcEndpoint
      Path: /
      Policies:
        - PolicyName: karpenter-controller
          PolicyDocument: !Sub
            - |
              {
                  "Statement": [
                      {
                          "Action": [
                              "ssm:GetParameter",
                              "ec2:DescribeImages",
                              "ec2:RunInstances",
                              "ec2:DescribeSubnets",
                              "ec2:DescribeSecurityGroups",
                              "ec2:DescribeLaunchTemplates",
                              "ec2:DescribeInstances",
                              "ec2:DescribeInstanceTypes",
                              "ec2:DescribeInstanceTypeOfferings",
                              "ec2:DescribeAvailabilityZones",
                              "ec2:DeleteLaunchTemplate",
                              "ec2:CreateTags",
                              "ec2:CreateLaunchTemplate",
                              "ec2:CreateFleet",
                              "ec2:DescribeSpotPriceHistory",
                              "pricing:GetProducts"
                          ],
                          "Effect": "Allow",
                          "Resource": "*",
                          "Sid": "Karpenter"
                      },
                      {
                          "Action": "ec2:TerminateInstances",
                          "Condition": {
                              "StringLike": {
                                  "ec2:ResourceTag/karpenter.sh/provisioner-name": "*"
                              }
                          },
                          "Effect": "Allow",
                          "Resource": "*",
                          "Sid": "ConditionalEC2Termination"
                      },
                      {
                          "Effect": "Allow",
                          "Action": "iam:PassRole",
                          "Resource": "arn:${AWS::Partition}:iam::${AWS::AccountId}:role/${NodeRole}",
                          "Sid": "PassNodeIAMRole"
                      },
                      {
                          "Effect": "Allow",
                          "Action": "eks:DescribeCluster",
                          "Resource": "arn:${AWS::Partition}:eks:${AWS::Region}:${AWS::AccountId}:cluster/${ClusterName}",
                          "Sid": "EKSClusterEndpointLookup"
                      }
                  ],
                  "Version": "2012-10-17"
              }
            - ClusterName: !Ref EksClusterName
        - PolicyName: karpenter-interruption
          PolicyDocument:
            Version: "2012-10-17"
            Statement:
              - Effect: Allow
                Action:
                  - "sqs:DeleteMessage"
                  - "sqs:GetQueueAttributes"
                  - "sqs:GetQueueUrl"
                  - "sqs:ReceiveMessage"
                Resource: !Sub
                  - "arn:${AWS::Partition}:sqs:${AWS::Region}:${AWS::AccountId}:${SqsQueueName}"
                  - SqsQueueName: !GetAtt [ SqsQueue, "QueueName" ]
  SqsQueue: 
    Type: AWS::SQS::Queue
    Properties: 
      QueueName: !Ref QueueName
      MessageRetentionPeriod: 300
      SqsManagedSseEnabled: true
  SqsQueuePolicy:
    Type: AWS::SQS::QueuePolicy
    Properties:
      Queues:
        - !Ref SqsQueue
      PolicyDocument:
        Statement:
          - Action: 
              - "SQS:SendMessage" 
            Effect: "Allow"
            Resource: !Sub
              - "arn:${AWS::Partition}:sqs:${AWS::Region}:${AWS::AccountId}:${SqsQueueName}"
              - SqsQueueName: !GetAtt [ SqsQueue, "QueueName" ]
            Principal:  
              Service:
                - "events.amazonaws.com"
                - "sqs.amazonaws.com"

  HealthEventRule:
    Type: 'AWS::Events::Rule'
    Properties:
      Description: AWS Health Event to SQS Queue
      Name: !Join [ "-", [ !Ref EventBridgeRulePrefix, "HE" ] ]
      State: ENABLED
      EventPattern:
        detail:
          service:
          - EC2
        detail-type:
        - AWS Health Event
        source:
        - aws.health
      Targets:
        - Arn: !GetAtt [ SqsQueue, "Arn" ]
          Id: 'KarpenterInterruptionQueueTarget'

  SpotInterruptionRule:
    Type: 'AWS::Events::Rule'
    Properties:
      Description: AWS Spot Interruption to SQS Queue
      Name: !Join [ "-", [ !Ref EventBridgeRulePrefix, "SI" ] ]
      State: ENABLED
      EventPattern:
        detail-type:
        - EC2 Spot Instance Interruption Warning
        source:
        - aws.ec2
      Targets:
        - Arn: !GetAtt [ SqsQueue, "Arn" ]
          Id: 'KarpenterInterruptionQueueTarget'

  InstanceRebalanceRule:
    Type: 'AWS::Events::Rule'
    Properties:
      Description: AWS Instance Rebalance to SQS Queue
      Name: !Join [ "-", [ !Ref EventBridgeRulePrefix, "IR" ] ]
      State: ENABLED
      EventPattern:
        detail-type:
        - EC2 Instance Rebalance Recommendation
        source:
        - aws.ec2
      Targets:
        - Arn: !GetAtt [ SqsQueue, "Arn" ]
          Id: 'KarpenterInterruptionQueueTarget'

  InstanceStateChangeRule:
    Type: 'AWS::Events::Rule'
    Properties:
      Description: AWS Instance State Change to SQS Queue
      Name: !Join [ "-", [ !Ref EventBridgeRulePrefix, "ISC" ] ]
      State: ENABLED
      EventPattern:
        detail:
          state:
          - stopping
          - terminated
          - shutting-down
          - stopped
        detail-type:
        - EC2 Instance State-change Notification
        source:
        - aws.ec2
      Targets:
        - Arn: !GetAtt [ SqsQueue, "Arn" ]
          Id: 'KarpenterInterruptionQueueTarget'

Outputs:
  NodeRoleARN:
    Value: !GetAtt [ NodeRole, "Arn" ]
  InstanceRoleProfileName:
    Value: !Ref NodeInstanceProfile
  ControllerRoleARN:
    Value: !GetAtt [ ControllerRole, "Arn" ]
  QueueName:
    Value: !GetAtt [ SqsQueue, "QueueName" ]
```

**Versions**:
- Karpenter: v0.31.0
- Chart Version: 0.16.3
- Kubernetes Version (`kubectl version`): 1.25

jmdeal:
Fixes #664 

**Description**
Karpenter doesn't correctly handle two StorageClass cases: when the StorageClass is **explicitly** `""` and when the StorageClass does not exist. A pod with a PVC that falls into one of these cases should be able to successfully schedule if the PVC is already bound to a PV. Additionally, currently Karpenter treats both a `nil` StorageClass and `""` StorageClass as the default StorageClass. This should only be the case when the StorageClass is `nil`.

`make test`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/volumetopology.go
+++ b/pkg/controllers/provisioning/scheduling/volumetopology.go
@@ -23,10 +23,9 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/logging"
-	"knative.dev/pkg/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	"sigs.k8s.io/karpenter/pkg/scheduling"
+	volumeutil "sigs.k8s.io/karpenter/pkg/utils/volume"
 )
 
 func NewVolumeTopology(kubeClient client.Client) *VolumeTopology {
@@ -76,30 +75,15 @@ func (v *VolumeTopology) Inject(ctx context.Context, pod *v1.Pod) error {
 }
 
 func (v *VolumeTopology) getRequirements(ctx context.Context, pod *v1.Pod, volume v1.Volume) ([]v1.NodeSelectorRequirement, error) {
-	defaultStorageClassName, err := scheduling.DiscoverDefaultStorageClassName(ctx, v.kubeClient)
+	pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, v.kubeClient, pod, volume)
 	if err != nil {
-		return nil, fmt.Errorf("discovering default storage class, %w", err)
+		return nil, fmt.Errorf("discovering persistent volume claim, %w", err)
 	}
-
-	// Get VolumeName and StorageClass name from PVC
-	pvc := &v1.PersistentVolumeClaim{}
-	switch {
-	case volume.PersistentVolumeClaim != nil:
-		if err = v.kubeClient.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: volume.PersistentVolumeClaim.ClaimName}, pvc); err != nil {
-			return nil, fmt.Errorf("discovering persistent volume claim, %w", err)
-		}
-	case volume.Ephemeral != nil:
-		// generated name per https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#persistentvolumeclaim-naming
-		if err = v.kubeClient.Get(ctx, types.NamespacedName{Namespace: pod.Namespace, Name: fmt.Sprintf("%s-%s", pod.Name, volume.Name)}, pvc); err != nil {
-			return nil, fmt.Errorf("discovering persistent volume claim for ephemeral volume, %w", err)
-		}
-	default:
+	// Not all volume types have PVCs, e.g. emptyDir, hostPath, etc.
+	if pvc == nil {
 		return nil, nil
 	}
-	storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
-	if storageClassName == "" {
-		storageClassName = defaultStorageClassName
-	}
+
 	// Persistent Volume Requirements
 	if pvc.Spec.VolumeName != "" {
 		requirements, err := v.getPersistentVolumeRequirements(ctx, pod, pvc.Spec.VolumeName)
@@ -109,8 +93,8 @@ func (v *VolumeTopology) getRequirements(ctx context.Context, pod *v1.Pod, volum
 		return requirements, nil
 	}
 	// Storage Class Requirements
-	if storageClassName != "" {
-		requirements, err := v.getStorageClassRequirements(ctx, storageClassName)
+	if sc := lo.FromPtr(pvc.Spec.StorageClassName); sc != "" {
+		requirements, err := v.getStorageClassRequirements(ctx, sc)
 		if err != nil {
 			return nil, err
 		}
@@ -153,45 +137,33 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
 	return requirements, nil
 }
 
-func (v *VolumeTopology) getPersistentVolumeClaim(ctx context.Context, pod *v1.Pod, volume v1.Volume) (*v1.PersistentVolumeClaim, error) {
-	if volume.PersistentVolumeClaim == nil {
-		return nil, nil
-	}
-	pvc := &v1.PersistentVolumeClaim{}
-	if err := v.kubeClient.Get(ctx, types.NamespacedName{Name: volume.PersistentVolumeClaim.ClaimName, Namespace: pod.Namespace}, pvc); err != nil {
-		return nil, fmt.Errorf("getting persistent volume claim %q, %w", volume.PersistentVolumeClaim.ClaimName, err)
-	}
-	return pvc, nil
-}
-
 // ValidatePersistentVolumeClaims returns an error if the pod doesn't appear to be valid with respect to
 // PVCs (e.g. the PVC is not found or references an unknown storage class).
 func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod *v1.Pod) error {
 	for _, volume := range pod.Spec.Volumes {
-		var storageClassName *string
-		var volumeName string
-		if volume.PersistentVolumeClaim != nil {
-			// validate the PVC if it exists
-			pvc, err := v.getPersistentVolumeClaim(ctx, pod, volume)
-			if err != nil {
-				return err
-			}
-			// may not have a PVC
-			if pvc == nil {
-				continue
-			}
-
-			storageClassName = pvc.Spec.StorageClassName
-			volumeName = pvc.Spec.VolumeName
-		} else if volume.Ephemeral != nil {
-			storageClassName = volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName
+		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, v.kubeClient, pod, volume)
+		if err != nil {
+			return err
+		}
+		// Not all volume types have PVCs, e.g. emptyDir, hostPath, etc.
+		if pvc == nil {
+			continue
 		}
 
+		if pvc.Spec.VolumeName != "" {
+			if err := v.validateVolume(ctx, pvc.Spec.VolumeName); err != nil {
+				return fmt.Errorf("failed to validate pvc %q with volume %q, %w", pvc.Name, pvc.Spec.VolumeName, err)
+			}
+			continue
+		}
+
+		// PVC is unbound, we can't schedule unless the pod defines a valid storage class
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
+		if storageClassName == "" {
+			return fmt.Errorf("unbound pvc %s must define a storage class", pvc.Name)
+		}
 		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
-			return err
-		}
-		if err := v.validateVolume(ctx, volumeName); err != nil {
-			return err
+			return fmt.Errorf("failed to validate pvc %q with storage class %q, %w", pvc.Name, storageClassName, err)
 		}
 	}
 	return nil
@@ -208,13 +180,10 @@ func (v *VolumeTopology) validateVolume(ctx context.Context, volumeName string) 
 	return nil
 }
 
-func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassName *string) error {
-	// we have a storage class name, so ensure that it exists
-	if ptr.StringValue(storageClassName) != "" {
-		storageClass := &storagev1.StorageClass{}
-		if err := v.kubeClient.Get(ctx, types.NamespacedName{Name: ptr.StringValue(storageClassName)}, storageClass); err != nil {
-			return err
-		}
+func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassName string) error {
+	storageClass := &storagev1.StorageClass{}
+	if err := v.kubeClient.Get(ctx, types.NamespacedName{Name: storageClassName}, storageClass); err != nil {
+		return err
 	}
 	return nil
 }
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -21,12 +21,15 @@ import (
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	csitranslation "k8s.io/csi-translation-lib"
 	"k8s.io/csi-translation-lib/plugins"
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	volumeutil "sigs.k8s.io/karpenter/pkg/utils/volume"
 )
 
 //go:generate controller-gen object:headerFile="../../hack/boilerplate.go.txt" paths="."
@@ -77,39 +80,32 @@ func (u Volumes) Insert(volumes Volumes) {
 func GetVolumes(ctx context.Context, kubeClient client.Client, pod *v1.Pod) (Volumes, error) {
 	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", pod.Name))
 	podPVCs := Volumes{}
-	defaultStorageClassName, err := DiscoverDefaultStorageClassName(ctx, kubeClient)
-	if err != nil {
-		return nil, fmt.Errorf("discovering default storage class, %w", err)
-	}
 	for _, volume := range pod.Spec.Volumes {
 		ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("volume", volume.Name))
-		var pvcID, storageClassName, volumeName string
-		var pvc v1.PersistentVolumeClaim
-		if volume.PersistentVolumeClaim != nil {
-			if err = kubeClient.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: volume.PersistentVolumeClaim.ClaimName}, &pvc); err != nil {
-				return nil, err
+		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, pod, volume)
+		// If the PVC is not found it was manually deleted and its finalizer removed. We should ignore this volume when
+		// computing limits, otherwise Karpenter may never be able to update its cluster state.
+		if err != nil {
+			if errors.IsNotFound(err) {
+				logging.FromContext(ctx).Errorf("failed updating volume limits for volume, %w", err)
+				continue
 			}
-			pvcID = fmt.Sprintf("%s/%s", pod.Namespace, volume.PersistentVolumeClaim.ClaimName)
-			storageClassName = lo.FromPtr(pvc.Spec.StorageClassName)
-			volumeName = pvc.Spec.VolumeName
-		} else if volume.Ephemeral != nil {
-			// generated name per https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#persistentvolumeclaim-naming
-			pvcID = fmt.Sprintf("%s/%s-%s", pod.Namespace, pod.Name, volume.Name)
-			storageClassName = lo.FromPtr(volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName)
-			volumeName = volume.Ephemeral.VolumeClaimTemplate.Spec.VolumeName
-		} else {
+			return nil, fmt.Errorf("failed updating volume limits, %w", err)
+		}
+		// Not all volume types have PVCs, e.g. emptyDir, hostPath, etc.
+		if pvc == nil {
 			continue
 		}
-		if storageClassName == "" {
-			storageClassName = defaultStorageClassName
-		}
-		driverName, err := resolveDriver(ctx, kubeClient, volumeName, storageClassName)
+		ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pvc", pvc.Name))
+
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
+		driverName, err := resolveDriver(ctx, kubeClient, pvc.Spec.VolumeName, storageClassName)
 		if err != nil {
 			return nil, err
 		}
 		// might be a non-CSI driver, something we don't currently handle
 		if driverName != "" {
-			podPVCs.Add(driverName, pvcID)
+			podPVCs.Add(driverName, client.ObjectKeyFromObject(pvc).String())
 		}
 	}
 	return podPVCs, nil
@@ -130,18 +126,30 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
 			return driverName, nil
 		}
 	}
-	if storageClassName != "" {
-		driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
-		if err != nil {
-			return "", err
-		}
-		if driverName != "" {
-			return driverName, nil
-		}
+
+	// This can occur in two scenarios:
+	//  1. The storage class was explicitly set to "" to disable dynamic provisioning
+	//  2. The storage class was not set but the cluster doesn't have a default storage class
+	// In either of these cases, a PV must have been previously bound to the PVC and has since been removed. We can
+	// ignore this PVC while computing limits and continue.
+	if storageClassName == "" {
+		logging.FromContext(ctx).Errorf("failed updating volume limits for volume with unbound PVC, no storage class specified")
+		return "", nil
 	}
-	// Driver name wasn't able to resolve for this volume. In this case, we just ignore the
-	// volume and move on to the other volumes that the pod has
-	return "", nil
+
+	driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
+	if err != nil {
+		// There are two scenarios where a StorageClass may be defined but not found:
+		//  1. The StorageClass was manually deleted and the finalizer removed
+		//  2. The StorageClass never existed and was used to bind the PVC to an existing PV, but that PV was removed
+		// In either of these cases, we should ignore the PVC while computing limits and continue.
+		if errors.IsNotFound(err) {
+			logging.FromContext(ctx).With("storageclass", storageClassName).Errorf("failed updating volume limits for volume with unbound PVC, %w", err)
+			return "", nil
+		}
+		return "", err
+	}
+	return driverName, nil
 }
 
 // driverFromSC resolves the storage driver name by getting the Provisioner name from the StorageClass
--- a/pkg/controllers/provisioning/scheduling/suite_test.go
+++ b/pkg/controllers/provisioning/scheduling/suite_test.go
@@ -2829,8 +2829,9 @@ var _ = Context("NodePool", func() {
 
 			initialPod := test.UnschedulablePod(test.PodOptions{})
 			// Pod has an ephemeral volume claim that has a specified storage class, so it should use the one specified
+			volumeName := "tmp-ephemeral"
 			initialPod.Spec.Volumes = append(initialPod.Spec.Volumes, v1.Volume{
-				Name: "tmp-ephemeral",
+				Name: volumeName,
 				VolumeSource: v1.VolumeSource{
 					Ephemeral: &v1.EphemeralVolumeSource{
 						VolumeClaimTemplate: &v1.PersistentVolumeClaimTemplate{
@@ -2849,7 +2850,14 @@ var _ = Context("NodePool", func() {
 					},
 				},
 			})
-			ExpectApplied(ctx, env.Client, nodePool, sc, sc2, initialPod)
+			pvc := test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: initialPod.Namespace,
+					Name:      fmt.Sprintf("%s-%s", initialPod.Name, volumeName),
+				},
+				StorageClassName: lo.ToPtr(sc.Name),
+			})
+			ExpectApplied(ctx, env.Client, nodePool, sc, sc2, pvc, initialPod)
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, initialPod)
 			node := ExpectScheduled(ctx, env.Client, initialPod)
 			csiNode := &storagev1.CSINode{
@@ -2881,7 +2889,7 @@ var _ = Context("NodePool", func() {
 			pod := test.UnschedulablePod(test.PodOptions{})
 			// Pod has an ephemeral volume claim that has a specified storage class, so it should use the one specified
 			pod.Spec.Volumes = append(pod.Spec.Volumes, v1.Volume{
-				Name: "tmp-ephemeral",
+				Name: volumeName,
 				VolumeSource: v1.VolumeSource{
 					Ephemeral: &v1.EphemeralVolumeSource{
 						VolumeClaimTemplate: &v1.PersistentVolumeClaimTemplate{
@@ -2900,7 +2908,14 @@ var _ = Context("NodePool", func() {
 					},
 				},
 			})
-			ExpectApplied(ctx, env.Client, nodePool, pod)
+			pvc = test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: pod.Namespace,
+					Name:      fmt.Sprintf("%s-%s", pod.Name, volumeName),
+				},
+				StorageClassName: lo.ToPtr(sc.Name),
+			})
+			ExpectApplied(ctx, env.Client, nodePool, pvc, pod)
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
 			node2 := ExpectScheduled(ctx, env.Client, pod)
 			Expect(node.Name).ToNot(Equal(node2.Name))
@@ -2919,8 +2934,9 @@ var _ = Context("NodePool", func() {
 
 			initialPod := test.UnschedulablePod(test.PodOptions{})
 			// Pod has an ephemeral volume claim that has NO storage class, so it should use the default one
+			volumeName := "tmp-ephemeral"
 			initialPod.Spec.Volumes = append(initialPod.Spec.Volumes, v1.Volume{
-				Name: "tmp-ephemeral",
+				Name: volumeName,
 				VolumeSource: v1.VolumeSource{
 					Ephemeral: &v1.EphemeralVolumeSource{
 						VolumeClaimTemplate: &v1.PersistentVolumeClaimTemplate{
@@ -2938,7 +2954,13 @@ var _ = Context("NodePool", func() {
 					},
 				},
 			})
-			ExpectApplied(ctx, env.Client, nodePool, sc, initialPod)
+			pvc := test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: initialPod.Namespace,
+					Name:      fmt.Sprintf("%s-%s", initialPod.Name, volumeName),
+				},
+			})
+			ExpectApplied(ctx, env.Client, nodePool, sc, initialPod, pvc)
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, initialPod)
 			node := ExpectScheduled(ctx, env.Client, initialPod)
 			csiNode := &storagev1.CSINode{
@@ -2963,7 +2985,7 @@ var _ = Context("NodePool", func() {
 			pod := test.UnschedulablePod(test.PodOptions{})
 			// Pod has an ephemeral volume claim that has NO storage class, so it should use the default one
 			pod.Spec.Volumes = append(pod.Spec.Volumes, v1.Volume{
-				Name: "tmp-ephemeral",
+				Name: volumeName,
 				VolumeSource: v1.VolumeSource{
 					Ephemeral: &v1.EphemeralVolumeSource{
 						VolumeClaimTemplate: &v1.PersistentVolumeClaimTemplate{
@@ -2981,8 +3003,14 @@ var _ = Context("NodePool", func() {
 					},
 				},
 			})
+			pvc = test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: pod.Namespace,
+					Name:      fmt.Sprintf("%s-%s", pod.Name, volumeName),
+				},
+			})
 
-			ExpectApplied(ctx, env.Client, sc, nodePool, pod)
+			ExpectApplied(ctx, env.Client, sc, nodePool, pod, pvc)
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
 			node2 := ExpectScheduled(ctx, env.Client, pod)
 			Expect(node.Name).ToNot(Equal(node2.Name))
@@ -3016,10 +3044,11 @@ var _ = Context("NodePool", func() {
 			time.Sleep(time.Second * 2)
 			ExpectApplied(ctx, env.Client, sc2)
 
+			volumeName := "tmp-ephemeral"
 			initialPod := test.UnschedulablePod(test.PodOptions{})
 			// Pod has an ephemeral volume claim that has NO storage class, so it should use the default one
 			initialPod.Spec.Volumes = append(initialPod.Spec.Volumes, v1.Volume{
-				Name: "tmp-ephemeral",
+				Name: volumeName,
 				VolumeSource: v1.VolumeSource{
 					Ephemeral: &v1.EphemeralVolumeSource{
 						VolumeClaimTemplate: &v1.PersistentVolumeClaimTemplate{
@@ -3037,7 +3066,13 @@ var _ = Context("NodePool", func() {
 					},
 				},
 			})
-			ExpectApplied(ctx, env.Client, nodePool, sc, initialPod)
+			pvc := test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: initialPod.Namespace,
+					Name:      fmt.Sprintf("%s-%s", initialPod.Name, volumeName),
+				},
+			})
+			ExpectApplied(ctx, env.Client, nodePool, sc, initialPod, pvc)
 			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, initialPod)
 			node := ExpectScheduled(ctx, env.Client, initialPod)
 			csiNode := &storagev1.CSINode{
@@ -3069,7 +3104,7 @@ var _ = Context("NodePool", func() {
 			pod := test.UnschedulablePod(test.PodOptions{})
 			// Pod has an ephemeral volume claim that has NO storage class, so it should use the default one
 			pod.Spec.Volumes = append(pod.Spec.Volumes, v1.Volume{
-				Name: "tmp-ephemeral",
+				Name: volumeName,
 				VolumeSource: v1.VolumeSource{
 					Ephemeral: &v1.EphemeralVolumeSource{
 						VolumeClaimTemplate: &v1.PersistentVolumeClaimTemplate{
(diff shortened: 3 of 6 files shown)


pkg/scheduling/volumeusage.go
```diff
@@ -132,7 +131,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
 	}
 	if storageClassName != "" {
 		driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
-		if err != nil {
+		if err != nil && !errors.IsNotFound(err) {
 			return "", err
```
jonathan-innis, line 135: I'm assuming that we should probably log if we don't find the storageClassName but the storage class name is specified. Also, I'd think that we should exit out of this function as soon as we get the `NotFound` error but we should just return a `nil` error when we are getting this data as part of state tracking

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -181,15 +182,32 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
 				continue
 			}
 
-			storageClassName = pvc.Spec.StorageClassName
+			storageClassName = lo.FromPtr(pvc.Spec.StorageClassName)
```
jonathan-innis, line 185: Is there ever a case where this should be empty? Why do we need to convert it from a pointer?

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -181,15 +182,32 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
...
 			volumeName = pvc.Spec.VolumeName
 		} else if volume.Ephemeral != nil {
```
jonathan-innis, line 187: Is it possible to just rely on the PVC directly always? The volumeClaimTemplate generates a PVC which has a known name. We should be able to look-up that known name and then use that to check the storage class name and the volume name. I believe in that case, the storage class name should never not be set and we can handle the cases which are:
1.  The storage class name is empty and the volume name needs to exist
2. The storage class name is not empty but the volume name exists (so this overrides the storage class in this case)
3. The storage class name is not empty AND the volume name is empty (so we should check if the storage class exists)

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: There's a couple of cases:
1) There was no storage class specified and the default hasn't been applied yet by the persistent volume controller
2) There was no storage class specified and there is no default storage class

pkg/scheduling/volumeusage.go
jmdeal: So the reason I'm not logging it is because there is a valid case for the storage class not existing but being specified. If you create a PV that specifies the same StorageClass as a PVC and all of the other requirements have been met, they can bind even if the StorageClass doesn't exist. I haven't been able to find solid documentation on **why** this is the case (it seems pretty counter-intuitive to me) but it is demonstrated in the kubernetes docs ([ref](https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/)) and I've been able to verify that in my own cluster.

jmdeal: I think the reason I didn't return `nil` immediately in this case is because I didn't think it followed the return style I had seen elsewhere in the project so I was trying to be consistent. Probably easier to talk offline about what I mean by this, but I'll give it a brief shot. I've seen quite a few examples of this:
```go
if err := someLogic(); err != nil {
  return err
}
return nil
```
rather than this:
```go
return someLogic()
```
I interpreted this as a way to make it simple to add more cases in the future, which I think waiting to return here does. I'm completely open to changing this to return earlier though.

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: We should be able to, thats what I was calling out in the TODO that I was planning to do in a follow-up PR. It does make sense to bring in scope here though,  I don't think I did because I tied it to removing default storage class discovery.

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: This may not be the case, I found out there's a [default storage class admission controller](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#defaultstorageclass) which will populate the default. I'd have to take a closer look to see what its behavior is when there is no default storage class but this eliminates the first case at least. Not sure if we can rely on the presence of that admission controller but its compiled into the API server so it seems like a safe bet?

jmdeal: Seems like the second case is still possible, the admission controller won't update the PVC if there is no default storage class ([ref](https://github.com/kubernetes/kubernetes/blob/8509ab82b96caa2365552efa08c8ba8baf11c5ec/plugin/pkg/admission/storage/storageclass/setdefault/admission.go#L112)). Makes sense, the persistent volume controller will also add the default storage class if one exists so that probably covers the case of a default being created after the PVC.

pkg/scheduling/volumeusage.go
```diff
@@ -77,38 +80,25 @@ func (u Volumes) Insert(volumes Volumes) {
...
-		} else if volume.Ephemeral != nil {
-			// generated name per https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#persistentvolumeclaim-naming
-			pvcID = fmt.Sprintf("%s/%s-%s", pod.Namespace, pod.Name, volume.Name)
-			storageClassName = lo.FromPtr(volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName)
-			volumeName = volume.Ephemeral.VolumeClaimTemplate.Spec.VolumeName
-		} else {
-			continue
+		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, pod, volume)
+		if err != nil {
+			return nil, err
```
jonathan-innis, line 87: If we fail to find the PVC but the node is bound, is that enough to cause us to continue to retry? Is there a case here where we just never sync the cluster if there is a pod that is bound to a node that is referencing a PVC that no longer exists for some reason?

pkg/scheduling/volumeusage.go
```diff
@@ -132,7 +122,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
 	}
```
jonathan-innis, line 122: Should we fail to sync nodes if we fail to find the PV that is referenced by the PVC? Is it possible for a PVC to reference a PV that doesn't exist and the pod still to be bound?

pkg/scheduling/volumeusage.go
jonathan-innis: > If you create a PV that specifies the same StorageClass as a PVC and all of the other requirements have been met, they can bind even if the StorageClass doesn't exist

Let's discuss this one a bit more. I'm trying to understand what this looks like exactly and why someone would want to do this.

pkg/scheduling/volumeusage.go
```diff
@@ -132,7 +122,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
 	if storageClassName != "" {
 		driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
-		if err != nil {
+		if err != nil && !errors.IsNotFound(err) {
```
jonathan-innis, line 125: We should think about how to solve the `NotFound` issue wholistically for all of the storage resolution logic that happens in `state.Cluster`. As it stands right now: if there is ever an error that is persistent and continues to happen (like a NotFound error) that is going to hang the cluster state while it's waiting for that node to get propagated.

pkg/scheduling/volumeusage.go
```diff
@@ -77,38 +80,25 @@ func (u Volumes) Insert(volumes Volumes) {
...
 		}
+		volumeName := pvc.Spec.VolumeName
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
 		driverName, err := resolveDriver(ctx, kubeClient, volumeName, storageClassName)
 		if err != nil {
 			return nil, err
 		}
 		// might be a non-CSI driver, something we don't currently handle
 		if driverName != "" {
+			pvcID := fmt.Sprintf("%s/%s", pod.Namespace, pvc.ObjectMeta.Name)
```
jonathan-innis, line 101:
```suggestion
			pvcID := fmt.Sprintf("%s/%s", pod.Namespace, pvc.Name)
```

No need to specify the `ObjectMeta` piece. Easier to just do `pvc.Name`

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -76,30 +76,18 @@ func (v *VolumeTopology) Inject(ctx context.Context, pod *v1.Pod) error {
...
-			return nil, fmt.Errorf("discovering persistent volume claim for ephemeral volume, %w", err)
-		}
-	default:
+	// may not have a pvc
+	if pvc == nil {
 		return nil, nil
 	}
+
+	// Get VolumeName and StorageClass name from PVC
 	storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
```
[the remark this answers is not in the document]
jmdeal: Probably not, no. I can go ahead and tear that out 👍

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,47 +141,39 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
 
-		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
+		err = v.validateStorageClass(ctx, storageClassName)
+		if err != nil && !errors.IsNotFound(err) {
 			return err
 		}
+		// If the StorageClass doesn't exist the pod will still be able to schedule if its PVC is already bound.
+		// Example: https://kubernetes.io/docs/tasks/configure-pod-container/configure-persistent-volume-storage/
+		if missingStorageClass := err != nil || storageClassName == ""; missingStorageClass && volumeName == "" {
```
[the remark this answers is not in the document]
jmdeal: That seems reasonable to me, I don't think it would change any behavior but would clean up some of the logic. Happy to discuss this more.

pkg/scheduling/volumeusage.go
jmdeal: The only case I'm aware of off the top of my head is if a user removes a finalizer and deletes the PVC. IIRC (I'll have to verify) the pod will continue running without a PVC in that case. At that point though maybe its acceptable for us to lock up here? What is the expected behavior for Kubernetes controllers when a resource they're managing has its finalizer removed? Are they expected to handle it gracefully?

pkg/scheduling/volumeusage.go
jmdeal: Resolving, discussed and aligned offline. The referenced Kubernetes doc above in the thread is the best example I could find of a use case.

pkg/controllers/provisioning/scheduling/volumetopology.go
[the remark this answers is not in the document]
jmdeal: Discussed offline, exiting early if the PVC is already bound cleans up logic significantly.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,47 +141,39 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
+			if err != nil {
+				retErr = fmt.Errorf("%w, %w", retErr, err)
+			}
+			return retErr
+		}
 		if err := v.validateVolume(ctx, volumeName); err != nil {
 			return err
 		}
 	}
+	logging.FromContext(ctx).Debugf("validated pvcs for pod %q", pod.Name)
```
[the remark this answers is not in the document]
jmdeal: Good catch, I missed removing that. I had it for debugging.

pkg/scheduling/volumeusage.go
jmdeal: Conclusion here was to ignore pods with missing PVCs when updating volume limits to ensure Karpenter does not lock up in that scenario.

pkg/scheduling/volumeusage.go
jmdeal: Yes, if the PV is dynamically provisioned via a storage class the PV is not provisioned until the pod is bound.

pkg/scheduling/volumeusage.go
jmdeal: We resolved this offline, the current approach will be to ignore volume limits for a pod when we encounter `NotFound` errors. This ensures that we can update cluster state and progress.

pkg/controllers/provisioning/scheduling/volumetopology.go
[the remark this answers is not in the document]
jmdeal: Actually I'm going to backpedal here, the storage class could still be `nil` if there was no default storage class when the PVC was created. If a default exists when the PVC is created it will be added via an admission controller, otherwise it can be reconciled on later.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,44 +137,31 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
-			volumeName = pvc.Spec.VolumeName
-		} else if volume.Ephemeral != nil {
-			storageClassName = volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName
+			continue
 		}
 
-		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
-			return err
+		// PVC is unbound, we can't schedule unless the pod defines a valid storage class
+		if lo.FromPtr(pvc.Spec.StorageClassName) == "" {
```
jonathan-innis, line 161: nit: store the storage class name in a variable and convert from a pointer one time?

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,44 +137,31 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
-		var volumeName string
-		if volume.PersistentVolumeClaim != nil {
-			// validate the PVC if it exists
-			pvc, err := v.getPersistentVolumeClaim(ctx, pod, volume)
-			if err != nil {
+		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, v.kubeClient, pod, volume)
+		if err != nil {
+			return err
+		}
+		// may not have a pvc
```
jonathan-innis, line 148: Can we callout when it's possible that the volume doesn't have a PVC. IIUC, it's a case of when it's using an in-tree provider that isn't using CSI. In this case, we are just ignoring the volume usage since we have no ability to discover it

pkg/scheduling/volumeusage.go
```diff
@@ -77,39 +80,29 @@ func (u Volumes) Insert(volumes Volumes) {
...
-			storageClassName = lo.FromPtr(pvc.Spec.StorageClassName)
-			volumeName = pvc.Spec.VolumeName
-		} else if volume.Ephemeral != nil {
-			// generated name per https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#persistentvolumeclaim-naming
-			pvcID = fmt.Sprintf("%s/%s-%s", pod.Namespace, pod.Name, volume.Name)
-			storageClassName = lo.FromPtr(volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName)
-			volumeName = volume.Ephemeral.VolumeClaimTemplate.Spec.VolumeName
-		} else {
+		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, pod, volume)
+		if errors.IsNotFound(err) {
```
jonathan-innis, line 86: Why do we chose to ignore the error here? When we have handling logic like this we should be really explicit about why we are choosing to swallow the error.

jonathan-innis, line 86: nit: Typically the syntax for error checking looks like this since it's slightly more readable

```go
if err != nil {
  if errors.IsNotFound(err) {
     // do something
  }
  // handle the error if not a NotFound 
}
```

pkg/scheduling/volumeusage.go
```diff
@@ -130,22 +123,24 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
-		if err != nil {
-			return "", err
-		}
-		if driverName != "" {
-			return driverName, nil
-		}
+
+	driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
+	if storageClassName == "" || errors.IsNotFound(err) {
+		logging.FromContext(ctx).Warnf("failed to update volume limits for volume with unbound PVC, storage class not found \"%s\"", storageClassName)
```
jonathan-innis, line 129: We should avoid warnings at all costs. In general, it's either an error or an info/debug. If we have a storage class that doesn't exist and doesn't have a bound volume, this should be an error

pkg/scheduling/volumeusage.go
```diff
@@ -130,22 +123,24 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
-		driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
-		if err != nil {
-			return "", err
-		}
-		if driverName != "" {
-			return driverName, nil
-		}
+
+	driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
+	if storageClassName == "" || errors.IsNotFound(err) {
```
jonathan-innis, line 128: This feels like a bunch of extra checking for empty strings and sees odd that we are doing a check both for the empty string AND also that the error is a NotFound error. Seems like there's a way that we could clean this error handling logic up a bit

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: That's what the comment is calling out on 148, I can be more verbose about it.

pkg/scheduling/volumeusage.go
jmdeal: I believe this was the case we discussed offline where for this to occur a finalizer would have to be removed from the PVC. In that case it won't naturally recover but the node will stay scheduled. This ensures we are able to make progress.  I can add a comment about that here since it's not clear.

pkg/scheduling/volumeusage.go
jmdeal: Got it, I figured since we could still progress a warning may have been more appropriate but I'm happy to change it to an error.

pkg/scheduling/volumeusage.go
jmdeal: Agreed, this could be cleaned up by having `driverFromSC` return a not found error if the storage class is empty.

jmdeal: On second thought, its probably best to handle this as a separate error.

pkg/controllers/provisioning/scheduling/volumetopology.go
jonathan-innis: Sure, I got that, but I still don't know the why around when something doesn't have a PVC

jmdeal: Ah I get what you mean, idk what I was thinking when I initially responded. That scenario sounds correct, I can lookup the full list of volume types which don't need a PVC.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -208,11 +179,10 @@ func (v *VolumeTopology) validateVolume(ctx context.Context, volumeName string)
 	return nil
 }
 
-func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassName *string) error {
-	// we have a storage class name, so ensure that it exists
-	if ptr.StringValue(storageClassName) != "" {
+func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassName string) error {
+	if storageClassName != "" {
```
jonathan-innis, line 183: When will this ever be an empty string now that we are checking that the storage class name can't be empty?

pkg/scheduling/volumeusage.go
```diff
@@ -77,39 +81,32 @@ func (u Volumes) Insert(volumes Volumes) {
 func GetVolumes(ctx context.Context, kubeClient client.Client, pod *v1.Pod) (Volumes, error) {
 	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", pod.Name))
 	podPVCs := Volumes{}
-	defaultStorageClassName, err := DiscoverDefaultStorageClassName(ctx, kubeClient)
-	if err != nil {
-		return nil, fmt.Errorf("discovering default storage class, %w", err)
-	}
 	for _, volume := range pod.Spec.Volumes {
 		ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("volume", volume.Name))
```
jonathan-innis, line 85: Thinking about the logging that we are doing here, does it make sense to inject the PVC name that we are looking up into the logger?

pkg/scheduling/volumeusage.go
```diff
@@ -130,22 +127,32 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
-	if storageClassName != "" {
-		driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
-		if err != nil {
-			return "", err
-		}
-		if driverName != "" {
-			return driverName, nil
+
+	// This state shouldn't be possible during normal operations but if a PV is deleted and its finalizer is removed it is possible.
+	if storageClassName == "" {
```
jonathan-innis, line 132: Is this state possible if we fail to resolve the driver to a non-empty string on L122 and then fall down into this conditional statement?

pkg/scheduling/volumeusage.go
```diff
@@ -130,22 +127,32 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
+		logging.FromContext(ctx).Errorf("failed to update volume limits for volume with unbound PVC, no storage class specified")
+		return "", nil
+	}
+	driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
+	if err != nil {
+		// There are two scenarios where a StorageClass may be defined but not found:
+		//  1. The StorageClass was manually deleted and the finalizer removed
+		//  2. The StorageClass never existed and was used to bind the PVC to an existing PV, but that PV was removed
+		if errors.IsNotFound(err) {
+			logging.FromContext(ctx).Errorf("failed to update volume limits for volume with unbound PVC, %w", storageClassName)
```
jonathan-innis, line 142:
```suggestion
			logging.FromContext(ctx).With("storageclass", storageClassName).Errorf("failed to update volume limits for volume with unbound PVC, %w", err)
```

pkg/scheduling/volumeusage.go
```diff
@@ -130,22 +127,32 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
 	}
-	// Driver name wasn't able to resolve for this volume. In this case, we just ignore the
-	// volume and move on to the other volumes that the pod has
-	return "", nil
+	return driverName, nil
 }
 
 // driverFromSC resolves the storage driver name by getting the Provisioner name from the StorageClass
 func driverFromSC(ctx context.Context, kubeClient client.Client, storageClassName string) (string, error) {
+	if storageClassName == "" {
```
jonathan-innis, line 152: Given our current codepaths, is this ever actually possible?

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: Good point, this check can be removed 👍

pkg/scheduling/volumeusage.go
jmdeal: Good point, we could potentially remove the empty string check on line 126. I'll need to find out if there's a case where we can't resolve the driver name from a bound volume but we can from it's storage class.

pkg/scheduling/volumeusage.go
jmdeal: I don't believe it is anymore, I'll verify and remove it 👍

pkg/scheduling/volumeusage.go
jmdeal: Sounds reasonable to me 👍

pkg/scheduling/volumeusage.go
jmdeal: I'm actually thinking it may be best to leave this as is, I'm not familiar with the in-tree plugins / storage classes so I don't want to miss an edge case with those so I've just changed the comment. There are a couple of more things I need to validate before I think this is completely ready to merge, if I determine its totally safe to remove I can make the update.

pkg/scheduling/volumeusage.go
```diff
@@ -126,22 +122,34 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
 		if err != nil {
 			return "", err
 		}
-		if driverName != "" {
+		if driverName == "" {
```
jonathan-innis, line 125: This seems incorrect. Unless I'm misunderstanding something

pkg/scheduling/volumeusage.go
```diff
@@ -126,22 +122,34 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
-		if driverName != "" {
-			return driverName, nil
+
+	// This can occur in two scenarios:
+	//  1. The storage class was explicitly set to "" to disable dynamic provisioning
+	//  2. The storage class was not set but the cluster doesn't have a default storage class
+	// In either of these cases, a PV must have been previously bound to the PVC and has since been removed. We can
+	// ignore this PVC while computing limits and continue.
+	if storageClassName == "" {
+		logging.FromContext(ctx).Errorf("failed to update volume limits for volume with unbound PVC, no storage class specified")
```
jonathan-innis, line 136:
```suggestion
		logging.FromContext(ctx).Errorf("failed updating volume limits for volume with unbound PVC, no storage class specified")
```

nit: style is to use the gerund

pkg/scheduling/volumeusage.go
```diff
@@ -126,22 +122,34 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
...
+	}
+
+	driverName, err := driverFromSC(ctx, kubeClient, storageClassName)
+	if err != nil {
+		// There are two scenarios where a StorageClass may be defined but not found:
+		//  1. The StorageClass was manually deleted and the finalizer removed
+		//  2. The StorageClass never existed and was used to bind the PVC to an existing PV, but that PV was removed
+		// In either of these cases, we should ignore the PVC while computing limits and continue.
+		if errors.IsNotFound(err) {
+			logging.FromContext(ctx).With("storageclass", storageClassName).Errorf("failed to update volume limits for volume with unbound PVC, %w", err)
```
jonathan-innis, line 147:
```suggestion
			logging.FromContext(ctx).With("storageclass", storageClassName).Errorf("failed updating volume limits for volume with unbound PVC, %w", err)
```

nit: error style

pkg/scheduling/volumeusage.go
```diff
@@ -77,39 +80,32 @@ func (u Volumes) Insert(volumes Volumes) {
...
-		var pvc v1.PersistentVolumeClaim
-		if volume.PersistentVolumeClaim != nil {
-			if err = kubeClient.Get(ctx, client.ObjectKey{Namespace: pod.Namespace, Name: volume.PersistentVolumeClaim.ClaimName}, &pvc); err != nil {
-				return nil, err
+		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, pod, volume)
+		// If the PVC is not found it was manually deleted and its finalizer removed. We should ignore this volume when
+		// computing limits, otherwise Karpenter may never be able to update its cluster state.
+		if err != nil {
+			if errors.IsNotFound(err) {
+				logging.FromContext(ctx).Errorf("failed to update volume limits for volume, %w", err)
```
jonathan-innis, line 90:
```suggestion
				logging.FromContext(ctx).Errorf("failed updating update volume limits for volume, %w", err)
```

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,44 +137,32 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
-			storageClassName = volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName
+			continue
 		}
 
-		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
-			return err
+		// PVC is unbound, we can't schedule unless the pod defines a valid storage class
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
+		if storageClassName == "" {
+			return fmt.Errorf("unbound PVC must define a storage class")
```
jonathan-innis, line 163: Should we give any detail on which PVC that this is?

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,44 +137,32 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
+			return err
+		}
+		// Not all volume types have PVCs, e.g. emptyDir, hostPath, etc.
+		if pvc == nil {
+			continue
+		}
+
+		if pvc.Spec.VolumeName != "" {
+			if err := v.validateVolume(ctx, pvc.Spec.VolumeName); err != nil {
 				return err
```
jonathan-innis, line 155: Wrap the error to give detail on which PVC and PV we are looking at?

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,44 +137,32 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
-		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
-			return err
+		// PVC is unbound, we can't schedule unless the pod defines a valid storage class
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
+		if storageClassName == "" {
+			return fmt.Errorf("unbound PVC must define a storage class")
 		}
-		if err := v.validateVolume(ctx, volumeName); err != nil {
+		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
 			return err
```
jonathan-innis, line 166: Wrap the error to give detail on which PVC and SC we are looking at?

pkg/scheduling/volumeusage.go
jmdeal: Definitely a typo...fixed now

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,45 +137,33 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
+			continue
+		}
 
-			storageClassName = pvc.Spec.StorageClassName
-			volumeName = pvc.Spec.VolumeName
-		} else if volume.Ephemeral != nil {
-			storageClassName = volume.Ephemeral.VolumeClaimTemplate.Spec.StorageClassName
+		if pvc.Spec.VolumeName != "" {
+			if err := v.validateVolume(ctx, pvc.Spec.VolumeName); err != nil {
+				return fmt.Errorf("failed to validate pvc %s with volume %s, %w", pvc.Name, pvc.Spec.VolumeName, err)
```
jonathan-innis, line 155:
```suggestion
				return fmt.Errorf("failed to validate pvc %q with volume %q, %w", pvc.Name, pvc.Spec.VolumeName, err)
```

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -153,45 +137,33 @@ func (v *VolumeTopology) getPersistentVolumeRequirements(ctx context.Context, po
...
-			return err
+		// PVC is unbound, we can't schedule unless the pod defines a valid storage class
+		storageClassName := lo.FromPtr(pvc.Spec.StorageClassName)
+		if storageClassName == "" {
+			return fmt.Errorf("unbound pvc %s must define a storage class", pvc.Name)
 		}
-		if err := v.validateVolume(ctx, volumeName); err != nil {
-			return err
+		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
+			return fmt.Errorf("failed to validate pvc %s with storage class %s, %w", pvc.Name, storageClassName, err)
```
jonathan-innis, line 166:
```suggestion
			return fmt.Errorf("failed to validate pvc %q with storage class %q, %w", pvc.Name, storageClassName, err)
```

jonathan-innis:
We have the E2E tests in place and will bump the Karpenter version that contains this fix before we merge the E2E testing in the AWS cloud provider repo: https://github.com/aws/karpenter-provider-aws/pull/5229

PR #786, opened 2023-11-15
chore: Add NodePool readiness defined by the CloudProvider

jonathan-innis:
Fixes #N/A 

**Description**

This PR adds the `IsReady()` interface function to the CloudProvider. The plan is to eventually move this interface function into the status conditions with #493; however, we can keep this concept internal for now by calling out to the CloudProvider as a first iteration.

`make presubmit`

### DIFF
--- a/pkg/cloudprovider/fake/cloudprovider.go
+++ b/pkg/cloudprovider/fake/cloudprovider.go
@@ -51,6 +51,7 @@ type CloudProvider struct {
 
 	CreatedNodeClaims map[string]*v1beta1.NodeClaim
 	Drifted           cloudprovider.DriftReason
+	Ready             error
 }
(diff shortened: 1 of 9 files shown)


ellistarn:
Why not just start with the condition?

jonathan-innis:
> Why not just start with the condition

I'd like to come to quicker resolution on the eventual consistency issues that we currently have with the EC2NodeClass. I don't think that we should rush out thinking about how exactly we want the status conditions to be defined; however, I think an internal contract as a starting point that we can build-on makes a lot of sense for handling these eventual consistency errors in the short-term.

jonathan-innis:
Closing this for now since we're going to take the path for just outright implementing #493

PR #789, opened 2023-11-15
fix: Create Generic Pod Filtering in Utils and Collapse Disruption Filtering

### LINKED ISSUE #643: Use the same helper functions for commonly used kubeClient List/Update/Delete calls.
### Description

We use `kubeClient` calls throughout the codebase, but we could align on all kubeClient.List() calls for all pods bound to a node, and pass in filters for options of which pods to return, such as filtering out DS when we don't care about them.

### LINKED ISSUE #863: Respect do-not-disrupt annotation for static pods
### Description

Presently, the `karpenter.sh/do-not-disrupt: "true"` annotation will prevent consolidation when present on K8s resources, with the exception of terminating pods, terminal pods, DaemonSet pods and static pods.

For the use case of Airflow on Kubernetes, jobs are spawned as static pods and consolidation causes them to receive sigterm from being evicted. In this use case, it is desirable to exempt these pods from consolidation-initiated eviction.

To that end, this feature request is to implement support for the annotation `karpenter.sh/do-not-disrupt: "true"` on static pods.

Workarounds exist (disabling consolidation, scheduling to a nodegroup without consolidation), however given the transient nature of Airflow workloads, it is desirable to have it execute alongside other workloads, and making full use of consolidation is an enormous boon to efficiency.

jonathan-innis:
Fixes #643
Fixes #863

**Description**

This PR creates the following node helper functions to grab the pods on the cluster or attached to nodes:
1. `IsActive()` - This filtering function is used for PDBs and `karpenter.sh/do-not-disrupt` checking to evaluate if a node can be determined based on the active pods that are running on it
2. `IsReschedulable()` - This filtering function is used by the provisioner to determine the pods that we should attempt to reschedule based on the pods that are currently scheduled to the existing node
3. `IsEvictable()` - This filtering function is used by the node terminator and for PDB evaluation to determine the pods that can be evicted based on whether the pod is a mirror pod, is terminating, is already completed, or tolerates the disruption taint such that it would reschedule after being evicted
4. `IsWaitingEviction()` - This filtering function is used by the node terminator to determine the pods that are waiting to be evicted from the node based on whether the pod is a mirror pod, is past its termination grace period, is already completed, or tolerates the disruption taint such that it would reschedule after being evicted
5. `IsProvisionable()` - This filtering function is used to filter all pods on the cluster to determine which pods are marked in a `Pending` state and need to be scheduled by the provsioner

This PR resolves the following issues that were present in the karpenter code:
1. Karpenter was evaluating the reschedulable pods for `karpenter.sh/do-not-disrupt`. This meant that any DaemonSet pods or Mirror pods were not considered for `karpenter.sh/do-not-disrupt`. Though this isn't a common scenario for using this annotation on pods, there were asks in #863 for support for mirror pods, which would have already been the case had we been getting all of the active pods that were bound to a candidate node
2. Karpenter was evaluating the reschedulable pods for PDBs. This mean that any DaemonSet pods or mirror pods were not considered for PDBs. This isn't a problem for mirror pods, since mirror pods aren't evaluated for eviction, which means that we won't check for PDB evaluation when we are draining the node and going through the node termination; however, this is a problem for DaemonSet pods. You can create a PDB against DaemonSets. Though this isn't a common use-case at all for using PDBs, we should strive to be correct here. Since we evaluate PDBs when we drain, we should also do the same check when we are evaluating for disruption. This PR changes the logic to consider all pods (including DaemonSet pods) as long as we will run the same pods through the eviction API when we terminate (this means that mirror pods or any pod that tolerates `karpenter.sh/disruption=disrupting` shouldn't be evaluated here). 
3. Adds common pods functions for evaluating each of the conditions above. This should help anyone who is looking at changing the relevant areas of this code to better-understand what filtering is done pods that are bound to nodes and why that filtering is done.

`make presubmit`
Manual validation of consistency between the eviction API and Karpenter

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -1016,9 +1016,6 @@ var _ = Describe("Consolidation", func() {
 			// Evict the pods off of the node
 			for _, p := range pods {
 				// Trigger an eviction to set the deletion timestamp but not delete the pod
-				Expect(env.KubernetesInterface.PolicyV1().Evictions(p.Namespace).Evict(ctx, &policyv1.Eviction{
-					ObjectMeta: metav1.ObjectMeta{Name: p.Name, Namespace: p.Namespace},
-				})).To(Succeed())
 				ExpectEvicted(ctx, env.Client, p)
 				ExpectExists(ctx, env.Client, p)
 			}
@@ -1080,9 +1077,6 @@ var _ = Describe("Consolidation", func() {
 			ExpectManualBinding(ctx, env.Client, pod, node)
 
 			// Trigger an eviction to set the deletion timestamp but not delete the pod
-			Expect(env.KubernetesInterface.PolicyV1().Evictions(pod.Namespace).Evict(ctx, &policyv1.Eviction{
-				ObjectMeta: metav1.ObjectMeta{Name: pod.Name, Namespace: pod.Namespace},
-			})).To(Succeed())
 			ExpectEvicted(ctx, env.Client, pod)
 			ExpectExists(ctx, env.Client, pod)
 
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -919,6 +919,194 @@ var _ = Describe("Consolidation", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodeClaim)
 		})
+		It("will consider a node with a DaemonSet pod as empty", func() {
+			// assign the nodeclaims to the least expensive offering so we don't get a replacement
+			nodeClaim.Labels = lo.Assign(nodeClaim.Labels, map[string]string{
+				v1.LabelInstanceTypeStable:   leastExpensiveInstance.Name,
+				v1beta1.CapacityTypeLabelKey: leastExpensiveOffering.CapacityType,
+				v1.LabelTopologyZone:         leastExpensiveOffering.Zone,
+			})
+			node.Labels = lo.Assign(node.Labels, map[string]string{
+				v1.LabelInstanceTypeStable:   leastExpensiveInstance.Name,
+				v1beta1.CapacityTypeLabelKey: leastExpensiveOffering.CapacityType,
+				v1.LabelTopologyZone:         leastExpensiveOffering.Zone,
+			})
+
+			ds := test.DaemonSet()
+			ExpectApplied(ctx, env.Client, ds, nodeClaim, node, nodePool)
+
+			// Pods owned by a Deployment
+			pod := test.Pod(test.PodOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion:         "apps/v1",
+							Kind:               "DaemonSet",
+							Name:               ds.Name,
+							UID:                ds.UID,
+							Controller:         ptr.Bool(true),
+							BlockOwnerDeletion: ptr.Bool(true),
+						},
+					},
+				},
+				Conditions: []v1.PodCondition{{Type: v1.PodReady, Status: v1.ConditionTrue}},
+			})
+			ExpectApplied(ctx, env.Client, pod)
+
+			ExpectManualBinding(ctx, env.Client, pod, node)
+
+			// inform cluster state about nodes and nodeclaims
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+			fakeClock.Step(10 * time.Minute)
+
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
+			wg.Wait()
+
+			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
+
+			// Cascade any deletion of the nodeclaim to the node
+			ExpectNodeClaimsCascadeDeletion(ctx, env.Client, nodeClaim)
+
+			// we should delete the empty node
+			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(0))
+			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(0))
+			ExpectNotFound(ctx, env.Client, nodeClaim, node)
+		})
+		It("will consider a node with terminating Deployment pods as empty", func() {
+			// assign the nodeclaims to the least expensive offering so we don't get a replacement
+			nodeClaim.Labels = lo.Assign(nodeClaim.Labels, map[string]string{
+				v1.LabelInstanceTypeStable:   leastExpensiveInstance.Name,
+				v1beta1.CapacityTypeLabelKey: leastExpensiveOffering.CapacityType,
+				v1.LabelTopologyZone:         leastExpensiveOffering.Zone,
+			})
+			node.Labels = lo.Assign(node.Labels, map[string]string{
+				v1.LabelInstanceTypeStable:   leastExpensiveInstance.Name,
+				v1beta1.CapacityTypeLabelKey: leastExpensiveOffering.CapacityType,
+				v1.LabelTopologyZone:         leastExpensiveOffering.Zone,
+			})
+
+			rs := test.ReplicaSet()
+			ExpectApplied(ctx, env.Client, rs, nodeClaim, node, nodePool)
+
+			// Pod owned by a Deployment
+			pods := test.Pods(3, test.PodOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion:         "apps/v1",
+							Kind:               "ReplicaSet",
+							Name:               rs.Name,
+							UID:                rs.UID,
+							Controller:         ptr.Bool(true),
+							BlockOwnerDeletion: ptr.Bool(true),
+						},
+					},
+				},
+				Conditions: []v1.PodCondition{{Type: v1.PodReady, Status: v1.ConditionTrue}},
+			})
+			ExpectApplied(ctx, env.Client, lo.Map(pods, func(p *v1.Pod, _ int) client.Object { return p })...)
+
+			for _, p := range pods {
+				ExpectManualBinding(ctx, env.Client, p, node)
+			}
+
+			// Evict the pods off of the node
+			for _, p := range pods {
+				// Trigger an eviction to set the deletion timestamp but not delete the pod
+				Expect(env.KubernetesInterface.PolicyV1().Evictions(p.Namespace).Evict(ctx, &policyv1.Eviction{
+					ObjectMeta: metav1.ObjectMeta{Name: p.Name, Namespace: p.Namespace},
+				})).To(Succeed())
+				ExpectEvicted(ctx, env.Client, p)
+				ExpectExists(ctx, env.Client, p)
+			}
+
+			// inform cluster state about nodes and nodeclaims
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+			fakeClock.Step(10 * time.Minute)
+
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
+			wg.Wait()
+
+			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
+
+			// Cascade any deletion of the nodeclaim to the node
+			ExpectNodeClaimsCascadeDeletion(ctx, env.Client, nodeClaim)
+
+			// we should delete the empty node
+			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(0))
+			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(0))
+			ExpectNotFound(ctx, env.Client, nodeClaim, node)
+		})
+		It("will not consider a node with a terminating StatefulSet pod as empty", func() {
+			// assign the nodeclaims to the least expensive offering so we don't get a replacement
+			nodeClaim.Labels = lo.Assign(nodeClaim.Labels, map[string]string{
+				v1.LabelInstanceTypeStable:   leastExpensiveInstance.Name,
+				v1beta1.CapacityTypeLabelKey: leastExpensiveOffering.CapacityType,
+				v1.LabelTopologyZone:         leastExpensiveOffering.Zone,
+			})
+			node.Labels = lo.Assign(node.Labels, map[string]string{
+				v1.LabelInstanceTypeStable:   leastExpensiveInstance.Name,
+				v1beta1.CapacityTypeLabelKey: leastExpensiveOffering.CapacityType,
+				v1.LabelTopologyZone:         leastExpensiveOffering.Zone,
+			})
+
+			ss := test.StatefulSet()
+			ExpectApplied(ctx, env.Client, ss, nodeClaim, node, nodePool)
+
+			// Pod owned by a StatefulSet
+			pod := test.Pod(test.PodOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion:         "apps/v1",
+							Kind:               "StatefulSet",
+							Name:               ss.Name,
+							UID:                ss.UID,
+							Controller:         ptr.Bool(true),
+							BlockOwnerDeletion: ptr.Bool(true),
+						},
+					},
+				},
+				Conditions: []v1.PodCondition{{Type: v1.PodReady, Status: v1.ConditionTrue}},
+			})
+			ExpectApplied(ctx, env.Client, pod)
+
+			ExpectManualBinding(ctx, env.Client, pod, node)
+
+			// Trigger an eviction to set the deletion timestamp but not delete the pod
+			Expect(env.KubernetesInterface.PolicyV1().Evictions(pod.Namespace).Evict(ctx, &policyv1.Eviction{
+				ObjectMeta: metav1.ObjectMeta{Name: pod.Name, Namespace: pod.Namespace},
+			})).To(Succeed())
+			ExpectEvicted(ctx, env.Client, pod)
+			ExpectExists(ctx, env.Client, pod)
+
+			// inform cluster state about nodes and nodeclaims
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+			fakeClock.Step(10 * time.Minute)
+
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
+			wg.Wait()
+
+			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
+
+			// Cascade any deletion of the nodeclaim to the node
+			ExpectNodeClaimsCascadeDeletion(ctx, env.Client, nodeClaim)
+
+			// we shouldn't delete the node due to emptiness with a statefulset terminating pod
+			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(1))
+			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
+			ExpectExists(ctx, env.Client, nodeClaim)
+			ExpectExists(ctx, env.Client, node)
+		})
 	})
 	Context("Replace", func() {
 		DescribeTable("can replace node",
--- a/pkg/controllers/disruption/pdblimits.go
+++ b/pkg/controllers/disruption/pdblimits.go
@@ -24,6 +24,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	podutil "github.com/aws/karpenter-core/pkg/utils/pod"
 )
 
 // PDBLimits is used to evaluate if evicting a list of pods is possible.
@@ -56,8 +58,12 @@ func NewPDBLimits(ctx context.Context, kubeClient client.Client) (*PDBLimits, er
 
 // CanEvictPods returns true if every pod in the list is evictable. They may not all be evictable simultaneously, but
 // for every PDB that controls the pods at least one pod can be evicted.
+// nolint:gocyclo
 func (s *PDBLimits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 	for _, pod := range pods {
+		if !podutil.IsActive(pod) {
+			continue
+		}
 		for _, pdb := range s.pdbs {
 			if pdb.name.Namespace == pod.ObjectMeta.Namespace {
 				if pdb.selector.Matches(labels.Set(pod.Labels)) {
--- a/pkg/controllers/disruption/pdblimits.go
+++ b/pkg/controllers/disruption/pdblimits.go
@@ -18,6 +18,7 @@ package disruption
 
 import (
 	"context"
+	"time"
 
 	v1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
@@ -25,7 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	podutil "github.com/aws/karpenter-core/pkg/utils/pod"
+	podutil "sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
 // PDBLimits is used to evaluate if evicting a list of pods is possible.
@@ -61,11 +62,13 @@ func NewPDBLimits(ctx context.Context, kubeClient client.Client) (*PDBLimits, er
 // nolint:gocyclo
 func (s *PDBLimits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 	for _, pod := range pods {
-		if !podutil.IsActive(pod) {
+		// If the pod isn't an "active" pod or isn't eligible for being evicted, then a fully blocking PDB doesn't matter
+		// This is due to the fact that we won't call the eviction API on these pods when we are disrupting the node
+		if !podutil.IsActive(pod) || !podutil.IsEvictable(pod, time.Now()) {
 			continue
 		}
 		for _, pdb := range s.pdbs {
-			if pdb.name.Namespace == pod.ObjectMeta.Namespace {
+			if pdb.key.Namespace == pod.ObjectMeta.Namespace {
 				if pdb.selector.Matches(labels.Set(pod.Labels)) {
 
 					// if the PDB policy is set to allow evicting unhealthy pods, then it won't stop us from
@@ -81,7 +84,7 @@ func (s *PDBLimits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 					}
 
 					if !ignorePod && pdb.disruptionsAllowed == 0 {
-						return pdb.name, false
+						return pdb.key, false
 					}
 				}
 			}
@@ -91,7 +94,7 @@ func (s *PDBLimits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 }
 
 type pdbItem struct {
-	name                        client.ObjectKey
+	key                         client.ObjectKey
 	selector                    labels.Selector
 	disruptionsAllowed          int32
 	canAlwaysEvictUnhealthyPods bool
@@ -108,7 +111,7 @@ func newPdb(pdb policyv1.PodDisruptionBudget) (*pdbItem, error) {
 		canAlwaysEvictUnhealthyPods = true
 	}
 	return &pdbItem{
-		name:                        client.ObjectKeyFromObject(&pdb),
+		key:                         client.ObjectKeyFromObject(&pdb),
 		selector:                    selector,
 		disruptionsAllowed:          pdb.Status.DisruptionsAllowed,
 		canAlwaysEvictUnhealthyPods: canAlwaysEvictUnhealthyPods,
--- a/pkg/controllers/disruption/pdblimits.go
+++ b/pkg/controllers/disruption/pdblimits.go
@@ -18,12 +18,12 @@ package disruption
 
 import (
 	"context"
-	"time"
 
 	v1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	podutil "sigs.k8s.io/karpenter/pkg/utils/pod"
@@ -32,13 +32,15 @@ import (
 // PDBLimits is used to evaluate if evicting a list of pods is possible.
 type PDBLimits struct {
 	ctx        context.Context
+	clk        clock.Clock
 	kubeClient client.Client
 	pdbs       []*pdbItem
 }
 
-func NewPDBLimits(ctx context.Context, kubeClient client.Client) (*PDBLimits, error) {
+func NewPDBLimits(ctx context.Context, clk clock.Clock, kubeClient client.Client) (*PDBLimits, error) {
 	ps := &PDBLimits{
 		ctx:        ctx,
+		clk:        clk,
 		kubeClient: kubeClient,
 	}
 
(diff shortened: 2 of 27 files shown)


pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -79,32 +80,14 @@ func (t *Terminator) Taint(ctx context.Context, node *v1.Node) error {
 // Drain evicts pods from the node and returns true when all pods are evicted
 // https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
 func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
-	// Get evictable pods
-	pods := &v1.PodList{}
-	if err := t.kubeClient.List(ctx, pods, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+	pods, err := nodeutil.GetEvictablePods(ctx, t.kubeClient, t.clock.Now(), node)
```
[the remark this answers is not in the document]
njtran: It would be nice if we could put the !Active() call inside of the GetEvictablePods(), but I realize that IsEvictable() is wired through here, which needs to have the list of active pods to know when the node is fully drained. Any thoughts on how we can simplify here, as we're still stuck in the same situation (albeit better) as before, where we have distinct pod filtering logic that requires a non-shallow understanding of how everything wires together to know how if it's right.

pkg/controllers/nodeclaim/consistency/termination.go
```diff
@@ -48,7 +48,7 @@ func (t *Termination) Check(ctx context.Context, node *v1.Node, nodeClaim *v1bet
 	if err != nil {
 		return nil, err
 	}
-	pods, err := nodeutils.GetNodePods(ctx, t.kubeClient, node)
+	pods, err := nodeutils.GetPods(ctx, t.kubeClient, node)
```
njtran, line 51: Why can't this be IsActive()? We use the same criteria in the pdblimits code.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -154,45 +155,38 @@ func (p *Provisioner) CreateNodeClaims(ctx context.Context, nodeClaims []*schedu
...
 // constraints and consolidation
-func (p *Provisioner) consolidationWarnings(ctx context.Context, po v1.Pod) {
+func (p *Provisioner) consolidationWarnings(ctx context.Context, po *v1.Pod) {
 	// We have pending pods that have preferred anti-affinity or topology spread constraints.  These can interact
 	// unexpectedly with consolidation so we warn once per hour when we see these pods.
 	if po.Spec.Affinity != nil && po.Spec.Affinity.PodAntiAffinity != nil {
 		if len(po.Spec.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution) != 0 {
 			if p.cm.HasChanged(string(po.UID), "pod-antiaffinity") {
-				logging.FromContext(ctx).Infof("pod %s has a preferred Anti-Affinity which can prevent consolidation", client.ObjectKeyFromObject(&po))
+				logging.FromContext(ctx).Infof("pod %s has a preferred Anti-Affinity which can prevent consolidation", client.ObjectKeyFromObject(po))
```
njtran, line 182: adding suggestion to go with your other changes
```suggestion
				logging.FromContext(ctx).Infof("pod %q has a preferred Anti-Affinity which can prevent consolidation", client.ObjectKeyFromObject(po))
```

pkg/controllers/provisioning/provisioner.go
```diff
@@ -154,45 +155,38 @@ func (p *Provisioner) CreateNodeClaims(ctx context.Context, nodeClaims []*schedu
...
 			}
 		}
 	}
 	for _, tsc := range po.Spec.TopologySpreadConstraints {
 		if tsc.WhenUnsatisfiable == v1.ScheduleAnyway {
 			if p.cm.HasChanged(string(po.UID), "pod-topology-spread") {
-				logging.FromContext(ctx).Infof("pod %s has a preferred TopologySpreadConstraint which can prevent consolidation", client.ObjectKeyFromObject(&po))
+				logging.FromContext(ctx).Infof("pod %s has a preferred TopologySpreadConstraint which can prevent consolidation", client.ObjectKeyFromObject(po))
```
njtran, line 189:
```suggestion
				logging.FromContext(ctx).Infof("pod %q has a preferred TopologySpreadConstraint which can prevent consolidation", client.ObjectKeyFromObject(po))
```

pkg/controllers/disruption/pdblimits.go
```diff
@@ -85,7 +94,7 @@ func (s *PDBLimits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 }
 
 type pdbItem struct {
-	name                        client.ObjectKey
+	key                         client.ObjectKey
```
njtran, line 97: nit: I think it makes sense to name it the type that it is (or its inferred type of NamespacedName) for ease of understanding
```suggestion
	objectKey                         client.ObjectKey
```

pkg/controllers/disruption/pdblimits.go
```diff
@@ -56,10 +59,16 @@ func NewPDBLimits(ctx context.Context, kubeClient client.Client) (*PDBLimits, er
 
 // CanEvictPods returns true if every pod in the list is evictable. They may not all be evictable simultaneously, but
 // for every PDB that controls the pods at least one pod can be evicted.
+// nolint:gocyclo
 func (s *PDBLimits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 	for _, pod := range pods {
+		// If the pod isn't an "active" pod or isn't eligible for being evicted, then a fully blocking PDB doesn't matter
+		// This is due to the fact that we won't call the eviction API on these pods when we are disrupting the node
+		if !podutil.IsActive(pod) || !podutil.IsEvictable(pod, time.Now()) {
```
njtran, line 67: commenting here again just to make sure it's caught, I wonder if there's a way we can just encapsulate the IsActive() call here within the IsEvictable call, so that we don't have two interpretations of the same helper.

njtran (commented):
Thanks for doing this! Just looked at the code and not the tests.

pkg/utils/pod/scheduling.go
```diff
@@ -25,10 +27,51 @@ import (
...
+// IsEvictable checks if a pod is evictable by Karpenter by ensuring that the pod:
+// - Doesn't tolerate the "karepnter.sh/disruption=disrupting" taint
+// - Isn't a terminal pod (Failed or Succeeded)
+// - Isn't a pod that has been terminating past its terminationGracePeriodSeconds
+// - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
+func IsEvictable(pod *v1.Pod, now time.Time) bool {
+	return !ToleratesDisruptionNoScheduleTaint(pod) &&
+		!IsTerminal(pod) &&
+		!IsStuckTerminating(pod, now) &&
+		// Mirror pods cannot be deleted through the apiserver since they can't be controller
```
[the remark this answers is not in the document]
jonathan-innis: I'll clarify. I wrote this a while ago and reading this back I also don't know what I meant :D

pkg/utils/node/node.go
```diff
@@ -19,37 +19,63 @@ package node
...
+
+// GetProvisionablePods grabs all the pods from the passed nodes that satisfy the IsProvisionable criteria
+func GetProvisionablePods(ctx context.Context, kubeClient client.Client) ([]*v1.Pod, error) {
+	var podList v1.PodList
+	if err := kubeClient.List(ctx, &podList, client.MatchingFields{"spec.nodeName": ""}); err != nil {
+		return nil, fmt.Errorf("listing pods, %w", err)
+	}
+	return lo.FilterMap(podList.Items, func(p v1.Pod, _ int) (*v1.Pod, bool) {
+		return &p, pod.IsProvisionable(&p)
+	}), nil
```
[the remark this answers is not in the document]
jonathan-innis: Seems like a reasonable stance to say that the filtering function can stand on its own without needing to know about the pods that are passed to it. From the standpoint of getting all the pods from the node, I could see an argument for just getting _all_ of the pods, but it doesn't make sense to me if we could just use less memory and do less computational work above by just using the field selector to get the subset that we already need.

pkg/utils/node/node.go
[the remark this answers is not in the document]
jonathan-innis: No, wasn't needed given what filters are being used when we are doing scheduling. I could see an argument to add them just for symmetry, but I'd rather just add it as-needed rather than adding it before it's used.

pkg/controllers/node/termination/terminator/terminator.go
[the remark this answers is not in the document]
jonathan-innis: > that requires a non-shallow understanding of how everything wires together to know how if it's right

It's a good callout. The best thing that I can think about is moving the pod length check up and then filtering out the non-active pods before calling evict.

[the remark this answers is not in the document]
jonathan-innis: Agreed. I'm not sure there's a better way than throwing a comment in there to explain that it's for efficiency (since it really is just to avoid redundant eviction calls).

pkg/controllers/disruption/validation.go
```diff
@@ -87,10 +87,7 @@ func (v *Validation) IsValid(ctx context.Context, cmd Command) (bool, error) {
 	// Get the current representation of the proposed candidates from before the validation timeout
 	// We do this so that we can re-validate that the candidates that were computed before we made the decision are the same
 	// We perform filtering here to ensure that none of the proposed candidates have blocking PDBs or do-not-evict/do-not-disrupt pods scheduled to them
```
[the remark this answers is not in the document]
jonathan-innis: Good callout 👍🏼

pkg/controllers/disruption/types.go
```diff
@@ -113,31 +116,41 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
...
-		StateNode:    node.DeepCopy(),
-		instanceType: instanceType,
-		nodePool:     nodePool,
-		capacityType: node.Labels()[v1beta1.CapacityTypeLabelKey],
-		zone:         node.Labels()[v1.LabelTopologyZone],
-		pods:         pods,
+	for _, po := range pods {
+		// We only consider pods that are actively running for "karpenter.sh/do-not-disrupt"
+		// This means that we will allow Mirror Pods and DaemonSets to block disruption using this annotation
+		if pod.IsActive(po) && pod.HasDoNotDisrupt(po) {
```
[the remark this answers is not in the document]
jonathan-innis: Good idea. I like it 👍🏼

pkg/utils/node/node.go
```diff
@@ -19,37 +19,63 @@ package node
...
 
 	"sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
-// GetNodePods gets the list of schedulable pods from a variadic list of nodes
-// It ignores pods that are owned by the node, a daemonset or are in a terminal
-// or terminating state
-func GetNodePods(ctx context.Context, kubeClient client.Client, nodes ...*v1.Node) ([]*v1.Pod, error) {
+// GetPods grabs all pods that are currently bound to the passed nodes
+func GetPods(ctx context.Context, kubeClient client.Client, nodes ...*v1.Node) ([]*v1.Pod, error) {
```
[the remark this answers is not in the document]
jonathan-innis: I think this makes sense to export given it provides an easy function to grab all the pods attached to multiple nodes and this is being leveraged by the StateNode.Pods() member funtion

pkg/controllers/node/termination/terminator/terminator.go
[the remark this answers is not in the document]
jonathan-innis: I tried creating two separate functions `IsWaitingEviction()` and `IsEvictable()` but I thought that this added more confusion than the current implementation. Happy to discuss more if you have any more thoughts here.

pkg/controllers/nodeclaim/consistency/termination.go
jonathan-innis: I'm mainly trying to keep this consistent with the disruption code which grabs all pods and then passes the full set of pods into the PDB evaluation function

pkg/controllers/disruption/pdblimits.go
jonathan-innis: Sure, I could go either way on this. We call it `key` elsewhere in the code too rather than `objectKey`

pkg/controllers/disruption/pdblimits.go
jonathan-innis: Good callout. Changed the `IsEvictable` function. You've convinced me that we should just have an `IsEvictable` and `IsWaitingEviction` function

pkg/controllers/disruption/types.go
```diff
@@ -113,31 +116,41 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
...
+	}
+	return &Candidate{
+		StateNode:         node.DeepCopy(),
+		instanceType:      instanceType,
+		nodePool:          nodePool,
+		capacityType:      node.Labels()[v1beta1.CapacityTypeLabelKey],
+		zone:              node.Labels()[v1.LabelTopologyZone],
+		reschedulablePods: lo.Filter(pods, func(p *v1.Pod, _ int) bool { return pod.IsReschedulable(p) }),
+		// We get the disruption cost from all pods in the candidate, not just the reschedulable pods
+		disruptionCost: disruptionCost(ctx, pods) * lifetimeRemaining(clk, nodePool, node.Node),
```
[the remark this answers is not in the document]
jonathan-innis: Seems like we should just do it in a separate PR now. I can't think of a good reason not to -- seems like a good idea to me

pkg/utils/pod/scheduling.go
```diff
@@ -25,10 +27,51 @@ import (
...
+
+// IsReschedulable checks if a Karpenter should consider this pod when re-scheduling to new capacity by ensuring that the pod:
+// - Isn't a terminal pod (Failed or Succeeded)
+// - Isn't actively terminating
+// - Isn't owned by a DaemonSet
+// - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
+func IsReschedulable(pod *v1.Pod) bool {
+	// these pods don't need to be rescheduled
+	return !IsTerminal(pod) &&
+		!IsTerminating(pod) &&
```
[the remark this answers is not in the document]
jonathan-innis: +1 to collapsing this down to IsActive()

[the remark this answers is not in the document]
jonathan-innis: For the second question: I think that it needs a little more thought. It's a good question to ask and a good callout.

pkg/utils/pod/scheduling.go
```diff
@@ -25,10 +27,51 @@ import (
...
+// IsProvisionable checks if a pod needs to be scheduled to new capacity by Karpenter by ensuring that the pod:
+// - Has been marked as "Pending" by the kube-scheduler
+// - Has not been bound to a node
+// - Isn't currently preempting other pods on the cluster and about to schedule
+// - Isn't owned by a DaemonSet
+// - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
 func IsProvisionable(pod *v1.Pod) bool {
-	return !IsScheduled(pod) &&
+	return FailedToSchedule(pod) &&
+		!IsScheduled(pod) &&
```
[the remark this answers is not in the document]
jonathan-innis: Good callout. I didn't want to touch this since I didn't fully understand the ramifications of what removing one or the other would do and it didn't seem worth changing. I'll dig in to see when this was first added and if there was strong rationale for adding it in

pkg/utils/pod/scheduling.go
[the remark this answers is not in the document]
jonathan-innis: > This does beg the question: do we need to consider pods that are terminating as reschedulable

After messing around with terminating pod with pre-stop hooks, the answer here should be: no. As soon as a controlled pod gets removed and goes into a terminating state (evicted) it is replaced with a new version of the pod. If we counted terminating pods, we would be double-counting.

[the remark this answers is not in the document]
njtran: Synced offline, and we're aligned here. Generally this means any terminating pod is not reschedulable, unless it is owned by a stateful set.

pkg/utils/pod/scheduling.go
```diff
@@ -17,22 +17,85 @@ limitations under the License.
...
+func IsEvictable(pod *v1.Pod) bool {
+	return IsActive(pod) &&
+		!ToleratesDisruptionNoScheduleTaint(pod) &&
+		!IsOwnedByNode(pod)
+}
+
+// IsWaitingEviction checks if this is a pod that we are waiting to be removed from the node by ensuring that the pod:
+// - Isn't a terminal pod (Failed or Succeeded)
+// - Isn't a pod that has been terminating past its terminationGracePeriodSeconds
+// - Doesn't tolerate the "karepnter.sh/disruption=disrupting" taint
```
njtran:
```suggestion
// - Doesn't tolerate the "karpenter.sh/disruption=disrupting" taint
```

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -113,9 +101,6 @@ func (t *Terminator) Evict(pods []*v1.Pod) {
 	// 1. Prioritize noncritical pods, non-daemon pods https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
 	var criticalNonDaemon, criticalDaemon, nonCriticalNonDaemon, nonCriticalDaemon []*v1.Pod
 	for _, pod := range pods {
-		if !pod.DeletionTimestamp.IsZero() {
-			continue
-		}
```
njtran, line 118: Why are we skipping this now? Seems like we'll be enqueuing more eviction calls now, which I think is redundant.

pkg/test/expectations/expectations.go
```diff
@@ -568,3 +568,14 @@ func ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx context.Context,
...
 	}
 }
+
+func ExpectEvicted(ctx context.Context, c client.Client, pods ...*v1.Pod) {
+	GinkgoHelper()
+	for _, pod := range pods {
+		Eventually(func() bool {
+			return ExpectPodExists(ctx, c, pod.Name, pod.Namespace).GetDeletionTimestamp().IsZero()
+		}, ReconcilerPropagationTime, RequestInterval).Should(BeFalse(), func() string {
+			return fmt.Sprintf("expected %s/%s to be evicting, but it isn't", pod.Namespace, pod.Name)
```
njtran, line 578: nit: I was originally confused because I saw that we were expecting it to exist and not have deletion timestamp, but saw that we were expecting it to be false. I think the converse would semantically make more sense, but it's up to you

pkg/utils/node/node.go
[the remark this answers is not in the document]
njtran: Yeah that's fair. I think even though the code looks redundant, it is helping. Good point.

pkg/utils/pod/scheduling.go
```diff
@@ -58,6 +121,20 @@ func IsTerminating(pod *v1.Pod) bool {
 	return pod.DeletionTimestamp != nil
 }
 
+func IsStuckTerminating(pod *v1.Pod, clk clock.Clock) bool {
+	// The pod DeletionTimestamp will be set to the time the pod was deleted plus its
+	// grace period in seconds. We give an additional minute as a buffer to allow
+	// pods to force delete off the node before we actually go and deprovision the node
```
njtran, line 127: nit: i think we should migrate off the deprovision vocab since we are using disrupt/terminate
```suggestion
	// pods to force delete off the node before we actually go and terminate the node
```

pkg/controllers/disruption/consolidation_test.go
njtran, line 1: nice test coverage!

pkg/test/expectations/expectations.go
jonathan-innis: Agreed. Funny enough, I also got confused and thought that the logic was wrong here. This is technically not a change from the existing code, though; this is just a move to get this function over into the expectations file

pkg/controllers/node/termination/terminator/terminator.go
jonathan-innis: It's ruled out during the "evictablePods" check

Makefile
```diff
@@ -44,7 +44,7 @@ delete: ## Delete the controller from your ~/.kube/config cluster
 test: ## Run tests
 	go test ./... \
 		-race \
-		-timeout 10m \
+		-timeout 15m \
```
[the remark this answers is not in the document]
jonathan-innis: Agree, I think having to bump this timeout is a good signal that we should try and drop these down. I don't want to scope this in here but I do think that we need to get to this at some point.

pkg/controllers/disruption/consolidation_test.go
jonathan-innis: Thanks!

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		ExpectManualBinding(ctx, env.Client, pod, node)
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+		Expect(cluster.Nodes()).To(HaveLen(1))
+		_, err := disruption.NewCandidate(ctx, env.Client, recorder, fakeClock, cluster.Nodes()[0], pdbLimits, nodePoolMap, nodePoolInstanceTypeMap, queue)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(Equal(fmt.Sprintf(`pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(pod))))
+		Expect(recorder.DetectedEvent(fmt.Sprintf(`Cannot disrupt Node: Pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(pod)))).To(BeTrue())
+	})
+	It("should not consider candidates that have do-not-disrupt pods scheduled to mirror pods", func() {
```
njtran, line 517:
```suggestion
	It("should not consider candidates that have do-not-disrupt mirror pods scheduled", func() {
```

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		ExpectManualBinding(ctx, env.Client, pod, node)
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+		Expect(cluster.Nodes()).To(HaveLen(1))
+		_, err := disruption.NewCandidate(ctx, env.Client, recorder, fakeClock, cluster.Nodes()[0], pdbLimits, nodePoolMap, nodePoolInstanceTypeMap, queue)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(Equal(fmt.Sprintf(`pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(pod))))
+		Expect(recorder.DetectedEvent(fmt.Sprintf(`Cannot disrupt Node: Pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(pod)))).To(BeTrue())
+	})
+	It("should not consider candidates that have do-not-disrupt pods scheduled to DaemonSet pods", func() {
```
njtran, line 555:
```suggestion
	It("should not consider candidates that have do-not-disrupt daemonset pods scheduled", func() {
```

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		ExpectManualBinding(ctx, env.Client, pod, node)
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+		Expect(cluster.Nodes()).To(HaveLen(1))
+		_, err := disruption.NewCandidate(ctx, env.Client, recorder, fakeClock, cluster.Nodes()[0], pdbLimits, nodePoolMap, nodePoolInstanceTypeMap, queue)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(Equal(fmt.Sprintf(`pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(pod))))
+		Expect(recorder.DetectedEvent(fmt.Sprintf(`Cannot disrupt Node: Pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(pod)))).To(BeTrue())
+	})
+	It("should consider candidates that have do-not-disrupt pods on terminating pods", func() {
```
njtran, line 594:
```suggestion
	It("should consider candidates that have do-not-disrupt terminating pods", func() {
```

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+
+		ExpectDeletionTimestampSet(ctx, env.Client, pod)
+
+		Expect(cluster.Nodes()).To(HaveLen(1))
+		c, err := disruption.NewCandidate(ctx, env.Client, recorder, fakeClock, cluster.Nodes()[0], pdbLimits, nodePoolMap, nodePoolInstanceTypeMap, queue)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(c.NodeClaim).ToNot(BeNil())
+		Expect(c.Node).ToNot(BeNil())
+	})
+	It("should consider candidates that have do-not-disrupt pods on terminal pods", func() {
```
njtran, line 624: Seems like this is how you wanted it done, but I think they each read better like this
```suggestion
	It("should consider candidates that have do-not-disrupt terminal pods", func() {
```

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+				Labels: map[string]string{
+					v1beta1.NodePoolLabelKey:     nodePool.Name,
+					v1.LabelInstanceTypeStable:   mostExpensiveInstance.Name,
+					v1beta1.CapacityTypeLabelKey: mostExpensiveOffering.CapacityType,
+					v1.LabelTopologyZone:         mostExpensiveOffering.Zone,
+				},
+			},
+		})
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{}, []*v1beta1.NodeClaim{nodeClaim})
```
njtran, line 927: nit
```suggestion
		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, nil, []*v1beta1.NodeClaim{nodeClaim})
```

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+
+		Expect(cluster.Nodes()).To(HaveLen(1))
+		_, err := disruption.NewCandidate(ctx, env.Client, recorder, fakeClock, cluster.Nodes()[0], pdbLimits, nodePoolMap, nodePoolInstanceTypeMap, queue)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(Equal(`instance type "" can't be resolved`))
+		Expect(recorder.DetectedEvent(`Cannot disrupt Node: Instance type "" not found`)).To(BeTrue())
+	})
+	It("should not consider candidates who's instance type cannot be resolved", func() {
```
njtran, line 1118:
```suggestion
	It("should not consider candidates whose instance type cannot be resolved", func() {
```

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,700 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		// Mock the InstanceType not existing by removing it from the nodePoolInstanceTypes map
+		delete(nodePoolInstanceTypeMap[nodePool.Name], mostExpensiveInstance.Name)
+
+		Expect(cluster.Nodes()).To(HaveLen(1))
+		_, err := disruption.NewCandidate(ctx, env.Client, recorder, fakeClock, cluster.Nodes()[0], pdbLimits, nodePoolMap, nodePoolInstanceTypeMap, queue)
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(Equal(fmt.Sprintf("instance type %q can't be resolved", mostExpensiveInstance.Name)))
+		Expect(recorder.DetectedEvent(fmt.Sprintf("Cannot disrupt Node: Instance type %q not found", mostExpensiveInstance.Name))).To(BeTrue())
+	})
+	It("should not consider candidates who are actively being processed in the queue", func() {
```
njtran, line 1141:
```suggestion
	It("should not consider candidates that are actively being processed in the queue", func() {
```

pkg/controllers/nodeclaim/disruption/emptiness_test.go
```diff
@@ -78,6 +80,77 @@ var _ = Describe("Emptiness", func() {
...
+			NodeName:   node.Name,
+			Conditions: []v1.PodCondition{{Type: v1.PodReady, Status: v1.ConditionTrue}},
+		})
+		ExpectApplied(ctx, env.Client, lo.Map(pods, func(p *v1.Pod, _ int) client.Object { return p })...)
+
+		for _, p := range pods {
+			// Trigger an eviction to set the deletion timestamp but not delete the pod
+			Expect(env.KubernetesInterface.PolicyV1().Evictions(p.Namespace).Evict(ctx, &policyv1.Eviction{
+				ObjectMeta: metav1.ObjectMeta{Name: p.Name, Namespace: p.Namespace},
+			})).To(Succeed())
```
njtran, line 113: This seems like something we could just model in ExpectEvicted, right?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3386,6 +3387,222 @@ var _ = Context("NodePool", func() {
...
+
+			// Mark for deletion so that we consider all pods on this node for reschedulability
+			cluster.MarkForDeletion(node.Spec.ProviderID)
+
+			// Trigger an eviction to set the deletion timestamp but not delete the pod
+			Expect(env.KubernetesInterface.PolicyV1().Evictions(pod.Namespace).Evict(ctx, &policyv1.Eviction{
+				ObjectMeta: metav1.ObjectMeta{Name: pod.Name, Namespace: pod.Namespace},
+			})).To(Succeed())
+			ExpectEvicted(ctx, env.Client, pod)
+			ExpectExists(ctx, env.Client, pod)
```
njtran, line 3438: Same here on putting the eviction call in the ExpectEvicted() block. It seems like ExpectEvicted is almost similar if not the same as a ExpectDeleting(obj)

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3386,6 +3387,222 @@ var _ = Context("NodePool", func() {
...
+
+			// Trigger a provisioning loop and expect that we don't create more nodes since we don't consider
+			// generic terminating pods
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov)
+
+			// We shouldn't create an additional node here because this is a standard pod
+			nodes := ExpectNodes(ctx, env.Client)
+			Expect(nodes).To(HaveLen(1))
+		})
+		It("should not reschedule pods from a deleting node when pods are not active and they are owned by a ReplicaSet", func() {
```
njtran, line 3508: Should we actually be rescheduling pods from a deleting node if the previous replica gets force terminated before the new replica actually goes active? I think right now we'd undershoot here, by not including this pod.

pkg/test/expectations/expectations.go
```diff
@@ -568,3 +568,14 @@ func ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx context.Context,
 		ExpectReconcileSucceeded(ctx, nodeClaimStateController, client.ObjectKeyFromObject(m))
 	}
 }
+
+func ExpectEvicted(ctx context.Context, c client.Client, pods ...*v1.Pod) {
```
njtran, line 572: I think this should just have the evict() call in here, and call an ExpectDeleted() or ExpectDeleting()

njtran (commented):
Nice work! Some nitty comments but one real comment

pkg/controllers/provisioning/scheduling/suite_test.go
jonathan-innis: > I think right now we'd undershoot here, by not including this pod

I'm not sure how we would know this. In the general case the new replicas get created immediately (or as quickly as they can be). Force deleting is a rare occurrence and, in the general case for replicasets and deployments, you are going to get a pending pod spun up as soon as the old pod goes into a terminating state.

njtran: So the idea is potentially undershooting, but possibly getting pulled into the provisioning loop anyway through the batching, is that right?

jonathan-innis: > but possibly getting pulled into the provisioning loop anyway through the batching, is that right

I don't think we will generally undershoot. In the worst-case, yes we may undershoot, but this should be rare. Realistically, this is better than the alternative where we overshoot and then have to pull it back through consolidation. Or worse -- they haven't enabled consolidation and then they are just overpaying.

jonathan-innis:
/unhold

Confirm the logic and everything LGTM

PR #791, opened 2023-11-15
staging: add budgets helpers and more validations

njtran:
Fixes #N/A 

**Description**
This adds helpers that are going to be used when considering disruption budgets for disruption. Adds more CEL validations on the values of durations.

make presubmit


pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
...
+	}
+	// return the values, defaulting to -1 if the value is MaxInt64
+	return intstr.FromInt(lo.Ternary(minIntVal == math.MaxInt64, -1, minIntVal)), intstr.FromString(fmt.Sprintf("%d%%", lo.Ternary(minPercentVal == math.MaxInt64, -1, minPercentVal))), nil
+}
+
+// GetAllowedDisruptions returns an intstr.IntOrString that can be used a comparison
+// for calculating if a disruption action is allowed. It returns an error if the
+// crontab is invalid. This returns -1 if the value is unbounded.
+func (b *Budget) GetAllowedDisruptions(c clock.Clock) (intstr.IntOrString, error) {
+	active, err := b.IsActive(c)
```
jonathan-innis, line 256: When does this error? Is there ever a valid case where this errors out?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
 		return ptr.Int32Value(pl.Items[a].Spec.Weight) > ptr.Int32Value(pl.Items[b].Spec.Weight)
 	})
 }
+
+// GetAllowedDisruptions returns the minimum allowed disruptions across all disruption budgets for a given node pool.
+// This returns two values as the resolved value for a percent depends on the number of current node claims.
+func (in *NodePool) GetAllowedDisruptions(ctx context.Context, c clock.Clock) (intstr.IntOrString, intstr.IntOrString, error) {
+	vals := make([]intstr.IntOrString, len(in.Spec.Disruption.Budgets))
+	errs := make([]error, len(in.Spec.Disruption.Budgets))
+	workqueue.ParallelizeUntil(ctx, len(in.Spec.Disruption.Budgets), len(in.Spec.Disruption.Budgets), func(i int) {
```
jonathan-innis, line 219: Any particular reason that we want to parallelize this? Do we expect this to be complicated or take a long time?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
 		return ptr.Int32Value(pl.Items[a].Spec.Weight) > ptr.Int32Value(pl.Items[b].Spec.Weight)
 	})
 }
+
+// GetAllowedDisruptions returns the minimum allowed disruptions across all disruption budgets for a given node pool.
+// This returns two values as the resolved value for a percent depends on the number of current node claims.
+func (in *NodePool) GetAllowedDisruptions(ctx context.Context, c clock.Clock) (intstr.IntOrString, intstr.IntOrString, error) {
```
jonathan-innis, line 216: I wasn't quite getting this from the description, but my assumption is that one of these values is going to be a percentage value and the other is going to be an integer value. Is there a reason that they are both `IntOrString` since one should always be a string and the other should always be an integer?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
...
+		vals[i] = val
+	})
+	if err := multierr.Combine(errs...); err != nil {
+		return intstr.IntOrString{}, intstr.IntOrString{}, err
+	}
+	minIntVal, minPercentVal := math.MaxInt64, math.MaxInt64
+	for i := range vals {
+		val := vals[i]
+		// The crontab wasn't active
+		if val.IntVal == -1 {
```
jonathan-innis, line 233: Does -1 mean not active or does -1 mean infinite? If I look below, I thought that you had mentioned that -1 was infinite. I was expecting an inactive budget to have an integer value of 0

pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
...
+			continue
+		}
+		// This returns the percent value if it's a string, and the raw value if it's an int.
+		temp, err := intstr.GetScaledValueFromIntOrPercent(lo.ToPtr(val), 100, false)
+		if err != nil {
+			// Should almost never happen since this is validated when the nodepool is applied
```
jonathan-innis, line 239: Can we guarantee that this should never happen? I'd like to avoid having errors that "pretty much can't ever happen" since I question why we should be handling the error in the code at all

pkg/apis/v1beta1/nodepool.go
jonathan-innis, line 216: Rather than handling everything as `IntOrString` does it make sense to just divide up the budgets into percentage-based ones and integer-based ones and take the minimum that way rather than handling both cases simultaneously and trying to split it later?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
...
+// If the last crontab hit is exactly the duration in the past, this means the
+// schedule is active, as any more crontab hits in between would only extend this
+// window. This ensures that any previous crontab hits for a schedule are considered.
+func (b *Budget) IsActive(c clock.Clock) (bool, error) {
+	if b.Crontab == nil && b.Duration == nil {
+		return true, nil
+	}
+	schedule, err := cron.ParseStandard(lo.FromPtr(b.Crontab))
+	if err != nil {
+		return false, fmt.Errorf("parsing crontab, %w", err)
```
jonathan-innis, line 275: Same question here. Should this ever fail?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -202,3 +210,72 @@ func (pl *NodePoolList) OrderByWeight() {
...
+	}
+	// Walk back in time for the duration associated with the crontab
+	checkPoint := c.Now().Add(-lo.FromPtr(b.Duration).Duration)
+	nextHit := schedule.Next(checkPoint)
```
jonathan-innis, line 279: Is there a clearer way to do this logic? I get how it's working but the way it's written makes it a little hard to follow. Perhaps an example in the comment could help clear it up but I'm also curious if there's a concept of "between" where we can get the last crontab and see if we are currently between that time and the duration after it

pkg/apis/v1beta1/nodepool_test.go
```diff
@@ -0,0 +1,177 @@
+/*
```
jonathan-innis, line 1: optional: Should this file be `nodepool_budgets_test`?

pkg/apis/v1beta1/nodepool_validation_cel_test.go
```diff
@@ -105,15 +105,49 @@ var _ = Describe("CEL/Validation", func() {
 			nodePool.Spec.Disruption.Budgets = []Budget{{
```
jonathan-innis, line 105: Should we be making the same test changes to the `nodepool_validation_webhook_test`?

pkg/apis/v1beta1/nodepool.go
njtran: Was just doing this to make it quicker and do the multierr handling like we do elsewhere. Don't mind moving it to a for loop with err.Append() calls if that's what you prefer.

pkg/apis/v1beta1/nodepool.go
njtran: The IntOrString has a Type value that allows the GetScaledValue to infer whether it needs to do a percentage or a raw value. I could probably put both of the values in the return value here, but we'd need to in-place modify the type to compare the two resolved values together, where I think having two values may just more deliberate here.

pkg/apis/v1beta1/nodepool.go
njtran: not active and infinite is the same. If a schedule isn't active, there's no budget that should be applied. An inactive budget defaulting to 0 means that when the budget isn't active, no disruption can happen, which is the opposite of what we want here.

pkg/apis/v1beta1/nodepool.go
njtran: If the validation mechanism isn't able to validate/default an object for some reason, a user could get themselves into config error. What are your thoughts on this?

pkg/apis/v1beta1/nodepool.go
njtran: I could see an argument for this. This is essentially what I do for the comparison in this function, but if we don't return an intstr.IntOrString() here, we have to create a new one in the caller function to then create the intstr.IntOrString() as an input to GetScaledValue().

pkg/apis/v1beta1/nodepool.go
njtran: If you have a suggestion on how to make it clearer in code, I'm open to it :D 

Let me know if this example helps:
```
Crontab = @hourly == 0 * * * * and time.Now() = 10:15 and duration = 30m, we'll walk back to 9:45, get the next crontab hit from there, which should be 10, which is before 10:15, meaning the cron is active. 
Doing this with the same but time.Now() = 10:45, we walk back in time to 10:15 and then get the next crontab hit, which is 11:00. Since this is after 10:45, we say that the cron is not active.
```

pkg/apis/v1beta1/nodepool_validation_cel_test.go
njtran: Will do in a follow up when I add those tests.

jonathan-innis: I see. So we're not adding them right now because we don't yet have the webhook validations. Makes sense.

pkg/apis/v1beta1/nodepool.go
garvinp-stripe: I think (maybe thats just me) the naming can be a bit confusing. Rather than calling it `nextHit` maybe it should be something like `closestLastHit`?

garvinp-stripe: I think its difficult given the cron parser only gives you `Next` based on a time and finding the last crontab is quite hard to do. I think this method make sense even though is slightly complicated to understand

pkg/apis/v1beta1/nodepool.go
njtran: It errors out when the the cron parser catches something that validation didn't. Since regex is notorious for having edge cases, I'd like to keep this here regardless of how fleshed out our validation is to ensure that we're always getting a schedule that's valid.

pkg/apis/v1beta1/nodepool.go
njtran: If there's an edge case with our regex that the cron parser catches but we don't, then yes.

njtran:
closed and reopened to re-trigger presubmits that weren't showing up

pkg/apis/v1beta1/nodepool.go
```diff
@@ -98,20 +104,25 @@ type Budget struct {
 	// MaxUnavailable dictates how many NodeClaims owned by this NodePool
 	// can be terminating at once. It must be set.
 	// This only considers NodeClaims with the karpenter.sh/disruption taint.
-	// +kubebuilder:validation:XIntOrString
+	// We can't use an intstr.IntOrString since kubebuilder doesn't support pattern
+	// checking for int values for IntOrString values.
+	// Ref: https://github.com/kubernetes-sigs/controller-tools/blob/55efe4be40394a288216dab63156b0a64fb82929/pkg/crd/markers/validation.go#L379-L388
+	// +kubebuilder:validation:Type="string"
```
jonathan-innis, line 110: Do you need this validation type now that you are explicitly defining this value as a string?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -199,3 +210,79 @@ func (pl *NodePoolList) OrderByWeight() {
...
+			minPercentVal = lo.Ternary(temp < minPercentVal, temp, minPercentVal)
+		}
+	}
+	// return the values, defaulting to -1 if the value is MaxInt64
+	return intstr.FromInt(lo.Ternary(minIntVal == math.MaxInt64, -1, minIntVal)), intstr.FromString(fmt.Sprintf("%d%%", lo.Ternary(minPercentVal == math.MaxInt64, -1, minPercentVal))), nil
+}
+
+// GetAllowedDisruptions returns an intstr.IntOrString that can be used a comparison
+// for calculating if a disruption action is allowed. It returns an error if the
+// crontab is invalid. This returns -1 if the value is unbounded.
```
jonathan-innis, line 254: I wonder if it makes more sense to use MAX_INT here so that what's stored in the value is more indicative of what's actually happening here. Otherwise, we are having to reason about what `-1` means here. Thoughts?

jonathan-innis (approved):
LGTM 🚀

PR #794, opened 2023-11-16
fix: Fix `RuntimeValidate` check to only check NodePools

jonathan-innis:
Releated to https://github.com/aws/karpenter/issues/5055

**Description**

This PR ensures that `RuntimeValidate()` calls only run against NodePools since Provisioners should be validated using the webhooks and we shouldn't have the same validation logic that we use against NodePools happening against Provisioners

`make presubmit`
`make apply`

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -40,6 +40,14 @@ linters-settings:
     min-complexity: 11
   govet:
     check-shadowing: true
+  revive:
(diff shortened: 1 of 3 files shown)


engedaam (dismissed):
LGTM 🚀

engedaam (approved):
LGTM 🚀

PR #801, opened 2023-11-17
ci: Pin Github Action versions to hashes

jonathan-innis:
Fixes #N/A 

**Description**

This PR pins all of our third-party actions to a full-length SHA to ensure they are not vulnerable to a supply-chain attack. This is the best practice for third-party GHA as described here: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions


jmdeal (approved):
LGTM 🚀

PR #813, opened 2023-11-22
chore: bumping length of the message we store on the nodeclaim

Bryce-Soghigian:
Fixes #N/A 

**Description**
Azure Returns back long errors on the nodeclaim, and they are truncated to 300 characters. The Azure SDK For Go error messages have some wrappers around them that add quite a few additional characters onto the error message. 

So as a result, we get a very abruptly truncated error message. 
```
Status:
  Conditions:
    Last Transition Time:  2023-11-21T22:06:39Z
    Message:               Node not launched
    Reason:                NotLaunched
    Status:                False
    Type:                  Initialized
    Last Transition Time:  2023-11-21T22:06:39Z
    Message:               creating instance, virtualMachine.BeginCreateOrUpdate for VM "aks-general-purpose-ljm95" failed: PUT https://management.azure.com/subscriptions/xxxxxxx1/resourceGroups/MC_sillygoose-rg-2_sillygoose-nap-cluster_uksouth/providers/Microsoft.Compute/virtualMachines/aks-genera...
    Reason:                LaunchFailed
    Status:                False
    Type:                  Launched
    Last Transition Time:  2023-11-21T22:06:39Z
```
    
    I would prefer to extend this to 1000 characters to show the important bits of the azure errors. 
    

- make presubmit
- inflate scale up to view nodeclaim. 

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/launch.go
+++ b/pkg/controllers/nodeclaim/lifecycle/launch.go
@@ -148,8 +148,8 @@ func PopulateNodeClaimDetails(nodeClaim, retrieved *v1beta1.NodeClaim) *v1beta1.
 }
 
 func truncateMessage(msg string) string {
-	if len(msg) < 300 {
+	if len(msg) < 1000 {
 		return msg
 	}
-	return msg[:300] + "..."
+	return msg[:1000] + "..."
 }


jonathan-innis:
Discussed offline. Looking at the upstream API guidance documentation, there is an ask that the message that is returned by the status condition be in a ["human readable"](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) format. I wonder if we should parse down some of the information that we are sending back from the CloudProvider Create call right now since it looks like we are sending back a JSON body which might not fit the standard for [human readable](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties).

Bryce-Soghigian:
@jonathan-innis I agree and its not exactly great to increase etcd load at scale to store long messages. Wondering how these crds perform at 5k cluster scale. 

I think the real solution is to add some string parsing and construct our own error messages. Not returning the sdk errors directly, and instead return a more selective message in the cloud provider. Since we detect the type, its fairly easy to return our own human readable error message.

PR #818, opened 2023-11-24
feat: Nodeclass readiness into nodepool status

### LINKED ISSUE #493: Surface health conditions of Karpenter resources as status conditions

sadath-12:
Fixes #493  (nodeclass part)  

**Description**

Users can inspect there nodeclass readiness in the status conditions of nodepool

```make verify```

### CODE REVIEW

File: pkg/controllers/nodeclaim/lifecycle/launch.go [hunk 1]
```diff
@@ -61,7 +62,22 @@ func (l *Launch) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (r
 		created, err = l.linkNodeClaim(ctx, nodeClaim)
 	} else {
 		created, err = l.launchNodeClaim(ctx, nodeClaim)
+		npKey := nodepoolutil.Key{Name: nodeClaim.Labels[v1beta1.NodePoolLabelKey], IsProvisioner: nodeClaim.IsMachine}
+		if npKey.Name != "" {
```
sadath-12, line 66: ```if npKey.Name != "" ``` this condition is added since tests panic . I think in real scenario it wun't be the case we can either leave as it is or apply nodepool in our tests wherever nodeclaim is applied

pkg/controllers/nodeclaim/lifecycle/launch.go
```diff
@@ -61,7 +62,22 @@ func (l *Launch) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (r
 		created, err = l.linkNodeClaim(ctx, nodeClaim)
 	} else {
 		created, err = l.launchNodeClaim(ctx, nodeClaim)
+		npKey := nodepoolutil.Key{Name: nodeClaim.Labels[v1beta1.NodePoolLabelKey], IsProvisioner: nodeClaim.IsMachine}
```
jonathan-innis, line 65: We shouldn't add the NodePool status condition update here since this controller is intended for the NodeClaim reconciliation. We basically need a NodePool controller that reconciles and checks the status condition of the `nodeClaimRef` and use that reconcile result to update the `NodeClassReady` condition which should further update the top-level `Ready` condition

sadath-12:
```NodeclassRef``` is under nodeclaim and so we would update nodepool at every update of nodeclaim ( for the ready condition in nodeclassRef under nodeclaimSpec which we can add  as well the launched condition ). So if we would make a new controller for nodepool we would still be checking for nodeclaim updates . So is it better to do within nodeclaim lifecycle controllers itself ?

**Comment by jonathan-innis** [comment]
> we would still be checking for nodeclaim updates

This isn't true. A NodePool can exist without a NodeClaim. As a general principle, the owner of the condition (most often where the condition exists) should be the thing that is being reconciled when determining the state of the condition. In this case, the NodePool is the object that has the `Ready` condition so we should build a controller that reconciles _that_ object without assuming any details of other objects that propagate off of it.

**Comment by sadath-12** [comment]
Ok work in progress .

File: pkg/apis/v1beta1/nodeclaim.go [hunk 3]
```diff
@@ -138,13 +138,13 @@ type KubeletConfiguration struct {
 
 type NodeClassReference struct {
 	// Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"
-	// +optional
+	// +required
```
jonathan-innis, line 141: We can't just make these fields required outright since this will break users if we change this all-of-a-sudden. We need a migration path to ensuring that these fields always exist. We should look into defaulting through OpenAPI as an example for the AWS cloudprovider implementation to ensure that users aren't broken.

pkg/apis/v1beta1/nodepool_status.go
```diff
@@ -23,4 +23,35 @@ type NodePoolStatus struct {
 	// Resources is the list of resources that have been provisioned.
 	// +optional
 	Resources v1.ResourceList `json:"resources,omitempty"`
+	// Conditions represent the latest available observations of a NodePool's current state.
+	// +optional
+	Conditions []NodePoolCondition `json:"conditions,omitempty"`
+}
+
+// NodePoolConditionType represents a NodePool condition value.
+type NodePoolCondition struct {
```
jonathan-innis, line 32: There's already a generalized condition type that we use from knative. See the conditions field in the NodeClaim

pkg/controllers/nodepool/lifecycle/controller.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+
+func (c *Controller) Reconcile(ctx context.Context, np *v1beta1.NodePool) (reconcile.Result, error) {
+    nodeClassRef := np.Spec.Template.Spec.NodeClassRef
+    if nodeClassRef == nil {
+        return reconcile.Result{}, fmt.Errorf("nodeClassRef is nil")
+    }
+    // Fetch the resource dynamically using Unstructured
+    config := controllerruntime.GetConfigOrDie()
```
jonathan-innis, line 58: We shouldn't build the client on each reconcile loop. This building of the dynamic client should probably happen in the `operator.go` file and then we pass through the dynamic client in the constructor

pkg/controllers/nodepool/lifecycle/controller.go
```diff
@@ -0,0 +1,137 @@
...
+    dynamicClient, err := dynamic.NewForConfig(config)
+    if err != nil {
+        return reconcile.Result{}, fmt.Errorf("failed to create dynamic client: %v", err)
+    }
+
+    gvr := schema.GroupVersionResource{
+        Group:    strings.Split(nodeClassRef.APIVersion,"/")[0],
```
jonathan-innis, line 65: Rather than parsing like this, we should make use of upstream utility functions. It always feels a bit risky to roll our own logic like this

pkg/controllers/nodepool/lifecycle/controller.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func NewNodePoolController(kubeClient client.Client) corecontroller.Controller {
+	return corecontroller.Typed[*v1beta1.NodePool](kubeClient, &NodePoolController{
+		Controller: NewController(kubeClient),
+	})
+}
+
+func (c *NodePoolController) Name() string {
+	return "nodepool.lifecycle"
```
jonathan-innis, line 127: This feels like the wrong name for this controller. Maybe `nodepool.status`?

pkg/controllers/nodepool/lifecycle/controller.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func (c *NodePoolController) Builder(_ context.Context, m manager.Manager) corecontroller.Builder {
+	return corecontroller.Adapt(controllerruntime.
+		NewControllerManagedBy(m).
```
jonathan-innis, line 132: How do we make sure that we requeue when there is a status change in the NodeClass?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -131,6 +131,14 @@ func (p *Provisioner) Reconcile(ctx context.Context, _ reconcile.Request) (resul
 	if len(results.NewNodeClaims) == 0 {
 		return reconcile.Result{}, nil
 	}
+	np , err := nodepoolutil.Get(ctx, p.kubeClient, results.NewNodeClaims[0].OwnerKey)
```
jonathan-innis, line 134:
```suggestion
	np, err := nodepoolutil.Get(ctx, p.kubeClient, results.NewNodeClaims[0].OwnerKey)
```

nit

jonathan-innis, line 134: I think this change may need a rebase since the `OwnerKey` field shouldn't exist anymore with some recent changes

jonathan-innis, line 134: We shouldn't grab the NodePool here. I think the right place to do this is in the filtering logic on L212, where we already remove NodePools that are deleting, we can just add the logic that filters out NodePools that aren't ready

pkg/controllers/provisioning/provisioner.go
```diff
@@ -131,6 +131,14 @@ func (p *Provisioner) Reconcile(ctx context.Context, _ reconcile.Request) (resul
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+	if !np.GetCondition(v1beta1.NodeClassConditionTypeReady){
+		logging.FromContext(ctx).Debugf("waiting for nodeclass to become ready")
```
jonathan-innis, line 139: I think the readiness condition in the NodePool should be enough for observability, not sure if we need a log line here

pkg/utils/nodepool/nodepool.go
```diff
@@ -132,3 +135,38 @@ func PatchStatus(ctx context.Context, c client.Client, stored, nodePool *v1beta1
 func HashAnnotation(nodePool *v1beta1.NodePool) map[string]string {
 	return map[string]string{v1beta1.NodePoolHashAnnotationKey: nodePool.Hash()}
 }
+
+func UpdateStatusCondition(ctx context.Context, c client.Client, np *v1beta1.NodePool, conditionType v1beta1.NodePoolConditionType, condition v1.ConditionStatus) {
```
jonathan-innis, line 139: A lot of these functions are already shipped as part of `knative.dev/pkg`. We should just use the same logic that we are using elsewhere in the codebase. See things like `MarkTrue` that's present in some of the NodeClaim lifecycle controllers.

pkg/controllers/nodepool/lifecycle/controller.go
sadath-12: yeah, seems right

PR #831, opened 2023-11-29
chore: fallback to sudo

jackfrancis:
Fixes #N/A 

**Description**

This PR replaces explicit sudo usages with a fallback to sudo only if the non-sudo equivalent fails. This is to address execution runtimes where the sudo command doesn't exist and the user context is root-equivalent.


jonathan-innis (approved):
/lgtm
/approve

PR #833, opened 2023-11-30
ci: use kube-apiserver v1.25.16 for k8s 1.25 CI

### LINKED ISSUE #803: Transient Failures in CI for `K8S_VERSION` 1.25
### Problem

We are transient CI failures in our system due to a bug in the kube-apiserver that is causing a panic on the apiserver due to a race in the evaluation of CEL expression. This race was first reported here: https://github.com/kubernetes/kubernetes/issues/114661 and was later fixed by a Kubernetes PR here: https://github.com/kubernetes/kubernetes/pull/114857 that bumped the package version of the CEL evaluation package where this bug was originating: https://github.com/google/cel-go/pull/620.

Screenshot 2023-11-17 at 10 55 54 AM

Kubernetes did back-port patch versions that contained the fix for this bug in the apiserver. These can be seen at the bottom of the originally reported issue.

Screenshot 2023-11-17 at 10 58 32 AM

You can also see these cherry-picks by viewing the Kubernetes changelog for 1.24, 1.25, and 1.26. 

- 1.24: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#changelog-since-v1249
- 1.25: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.25.md#changelog-since-v1255
- 1.26: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.26.md#changelog-since-v1260

setup-envtest does not currently surface every kubernetes version through its mirror where it pulls the etcd and apiserver binaries. These mirrors can be seen here: https://storage.googleapis.com/kubebuilder-tools. As a result, we will continue to see these transient failures on 1.25 until they publish a newer patch version of these binaries that contains the fix. 

We are only seeing this version on Kubernetes 1.25 because we have disabled CEL on versions less than 1.25 (because the feature was not yet in beta and wasn’t enabled on EKS clusters) and the bug is fixed on versions greater than 1.25 because setup-envtest has a new enough binary that contains the patch fix.

### Solutions

1. Upstream `controller-runtime` can support these newer patch versions in their mirror directly. I've currently opened an issue to ask for this here: https://github.com/kubernetes-sigs/controller-runtime/issues/2583. 
2. We can download and install the binaries into the correct directories ourselves for 1.25. `envtest` would then just use these binaries directly. 

For now, we can continue to retry when we see failures on 1.25 since these errors are transient and resolve after some re-runs; however, we should consider fixing this so our CI doesn't become flaky for 1.25 due to this interaction with CEL.

jmdeal:
Fixes #803  

**Description**
This PR downloads kube-apiserver v1.25.16 for use with envtest when running CI jobs for k8s 1.25. 

Running GitHub actions after opening the PR.

### DIFF
--- a/hack/toolchain.sh
+++ b/hack/toolchain.sh
@@ -35,6 +35,15 @@ kubebuilder() {
     arch=$(go env GOARCH)
     ln -sf $(setup-envtest use -p path "${K8S_VERSION}" --arch="${arch}" --bin-dir="${KUBEBUILDER_ASSETS}")/* ${KUBEBUILDER_ASSETS}
     find $KUBEBUILDER_ASSETS
+
+    # Install latest binaries for 1.25.x (contains CEL fix)
+    if [[ "${K8S_VERSION}" = "1.25.x" ]] && [[ "$OSTYPE" == "linux"* ]]; then
+        for binary in 'kube-apiserver' 'kubectl'; do
+            rm $KUBEBUILDER_ASSETS/$binary
+            wget -P $KUBEBUILDER_ASSETS dl.k8s.io/v1.25.16/bin/linux/${arch}/${binary}
+            chmod +x $KUBEBUILDER_ASSETS/$binary
+        done
+    fi
 }
 
 main "$@"


.github/workflows/presubmit.yaml
```diff
@@ -17,6 +17,14 @@ jobs:
     - uses: ./.github/actions/install-deps
       with:
         k8sVersion: ${{ matrix.k8sVersion }}
+    - name: Install kube-apiserver (k8s v1.25 only)
+      # Install kube-apiserver v1.25.16
+      if: ${{ matrix.k8sVersion == '1.25.x' }}
+      run: |
+          tempout=$(mktemp)
+          wget -P "$tempout" dl.k8s.io/v1.25.16/bin/linux/amd64/kube-apiserver
```
jonathan-innis, line 25: Can this be in the `toolchain.sh` file rather than just in the Github actions? We are going to move away from GHA to prow long-term AND we want to make sure that anyone who is installing the toolchain with `1.25.x` just gets the behavior out-of-the-box.

jonathan-innis, line 25: Can we also make sure we download all of the associated binaries. There should be three that are downloaded with `setup-envtest` (`kubectl`, `etcd`, and `kube-apiserver`)

jmdeal: Is there a need to grab the other binaries or is it just for consistency? My thought was to leave as much as possible envtests default distribution. (plus etcd isn't distributed or versioned with the other binaries)

jmdeal: Discussed offline, now includes `kubectl`. I haven't included `etcd` since setup-envtest already installs the correct version.

jonathan-innis (approved):
LGTM 🚀 Thanks so much! Awesome work here!

Issue #743, opened 2022-09-08
support drain timeout


grosser:
**Tell us about your request**
Sometimes nodes get stuck when a pod could not be drained, so we have to alert ourselves and then manually kill it, which is not ideal.

It would be nice to set an optional `drainTimeout` so we can ensure nodes always die after x hours/days.

**Are you currently working around this issue?**
Helper pod that kills pods on stuck nodes so they finish draining.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

wmgroot:
Just for reference, this is the field that supports this behavior in ClusterAPI.
```
$ kubectl explain machinedeployment.spec.template.spec.nodeDrainTimeout
KIND:     MachineDeployment
VERSION:  cluster.x-k8s.io/v1beta1

FIELD:    nodeDrainTimeout <string>

DESCRIPTION:
     NodeDrainTimeout is the total amount of time that the controller will spend
     on draining a node. The default value is 0, meaning that the node can be
     drained without any time limitations. NOTE: NodeDrainTimeout is different
     from `kubectl drain --timeout`
```

In our case, we set this value to `24h` in all clusters, production or otherwise. We make it clear to our users that they must tolerate graceful disruption of their pods, and to alert themselves if their pods are in an unhealthy state for a long period of time in production. This allows us to ensure we can keep up to date with security patches across all of our clusters.

We've discussed lowering the value from 24h to something much shorter once we have a better story for automated traffic failover between our clusters.

wmgroot:
Should support for volume detach timeouts be opened as another issue? We use EBS CSI and find that some volumes refuse to detach due to timeouts that aren't handled well by the EBS CSI controller. We've opened several issues there, but these problems still occur after years of using EBS CSI. This blocks removal of nodes and can greatly delay the process of updating nodes in a cluster if left to manual intervention.
```
$ kubectl explain machinedeployment.spec.template.spec.nodeVolumeDetachTimeout --context mgmt-use2
KIND:     MachineDeployment
VERSION:  cluster.x-k8s.io/v1beta1

FIELD:    nodeVolumeDetachTimeout <string>

[...]

Issue #758, opened 2023-10-31
v1 Laundry List


jonathan-innis:
### Description

This issue contains a list of breaking API changes that we want to make for v1.

Linked Cloud Provider v1 Laundry Lists
- AWS: https://github.com/aws/karpenter/issues/5006

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
Align the `karpenter_interruption_actions_performed` values with the `karpenter_disruption_actions_performed` values. Currently the values are different and (since these components have similar names) we should make them similar

njtran:
Drop support for do-not-consolidate and do-not-evict annotations from v1alpha5.

garvinp-stripe:
I think having finalizer in nodeclaim make sense that way as operators we can force the node to terminate outside of the finalizer when necessary. https://github.com/aws/karpenter-core/issues/796 / https://github.com/aws/karpenter-core/issues/743

garvinp-stripe:
Can we move the Karpenter docs from aws provider github to the k8s sigs Karpenter?

Bryce-Soghigian:
@garvinp-stripe  there is an ongoing discussion on this https://github.com/kubernetes-sigs/karpenter/issues/832

jonathan-innis:
We should consider requiring `apiVersion` and `kind` in the `nodeClassRef` so that we can dynamically look-up the NodeClass from the NodePool in the neutral code.

jonathan-innis:
Consider changing the `apiVersion` field in the `nodeClassRef` to `group` to match what is done in [Gateway API](https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1.RouteGroupKind). Referencing references by `apiVersion` is also not useful here since the reference is only going to be reasoned about in terms of a single version anyways. We really only need the `group` parameter here.

sftim:
We should check that the v1 API will work with `InPlacePodVerticalScaling` if / when that becomes stable
https://github.com/kubernetes-sigs/karpenter/issues/829

sftim:
[...]

Issue #655, opened 2023-11-01
Termination leaks pods that tolerate the disruption taint that Karpenter will not evict


njtran:
### Description

**Observed Behavior**:
Karpenter should terminate and evict all pods scheduled to a node when terminating a Node/NodeClaim. 

Karpenter skips eviction of [pods that will reschedule](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/node/termination/terminator/terminator.go#L93-L100), meaning that those pods are not evicted and cleaned up prior to instance deletion. After Karpenter successfully sends a delete request for the instance, initiating kubelet shutdown, it will immediately after remove the finalizer, thus deleting the node. This race can result in Kubelet failing to evict daemonsets and failing to deregister itself. These pods are then leaked, relying on garbage collection to clean them up. 

**Expected Behavior**:
All pods are deleted and evicted from the node, then the kubelet is able to fully gracefully delete with no leaks.

This can be solved with #621 by adding a NoExecute Taint to nodes, allowing TaintManagerEviction to delete pods.
Additionally, Karpenter should wait to remove the node finalizer until the underlying CP instance is Terminated (as opposed to Terminating). 

**Versions**:
- Chart Version: 
- Kubernetes Version (`kubectl version`): 

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #834, opened 2023-12-01
docs: RFC for disruption.terminationGracePeriod feature

wmgroot:
Design for #743 
Implementation PR: https://github.com/kubernetes-sigs/karpenter/pull/916

**Description**
RFC for support for `NodeClaim.spec.terminationGracePeriod` (nodeDrainTimeout)

### DIFF
--- a/designs/nodeclaim-termination-grace-period.md
+++ b/designs/nodeclaim-termination-grace-period.md
@@ -1,0 +1,167 @@
+# NodeClaim Termination Grace Period
+
+## Motivation
+Users are requesting the ability to control how long karpenter will wait for the deletion of an individual node to complete before forcefully terminating the node regardless of pod status. [#743](https://github.com/kubernetes-sigs/karpenter/issues/743). This supports two primary use cases.
+* Cluster admins who want to ensure that nodes are cycled after a given period of time, regardless of user-defined disruption controls (such as PDBs or preStop hooks) that might prevent eviction of a pod beyond the configured limit. This could be to satisfy security requirements or for convenience.
+* Cluster admins who want to allow users to protect long-running jobs from being interrupted by node disruption, up to a configured limit.
+
+This design adds in a new `terminationGracePeriod` field to the `NodeClaim` resource. The field denotes the maximum time a node will wait for graceful termination of its pod before being forcefully torn down.
+This new field is immutable on each NodeClaim, and requires cycling nodes in order to change the value.
+
+## Proposed Spec
+
+```yaml
+apiVersion: karpenter.sh/v1beta1
+kind: NodeClaim
+spec:
+  terminationGracePeriod: 24h   # metav1.Duration, nil if not set
+```
+
+## Code Definition
+
+```go
+type NodeClaimSpec struct {
+  {...}
+	// TerminationGracePeriod is the maximum duration the controller will wait before forcefully deleting the pods on a node, measured from when deletion is first initiated.
+	//
+	// Warning: this feature takes precedence over a Pod's terminationGracePeriodSeconds value, and bypasses any blocked PDBs or the karpenter.sh/do-not-disrupt annotation.
+	//
+	// This field is intended to be used by cluster administrators to enforce that nodes can be cycled within a given time period.
+	// When set, drifted nodes will begin draining even if there are pods blocking eviction. Draining will respect PDBs and the do-not-disrupt annotation until the TGP is reached.
+	//
+	// Karpenter will preemptively delete pods so their terminationGracePeriodSeconds align with the node's terminationGracePeriod.
+	// If a pod would be terminated without being granted its full terminationGracePeriodSeconds prior to the node timeout,
+	// that pod will be deleted at T = node timeout - pod terminationGracePeriodSeconds.
+	//
+	// The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
+	// If left undefined, the controller will wait indefinitely for pods to be drained.
+	// +kubebuilder:validation:Pattern=`^([0-9]+(s|m|h))+$`
+	// +kubebuilder:validation:Type="string"
+	// +optional
+	TerminationGracePeriod *metav1.Duration `json:"terminationGracePeriod,omitempty"`
+}
+```
+
+## Alternative Spec
+Rather than defining the `terminationGracePeriod` on the NodeClaim spec, it could be added to the NodePool disruption block instead.
+
+```yaml
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec:
+  disruption:
+    terminationGracePeriod: 24h   # metav1.Duration, nil if not set
+```
+
+#### Benefits
+1. Making updates to the `terminationGracePeriod` in a NodeClaim template is cumbersome and disruptive to users of the cluster since it requires all nodes to be cycled. The NodePool disruption block can be considered mutable and allow cluster admins to change the value without replacing all nodes.
+
+#### Drawbacks
+1. Changes to the `terminationGracePeriod` would be more surprising, because current live nodes would be impacted, and might result in immediate deletion of pods bypassing PDBs if the grace period is too low or pod terminationGracePeriodSeconds is too high.
+2. Behavior would no longer match the behavior of a Pod's `terminationGracePeriodSeconds`, which is immutable and can only be changed with a full pod rollout.
+3. Users of NodeClaims without NodePools would not be able to take advantage of this feature.
+
+## Validation/Defaults
+The `terminationGracePeriod` fields accepts a common duration string which defaults to a value of `nil`. Omitting the field results in the default nil value instructing the controller to wait indefinitely for pods to drain gracefully, maintaining the existing behavior.
+A value of 0 instructs the controller to evict all pods by force immediately, behaving similarly to `pod.spec.terminationGracePeriodSeconds=0`.
+
+## Prior Art
+This is the field from CAPI's MachineDeployment spec which implements similar behavior.
+```
+$ kubectl explain machinedeployment.spec.template.spec.nodeDrainTimeout
+KIND:     MachineDeployment
+VERSION:  cluster.x-k8s.io/v1beta1
+
+FIELD:    nodeDrainTimeout <string>
+
+DESCRIPTION:
+     NodeDrainTimeout is the total amount of time that the controller will spend
+     on draining a node. The default value is 0, meaning that the node can be
+     drained without any time limitations. NOTE: NodeDrainTimeout is different
+     from `kubectl drain --timeout`
+```
+
+```
+$ kubectl explain pod.spec.terminationGracePeriod
+KIND:     Pod
+VERSION:  v1
+
+FIELD:    terminationGracePeriod <integer>
+
+DESCRIPTION:
+     Optional duration in seconds the pod needs to terminate gracefully. May be
+     decreased in delete request. Value must be non-negative integer. The value
+     zero indicates stop immediately via the kill signal (no opportunity to shut
+     down). If this value is nil, the default grace period will be used instead.
+     The grace period is the duration in seconds after the processes running in
+     the pod are sent a termination signal and the time when the processes are
+     forcibly halted with a kill signal. Set this value longer than the expected
+     cleanup time for your process. Defaults to 30 seconds.
+```
+
+## Implementation
+
+### Termination Grace Period Expiration Detection
+Because Karpenter already ensures that nodes and their nodeclaims are in a deleting state before performing a node drain during node deletion, we are able to leverage existing `deletionTimestamp` fields to avoid the need for an addition annotation or other tracking label.
+We'll use the NodeClaim's `deletionTimestamp` specifically to avoid depending on the Node's `deletionTimestamp`, because [Kubernetes docs](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/) recommend draining a node *before* deleting it, which is counter to how Karpenter behaves today (relying on Node finalizers to handle cleanup).
+
+### Termination Grace Period Expiration Behavior
+1. NodeClaim deletion occurs (user initiated via Node finalizer,or when Karpenter begins node rotation from drift detection, etc).
+2. NodeClaim deletionTimestamp is set.
+3. Standard node drain process begins.
+4. Drain attempts fail (due to PDBs, preStop hooks, etc). Karpenter adds an Event to the Node noting that it will be forcibly terminated at X time.
+5. Reconciliation of the Node/NodeClaim is requeued (potentially many times).
+6. Time passes, and the configured terminationGracePeriod for the NodeClaim is exceeded.
+7. Any remaining pods are forcibly drained, ignoring eviction (bypassing PDBs, preStop hooks, pod terminationGracePeriod, etc).
+8. The standard Karpenter reconciliation loop for the termination controller succeeds, and the node is formally removed from the cloud provider.
+
+### Eventual Disruption
+Karpenter automatically excludes nodes containing do-not-disrupt annotated pods or pods with blocked PDBs from being considered as candidates for disruption. Implementation of this feature will require changing that behavior to allow disruption of those nodes.
+
+### Optimizing Pod Deletion Behavior When Eviction is Blocked
+Some pod protection semantics block or delay termination of pods, which could result in a pod being forcibly evicted without any chance for graceful shutdown while using this feature.
+
+Consider the following scenario.
+1. A Deployment has 3 replicas, each with a terminationGracePeriodSeconds of 10m.
+2. Each of these pods requires the full 10m to terminate gracefully.
+3. The Deployment has an associated PDB with maxUnavailable 1 requiring that only one of these pods be disrupted at a time.
+
+1. The cluster admin has configured NodePool.spec.disruption.terminationGracePeriod at 15m.
+2. A full node reconcilliation has started which has caused all nodes which contain the pods above to enter a deleting state at the same time.
+
+During the standard node deletion process, k8s would:
+1. Evict pod A only, which takes 10m of the available 15m to terminate gracefully. Pod B and C are not evicted, due to their PDB only allowing 1 disruption at a time.
+2. After pod A starts up healthy on a new node, pod B is evicted, but has only 5m of the needed 10m to terminate. Both pods B and C are evicted before graceful termination could complete, and pod C never received a termination signal.
+
+In order to improve this case for users of the cluster, we can assume that a pod's terminationGracePeriodSeconds is a good estimate of the time it takes to terminate gracefully and start the eviction process early if the pod would not be able to terminate within its terminationGracePeriodSeconds when a node timeout is looming.
+Pods that are deleted in this way should have a clear event log of what deleted their pod (karpenter) and why (current time > node terminationGracePeriod - pod terminationGracePeriod).
+
+Risks
+1. This behavior could be confusing to cluster administrators, and could result in immediate pod evictions when a NodePool terminationGracePeriod is set, even to a value of hours.
+2. This behavior could be confusing to users, who may have set their pod terminationGracePeriodSeconds to higher than necessary values and have their pods suddenly deleted when a node begins deletion.
+
+### Documented Warnings
+This feature carries some risk for cluster administrators given that it can bypass the configured grace periods and PDBs set by users of a cluster. These risks should be called out explicitly in the docs for the feature to minimize the risk of a cluster administrator unintentionally disrupting their users' workloads.
+1. Enabling this feature by setting NodePool.spec.template.terminationGracePeriod to any value will result in a drift reconciliation that will replace all nodes. The new terminationGracePeriod value will not be active until new NodeClaims have been provisioned.
+2. Enabling this feature can result in termination of user pods, violating their configured grace periods or PDBs. If a pod has a terminationGracePeriodSeconds greater than the NodeClaim's TGP, it will be deleted as soon as the Node begins disruption, effectively limiting a pod's maximum TGPS to the NodeClaim's TGP.
+3. Because of the optimization stated above, user pods could be terminated in violdation of their PDBs **before** the configured NodeClaim grace period is reached.
+
+## Manual Intervention
+If needed, a cluster-admin can prematurely forcefully terminate a node by starting a `kubectl delete node` followed by `kubectl drain --disable-eviction=true`.
+This is already possible today with Karpenter, and means that we shouldn't need to consider manual forced termination as part of this design.
+
+```
+$ kubectl drain my-node --grace-period=0 --disable-eviction=true
+
+$ kubectl drain --help
+Options:
+    --grace-period=-1:
+	Period of time in seconds given to each pod to terminate gracefully. If negative, the default value specified
+	in the pod will be used.
+
+    --disable-eviction=false:
+	Force drain to use delete, even if eviction is supported. This will bypass checking PodDisruptionBudgets, use
+	with caution.
+```


designs/disruption-controls-drain-timeout.md
```diff
@@ -0,0 +1,78 @@
...
+     NodeDrainTimeout is the total amount of time that the controller will spend
+     on draining a node. The default value is 0, meaning that the node can be
+     drained without any time limitations. NOTE: NodeDrainTimeout is different
+     from `kubectl drain --timeout`
+```
+
+## Implementation
+
+### Force Drain Detection
+If forceDrainAfter is set, I think it should be straightforward to have the node termination controller add a timestamp as an annotation on the Node resource denoting when the Node will be force drained at a specific time (is RFC-3339 the preferred standard?)
```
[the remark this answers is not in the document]
garvinp-stripe: One more note is if we move the finalizer out if node then the above comment won't work https://github.com/kubernetes-sigs/karpenter/issues/758#issuecomment-1790931682 since deletion won't happen on the node first -> deletion timestamp won't exist on the node. However we likely won't just consult the nodeclaim if this migration happens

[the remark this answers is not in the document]
wmgroot: I did want to avoid the use of an extra annotation for state tracking, but this is why I didn't think `deletionTimestamp` would work.

```
$ kubectl get node ip-10-117-32-4.us-east-2.compute.internal -o yaml | grep deletionTimestamp
<empty>

$ kubectl drain ip-10-117-32-4.us-east-2.compute.internal --delete-emptydir-data --ignore-daemonsets
node/ip-10-117-32-4.us-east-2.compute.internal already cordoned
...
<all pods cleanly evicted>
...
node/ip-10-117-32-4.us-east-2.compute.internal drained

$ kubectl get node ip-10-117-32-4.us-east-2.compute.internal -o yaml | grep deletionTimestamp
<empty>

<node is reaped by node autoscaler for being unschedulable and empty>
```
`deletionTimestamp` isn't set until the Node resource is deleted, which doesn't occur during a normal drain.

1. Does Karpenter always perform a Node deletion as part of its node deprovisioning process?
2. If so, does Karpenter perform the Node deletion at the start of the deprovisioning process, or it is only performed after all pods have been successfully drained from the node?

jackfrancis:
@wmgroot thank you for letting me pollute your PR thread w/ some prow tests, just testing out some recent additions, nothing for you to worry about here :)

designs/disruption-controls-drain-timeout.md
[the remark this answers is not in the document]
garvinp-stripe: https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/#understanding-finalizers about deletion with finalizers 
```
Kubernetes will report back that the object has been deleted, however, it hasn’t been deleted in a traditional sense. Rather, it’s in the process of deletion. When we attempt to get that object again, we discover the object has been modified to include the deletion timestamp.```

File: designs/disruption-controls-drain-timeout.md [hunk 2]
```diff
@@ -0,0 +1,78 @@
+# Disruption Controls - (in v1beta1+)
```
ellistarn, line 1: Thoughts on integrating with https://kubernetes.io/blog/2022/12/16/kubernetes-1-26-non-graceful-node-shutdown-beta/#how-does-it-work

designs/disruption-controls-drain-timeout.md
```diff
@@ -0,0 +1,78 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec: # This is not a complete NodePool Spec.
+  disruption:
+    consolidationPolicy: WhenUnderutilized || WhenEmpty
+    consolidateAfter: 10m || Never # metav1.Duration
+    expireAfter: 10m || Never # Equivalent to v1alpha5 TTLSecondsUntilExpired
+    forceDrainAfter: 24h || Never
```
ellistarn, line 22: re: 
```
$ kubectl drain my-node --grace-period=0 --disable-eviction=true
```

thoughts on spec.disruption.gracePeriod? "The disruption grace period".

designs/disruption-controls-drain-timeout.md
```diff
@@ -0,0 +1,78 @@
...
+$ kubectl explain machinedeployment.spec.template.spec.nodeDrainTimeout
+KIND:     MachineDeployment
+VERSION:  cluster.x-k8s.io/v1beta1
+
+FIELD:    nodeDrainTimeout <string>
+
+DESCRIPTION:
+     NodeDrainTimeout is the total amount of time that the controller will spend
+     on draining a node. The default value is 0, meaning that the node can be
+     drained without any time limitations. NOTE: NodeDrainTimeout is different
```
ellistarn, line 51: This `NOTE` suggests that there has been some confusion with this semantic. Any learnings here @jackfrancis?

designs/disruption-controls-drain-timeout.md
wmgroot: This looks promising, it looks like it also handles the volume detachment case too.

Once the timeout is reach, I think karpenter would need to set the taint, then watch for all pods to be forcibly terminated, then clean up the node on the cloud provider's side.

designs/disruption-controls-drain-timeout.md
wmgroot: The `grace-period=0` is actually not the necessary part of the drain command fwiw, the `disable-eviction` is the real lifter there. But I think that arg name is pretty unintuitive.

I'm not attached to a particular name and think something like `terminationGracePeriod` would be intuitive for users.

designs/disruption-controls-drain-timeout.md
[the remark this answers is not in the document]
wmgroot: I think the standard best practice for K8s node cleanup is to completely drain the Node before any deletion occurs.
https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/

I think that agrees with the comment you linked above?
> Minimally, we need to stop deleting the node and allow the kubelet to deregister as part of our standard termination workflow, as it's currently causing daemons to leak.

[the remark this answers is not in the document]
garvinp-stripe: Agreed on thats what we should do however that isn't what Karpenter is doing atm,  until the drain happens at nodeclaim first then node. I believe this isn's something that's settled on yet. So this depends on how we want to implement this to match the future of Karpenter. It would be good to know if Karpenter will end up migrating the finalizer. (I am ok with annotation for now).

Ideally I think it shouldn't be an annotation but rather a status on the nodeclaim.

designs/disruption-controls-drain-timeout.md
ellistarn: TerminationGracePeriod I like. There's also `metadata.DeletionGracePeriodSeconds` in every Kubernetes object. In general, we've been using the term disruption, rather than deletion or termination, which is why I suggested `disruption.gracePeriod`. That said, this semantic is a bit awkward with the `Never` bit of our NillableDurations.

njtran: Thoughts on having this be drainTimeout or terminationTimeout? We've been using the -After syntax to indicate a time when a node should be disrupted. I think it can apply here as well, but I think a timeout is a little more deliberate in the meaning and usage.

Alternatively, I like the GracePeriodSeconds + its derivatives.

designs/disruption-controls-drain-timeout.md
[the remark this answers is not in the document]
njtran: The finalizer should be both on the node and nodeclaim. Whichever receives the deletion request, either by the user or by Karpenter, should always trigger the deletion behavior. The [termination code](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/controller.go#L75) requires this deletion timestamp to be set to know when a delete request has been made.  

I don't see a reason why we can't rely on the deletion timestamp here. If there's a debate on whether we should rely on the nodeclaim's or node's deletion timestamp, Karpenter deletes the nodeclaim to then delete the node, whereas a user could do either. Since both nodeclaim and node deletion trigger deletion of the other linked resource, I could imagine picking one of the two as the source of truth (which I would argue we should do NodeClaim, since it's the resource that Karpenter chooses to represent both the node and underlying machine).

designs/disruption-controls-drain-timeout.md
```diff
@@ -0,0 +1,78 @@
...
+     from `kubectl drain --timeout`
+```
+
+## Implementation
+
+### Force Drain Detection
+If forceDrainAfter is set, I think it should be straightforward to have the node termination controller add a timestamp as an annotation on the Node resource denoting when the Node will be force drained at a specific time (is RFC-3339 the preferred standard?)
+
+When a drain fails and would be requeued to be attempted again, if the current time is after the timestamp of the forceDrainAt annotation, a requeue should be skipped and the force drain logic should be triggered instead.
```
njtran, line 60: I'm not sure I'm understanding this bit about requeueing. If a node has begun draining (i.e. begun terminating), then Karpenter treats this as past the point of no return. We'll keep trying indefinitely to evict the pods, and then once the timeout is reached, we'll delete the pods (not using the evictionAPI, thus not respecting PDBs), causing the pods to be terminated and the node eventually drained. 

One bit I think that we should include here is what the drain timeout means. Does it mean that after the timeout, if the node isn't fully drained, that we begin deleting all pods, respecting each of those pods grace period seconds? If not, should we proactively begin terminating pods in accordance with their grace period seconds so that they match up with the grace period timeout? I'm of the opinion that it should be the former, but curious to hear other opinions.

designs/disruption-controls-drain-timeout.md
```diff
@@ -0,0 +1,78 @@
...
+    --grace-period=-1:
+	Period of time in seconds given to each pod to terminate gracefully. If negative, the default value specified
+	in the pod will be used.
+
+    --disable-eviction=false:
+	Force drain to use delete, even if eviction is supported. This will bypass checking PodDisruptionBudgets, use
+	with caution.
+```
+
+It might sufficient to skip straight to termination of the node at the cloud provider rather than waiting for a successful force drain to complete.
```
njtran, line 78: The problem with this is similar to https://github.com/kubernetes-sigs/karpenter/issues/655. You'll get leaked pods, which is a hefty consideration if we decide to go this way.

njtran (commented):
Nice design! Thanks for writing this out. Eager to talk more about this.

designs/disruption-controls-drain-timeout.md
garvinp-stripe: I think keeping `grace` is more clear because this isn't a timeout on termination necessarily nor does it need to be `drain` related? It is simply indicating a graceful period of time for termination was exceed and now we switch modes to ungracefully terminating the node and its pods? 

Grammatically I think we got forced into a corner if we have to stick with `Never`? Can it be something like `disruption.graceTimeout`

wmgroot: We can always disallow `Never` and have 0 refer to an unlimited grace period like CAPI does if it helps avoid awkwardness.

designs/disruption-controls-drain-timeout.md
wmgroot: > We'll keep trying indefinitely to evict the pods

My understanding is that this indefinite retry behavior is handled by a requeue mechanism when a drain fails, but I could be misunderstanding the controller's current behavior.

> Does it mean that after the timeout, if the node isn't fully drained, that we begin deleting all pods, respecting each of those pods grace period seconds?

No, a user can define an arbitrarily high `terminationGracePeriodSeconds` for their pods. In our use cases, users set the value to 24h and include a preStop hook that attempts to clean up and terminate long running jobs to minimize disruption during node rollouts. We use the node's `terminationGracePeriodSeconds` as a hard-stop mechanism that cannot be avoided. I think this behaves similarly to the pod's `terminationGracePeriodSeconds`, where you cannot do something like add a sleep greater than the grace period to prolong the life of the pod; it's a hard limit.

designs/disruption-controls-drain-timeout.md
wmgroot: Updated and added the non-graceful node shutdown doc as a consideration in the behavior implementation section.

designs/disruption-controls-drain-timeout.md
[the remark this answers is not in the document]
wmgroot: I think using the NodeClaim's `deletionTimestamp` should be sufficient.

designs/disruption-termination-grace-period.md
```diff
@@ -0,0 +1,123 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec: # This is not a complete NodePool Spec.
+  disruption:
+    consolidationPolicy: WhenUnderutilized || WhenEmpty
+    consolidateAfter: 10m || Never # metav1.Duration
+    expireAfter: 10m || Never # Equivalent to v1alpha5 TTLSecondsUntilExpired
+    terminationGracePeriodSeconds: 86400 || 24h || nil
```
garvinp-stripe, line 22: if we are specifying seconds we shouldn't allow 24h?

wmgroot: Yeah, you're correct. I'm going to update it to support standard duration formats for convenience so we don't have to google how many seconds are in a day all the time.

designs/disruption-controls-drain-timeout.md
njtran: So essentially this defines the contract for maximum length of drain to be `node.terminationGracePeriodSeconds + max(pod.terminationGracePeriodSeconds)`, is that right?

designs/disruption-controls-drain-timeout.md
njtran: Daemonsets tolerating all taints I see as a separate issue. Many daemonsets do this because it's the easiest way to ensure that the pods are running as long as possible, without having to think about all the different addons that may add taints. Whichever ones terminate all taints will either need to be okay with being leaked or we may need to respect their terminationGracePeriodSeconds as well.

designs/disruption-termination-grace-period.md
njtran: What about terminationGracePeriodDuration? Would make it easier to do the thing you're trying to do.

designs/disruption-controls-drain-timeout.md
garvinp-stripe: Thats a good question. I think at this time the maximum length of a drain is just `node.terminationGracePeriodSeconds` and the assumption is within `node.terminationGracePeriodSeconds` all pods should be drained? (At least from what I understood)

wmgroot: Yes, @garvinp-stripe has the correct understanding. A pod's `terminationGracePeriodSeconds` is not considered.

If a pod had a `terminationGracePeriodSeconds` of 2 hours, and the node's `terminationGracePeriod` was 1 hour, deletion of the node could cause a pod to be forcibly terminated an hour before the pod's `terminationGracePeriodSeconds` would expire.

designs/disruption-termination-grace-period.md
garvinp-stripe: > What about terminationGracePeriodDuration? Would make it easier to do the thing you're trying to do.
Happy with that or just terminationGracePeriod

designs/disruption-termination-grace-period.md
```diff
@@ -0,0 +1,122 @@
...
+    expireAfter: 10m || Never # Equivalent to v1alpha5 TTLSecondsUntilExpired
+    terminationGracePeriod: 24h || nil # metav1.Duration
+```
+
+```
+$ kubectl explain nodepool.spec.disruption.terminationGracePeriod
+KIND:     NodePool
+VERSION:  v1beta1
+
+FIELD:    terminationGracePeriod <integer>
```
garvinp-stripe, line 30: would it be a string instead since it support duration

garvinp-stripe:
Overall looks good to me

designs/disruption-termination-grace-period.md
wmgroot: I think I slightly prefer `terminationGracePeriod` just for brevity, but I don't have very strong opinions.

designs/disruption-termination-grace-period.md
```diff
@@ -0,0 +1,122 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+spec: # This is not a complete NodePool Spec.
+  disruption:
+    consolidationPolicy: WhenUnderutilized || WhenEmpty
+    consolidateAfter: 10m || Never # metav1.Duration
+    expireAfter: 10m || Never # Equivalent to v1alpha5 TTLSecondsUntilExpired
+    terminationGracePeriod: 24h || nil # metav1.Duration
```
ellistarn, line 22: Since we already are scoped to disruption, I'd suggest dropping the termination bit. We typically try to design the API such that you can speak the dotted expansion of the config, E.g.

disruption.gracePeriod is the "node pool's disruption grace period" 

disruption.terminationGracePeriod is nonsensical to speak, and intermingled concepts of disruption and termination.

wmgroot: The disruption field isn't limited to node termination behavior. I think it's reasonable and leaves no room for misunderstanding to keep the full name `disruption.terminationGracePeriod`.

designs/disruption-controls-drain-timeout.md
njtran: > If a pod had a terminationGracePeriodSeconds of 2 hours, and the node's terminationGracePeriod was 1 hour, deletion of the node could cause a pod to be forcibly terminated an hour before the pod's terminationGracePeriodSeconds would expire.

Does this consider the case where we have a pod that never receives an eviction request during the terminationGracePeriod for the node? For example: (1) pods with blocking PDBs, (2) do-not-disrupt pods, or (3) pods that are opted out of termination by default. In this 1 hour period, these pods would never receive a delete call, which means that as soon as forceful termination kicks in, these pods are _never_ given a chance to gracefully terminate. 

This is the core question I am still undecided about: what should the maximum time a node is terminating be? I can imagine taking a stance of being application-safe, by saying that it's `node.terminationGracePeriodSeconds + max(pod.terminationGracePeriodSeconds)`, but can also imagine that users who don't want to allow applications to extend the lifetime of a node to say it just means `node.terminationGracPeriodSeconds`, which may result in the pod terminationGracePeriodSeconds a best effort setting. in the case of spot instance reclamation, this is how it's treated.

designs/disruption-termination-grace-period.md
njtran: Disruption encompasses both the disruption and termination controllers, which a disruptionGracePeriod in the context of the disruption controller doesn't really make much sense. Prefacing with `termination` makes it more explicit in what process it's referencing. WDYT @ellistarn ?

designs/disruption-controls-drain-timeout.md
wmgroot: That's an interesting point. At my company we handle that case by preemptively identifying situations where Deployments and other resources are at their "PDB threshold" for extended periods of time, which is considered unusual behavior. You'd want to be checking for those situations as a workload maintainer in K8s regardless of this feature existing or not. You also have to tolerate the case where pods are terminated without a proper SIGTERM, because of power loss scenarios and other extenuating cases.

I think we can put some guidance in the docs for this feature to caution cluster admins that applications with startup times longer than the `terminationGracePeriod` could be at high risk for eviction without proper graceful pod termination every time nodes are cycled.

As the entire point of this feature is to be a hard limit that cannot be avoided from the user side, I think any guarantee based on a pod's `terminationGracePeriod` is a non-starter.

designs/disruption-termination-grace-period.md
ellistarn: Makes me wonder if it should even be part of disruption, then? What other fields in disruption control termination?

designs/disruption-controls-drain-timeout.md
garvinp-stripe: To add to this point. I agree this is dangerous however there is a risk of letting users "set" a termination deadline and I think its a more problematic approach. We give users a very specific SLA on how long a node will provide a graceful termination period before we hard shut down the node. The reason for this is we have compliance standards to meet. For example, its pretty common for companies to enforce daily or weekly node refresh to minimize, in the case of a compromised user application, the time a compromised application has access to the node. It also allows us to enforce an SLA on when nodes will get updates. 

IMO, allowing cluster operators to manage this risk rather than not is a better approach. I think its unreasonable to let application users to dictate a host's graceful termination and instead cluster operator should set SLA for users on their platform.  We definitely should clearly state that this hard termination period ignores any pod level termination period for those who don't operate the same way.

designs/disruption-termination-grace-period.md
jonathan-innis, line 22: Is there a reasonable default to get set here? `24h` or something shorter?

jonathan-innis: > Makes me wonder if it should even be part of disruption, then

Agree with @ellistarn. Is this just a property of the NodeClaim? Can we imagine that if you initiate a deletion request on a NodeClaim as a standalone item that it just deprovisions on its own? I'd argue, yes. I wonder if (similar to pod) this fits best in `spec.template.spec.terminationGracePeriod`?

designs/disruption-controls-drain-timeout.md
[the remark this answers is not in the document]
jonathan-innis: > The finalizer should be both on the node and nodeclaim

Having the finalizer on the NodeClaim gives us the benefit of removing the finalizer on the Node later. We've talked a little bit about potentially removing it since it's caused some node hanging in cases where the karpenter controller has been removed. I'm not fully convinced that we want to remove the node finalizer, but I do think there's quite a bit of value in pushing more detail into the NodeClaim management object over the Node itself

designs/disruption-termination-grace-period.md
wmgroot: Whether or not it's a NodeClaim property depends on the behavior we'd like to see. If you want to be able to lower the `terminationGracePeriod` in order to force deletion of live nodes, you'd want it to not be a NodeClaim field, since that wouldn't be changed if the NodePool's `terminationGracePeriod` was updated. Are NodeClaim's mutuble? Would a cluster admin be able to modify the NodeClaim for a Node in order to force early termination of a Node if desired? Maybe it's sufficient for the cluster admin to have to perform a force drain on a terminating Node if they want to force early termination.

In my company, we run 24h `terminationGracePeriod`'s in production, and want the ability to forcefully tear down the Node earlier if desired. Seems like that would still be possible if it's a field on the NodeClaim via `kubectl drain`, but I just wanted to lay out the implications of one choice over another.

wmgroot: No, the default should be "never terminate" (nil). If we set a default to any value, that would be surprising new behavior for Karpenter users who are not aware this capability exists. It requires cluster admins to communicate platform expectations on when their users can expect forced pod disruption that could violate their users' PDBs.

jonathan-innis: > Are NodeClaim's mutuble

jonathan-innis: > Are NodeClaim's mutuble

Yes, NodeClaims are mutable; though, mutating certain fields on them may not have the behavior that's expected today since mutating some things like the labels/annotations won't necessarily have any affect on Node that they own.

 > you'd want it to not be a NodeClaim field, since that wouldn't be changed if the NodePool's terminationGracePeriod was updated

One option here is to let this field be driftable, if a NodePool's value for this changes, then all NodeClaims that are owned by that NodePool are drifted.

designs/disruption-termination-grace-period.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+## Code Definition
+
+```go
+type Disruption struct {
+    {...}
+    // terminationGracePeriod is a nillable duration, parsed as an integer duration of number of seconds.
+    // A nil value means there is no timeout before pods will be forcibly evicted during node deletion.
+    terminationGracePeriod *NillableDuration `json:"terminationGracePeriod" hash:"ignore"`
```
jonathan-innis, line 50:
```suggestion
    terminationGracePeriod *metav1.Duration `json:"terminationGracePeriod" hash:"ignore"`
```

Seems like this could just be a standard `metav1.Duration` if we don't need the `Never` value

designs/disruption-termination-grace-period.md
wmgroot: I kinda dislike that updating a NodePool could both have and not have large implications for all the NodeClaims that belong to it. 😅 Might be worth trying to aim for consistency of behavior there to avoid surprises for Karpenter users.

Even if the field is immutable on the NodeClaim, I think a viable path exists for a cluster-admin to force termination of the Node via `kubectl drain`. It's the same path that already exists for Karpenter users today too. But I'm interested in hearing what others think about what the expectations should be here.
1. Updating NodePool's `terminationGracePeriod` immediately changes the timeout for all existing Nodes that belong to the NodePool, which could result in mass termination of Nodes (possibly accidentally). Extension question, can this termination bypass the `nodepool.spec.disruption.budgets.nodes` value configured for the NodePool?
2. Updating the NodePool's `terminationGracePeriod` only changes the `terminationGracePeriod` for subsequently created NodeClaims and Nodes.

designs/disruption-termination-grace-period.md
```diff
@@ -0,0 +1,122 @@
...
+     cleanup time for your process. Defaults to 30 seconds.
+```
+
+## Implementation
+
+### Termination Grace Period Expiration Detection
+Because Karpenter already ensures that nodes and their nodeclaims are in a deleting state before performing a node drain during node deletion, we should be able to leverage existing `deletionTimestamp` fields to avoid the need for an addition annotation or other tracking label.
+I believe we should use the NodeClaim `deletionTimestamp` specifically to avoid depending on the Node's `deletionTimestamp`, because [Kubernetes docs](https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/) recommend draining a node *before* deleting it, which is counter to how Karpenter behaves today (relying on Node finalizers to handle cleanup).
+
+If drains are not already periodically requeued, we may need to modify the current drain wait logic to periodically check if a node has exceeded its termination grace period.
```
jonathan-innis, line 98: I think we can drop this line since Karpenter does do the requeueing today, so checking whether we are past the timeout should be pretty straightforward. See: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/controller.go#L98

designs/disruption-termination-grace-period.md
```diff
@@ -0,0 +1,122 @@
...
+
+### Termination Grace Period Expiration Behavior
+1. Node deletion occurs (user initiated, node rotation from drift, etc).
```
jonathan-innis, line 101: To clarify Karpenter's process here: Karpenter initiates a deletion on the NodeClaim first before it initiates the delete of the Node. This might be slightly different if a user initiates the delete on the node directly, but in the automated way, the ordering is:
1. Karpenter initiates NodeClaim deletion
2. Karpenter initiates Node deletion which triggers the drain

designs/disruption-termination-grace-period.md
wmgroot: Agreed, I've confirmed this works as expected in my current PR testing.
https://github.com/kubernetes-sigs/karpenter/pull/916

designs/disruption-termination-grace-period.md
wmgroot: As long as `metav1.Duration` can be nil we should be good to go.

designs/disruption-termination-grace-period.md
jonathan-innis: IMO, if we decided to add this field directly on the NodeClaim, having one field in the NodePool that behaves differently in the `template` block than the other fields in the `template` block is pretty unexpected. Today, if you make a change to the `template` block, `NodeClaims` get marked drifted and you will get nodes rolling to bring your Nodes into the new desired state that you set in the `NodePool`. We are communicating a relationship similar to the Pod/Deployment objects with the NodeClaim/NodePool through the `template` field. If you update the `terminationGracePeriodSeconds` of the pod in the deployment object, you get a pod roll, I'd expect similar behavior here.

designs/disruption-controls-drain-timeout.md
njtran: I can totally understand both of your perspectives here, where as cluster admins, you essentially want an override over application developers who may be acting in bad practice. The difference I think we should call out is that when Karpenter has a choice, it always aims to maximize application availability. 

For instance:
- Disruption for Expiration/Drift/Consolidation won't disrupt/replace a node if it isn't able to re-schedule _all_ of the pods on the node. 
- Karpenter will wait for a replacement node to initialize before deleting nodes so that pod downtime can be minimized. 

Generally, I agree that it would be nice to have a terminationGracePeriod that can be a sure-fire way to know "after x amount of time, I'll have this node cleaned up, regardless of what's on it" but I worry about the implications of this with scheduling. 

Users may need to be more rigorous about the nodepools that have scheduling constraints compatible with their pods but may not have terminationGracePeriods that are compatible with their applications graceful termination period. If they had a pod that was accidentally scheduled to a nodePool with a shorter terminationGracePeriod, the pod could still be running, but would not have enough time to cleanup. Ensuring that the pods always have a say here will ensure that applications clean up with the allotted time they're given, regardless of how they are scheduled. We've made this situation more likely with NodePool weights, where users can define a "preference" of nodePool schedulability, which may not map perfectly with terminationGracePeriod. 

Additionally, not respecting the terminationGracePeriod of pods that want to be the last to run on a node, which may not be included in the core drain function, will never get a chance to terminate, should we choose to implement it like this, which we've thought about implementing with a `NoExecute` taint. Adding a NoExecute taint at the end of the terminationGracePeriod may be a better k8s-native way to handle this logic too. 

I would be curious to know what @sftim 's thoughts are on this thread.

designs/disruption-termination-grace-period.md
njtran: We use NillableDuration if there's a non-nil value that we want the field to default to.

designs/disruption-termination-grace-period.md
sftim: > So, the other line that we are probably drawing here is:
> 
> 1. If we want to be able to make a change to `terminationGracePeriod` that automatically affects all NodeClaims, that field needs to live in the NodePool, outside of the `template` block

That fits, _except_ that we might want to allow changes to the `.spec.template.metadata` of a NodePool, which triggers a metadata change via updates to the NodeClaims that it owns. For example, adding an annotation `whatever.fictional.example/soft-node-drain-after: 2024-01-01T00:00:00+0000` (which no cluster operator would thank you for).

> 2. If we are alright having a rolling update for `terminationGracePeriod` if that value is changed, then I think it's fine in the `template` block

Sure.

---

We should avoid having any annotations or labels that mutate a field that is in the `spec` of a NodeClaim. So 
`soft-node-drain-after` is fine (because it doesn't override `terminationGracePeriod` or any other field), but `terminationGracePeriodAdjustment` would be a poor choice.

sftim:
_Following up on https://github.com/kubernetes-sigs/karpenter/pull/834#discussion_r1438324591_

I can see the benefit of a NodePool-level setting to define grace behavior. I would have an enum with 2 values defined:

- `RespectNodeAndPodGracePeriods`
- `TreatNodeClaimGracePeriodAsDefinitive`

These are wordy but hard to misunderstand. Then, Karpenter uses this setting to influence what `gracePeriodSeconds` to specify for a `DELETE` of the Node, and also uses that for its internal drain-and-finalize behavior.

Cluster operators can serve different developer needs by having different node pools with different settings. If you set `RespectNodeAndPodGracePeriods`, you're allowing anything that can get a pod admitted to delay a node-level operation, and maybe that's OK. Separate mechanisms can limit which namespaces get to use which NodePools.

Whilst the NodePool API is still beta, we could limit the enum to just one value, and have a plan to implement the other enum option ahead of v1 / general availability.

To help people avoid scheduling Pods to nodes that will enforce termination before their app might be ready for that, I think we should advise them to prefer (or require) specific NodePools using `nodeAffinity`, if that outcome is important to them.

sftim:
> Adding a NoExecute taint at the end of the `terminationGracePeriod` may be a better k8s-native way to handle this logic too.

Mmm, yes. I would implement that by defining a field on NodeClaim to specify the duration between:
- the point in time at which this taint appears
- the point at which Karpenter will enforce shutdown / termination

Then, you put that field in your NodePool's template. I think for this one we could have a default / heuristic behavior: the Pod is already supposed to have drained, _and_ any pod that wants to tolerate that taint can specify how long it'd like to tolerate it for.

Does this mean two taints, then?
- one taint to mean “Node is draining for termination, don't schedule here”
- one taint to mean “Node is draining for termination, stop executing now”

or the same taint twice (NoSchedule and NoExecute)?

This also has potential to tie in to [Declarative Node Maintenance](https://kep.k8s.io/4212) if the KEP moves forward. If the declarative node maintenance API were available in a cluster, I think Karpenter should discover and use it (suppressing its built-in drain logic).

designs/disruption-termination-grace-period.md
wmgroot: The first thing the node deletion finalizer does is delete all NodeClaims for the Node that's being deleted, so I think both of those paths should operate identically for the purposes of this feature.
https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/controller.go#L76-L82

designs/disruption-controls-drain-timeout.md
wmgroot: The current implementation PR makes use of the official "out-of-service" NoExecute taint (requiring K8s 1.26 or feature gates). https://github.com/kubernetes-sigs/karpenter/pull/916
https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown

designs/disruption-termination-grace-period.md
wmgroot: Updated to `metav1.Duration`.

designs/disruption-termination-grace-period.md
wmgroot: Updated the expected behavior here to cover both cases (NodeClaim vs Node deleted first).

wmgroot:
> I can see the benefit of a NodePool-level setting to define grace behavior. I would have an enum with 2 values defined:

Rather than an enum with implementation options, could we add a new separate timeout field later for the `RespectNodeAndPodGracePeriods` behavior, such as `NodeClaim.spec.softTerminationGracePeriod`?

I'm not sure I've seen any Karpenter users requesting this style of behavior, and I'm hesitant to complicate the design of this feature for a future addition that isn't something we think we're likely to need.

designs/disruption-termination-grace-period.md
ellistarn: Does modifying a pod's terminationGracePeriodSeconds in a deployment's pod template spec result in a rollout? If not, I could see the case for following suit and special casing drift.

wmgroot: Yes, it does. Just double checked that modifying a pod's terminationGracePeriodSeconds from 30 -> 60 in the Deployment.spec.template.spec did roll the pods.

```
$ kubectl edit deploy -n default
deployment.apps/hello-world-nginx edited

$ kubectl get pod -n default    
NAME                                 READY   STATUS              RESTARTS   AGE
hello-world-nginx-5fd5b65b9d-zpksw   1/1     Running             0          32h
hello-world-nginx-84556d7755-cgzzf   1/1     Running             0          3s
hello-world-nginx-84556d7755-d6sbh   0/1     ContainerCreating   0          1s
```

ellistarn: It's not unreasonable to follow suit from a principle of least surprise perspective. Perhaps nodes of a given spec may need different amounts of time to terminate. I could imagine cluster admins being annoyed at this behavior in the rare case it needs to be changed. I do like the fact it applies to the Node claim, rather than the node pool, as a standalone nodeclaim could have similar drain requirements.

jonathan-innis: > I could imagine cluster admins being annoyed at this behavior in the rare case it needs to be changed

The more I think about this, the more I'm convinced that it makes sense just to one-off this as something that can be updated in-place. I can see updating the value causing a rollout across all nodes as extremely annoying and the value that it gives us is pretty minimal.

Realistically, I'm failing to think of strong sets of use-cases where a user would want to rollout a change to terminationGracePeriod that needs to be closely tied to other changes that they are making to the NodePool at the same time.

jonathan-innis:
> I can see the benefit of a NodePool-level setting to define grace behavior

I'd like to avoid opening up too many options here and just enforcing a single behavior that achieves the persona isolation and forceful drain logic that we think makes the most semantic sense. I'd argue that it's on the cluster admin who is configuring the `terminationGracePeriod` to ensure that this value is above the `terminationGracePeriodSeconds` of pods that it manages OR that there be enough coordination between users and cluster admins to ensure that stateful pods aren't forcefully dropped before they can safely drain. I don't think we should give users tools to shoot themselves in the foot if they can avoid it, but we should also give users enough knobs to accept the risk of what they are doing and do what they need to do to fulfill their requirements.

> Does this mean two taints, then

I think the way we are currently thinking about handling this (could consider making this more cohesive with the [Declarative Node Maintenance KEP](https://kep.k8s.io/4212) going forward) would be to add the `karpenter.sh/disruption=disrupting:NoSchedule` taint like we currently do when we are actively disrupting the node (this is similar to the cordoning logic that currently happens during a standard node drain, except it allows us to drain DaemonSets as well). Then, once we hit our node `terminationGracePeriod`, we add the [out-of-service] taint that already has built-in force logic built-in to the control plane for rescheduling pods and detaching volumes based on `NoExecute`. Here, once we add this taint, we should respect any `tolerationSeconds` that's specified and including waiting for all pods to be removed from the node that either 1) Don't tolerate the `NoExecute` taint or 2) Tolerate the `NoExecute` taint, but only for a specified `tolerationSeconds`. After all of these pods are removed, and we are only left with pods that tolerate the taint, we force delete the Node.

This allows some nice flexibility since the default behavior just terminates the node in a forceful manner once it exceeds the grace period, but still allows standard taint toleration flexibility around `NoExecute`. This also (IMO) makes the most semantic sense since it will forcefully start draining the node past its grace period (though it will still allow one-off exceptions to tolerate taints a bit longer with `tolerationSeconds`).

The one thing we need to think about here, though: Does this still provide the same persona-based isolation that this feature is trying to enforce? Respecting the `tolerationSeconds` for a pod toleration for `out-of-service` still means that there needs to be some policy enforcement to ensure that users aren't setting this to an unnecessary high value or just to ensure that they aren't tolerating this taint _at all_ and only let cluster admins selectively tolerate this taint.

designs/nodepool-termination-grace-period.md
```diff
@@ -0,0 +1,127 @@
+# NodePool Termination Grace Period
+
+## Motivation
+Users are requesting the ability to control how long karpenter will wait for the deletion of an individual node to complete before forcefully terminating the node regardless of pod status. [#743](https://github.com/kubernetes-sigs/karpenter/issues/743). This supports two primary use cases.
+* Cluster admins who want to ensure that nodes are cycled after a given period of time, regardless of user-defined disruption controls (such as PDBs or preStop hooks) that might prevent eviction of a pod beyond the configured limit. This could be to satisfy security requirements or for convenience.
```
jonathan-innis, line 5: When you are forcefully draining nodes in this way, how do you coordinate with the app devs on your clusters? Do you inform them that this change is incoming? Is there a way that you can enforce that they don't do things like a fully-blocking PDB through policy or just discussing with them or is it just outright untenable to keep up with the number of ways that users might mis-configure their nodes?

designs/disruption-controls-drain-timeout.md
jonathan-innis: > requiring K8s 1.26 or feature gates

IIUC, this should still work for all versions of K8s; the non-graceful feature just gives us the extra benefit of the node getting force-drained if the node's taint eviction manager is down. So, for the vast majority of cases, this should handle the same as a typical eviction flow when a `NoExecute` is added.

designs/nodepool-termination-grace-period.md
wmgroot: We have internal documentation dedicated to describing our node rollout processes for handling upgrades, how to control pod disruption with PDBs, preStop hooks, etc. We make it clear in these docs that voluntary disruptions must be tolerated by users of our clusters, and that they may delay eviction by up to X time before their pods are evicted forcefully (24h currently).

wmgroot: > there needs to be some policy enforcement to ensure that users aren't setting this to an unnecessary high value or just to ensure that they aren't tolerating this taint at all and only let cluster admins selectively tolerate this taint.

1. I'm not sure if the out-of-service taint is handled specially by K8s. This can be tested.
2. If it's a problem, it can be solved by validating admission webhooks such as OPA and the like.

K8s has a lot of fields that are potentially disruptive to other users of a cluster, such as priorityClass, terminationGracePeriodSeconds, etc which often require additional validation controls to support multi-tenancy.

sftim:
I presume that any Pod that tries to tolerate everything gets shut down, ordered based on its `priority`, at the last possible moment (at least on Linux). That outcome matches our understanding of the intent of whoever defined that Pod. It might be a surprise to the app but it people wanted more notice, they shouldn't tolerate all the taints that could have given them that extra notice.

jonathan-innis:
@wmgroot @sftim I was thinking more about the ordered drain problem with this force drain operation, it seems like we can collapse the behavior that we want here with the force drain logic into the following tenents:
1. A cluster admin can set `terminationGracePeriod` on their NodeClaims; this value, is the maximum time that the NodeClaim can be alive after the deletionTimestamp is set. This effectively means the maximum amount of time that the Node is draining, but just because of eventual consistency and processing, there might be a 1-2s wiggle room in here.
2. Karpenter will drain pods after the deletion timestamp is set with a requeue mechanism and with the following batching:
    1. Any pod who has a `terminationGracePeriodSeconds` <= Remaining time left in the `terminationGracePeriod` of the NodeClaim gets force drained. We won't call the eviction API in this case since if the following statement is true we don't care about blocking PDBs. We just want to give the pod the maximum amount of time to spin-down.
    2. Non-Critical Non-Daemonset pods are added to the set of evictable pods in this eviction loop
    3. If there are no pods in #2, add Non-Critical Daemonset pods to the set of evictable pods in this eviction loop
    4. If there are no pods in #3, add Critical Non-Daemonset pods to the set of evictable pods in the eviction loop
    5. If there are no pods in #4, add Critical Daemonset pods to the set of evictable pods in the eviction loop

Once there are no more pods or the `terminationGracePeriod` of the NodeClaim is exhausted, we add the non-graceful termination taint to the Node. If there are any pods that tolerate this taint remaining, we wait for these pods to be removed and terminated off of the node. Once it's only pods that tolerate this NoExeucte taint remaining, we terminate the instance and remove the finalizer from the NodeClaim.

wmgroot:
> @wmgroot @sftim I was thinking more about the ordered drain problem with this force drain operation, it seems like we can collapse the behavior that we want here with the force drain logic into the following tenents:
> 
>     1. A cluster admin can set `terminationGracePeriod` on their NodeClaims; this value, is the maximum time that the NodeClaim can be alive after the deletionTimestamp is set. This effectively means the maximum amount of time that the Node is draining, but just because of eventual consistency and processing, there might be a 1-2s wiggle room in here.
> 
>     2. Karpenter will drain pods after the deletion timestamp is set with a requeue mechanism and with the following batching:
>        
>        1. Any pod who has a `terminationGracePeriodSeconds` <= Remaining time left in the `terminationGracePeriod` of the NodeClaim gets force drained. We won't call the eviction API in this case since if the following statement is true we don't care about blocking PDBs. We just want to give the pod the maximum amount of time to spin-down.
>        2. Non-Critical Non-Daemonset pods are added to the set of evictable pods in this eviction loop
>        3. If there are no pods in ii, add Non-Critical Daemonset pods to the set of evictable pods in this eviction loop
>        4. If there are no pods in iii, add Critical Non-Daemonset pods to the set of evictable pods in the eviction loop
>        5. If there are no pods in iv, add Critical Daemonset pods to the set of evictable pods in the eviction loop
> 
> 
> Once there are no more pods or the `terminationGracePeriod` of the NodeClaim is exhausted, we add the non-graceful termination taint to the Node. If there are any pods that tolerate this taint remaining, we wait for these pods to be removed and terminated off of the node. Once it's only pods that tolerate this NoExeucte taint remaining, we terminate the instance and remove the finalizer from the NodeClaim.

I think anything smarter than just adding the official k8s taint can be added in a further update to this behavior if a need is determined for it. The PR I have open is dead simple and works in my testing. I think any kind of special handling of the order of force deletes on pods prior to setting the taint may be over-engineering.

Testing what behavior occurs if a pod tolerates all taints to ensure it doesn't block node deletion is worth because that's a very common possibility.

jonathan-innis:
> but as an addendum I think we need to shift this all before the deletion timestamp is placed on the node claim

This one I'm less sure about. I agree that we need a mechansim to be able to force the delete of all NodeClaims for uninstall scenarios, but I'm not sure that removing the finalizer is the right way to go about it. I could likewise see that you orchestrate an update that sets the terminationGracePeriod of all NodeClaims to `0s` and then performs the NodeClaim deletion as another way to orchestrate the deletion of the entire CRD without hanging. It feels strange to me that we would be adding a `terminationGracePeriod` for the NodeClaim but then not actually be performing a deletion of the NodeClaim until after the grace period. This is counter to how grace period works elsewhere in K8s.

jonathan-innis:
> I think anything smarter than just adding the official k8s taint can be added in a further update to this behavior if a need is determined for it

I disagree here. A lot of the ordering logic that's built-in above is already handled by the current termination controller. The only difference that is really being proposed with this change, is to ensure that pods receive appropriate heads-up before the terminationGracePeriod is met and the pods are force-drained due to non-graceful node shutdown taint. If we don't add the ability to terminate pods and respect their grace period, pods may never receive an eviction call due to fully blocking PDBs and interrupted with no time to clean-up.

wmgroot:
> > I think anything smarter than just adding the official k8s taint can be added in a further update to this behavior if a need is determined for it
> 
> I disagree here. A lot of the ordering logic that's built-in above is already handled by the current termination controller. The only difference that is really being proposed with this change, is to ensure that pods receive appropriate heads-up before the terminationGracePeriod is met and the pods are force-drained due to non-graceful node shutdown taint. If we don't add the ability to terminate pods and respect their grace period, pods may never receive an eviction call due to fully blocking PDBs and interrupted with no time to clean-up.

I think the extra eviction behavior you're asking for would be better implemented as a general enhancement to this feature instead of being limited to Karpenter behavior.
https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown

jonathan-innis:
> I think the extra eviction behavior you're asking for would be better implemented as a general enhancement to this feature instead of being limited to Karpenter behavior

How would that function with the `out-of-service` taint? The pods need to be aware that they need to be removed before the `out-of-service` taint is added in this case. I agree that we shouldn't over-engineer the solution here but I also want to be cautious that pods with blocking PDBs that are force terminated still get the requisite time that they have requested to do their clean-up if we have the chance to be smarter about their termination behavior.

wmgroot:
> How would that function with the `out-of-service` taint? The pods need to be aware that they need to be removed before the `out-of-service` taint is added in this case. I agree that we shouldn't over-engineer the solution here but I also want to be cautious that pods with blocking PDBs that are force terminated still get the requisite time that they have requested to do their clean-up if we have the chance to be smarter about their termination behavior.

How long is the "requisite time" that a pod needs to do cleanup? `terminationGracePeriodSeconds` is user-controlled and could be set to hours or days.

jonathan-innis:
> How long is the "requisite time" that a pod needs to do cleanup? terminationGracePeriodSeconds is user-controlled and could be set to hours or days

Right, so the time that the pod gets to drain is equal to the following equation:

```
podTerminationTime = min(pod.terminationGracePeriodSeconds, nodeclaim.terminationGracePeriod)
```

This ensures that if the nodeClaim terminationGracePeriod is ever less than the pod termination grace period, those pods will get force drained and start terminating immediately since they have requested such a long grace period. This means they might not get the _full_ grace period, but this ensures that the NodeClaim is fully deleted by the time the cluster admin asked for

wmgroot:
Isn't that already how it works without any extra logic that references the pods' terminationGracePeriodSeconds?

When a Node or NodeClaim is deleted, it starts a drain on the node, which attempts to evict a pod and puts it into terminating status.

A. If the pod's terminationGracePeriodSeconds is less than the NodeClaim's terminationGracePeriod, it will be forcibly terminated before the out-of-service taint is applied.
B. If it's greater (and the pod doesn't terminate until it reaches it's maximum tGPS), the out-of-service taint will be applied first, forcibly removing the pod before it can reach its tGPS, but still giving the pod the full NodeClaim tGP to terminate.

I'm not understanding how what you're proposing differs from this behavior.

ellistarn:
> I could likewise see that you orchestrate an update that sets the terminationGracePeriod of all NodeClaims to 0s

Giving/requiring installers to have write permissions to NodeClaims is likely a nonstarter. Alternatively, you could have the finalizer attempt to introspect the CRD to check if it's a cascading delete, but this is another permissions boundary issue. I don't really see a good alternative to cascading delete.

Further triggering cordon/drain via a taint as @wmgroot states has precedence in k8s, though I think we have flexibility on whether or not we must use the upstream taint. I could see an argument that it doesn't meet our needs, but would need to dig in a bit.

What's your concern with using taints as the driver of cordon/drain?

ellistarn:
> I think the extra eviction behavior you're asking for would be better implemented as a general enhancement to this feature instead of being limited to Karpenter behavior.

I can see this as a long term approach, but I don't think we necessarily want to block on upstreaming in the interim.

njtran:
Another thing I'd like to discuss (maybe I'm missing some discussion before this), is where did we land for configuration surface?

My thought is that we shouldn't make this be driftable for nodes, even if that's how deployments and pods work. Pod restarts are inherently much less expensive than node starts, and I think this is the place to draw the line. I think adding the `terminationGracePeriodDuration` into the `nodePool.Spec.Disruption` block is the best place for this to live, we can infer from the NodePool's setting what the value should be here, and punt on adding this as a field into the nodeclaim itself.

jonathan-innis:
> we can infer from the NodePool's setting what the value should be here, and punt on adding this as a field into the nodeclaim itself

After some discussion with @njtran, I'm +1 on this proposal. We don't have a ton of utility for standalone NodeClaims today. We can always add this as an additional field on the NodeClaims down-the-line.

njtran:
Also just a ping here: would love to talk about this at the working group today @ 2PM PST if everyone's planning on coming.

designs/nodepool-termination-grace-period.md
```diff
@@ -0,0 +1,127 @@
+# NodePool Termination Grace Period
```
ellistarn, line 1: Thanks for this RFC! We've gone back and forth on a bunch of options (in comments) for the spec.template, spec.disruption, or just spec. Can you lay out those options, their pros and cons, and then make a recommendation?

wmgroot: Updated with a preferred and alternative spec for the location of the `terminationGracePeriod` field.

wmgroot:
Followed up with @jonathan-innis and @njtran in slack about the nuances of the optimized pod deletion behavior they're looking for. I've included a rundown of how it would behave in the proposal.

It's _possible_ that we'd want to put it behind a flag of some sort, but I'm ok with including it as the default behavior and addressing making it optional in the future if there's a need for it.

designs/nodepool-termination-grace-period.md
```diff
@@ -0,0 +1,175 @@
...
+    spec:
+      terminationGracePeriod: 24h   # metav1.Duration, nil if not set
+```
+
+#### Benefits
+1. Changes to the `terminationGracePeriod` would be less surprising, because current live nodes would not be impacted, the change would only take effect after new Nodes/NodeClaims had replaced old nodes.
+2. Behavior matches the behavior of a Pod's `terminationGracePeriodSeconds` more closely, which is not active until new pods have been created with a full pod rollout.
+
+#### Drawbacks
+1. Making updates to the `terminationGracePeriod` for a NodePool is cumbersome and disruptive to users of the cluster.
```
ellistarn, line 72: I have a moderate preference to making this part of template, as it creates extensibility to standalone nodes. I agree that we should exclude it from drift, which is a slight deviation from how pods work, which is slightly surprising. However, both solutions deviate slightly in surprising ways, so I don't see the proposal and an advantage on this point. 

Making it part of template and excluding from drift is simply more powerful/extensible, without drawback compared to the proposal.

jonathan-innis:
> It's possible that we'd want to put it behind a flag of some sort, but I'm ok with including it as the default behavior and addressing making it optional in the future if there's a need for it

In a lot of ways, enabling the field itself _is_ the flag here.

designs/nodepool-termination-grace-period.md
jonathan-innis: >  I agree that we should exclude it from drift, which is a slight deviation from how pods work, which is slightly surprising

I have a few questions/comments that stem off the top-of-my-head if we choose to go the template route:
1. I do like that it's a standalone thing that just works if NodeClaims are generalized outside of NodePools. Up to this point, we have taken the stance that a NodeClaim can live on its own and serve as declarative state for a node launch and a node teardown. Having `terminationGracePeriod` in there is definitely nice
2. It raises a question in my head about what else is a stand-alone concept that we would put in there. `minLifetime`, `maxLifetime`? How different is `maxLifetime` from `expireAfter` today in the `disruption` block. For that matter, why is `expireAfter` in the disruption block if the concept can live on its own and we start putting other behavioral fields in the NodeClaim
3. The last one -- perhaps the one that I have the biggest question about -- if we don't drift on the value, what happens when someone updates the value in their NodePool that is different from the value in the NodeClaims? One option here is we reconcile the new value onto all of the owned NodeClaims. The downside of that approach is now how do you manually update the NodeClaim `terminationGracePeriod` if you want to shorten it in one-off cases. If the value is reconciled down, then you don't really have that option since whatever you change will just get reconciled back to whatever is in the NodePool. That bit feels awkward to me, since I'd like to think that the reason that we would choose to put it in the NodeClaim is so that manual interaction _is_ possible. The other options is you drift on it. You accept the painpoint _but_ you get the nice behavior that you can make manual changes to the NodeClaim and they won't be overridden after the NodeClaim's initial creation

ellistarn: For 3, there's another option where we don't reconcile it down, and the controller just uses whatever happens to be in the nodeclaim. This may be "out of date" and a point in time value from the launch.

jonathan-innis: That's true. I question, though, if that behavior is more surprising than just drifting to reconcile the state that the user requested in their NodePool.

ellistarn: I could go either way on this one. At the end of the day, as long as we document the behavior, users will have to work with whether it drifts (or not). The value of having it on the NodeClaim seems clear.

garvinp-stripe: This is great because I have always wonder what is "NodeClaim" and what does it represent. The Karpenter docs doesn't have section on it. 

From my perspective I have always seen NodeClaim as just a Karpenter representation of a node. From that perspective I have always imagine all disruption related logic falls outside of NodeClaim as it is not an attribute of the node itself but rather the management of the node. Following that I would to say NodeClaims spec shouldn't have any information about disruption including maxLife or graceful node termination since these are actions taken not by the node itself but a service that manages the node. I wouldn't mind having annotations that have these information but thats purely for convenience

jonathan-innis: > Following that I would to say NodeClaims spec shouldn't have any information about disruption including maxLife or graceful node termination since these are actions taken not by the node itself but a service that manages the node

I could see NodeClaim being a step above a "representation of the node". It's the node's manager IMO. I definitely think there's room to say that a NodeClaim is the declarative state for a Node, including providing details for the launch request as well as how the node gets spun down.

> since these are actions taken not by the node itself but a service that manages the node

The NodeClaim does orchestrate some of this logic today, at least for draining.

> The Karpenter docs doesn't have section on it.

I whole-heartedly agree that we should have a section in the docs. There's an issue that's open in the `aws/karpenter-provider-aws` repo for this, but this just isn't something that we have had the chance to get around to yet: https://github.com/aws/karpenter-provider-aws/issues/5144.

garvinp-stripe: > I could see NodeClaim being a step above a "representation of the node". It's the node's manager IMO. I definitely think there's room to say that a NodeClaim is the declarative state for a Node, including providing details for the launch request as well as how the node gets spun down.

I see. Thanks for the clarification.

I think its important to establish what is the responsibility of NodeClaims in order to know if NodeClaims should get disruption related information propagated to it. Agreed that I think it make sense that NodeClaim manages spin up and down however the current terminator controller logic seems to jump between interacting directly with the Node itself and NodeClaim's terminator controller which makes it very difficult to understand who is responsible.

From a tactical point of view, propagating termination related information down to NodeClaim is a one way door because it means NodeClaim should have the authority on how the termination happens. If we are undecided or need more time to determine if this is the right way, wouldn't it be safer to not do this right now? If we eventually decide that NodeClaim should have this responsibility, we should do it with fully conviction and completely migrate all termination logic into NodeClaim rather than partially doing now which continues building tech debt for little near term benefit.

jonathan-innis: > is a one way door because it means NodeClaim should have the authority on how the termination happens

Agreed. This is something that we are thinking about for v1. It actually raises an even deeper question around whether expiration is a concept that should be stored strictly on the NodePool or if it also lives as a standalone concept on the NodeClaim. If you reframe the concept into `maxAge` or `maxLifetime`, all of a sudden you've got something that sounds like it might belong on the NodeClaim. I agree that if we take the step to push the `terminationGracePeriod` to the NodeClaim, we should do so with the plan to introduce expiration at the NodeClaim-level and drop it out of the NodePool.

jonathan-innis:
@wmgroot Just a heads-up. Given the way that this conversation has gone in terms of API surface, @njtran and I are convinced that we should pull figuring out where this goes into the v1 API design that we are planning to put out relatively soon on Karpenter's path to full stability.

With that being said, I want to settle out the API question before we merge this. I think it may take us a couple weeks or so to settle-out these questions. Happy to involve you in any discussions as we have them, but just wanted to give you a heads-up that this RFC may not get any movement until we figure that bit out.

Otherwise, I'm super excited about getting this feature in. I'll still take a look at your implementation so that we can get all the minutia of the code squared away so we are ready to merge once the API stuff is sorted.

wmgroot:
Sounds good to me. We should be able to work on the implementation details in parallel since switching between NodePool vs NodeClaim spec is a fairly trivial part of the actual implementation.

designs/nodepool-termination-grace-period.md
wmgroot: I'm in favor of putting terminationGracePeriod back into the NodeClaim spec simply to support users of NodeClaims without NodePools (I also think expiration should belong there too). I've updated the design to reflect that.

jonathan-innis:
cc: @akestner

designs/nodeclaim-termination-grace-period.md
```diff
@@ -0,0 +1,168 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodeClaim
+spec:
+  terminationGracePeriod: 24h   # metav1.Duration, nil if not set
+```
+
+## Code Definition
+
+```go
+type NodeClaim struct {
```
jmdeal, line 23:
```suggestion
type NodeClaimSpec struct {
```

designs/nodeclaim-termination-grace-period.md
```diff
@@ -0,0 +1,168 @@
...
+
+### Termination Grace Period Expiration Behavior
+1. NodeClaim deletion occurs (user initiated via Node finalizer,or when Karpenter begins node rotation from drift detection, etc).
+2. NodeClaim deletionTimestamp is set.
+3. Standard node drain process begins.
+4. Drain attempts fail (due to PDBs, preStop hooks, etc). Karpenter adds an Event to the Node noting that it will be forcibly terminated at X time.
+5. Reconciliation of the Node/NodeClaim is requeued (potentially many times).
+6. Time passes, and the configured terminationGracePeriod for the NodeClaim is exceeded.
+7. Karpenter adds the official ["out-of-service" taint](https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown), instructing Kubernetes to forcefully remove all remaining pods. This requires K8s 1.26 to be available without setting a feature gate.
```
jmdeal, line 117: We can drop this step

designs/nodeclaim-termination-grace-period.md
```diff
@@ -0,0 +1,168 @@
...
+2. This behavior could be confusing to users, who may have set their pod terminationGracePeriodSeconds to higher than necessary values and have their pods suddenly deleted when a node begins deletion.
+
+### Documented Warnings
+This feature carries some risk for cluster administrators given that it can bypass the configured grace periods and PDBs set by users of a cluster. These risks should be called out explicitly in the docs for the feature to minimize the risk of a cluster administrator unintentionally disrupting their users' workloads.
+1. Enabling this feature by setting NodePool.spec.template.terminationGracePeriod to any value will result in a drift reconciliation that will replace all nodes. The new terminationGracePeriod value will not be active until new NodeClaims have been provisioned.
+2. Enabling this feature can result in termination of user pods in violation of their configured grace periods or PDBs.
+3. Because of the optimization stated above, user pods could be terminated in violdation of grace periods and PDBs **before** the configured NodeClaim grace period is reached.
```
jmdeal, line 150: This shouldn't be in violation of grace periods right, just PDBs? Grace periods would only be disrespected when the TGP has expired.

jmdeal (commented):
Couple of small things to bring this inline with what we merged in #916, but other than that this looks good to go.

designs/nodeclaim-termination-grace-period.md
wmgroot: Grace periods could be violated if the pod's TGPS is greater than the NodeClaim's TGP. I'll reword this.

njtran (commented):
/lgtm
/approve

Bryce-Soghigian:
great work!

PR #799, opened 2023-11-16
chore: Revert staging changes for budgets

njtran:
Fixes #N/A 

**Description**
Revert staging changes for budgets until it's ready to be released

The changes that are being reverted are these two, which are only API and validations/defaults. 
- https://github.com/aws/karpenter-core/commit/707bb6b8cdd2136411a8564f0184218f6aafa309
- https://github.com/aws/karpenter-core/commit/f4265df62b4ee6cdb33b111d7e91269846537032

**How was this change tested?**
make presubmit

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #839, opened 2023-12-05
Revert "chore: Revert staging changes for budgets (#799)"

njtran:
This reverts commit 331c506f207e350a3c417ee8a9e4a4cb27cea1a5.

Fixes #N/A 

**Description**
This reverts the API reverts from a couple weeks back to begin staging the changes for budgets again.

make presubmit


pkg/apis/v1beta1/nodepool.go
```diff
@@ -78,6 +79,40 @@ type Disruption struct {
...
+
+// Budget defines when Karpenter will restrict the
+// number of Node Claims that can be terminating simultaneously.
+type Budget struct {
+	// MaxUnavailable dictates how many NodeClaims owned by this NodePool
+	// can be terminating at once. It must be set.
+	// This only considers NodeClaims with the karpenter.sh/disruption taint.
+	// +kubebuilder:validation:XIntOrString
+	// +kubebuilder:default:="10%"
+	MaxUnavailable intstr.IntOrString `json:"maxUnavailable" hash:"ignore"`
```
ellistarn, line 101: didn't we discuss "value" for this?

pkg/apis/v1beta1/nodepool.go
```diff
@@ -78,6 +79,40 @@ type Disruption struct {
...
+	// Crontab specifies when a budget begins being active,
+	// using the upstream cronjob syntax. If omitted, the budget is always active.
+	// Currently timezones are not supported.
+	// This is required if Duration is set.
+	// +kubebuilder:validation:Pattern:=`^(@(annually|yearly|monthly|weekly|daily|midnight|hourly))|((.*)\s(.*)\s(.*)\s(.*)\s(.*))$`
+	// +optional
+	Crontab *string `json:"crontab,omitempty" hash:"ignore"`
```
ellistarn, line 108: Cronjob uses `schedule` https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/

pkg/apis/v1beta1/nodepool.go
njtran: The consensus was to do MaxUnavailable. It matches really nicely with the fact that the parent struct name is budgets, relying on the established parallel with PDB -> MaxUnavailable

pkg/apis/v1beta1/nodepool.go
njtran: Yep, I think schedule may be misinterpreted, since the actual schedule of this budget is determined by the combination of this crontab and the duration field. One could argue that it should be something like `starts` or `beginnings` but I think a field name of `crontab` indicates the syntax of the string more deliberately.

jonathan-innis (approved):
/lgtm
/approve

pkg/apis/v1beta1/nodepool.go
ellistarn: Crontab seems a bit awkward to me, since a crontab is a file with a schedule + command: https://man7.org/linux/man-pages/man5/crontab.5.html

pkg/apis/v1beta1/nodepool.go
ellistarn: MaxUnavailable only make sense in comparison to MinAvailable.

ellistarn: Also -- MaxUnavailable is a bit confusing to me, since it implies what we know what the maximum value is.

PR #840, opened 2023-12-05
chore: remove the remainder of the alpha references

njtran:
Fixes #N/A 

**Description**
Removes the alpha directory and all associated removals associated with Provisioners or Machines.

make presubmit


pkg/controllers/metrics/node/controller.go
```diff
@@ -106,7 +104,6 @@ func nodeLabelNames() []string {
 		sets.New(lo.Values(wellKnownLabels)...).UnsortedList(),
 		resourceType,
 		nodeName,
-		nodeProvisioner,
```
jonathan-innis, line 109: Should this be `nodeNodePool` now?

pkg/controllers/metrics/node/controller.go
```diff
@@ -176,7 +173,6 @@ func getNodeLabels(node *v1.Node, resourceTypeName string) prometheus.Labels {
 	metricLabels := prometheus.Labels{}
 	metricLabels[resourceType] = resourceTypeName
 	metricLabels[nodeName] = node.Name
-	metricLabels[nodeProvisioner] = node.Labels[v1alpha5.ProvisionerNameLabelKey]
```
jonathan-innis, line 179: `metricLabels[nodeNodePool] = node.Labels[v1beta1.NodePoolLabelKey]`

pkg/controllers/metrics/node/controller.go
njtran: I intentionally left this out since we have this label in the well known labels. If we add it here, this panics for having duplicates. I'm happy to lo.Reject() the above call on L104 and add it back in here, but I think just having a comment here is better than having redundant code. Thoughts?

pkg/controllers/metrics/node/controller.go
njtran: Good catch!

pkg/controllers/metrics/node/controller.go
jonathan-innis: Yep, that makes sense to me. IIRC, that's the reason why I added the nodeProvisioner in here separately to begin with.

jonathan-innis (approved):
/lgtm

jonathan-innis (approved):
/lgtm

Issue #740, opened 2022-10-20
Custom drain flow


dorsegal:
### Tell us about your request

Add a `rollout` flag when using drain. It will be used when consolidation and native termination handler (https://github.com/aws/karpenter/pull/2546)  will be ready.
The custom drain flow is like this:

1. Cordon the node
2. Do a rolling restart of the deployments that have pods running on the node.
3. Drain the node.

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Currently when using consolidation feature or[ aws-node-termination-handler](https://github.com/aws/aws-node-termination-handler) we can result of downtime or heavy performance degradation with the current implementation of `kubectl drain`

Current drain will terminate all workloads on a node and the scheduler will try to create those workloads on available nodes and if not any Karpenter will provision new node.
Even with PDB there is some level or degradation. 

### Are you currently working around this issue?

having a custom bash script that implements an alternative to `kubectl drain`

https://gist.github.com/juliohm1978/1f24f9259399e1e1edf092f1e2c7b089

### Additional Context

kubectl drain leads to downtime even with a PodDisruptionBudget https://github.com/kubernetes/kubernetes/issues/48307

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Bharath509:
Please open this issue

vainkop:
Would be nice to see a proper solution

nikoul:
This issue also affects our single-replica deployments that are accessed through a service. Initially, I considered using a preStop hook, but this solution proved ineffective since it still transitions the pod to a terminating state and immediately removes it from the service endpoints.

EvgeniGordeev:
https://github.com/kubernetes-sigs/karpenter/discussions/2532

Issue #622, opened 2022-12-05
Cordon node with a do-not-evict pod when ttlSecondsUntilExpired is met


cest-pas-faux:
### Tell us about your request

Regarding the do-not-evict annotation, it currrently prevents some nodes to be deprovisioned on our clusters.

I know, that's the goal and i'm fine with it, but would it be possible to cordon the node at least when the `ttlSecondsUntilExpired` is met ?

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

We provision expensive nodes as a failback for some workloads with a ttl of 3 hours, and we saw some still running with a 6 hours uptime.
The annotation was preventing the drain/deletion as cronjobs kept being scheduled on it, but a manual cordon fixed it within minutes.

### Are you currently working around this issue?

No

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jukie:
Has there been any more discussions on this? Would PRs adding this behavior be accepted?

jonathan-innis:
> Has there been any more discussions on this

I think we should consider discussing this in the next working group. I'd like to see a design on this one considering the impact that this could have to the cluster, given there could be cases if we're not careful where all nodes on the cluster get tainted all at once. 

As an example, if we have a set of nodes that are expired all at once and we cordon them all at once, is that a reasonable stance to take, knowing that those nodes will get eventually disrupted? Probably, assuming that our disruption logic can handle the disruption quick enough that have underutilization on that node isn't a problem, but what if that node has a fully-blocking PDB? What if that node has a `do-not-evict` pod and that pod never goes away? Do we need some forceful termination that happens after a node has been cordoned with a blocking pod for some amount of time? There might be some interactions here that we should consider with #743

jukie:
Agreed there are some risks and also some relation to other proposed features under https://github.com/aws/karpenter/issues/1738 but I don't think that should limit from implementing this through a feature gate or by disabling as default.

How do I see the schedule for bi-weekly's? Looks like the notes haven't been updated in awhile so it's unclear whether it's this coming week or next.

njtran:
[...]

PR #843, opened 2023-12-06
docs: RFC for Customizable Node Finalizer

garvinp-stripe:
Fixes  [#743](https://github.com/kubernetes-sigs/karpenter/issues/743), [#622](https://github.com/kubernetes-sigs/karpenter/issues/622), [#740](https://github.com/kubernetes-sigs/karpenter/issues/740), [#690](https://github.com/kubernetes-sigs/karpenter/issues/690), [#5232](https://github.com/aws/karpenter/issues/5232)
**Description**
RFC for adding customizable finalizer for nodes


designs/node-termination-hook.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Custom Finalizer behavior/ expectation
+During termination, Karpenter will not perform the finalization for both node claims and nodes but allow the user's custom controller to perform 
+the finalization of node and nodeclaim. We will allow users to set a grace termination period from [#834](https://github.com/kubernetes-sigs/karpenter/pull/834) where 
+if set the and breached then Karpenter's termination will occur ungracefully and force terminate both the node and node claim. 
+
+##Concerns
+- I am not completely sure if allowing users have control over the finalizer of nodeclaim fully make sense so I am open to changing this. It could be that user defines
+a custom finalizer just at the node level and when Karpenter addFinalizer on nodes it uses the custom finalizer rather than the one provided by Karpenter
+- Alternatively Karpenter could provide an option where no node finalizer is set by Karpenter and that way it allows the users to inject their own finalizer without
```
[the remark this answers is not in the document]
garvinp-stripe: When you say "this finalizer" is it the finalizer on nodeclaim or node?

designs/node-termination-hook.md
```diff
@@ -0,0 +1,72 @@
+# Custom Node termination (custom finalizer)
+
+## Motivation
```
jonathan-innis, line 3: I'm curious why we think that this interface needs to be "pluggable" rather than supporting these features natively in Karpenter directly. I know we have linked out a few features here, but I'd be interested to hear the tangible ways that controllers are going to plug-in to this new system, so that we know that we are actually solving these use-cases if we were to implement this interface. IMO, a lot of these features make sense to be part of Karpenter natively, we just haven't had the time or resources to get some of them into the project. There's also some design thinking that needs to go into things like "cordon nodes after expiry."

I think there's two buckets of features that are being listed here: 1) Features that we know that we want to do, but need a little design and direction to determine the exact best way to move forward with it (https://github.com/kubernetes-sigs/karpenter/issues/743, https://github.com/kubernetes-sigs/karpenter/issues/622) and 2) Features that we're less sure that we want to pursue, with design that might touch both upstream and Karpenter (https://github.com/kubernetes-sigs/karpenter/issues/740, https://github.com/kubernetes-sigs/karpenter/issues/690, https://github.com/aws/karpenter-provider-aws/issues/5232).

The other thing to mention, is that any controller today has the ability to plug an additional finalizer into the Node or the NodeClaim to add additional finalization logic. Granted, this won't stop the Karpenter finalizer from taking action, but you can layer additional things on top of the typical finalization flow.

garvinp-stripe: _TLDR: Can we modularize finalizer architecture to be more Singular Responsibility so contribution may be less complex/ less risky and users can attempt to take over parts they want to customize?_

Good question, sorry I never answer similar ask from @njtran. 
Initially , we were thinking about a way to notify operators that a node is stuck in Production rather than relying on a timeout to "do something" with the node that's stuck. In theory we could update Karpenter's finalizer to send a more clear event and we can capture that to notify our operators.  However this begs the question if there are ways we can have better control drain and node termination. 

Specifically, I understand the desire for Karpenter to natively do more advance termination logic however it is difficult to do this right now because the termination logic is tightly wrapped into a single finalizer. I wonder if alternative approach could be breaking out the finalizer into "[stages](https://github.com/kubernetes-sigs/karpenter/pull/843/files#r1421083940)" and allowing contributors to contribute logic that can perform these stages differently?

designs/node-termination-hook.md
```diff
@@ -0,0 +1,72 @@
...
+	} else {
+		controllerutil.AddFinalizer(node, v1beta1.TerminationFinalizer) 
+	}...
+}
+```
+
+### Custom Finalizer behavior/ expectation
+During termination, Karpenter will not perform the finalization for both node claims and nodes but allow the user's custom controller to perform 
+the finalization of node and nodeclaim. We will allow users to set a grace termination period from [#834](https://github.com/kubernetes-sigs/karpenter/pull/834) where 
+if set the and breached then Karpenter's termination will occur ungracefully and force terminate both the node and node claim.
```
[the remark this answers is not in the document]
garvinp-stripe: Make sense can clean up

designs/node-termination-hook.md
garvinp-stripe: Thinking about this more, I suspect my thinking is bias towards how we did things pre-Karpenter. Since we use ASGs, we rely heavily on lifecycle hooks which in a way I am trying to replace with finalizer. I think finalizer isn't completely the replacement and maybe it shouldn't be. 

I think we could move a majority of the logic to be on a daemonset that interacts with host level processes since finalizer does wait for daemonsets to terminate. But this raise a question around [graceful termination timeout](https://github.com/kubernetes-sigs/karpenter/pull/834). I see that the finalizer [orders](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/terminator/terminator.go#L113) it termination sequence. If we want do move in that direction (termination lifecycle hook logic -> daemonset) we need our daemonset to gracefully terminate which means the graceful termination period will likely cause issues if user pods drain the full termination period. I think we (my team) may want to separate out finalizer between user application vs cluster operator applications graceful termination periods.

Does it make sense for finalizer to capture all these complexity? I am just trying to wrap my head around how to keep Karpenter's API surface clean but yet not be overly complex in what it manages because of user needs.

njtran: I think this RFC is prescribing a solution rather than tackling the problem as a whole, which I understand to be a custom drain flow. I think a custom finalizer could be one of the solutions here, but I think we're leaving out some considerations of other alternatives and an understanding of the use-cases here. 

Here are some questions I find myself asking, which I'm happy to talk over slack if that makes the discussion easier:
- What are the alternatives for API Spec here? 
  - A setting for enabling/disabling the termination controller response to finalizers? 
  - A new finalizer to indicate a custom drain flow? 
  - A setting to say which part of each termination action needs to be different?
- Do all users who want custom termination flow have the ability to run a custom controller that handles the termination flow themselves? Is full extensibility the right answer? 
  - If users want all existing features of Karpenter, but a change in one part of the code, does a fully custom termination flow solve this issue for them, or would that be not worth the effort for them?
- What termination controller features do _you_ need implemented in Karpenter? Can we do any of those without adding an entirely new controller? Are any of those use-cases not covered in an issue?

Issue #688, opened 2023-07-07
Custom Deprovisioning Trigger Mechanism


sidewinder12s:
### Description

Allow users to trigger node Drift

**What problem are you trying to solve?**

In Karpenter pre-v0.28, I had started using the karpenter drift annotation `karpenter.sh/voluntary-disruption=drifted` as a way to force nodes to get replaced in an orderly fashion when I changed configuration that was not supported by Karpenters Drift detection. 

In v0.28 this was removed and now the annotation is simply removed by Karpenter. 

I found the ability to trigger drift useful in testing and in filling in the gaps in Drift support. I'd also assume long term, there may be corner cases users would want to trigger replacement on that drift cannot detect or detect easily. 

Perhaps just another annotation indicating user requested drift so that Karpenter can replace nodes in an orderly manner and while respecting deprovisioning controls. 

**How important is this feature to you?**

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

njtran:
Posting thoughts here for you on how to approach this: 

**Design**
- How should we allow users to signal a node being terminated? 
  - Should we use an annotation or a status condition? We mark nodeClaims with status conditions to indicate when they're expired/drifted. Maybe we could do the same? But that's tougher to do from a user's perspective. In my eyes, annotation on the node/NodeClaim may be the easiest and make the most sense. 
- Should this have any special disruption behavior from what's listed [here in the docs](https://karpenter.sh/docs/concepts/disruption/#disruption-controller)? 

**Implementation**
Currently, we iterate through the disruption, [one method at a time](https://github.com/aws/karpenter-core/blob/main/pkg/controllers/disruption/controller.go#L88-L100). 

There's a `Method` interface() that consists of 
- ShouldDisrupt(): a func used to filter if a node should be disrupted
- ComputeCommand(): returns a Command, consisting of a list of commands, and a list of replacements. this should consist of a scheduling simulation that determines if we need to spin up a replacement. 
- Type(): just a string representation for metrics/logs
- ConsolidationType(): used for Consolidations, can just ignore.

jeesmon:
[...]

PR #845, opened 2023-12-07
[WIP]feat: Add UserAnnotedExpiration as a new disruption method 

tasdikrahman:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/688

This PR introduces a custom disruption method which is evaluating candidates which need to be deprovisioned, which have been added an annotation 

```
karpenter.sh/voluntary-disruption=drifted
```
to the Node object. 

### DIFF
--- a/pkg/controllers/disruption/custom_expiration.go
+++ b/pkg/controllers/disruption/custom_expiration.go
@@ -27,9 +27,9 @@ import (
 	"github.com/aws/karpenter-core/pkg/metrics"
 )
 
-// CustomExpiration is a subreconciler that deletes custom marked candidates, based on node annotations.
+// UserAnnotedExpiration is a subreconciler that deletes custom marked candidates, based on node annotations.
 // Expiration will respect TTLSecondsAfterEmpty
-type CustomExpiration struct {
+type UserAnnotedExpiration struct {
 	clock       clock.Clock
 	kubeClient  client.Client
 	cluster     *state.Cluster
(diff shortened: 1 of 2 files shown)


njtran:
Hey @tasdikrahman, as I recall, we wanted to start off with a small design before continuing with this. Was there anyone you had talked to about this while I was on vacation? If not, let's see if we can sync up and get a design in this PR. This should help align us on naming and making sure we're considering how this impacts any other controllers in the project.

pkg/controllers/disruption/custom_expiration.go
```diff
@@ -0,0 +1,74 @@
+/*
```
jonathan-innis, line 1: I'm anticipating that we probably don't want to do something as verbose as `karpenter.sh/voluntarily-disrupted=drifted`. Realistically, it should probably something simple like `karpenter.sh/disrupt=true` since we've already got an annotation that's `karpenter.sh/do-not-disrupt=false`

garvinp-stripe:
I wonder if it would be better to start off with a quick design doc for RFC? This is adding a new controller and method of disruption. I am not fully sure how this fits in with other methods

tasdikrahman:
Hey folks, sorry for the delay, I will work on sharing a design doc first on this and share it here to align on the approach and changes. I opened the PR as a wip. @njtran @garvinp-stripe @jonathan-innis

njtran:
Thanks. @tasdikrahman I'll set this as a Draft in the meantime.

njtran:
Hey @tasdikrahman how are you doing? Happy to keep you assigned if you want, but want to make sure that you're making progress. I want to make sure the issue is tracked properly based on if you're getting progress here.

njtran:
@tasdikrahman hey, it seems like it's been a while since you've been able to work on this. I'll close this out and unassign you from the issue, but feel free to take it up when you have more time.

PR #846, opened 2023-12-07
chore: add webhook validation for budgets

njtran:
Fixes #N/A 

**Description**
Adds in validations for budgets and the associated tests

make presubmit


pkg/apis/v1beta1/nodepool_validation.go
```diff
@@ -94,18 +98,52 @@ func (in *NodeClaimTemplate) validateRequirementsNodePoolKeyDoesNotExist() (errs
...
-		return errs.Also(apis.ErrInvalidValue("cannot be negative", "consolidationTTL"))
+		return errs.Also(apis.ErrInvalidValue(in.ConsolidateAfter.Duration, "consolidationTTL", "cannot be negative"))
 	}
 	if in.ConsolidateAfter != nil && in.ConsolidateAfter.Duration != nil && in.ConsolidationPolicy == ConsolidationPolicyWhenUnderutilized {
 		return errs.Also(apis.ErrGeneric("consolidateAfter cannot be combined with consolidationPolicy=WhenUnderutilized"))
 	}
 	if in.ConsolidateAfter == nil && in.ConsolidationPolicy == ConsolidationPolicyWhenEmpty {
 		return errs.Also(apis.ErrGeneric("consolidateAfter must be specified with consolidationPolicy=WhenEmpty"))
 	}
+	if len(in.Budgets) > 50 {
```
jonathan-innis, line 115: Is the maxItems parameter enforced across clusters that are pre-1.25 so that we don't need to check the budget length here as well. Shouldn't be a blocker to merge this (since the worst is that we will do this check twice), but would be good to know if it's necessary.

pkg/apis/v1beta1/nodepool_validation.go
```diff
@@ -94,18 +98,52 @@ func (in *NodeClaimTemplate) validateRequirementsNodePoolKeyDoesNotExist() (errs
...
+	}
+	return errs
+}
+
+//nolint:gocyclo
+func (in *Budget) validate() (errs *apis.FieldError) {
+	if (in.Crontab != nil && in.Duration == nil) || (in.Crontab == nil && in.Duration != nil) {
+		return apis.ErrGeneric("crontab and duration must be specified together")
+	}
+	if matched, err := regexp.MatchString("^((100|[0-9]{1,2})%|[0-9]+)$", in.MaxUnavailable); err != nil || !matched {
```
jonathan-innis, line 132: This pattern is enforced by the OpenAPI spec directly so I don't think that we need to do validation here.

pkg/apis/v1beta1/nodepool_validation.go
```diff
@@ -94,18 +98,52 @@ func (in *NodeClaimTemplate) validateRequirementsNodePoolKeyDoesNotExist() (errs
...
+		return apis.ErrInvalidValue(in.MaxUnavailable, "maxUnavailable", "must be a valid intorstring")
+	}
+	if in.Crontab != nil {
+		if _, err := cron.ParseStandard(lo.FromPtr(in.Crontab)); err != nil {
+			return apis.ErrInvalidValue(in.Crontab, "crontab", fmt.Sprintf("invalid crontab %s", err))
+		}
+	}
+	if in.Duration != nil {
```
jonathan-innis, line 140: Can we double-check if we actually need to check some of these things, given that things that are just OpenAPI validations and aren't part of CEL can just be passed off to the apiserver. I'd like to avoid duplicating checks here twice if the request wouldn't even succeed at the apiserver

jonathan-innis: Discussed offline. Checking that durations are under a minute should be enforced at the pattern matching layer. I'm fine either way but we probably don't need this in the webhook validation.

jonathan-innis (approved):
Nice work 🎉 
/lgtm
/approve

PR #849, opened 2023-12-08
feat: Disruption Budgets

njtran:
Fixes #N/A 

**Description**
This adds in the check within the disruption controller to take into account NodePool.Disruption.Budgets[] when considering if a node can be disrupted. This directly pertains to the nodes that have a deletionTimestamp set and that have the NoSchedule Termination Taint. 

draft at the moment.


pkg/controllers/state/statenode.go
```diff
@@ -296,6 +296,7 @@ func (in *StateNode) MarkedForDeletion() bool {
 	return in.markedForDeletion ||
 		(in.NodeClaim != nil && !in.NodeClaim.DeletionTimestamp.IsZero()) ||
 		(in.Node != nil && in.NodeClaim == nil && !in.Node.DeletionTimestamp.IsZero())
+	// || in.Taints().Has(RequireNoScheduleTaint())
```
jonathan-innis, line 299: Is this still meant to be commented out? Or are we just handling this explicitly through mark/unmark calls in the code?

pkg/controllers/disruption/emptynodeconsolidation.go
```diff
@@ -79,14 +90,21 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, candidates
 	// We do this so that we can re-validate that the candidates that were computed before we made the decision are the same
 	candidatesToDelete := mapCandidates(cmd.candidates, validationCandidates)
 
+	postValidationMapping, err := buildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient)
```
jonathan-innis, line 93: This validation delay here definitely scares me. What are the edge-cases here w.r.t. to re-building the disruption budgets mapping? Are there any risks that this might continually fail for some reason?

pkg/controllers/disruption/helpers.go
```diff
@@ -191,6 +191,45 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
...
+		}
+		numNodes[nodePool]++
+	}
+
+	for i := range nodePoolList.Items {
+		nodePool := nodePoolList.Items[i]
+		disruptions, err := nodePool.GetAllowedDisruptions(ctx, clk, numNodes[nodePool.Name])
+		if err != nil {
+			// Don't return here as we shouldn't block all disruption within the cluster on one misconfigured nodepool.
+			logging.FromContext(ctx).With("nodePool", nodePool.Name).Debugf("getting allowed disruptions, %w", err)
```
jonathan-innis, line 225:
```suggestion
			logging.FromContext(ctx).With("nodepool", nodePool.Name).Errorf("getting allowed disruptions, %v", err)
```

Shouldn't this be an error? Since we should scream about this not being correct?

pkg/test/expectations/expectations.go
```diff
@@ -561,10 +561,12 @@ func ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx context.Context,
 	ExpectMakeNodeClaimsInitialized(ctx, c, nodeClaims...)
 
 	// Inform cluster state about node and nodeclaim readiness
-	for _, n := range nodes {
+	for i := range nodes {
+		n := nodes[i]
```
jonathan-innis, line 565: Not a big deal, mainly just curious why we chose to change this

pkg/controllers/state/statenode.go
njtran: Good question. I was thinking about adding it in, but I don't think it honestly gets any value, considering we add and remove taints whenever we mark and unmark for deletion. What are your thoughts here?

pkg/controllers/disruption/helpers.go
njtran: Good catch!

pkg/test/expectations/expectations.go
njtran: mistaken change for testing, will move it back 👍

pkg/controllers/disruption/emptynodeconsolidation.go
njtran: What about the delay scares you? We're checking the budgets before and after the validation, which I think is the best we can do.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -58,6 +59,10 @@ var _ = Describe("Consolidation", func() {
 			Spec: v1beta1.NodePoolSpec{
 				Disruption: v1beta1.Disruption{
 					ConsolidationPolicy: v1beta1.ConsolidationPolicyWhenUnderutilized,
+					// Disrupt away!
```
jonathan-innis, line 62: 😆

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -71,11 +76,245 @@ var _ = Describe("Consolidation", func() {
...
+				ExpectApplied(ctx, env.Client, pods[i])
+				ExpectManualBinding(ctx, env.Client, pods[i], nodes[i])
+			}
+
+			// inform cluster state about nodes and nodeclaims
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, nodes, nodeClaims)
+
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			// Reconcile 5 times, enqueuing 3 commands total.
```
jonathan-innis, line 248: This comment doesn't match-up. Just reading over it, looks like we requeue 15 times here.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -71,11 +76,245 @@ var _ = Describe("Consolidation", func() {
...
+			for i := 0; i < 15; i++ {
+				ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
```
jonathan-innis, line 250: Will this re-trigger the verify action each time? Just reading through it, I'm assuming that each of these needs a verify action and it looks like we are only sending the verification out here once.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -71,11 +76,245 @@ var _ = Describe("Consolidation", func() {
...
+			}
+			wg.Wait()
+
+			// Execute all commands in the queue, only deleting 3 nodes
+			for i := 0; i < 15; i++ {
+				ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
+			}
+			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(7))
+		})
+		It("should allow 2 nodes from each nodePool to be deleted", func() {
```
jonathan-innis, line 260: Really like this test-case. Super good coverage here 🎉

pkg/controllers/disruption/drift_test.go
```diff
@@ -73,505 +78,616 @@ var _ = Describe("Drift", func() {
...
-		ExpectExists(ctx, env.Client, nodeClaim)
-	})
-	It("should continue to the next drifted node if the first cannot reschedule all pods", func() {
-		pod := test.Pod(test.PodOptions{
-			ResourceRequirements: v1.ResourceRequirements{
-				Requests: v1.ResourceList{
-					v1.ResourceCPU: resource.MustParse("150"),
-				},
-			},
+	Context("Budgets", func() {
```
jonathan-innis, line 81: Do we need any handlings for budgets where there are pods bound?

pkg/controllers/disruption/expiration_test.go
```diff
@@ -72,476 +75,585 @@ var _ = Describe("Expiration", func() {
 		})
 		nodeClaim.StatusConditions().MarkTrue(v1beta1.Expired)
 	})
-	It("should ignore nodes without the expired status condition", func() {
-		_ = nodeClaim.StatusConditions().ClearCondition(v1beta1.Expired)
-		ExpectApplied(ctx, env.Client, nodeClaim, node, nodePool)
-
-		// inform cluster state about nodes and nodeclaims
-		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{node}, []*v1beta1.NodeClaim{nodeClaim})
+	Context("Budgets", func() {
```
jonathan-innis, line 78: Same question here as with drift: do we need test coverage for cases where we check that we can disrupt with budgets when the pods are bound?

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
```
jonathan-innis, line 28: Maybe you'll cover this with the helper testing, but I'm curious if we should have coverage over checking cases where we have nodes getting marked for deletion for some other reason and these should be counted against the disruption budget.

pkg/controllers/disruption/helpers.go
```diff
@@ -191,6 +191,45 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
...
+
+	for i := range nodePoolList.Items {
+		nodePool := nodePoolList.Items[i]
+		disruptions, err := nodePool.GetAllowedDisruptions(ctx, clk, numNodes[nodePool.Name])
+		if err != nil {
+			// Don't return here as we shouldn't block all disruption within the cluster on one misconfigured nodepool.
+			logging.FromContext(ctx).With("nodePool", nodePool.Name).Errorf("getting allowed disruptions, %w", err)
+		}
+		// Subtract the allowed number of disruptions from the number of already deleting nodes.
+		disruptionBudgetMapping[nodePool.Name] = disruptions - deleting[nodePool.Name]
```
jonathan-innis, line 228: Discussed offline but this has the potential to go negative, which makes other logic in the code that assumes that this value _can't_ go negative, go haywire.

pkg/controllers/disruption/consolidation_test.go
njtran: Good catch, I hadn't pushed up my recent testing changes. The good thing is that test-wise, this shouldn't impact anything.

pkg/controllers/disruption/consolidation_test.go
njtran: Good catch! This would have definitely slowed down testing if not caught.

pkg/controllers/disruption/consolidation_test.go
njtran: We only do MarkForDeletion within the disruption controller, so I don't think we need to worry about this. Even if there were, I don't think it'd be unreasonable to expect this to be considered as part of the budget.

pkg/controllers/disruption/helpers.go
```diff
@@ -191,6 +191,45 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
 	return lo.Filter(candidates, func(c *Candidate, _ int) bool { return shouldDeprovision(ctx, c) }), nil
 }
 
+// BuildDisruptionBudgets will return a map for nodePoolName -> numAllowedDisruptions and an error
+func BuildDisruptionBudgets(ctx context.Context, cluster *state.Cluster, clk clock.Clock, kubeClient client.Client) (map[string]int, error) {
```
ellistarn, line 195: I wonder if it would simplify things to build a `state.budgets` struct that tracks this state (w/ state.cluster) as a member variable. You could write queries against this, and potentially keep errorable API calls out of the critical path.

pkg/controllers/disruption/helpers.go
```diff
@@ -191,6 +191,45 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
...
+		nodePool := node.Labels()[v1beta1.NodePoolLabelKey]
+		if node.MarkedForDeletion() {
+			deleting[nodePool]++
+		}
+		numNodes[nodePool]++
+	}
+
+	for i := range nodePoolList.Items {
+		nodePool := nodePoolList.Items[i]
+		disruptions, err := nodePool.GetAllowedDisruptions(ctx, clk, numNodes[nodePool.Name])
```
ellistarn, line 222: Is there any possible way to make this nonerrorable? Errors that are ignored and logged has been a source of invariant violations in the past.

pkg/controllers/disruption/helpers.go
jonathan-innis: > You could write queries against this, and potentially keep errorable API calls out of the critical path

pkg/controllers/disruption/helpers.go
```diff
@@ -191,6 +191,45 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
...
+		if err != nil {
+			// Don't return here as we shouldn't block all disruption within the cluster on one misconfigured nodepool.
+			logging.FromContext(ctx).With("nodePool", nodePool.Name).Errorf("getting allowed disruptions, %w", err)
+		}
+		// Subtract the allowed number of disruptions from the number of already deleting nodes.
+		disruptionBudgetMapping[nodePool.Name] = lo.Clamp(disruptions-deleting[nodePool.Name], 0, math.MaxInt32)
```
jonathan-innis, line 228: Worth some comment on why we need to clamp here?

pkg/controllers/disruption/suite_test.go
```diff
@@ -416,6 +278,133 @@ var _ = Describe("Disruption Taints", func() {
 	})
 })
 
+/*
+	nodePoolList, err := nodepoolutil.List(ctx, kubeClient)
```
jonathan-innis, line 282: I figure you will remove this later, but just want to callout that this is still commented out

pkg/controllers/disruption/consolidation_test.go
njtran: Ah I thought you meant if there were other controllers calling `statenode.MarkForDeletion`. I've added tests that incorporate if a node is deleting, which would then be updated in ClusterState to be seen as MarkedForDeletion.

pkg/controllers/disruption/helpers.go
njtran: Talked offline, considering validating these budgets at the start of the loop and removing this error path.

pkg/controllers/disruption/suite_test.go
njtran: Thanks for reminder 👍

pkg/apis/v1beta1/nodepool.go
```diff
@@ -248,12 +241,19 @@ func (in *Budget) GetAllowedDisruptions(c clock.Clock, numNodes int) (int, error
...
+	// sometimes exceed the disruption budget. This is the same as how Kubernetes
+	// handles MaxUnavailable with PDBs. Take the case with 5% disruptions, but
+	// 10 nodes. Karpenter will opt to allow 1 node to be disrupted, rather than
+	// blocking all disruptions for this nodepool.
+	res, err := intstr.GetScaledValueFromIntOrPercent(lo.ToPtr(val), numNodes, true)
 	if err != nil {
-		// Should almost never happen since this is validated when the nodepool is applied
-		return 0, fmt.Errorf("getting intstr scaled value, %w", err)
+		// Should never happen since this is validated when the nodepool is applied
+		// If this value is incorrectly formatted, fail close, since we don't know what
```
jonathan-innis, line 252:
```suggestion
		// If this value is incorrectly formatted, fail closed, since we don't know what
```
nit

pkg/apis/v1beta1/nodepool_budgets_test.go
```diff
@@ -72,56 +72,48 @@ var _ = Describe("Budgets", func() {
...
 			budgets[1].Schedule = lo.ToPtr("@yearly")
 			budgets[2].Schedule = lo.ToPtr("@yearly")
 			budgets[3].Schedule = lo.ToPtr("@yearly")
-			min, err := nodePool.GetAllowedDisruptions(ctx, fakeClock, 100)
-			Expect(err).To(Succeed())
+			min := nodePool.GetAllowedDisruptions(ctx, fakeClock, 100)
 			Expect(min).To(BeNumerically("==", math.MaxInt32))
 		})
-		It("should fail and return zero values if a schedule is invalid", func() {
+		It("should return zero values if a schedule is invalid", func() {
```
jonathan-innis, line 93: Looks like we are testing the same test case twice?

jonathan-innis: Should we also have a check when the value is invalid, not just the schedule?

pkg/controllers/disruption/controller.go
```diff
@@ -242,3 +248,36 @@ func (c *Controller) logAbnormalRuns(ctx context.Context) {
...
+
+// logInvalidBudgets will log if there are any invalid schedules detected
+func (c *Controller) logInvalidBudgets(ctx context.Context) {
+	nodePoolList, err := nodepoolutil.List(ctx, c.kubeClient)
+	if err != nil {
+		logging.FromContext(ctx).Debugf("listing nodepools, %s", err)
+	}
+	var buf bytes.Buffer
+	for _, np := range nodePoolList.Items {
+		for _, budget := range np.Spec.Disruption.Budgets {
```
jonathan-innis, line 260: Seems like a lot of parsing the same thing over again. Is there a way to unify this code into `v1beta1` so that we are using duplications of the code (one just returns an error while the other checks for the presence of the error and fails closed?)

pkg/apis/v1beta1/nodepool_budgets_test.go
njtran: one is the budget test, and one is the nodepool test. It's effectively the same, so I can go either way on including it.

pkg/controllers/disruption/helpers.go
ellistarn: Feel free to ignore if it doesn't make sense. This just set off my static helper smell detector.

pkg/controllers/disruption/drift.go
```diff
@@ -82,16 +82,35 @@ func (d *Drift) ComputeCommand(ctx context.Context, candidates ...*Candidate) (C
 		consolidationTypeLabel: d.ConsolidationType(),
 	}).Set(float64(len(candidates)))
 
+	// Do a quick check through the candidates to see if they're empty.
+	// For each candidate that is empty with a nodePool allowing its disruption
+	// add it to the existing command.
+	empty := make([]*Candidate, 0, len(candidates))
```
garvinp-stripe, line 88: I see this code in drift and expiration. Is there a reason to do this per controller vs just moving emptiness disruption evaluation higher https://github.com/kubernetes-sigs/karpenter/blob/534b2711a96130aff85b0addb7ca0173f6135831/pkg/controllers/disruption/controller.go#L72?

I assume atm disruption controller runs through all the methods in order https://github.com/kubernetes-sigs/karpenter/blob/534b2711a96130aff85b0addb7ca0173f6135831/pkg/controllers/disruption/controller.go#L127

pkg/controllers/disruption/helpers.go
garvinp-stripe, line 227: Sidenote: I think we know about this implicitly because this is how PDB works but we may want to clarify how we count "disruptions" per cycle. For example, this code would count a node in deleting state that was marked for deletion in the prior disruption cycle and will continue to drain the budget until it resolves. I believe this is the same behavior as PDB but I don't think it was document explicitly.

pkg/controllers/disruption/helpers.go
```diff
@@ -191,6 +191,44 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
...
+	for i := range nodePoolList.Items {
+		nodePool := nodePoolList.Items[i]
+		disruptions := nodePool.GetAllowedDisruptions(ctx, clk, numNodes[nodePool.Name])
+		// Subtract the allowed number of disruptions from the number of already deleting nodes.
+		// Floor the value since the number of deleting nodes can exceed the number of allowed disruptions.
+		// Allowing this value to be negative breaks assumptions in the code used to calculate how
+		// many nodes can be disrupted.
+		disruptionBudgetMapping[nodePool.Name] = lo.Clamp(disruptions-deleting[nodePool.Name], 0, math.MaxInt32)
+	}
+	return disruptionBudgetMapping, nil
```
garvinp-stripe, line 229: Is it possible to add a debug log statement here that dumps the disruption budget mapping? It would likely help with debugging if we were not seeing nodes being killed

pkg/controllers/disruption/drift.go
jonathan-innis: +1. I think that we should think about how to coalesce a lot of this going forward. Feels out of scope of this particular PR but definitely something that we should do

pkg/controllers/disruption/helpers.go
jonathan-innis: This is a great callout. We should think about metrics on for the number of nodes that we can currently disrupt

pkg/controllers/disruption/drift.go
garvinp-stripe: Unless I am misunderstanding that a node that is drifted/ expired + empty is at a higher priority than just an empty node.

njtran: Agreed. Something we definitely want to do, which could be blended into the custom termination reason as well.

pkg/controllers/disruption/drift.go
```diff
@@ -82,16 +82,35 @@ func (d *Drift) ComputeCommand(ctx context.Context, candidates ...*Candidate) (C
...
 		return Command{
 			candidates: empty,
 		}, nil
 	}
 
 	for _, candidate := range candidates {
+		// If the disruption budget doesn't allow this candidate to be disrupted,
+		// continue to the next candidate. We don't need to decrement any budget
+		// counter since drift commands can only have one candidate.
+		if disruptionBudgetMapping[candidate.nodePool.Name] == 0 {
```
garvinp-stripe, line 111: Is it possible to emit a debug log or metric around the number of candidates that were "left behind" because budget has been drained.

garvinp-stripe, line 111: Is it possible to emit a debug log or metric around the number of candidates that were "left behind" because budget has been drained.

pkg/controllers/disruption/emptiness.go
```diff
@@ -48,17 +48,32 @@ func (e *Emptiness) ShouldDisrupt(_ context.Context, c *Candidate) bool {
...
+			continue
+		}
+		// If there's disruptions allowed for the candidate's nodepool,
+		// add it to the list of candidates, and decrement the budget.
+		if disruptionBudgetMapping[candidate.nodePool.Name] > 0 {
+			empty = append(empty, candidate)
+			disruptionBudgetMapping[candidate.nodePool.Name]--
+		}
+	}
+
```
garvinp-stripe, line 74: similar comment around log/ metric around number of left behind candidates

pkg/controllers/disruption/drift.go
```diff
@@ -82,16 +82,35 @@ func (d *Drift) ComputeCommand(ctx context.Context, candidates ...*Candidate) (C
 		consolidationTypeLabel: d.ConsolidationType(),
 	}).Set(float64(len(candidates)))
 
+	// Do a quick check through the candidates to see if they're empty.
+	// For each candidate that is empty with a nodePool allowing its disruption
+	// add it to the existing command.
+	empty := make([]*Candidate, 0, len(candidates))
+	for _, candidate := range candidates {
+		if len(candidate.pods) > 0 {
```
garvinp-stripe, line 90: It feels like this emptiness function can be refactor out? It looks the same in all the controllers

pkg/controllers/disruption/helpers.go
njtran: We'll definitely document this explicitly in the website. Would you like to see this comment expanded? 

I was hoping to drive the definition of this through the API comment itself here: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1beta1/nodepool.go#L104-L112 but I'm realizing this comment is not too descriptive and partly wrong, so I'll make sure to update this too.

pkg/controllers/disruption/helpers.go
njtran: Definitely. There's an `eligibility` metric here which may be conflated with if it can be disrupted w.r.t. a disruption budget. I definitely think we should include this, and I'll make sure we think about this in a follow-up.

pkg/controllers/disruption/helpers.go
garvinp-stripe:
```
// Nodes dictates how many NodeClaims owned by this NodePool
// can be terminating at once
``` 
	
I think is just a little to broad/ up for interpretation because what we are actually doing isn't terminating all these node at once and some nodes can take longer to terminate. I think should be something along the lines of `maximum number of nodes a node pool can tolerate being disrupted`

njtran: Will adjust. Thanks for the feedback!

pkg/controllers/disruption/drift.go
njtran: Definitely will emit a metric as a followup. I worry about a log here, since this may get very noisy, and we'd need to duplicate this across each method, and it would probably need to log per-provisioner, per-method, per-disruption loop, which I could envision getting noisy. @jonathan-innis and I were talking about modeling this as a status condition per-node. Thoughts on this - metric for followup, and status condition for node - rather than emitting a log?

pkg/controllers/disruption/drift.go
njtran: Agreed. Good callout.

pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -59,11 +59,11 @@ spec:
                             type: string
                           nodes:
                             default: 10%
-                            description: 'Nodes dictates how many NodeClaims owned by this NodePool can be terminating at once. It must be set. This only considers NodeClaims with the karpenter.sh/disruption taint. We can''t use an intstr.IntOrString since kubebuilder doesn''t support pattern checking for int nodes for IntOrString nodes. Ref: https://github.com/kubernetes-sigs/controller-tools/blob/55efe4be40394a288216dab63156b0a64fb82929/pkg/crd/markers/validation.go#L379-L388'
+                            description: 'Nodes dictates the maximum number of NodeClaims owned by this Node that can be terminating at once. This is calculated by counting nodes that have a deletion timestamp set, or are marked for deletion in memory. This field is required. This cannot be of type intstr.IntOrString since kubebuilder doesn''t support pattern checking for int nodes for IntOrString nodes. Ref: https://github.com/kubernetes-sigs/controller-tools/blob/55efe4be40394a288216dab63156b0a64fb82929/pkg/crd/markers/validation.go#L379-L388'
```
jonathan-innis, line 62:
```suggestion
                            description: 'Nodes dictates the maximum number of NodeClaims owned by this NodePool that can be terminating at once. This is calculated by counting nodes that have a deletion timestamp set, or are actively being deleted by Karpenter. This field is required when specifying a budget. This cannot be of type intstr.IntOrString since kubebuilder doesn''t support pattern checking for int nodes for IntOrString nodes. Ref: https://github.com/kubernetes-sigs/controller-tools/blob/55efe4be40394a288216dab63156b0a64fb82929/pkg/crd/markers/validation.go#L379-L388'
```

I think that the marked for deletion construct is probably not well-known to someone who is just reading through the CRD docs. Best to keep this written in terms of well-known API concepts

pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -59,11 +59,11 @@ spec:
...
                             pattern: ^((100|[0-9]{1,2})%|[0-9]+)$
                             type: string
                           schedule:
-                            description: Schedule specifies when a budget begins being active, using the upstream cronjob syntax. If omitted, the budget is always active. Currently timezones are not supported. This is required if Duration is set.
+                            description: Schedule specifies when a budget begins being active, following the upstream cronjob syntax. If omitted, the budget is always active. Like in upstream, timezones are not supported. This is required if Duration is set.
```
jonathan-innis, line 66:
```suggestion
                            description: Schedule specifies when a budget begins being active. If omitted, the budget is always active. Timezones are not supported. This field is required if Duration is set.
```

You say upstream a few times and I feel like we don't need to say that we are following someone else. I get that this might make it easier to reason about, but we effectively _are_ upstream now, so we are just following convention in general

pkg/apis/v1beta1/nodepool_budgets_test.go
```diff
@@ -70,43 +70,45 @@ var _ = Describe("Budgets", func() {
 			},
 		}
 	})
-	Context("NodePool/AllowedDisruptions", func() {
+	Context("NodePool/MustGetAllowedDisruptions", func() {
```
jonathan-innis, line 73: Do these `Context` blocks need to have `NodePool` in front of them? This is all NodePool testing so I think it's fine to just say `MustGetAllowedDisruptions`

pkg/controllers/disruption/controller.go
```diff
@@ -242,3 +245,20 @@ func (c *Controller) logAbnormalRuns(ctx context.Context) {
...
+
+// logInvalidBudgets will log if there are any invalid schedules detected
+func (c *Controller) logInvalidBudgets(ctx context.Context) {
+	nodePoolList, err := nodepoolutil.List(ctx, c.kubeClient)
+	if err != nil {
+		logging.FromContext(ctx).Debugf("listing nodepools, %s", err)
+	}
+	var buf bytes.Buffer
+	for _, np := range nodePoolList.Items {
+		if _, err := np.GetAllowedDisruptions(ctx, c.clock, 100); err != nil {
```
jonathan-innis, line 257: nit: Maybe a comment here to say that you are sending through a dummy value to check the validity of the disruption; otherwise, I'm left guessing where 100 came from

pkg/controllers/disruption/drift.go
garvinp-stripe: Ya that sounds good, I think I just want a way to know if we are falling behind and by how much in addition to which are the nodes falling behind

jonathan-innis (approved):
/lgtm
/approve

PR #852, opened 2023-12-08
feat: support new topologySpread scheduling constraints

### LINKED ISSUE #430: Support new topologySpread scheduling constraints
### Description

**Observed Behavior**:

Karpenter doesn't support:
- matchLabelKeys
- nodeAffinityPolicy
- nodeTaintsPolicy

These fields were introduced into beta in 1.27.

**Expected Behavior**:

**Versions**:
- Chart Version: 

jmdeal:
Fixes #430 

**Description**
This PR adds support for the following topology spread constraint fields:
- `matchLabelKeys`
- `nodeAffinityPolicy`
- `nodeTaintsPolicy`

`make test`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/topology.go
+++ b/pkg/controllers/provisioning/scheduling/topology.go
@@ -22,6 +22,7 @@ import (
 	"math"
 
 	"github.com/awslabs/operatorpkg/option"
+	"github.com/samber/lo"
 	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -32,6 +33,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 	"sigs.k8s.io/karpenter/pkg/utils/pod"
@@ -40,31 +43,41 @@ import (
 
 type Topology struct {
 	kubeClient client.Client
-	// Both the topologies and inverseTopologies are maps of the hash from TopologyGroup.Hash() to the topology group
+	// Both the topologyGroups and inverseTopologies are maps of the hash from TopologyGroup.Hash() to the topology group
 	// itself. This is used to allow us to store one topology group that tracks the topology of many pods instead of
 	// having a 1<->1 mapping between topology groups and pods owned/selected by that group.
-	topologies map[uint64]*TopologyGroup
+	topologyGroups map[uint64]*TopologyGroup
 	// Anti-affinity works both ways (if a zone has a pod foo with anti-affinity to a pod bar, we can't schedule bar to
 	// that zone, even though bar has no anti affinity terms on it. For this to work, we need to separately track the
 	// topologies of pods with anti-affinity terms, so we can prevent scheduling the pods they have anti-affinity to
 	// in some cases.
-	inverseTopologies map[uint64]*TopologyGroup
+	inverseTopologyGroups map[uint64]*TopologyGroup
 	// The universe of domains by topology key
 	domainGroups map[string]TopologyDomainGroup
 	// excludedPods are the pod UIDs of pods that are excluded from counting.  This is used so we can simulate
 	// moving pods to prevent them from being double counted.
 	excludedPods sets.Set[string]
 	cluster      *state.Cluster
+	stateNodes   []*state.StateNode
 }
 
-func NewTopology(ctx context.Context, kubeClient client.Client, cluster *state.Cluster, domainGroups map[string]TopologyDomainGroup, pods []*corev1.Pod) (*Topology, error) {
+func NewTopology(
+	ctx context.Context,
+	kubeClient client.Client,
+	cluster *state.Cluster,
+	stateNodes []*state.StateNode,
+	nodePools []*v1.NodePool,
+	instanceTypes map[string][]*cloudprovider.InstanceType,
+	pods []*corev1.Pod,
+) (*Topology, error) {
 	t := &Topology{
-		kubeClient:        kubeClient,
-		cluster:           cluster,
-		domainGroups:      domainGroups,
-		topologies:        map[uint64]*TopologyGroup{},
-		inverseTopologies: map[uint64]*TopologyGroup{},
-		excludedPods:      sets.New[string](),
+		kubeClient:            kubeClient,
+		cluster:               cluster,
+		stateNodes:            stateNodes,
+		domainGroups:          buildDomainGroups(nodePools, instanceTypes),
+		topologyGroups:        map[uint64]*TopologyGroup{},
+		inverseTopologyGroups: map[uint64]*TopologyGroup{},
+		excludedPods:          sets.New[string](),
 	}
 
 	// these are the pods that we intend to schedule, so if they are currently in the cluster we shouldn't count them for
@@ -81,6 +94,46 @@ func NewTopology(ctx context.Context, kubeClient client.Client, cluster *state.C
 		return nil, errs
 	}
 	return t, nil
+}
+
+func buildDomainGroups(nodePools []*v1.NodePool, instanceTypes map[string][]*cloudprovider.InstanceType) map[string]TopologyDomainGroup {
+	nodePoolIndex := lo.SliceToMap(nodePools, func(np *v1.NodePool) (string, *v1.NodePool) {
+		return np.Name, np
+	})
+	domainGroups := map[string]TopologyDomainGroup{}
+	for npName, its := range instanceTypes {
+		np := nodePoolIndex[npName]
+		for _, it := range its {
+			// We need to intersect the instance type requirements with the current nodePool requirements.  This
+			// ensures that something like zones from an instance type don't expand the universe of valid domains.
+			requirements := scheduling.NewNodeSelectorRequirementsWithMinValues(np.Spec.Template.Spec.Requirements...)
+			requirements.Add(scheduling.NewLabelRequirements(np.Spec.Template.Labels).Values()...)
+			requirements.Add(it.Requirements.Values()...)
+
+			for topologyKey, requirement := range requirements {
+				if _, ok := domainGroups[topologyKey]; !ok {
+					domainGroups[topologyKey] = NewTopologyDomainGroup()
+				}
+				for _, domain := range requirement.Values() {
+					domainGroups[topologyKey].Insert(domain, np.Spec.Template.Spec.Taints...)
+				}
+			}
+		}
+
+		requirements := scheduling.NewNodeSelectorRequirementsWithMinValues(np.Spec.Template.Spec.Requirements...)
+		requirements.Add(scheduling.NewLabelRequirements(np.Spec.Template.Labels).Values()...)
+		for key, requirement := range requirements {
+			if requirement.Operator() == corev1.NodeSelectorOpIn {
+				if _, ok := domainGroups[key]; !ok {
+					domainGroups[key] = NewTopologyDomainGroup()
+				}
+				for _, value := range requirement.Values() {
+					domainGroups[key].Insert(value, np.Spec.Template.Spec.Taints...)
+				}
+			}
+		}
+	}
+	return domainGroups
 }
 
 // topologyError allows lazily generating the error string in the topology error.  If a pod fails to schedule, most often
@@ -101,7 +154,7 @@ func (t topologyError) Error() string {
 // relaxation of a preference to properly break the topology <-> owner relationship so that the preferred topology will
 // no longer influence scheduling.
 func (t *Topology) Update(ctx context.Context, p *corev1.Pod) error {
-	for _, topology := range t.topologies {
+	for _, topology := range t.topologyGroups {
 		topology.RemoveOwner(p.UID)
 	}
 
@@ -120,11 +173,11 @@ func (t *Topology) Update(ctx context.Context, p *corev1.Pod) error {
 	for _, tg := range append(topologies, affinities...) {
 		hash := tg.Hash()
 		// Avoid recomputing topology counts if we've already seen this group
-		if existing, ok := t.topologies[hash]; !ok {
+		if existing, ok := t.topologyGroups[hash]; !ok {
 			if err := t.countDomains(ctx, tg); err != nil {
 				return err
 			}
-			t.topologies[hash] = tg
+			t.topologyGroups[hash] = tg
 		} else {
 			tg = existing
 		}
@@ -134,27 +187,27 @@ func (t *Topology) Update(ctx context.Context, p *corev1.Pod) error {
 }
 
 // Record records the topology changes given that pod p schedule on a node with the given requirements
-func (t *Topology) Record(p *corev1.Pod, requirements scheduling.Requirements, compatabilityOptions ...option.Function[scheduling.CompatibilityOptions]) {
+func (t *Topology) Record(p *corev1.Pod, taints []corev1.Taint, requirements scheduling.Requirements, compatabilityOptions ...option.Function[scheduling.CompatibilityOptions]) {
 	// once we've committed to a domain, we record the usage in every topology that cares about it
-	for _, tc := range t.topologies {
-		if tc.Counts(p, requirements, compatabilityOptions...) {
-			domains := requirements.Get(tc.Key)
-			if tc.Type == TopologyTypePodAntiAffinity {
+	for _, tg := range t.topologyGroups {
+		if tg.Counts(p, taints, requirements, compatabilityOptions...) {
+			domains := requirements.Get(tg.Key)
+			if tg.Type == TopologyTypePodAntiAffinity {
 				// for anti-affinity topologies we need to block out all possible domains that the pod could land in
-				tc.Record(domains.Values()...)
+				tg.Record(domains.Values()...)
 			} else {
 				// but for affinity & topology spread, we can only record the domain if we know the specific domain we land in
 				if domains.Len() == 1 {
-					tc.Record(domains.Values()[0])
+					tg.Record(domains.Values()[0])
 				}
 			}
 		}
 	}
 	// for anti-affinities, we record where the pods could be, even if
 	// requirements haven't collapsed to a single value.
-	for _, tc := range t.inverseTopologies {
-		if tc.IsOwnedBy(p.UID) {
-			tc.Record(requirements.Get(tc.Key).Values()...)
+	for _, tg := range t.inverseTopologyGroups {
+		if tg.IsOwnedBy(p.UID) {
+			tg.Record(requirements.Get(tg.Key).Values()...)
 		}
 	}
 }
@@ -163,9 +216,9 @@ func (t *Topology) Record(p *corev1.Pod, requirements scheduling.Requirements, c
 // affinities, anti-affinities or inverse anti-affinities.  The nodeHostname is the hostname that we are currently considering
 // placing the pod on.  It returns these newly tightened requirements, or an error in the case of a set of requirements that
 // cannot be satisfied.
-func (t *Topology) AddRequirements(podRequirements, nodeRequirements scheduling.Requirements, p *corev1.Pod, compatabilityOptions ...option.Function[scheduling.CompatibilityOptions]) (scheduling.Requirements, error) {
+func (t *Topology) AddRequirements(p *corev1.Pod, taints []corev1.Taint, podRequirements, nodeRequirements scheduling.Requirements, compatabilityOptions ...option.Function[scheduling.CompatibilityOptions]) (scheduling.Requirements, error) {
 	requirements := scheduling.NewRequirements(nodeRequirements.Values()...)
-	for _, topology := range t.getMatchingTopologies(p, nodeRequirements, compatabilityOptions...) {
+	for _, topology := range t.getMatchingTopologies(p, taints, nodeRequirements, compatabilityOptions...) {
 		podDomains := scheduling.NewRequirement(topology.Key, corev1.NodeSelectorOpExists)
 		if podRequirements.Has(topology.Key) {
 			podDomains = podRequirements.Get(topology.Key)
@@ -189,26 +242,26 @@ func (t *Topology) AddRequirements(podRequirements, nodeRequirements scheduling.
 
 // Register is used to register a domain as available across topologies for the given topology key.
 func (t *Topology) Register(topologyKey string, domain string) {
-	for _, topology := range t.topologies {
-		if topology.Key == topologyKey {
-			topology.Register(domain)
+	for _, tg := range t.topologyGroups {
+		if tg.Key == topologyKey {
+			tg.Register(domain)
 		}
 	}
-	for _, topology := range t.inverseTopologies {
-		if topology.Key == topologyKey {
-			topology.Register(domain)
+	for _, tg := range t.inverseTopologyGroups {
+		if tg.Key == topologyKey {
+			tg.Register(domain)
 		}
 	}
 }
 
 // Unregister is used to unregister a domain as available across topologies for the given topology key.
 func (t *Topology) Unregister(topologyKey string, domain string) {
-	for _, topology := range t.topologies {
+	for _, topology := range t.topologyGroups {
 		if topology.Key == topologyKey {
 			topology.Unregister(domain)
 		}
 	}
-	for _, topology := range t.inverseTopologies {
+	for _, topology := range t.inverseTopologyGroups {
 		if topology.Key == topologyKey {
 			topology.Unregister(domain)
 		}
@@ -248,8 +301,8 @@ func (t *Topology) updateInverseAntiAffinity(ctx context.Context, pod *corev1.Po
 		tg := NewTopologyGroup(TopologyTypePodAntiAffinity, term.TopologyKey, pod, namespaces, term.LabelSelector, math.MaxInt32, nil, nil, nil, t.domainGroups[term.TopologyKey])
 
 		hash := tg.Hash()
-		if existing, ok := t.inverseTopologies[hash]; !ok {
-			t.inverseTopologies[hash] = tg
+		if existing, ok := t.inverseTopologyGroups[hash]; !ok {
+			t.inverseTopologyGroups[hash] = tg
 		} else {
 			tg = existing
 		}
@@ -280,25 +333,26 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 
 	// capture new domain values from existing nodes that may not have any pods selected by the topology group
 	// scheduled to them already
-	t.cluster.ForEachNode(func(n *state.StateNode) bool {
+	// Note: long term we should handle this when constructing the domain groups, but that would require domain groups
+	// to handle affinity in addition to taints / tolerations.
+	for _, n := range t.stateNodes {
 		// ignore state nodes which are tracking in-flight NodeClaims
 		if n.Node == nil {
-			return true
+			continue
 		}
 		// ignore the node if it doesn't match the topology group
-		if !tg.nodeFilter.Matches(n.Node) {
-			return true
+		if !tg.nodeFilter.Matches(n.Node.Spec.Taints, scheduling.NewLabelRequirements(n.Node.Labels)) {
+			continue
 		}
 		domain, exists := n.Labels()[tg.Key]
 		if !exists {
-			return true
+			continue
 		}
-		// ensure we at least have a count of zero for this potentially new topology domain
-		if _, countExists := tg.domains[domain]; !countExists {
+		if _, ok := tg.domains[domain]; !ok {
 			tg.domains[domain] = 0
+			tg.emptyDomains.Insert(domain)
 		}
-		return true
-	})
+	}
 
 	for i, p := range pods {
 		if IgnoredForTopology(&pods[i]) {
@@ -334,7 +388,7 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 		}
 		// nodes may or may not be considered for counting purposes for topology spread constraints depending on if they
 		// are selected by the pod's node selectors and required node affinities.  If these are unset, the node always counts.
-		if !tg.nodeFilter.Matches(node) {
+		if !tg.nodeFilter.Matches(node.Spec.Taints, scheduling.NewLabelRequirements(node.Labels)) {
 			continue
 		}
 		tg.Record(domain)
@@ -344,18 +398,28 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 
 func (t *Topology) newForTopologies(p *corev1.Pod) []*TopologyGroup {
 	var topologyGroups []*TopologyGroup
-	for _, cs := range p.Spec.TopologySpreadConstraints {
-		for _, key := range cs.MatchLabelKeys {
-			if value, ok := p.ObjectMeta.Labels[key]; ok {
-				cs.LabelSelector.MatchExpressions = append(cs.LabelSelector.MatchExpressions, metav1.LabelSelectorRequirement{
+	for _, tsc := range p.Spec.TopologySpreadConstraints {
+		for _, key := range tsc.MatchLabelKeys {
+			if value, ok := p.Labels[key]; ok {
+				tsc.LabelSelector.MatchExpressions = append(tsc.LabelSelector.MatchExpressions, metav1.LabelSelectorRequirement{
 					Key:      key,
 					Operator: metav1.LabelSelectorOpIn,
 					Values:   []string{value},
 				})
 			}
 		}
-		topologyGroups = append(topologyGroups, NewTopologyGroup(TopologyTypeSpread, cs.TopologyKey, p, sets.New(p.Namespace),
-			cs.LabelSelector, cs.MaxSkew, cs.MinDomains, cs.NodeTaintsPolicy, cs.NodeAffinityPolicy, t.domainGroups[cs.TopologyKey]))
+		topologyGroups = append(topologyGroups, NewTopologyGroup(
+			TopologyTypeSpread,
+			tsc.TopologyKey,
+			p,
+			sets.New(p.Namespace),
+			tsc.LabelSelector,
+			tsc.MaxSkew,
+			tsc.MinDomains,
+			tsc.NodeTaintsPolicy,
+			tsc.NodeAffinityPolicy,
+			t.domainGroups[tsc.TopologyKey],
+		))
 	}
 	return topologyGroups
 }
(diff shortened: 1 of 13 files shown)


pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1311,20 +1310,26 @@ var _ = Describe("Topology", func() {
...
+				return test.UnschedulablePods(test.PodOptions{
+					ObjectMeta: metav1.ObjectMeta{
+						Labels: labels,
 					},
 					TopologySpreadConstraints: topology,
-				}, 5)...,
-			)
-			// and should schedule all of the pods on the same node
-			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(5))
+					Tolerations: []v1.Toleration{{
```
tzneal, line 1370: This may work, but I think the test will be clearer with:
- 1 node pool with a value for a particular label and taints
- 1 node pool with a different value for that label and not taints

and then schedule multiple pods with no tolerations and a required topology spread over the label, one should schedule against the non-tainted node pool and the others remain pending.

A second test has identical setup but taints policy is ignore and all pods schedule against the non-tainted node pool since they don't discover the tainted node pool.

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -0,0 +1,67 @@
...
+// GetDomains returns a slice of all the domains in the topology tracked by this TopologyDomainGroup.
+func (t TopologyDomainGroup) GetDomains() []string {
+	return lo.Keys(t)
+}
+
+// GetToleratedDomains returns a slice of domains that the pod can potentially schedule to, given its tolerations.
+func (t TopologyDomainGroup) GetToleratedDomains(pod *v1.Pod) []string {
+	return lo.Filter(lo.Keys(t), func(domain string, _ int) bool {
+		for _, taints := range t[domain] {
+			if err := scheduling.Taints(taints.UnsortedList()).ToleratesPod(pod); err == nil {
```
tzneal, line 61: If we can void calling keys/unsortedlisted & friends in the scheduling code, it's better.  They construct new lists and we spend extra time on GC that we don't need to.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -86,6 +82,20 @@ func NewTopologyGroup(topologyType TopologyType, topologyKey string, pod *v1.Pod
 		nodeSelector = MakeTopologyNodeFilter(pod, nodeTaintsPolicy, nodeAffinityPolicy)
 	}
 
+	domainCounts := lo.FromEntries(lo.Map(
```
tzneal, line 85: Same here,

1. lo.FromEntries creates a new map
2. lo.Map creates a new slice

pkg/controllers/provisioning/scheduling/topology_test.go
jmdeal: That makes sense and IIUC I think that's more or less what you had originally for this test (+ the first set of pods). I modified the original test to catch a case I had missed when I originally implemented the filter for which domains a pod should consider based on taints. Basically I mistakenly thought I could just do a set intersection of the taints for all of the nodepools, which would result in a pod with no tolerations considering `foo` as a domain in this case (instead of needing to tolerate `nodepool-1` or `nodepool-2`. To catch this I needed to have multiple NodePools with distinct taints that shared at least 1 domain. I'd be happy to add in an additional test, I agree that this one isn't super clear.

tzneal: Sure that works, let's add the other test and keep this as well.

jmdeal:
Holding to include `matchLabelKeys` for pod affinity with k8s v1.29.

jmdeal:
Dropped the 1.29 version bumps from this PR in favor of adding in #886, build will fail until that is merged in and this PR is rebased.

jonathan-innis:
PR is blocked on envtest support for 1.29

tzneal:
Let's run a benchmark of the scheduler before/after this change and compare just to check performance.

Bryce-Soghigian:
Is there a way to keep stalebot away for PRs?

jmdeal:
Benchmarking results:
```
goos: darwin
goarch: arm64
                  │ old_results.txt │           new_results.txt            │
                  │     sec/op      │    sec/op      vs base               │
Scheduling1-10         99.67µ ± 42%   103.39µ ± 30%       ~ (p=0.989 n=20)
Scheduling50-10        305.6µ ±  6%    305.5µ ±  7%       ~ (p=0.265 n=20)
Scheduling100-10       488.9µ ±  6%    451.0µ ±  5%  -7.76% (p=0.001 n=20)
Scheduling500-10       1.929m ±  1%    1.859m ±  2%  -3.66% (p=0.002 n=20)
Scheduling1000-10      3.955m ±  3%    3.999m ±  7%       ~ (p=0.478 n=20)
Scheduling2000-10      7.518m ±  7%    8.072m ±  4%  +7.38% (p=0.013 n=20)
Scheduling5000-10      19.45m ±  9%    17.96m ±  4%  -7.69% (p=0.038 n=20)
geomean                1.494m          1.477m        -1.13%

                  │ old_results.txt │           new_results.txt           │
                  │      nodes      │   nodes     vs base                 │
Scheduling1-10           1.000 ± 0%   1.000 ± 0%       ~ (p=1.000 n=20) ¹
Scheduling50-10          5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=20) ¹
Scheduling100-10         5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=20) ¹
Scheduling500-10         5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=20) ¹
Scheduling1000-10        5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=20) ¹
Scheduling2000-10        5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=20) ¹
Scheduling5000-10        5.000 ± 0%   5.000 ± 0%       ~ (p=1.000 n=20) ¹
geomean                  3.973        3.973       +0.00%
¹ all samples are equal

                  │ old_results.txt │           new_results.txt            │
                  │      pods       │    pods      vs base                 │
Scheduling1-10          1.000 ±  0%   1.000 ±  0%       ~ (p=1.000 n=20) ¹
Scheduling50-10         29.00 ± 17%   29.00 ± 24%       ~ (p=0.814 n=20)
Scheduling100-10        25.50 ± 25%   26.00 ± 23%       ~ (p=0.732 n=20)
Scheduling500-10        28.00 ± 50%   25.00 ± 32%       ~ (p=0.587 n=20)
Scheduling1000-10       35.50 ± 21%   29.00 ± 21%       ~ (p=0.279 n=20)
Scheduling2000-10       29.50 ± 39%   28.00 ± 29%       ~ (p=0.815 n=20)
Scheduling5000-10       34.00 ± 41%   31.00 ± 35%       ~ (p=0.344 n=20)
geomean                 18.48         17.36        -6.10%
¹ all samples are equal

                  │ old_results.txt │           new_results.txt           │
                  │    pods/sec     │   pods/sec    vs base               │
Scheduling1-10        10.044k ± 30%   9.672k ± 24%       ~ (p=0.984 n=20)
Scheduling50-10        163.6k ±  6%   163.7k ±  8%       ~ (p=0.265 n=20)
Scheduling100-10       204.5k ±  6%   221.8k ±  5%  +8.42% (p=0.001 n=20)
Scheduling500-10       259.2k ±  1%   269.0k ±  2%  +3.80% (p=0.002 n=20)
Scheduling1000-10      252.9k ±  3%   250.0k ±  8%       ~ (p=0.478 n=20)
Scheduling2000-10      266.1k ±  7%   247.8k ±  4%  -6.91% (p=0.013 n=20)
Scheduling5000-10      257.0k ± 10%   278.4k ±  3%  +8.32% (p=0.038 n=20)
geomean                147.3k         149.0k        +1.12%
```

abrazouski:
Sorry @jmdeal for bothering, but any plans for moving forward with this?
We're really rely on this feature

hamishforbes:
Hi @jmdeal is there anything blocking this or any way I can help to move this on (testing a custom build or something)?

jonathan-innis:
> Ran this comparison a few times, this was the set of results with the largest difference. Marginally faster overall surprisingly, with a slight decrease in first round scheduled pods (though I noticed this stat varied wildly between runs).

And I think I figured out why! Turns out that we really haven't been doing our scheduling benchmarking correctly in a while! See #1930

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -69,20 +69,42 @@ type TopologyGroup struct {
 	emptyDomains sets.Set[string]       // domains for which we know that no pod exists
 }
 
-func NewTopologyGroup(topologyType TopologyType, topologyKey string, pod *v1.Pod, namespaces sets.Set[string], labelSelector *metav1.LabelSelector, maxSkew int32, minDomains *int32, domains sets.Set[string]) *TopologyGroup {
-	domainCounts := map[string]int32{}
-	for domain := range domains {
-		domainCounts[domain] = 0
-	}
+func NewTopologyGroup(topologyType TopologyType, topologyKey string, pod *v1.Pod, namespaces sets.Set[string], labelSelector *metav1.LabelSelector, maxSkew int32, minDomains *int32, taintPolicy *v1.NodeInclusionPolicy, affinityPolicy *v1.NodeInclusionPolicy, domainGroup TopologyDomainGroup) *TopologyGroup {
```
jonathan-innis, line 72:
```suggestion
func NewTopologyGroup(topologyType TopologyType, topologyKey string, pod *v1.Pod, namespaces sets.Set[string], labelSelector *metav1.LabelSelector, maxSkew int32, minDomains *int32, taintPolicy, affinityPolicy *v1.NodeInclusionPolicy, domainGroup TopologyDomainGroup) *TopologyGroup {
```

the tiniest nit known to man

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -69,20 +69,42 @@ type TopologyGroup struct {
...
+		domainGroup.ForEachToleratedDomain(pod, func(domain string) {
+			domains[domain] = 0
+		})
+	} else {
+		domainGroup.ForEachDomain(func(domain string) {
+			domains[domain] = 0
+		})
+	}
+
+	emptyDomains := sets.New[string]()
```
jonathan-innis, line 103: So, does emptyDomains need to match the domains that are marked above? If a domain isn't tolerated, is it going to be problematic if we allow it to be an empty domain -- if not practically, what about semantically?

pkg/utils/pod/scheduling.go
```diff
@@ -180,9 +180,14 @@ func HasDoNotDisrupt(pod *corev1.Pod) bool {
 	return pod.Annotations[v1.DoNotDisruptAnnotationKey] == "true"
 }
 
-// ToleratesDisruptedNoScheduleTaint returns true if the pod tolerates karpenter.sh/disrupted:NoSchedule taint
+// ToleratesUnschedulableTaint returns true if the pod tolerates node.kubernetes.io/unschedulable taint
+func ToleratesUnschedulableTaint(pod *corev1.Pod) bool {
```
jonathan-innis, line 184: This might have been a rebase error -- I don't see anywhere in this PR that you're using this

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -91,8 +113,8 @@ func NewTopologyGroup(topologyType TopologyType, topologyKey string, pod *v1.Pod
 		rawSelector:  labelSelector,
 		nodeFilter:   nodeSelector,
```
jonathan-innis, line 114:
```suggestion
		nodeFilter:   nodeFilter,
```

nit: Consider calling this `nodeFilter` now since it's no longer just a nodeSelector

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+// TopologyDomainGroup tracks the domains for a single topology. Additionally, it tracks the taints associated with
+// each of these domains. This enables us to determine which domains should be considered by a pod if its
+// NodeTaintPolicy is honor.
+type TopologyDomainGroup map[string][][]v1.Taint
```
jonathan-innis, line 28: Does this always work? I'm concocting a scenario where there is a single node in a domain (let's say zone) that has a pod scheduled to it and a taint and another node in a domain (same zone) that has a pod scheduled to it that matches the nodeSelector. This code would mean that we would count both pods when determining domain selection -- since there is one set of taints that we tolerate in the domain. Do I have that right?

jonathan-innis:
First pass -- I didn't look at everything and have some starting questions

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
jmdeal: If I understand correctly, I believe I have a [test](https://github.com/kubernetes-sigs/karpenter/blob/01de6ac7643b419957839b1a66e9aa33ab6856c6/pkg/controllers/provisioning/scheduling/topology_test.go#L1459-L1525) for this exact scenario. 

The role of the `TopologyDomainGroup` is solely to determine the eligible domains when constructing a `TopologyGroup`. A domain is eligible if the pod tolerates any of the groups of taints stored in the domain group. This isn't actually responsible for determining which nodes are counted as part of a domain, that's still going to be handled by the `TopologyNodeFilter`, where we still verify that the node is tolerated by the pod for which the `TopologyGroup` was constructed.

pkg/controllers/provisioning/scheduling/topologygroup.go
jmdeal: Discussed offline, while this makes some sense since `emptyDomains` is only used for affinity based topology groups, which don't support an honor taints policy, mapping `emptyDomains` directly to the keys of `domains` does make the intention clearer. It shouldn't make a functional difference, so we'll take in the change.

pkg/controllers/provisioning/scheduling/topologygroup.go
jmdeal: I've reformatted to put each argument on it's own line, since this is beginning to wrap off of my screen. Given that, I think leaving both types in is more readable, but let me know what you think. Personally, I prefer always including the types as I find it a little more readable, but I don't feel to strongly.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -28,43 +28,66 @@ import (
...
+			TaintPolicy:    taintPolicy,
+			AffinityPolicy: affinityPolicy,
+			Tolerations:    p.Spec.Tolerations,
+		}
 	}
 
 	// otherwise, we need to match the combination of label selector and any term of the required node affinities since
 	// those terms are OR'd together
-	var filter TopologyNodeFilter
+	filter := TopologyNodeFilter{
```
jonathan-innis, line 52: We don't attach tolerations in this codepath? Seems wrong that we wouldn't attach tolerations here -- also, since there isn't a test failing for this scenario seems like we need to add one

pkg/utils/pod/scheduling.go
```diff
@@ -180,9 +180,9 @@ func HasDoNotDisrupt(pod *corev1.Pod) bool {
 	return pod.Annotations[v1.DoNotDisruptAnnotationKey] == "true"
 }
 
-// ToleratesDisruptedNoScheduleTaint returns true if the pod tolerates karpenter.sh/disrupted:NoSchedule taint
+// ToleratesDisruptionNoScheduleTaint returns true if the pod tolerates karpenter.sh/disruption:NoSchedule=Disrupting taint
```
jonathan-innis, line 183: This doesn't look right -- I think we are still using `karpenter.sh/disrupted:NoSchedule`

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -0,0 +1,67 @@
...
+
+package scheduling
+
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+// TopologyDomainGroup tracks the domains for a single topology. Additionally, it tracks the taints associated with
```
jonathan-innis, line 25: It just tracks the eligible domains, right?

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -0,0 +1,67 @@
...
+
+func NewTopologyDomainGroup() TopologyDomainGroup {
+	return map[string][][]v1.Taint{}
+}
+
+// Insert either adds a new domain to the TopologyDomainGroup or updates an existing domain.
+func (t TopologyDomainGroup) Insert(domain string, taints ...v1.Taint) {
+	// Note: This could potentially be improved by removing any set of which the new set of taints is a proper subset.
+	// Currently this is only handled when the incoming set is the empty set.
+	if _, ok := t[domain]; !ok || len(taints) == 0 {
```
jonathan-innis, line 38: It still took me a second to parse this -- it might be worth pushing that comment down below up here or maybe add a second one -- just because I didn't get it until I saw the comment down below

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -0,0 +1,67 @@
...
+// ForEachDomain calls f on each domain tracked by the topology group
+func (t TopologyDomainGroup) ForEachDomain(f func(domain string)) {
+	for domain := range t {
+		f(domain)
+	}
+}
+
+// ForEachToleratedDomain calls f on each domain tracked by the TopologyDomainGroup which are also tolerated by the provided pod.
+func (t TopologyDomainGroup) ForEachToleratedDomain(pod *v1.Pod, f func(domain string)) {
+	for domain, taintGroups := range t {
```
jonathan-innis, line 59: Consider marking down that we could potentially improve this code by just storing the domains that map to certain cached tolerations -- small improvement but maybe worth doing eventually

pkg/controllers/provisioning/scheduling/topologygroup.go
jonathan-innis: Maybe just a personal preference thing but it makes more sense in my head to just inject into empty domains right next to where you are setting the value to 0 (which means you will have two lines vs. one) but I think it's worth it just to reason about the logic that's mapped together

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -278,6 +278,28 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 		pods = append(pods, podList.Items...)
 	}
 
+	// capture new domain values from existing nodes that may not have any pods selected by the topology group
+	// scheduled to them already
+	t.cluster.ForEachNode(func(n *state.StateNode) bool {
```
jonathan-innis, line 283: Noting down that we have plans to move this over into domainGroups since domainGroups should really capture _all_ of the domains that are eligible across the cluster (including NodePools) rather than just the NodePools (which is what they are doing right now)

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -278,6 +278,28 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 		pods = append(pods, podList.Items...)
 	}
 
+	// capture new domain values from existing nodes that may not have any pods selected by the topology group
```
jonathan-innis, line 281: After some discussion with @jmdeal, we should consider passing state nodes directly into this function rather than using the t.cluster.ForEach and holding a read lock throughout the time

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -278,6 +278,28 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
...
+		// ignore the node if it doesn't match the topology group
+		if !tg.nodeFilter.Matches(n.Node) {
+			return true
+		}
+		domain, exists := n.Labels()[tg.Key]
+		if !exists {
+			return true
+		}
+		// ensure we at least have a count of zero for this potentially new topology domain
+		if _, countExists := tg.domains[domain]; !countExists {
```
jonathan-innis, line 297: nit: Since this is more standard
```suggestion
		if _, ok := tg.domains[domain]; !ok {
```

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -323,7 +345,17 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 func (t *Topology) newForTopologies(p *corev1.Pod) []*TopologyGroup {
 	var topologyGroups []*TopologyGroup
 	for _, cs := range p.Spec.TopologySpreadConstraints {
-		topologyGroups = append(topologyGroups, NewTopologyGroup(TopologyTypeSpread, cs.TopologyKey, p, sets.New(p.Namespace), cs.LabelSelector, cs.MaxSkew, cs.MinDomains, t.domains[cs.TopologyKey]))
+		for _, key := range cs.MatchLabelKeys {
+			if value, ok := p.ObjectMeta.Labels[key]; ok {
```
jonathan-innis, line 349:
```suggestion
			if value, ok := p.Labels[key]; ok {
```

super super minor nit

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -323,7 +345,17 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
 func (t *Topology) newForTopologies(p *corev1.Pod) []*TopologyGroup {
 	var topologyGroups []*TopologyGroup
 	for _, cs := range p.Spec.TopologySpreadConstraints {
```
jonathan-innis, line 347:
```suggestion
	for _, tsc := range p.Spec.TopologySpreadConstraints {
```

this is also minor but why not?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -278,6 +278,28 @@ func (t *Topology) countDomains(ctx context.Context, tg *TopologyGroup) error {
...
+		if !tg.nodeFilter.Matches(n.Node) {
+			return true
+		}
+		domain, exists := n.Labels()[tg.Key]
+		if !exists {
+			return true
+		}
+		// ensure we at least have a count of zero for this potentially new topology domain
+		if _, countExists := tg.domains[domain]; !countExists {
+			tg.domains[domain] = 0
```
jonathan-innis, line 298: We missed empty domains here

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+			}, count)
+			pods = append(pods, test.UnschedulablePods(test.PodOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: lo.Assign(labels, map[string]string{matchedLabel: "value-b"}),
+				},
+				TopologySpreadConstraints: topology,
+			}, count)...)
+			ExpectApplied(ctx, env.Client, nodePool)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(2, 2))
```
jonathan-innis, line 1160: I find this slightly confusing -- is this implying that one pod of each scheduled on the first node and one pod of each scheduled on the second? If so, probably worth adding a comment to clarify that -- might also be nice to check node length

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+				}})
+			ExpectApplied(ctx, env.Client, nodePool, node1, node2)
+
+			ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node1))
+			ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node2))
+
+			// there are now three domains for spreadLabel that Karpenter should know about, foo/bar from the two existing
+			// nodes and baz from the node pool
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov)
+			ignore := corev1.NodeInclusionPolicyIgnore
```
jonathan-innis, line 1238: nit: Do you need to set this to a variable -- can't you just take the address of the string from upstream?

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+				}})
+			ExpectApplied(ctx, env.Client, nodePool, node1, node2)
+
+			ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node1))
+			ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node2))
+
+			// since two of the nodes are tainted, Karpenter should not consider them for topology domain discovery
+			// purposes
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov)
+			honor := corev1.NodeInclusionPolicyHonor
```
jonathan-innis, line 1309: Same comment here

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+										Values:   []string{"bar"},
+									},
+								},
+							},
+						},
+					},
+				},
+			})
+
+			honor := corev1.NodeInclusionPolicyIgnore
```
jonathan-innis, line 1374: This looks like it's mislabeled and also isn't necessary as well as a variable assignment

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				TopologySpreadConstraints: topology,
+			}, 2)
+
+			ExpectApplied(ctx, env.Client, nodePool, taintedNodePool)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+
+			// should fail to schedule both pods, one pod is scheduled to domain "foo" but the other can't be scheduled to domain "bar"
```
jonathan-innis, line 1393: I think this comment may not be accurate -- it only fails to schedule one of pods

jonathan-innis: Also, I didn't realize that you got these cross-NodePool constraints today -- that's interesting -- I guess that makes sense if your pod supports both of them

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+					TopologySpreadConstraints: topology,
+				}, 5)...,
+			)
+			// and should schedule all of the pods on the same node
+			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(5))
+		})
+		It("should balance pods across a label when discovered from the provisioner (NodeTaintsPolicy=ignore)", func() {
+			const spreadLabel = "fake-label"
+			const taintKey = "taint-key"
+			nodePool.Spec.Template.Spec.Requirements = append(nodePool.Spec.Template.Spec.Requirements, v1.NodeSelectorRequirementWithMinValues{
```
jonathan-innis, line 1336: For the tests that are validating schedulability of a pod against a nodepool or node, it might be worth it to add taints and tolerations for the nodepool/node that we are able to schedule to as well so that we ensure that we are exercising the tolerates checks that we have in the topology code

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+				TopologySpreadConstraints: topology,
+			}, 2)
+
+			ExpectApplied(ctx, env.Client, nodePool, taintedNodePool)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+
+			// should fail to schedule both pods, one pod is scheduled to domain "foo" but the other can't be scheduled to domain "bar"
+			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(1))
+		})
+		It("should balance pods across a label when discovered from the provisioner (NodeTaintsPolicy=honor)", func() {
```
jonathan-innis, line 1396:
```suggestion
		It("should balance pods across a label when discovered from the NodePool (NodeTaintsPolicy=honor)", func() {
```

consider replacing the use of the word provisioner throughout this file with NodePool :P we got some legacy wording that got kept around due to how long this has been open

pkg/controllers/provisioning/scheduling/topology_test.go
jonathan-innis: I could also see an argument that this is tested below but it probably doesn't hurt -- I could go either way

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+						Effect: corev1.TaintEffectNoSchedule,
+						Value:  np.Spec.Template.Spec.Taints[0].Value,
+					}},
+				}, 2)
+			}))
+
+			ExpectApplied(ctx, env.Client, nodePools[0], nodePools[1])
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+
+			// Expect 3 total nodes provisioned, 2 pods schedule to foo, 1 to bar, and 1 to baz
```
jonathan-innis, line 1522: nit: consider adding a bit more to this comment (1 pod that tolerated the nodepool-1 taint on fot, 1 that tolerated the nodepool-1 taint on bar, etc.

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+					Labels: labels,
+				},
+				TopologySpreadConstraints: topology,
+			}, 2)
+
+			ExpectApplied(ctx, env.Client, nodePool, taintedNodePool)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+
+			// should schedule all pods to domain "foo", ignoring bar since pods don't tolerate
+			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(2))
```
jonathan-innis, line 1457: What about a test for a set of pods that have two different tolerations and two different NodePools with mutually exclusive taints?

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,544 @@ var _ = Describe("Topology", func() {
...
+				Allocatable: corev1.ResourceList{
+					corev1.ResourceCPU: resource.MustParse("100m"),
+				}})
+			ExpectApplied(ctx, env.Client, nodePool, node1, node2)
+
+			ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node1))
+			ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node2))
+
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov)
+			ignore := corev1.NodeInclusionPolicyHonor
```
jonathan-innis, line 1638: nit: This naming is wrong from the copy

jonathan-innis:
This is awesome work! This is definitely getting close! I think mostly a few small things -- we should write-down things that we want to refactor here since there were some ideas thrown out about how we could move existing nodes into `topologyGroups` and avoid iterating through stateNodes in countDomains -- also some things around naming of functions -- also also capturing nodeAffinities in eligible domains for topologyDomainGroups

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
jmdeal: I'm not sure what the distinction here is - the domain group tracks all possible domains for a given topology, along with the taints which are one aspect of determining if a domain is eligible for a given topology group.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
jmdeal: Good catch, looks like we need a test that involves pods with node selectors and an honor taint TSC.

pkg/controllers/provisioning/scheduling/topology_test.go
jmdeal: That should be handled by this test, right? We're you thinking of something else? https://github.com/kubernetes-sigs/karpenter/blob/3feab07dfd89e5b27bcea06f54f22610ff8529ad/pkg/controllers/provisioning/scheduling/topology_test.go#L1455

pkg/controllers/provisioning/scheduling/topology_test.go
jmdeal: I've updated the test below this one to exercise this, and it did actually catch a miss in the check for taints / tolerations in `topology.Record`.

jonathan-innis: Nice! Good to hear!

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1125,6 +1126,550 @@ var _ = Describe("Topology", func() {
...
+
+			topology := []corev1.TopologySpreadConstraint{{
+				TopologyKey:       spreadLabel,
+				WhenUnsatisfiable: corev1.DoNotSchedule,
+				LabelSelector:     &metav1.LabelSelector{MatchLabels: labels},
+				MaxSkew:           1,
+				NodeTaintsPolicy:  lo.ToPtr(corev1.NodeInclusionPolicyHonor),
+			}}
+
+			pods := lo.Flatten(lo.Map(nodePools, func(np *v1.NodePool, i int) []*corev1.Pod {
```
jonathan-innis, line 1502: nit: Comment -- two pods that tolerate the first NodePool, four pods that tolerate the second

jonathan-innis (commented):
Nice work!

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -83,6 +96,46 @@ func NewTopology(ctx context.Context, kubeClient client.Client, cluster *state.C
 	return t, nil
 }
 
+func buildDomainGroups(nodePools []*v1.NodePool, instanceTypes map[string][]*cloudprovider.InstanceType) map[string]TopologyDomainGroup {
+	nodePoolIndex := lo.SliceToMap(nodePools, func(np *v1.NodePool) (string, *v1.NodePool) {
```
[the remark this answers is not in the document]
jmdeal: O(n) vs O(n^2). Honestly can go either way on this one, in all likelihood it makes no real runtime difference since this is just a one off loop at the beginning of each scheduling loop.

pkg/controllers/provisioning/scheduling/topologynodefilter.go
```diff
@@ -28,43 +28,67 @@ import (
...
+		filter.Requirements = append(filter.Requirements, requirements)
 	}
 
 	return filter
 }
 
 // Matches returns true if the TopologyNodeFilter doesn't prohibit node from the participating in the topology
-func (t TopologyNodeFilter) Matches(node *v1.Node) bool {
-	return t.MatchesRequirements(scheduling.NewLabelRequirements(node.Labels))
+func (t TopologyNodeFilter) Matches(taints []corev1.Taint, requirements scheduling.Requirements, compatibilityOptions ...option.Function[scheduling.CompatibilityOptions]) bool {
```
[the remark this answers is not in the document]
jmdeal: Right now we only allow missing WellKnownLabels when evaluating if an inflight NodeClaim should count against a domain. I don't think we should drop that here because that would result in us **not** counting a nodeclaim for a topology group when we should have. I think this is the right side for us to error on - being to conservative and overcounting NodeClaims rather than undercounting them.

pkg/controllers/provisioning/scheduling/topologygroup.go
jonathan-innis: Nice! This looks great!

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
/hold 

Wait for confirmation from @jmdeal that this one is good to merge

jmdeal:
Should be good to go 👍

codeeong:
When will a new release with this change be rolled out? Thanks!

PR #860, opened 2023-12-13
chore: Fix consistent ordering of NodePools and Provisioners

jonathan-innis:
Fixes #N/A 

**Description**

This changes the ordering of NodePools and their sorting so that NodePools of equal weight are always moved in front of Provisioners. This particularly helps with the migration story, where we want the same NodePool that has the same weight as Provisioner to always come first

`make presubmit`

### DIFF
--- a/pkg/apis/v1beta1/nodepool.go
+++ b/pkg/apis/v1beta1/nodepool.go
@@ -160,10 +160,28 @@ type NodePoolList struct {
 	Items           []NodePool `json:"items"`
 }
 
-// OrderByWeight orders the provisioners in the NodePoolList
-// by their priority weight in-place
+// OrderByWeight orders the NodePools in the NodePoolList by their priority weight in-place.
+// NodePools currently have an isProvisioner flag which defines whether the NodePool is actually a NodePool object
+// at the apiserver or whether it's just a NodePool representation in memory and is actually a Provisioner at
+// the apiserver. This conversion in memory follows an MVC model and helps the scheduler reason about options.
+// This priority evaluates the following things in precedence order:
+//  1. NodePools that have a larger weight are ordered first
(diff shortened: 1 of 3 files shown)


pkg/apis/v1beta1/nodepool.go
```diff
@@ -160,10 +160,27 @@ type NodePoolList struct {
 	Items           []NodePool `json:"items"`
 }
 
-// OrderByWeight orders the provisioners in the NodePoolList
+// OrderByWeight orders the NodePools in the NodePoolList
 // by their priority weight in-place
+// This priority evaluates the following things in precedence order:
+//  1. NodePools that have a larger weight are ordered first
+//  2. If two NodePools have the same weight, a NodePool that is actually a true NodePool on the apiserver
+//     and isn't a converted NodePool from a Provisioner will come next
```
njtran, line 168: Just a clarifying nit. When I first read this, I thought it could come off as "If i've used the karpenter-convert tool, this means that nodepool would not be prioritized"
```suggestion
//  2. If two NodePools have the same weight, a NodePool that is actually a true NodePool on the apiserver
//     and isn't a converted NodePool from a Provisioner will come next in memory
```

njtran (approved):
Really nice coverage on the tests. One comment.

pkg/apis/v1beta1/nodepool.go
jonathan-innis: Updated wording to be more explicit about the `isProvisioner` flag. Let me know if it's still not clear.

njtran (approved):
/lgtm /approve

njtran:
Had to merge instead of Prowbot since it's looking for an owners.md file in this branch that doesn't exist.

PR #862, opened 2023-12-13
chore: Drop events for consolidation disabled when we have a different policy

jonathan-innis:
Fixes #N/A 

**Description**

This drops the "Consolidation disabled" event when we have `consolidationPolicy: WhenEmpty`. This error is confusing for users since (from their perspective), consolidation is in fact enabled with the `consolidationPolicy`


engedaam (commented):
/lgtm

PR #864, opened 2023-12-14
fix: remove event filter for nodepool hash controller

njtran:
Fixes #N/A 

**Description**
If a user removes the hash from the nodepool, the hash won't be added again until another event occurs, this could prevent drift from happening

There are four other controllers that filter out update events that don't modify the spec, which are all the correct behavior. One difference is we'll now enqueue for the nodeclaim/disruption controller so that if a user removes the status condition on a nodeclaim, we'll add it back in if necessary.

tested manually

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/controller.go
+++ b/pkg/controllers/nodeclaim/disruption/controller.go
@@ -25,6 +25,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/clock"
+	"knative.dev/pkg/apis"
 	controllerruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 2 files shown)


pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -134,6 +135,14 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 								return true
 							}
 						}
+						for _, cond := range []apis.ConditionType{v1beta1.Expired, v1beta1.Empty, v1beta1.Drifted} {
```
jonathan-innis, line 138: Do we just outright remove the event filtering logic here and just requeue across all updates? I'm trying to think of a case where that wouldn't make sense as long as we are ensuring that our updates don't cause increased reconciles

jonathan-innis (commented):
/lgtm
/approve

PR #866, opened 2023-12-15
fix: don't panic on shutdown for nil nodepools for logging budgets

njtran:
Fixes #N/A 

**Description**
On shutdown, we were getting panics since we weren't returning after a failed nodePool.List call.

make presubmit


jonathan-innis (approved):
LGTM 🚀

pkg/controllers/disruption/controller.go
```diff
@@ -250,7 +250,8 @@ func (c *Controller) logAbnormalRuns(ctx context.Context) {
 func (c *Controller) logInvalidBudgets(ctx context.Context) {
```
jonathan-innis, line 250: Discussed offline. I think the method signature should just be `ValdiateBudgets() error` so that we are less likely to miss error returns like this

PR #869, opened 2023-12-15
feat: Initialize the VEX feed

coder12git:
Fixes #N/A 
This PR is created to ensure the project has a VEX feed under the `Cloud Native Security Slam`

**Description**
This PR initialize the VEX feed

**How was this change tested?**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #870, opened 2023-12-15
ci: Added GitHub action to generate OpenVEX data with each release

sreeram-venkitesh:
Fixes #N/A 

**Description**

This PR adds a github action to generate openvex data on each release. This is a follow up to https://github.com/kubernetes-sigs/karpenter/pull/869. These two PRs are raised as part of the Cloud Native Security Slam. Please refer this [doc](https://docs.google.com/document/d/1TLahqW6OpWQXbaVJ-SR8z-o0yLSoGKYTHxPU0PyyjXo/) for more information.

I tested this by creating a test tag in my fork of the repo, similar to what @YashPimple has done in https://github.com/kubernetes-sigs/karpenter/pull/871. The job is getting created, but it is failing because the template directory created in https://github.com/kubernetes-sigs/karpenter/pull/869 is not present. I will rebase the code once #869 is merged and test the job once again.


.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
+        name: Run vexctl
+        with:
+            product: pkg:generic/hello@${{ env.RELEASE_VERSION }}
```
jonathan-innis, line 17: From what I can understand, this just generates the file, but doesn't actually attach it to the release. I think you need a separate upload step to attach the generated vex file to the release. See an example of that here: https://github.com/kubernetes-sigs/kueue/pull/1475

jonathan-innis (commented):
/ok-to-test

.github/workflows/openvex.yaml
sreeram-venkitesh: @jonathan-innis: I've referred the Kueue PR and made the workflow manual trigger only. I've added the upload step as well. I was not able to trigger the workflow manually in my fork to test this though. I couldn't any logs of this being done in the Kueue repository as well. Please let me know if the code changes are fine.

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  vexctl:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
```
mehabhalodiya, line 20: Hi @sreeram-venkitesh 
I am helping you to figure out what is different from this [kueue/pull/1475/files](https://github.com/kubernetes-sigs/kueue/pull/1475/files) PR, as a few of the CI checks are failing here.

Remove this `Set env` field.

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Set tag name
+        shell: bash
+        run: |
+          TAG=${{ github.event.inputs.tag }}
+          echo "TAG=$TAG" >> "$GITHUB_ENV"
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
+        name: Run vexctl
+        with:
+          product: pkg:generic/karpenter@${{ env.RELEASE_VERSION }}
```
mehabhalodiya, line 29: Please compare the `product` field with this [kueue/pull/1475/files#diff-R26](https://github.com/kubernetes-sigs/kueue/pull/1475/files#diff-c2573174fef0722ce49b7b168ac16b86d880224801aa6a45af9489ed4031648cR26).

I believe it needs a bit of change; maybe remove `RELEASE_VERSION` and instead use `TAG`.

mehabhalodiya (changes requested):

.github/workflows/openvex.yaml
jonathan-innis, line 20: Just from looking over this, I think that we are just using the tag here; so, we could probably use `github.ref_name` rather than having to parse the tag from the ref directly

.github/workflows/openvex.yaml
jonathan-innis, line 29:
```suggestion
          product: pkg:generic/karpenter@${{ github.ref_name }}
```

I'm assuming that we migth be able to just use the `ref_name` directly here rather than having to parse it out

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,35 @@
...
+          file: /tmp/karpenter.vex.json
+      - name: Upload openvex data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload $TAG /tmp/karpenter.vex.json
```
jonathan-innis, line 35:
```suggestion
          gh release upload ${{ github.ref_name }} /tmp/karpenter.vex.json
```

Does using the ref_name just work here?

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,35 @@
...
+          TAG=${{ github.event.inputs.tag }}
+          echo "TAG=$TAG" >> "$GITHUB_ENV"
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
+        name: Run vexctl
+        with:
+          product: pkg:generic/karpenter@${{ env.RELEASE_VERSION }}
+          file: /tmp/karpenter.vex.json
+      - name: Upload openvex data
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jonathan-innis, line 33: sanity check: Do we need to set this environment variable here or is the token just injected into the context when you are making calls in the bash script?

.github/workflows/openvex.yaml
jonathan-innis: Also, as an additional security measure, if we are going to use `ref_name` here, it's best practice to just store that value in an environment variable and reference it in the scripts to avoid script injection. Maybe we can set `TAG: ${{ github.ref_name }}` at the top of the script and then just reference the environment variable as we need it throughout

.github/workflows/openvex.yaml
sreeram-venkitesh: I made the changes mentioned in the review and tried triggering the [workflow in my fork](https://github.com/sreeram-venkitesh/karpenter/actions/runs/7323345505). It was able to [generate and upload the json file](https://github.com/sreeram-venkitesh/karpenter/releases/tag/v0.34.9) without me having to set any environment variable.

sreeram-venkitesh:
@jonathan-innis, @mehabhalodiya: I've made the changes mentioned in the reviews. I was also able to get the workflow to succeed in my fork of the repository and generate the openvex json file. Please take a look at the workflow run [here](https://github.com/sreeram-venkitesh/karpenter/actions/runs/7323345505) and the generated json file [here](https://github.com/sreeram-venkitesh/karpenter/releases/tag/v0.34.9)

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,27 @@
...
+jobs:
+  vexctl:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Set tag name
+        shell: bash
+        run: |
+          echo "TAG=${{ github.ref_name }}" >> "$GITHUB_ENV"
```
jonathan-innis, line 17: This still allows script injection. Can we directly assign the `${{ github.ref_name }}` in the env block like

```yaml
jobs:
  vexctl:
    runs-on: ubuntu-latest
    env:
      TAG: ${{ github.ref_name }}
    steps:
    ...
```

.github/workflows/openvex.yaml
jonathan-innis: Thanks @sreeram-venkitesh! One comment on how we should set our environment variables.

sreeram-venkitesh:
@jonathan-innis Please take a look!

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,26 @@
...
+        name: Run vexctl
+        with:
+          product: pkg:generic/karpenter@${{ env.TAG }}
+          file: /tmp/karpenter.vex.json
+      - name: Upload openvex data
+        env:
+          # Secrets for the workflow can be added at https://github.com/kubernetes-sigs/karpenter/settings/secrets/actions
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload ${{ env.TAG }} /tmp/karpenter.vex.json
```
jonathan-innis, line 26:
```suggestion
          gh release upload "$TAG" /tmp/karpenter.vex.json
```

This is safer from a script-injection perspective

jonathan-innis: Should this be part of `release.yaml` directly? I'm realizing now that this is a separate workflow, which is going to have a race with our other release workflow that is triggered on the same event. This should really be a file that we generate after we have already created the release. Or better yet, adding a file to the file block on release creation here: https://github.com/kubernetes-sigs/karpenter/blob/main/.github/workflows/release.yaml#L29

sreeram-venkitesh: Hmm, makes sense. Would moving the `Run vexctl` and `Upload openvex data` steps to after the `Create GitHub release` step in release.yaml be correct?

jonathan-innis: I think if you move the file generation before the actual release creation step, then you can just add the generated file to the `files:` block in the release creation.

sreeram-venkitesh: @jonathan-innis I've made the change. Tried testing the workflow from my repository, but I couldn't get the `Run tejolote` step to work properly. I had to manually create a release for that step to pass, and once the vex json file was also created, the workflow couldn't create a release since I had already created one with the same tag. I believe this wouldn't be an issue when triggering the run from the main repository, although I'd want to run a sanity check for that. The openvex json file is getting generated properly though, as seen in this run [here](https://github.com/sreeram-venkitesh/karpenter/actions/runs/7332501501/job/19966733748). Please let me know what you think!

.github/workflows/release.yaml
```diff
@@ -18,14 +20,20 @@ jobs:
         uses: kubernetes-sigs/release-actions/setup-tejolote@dd08496c83441d6477114cc0555b96d404dacff7 # v0.1.2
 
       - name: Run tejolote 
-        env:
-          TAG: ${{ github.ref_name }}
         run: |
-          tejolote attest --artifacts github://kubernetes-sigs/karpenter/"${TAG}" github://kubernetes-sigs/karpenter/"${{ github.run_id }}" --output karpenter.intoto.json --sign
+          tejolote attest --artifacts github://kubernetes-sigs/karpenter/${{ env.TAG }} github://kubernetes-sigs/karpenter/"${{ github.run_id }}" --output karpenter.intoto.json --sign
```
jonathan-innis, line 24:
```suggestion
          tejolote attest --artifacts "github://kubernetes-sigs/karpenter/$TAG" "github://kubernetes-sigs/karpenter/${{ github.run_id }}" --output karpenter.intoto.json --sign
```

This syntax avoids the script injection possibility

.github/workflows/release.yaml
```diff
@@ -18,14 +20,20 @@ jobs:
...
+
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
```
jonathan-innis, line 26: Can you add the version at the end here as a comment for `generate-vex`

.github/workflows/openvex.yaml
jonathan-innis: @sreeram-venkitesh Looking at it a little deeper, it looks like the `tejolote` attestation file needs to be generated _after_ the release is created. Testing in my branch, it looks like we need to move this below the release action to get everything to work appropriately.

jonathan-innis: I was able to get this working. There were a few misses. See the changes made here: https://github.com/kubernetes-sigs/karpenter/pull/914

.github/workflows/release.yaml
sreeram-venkitesh: Since this has been changed in https://github.com/kubernetes-sigs/karpenter/commit/d4ff37c70609c2da3afa85924707ecaef1e71a22, I've rebased the PR. Now it only has the changes related to openvex.

.github/workflows/release.yaml
sreeram-venkitesh: [openvex/generate-vex](https://github.com/openvex/generate-vex) doesn't have a release yet. I'm not sure what version we need to mention.

jonathan-innis (approved):
Nice work 🎉 

/lgtm
/approve

Issue #876, opened 2023-12-15
Security Slam: Kubernetes Lightning Round


ricardoapl:
### Description

Open tasks for Kubernetes Security Slam 2023

**Release to Community Infrastructure**

Tasks 1 and 2 are not relevant, because images are not part of the release process for Karpenter

- [x] Task 1: Releases are being made via CI (https://github.com/kubernetes-sigs/karpenter/issues/868)
- [x] Task 2: Check that Images are Staged and Promoted Using Community Infra (https://github.com/kubernetes-sigs/karpenter/issues/874)

**Automate Security Documentation**

Task 3 is not relevant, because Karpenter is not a separate distributable artifact

- [x] Task 3: Ensure SBOMs are Generated by Kubernetes BOM (https://github.com/kubernetes-sigs/karpenter/pull/871)
- [x] Task 4: Ensure SLSA Attestations are Generated when Possible (https://github.com/kubernetes-sigs/karpenter/pull/889)
- [x] Task 5: Ensure the Project has a VEX Feed (https://github.com/kubernetes-sigs/karpenter/pull/869 and https://github.com/kubernetes-sigs/karpenter/pull/870)

**CLOMonitor: Secure Development Practices**

- [x] Task 7: Set up CLOMonitor Tracking (https://github.com/cncf/clomonitor/pull/1385)
- [x] Task 8: Check for Binary Artifacts (no binaries found in the repo)
- [x] Task 9: Review the Code Review (all changesets reviewed)
- [x] Task 10: Dangerous Workflow (no dangerous workflow patterns detected)
- [x] Task 11: Security Insights (https://github.com/kubernetes-sigs/karpenter/issues/877 and https://github.com/kubernetes-sigs/karpenter/pull/878)
- [x] Task 12: Dependencies Policy (https://github.com/kubernetes-sigs/karpenter/pull/900)
- [x] Task 13: Dependency update tool (update tool detected, dependabot)
- [x] Task 15: Token Permissions (https://github.com/kubernetes-sigs/karpenter/pull/884)

This checklist was developed by @eddie-knight and @puerco to help guide contributions. When the project tracking kicks in on https://clomonitor.io/projects/cncf/karpenter then more of the security hygiene state will be visible as well.

coder12git:
@ricardoapl we need to update this

ricardoapl:
Thank you @coder12git, should be updated now

Please let me know if I missed something

/cc @YashPimple @sreeram-venkitesh

[...]

PR #884, opened 2023-12-16
ci(workflows): least privileged token permissions

ricardoapl:
Fixes #N/A 

Ref. https://github.com/kubernetes-sigs/karpenter/issues/876

**Description**

The `GITHUB_TOKEN` is an automatically generated secret to make authenticated calls to the GitHub API. GitHub recommends setting minimum token permissions for the `GITHUB_TOKEN`. Current configuration seems to be secure, but there is a chance that when a new job is added to one of the workflows, its job permissions could be left undefined because of human error. We can prevent that by setting permissions definitions in each workflow's yaml file as read-only at the [top level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions).

- [GitHub Security Guide](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow)
- [The Open Source Security Foundation (OpenSSF) Security Guide](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions)

### DIFF
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -7,6 +7,8 @@ on:
   pull_request:
   schedule:
     - cron: '0 12 * * *'
+permissions:
+  contents: read
 jobs:
   analyze-go:
     name: Analyze Go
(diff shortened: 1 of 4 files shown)


jonathan-innis (approved):
LGTM 🚀 We set the default token permissions to read-only by default but this adds another layer

/lgtm
/approve

PR #899, opened 2023-12-20
chore: Update ephemeral taints logic and drop unmanaged check on nodes

jonathan-innis:
Fixes #N/A 

**Description**

This PR changes the ephemeral taints logic that we use for removing certain taints for consideration when nodes managed by Karpenter haven't yet initialized. This changes the logic so that this logic doesn't also apply to unmanaged nodes as well as ensuring that we only consider these taints _up_ to the point that the node initializes. Once a node has reached its initialization point, we will begin considering these taints.

This PR drops an additional check that is unneeded now that we are no longer on a migration path. Nodes that would fall under the additional check (that is -- nodes that would have the `karpenter.sh/nodepool` label but don't have an associated NodeClaim) will only exist when the NodeClaim is deleted _immediately_ after creation (since the node would exist for a bit of time while the instance is terminating). In this case, we shouldn't consider the node managed since the instance is going away and the node will get deleted soon by CCM.

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -39,7 +39,6 @@ import (
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/events"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
-	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
 	"sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
@@ -111,11 +110,12 @@ func simulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 	}
 
 	results := scheduler.Solve(ctx, pods)
-	// check if the scheduling relied on an existing node that isn't ready yet, if so we fail
-	// to schedule since we want to assume that we can delete a node and its pods will immediately
-	// move to an existing node which won't occur if that node isn't ready.
 	for _, n := range results.ExistingNodes {
-		if !n.Initialized() || nodeutils.GetCondition(n.Node, v1.NodeReady).Status != v1.ConditionTrue {
+		// We consider existing nodes for scheduling. When these nodes are unmanaged, their taint logic should
+		// tell us if we can schedule to them or not; however, if these nodes are managed, we will still schedule to them
+		// even if they are still in the middle of their initialization loop. In the case of disruption, we don't want
+		// to proceed disrupting if our scheduling decision relies on nodes that haven't entered a terminal state.
+		if !n.Initialized() {
 			for _, p := range n.Pods {
 				results.PodErrors[p] = fmt.Errorf("would schedule against a non-initialized node %s", n.Name())
 			}
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -160,11 +160,6 @@ func (in *StateNode) Annotations() map[string]string {
 	return in.Node.Annotations
 }
 
-// GetLabels exists to conform to the GetLabels() function on the metav1.Object interface
-func (in *StateNode) GetLabels() map[string]string {
-	return in.Labels()
-}
-
 func (in *StateNode) Labels() map[string]string {
 	// If the nodeclaim exists and the state node isn't registered
 	// use the nodeclaim representation of the labels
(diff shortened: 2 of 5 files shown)


pkg/controllers/state/statenode.go
```diff
@@ -181,26 +176,33 @@ func (in *StateNode) Labels() map[string]string {
...
-	// re-appears on the node for a different reason (e.g. the node is cordoned) we will assume that pods can
-	// schedule against the node in the future incorrectly.
-	ephemeralTaints := scheduling.KnownEphemeralTaints
-	if !in.Initialized() && in.Managed() && in.NodeClaim != nil {
-		ephemeralTaints = append(ephemeralTaints, in.NodeClaim.Spec.StartupTaints...)
-	}
-
 	var taints []v1.Taint
-	if (!in.Registered() && in.NodeClaim != nil) || in.Node == nil {
+	if (!in.Registered() && in.Managed()) || in.Node == nil {
```
njtran, line 180: Worth a comment detailing when a nodeclaim could be not registered but there be a node object and vice versa?

pkg/controllers/state/statenode.go
```diff
@@ -181,26 +176,33 @@ func (in *StateNode) Labels() map[string]string {
...
+			}); found {
+				return true
+			}
+			return false
 		})
-		return found
-	})
+	} else {
+		return taints
+	}
```
njtran, line 205: nit: you don't need the else, right?

pkg/controllers/state/statenode.go
```diff
@@ -307,8 +309,7 @@ func (in *StateNode) Nominated() bool {
 }
 
 func (in *StateNode) Managed() bool {
-	return in.NodeClaim != nil ||
-		(in.Node != nil && in.Node.Labels[v1beta1.NodePoolLabelKey] != "")
+	return in.NodeClaim != nil
```
njtran, line 312: Does this remove the case where a nodeclaim could be deleted before a node is? In this case, now those nodes that are leaked are no longer seen as managed nodes, and get indefinitely leaked. If possible, this seems like a pretty gnarly edge case.

pkg/controllers/state/suite_test.go
```diff
@@ -1432,17 +1568,3 @@ func ExpectStateNodeCount(comparator string, count int) int {
 	Expect(c).To(BeNumerically(comparator, count))
 	return c
 }
-
-func ExpectStateNodeNotFoundForNodeClaim(nodeClaim *v1beta1.NodeClaim) *state.StateNode {
```
njtran, line 1436: goodbye friend

pkg/controllers/state/suite_test.go
```diff
@@ -1422,6 +1423,141 @@ var _ = Describe("Consolidated State", func() {
...
+			},
+		})
+	})
+	Context("Managed", func() {
+		It("should not consider ephemeral taints on a managed node that isn't initialized", func() {
+			node.Spec.Taints = []v1.Taint{
+				{Key: v1.TaintNodeNotReady, Effect: v1.TaintEffectNoSchedule},
+				{Key: v1.TaintNodeUnreachable, Effect: v1.TaintEffectNoSchedule},
+				{Key: cloudproviderapi.TaintExternalCloudProvider, Effect: v1.TaintEffectNoSchedule, Value: "true"},
+			}
```
njtran, line 1446: Worth using the same ephemeral taints list variable here so that this is automatically updated when we change that variable?

pkg/controllers/disruption/helpers.go
```diff
@@ -111,11 +110,12 @@ func simulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
...
-	// check if the scheduling relied on an existing node that isn't ready yet, if so we fail
-	// to schedule since we want to assume that we can delete a node and its pods will immediately
-	// move to an existing node which won't occur if that node isn't ready.
 	for _, n := range results.ExistingNodes {
-		if !n.Initialized() || nodeutils.GetCondition(n.Node, v1.NodeReady).Status != v1.ConditionTrue {
+		// We consider existing nodes for scheduling. When these nodes are unmanaged, their taint logic should
+		// tell us if we can schedule to them or not; however, if these nodes are managed, we will still schedule to them
+		// even if they are still in the middle of their initialization loop. In the case of disruption, we don't want
+		// to proceed disrupting if our scheduling decision relies on nodes that haven't entered a terminal state.
+		if !n.Initialized() {
```
njtran, line 118: The initialized logic isn't changed with respect to the NodeReady condition. I thought we were going to model initialization everywhere with the existing condition.

pkg/controllers/state/statenode.go
jonathan-innis: This seems like the correct concession to make. The only time that a Node should be deleted _after_ a NodeClaim is when the Node registers in a way that doesn't allow us to see the node while we are going through NodeClaim deletion. In this case, I don't think that it makes sense for us to consider this node a managed node. The node will be going away shortly since the instance has gotten removed and we can't do anything with the instance since we've lost its NodeClaim representation.

pkg/controllers/state/suite_test.go
jonathan-innis: I don't love using constants that are defined within your own package in testing since it can mask bugs. Imagine that we accidentally change the ephemeralTaints value in the state package and now it doesn't include one of these taints. That wouldn't cause a testing failure when I'd argue that it should since you are now not considering one of the taints that you had statically intended to consider.

pkg/controllers/disruption/helpers.go
jonathan-innis: Discussed offline: Nodes that aren't managed have a `kubernetes.io/not-ready` taint added to them which means that we wouldn't schedule against them. The initialized concept should _only_ be used as a Karpenter-managed node concept and it just means that the node hasn't been through its initialization loop once. Once a Karpenter node goes through its initialization loop, it is marked as initialized with a label and it can receive taints that will affect scheduling like any other node on the cluster that is unmanaged.

pkg/controllers/state/suite_test.go
jonathan-innis: Discussed offline: we both think that it's reasonable to use the static value in testing to avoid cases like this.

pkg/controllers/state/statenode.go
jonathan-innis: Discussed offline: we agreed that there are two scenarios that we need to address here:
1. The NodeClaim and Node both exist on the cluster when the NodeClaim is deleted. In this case, Karpenter will wait for the deletion of the nodes before fully removing the NodeClaim from the cluster. This ensures that the node is properly drained and terminated before cleaning up its owning NodeClaim (and its instance representation in the cloud provider)
2. The NodeClaim is deleted, doesn't see any Nodes, and then the Node joins after the NodeClaim has been deleted. In particular, we have seen this happen in AWS since the the AWS Cloud Provider doesn't wait for instances to be fully terminated before it actually deletes the NodeClaim. It just expects that the NodeClaim goes into a terminating state. The problem with only ensuring this is that userData on the EC2 instance may still run, which means that the Node may still register after the NodeClaim has gone. Ideally, we shouldn't hit this case; however, even if we do hit this case, we should just treat the Node as an unmanaged node. From our perspective, Cloud Controller Manager will eventually delete the node and the node is actually not a good place to schedule container capacity since the underlying instance is actually actively being removed.

njtran (approved):
/lgtm
/approve

Issue #895, opened 2023-12-19
Karpenter KwoK Cloud Provider


njtran:
### Description

**What problem are you trying to solve?**
Developers that want to make contributions to this repo must build and install with a cloud provider specific installation. This can be costly for users to develop and test their features, and may require maintainers to do that testing for them. Additionally, since this comes with Cloud Provider specific details, users contributing just to kubernetes-sigs/karpenter have to be aware of how the specific CP implementation may have decided to reason about general concepts. 

Creating a cloud provider using [KwoK nodes](https://github.com/kubernetes-sigs/kwok/tree/main/pkg) won't incur costs for users that need to test and validate their changes.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
So excited about this! This will be awesome when we get this in!

njtran:
For those interested, added a design doc in #905 to detail some of the info I used to drive what I thought was a good set of instance types for #904.

PR #905, opened 2023-12-22
RFC: add in kwok design living doc

njtran:
Fixes #N/A 

**Description**
Adds in the docs guide that will be referenced for the kwok cloud provider. Goes into the ideas of why the specific instance types were picked. 

Related: https://github.com/kubernetes-sigs/karpenter/issues/895

### DIFF
--- a/designs/kwok-provider.md
+++ b/designs/kwok-provider.md
@@ -2,7 +2,7 @@
 
 The kubernetes-sigs/karpenter repo uses a simulated kubernetes environment to test code changes. Users must use a cloud provider of their choice to test their changes, which can be costly and creates an unreasonable barrier to entry for contributors. Related issue: https://github.com/kubernetes-sigs/karpenter/issues/895.
 
-Creating a Cloud Provider neutral implementation of Karpenter within Karpenter-core requires creating a set of CP neutral examples of instance types and pricing that captures common tenets across CPs. 
+Creating a Cloud Provider neutral implementation of Karpenter in kubernetes-sigs/karpenter requires creating a set of CP neutral examples of instance types and pricing that captures common tenets across CPs. 
 
 This is meant to be a living document to in-depth go through the decisions on how kubernetes-sigs/karpenter models its fake instance types. 
 
@@ -22,15 +22,15 @@ The full set of instance types for the KwoK Cloud Provider should be at least th
 ```
 
 #### Size
-While AWS uses `medium`, `large`, `xlarge`, `2xlarge`, `4xlarge` and so on, AKS (include cpu value in name) and GCE (suffixed with cpu value) use a different naming convention, generally increasing in offering sizes by a factor of 1.5x or 2x. In addition, in case the factors of 1.5x and 2x are incorrectly modeled, we'll add in the 9x and 100x instance sizes. 
+While AWS uses `medium`, `large`, `xlarge`, `2xlarge`, `4xlarge` and so on, Azure (includes cpu value in name) and GCE (suffixed with cpu value) use a different naming convention, which all usually have a list of offerings that increase in size by a factor of 1.5x (32 -> 48) or 2x (16 -> 32). In addition, in case 1.5x and 2x are incorrectly modeled, we'll add in the 9x and 100x instance sizes. 
 
 #### Ratios
-AWS Instance Types fall under many "Instance Families" that give a high level overview to why a user may want to use the instance type. AWS `m` instance types are labeled General Purpose, and `c` and `r` represent Compute Optimized and Memory Optimized, most easily compared by the ratio of Memory per vCPU. Defining these ratios is important to test Karpenter's bin-packing value, where applications with more resource-skewed requirements can be more cost effectively bin-packed. 
+AWS Instance Types are categorized through "Instance Families" based on the instance type's attributes. AWS `m` instance types are labeled General Purpose, and `c` and `r` represent Compute Optimized and Memory Optimized, most easily compared by the ratio of Memory per vCPU. Defining these ratios is important to test Karpenter's bin-packing value, where applications with more resource-skewed requirements can be more cost effectively bin-packed. 
 
-These ratios seem to accurately depict a common ratios, where "standard/general" ratios for different cloud providers are seen below:
+These ratios seem to accurately depict common ratios, where "standard/general" ratios for different cloud providers are seen below:
 - AWS: General     = 1 vCPU : 4 GiB
-- AKS: D Family    = 2 vCPU : 8 GiB
-- GKE: C3 Standard = 4 vCPU : 16 GiB
+- Azure: D Family    = 2 vCPU : 8 GiB
+- GCE: C3 Standard = 4 vCPU : 16 GiB
 
 The Kwok provider will be able to select on `karpenter.sh/instance-memory` and `karpenter.sh/instance-cpu`.
 
@@ -38,11 +38,11 @@ The Kwok provider will be able to select on `karpenter.sh/instance-memory` and `
 Karpenter v1beta1 APIs use `on-demand` and `spot` as the options for `karpenter.sh/capacity-type`, reflecting how they're referenced in AWS and elsewhere. In my docs search, I've found that Cloud Providers have different names for on-demand (e.g. Regular for AKS and Standard for GKE). Since this is how Karpenter defines the capacity type labels, the KwoK CP will use the values defined in the project, regardless of how this value changes in the future.
 
 ### Pricing
-The pricing for each of the instance types will be a function of the size and capacity type.
+The pricing for each of the instance types will be a function of the size and capacity type. Instance types with the spot capacity type are guaranteed to be cheaper than their on-demand counterparts. Each cloud provider varies in the pricing discount for spot instances:
 
-EKS: Discount is [< 90%](https://aws.amazon.com/ec2/spot/pricing/#:~:text=Spot%20Instances%20are%20available%20at%20a%20discount%20of%20up%20to%2090%25%20off%20compared%20to%20On%2DDemand%20pricing.)
+AWS: Discount is [< 90%](https://aws.amazon.com/ec2/spot/pricing/#:~:text=Spot%20Instances%20are%20available%20at%20a%20discount%20of%20up%20to%2090%25%20off%20compared%20to%20On%2DDemand%20pricing.)
 AKS: Discount is [48-90%](https://azure.microsoft.com/en-us/pricing/details/virtual-machine-scale-sets/linux/)
-GKE: Discount is [60-91%](https://cloud.google.com/compute/docs/instances/create-use-spot#:~:text=Spot%20VMs%20are%20available%20at%20a%2060%2D91%25%20discount%20compared%20to%20the%20price%20of%20standard%20VMs.)
+GCE: Discount is [60-91%](https://cloud.google.com/compute/docs/instances/create-use-spot#:~:text=Spot%20VMs%20are%20available%20at%20a%2060%2D91%25%20discount%20compared%20to%20the%20price%20of%20standard%20VMs.)
 
 We'll define the pricing functions:
 
--- a/designs/kwok-provider.md
+++ b/designs/kwok-provider.md
@@ -13,7 +13,7 @@ The full set of instance types for the KwoK Cloud Provider should be at least th
 
 ```md
 1. Instance Type Name: (family)-(size)
-    a. size: 1x, 2x, 4x, 8x, 16x, 32x, 48x, 64x, 96x, 128x, 192x, 256x
+    a. size: 1x, 2x, 4x, 9x, 16x, 32x, 48x, 64x, 100x, 128x, 192x, 256x
     b. family: c (cpu), s (standard), m (memory)
         i. c ---> 1 vCPU : 2 GiB
         ii. s --> 1 vCPU : 4 GiB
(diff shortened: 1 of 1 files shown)


njtran:
cc: @tallaxes @jackfrancis

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+This is meant to be a living document to in-depth go through the decisions on how kubernetes-sigs/karpenter models its fake instance types. 
+
+## Proposal
+
+### Instance Types
+The full set of instance types for the KwoK Cloud Provider should be at least the cartesian product of the following dimensions, generating 60 instance types across 4 zones, totaling `12 * 3 * 2 = 64` in-memory Cloud Provider instance types.
+
+```md
+1. Instance Type Name: (family)-(size)
+    a. size: 1x, 2x, 4x, 8x, 16x, 32x, 48x, 64x, 96x, 128x, 192x, 256x
```
sftim, line 16:
```suggestion
    a. size: 1x, 2x, 4x, 9x, 32x, 100x, 128x, 256x
```
This deliberately doesn't match any particular cloud provider's size ratios, but does allow testing the logic side independent of what the real-world ratios might be (you can always find a node set to provide capacity for a workload).

If we accidentally coded in a power-of-two assumption, the 9x side ought to help pick that up.

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+    b. family: c (cpu), s (standard), m (memory)
+        i. c ---> 1 vCPU : 2 GiB
+        ii. s --> 1 vCPU : 4 GiB
+        iii. m -> 1 vCPU : 8 GiB
```
sftim, line 20: If possible, I'd like to have a test case with less than 1vCPU.

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+Creating a Cloud Provider neutral implementation of Karpenter within Karpenter-core requires creating a set of CP neutral examples of instance types and pricing that captures common tenets across CPs. 
+
+This is meant to be a living document to in-depth go through the decisions on how kubernetes-sigs/karpenter models its fake instance types. 
+
+## Proposal
+
+### Instance Types
+The full set of instance types for the KwoK Cloud Provider should be at least the cartesian product of the following dimensions, generating 60 instance types across 4 zones, totaling `12 * 3 * 2 = 64` in-memory Cloud Provider instance types.
+
+```md
```
sftim, line 14: I recommend making fake Linux nodes and fake Windows nodes, which means that `kubernetes.io/os` would be relevant too.

We can support fake GLADOS nodes too if we want, and then verify that neither Windows nor Linux Pods get scheduled there.

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+AKS: Discount is [48-90%](https://azure.microsoft.com/en-us/pricing/details/virtual-machine-scale-sets/linux/)
+GKE: Discount is [60-91%](https://cloud.google.com/compute/docs/instances/create-use-spot#:~:text=Spot%20VMs%20are%20available%20at%20a%2060%2D91%25%20discount%20compared%20to%20the%20price%20of%20standard%20VMs.)
+
+We'll define the pricing functions:
+
+```markdown
+1. price_of_base_size = (#vCPU * 0.25) + (#GiB * 0.01) # Roughly calculated based on guessing pricing functions from multiple cloud providers. 
+2. on_demand_price    = size * price_of_base_size()
+3. spot_price         = on_demand_price * .7 # Some discount factor to reflect savings somewhere in the middle of the referenced CPs.
+```
```
sftim, line 51: If we're doing abstract testing, let's plan to one day handle a free tier.

For example, a test scenario where you can have up to two on-demand `c-1x` instances for free, regardless of your other consumption, but any use beyond that costs you actual money.

So this comment is to mark that we might one day want to account for that, not a request to put it in immediately.

sftim (commented):
Some thoughts

designs/kwok-provider.md
njtran: Good point. I'll make the change to add in 9x. Any reason you think it should replace the 32x and the same for 100x for 48x, 64x, and 96x?

designs/kwok-provider.md
njtran: What's the thought behind having less than 1vCPU if we're not actually running the workloads? My thought was that one could scale their workload to the desired minimum node size.

designs/kwok-provider.md
njtran: I'm not quite following. Seems pretty bespoke to simulate a user trying out Karpenter, where I don't see much of the value in that.

sftim: Sure, don't worry about this detail.

njtran:
@Bryce-Soghigian curious if you have any thoughts here

designs/kwok-provider.md
njtran: Yeah currently it'll support linux and windows, so we should be good here. It would be pretty easy to add in new OS keys in the future too!

Bryce-Soghigian (commented):
Overall looks solid

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+
+These ratios seem to accurately depict a common ratios, where "standard/general" ratios for different cloud providers are seen below:
+- AWS: General     = 1 vCPU : 4 GiB
+- AKS: D Family    = 2 vCPU : 8 GiB
+- GKE: C3 Standard = 4 vCPU : 16 GiB
+
+#### Capacity Type Names
+Karpenter v1beta1 APIs use `on-demand` and `spot` as the options for `karpenter.sh/capacity-type`, reflecting how they're referenced in AWS and elsewhere. In my docs search, I've found that Cloud Providers have different names for on-demand (e.g. Regular for AKS and Standard for GKE). Since this is how Karpenter defines the capacity type labels, the KwoK CP will use the values defined in the project, regardless of how this value changes in the future.
```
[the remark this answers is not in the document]
njtran: I think a doc that highlights core differences in the providers would be nice to have, especially from a maintainer's perspective, to know what assumptions can be taken that's true across all providers!

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Proposal
+
+### Instance Types
+The full set of instance types for the KwoK Cloud Provider should be at least the cartesian product of the following dimensions, generating 60 instance types across 4 zones, totaling `12 * 3 * 2 = 64` in-memory Cloud Provider instance types.
+
+```md
+1. Instance Type Name: (family)-(size)
+    a. size: 1x, 2x, 4x, 9x, 16x, 32x, 48x, 64x, 100x, 128x, 192x, 256x
+    b. family: c (cpu), s (standard), m (memory)
```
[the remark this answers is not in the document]
njtran: I had considered adding an `exotic` or `gpu` type, but was planning on keeping it simple with just the basic mem/cpu/pods resources. Definitely open to adding this in after.

designs/kwok-provider.md
```diff
@@ -0,0 +1,51 @@
...
+        i. c ---> 1 vCPU : 2 GiB
+        ii. s --> 1 vCPU : 4 GiB
+        iii. m -> 1 vCPU : 8 GiB
+2. Capacity Type: [on-demand, spot]
```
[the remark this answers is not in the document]
njtran: Totally agree. This is a good callout. It would be super useful to have a shared list here. 

The kwok provider should definitely be able to select by vCPU and GPU (when GPUs are implemented)

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
+# Karpenter KwoK Cloud Provider - v0.34+
+
+The kubernetes-sigs/karpenter repo uses a simulated kubernetes environment to test code changes. Users must use a cloud provider of their choice to test their changes, which can be costly and creates an unreasonable barrier to entry for contributors. Related issue: https://github.com/kubernetes-sigs/karpenter/issues/895.
+
+Creating a Cloud Provider neutral implementation of Karpenter within Karpenter-core requires creating a set of CP neutral examples of instance types and pricing that captures common tenets across CPs.
```
jonathan-innis, line 5:
```suggestion
Creating a Cloud Provider neutral implementation of Karpenter within `kubernetes-sigs/karpenter` requires creating a set of CP neutral examples of instance types and pricing that captures common tenets across CPs. 
```

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
...
+    a. size: 1x, 2x, 4x, 9x, 16x, 32x, 48x, 64x, 100x, 128x, 192x, 256x
+    b. family: c (cpu), s (standard), m (memory)
+        i. c ---> 1 vCPU : 2 GiB
+        ii. s --> 1 vCPU : 4 GiB
+        iii. m -> 1 vCPU : 8 GiB
+2. Capacity Type: [on-demand, spot]
+```
+
+#### Size
+While AWS uses `medium`, `large`, `xlarge`, `2xlarge`, `4xlarge` and so on, AKS (include cpu value in name) and GCE (suffixed with cpu value) use a different naming convention, generally increasing in offering sizes by a factor of 1.5x or 2x. In addition, in case the factors of 1.5x and 2x are incorrectly modeled, we'll add in the 9x and 100x instance sizes.
```
jonathan-innis, line 25:
```suggestion
While EKS uses `medium`, `large`, `xlarge`, `2xlarge`, `4xlarge` and so on, AKS (include cpu value in name) and GKE (suffixed with cpu value) use a different naming convention, generally increasing in offering sizes by a factor of 1.5x or 2x. In addition, in case the factors of 1.5x and 2x are incorrectly modeled, we'll add in the 9x and 100x instance sizes. 
```

nit: Either reference the service for each CP or reference the cloud name for each, I wouldn't mix them

jonathan-innis, line 25: I'm not sure that I understand what this statement means "generally increasing in offering sizes by a factor of 1.5x or 2x."

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
...
+        ii. s --> 1 vCPU : 4 GiB
+        iii. m -> 1 vCPU : 8 GiB
+2. Capacity Type: [on-demand, spot]
+```
+
+#### Size
+While AWS uses `medium`, `large`, `xlarge`, `2xlarge`, `4xlarge` and so on, AKS (include cpu value in name) and GCE (suffixed with cpu value) use a different naming convention, generally increasing in offering sizes by a factor of 1.5x or 2x. In addition, in case the factors of 1.5x and 2x are incorrectly modeled, we'll add in the 9x and 100x instance sizes. 
+
+#### Ratios
+AWS Instance Types fall under many "Instance Families" that give a high level overview to why a user may want to use the instance type. AWS `m` instance types are labeled General Purpose, and `c` and `r` represent Compute Optimized and Memory Optimized, most easily compared by the ratio of Memory per vCPU. Defining these ratios is important to test Karpenter's bin-packing value, where applications with more resource-skewed requirements can be more cost effectively bin-packed.
```
jonathan-innis, line 28:
```suggestion
AWS Instance Types are categorized into "Instance Families" based on the characteristics of the instance types. AWS `m` instance types are labeled General Purpose, and `c` and `r` represent Compute Optimized and Memory Optimized, most easily compared by the ratio of Memory per vCPU. Defining these ratios is important to test Karpenter's bin-packing value, where applications with more resource-skewed requirements can be more cost effectively bin-packed. 
```

nit: Clarity on the wording here

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
...
+2. Capacity Type: [on-demand, spot]
+```
+
+#### Size
+While AWS uses `medium`, `large`, `xlarge`, `2xlarge`, `4xlarge` and so on, AKS (include cpu value in name) and GCE (suffixed with cpu value) use a different naming convention, generally increasing in offering sizes by a factor of 1.5x or 2x. In addition, in case the factors of 1.5x and 2x are incorrectly modeled, we'll add in the 9x and 100x instance sizes. 
+
+#### Ratios
+AWS Instance Types fall under many "Instance Families" that give a high level overview to why a user may want to use the instance type. AWS `m` instance types are labeled General Purpose, and `c` and `r` represent Compute Optimized and Memory Optimized, most easily compared by the ratio of Memory per vCPU. Defining these ratios is important to test Karpenter's bin-packing value, where applications with more resource-skewed requirements can be more cost effectively bin-packed. 
+
+These ratios seem to accurately depict a common ratios, where "standard/general" ratios for different cloud providers are seen below:
```
jonathan-innis, line 30:
```suggestion
These ratios seem to accurately depict common ratios, where "standard/general" ratios for different cloud providers are seen below:
```

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
...
+- AWS: General     = 1 vCPU : 4 GiB
+- AKS: D Family    = 2 vCPU : 8 GiB
+- GKE: C3 Standard = 4 vCPU : 16 GiB
+
+The Kwok provider will be able to select on `karpenter.sh/instance-memory` and `karpenter.sh/instance-cpu`.
```
jonathan-innis, line 35:
```suggestion
The Kwok provider will be able to select on `karpenter.sh/instance-memory` and `karpenter.sh/instance-cpu`.
```

Interesting. Are we saying that this requirement should be pulled up into the neutral requirements logic by doing this?

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
...
+
+#### Capacity Type Names
+Karpenter v1beta1 APIs use `on-demand` and `spot` as the options for `karpenter.sh/capacity-type`, reflecting how they're referenced in AWS and elsewhere. In my docs search, I've found that Cloud Providers have different names for on-demand (e.g. Regular for AKS and Standard for GKE). Since this is how Karpenter defines the capacity type labels, the KwoK CP will use the values defined in the project, regardless of how this value changes in the future.
+
+### Pricing
+The pricing for each of the instance types will be a function of the size and capacity type.
+
+EKS: Discount is [< 90%](https://aws.amazon.com/ec2/spot/pricing/#:~:text=Spot%20Instances%20are%20available%20at%20a%20discount%20of%20up%20to%2090%25%20off%20compared%20to%20On%2DDemand%20pricing.)
```
jonathan-innis, line 43: Not sure I get what you mean by "discount" here. Is this referring to the spot discount? If so, that's not immediately clear from the doc

designs/kwok-provider.md
njtran: I meant it successively:  1x, 2x, 4x, 9x, 16x, 32x, 48x, 64x, 100x, 128x, 192x, 256x

for instance 32 -> 48 is multiplying by a factor of 1.5, 2 -> 4 is multiplying by a factor of 2. Let me adjust the language to make it clearer.

designs/kwok-provider.md
njtran: I'd consider this a separate effort, but @Bryce-Soghigian calls this out as something we may want to consider doing in the future.

designs/kwok-provider.md
```diff
@@ -0,0 +1,53 @@
+# Karpenter KwoK Cloud Provider - v0.34+
```
jonathan-innis, line 1: Are we deciding to keep the version in the title here?

designs/kwok-provider.md
jonathan-innis: Yeah, more of a musing. I was just realizing that if we were doing things like this, then it starts to probe that question

jonathan-innis (commented):
One comment on the title

jonathan-innis (approved):
/lgtm
/approve

PR #906, opened 2023-12-22
feat: add kwok install hack script

njtran:
Fixes #N/A 

**Description**
This adds in a hack/kwok-install.sh script that will allow users to install Kwok with 10 partitions in their cluster. 10 partitions are needed as Kwok has [scalability limitations](https://github.com/kubernetes-sigs/kwok#:~:text=Lightweight%3A%20You%20can%20simulate%20thousands%20of%20nodes%20on%20your%20laptop%20without%20significant%20consumption%20of%20CPU%20or%20memory%20resources.%20Currently%2C%20KWOK%20can%20reliably%20maintain%201k%20nodes%20and%20100k%20pods%20easily.)

This kwok-install.sh script can be improved by doing kustomization to only apply 10 different deployments, rather than 10 copies of all objects associated with the manifest. 

Related: https://github.com/kubernetes-sigs/karpenter/issues/895

`make install-kwok` and `make uninstall-kwok`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,11 @@ deflake: ## Run randomized, racing tests until the test fails to catch flakes
 		-v \
 		./pkg/...
 
+install-kwok: ## Install kwok into cluster
+	UNINSTALL_KWOK=false ./hack/install-kwok.sh
+uninstall-kwok: ## Install kwok provider
+	UNINSTALL_KWOK=true ./hack/install-kwok.sh
+
 vulncheck: ## Verify code vulnerabilities
 	@govulncheck ./pkg/...
 
--- a/hack/install-kwok.sh
+++ b/hack/install-kwok.sh
@@ -1,16 +1,23 @@
+# This script uses kustomize to generate the manifests of the latest 
+# version of kwok, and will then either apply/delete the resources 
+# depending on the UNINSTALL_KWOK environment variable. Set this to 
+# true if you want to delete the generated objects.
+
+# This will create 10 different partitions, so that Karpenter can scale 
+# to 10000 nodes, as each kwok controller can only reliably handle about
+# 1000 nodes.
 set -euo pipefail
 
 # get the latest version
 KWOK_REPO=kubernetes-sigs/kwok
 # KWOK_LATEST_RELEASE=$(curl "https://api.github.com/repos/${KWOK_REPO}/releases/latest" | jq -r '.tag_name')
-KWOK_LATEST_RELEASE="v0.3.0"
-
+KWOK_LATEST_RELEASE=v0.3.0
 # make a base directory for multi-base kustomization
 HOME_DIR=$(mktemp -d)
 BASE=${HOME_DIR}/base
 mkdir ${BASE}
 
-# allow it to tolerate everything, but not run on nodes that kwok launches
+# allow it to schedule to critical addons, but not schedule onto kwok nodes.
 cat <<EOF > "${BASE}/tolerate-all.yaml"
 apiVersion: apps/v1
 kind: Deployment
@@ -21,7 +28,9 @@ spec:
   template:
     spec:
       tolerations:
-      - operator: "Exists"
+      - operator: "Equals"
+        key: CriticalAddonsOnly 
+        effect: NoSchedule
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
@@ -31,6 +40,8 @@ spec:
                 operator: DoesNotExist
 EOF
 
+# TODO: Simplify the kustomize to only use one copy of the RBAC that all
+# controllers can use.  
 cat <<EOF > "${BASE}/kustomization.yaml"
   apiVersion: kustomize.config.k8s.io/v1beta1
   kind: Kustomization
--- a/hack/install-kwok.sh
+++ b/hack/install-kwok.sh
@@ -28,7 +28,7 @@ spec:
   template:
     spec:
       tolerations:
-      - operator: "Equals"
+      - operator: "Equal"
         key: CriticalAddonsOnly 
         effect: NoSchedule
       affinity:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -22,6 +22,11 @@ deflake: ## Run randomized, racing tests until the test fails to catch flakes
 		-v \
 		./pkg/...
 
+install-kwok: ## Install kwok into cluster
+	UNINSTALL_KWOK=false ./hack/install-kwok.sh
```
jonathan-innis, line 26: optional: I wonder if we should have this be a CLI flag as opposed to an environment variable.

hack/install-kwok.sh
```diff
@@ -0,0 +1,102 @@
+set -euo pipefail
+
+# get the latest version
+KWOK_REPO=kubernetes-sigs/kwok
+# KWOK_LATEST_RELEASE=$(curl "https://api.github.com/repos/${KWOK_REPO}/releases/latest" | jq -r '.tag_name')
```
jonathan-innis, line 5: Should this still be uncommented?

hack/install-kwok.sh
```diff
@@ -0,0 +1,102 @@
...
+  - name: registry.k8s.io/kwok/kwok
+    newTag: "${KWOK_LATEST_RELEASE}"
+  resources:
+  - "https://github.com/${KWOK_REPO}/kustomize/kwok?ref=${KWOK_LATEST_RELEASE}"
+  patchesStrategicMerge:
+  - tolerate-all.yaml
+EOF
+
+# Define 10 different kwok controllers to handle large load
+for let in partition-a partition-b partition-c partition-d partition-e partition-f partition-g partition-h partition-i partition-j
```
jonathan-innis, line 47: Should the number of partitions be configurable here or do we expect that everyone is going to need 10 containers running here that cover these different partitions?

hack/install-kwok.sh
```diff
@@ -0,0 +1,102 @@
...
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kwok-controller
+  namespace: kube-system
+spec:
+  template:
+    spec:
+      tolerations:
+      - operator: "Exists"
```
jonathan-innis, line 24: This is typically bad practice, since it means that this pod can never be evicted even if the node is experiencing issues. What was the reason we needed to tolerate everything?

hack/install-kwok.sh
```diff
@@ -0,0 +1,102 @@
+set -euo pipefail
```
jonathan-innis, line 1: Top-level script comment that describes script usage, what the script is intended to do, and script options/flags?

jonathan-innis (commented):
Nice work! I shifted some comments over from the larger PR into this one.

hack/install-kwok.sh
njtran: Kwok says each controller can reliably handle about 1k nodes and 100k pods. With 10 partitions, this makes it 10k nodes and 1 M pods, which I think is past kubernetes scalability limitation recommendations of 5k nodes. Even if the k8s doc is outdated, it's the best we've got, and I think 10 partitions would cover most scales that I imagine users are at.

- [Kwok reference](https://github.com/kubernetes-sigs/kwok/tree/main#:~:text=Currently%2C%20KWOK%20can%20reliably%20maintain%201k%20nodes%20and%20100k%20pods%20easily.)
- [Kubernetes reference](https://github.com/kubernetes/community/blob/master/sig-scalability/configs-and-limits/thresholds.md#:~:text=The%20table%20below%20is%20NOT%20exhaustive%20%2D%20more%20content%20is%20coming%20soon.)

hack/install-kwok.sh
njtran: Good callout, was doing this just for my testing, I'll fix this.

Makefile
njtran: Can go either way, what's your thought on the benefits of the CLI flag?

hack/install-kwok.sh
njtran: Nope, I'll fix this.

hack/install-kwok.sh
jonathan-innis: Yeah, it was more asking if we think that users are going to use this installation script to install kwok at a smaller scale. We are going to use this for _our_ scale testing, but I could see use-cases for installing kwok and running the kwok karpenter cloudprovider with a smaller-scale cluster. I'm fine waiting until we get that requested.

Makefile
jonathan-innis: CLI flags are a bit easier to reason about when it comes to parameters being passed to scripts. Without a CLI arg here, if I were coming at the script without the Makefile, it would be harder for me to understand what the script is capable of doing and what "special environment variable" I need to set to get Karpenter to uninstall with this script. Not blocking, though. Just a thought.

njtran: That's why I wanted to use the environment variables and have the two makefile targets of `install-kwok` and `uninstall-kwok`. This adds the environment variables for them, and the script will fail if this is unset anyway.

hack/install-kwok.sh
njtran: Can do

Makefile
jonathan-innis: Sounds fine. Only a thought, definitely not required.

hack/install-kwok.sh
njtran: Actually, on trying to use the latest version, this script doesn't work. In the spirit of time, I'll keep this commented out, and just depend on the v0.3.0 version, adjusting/upgrading this script when we need to. WDYT?

jonathan-innis: > Actually, on trying to use the latest version, this script doesn't work

Do you know why it doesn't work? I'm surprised that the installation docs that are vended by the kwok installation guide don't work correctly.

jonathan-innis (approved):
/lgtm
/approve

PR #911, opened 2023-12-25
chore: Drop year from Copyright Header

jonathan-innis:
Fixes #N/A 

**Description**

The latest guidance from the CNCF here (https://github.com/cncf/foundation/blob/main/copyright-notices.md) indicates that we no longer need the year associated with when the file was created in the header. As the guidance points out, this is tough to keep up with and has to be maintained year-over-year.


jonathan-innis:
> I imagine that we'll need to add the license to any files that get modifications in 2024, which we should also validate that the boilerplate will work properly for this

Copyright doesn't get added to existing files, should only get added to new files. Copyright exists as a "starting from" this date and gets extended out from that starting point.

jackfrancis (approved):
lgtm

jonathan-innis:
From a discussion that I had in the K8s slack, this is the position that K8s takes as well with its copyright: https://kubernetes.slack.com/archives/C09R23FHP/p1704227941184169

hack/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2023 The Kubernetes Authors.
+Copyright 2024 The Kubernetes Authors.
```
njtran, line 2: Can you add a check for this in the hack/boilerplate.sh? Maybe grep for The Kubernetes Authors?

jonathan-innis: Dropped the year entirely. Seems like the guidance indicates that the year is no longer required: https://github.com/cncf/foundation/blob/main/copyright-notices.md

njtran (commented):
/lgtm 
/approve

PR #913, opened 2023-12-26
ci: Drop `kind/support` from exempt and only use kind labels for exemption

jonathan-innis:
Fixes #N/A 

**Description**

Drops `kind/support` from the set of exempt labels to close-out stale questions and use only `kind/*` labels for staleness categorization

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -19,7 +19,7 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           stale-issue-message: 'This issue has been inactive for 14 days. StaleBot will close this stale issue after 14 more days of inactivity.'
-          exempt-issue-labels: 'bug,chore,feature,documentation,testing,operational-excellence,automation,roadmap,kind/cleanup,kind/documentation,kind/regression,kind/flake,kind/bug,kind/deprecation,kind/support,kind/feature,kind/failing-test,kind/api-change'
+          exempt-issue-labels: 'kind/cleanup,kind/testing,kind/documentation,kind/regression,kind/flake,kind/bug,kind/deprecation,kind/feature,kind/failing-test,kind/api-change,kind/automation'
           stale-issue-label: 'lifecycle/stale'
           close-issue-label: 'lifecycle/closed'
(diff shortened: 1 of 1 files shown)


njtran:
We need to make sure we're monitoring the existing issues so that we're not erroneously staling + closing out issues that should be open in the two weeks after this is merged.

jonathan-innis:
> We need to make sure we're monitoring the existing issues so that we're not erroneously staling + closing out issues that should be open in the two weeks after this is merged

You can check through all of the currently open issues. All issues currently have a `kind/*` and there are two issues that have `kind/support` where we are waiting on a response.

njtran:
There are a lot of operational-excellence issues that shouldn't be closed out, but you're right that it looks like each of these has a `kind/*` label. Can we make a new label for this - kind/operational-excellence?

jonathan-innis:
> Can we make a new label for this - kind/operational-excellence

I'm questioning the value of this label. Which issues that currently carry this label can't fit in another category that already exists within `kind/feature`, `kind/testing`, `kind/cleanup`? If we think there is a natural separation between issues that are `kind/operational-excellence` versus other kinds, I think this makes sense, but I'm having trouble coming up with examples of issues that fit in `kind/operational-excellence` and don't fit in another category as well.

njtran:
I feel like `kind/operation-excellence` will fit for issues related to repo management, release automation, security hardening, and maybe refactoring? I definitely think we should migrate all testing related things to `kind/test` though, as we used to use operational-excellence a lot in conjunction with test.

jonathan-innis:
I wonder if we should use other labels for these since it isn't a well-known upstream label. There's a little bit of prior art for creating your own custom labels in projects, but I'd like to minimize it as much as we think is reasonable to make it easier to hop between project to project in `kubernetes-sigs`. I feel like I would categorize the things that you mentioned above into the following buckets:
1. repo management: `kind/cleanup`
2. release automation: `kind/automation` (this is currently a custom label that I created for our repo)
3. security hardening: `kind/automation` or `kind/cleanup` (I'm assuming you mean this w.r.t. either CI/CD or security vulnerabilities in the code itself)
4. refactoring: `kind/cleanup`

njtran:
I can see the case for each one except the security hardening one. Happy to disagree and commit on this one, since I don't think it's a big issue if those issues get labeled like that.

njtran (commented):
/lgtm 
/approve

PR #921, opened 2024-01-03
feat: add support for ContainerLogMaxSize and ContainerLogMaxFiles

changhyuni:
**Description**
Add support for ContainerLogMaxSize and ContainerLogMaxFiles.  
We'll be adding this to the bootstrap code in the near future. https://github.com/aws/karpenter-provider-aws   
To set log rotation the current workaround with AWS Node Template is to use jq on node start to insert these args into the /etc/kubernetes/kubelet/kubelet-config.json.   

This method is hard to manage the kubelet config file as a manifest and I think we need to add the args feature

### DIFF
--- a/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
@@ -113,6 +113,14 @@ spec:
                       maximum: 100
                       minimum: 0
                       type: integer
+                    containerLogMaxSize:
+                      description: ContainerLogMaxSize is a quantity defining the maximum size of the container log file before it is rotated.
+                      type: string
+                    containerLogMaxFiles:
+                      description: ContainerLogMaxFiles specifies the maximum number of container log files that can be present for a container.
+                      format: int32
+                      minimum: 2
+                      type: integer
                     kubeReserved:
                       additionalProperties:
                         anyOf:
--- a/pkg/apis/crds/karpenter.sh_nodepools.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodepools.yaml
@@ -191,6 +191,14 @@ spec:
                               maximum: 100
                               minimum: 0
                               type: integer
+                            containerLogMaxSize:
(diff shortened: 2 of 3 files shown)


jonathan-innis:
> This method is hard to manage the kubelet config file as a manifest and I think we need to add the args feature

Can you describe what's difficult about it? We're trying to avoid managing the entire KubeletConfig file all over again and are just trying to focus on values in our `spec.kubelet` that affect the scheduling decisions that Karpenter would make.

changhyuni:
@jonathan-innis 
What I would like to say is that if the user needs to change the kubelet config, it should be set via `spec.kubelet` instead of relying on bootstrap. Like the `imageGCThresholdPercent` setting, the `containerLogMax**` setting should be set via `spec.kubelet`.  I think it's convenient and intuitive for users to manage "kubeletconfig" settings only through `spec.kubelet`.

jonathan-innis:
> Like the imageGCThresholdPercent setting, the containerLogMax** setting should be set via spec.kubelet

I agree that it is effectively easier to configure; however, it being something that is slightly easier to configure doesn't supercede the development and management burden that we have to push toward CloudProvider when we add these additional fields to the API.

We added the `imageGCThresholdPercent` before we had a more strict policy on things that we were going to accept over in `spec.kubelet`. I think our issue now is that CloudProviders have to keep-up with these updates. If we are taking in these changes piece-meal, then we're going to have an update, then each CloudProvider is going to have to rollout the logic to add that key into their userData logic as well as maintain testing to make sure that that value is actually being propagated and configured properly.

Given the number of fields in the [KubeletConfiguration](https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/), I don't think that this is really a feasible burden to push onto all Cloud Providers that support Karpenter. Unless we have a true, pressing need (like impact on scheduling), I'd like us to avoid pushing all of the configuration, updating, and test burden to all Cloud Providers.

changhyuni:
@jonathan-innis 
Thanks for the reply, I definitely understand. 
I'll close this PR
However, it will be much easier to overwrite the helm chart values if we manage it with `spec.kubelet` than using userdata as eof type.

PR #923, opened 2024-01-04
fix: add manual nodeclaim update into cluster state for new nodes

njtran:
Fixes #N/A 

**Description**
This calls `clusterState.UpdateNodeClaim` when Karpenter creates new node claims to ensure that Karpenter is aware of nodes that are being created immediately, avoiding eventual consistency delay races with the controller-runtime cache. 

make presubmit

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -359,6 +359,8 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 		metrics.ReasonLabel:   options.Reason,
 		metrics.NodePoolLabel: nodeClaim.Labels[v1beta1.NodePoolLabelKey],
 	}).Inc()
+	// Update the nodeclaim manually in state to avoid evenutal consistency delay races with our watcher
+	p.cluster.UpdateNodeClaim(nodeClaim)
 	if functional.ResolveOptions(opts...).RecordPodNomination {
 		for _, pod := range n.Pods {
 			p.recorder.Publish(scheduler.NominatePodEvent(pod, nil, nodeClaim))
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -359,7 +359,11 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 		metrics.ReasonLabel:   options.Reason,
 		metrics.NodePoolLabel: nodeClaim.Labels[v1beta1.NodePoolLabelKey],
 	}).Inc()
-	// Update the nodeclaim manually in state to avoid evenutal consistency delay races with our watcher
+	// Update the nodeclaim manually in state to avoid evenutal consistency delay races with our watcher.
+	// This is essential to avoiding races where disruption can create a replacement node, then immediately
+	// requeue. This can race with controller-runtime's internal cache as it watches events on the cluster
+	// to then trigger cluster state updates. Triggering it manually ensures that Karpenter waits for the
+	// internal cache to sync before moving onto another disruption loop.
 	p.cluster.UpdateNodeClaim(nodeClaim)
 	if functional.ResolveOptions(opts...).RecordPodNomination {
 		for _, pod := range n.Pods {
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -102,12 +102,16 @@ func (c *Cluster) Synced(ctx context.Context) bool {
 	stateNodeNames := sets.New(lo.Keys(c.nodeNameToProviderID)...)
 	c.mu.RUnlock()
 
-	nodeClaimNames := sets.New[string]()
-	for _, nodeClaim := range nodeClaimList.Items {
-		// If the nodeClaim hasn't resolved its provider id, then it hasn't resolved its status
-		if nodeClaim.Status.ProviderID == "" {
+	// Check to see if any node claim doesn't have a provider ID. If it doesn't, then the nodeclaim hasn't been
+	// launched, and we need to wait to see what the resolved values are before continuing.
+	for nc := range stateNodeClaimNames {
+		if c.nodeClaimNameToProviderID[nc] == "" {
 			return false
 		}
+	}
+
+	nodeClaimNames := sets.New[string]()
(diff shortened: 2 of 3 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -359,6 +359,8 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 		metrics.ReasonLabel:   options.Reason,
 		metrics.NodePoolLabel: nodeClaim.Labels[v1beta1.NodePoolLabelKey],
 	}).Inc()
+	// Update the nodeclaim manually in state to avoid evenutal consistency delay races with our watcher
```
jonathan-innis, line 362: I'd personally expound on this comment a bit more. I can imagine quickly forgetting the context on why we added this and this seems like a pretty critical line to ensure that we aren't forging ahead without knowing that we just created something

pkg/controllers/state/cluster.go
```diff
@@ -221,11 +221,10 @@ func (c *Cluster) UpdateNodeClaim(nodeClaim *v1beta1.NodeClaim) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	if nodeClaim.Status.ProviderID == "" {
-		return // We can't reconcile nodeclaims that don't yet have provider ids
+	if nodeClaim.Status.ProviderID != "" {
```
jonathan-innis, line 224: Add a comment here on why we are only doing this when nodeClaim status is non-empty

pkg/controllers/state/cluster.go
```diff
@@ -221,11 +221,10 @@ func (c *Cluster) UpdateNodeClaim(nodeClaim *v1beta1.NodeClaim) {
 	c.mu.Lock()
```
jonathan-innis, line 221: We're missing the update to the `Synced()` call here which validates that we don't have any `nodeClaimNametoProviderID` entries that have empty values

pkg/controllers/state/suite_test.go
```diff
@@ -1263,6 +1263,15 @@ var _ = Describe("Cluster State Sync", func() {
 		}
 		Expect(cluster.Synced(ctx)).To(BeFalse())
 	})
+	It("shouldn't consider the cluster state synced if a nodeclaim isn't launched", func() {
```
jonathan-innis, line 1266: I think this might be covered by a previous test case that syncs a bunch of nodes and then has a single node that doesn't have a provider id

jonathan-innis, line 1266: Should we have a test case that validates that when we call the UpdateNodeClaim() call directly (rather than calling the reconcile method, which would be triggered by controller-runtime), we mark the cluster state as not synced?

pkg/controllers/state/suite_test.go
```diff
@@ -1263,6 +1263,15 @@ var _ = Describe("Cluster State Sync", func() {
 		}
 		Expect(cluster.Synced(ctx)).To(BeFalse())
 	})
```
jonathan-innis, line 1265: Worthwhile to add a test case to ensure that we only store a single state Node when we call UpdateNodeClaim directly without a provider id, call reconcile with a Nodeclaim without a providerid, call reconcile with the same NodeClaim that resolved its provider id, and then delete the Nodeclaim to ensure it all cleans-up appropriately?

njtran: I don't think we can do exactly all of this, since we're only blocking Synced() on these nodeclaims. I'll add tests that ensure these nodeclaims are cleaned up on a deletion though.

pkg/controllers/state/suite_test.go
```diff
@@ -1263,6 +1263,24 @@ var _ = Describe("Cluster State Sync", func() {
...
+		Expect(cluster.Synced(ctx)).To(BeFalse())
+	})
+	It("shouldn't consider the cluster state synced if a nodeclaim without a providerID is deleted", func() {
+		nodeClaim := test.NodeClaim()
+		nodeClaim.Status.ProviderID = ""
+
+		cluster.UpdateNodeClaim(nodeClaim)
+		Expect(cluster.Synced(ctx)).To(BeFalse())
+
+		// Reconcile it without actually creating it, essentially making it deleted.
```
jonathan-innis, line 1280: I was confused at first by this statement. Might make more sense within the context of the test to just Create the thing, update it, then actually delete the thing and update

pkg/controllers/state/cluster.go
```diff
@@ -401,13 +409,13 @@ func (c *Cluster) newStateFromNodeClaim(nodeClaim *v1beta1.NodeClaim, oldNode *S
 }
 
 func (c *Cluster) cleanupNodeClaim(name string) {
+	delete(c.nodeClaimNameToProviderID, name)
```
jonathan-innis, line 412: Why change this?

jonathan-innis: Nvm, discussed offline. This is needed since we are checking `id != ""` below

pkg/controllers/state/suite_test.go
```diff
@@ -1263,6 +1263,28 @@ var _ = Describe("Cluster State Sync", func() {
...
+	It("shouldn't consider the cluster state synced if a nodeclaim without a providerID is deleted", func() {
+		nodeClaim := test.NodeClaim()
+		nodeClaim.Status.ProviderID = ""
+
+		cluster.UpdateNodeClaim(nodeClaim)
+		Expect(cluster.Synced(ctx)).To(BeFalse())
+
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		ExpectReconcileSucceeded(ctx, nodeClaimController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(cluster.Synced(ctx)).To(BeFalse())
```
jonathan-innis, line 1282:
```suggestion
		Expect(cluster.Synced(ctx)).To(BeTrue())
```

This should be true here, right?

jonathan-innis (commented):
One comment on the testing; otherwise, LGTM

pkg/controllers/state/suite_test.go
njtran: Don't think so, in this case, the node claim is still in our data structure.

jonathan-innis (approved):
Nice work! 🎉 

/lgtm
/approve

PR #925, opened 2024-01-05
fix: only include initialized towards the budget total

njtran:
Fixes #N/A 

**Description**
When Karpenter replaces nodes within disruption, Karpenter previously considered nodes that weren't deleting, which meant that nodes that were launching or registering were also counting towards the total. This means while processing another command, a NodePool's allowed disruptions could change when using percentage based budgets. We should only be using the nodes that Karpenter considers healthy when determining whether a budget threshold will be reached. 

make presubmit


jonathan-innis (approved):
LGTM 🚀

jonathan-innis (approved):
/lgtm
/approve

Issue #920, opened 2024-01-03
Minimal support for stacked/cooperative autoscaling


gilqubex:
### Description

**What problem are you trying to solve?**

(in continuation of slack conversation in #karpenter-dev on kubernetes slack)

As a 3rd party independent software vendor, I would like Karpenter to allow for cooperative autoscaling - that is, two or more autoscalers sharing responsibility for the same cluster.

While having more than one autoscaler start instances is generally benign (at most, one ends up with temporary overcapacity), more than one autoscaler trying to downscale at the same time may very well cause self-induced unavailability.

in v1alpha5, the "linking" mechanism enabled karpenter to "adopt" foreign instances and thus remain as the sole cluster downscaler. This however was incidental as this was not the original purpose of the linking mechanism, perhaps now a proper feature aimed at this goal can be designed.

In lieu of cooperative autoscaling, we have observed 3rd party vendors requiring customers to migrate off of karpenter (or cluster autoscaler for that matter), which (at least in our scenario as a specialized autoscaler) is undesirable.

as a 3rd party vendor, we plan on assuming all integration responsibility upon ourselves, but the most basic supporting mechanism (e.g. not rejecting foreign nodeclaims) should hopefully exist.

**How important is this feature to you?**

This is very important to us as we would like to continue supporting cooperative autoscaling from v1beta1 onwards.

Bryce-Soghigian:
Have you thought about natively implementing your instance launch in karpenter via an implementation of the cloud provider interface? Seems like bringing your own VMS means that effectively you take the responsiblity inside of karpenter for launching/creating instances(the responsibility of the cloud provider). 


---
Link to original slack thread: https://kubernetes.slack.com/archives/C04JW2J5J5P/p1703276514443079

gilqubex:
> Have you thought about natively implementing your instance launch in karpenter via an implementation of the cloud provider interface? Seems like bringing your own VMS means that effectively you take the responsiblity inside of karpenter for launching/creating instances(the responsibility of the cloud provider).
> 
> Link to original slack thread: https://kubernetes.slack.com/archives/C04JW2J5J5P/p1703276514443079

Before referring to the specific use case of our software, I have to say that I cannot confirm that this is a sufficient solution for all third party cooperative autoscalers. Even if we could "squeeze in", I believe this degree of freedom is very likely to be limiting for other use cases.

[...]

Issue #752, opened 2022-03-16
Add a gracePeriod for the `do-not-evict` pod annotation


stevehipwell:
**Tell us about your request**
I'd like the ability to taint a node after a given period via `ttlSecondsUntilTainted` to allow running pods to finish before the node is terminated, this should still respect the `ttlSecondsUntilExpired` and `ttlSecondsAfterEmpty`.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
When running jobs such as CI/CD pipelines in K8s there are long running jobs which shouldn't be terminated due to the high cost of re-running them, by adding `ttlSecondsUntilTainted` we can have nodes that expire and are replaced without the cost of killing potentially long running jobs.

**Are you currently working around this issue?**
Not using Karpenter.

**Additional context**
n/a

**Attachments**
n/a

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ssoriche:
I've been reading through the design document referenced above (https://github.com/aws/karpenter-core/pull/516) and read that to mean the controls for when a pod is going to be considered for eviction are placed on the NodePool. We have pods that take 20 minutes to start up and checkpoint every 5 minutes after that. The gracePeriod mentioned in this issue on the pod would allow us to block eviction of the pod for a significant time for the task to get traction before being considered for consolidation.
Having it on the NodePool level would mean that each workload that has a long start up time, and checkpoint would need its own NodePool. We experimented with separating our workloads into node pools and saw a more than 50% increase in our cluster costs immediately.

Please forgive me if this is not the right place for the above feedback, willing to move the conversation to the appropriate venue.

njtran:
> read that to mean the controls for when a pod is going to be considered for eviction are placed on the NodePool

Not necessarily, in the context of this issue, I could imagine the solution changing the value of the annotation to a `time.Duration` formatted string, which Karpenter could use with CreationTimestamp to know when it can be deleted.  Yet, for NodePool-level controls, I could see something like a `minNodeLifetime` be included, which would opt out nodes that aren't old enough for disruption.

[...]

PR #926, opened 2024-01-06
chore: add initialization buffer to disruptable nodes

njtran:
Fixes #N/A 

**Description**
Adds a 30s initialization buffer into candidacy for nodes for disruption. Only nodes that have been initialized for 30 seconds can be considered candidates for disruption, ensuring that since we're able to do multiple commands in parallel, that we don't disrupt replacement nodes for a previous action in a future action. 

make presubmit

### DIFF
--- a/pkg/controllers/disruption/types.go
+++ b/pkg/controllers/disruption/types.go
@@ -67,8 +67,8 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 	if node.MarkedForDeletion() {
 		return nil, fmt.Errorf("state node is marked for deletion")
 	}
-	// skip candidates that aren't initialized
-	if !node.Initialized() {
+	// skip candidates that haven't been initialized for a period time
+	if !node.Disruptable(clk) {
 		return nil, fmt.Errorf("state node isn't initialized")
 	}
 	// If the orchestration queue is already considering a candidate we want to disrupt, don't consider it a candidate.
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -232,7 +232,7 @@ func (in *StateNode) Disruptable(clock clock.Clock) bool {
 	// chosen for disruption as soon as they're initialized. We only need a small time frame
 	// to get one pod nomination event, as that will block candidacy further.
 	cond := in.NodeClaim.StatusConditions().GetCondition(v1beta1.Initialized)
-	return cond.IsTrue() && clock.Since(cond.LastTransitionTime.Inner.Time) < (30*time.Second)
+	return cond.IsTrue() && clock.Since(cond.LastTransitionTime.Inner.Time) >= (30*time.Second)
 }
 
 func (in *StateNode) Capacity() v1.ResourceList {
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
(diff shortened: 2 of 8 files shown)


pkg/controllers/state/statenode.go
```diff
@@ -221,6 +222,19 @@ func (in *StateNode) Initialized() bool {
...
+// Disruptable returns if a node has been initialized long enough for Karpenter to disrupt it
+func (in *StateNode) Disruptable(clock clock.Clock) bool {
+	if !in.Initialized() {
+		return false
+	}
+	// A node is disruptable only if it's been initialized for 30 seconds.
+	// This is critical to ensure that nodes that are created as part of a replacement aren't
+	// chosen for disruption as soon as they're initialized. We only need a small time frame
+	// to get one pod nomination event, as that will block candidacy further.
+	cond := in.NodeClaim.StatusConditions().GetCondition(v1beta1.Initialized)
```
jonathan-innis, line 234: You need to add this check for emptiness. Otherwise, we are going to mark NodeClaims as empty that are also within this buffer

pkg/controllers/state/statenode.go
```diff
@@ -221,6 +222,19 @@ func (in *StateNode) Initialized() bool {
 	return true
 }
 
+// Disruptable returns if a node has been initialized long enough for Karpenter to disrupt it
+func (in *StateNode) Disruptable(clock clock.Clock) bool {
```
jonathan-innis, line 226: I like the naming, but because we are going to use this with emptiness, I wonder if we have to rename this to something else like, "WithinInitializionWindow()" or something more verbose

pkg/controllers/disruption/types.go
```diff
@@ -67,8 +67,8 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 	if node.MarkedForDeletion() {
 		return nil, fmt.Errorf("state node is marked for deletion")
 	}
-	// skip candidates that aren't initialized
-	if !node.Initialized() {
+	// skip candidates that haven't been initialized for a period time
+	if !node.Disruptable(clk) {
```
Bryce-Soghigian, line 71: I like the idea of introducing Disruptable. Should it eventually be controlable beyond this static 30s value?

We have the ability to ignore nodes for disruption based in the DoNotDisrupt annotation but there is not temporal control there. 

https://github.com/kubernetes-sigs/karpenter/issues/920#issuecomment-1874946264 mentions an issue where this 3rd party autoscaler prevents handing the instances off to karpenter because it will disrupt too quickly.

I wonder if there is room in the disruption controls for something like "do not disrupt within first 5m of life" or a concept like this. 

Not saying its something we need to solve in this pr just food for thought

jonathan-innis: > I wonder if there is room in the disruption controls for something like "do not disrupt within first 5m of life"

We have this as an annotation that can get propagated onto the Node from the NodeClaim today. Seems like this might be a really nice place to do https://github.com/kubernetes-sigs/karpenter/issues/752 and push the `karpenter.sh/do-not-disrupt` annotation that can either be configured at the Pod spec level or at the NodeClaim level (which could be passed down from NodePool `spec.template.metadata.annotations` so you could do something like):

```yaml
spec:
  template:
     metadata:
       annotations:
          karpenter.sh/do-not-disrupt: 10m
```

> Should it eventually be controlable beyond this static 30s value

I think the annotation handling above gives us a nice path to this; however, the Node `karpenter.sh/do-not-disrupt` min lifetime differs a bit from the InitializationBuffer -- both in implementation and also in what it should be checking against. The `karpenter.sh/do-not-disrupt` with a time should be comparing the current time to the creation timestamp of that "thing" (whether that be a Pod or a NodeClaim) and then checking if the time since its creation timestamp is within the grace period associated with the `karpenter.sh/do-not-disrupt` annotation. In the case of the initialization buffer, I think we always need this to be a static value based off of the initialization time. 

The original intention of the initialization operation was to know when Karpenter could start to deprovision nodes and nodes could be considered "healthy" by the system. One issue here is that after a node becomes "healthy" to schedule nodes, pods may not schedule to it immediately (scheduler needs some time to see the pods, new node, and bind the pods as well as our watch on our nodes and the pods needs to see the binding). If someone has their `consolidationPolicy` set to `WhenEmpty` and they have their `consolidateAfter` value set to `0s`, now we have a race. We'll try to terminate new nodes that we create (which are valid for pods) a lot since we see that that node that is initialized doesn't have any pods scheduled to it (but might 2-5s later). 

Our current solve for that was to use scheduler nomination to know that pods were going to land on nodes. This worked okay, but 1) It's more complicated than a flat amount of time and 2) We didn't extend that scheduler nomination when we were launching nodes for disruption. This means that any nodes that we were launching as replacements wouldn't know that they were having pods scheduled against them, which could be a problem for the reasons stated above around race conditions.

Given all this, it seems we'll always need a flat buffer on top of our initialization time to ensuring that we aren't racing against the kube-scheduler

pkg/controllers/state/statenode.go
njtran: This is an interesting point. On one hand, we're expecting nodes to get pods scheduled onto them soon, which could result in a redundant double patch (mark as empty, then mark as not empty). On the other hand, the 30s is a heuristic on how long we're expecting this to be _good enough_ for pod scheduling. Since we're using the initialization buffer to hold on disrupting, I'm not sure I agree with holding on identifying status conditions of individual node claims. It may be preferable to have the emptiness condition progress while the buffer is going.

pkg/controllers/state/statenode.go
njtran: i can see the case for this if we have more Disruptable requirements. Not sure I understand why reasoning about emptiness motivates you to think of another name.

pkg/controllers/disruption/types.go
njtran: Well said!

njtran:
closing after discussion offline with @jonathan-innis to go another way to solve the core problem

PR #930, opened 2024-01-09
fix: Fix data race in cluster state

jonathan-innis:
Fixes #N/A 

**Description**

This solves a data race problem that we had in our cluster state `Synced()` call when we were access map data without holding the mutex

`make test` (which calls `go test` with the `-race` parameter)


njtran (approved):
/lgtm 
/approve
nice catch!

PR #933, opened 2024-01-10
fix: emit pod nomination events and add provisioning node tainted trigger controller

njtran:
Fixes #N/A 

**Description**
This is a breaking change for users that rely on messages emitted by the `provisioning.trigger` controller in their logging. This adds another reconciler so that the provisioning trigger consists of a `provisioning.nodetrigger` which will trigger the batching mechanism for any tainted nodes, and `provisioning.podtrigger` which is the same as the `provisioning.trigger`.

This fixes an issue where successive disruption controller reconciliations wouldn't consider the results of the previous (and currently active) disruption command. This could result in Karpenter disrupting a node that was recently the target of a scheduling loop, when Karpenter should really wait and consider another node for disruption. 

make presubmit
kwok cloud provider

### DIFF
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -215,14 +215,14 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
 //     a. There are at least 15 cheapest instance type replacement options to consolidate.
 //     b. The current candidate is NOT part of the first 15 cheapest instance types inorder to avoid repeated consolidation.
 func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, candidates []*Candidate, results *pscheduling.Results,
-	candidatePrice float64) (Command, error) {
+	candidatePrice float64) (Command, sets.Set[*pscheduling.ExistingNode], error) {
 
 	// Spot consolidation is turned off.
 	if !options.FromContext(ctx).FeatureGates.SpotToSpotConsolidation {
 		if len(candidates) == 1 {
 			c.recorder.Publish(disruptionevents.Unconsolidatable(candidates[0].Node, candidates[0].NodeClaim, "SpotToSpotConsolidation is disabled, can't replace a spot node with a spot node")...)
 		}
-		return Command{}, nil
+		return Command{}, nil, nil
 	}
 
 	// Since we are sure that the replacement nodeclaim considered for the spot candidates are spot, we will enforce it through the requirements.
@@ -239,8 +239,13 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 			c.recorder.Publish(disruptionevents.Unconsolidatable(candidates[0].Node, candidates[0].NodeClaim, "Can't replace spot node with a cheaper spot node")...)
 		}
 		// no instance types remain after filtering by price
-		return Command{}, nil
+		return Command{}, nil, nil
 	}
+
+	// Only return the existing nodes that had pods scheduled to it
+	nominatedNodes := sets.New[*pscheduling.ExistingNode](lo.Filter(results.ExistingNodes, func(n *pscheduling.ExistingNode, _ int) bool {
+		return len(n.Pods) > 0
+	})...)
 
 	// For multi-node consolidation:
 	// We don't have any requirement to check the remaining instance type flexibility, so exit early in this case.
@@ -248,7 +253,7 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 		return Command{
 			candidates:   candidates,
 			replacements: results.NewNodeClaims,
-		}, nil
+		}, nominatedNodes, nil
 	}
 
 	// For single-node consolidation:
@@ -258,7 +263,7 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 	if len(results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions) < MinInstanceTypesForSpotToSpotConsolidation {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(candidates[0].Node, candidates[0].NodeClaim, fmt.Sprintf("SpotToSpotConsolidation requires %d cheaper instance type options than the current candidate to consolidate, got %d",
 			MinInstanceTypesForSpotToSpotConsolidation, len(results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions)))...)
-		return Command{}, nil
+		return Command{}, nil, nil
 	}
 
 	// Restrict the InstanceTypeOptions for launch to 15 so we don't get into a continual consolidation situation.
@@ -273,7 +278,7 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 	return Command{
 		candidates:   candidates,
 		replacements: results.NewNodeClaims,
-	}, nil
+	}, nominatedNodes, nil
 }
 
 // getCandidatePrices returns the sum of the prices of the given candidates
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -116,7 +115,8 @@ func (c *consolidation) sortCandidates(candidates []*Candidate) []*Candidate {
 // computeConsolidation computes a consolidation action to take
 //
 // nolint:gocyclo
-func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...*Candidate) (Command, sets.Set[*pscheduling.ExistingNode], error) {
+func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...*Candidate) (Command, *pscheduling.Results, error) {
+	var err error
 	// Run scheduling simulation to compute consolidation option
 	results, err := SimulateScheduling(ctx, c.kubeClient, c.cluster, c.provisioner, candidates...)
 	if err != nil {
@@ -135,16 +135,12 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
 		}
 		return Command{}, nil, nil
 	}
-	// Only return the existing nodes that had pods scheduled to it
-	nominatedNodes := sets.New[*pscheduling.ExistingNode](lo.Filter(results.ExistingNodes, func(n *pscheduling.ExistingNode, _ int) bool {
-		return len(n.Pods) > 0
-	})...)
 
 	// were we able to schedule all the pods on the inflight candidates?
 	if len(results.NewNodeClaims) == 0 {
 		return Command{
 			candidates: candidates,
-		}, nominatedNodes, nil
+		}, results, nil
 	}
 
 	// we're not going to turn a single node into multiple candidates
@@ -206,7 +202,7 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
 	return Command{
 		candidates:   candidates,
 		replacements: results.NewNodeClaims,
-	}, nominatedNodes, nil
+	}, results, nil
 }
 
 // Compute command to execute spot-to-spot consolidation if:
@@ -215,7 +211,7 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
 //     a. There are at least 15 cheapest instance type replacement options to consolidate.
 //     b. The current candidate is NOT part of the first 15 cheapest instance types inorder to avoid repeated consolidation.
 func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, candidates []*Candidate, results *pscheduling.Results,
-	candidatePrice float64) (Command, sets.Set[*pscheduling.ExistingNode], error) {
+	candidatePrice float64) (Command, *pscheduling.Results, error) {
 
 	// Spot consolidation is turned off.
 	if !options.FromContext(ctx).FeatureGates.SpotToSpotConsolidation {
@@ -242,18 +238,13 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 		return Command{}, nil, nil
 	}
 
-	// Only return the existing nodes that had pods scheduled to it
-	nominatedNodes := sets.New[*pscheduling.ExistingNode](lo.Filter(results.ExistingNodes, func(n *pscheduling.ExistingNode, _ int) bool {
-		return len(n.Pods) > 0
-	})...)
-
 	// For multi-node consolidation:
 	// We don't have any requirement to check the remaining instance type flexibility, so exit early in this case.
 	if len(candidates) > 1 {
 		return Command{
 			candidates:   candidates,
 			replacements: results.NewNodeClaims,
-		}, nominatedNodes, nil
+		}, results, nil
 	}
 
 	// For single-node consolidation:
@@ -278,7 +269,7 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 	return Command{
 		candidates:   candidates,
 		replacements: results.NewNodeClaims,
-	}, nominatedNodes, nil
+	}, results, nil
 }
 
 // getCandidatePrices returns the sum of the prices of the given candidates
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -115,15 +116,15 @@ func (c *consolidation) sortCandidates(candidates []*Candidate) []*Candidate {
 // computeConsolidation computes a consolidation action to take
 //
 // nolint:gocyclo
-func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...*Candidate) (Command, error) {
+func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...*Candidate) (Command, sets.Set[*pscheduling.ExistingNode], error) {
 	// Run scheduling simulation to compute consolidation option
 	results, err := SimulateScheduling(ctx, c.kubeClient, c.cluster, c.provisioner, candidates...)
 	if err != nil {
 		// if a candidate node is now deleting, just retry
 		if errors.Is(err, errCandidateDeleting) {
-			return Command{}, nil
+			return Command{}, nil, nil
 		}
-		return Command{}, err
+		return Command{}, nil, err
 	}
 
 	// if not all of the pods were scheduled, we can't do anything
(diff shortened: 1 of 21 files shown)


pkg/controllers/disruption/emptynodeconsolidation.go
```diff
@@ -102,11 +103,13 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
...
 		if len(n.pods) != 0 || c.cluster.IsNodeNominated(n.ProviderID()) || postValidationMapping[n.nodePool.Name] == 0 {
 			logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd)
-			return Command{}, nil
+			return Command{}, nil, nil
 		}
 		postValidationMapping[n.nodePool.Name]--
 	}
-	return cmd, nil
+	// Return empty scheduling results since no empty nodes should be rescheduling any pods.
+	// Returning nil would require doing a check later in the core controller.
```
jonathan-innis, line 111: Having to write this here implies that you now have dependencies on the caller inside of your abstraction. I was trying to think if `scheduling.Results` should be `nil` or not and if we should just handle it gracefully outside....I think it may make sense.

In either case, having a dependency like this between your return arguments seems wrong to me.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -76,43 +76,44 @@ func (m *MultiNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
...
-		return Command{}, nil
+		return Command{}, nil, nil
 	}
 	min := 1
 	if len(candidates) <= max {
 		max = len(candidates) - 1
 	}
 
 	lastSavedCommand := Command{}
+	results := &scheduling.Results{}
```
jonathan-innis, line 116:
```suggestion
	lastSavedResults := &scheduling.Results{}
```

Should we call this `lastSavedResults` for parity and then we can just call it `results` in the binary search?

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,105 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		nodePool.Spec.Disruption.Budgets = []v1beta1.Budget{{
+			Nodes: "100%",
+		}}
+
+		// create our RS so we can link a pod to it
+		rs := test.ReplicaSet()
+		ExpectApplied(ctx, env.Client, rs)
+		Expect(env.Client.Get(ctx, client.ObjectKeyFromObject(rs), rs)).To(Succeed())
+
+		// Create a two pods that can all fit on one node.
```
jonathan-innis, line 485:
```suggestion
		// Create two pods that can all fit on one node.
```

nit

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,105 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+				},
+			},
+		})
+		nodeClaims[0].StatusConditions().MarkTrue(v1beta1.Drifted)
+		nodeClaims[1].StatusConditions().MarkTrue(v1beta1.Drifted)
+
+		ExpectApplied(ctx, env.Client, nodePool, pods[0], nodes[0], nodeClaims[0])
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*v1.Node{nodes[0]}, []*v1beta1.NodeClaim{nodeClaims[0]})
+
+		// Expect the first pod scheduled, and then cordon the node so that the other pod can schedule
```
jonathan-innis, line 530: Why do you need to cordon at all if consolidation should just take care of combining the nodes -- either A onto B or B onto A

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,105 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+		ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
+
+		var wg sync.WaitGroup
+		ExpectTriggerVerifyAction(&wg)
+		ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
+		wg.Wait()
+
+		nodes = ExpectNodes(ctx, env.Client)
+		Expect(len(nodes)).To(Equal(2))
+		tainted := lo.Filter(nodes, func(n *v1.Node, _ int) bool {
```
jonathan-innis, line 555: Should we have an `ExpectTainted` function like we do in our E2Es on the AWS provider?

pkg/controllers/disruption/suite_test.go
```diff
@@ -467,6 +467,105 @@ var _ = Describe("Pod Eviction Cost", func() {
...
+			return lo.Contains(n.Spec.Taints, v1beta1.DisruptionNoScheduleTaint)
+		})
+		Expect(len(tainted)).To(Equal(1))
+		nominatedNode, found := lo.Find(nodes, func(n *v1.Node) bool {
+			return n.Name != tainted[0].Name
+		})
+		Expect(found).To(BeTrue())
+		stateNode := ExpectStateNodeExists(cluster, nominatedNode)
+
+		Expect(stateNode.Nominated()).To(BeTrue())
```
jonathan-innis, line 565: Nice 🎉

pkg/controllers/disruption/types.go
```diff
@@ -37,11 +38,15 @@ import (
...
 	ShouldDisrupt(context.Context, *Candidate) bool
-	ComputeCommand(context.Context, map[string]int, ...*Candidate) (Command, error)
+	ComputeCommand(context.Context, map[string]int, ...*Candidate) (Command, *scheduling.Results, error)
 	Type() string
 	ConsolidationType() string
 }
 
+// NominatedNodes is a set of node/nodeclaim provider IDs that were nominated for a scheduling simulation.
+// This is used to emit nomination events when we finish executing a command.
+type NominatedNodes sets.Set[*scheduling.ExistingNode]
```
jonathan-innis, line 48: I didn't catch it. Where is this used?

pkg/controllers/provisioning/controller.go
```diff
@@ -71,6 +73,19 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 	return operatorcontroller.Adapt(controllerruntime.
 		NewControllerManagedBy(m).
 		For(&v1.Pod{}).
+		// Only enqueue pods that have the node name not set. We do this here so that we can enqueue the pod
+		// requests from nodes that are deleting
+		WithEventFilter(predicate.NewPredicateFuncs(func(o client.Object) bool {
```
jonathan-innis, line 78: Should we have changed L61? I don't see how this triggers reconciliation for tainted nodes now?

pkg/test/pods.go
```diff
@@ -143,6 +145,25 @@ func Pod(overrides ...PodOptions) *v1.Pod {
 			Phase:      options.Phase,
 		},
 	}
+	// If PreStopSleep is enabled, add it to each of the containers.
+	// Can't use v1.LifecycleHandler == v1.SleepAction as that's a feature gate in Alpha 1.29.
+	// https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-implementations
+	if options.PreStopSleep != nil {
```
jonathan-innis, line 151: I'm assuming that this is put here for the E2E test suite? I didn't see a functional reason why this would need to be part of the functional testing suite

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -74,6 +75,26 @@ func NodeEventHandler(c client.Client) handler.EventHandler {
 	})
 }
 
+func NodeEventHandlerToPods(c client.Client) handler.EventHandler {
```
jonathan-innis, line 78: High-level question on the triggering mechanism: Do we need to re-trigger on every pod or just once on the node getting tainted? Does it make sense to put this batch trigger in this existing controller or is this maybe a node controller triggering mechanism?

pkg/controllers/disruption/consolidation.go
```diff
@@ -138,29 +139,37 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
...
 	}
 
 	// get the current node price based on the offering
 	// fallback if we can't find the specific zonal pricing data
 	candidatePrice, err := getCandidatePrices(candidates)
 	if err != nil {
-		return Command{}, fmt.Errorf("getting offering price from candidate node, %w", err)
+		return Command{}, nil, fmt.Errorf("getting offering price from candidate node, %w", err)
+	}
+	results.NewNodeClaims[0].InstanceTypeOptions = filterByPrice(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
```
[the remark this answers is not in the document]
njtran: Yep, thanks for catching

pkg/controllers/disruption/drift.go
```diff
@@ -72,10 +73,12 @@ func (d *Drift) filterAndSortCandidates(ctx context.Context, candidates []*Candi
 // ComputeCommand generates a disruption command given candidates
 //
 //nolint:gocyclo
-func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[string]int, candidates ...*Candidate) (Command, error) {
-	candidates, err := d.filterAndSortCandidates(ctx, candidates)
+func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[string]int, candidates ...*Candidate) (Command, *scheduling.Results, error) {
+	var err error
+	results := &scheduling.Results{}
```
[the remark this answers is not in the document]
njtran: we do a check in controller.go
```
for _, node := range schedulingResults.ExistingNodes {
```

if we return nil here, this panics. We can easily gate this by an if statement, but I think defining it here and enforcing that we never return nil is better than the if statement. WDYT?

pkg/controllers/disruption/multinodeconsolidation.go
njtran: Good callout!

pkg/controllers/disruption/types.go
njtran: Artifact of past revision!

pkg/test/pods.go
njtran: Yeah, that's right

pkg/controllers/provisioning/controller.go
njtran: Yes good catch!

pkg/controllers/disruption/suite_test.go
njtran: You bring up a good point

pkg/utils/pod/scheduling.go
```diff
@@ -25,6 +25,13 @@ import (
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
+func CanTriggerProvisioning(pod *v1.Pod) bool {
+	return !IsPreempting(pod) &&
+		FailedToSchedule(pod) &&
```
jonathan-innis, line 30: Doesn't failed to schedule imply that the pod was marked as pending? Are you going to catch the pods that you want here? Also, do you still need this change?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -181,7 +181,7 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*v1.Pod) *Results {
 			delete(errors, k)
 		}
 	}
-	return &Results{
+	return Results{
```
jonathan-innis, line 184: Do we have benchmarking on this change or a heap profile? With results no longer being a pointer, this means that every time that it's passed around the inner values of the struct are copied. That should be fine as long as the things that are being stored in results are maps, arrays, and pointers

pkg/controllers/provisioning/controller.go
```diff
@@ -20,42 +20,44 @@ import (
...
 		kubeClient:  kubeClient,
 		provisioner: provisioner,
 		recorder:    recorder,
 	})
 }
 
-func (c *Controller) Name() string {
-	return "provisioner.trigger"
+func (*PodController) Name() string {
+	return "provisioner.podtrigger"
```
jonathan-innis, line 56: This is a breaking change for anyone who has logging around this. Something to think about scoping into v1?

pkg/controllers/disruption/suite_test.go
```diff
@@ -512,6 +595,18 @@ func ExpectTriggerVerifyAction(wg *sync.WaitGroup) {
 	}()
 }
 
+// ExpectTaintedNodes will assert the number of nodes and tainted nodes in the cluster and return the tainted nodes.
+func ExpectTaintedNodes(ctx context.Context, c client.Client, numNodes, numTaintedNodes int) ([]*v1.Node, []*v1.Node) {
```
jonathan-innis, line 599: Is it worth just having these as separate functions? One function that asserts the number of tainted nodes, another that asserts the number of nodes. Come to think of it, I think that we have a function that asserts the number of nodes already

pkg/controllers/disruption/emptynodeconsolidation.go
```diff
@@ -102,11 +103,12 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
 	for _, n := range candidatesToDelete {
 		if len(n.pods) != 0 || c.cluster.IsNodeNominated(n.ProviderID()) || postValidationMapping[n.nodePool.Name] == 0 {
 			logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd)
-			return Command{}, nil
+			return Command{}, scheduling.Results{}, nil
 		}
 		postValidationMapping[n.nodePool.Name]--
 	}
-	return cmd, nil
+	// Return empty scheduling results since no empty nodes should be rescheduling any pods.
```
jonathan-innis, line 110: nit: do we need this comment now? Seems pretty self-explanatory to me

pkg/controllers/disruption/emptiness.go
```diff
@@ -85,9 +86,10 @@ func (e *Emptiness) ComputeCommand(_ context.Context, disruptionBudgetMapping ma
 		}
 	}
 
+	// Return empty scheduling results since no empty nodes should be rescheduling any pods.
```
jonathan-innis, line 89: Same callout here

pkg/controllers/disruption/drift.go
```diff
@@ -98,10 +99,11 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
 		}
 	}
 	// Disrupt all empty drifted candidates, as they require no scheduling simulations.
+	// Return empty scheduling results since no empty nodes should be rescheduling any pods.
```
jonathan-innis, line 102: same comment here

pkg/utils/pod/scheduling.go
njtran: Sorry I forgot to press save!

pkg/controllers/provisioning/scheduling/scheduler.go
njtran: I'll definitely be doing benchmarking before we make this change. But yes, all the inner values are maps and arrays, so we should be fine here. It's going from 1 pointer to 3 pointers. 

```
// Results contains the results of the scheduling operation
type Results struct {
	NewNodeClaims []*NodeClaim
	ExistingNodes []*ExistingNode
	PodErrors     map[*v1.Pod]error
}
```

pkg/controllers/disruption/suite_test.go
njtran: I could go either way. I added it in here because the previous revision I believe you asked if it could make more sense as an ExpectTaintedNodes() function. I can pull out the nodes though

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: Sounds good to me. Interestingly enough: Slices are struct values under the hood that point to a combination of a pointer and a length. Not that that matters in this context for memory utilization, just thought it was interesting that it's slightly different from what I expected: https://go.dev/blog/slices

pkg/controllers/disruption/suite_test.go
```diff
@@ -512,6 +597,15 @@ func ExpectTriggerVerifyAction(wg *sync.WaitGroup) {
 	}()
 }
 
+// ExpectTaintedNodes will assert the number of nodes and tainted nodes in the cluster and return the tainted nodes.
+func ExpectTaintedNodes(ctx context.Context, c client.Client, numTainted int) []*v1.Node {
+	tainted := lo.Filter(ExpectNodes(ctx, c), func(n *v1.Node, _ int) bool {
```
jonathan-innis, line 602: nit: `GinkgoHelper()`?

pkg/controllers/provisioning/controller.go
njtran: Added a note in the description

pkg/utils/nodeclaim/nodeclaim.go
jonathan-innis, line 78: Is this being used? I don't see usage of it in the PR?

pkg/controllers/provisioning/controller.go
```diff
@@ -20,42 +20,44 @@ import (
...
 
 // Controller for the resource
-type Controller struct {
+type PodController struct {
 	kubeClient  client.Client
 	provisioner *Provisioner
 	recorder    events.Recorder
 }
 
 // NewController constructs a controller instance
```
jonathan-innis, line 46: nit: Comments need to be changed to match function name

pkg/controllers/provisioning/controller.go
```diff
@@ -20,42 +20,44 @@ import (
...
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 	"sigs.k8s.io/karpenter/pkg/events"
 	operatorcontroller "sigs.k8s.io/karpenter/pkg/operator/controller"
 	"sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
-var _ operatorcontroller.TypedController[*v1.Pod] = (*Controller)(nil)
+var _ operatorcontroller.TypedController[*v1.Pod] = (*PodController)(nil)
 
 // Controller for the resource
```
jonathan-innis, line 39: nit: Comments need to be changed to match struct name

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -124,14 +125,14 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
 				logging.FromContext(ctx).Debugf("stopping multi-node consolidation after timeout, returning last valid command %s", lastSavedCommand)
 			}
-			return lastSavedCommand, nil
+			return lastSavedCommand, lastSavedResults, nil
 		}
 		mid := (min + max) / 2
 		candidatesToConsolidate := candidates[0 : mid+1]
 
-		cmd, err := m.computeConsolidation(ctx, candidatesToConsolidate...)
+		cmd, newResults, err := m.computeConsolidation(ctx, candidatesToConsolidate...)
```
jonathan-innis, line 133:
```suggestion
		cmd, results, err := m.computeConsolidation(ctx, candidatesToConsolidate...)
```

nit: `results`?

pkg/controllers/disruption/helpers.go
```diff
@@ -85,18 +85,18 @@ func simulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 	if _, ok := lo.Find(deletingNodes, func(n *state.StateNode) bool {
 		return candidateNames.Has(n.Name())
 	}); ok {
-		return nil, errCandidateDeleting
+		return pscheduling.Results{}, errCandidateDeleting
```
jonathan-innis, line 88: nit: I think it's kind of awkward that we call this package `pscheduling` which is supposed to mean "provisioner scheduling" but we are using it in the context of disruption. I know that we do this elsewhere but I wonder if we just call this the "scheduling" package and the other thing is scheduling utils or something similar

pkg/controllers/provisioning/controller.go
jonathan-innis, line 56: musing: Is this `provisioner.podtrigger` or `provisioner.trigger.pod` with the directory structure modeling that too?

pkg/controllers/disruption/suite_test.go
```diff
@@ -512,6 +597,16 @@ func ExpectTriggerVerifyAction(wg *sync.WaitGroup) {
 	}()
 }
 
+// ExpectTaintedNodes will assert the number of nodes and tainted nodes in the cluster and return the tainted nodes.
+func ExpectTaintedNodes(ctx context.Context, c client.Client, numTainted int) []*v1.Node {
```
jonathan-innis, line 601:
```suggestion
func ExpectTaintedNodeCount(ctx context.Context, c client.Client, numTainted int) []*v1.Node {
```

nit

pkg/controllers/disruption/helpers.go
njtran: I'd rather just leave as is, since we follow the same naming scheme elsewhere. Could consider moving it, but also I like how the other reads `scheduling.Requirements` whereas this one is just `pscheduling.Results`, which I think is fine.

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

PR #934, opened 2024-01-11
perf: order pods by deleting first

njtran:
Fixes #N/A 

**Description**
This orders pods for disruption scheduling simulations so that we order pods that are on deleting nodes first. This ensures that our bin-packing will put pods from previous disruption actions on the same nodes that they've been scheduled to. This allows future disruption actions to rely on purely a new node for a replacement action.

This actually speeds up disruption performance depending on the flexibility of the budgets. With a tightly packed cluster with all nodes constantly expired, before this change, we could only expire one node per 30s (as we wait for the replacement to be initialized before being able to do another disruption action). After this, I was able to see at least 5 nodes being disrupted at a time with a budget of 100%. 

Testing here

### DIFF
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -23,7 +23,9 @@ import (
 	"sort"
 	"time"
 
+	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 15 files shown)


njtran:
Closing this as we don't need this at the moment.

PR #940, opened 2024-01-16
fix: report zero resources in empty nodepools

### LINKED ISSUE #935: Empty nodepools do not report usage metrics or have a status
### Description

**Observed Behavior**:
I have a nodepool responsible for autoscaling nodes to run tasks. When no tasks are needing to be run, the nodepool scales to 0 nodes. While there are 0 nodes in the pool, the nodepool does not have a status and does not report the `karpenter_nodepool_usage` metric. This causes some issues with monitoring, as alerts setup to monitor usage notify on no data whenever the nodepool scales to 0 nodes.

**Expected Behavior**:
While there are 0 nodes in the pool, I would expect the nodepool to show a status as follows:
```json
{
    "resources": {
      "cpu": "0",
      "ephemeral-storage": "0Ki",
      "memory": "0Ki",
      "pods": "0"
    }
}
```

I would also expect it to report the `karpenter_nodepool_usage` metric with a 0 for the relevant resource types.

1. Create a nodepool with disruption/consolidation enabled and a taint to only allow specific workloads to run on these nodes:
```yaml
kind: NodePool
metadata:
  name: tasks-nodepool
spec:
  template:
    spec:
      requirements:
          operator: In
          values:
            - c
            - m
            - r
            - t
          operator: In
          values:
            - "4"
            - "8"
            - "16"
        - key: kubernetes.io/arch
          operator: In
          values:
            - amd64
          operator: In
          values:
            - spot
            - on-demand
      nodeClassRef:
        name: default
      taints:
        - effect: NoSchedule
          key: workload
          value: tasks
  limits:
    cpu: 1000
  disruption:
    consolidationPolicy: WhenUnderutilized
    expireAfter: 720h
```
2. Ensure no nodes are created in this nodepool:
```sh
> kubectl get nodes -l karpenter.sh/nodepool=tasks-nodepool
No resources found
```
3. Check to see the status of the nodepool, and ensure an empty object is returned.
```sh
> kubectl get nodepool/tasks-nodepool -o json | jq .status
{}
```
4. Check the metrics endpoint (`karpenter.karpenter.svc.cluster.local:8000/metrics`) and ensure the `karpenter_nodepool_usage` gauge metric does not include data for the `tasks-nodepool` nodepool. 

**Versions**:
- Chart Version: 0.32.4
- Kubernetes Version (`kubectl version`): v1.28.4-eks-8cb36c9

mattsre:
Fixes #935 

**Description**
Create a default list of zeroed resources before merging resources from actual nodes into the zeroed list. This allows a nodepool with 0 nodes to still report metrics on usage

Unit tests were updated to follow the new behavior. A unit test which previously ensured the resources status was nil when no nodes were in the pool was updated to ensure the resources status reported zeroed resources.


pkg/utils/resources/resources.go
```diff
@@ -125,6 +125,18 @@ func MaxResources(resources ...v1.ResourceList) v1.ResourceList {
 	return resourceList
 }
 
+// ZeroResources returns a default resource list with zeroed resources
+func ZeroResources() v1.ResourceList {
```
ellistarn, line 129: This is definitely better, but still misses possible resources that could exist (e.g. gpus). I wonder if we could analyze/cache the results of GetInstanceTypes() to hydrate the full set of values available in the cloud provider.

jonathan-innis:
Given the fact that we have to discover these resources across instance types to get this right, I'm a little worried that we are adding a lot of additional overhead to just set this value to 0. I wonder if there are ways to just get the alerting piece of this right without leaking a bunch of detail into this `ZeroResources` function. 

I know that the original concern here was that alerting couldn't be configured on usage correctly. Are there ways to handle the nil value in monitoring and treat it as `0`?

ellistarn:
> I'm a little worried that we are adding a lot of additional overhead to just set this value to 0

Isn't this a (relatively) free cache read in the counter controller?

jonathan-innis:
> Isn't this a (relatively) free cache read in the counter controller

I'm less concerned with the computational overhead of doing this and more concerned about the trade-off here between the logic that we write and maintain and the value that we get from doing it. TLDR: I'm not opposed to doing this. I just want to check that we don't have alternatives here given this is not as straight-forward as I first thought.

sftim:
Reporting the time series even when a node pool is empty is _much_ easier to teach and to document.

pkg/utils/resources/resources.go
mattsre: Is `GetInstanceTypes` filtered by requirements set in the nodepool? If so, could be a good way to track real values of resources that a nodepool _might_ provision.

sftim:
If we build something that lets you, as a cluster operator, make a recording rule that gets you what you want, we can document that. (IMO) we don't have to make the metrics perfect so long as people can get a useful overall outcome.

mattsre:
Want to follow up on this, make sure it doesn't end up dead. I think the approach of checking the cloud provider to determine the resource types a node pool can provision is probably the best to cover as many scenarios as possible, but it does carry some overhead on both Karpenter's side and on the cloud provider (they have to give us the resource types somehow). Alternatively, we could merge this as is and leave other resources out. This implementation probably covers a large % of users without any complexity.

I'd think at a bare minimum we ship this current change since it's an objective improvement over the existing behavior, given I'm biased since I reported the original issue 😅

pkg/utils/resources/resources.go
jonathan-innis: It's not filtered by requirements. The reason we pass NodePool through is that the NodeClassReference and KubeletConfig in the NodePool _might_ impact available memory/cpu. 

As for allowing the controller to make a call to `GetInstanceTypes()`, I think it's fine after reviewing the discussion. The cloudprovider generally caches this so it shouldn't have a tangible impact on TPS, but we should probably cache the resources that we get here from a call for something like 12h and then re-poll when that cache expires

ellistarn: Why do we need another cache?

jonathan-innis: Seems like you can't know what the Cloud Provider is doing when you call GetInstanceTypes() so it's probably best practice to reduce your calls to that interface if you can.

mattsre: I'll take a look this week at updating the PR to use info from `GetInstanceTypes()` so we can support a wider variety of resources

ellistarn: Curious Jonathan, why isn't the caching already in the vendor neutral impl?

jonathan-innis: > why isn't the caching already in the vendor neutral impl

I think the real answer is that there hasn't been a proven need for it yet but practically you have to push a lot of the staling mechanisms that we have in the AWS cloudprovider side into the neutral caching mechanism. It's definitely something to explore, I don't disagree. There's definitely also a path to explore here around reconciling the instance types into state and referring to that rather than doing an imperative-based cache. cc: @engedaam

njtran (commented):
All LGTM here. Any more considerations @jonathan-innis ?

mattsre:
I spent some time over the past two weeks trying to modify this PR to use data from `GetInstanceTypes()` but I don't think I can use it without making larger changes to the code base than I'm comfortable with at this point.

I'm not sure when nodepools filter instance types based on their requirements, but is it possible to store a list of filtered instance types on the nodepool? From there, the metrics code could determine what resources might be provisioned by each nodepool. This filtered list could be re-used when evaluating which instance types to provision new pods on.

jonathan-innis:
> use it without making larger changes to the code base than I'm comfortable with at this point

We can also decouple these changes if you feel comfortable with that @mattsre. Happy to take-in an expanded set of resources that are coming from `GetInstanceTypes()` as a separate PR

PR #953, opened 2024-01-23
feat: add metric on pod evicted during node termination

chenfeilee:
Description

Emit metric on number of pods evicted during node termination

How was this change tested?

`make verify`

### DIFF
--- a/pkg/controllers/node/termination/terminator/metrics.go
+++ b/pkg/controllers/node/termination/terminator/metrics.go
@@ -14,28 +14,25 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package termination
+package terminator
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
 	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
-
-	"sigs.k8s.io/karpenter/pkg/metrics"
 )
 
 var (
-	TerminationSummary = prometheus.NewSummaryVec(
-		prometheus.SummaryOpts{
-			Namespace:  "karpenter",
-			Subsystem:  "nodes",
-			Name:       "termination_time_seconds",
-			Help:       "The time taken between a node's deletion request and the removal of its finalizer",
-			Objectives: metrics.SummaryObjectives(),
+	podEvictedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "karpenter",
+			Subsystem: "pods",
+			Name:      "evicted",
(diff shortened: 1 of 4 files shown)


jonathan-innis (commented):
Nice! This is often work. Only a couple of comments. Love that you already knocked out the tests for the metrics!

jonathan-innis:
cc: @njtran To look at these changes as well.

pkg/controllers/node/termination/terminator/metrics.go
```diff
@@ -0,0 +1,38 @@
...
+
+var (
+	PodEvictedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: "karpenter",
+			Subsystem: "pods",
+			Name:      "evicted",
+			Help:      "Number of pods evicted by Karpenter during node termination",
+		},
+		[]string{},
```
jonathan-innis, line 32: Should we consider setting NodePool as a label here? We do everywhere else and I think it generally makes sense to want to correlate the number of evictions that came from a given NodePool

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -89,6 +89,11 @@ var _ = Describe("Eviction/Queue", func() {
 		})
 	})
 
+	AfterEach(func() {
+		// Reset the metrics collectors
+		terminator.PodEvictedCounter.Reset()
+	})
```
njtran, line 95: nit: I think it's just as equivalent for testing purposes to put this in the BeforeEach() block, and it's where everything else like this is defined.

pkg/controllers/node/termination/terminator/metrics.go
njtran: Consider also doing namespace here as a label.

njtran (commented):
Thanks for the change! Once the comments are addressed, it all looks good to me.

njtran:
@chenfeilee let us know if you need any more help here!

chenfeilee:
Hi all, thanks for the review and so sorry that I didn't get the chance to get to this. Let me fix this within the next few days.

njtran:
@chenfeilee any update?

pkg/controllers/node/termination/terminator/suite_test.go
jonathan-innis: I think we generally put cleanup in the AfterEach block. Resetting the counter is probably something that should be done at the end of the test since the value will be changed during the test

njtran:
Closing, feel free to re-open when you get back to this!

PR #961, opened 2024-01-25
fix: don't mark cluster consolidated if constrained by budgets

njtran:
Fixes #N/A 

**Description**
Fixes a bug where we would mark the cluster as consolidated if budgets caused Consolidation to not occur. We only want to mark a cluster as consolidated if budgets aren't restricting the candidate set. 

make presubmit

### DIFF
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -88,8 +88,8 @@ func (c *consolidation) sortAndFilterCandidates(ctx context.Context, candidates 
 	return candidates, nil
 }
 
-// isConsolidated returns true if nothing has changed since markConsolidated was called.
-func (c *consolidation) isConsolidated() bool {
+// IsConsolidated returns true if nothing has changed since markConsolidated was called.
+func (c *consolidation) IsConsolidated() bool {
 	return c.lastConsolidationState.Equal(c.cluster.ConsolidationState())
 }
 
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -61,7 +61,7 @@ type consolidation struct {
 	lastConsolidationState time.Time
 }
 
-func makeConsolidation(clock clock.Clock, cluster *state.Cluster, kubeClient client.Client, provisioner *provisioning.Provisioner,
+func MakeConsolidation(clock clock.Clock, cluster *state.Cluster, kubeClient client.Client, provisioner *provisioning.Provisioner,
 	cloudProvider cloudprovider.CloudProvider, recorder events.Recorder, queue *orchestration.Queue) consolidation {
 	return consolidation{
 		queue:         queue,
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -211,95 +211,6 @@ var _ = Describe("Consolidation", func() {
 			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
 			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(0))
 		})
-		FIt("should not mark empty node consolidated if candidates can't be disrupted due to budgets", func() {
-			nodePool.Spec.Disruption.Budgets = []v1beta1.Budget{{Nodes: "0%"}}
-
-			ExpectApplied(ctx, env.Client, nodePool)
-			for i := 0; i < numNodes; i++ {
-				ExpectApplied(ctx, env.Client, nodeClaims[i], nodes[i])
-			}
-			// inform cluster state about nodes and nodeclaims
-			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, nodes, nodeClaims)
-
-			emptyConsolidation := disruption.NewEmptyNodeConsolidation(disruption.MakeConsolidation(fakeClock, cluster, env.Client, prov, cloudProvider, recorder, queue))
-			budgets, err := disruption.BuildDisruptionBudgets(ctx, cluster, fakeClock, env.Client)
-			Expect(err).To(Succeed())
-
-			candidates, err := disruption.GetCandidates(ctx, cluster, env.Client, recorder, fakeClock, cloudProvider, emptyConsolidation.ShouldDisrupt, queue)
-			Expect(err).To(Succeed())
-
-			var wg sync.WaitGroup
-			ExpectTriggerVerifyAction(&wg)
-			cmd, err := emptyConsolidation.ComputeCommand(ctx, budgets, candidates...)
-			Expect(err).To(Succeed())
-			Expect(cmd)
-			wg.Wait()
-
-			Expect(emptyConsolidation.IsConsolidated()).To(BeFalse())
-		})
-		FIt("should not mark empty node consolidated if all nodepools are blocked by budgets", func() {
-			// Create 10 nodepools
-			nps := test.NodePools(10, v1beta1.NodePool{
-				Spec: v1beta1.NodePoolSpec{
-					Disruption: v1beta1.Disruption{
-						ConsolidationPolicy: v1beta1.ConsolidationPolicyWhenUnderutilized,
-						Budgets: []v1beta1.Budget{{
-							Nodes: "0%",
-						}},
-					},
-				},
-			})
-			ExpectApplied(ctx, env.Client, nodePool)
-			for i := 0; i < len(nps); i++ {
-				ExpectApplied(ctx, env.Client, nps[i])
-			}
-			nodeClaims = make([]*v1beta1.NodeClaim, 0, 30)
-			nodes = make([]*v1.Node, 0, 30)
-			// Create 3 nodes for each nodePool
-			for _, np := range nps {
-				ncs, ns := test.NodeClaimsAndNodes(3, v1beta1.NodeClaim{
-					ObjectMeta: metav1.ObjectMeta{
-						Labels: map[string]string{
-							v1beta1.NodePoolLabelKey:     np.Name,
-							v1.LabelInstanceTypeStable:   mostExpensiveInstance.Name,
-							v1beta1.CapacityTypeLabelKey: mostExpensiveOffering.CapacityType,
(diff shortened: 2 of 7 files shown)


pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -62,12 +62,15 @@ func (m *MultiNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
...
 	disruptableCandidates := make([]*Candidate, 0, len(candidates))
+	constrainedByBudgets := false
 	for _, candidate := range candidates {
 		// If there's disruptions allowed for the candidate's nodepool,
 		// add it to the list of candidates, and decrement the budget.
 		if disruptionBudgetMapping[candidate.nodePool.Name] == 0 {
 			continue
 		}
+		// set constrainedByBudgets to true if any node was a candidate but was constrained by a budget
+		constrainedByBudgets = true
```
jonathan-innis, line 73: Does this make sense? Won't this always get called regardless if the candidate was constrained by a budget or not?

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -236,6 +238,8 @@ var _ = Describe("Consolidation", func() {
 			// Execute command, thus deleting all nodes
 			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
 			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(numNodes))
+			// Expect the cluster consolidated state to not change.
```
jonathan-innis, line 241: sanity check: I'm assuming that this failed prior to this change?

jonathan-innis: Should we test this in multiple places since we have 3 places where we are handling this logic but we only have a single test with emptiness where we are validating this behavior?

njtran: The event is only emitted for consolidation when we have no allowed disruptions due to a blocking budget. I can add another test with multiple nodepools here

pkg/controllers/disruption/controller.go
```diff
@@ -72,7 +72,7 @@ func NewController(clk clock.Clock, kubeClient client.Client, provisioner *provi
 		recorder:      recorder,
 		cloudProvider: cp,
 		lastRun:       map[string]time.Time{},
-		methods: []Method{
+		Methods: []Method{
```
jonathan-innis, line 75: Why the change to make methods publicly exposed?

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -227,6 +229,9 @@ var _ = Describe("Consolidation", func() {
 			ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
 			wg.Wait()
 
+			// Expect the cluster consolidated state to not change.
+			Expect(disruptionController.Methods[3].(*disruption.EmptyNodeConsolidation).IsConsolidated()).To(BeFalse())
```
jonathan-innis, line 233: This feels extremely dirty to me with exposing something publicly, getting a known ordering of it and then casting to get access to a member function. Is there a cleaner way to handle this that doesn't cause this many different hoops to jump through to get the testing to work-out?

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -62,12 +62,15 @@ func (m *MultiNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
 	// would have violated the budget anyway, preserving the ordering
 	// and only considering a number of nodes that can be disrupted.
 	disruptableCandidates := make([]*Candidate, 0, len(candidates))
+	constrainedByBudgets := false
 	for _, candidate := range candidates {
 		// If there's disruptions allowed for the candidate's nodepool,
 		// add it to the list of candidates, and decrement the budget.
 		if disruptionBudgetMapping[candidate.nodePool.Name] == 0 {
+			constrainedByBudgets = true
```
Bryce-Soghigian, line 70: Constained by budgets might make for an interesting event to create on the cluster. We can use the DisruptionBlocked with the reaason being constrained by budgets?

pkg/controllers/disruption/consolidation.go
```diff
@@ -88,8 +88,8 @@ func (c *consolidation) sortAndFilterCandidates(ctx context.Context, candidates
 	return candidates, nil
 }
 
-// isConsolidated returns true if nothing has changed since markConsolidated was called.
-func (c *consolidation) isConsolidated() bool {
+// IsConsolidated returns true if nothing has changed since markConsolidated was called.
```
Bryce-Soghigian, line 91: Is there a need for this to be public for anything other than unit tests?

pkg/controllers/disruption/emptynodeconsolidation.go
```diff
@@ -52,20 +52,29 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
...
 
 	empty := make([]*Candidate, 0, len(candidates))
+	constrainedByBudgets := false
 	for _, candidate := range candidates {
 		if len(candidate.pods) > 0 {
 			continue
 		}
+		if disruptionBudgetMapping[candidate.nodePool.Name] == 0 {
+			// set constrainedByBudgets to true if any node was a candidate but was constrained by a budget
+			constrainedByBudgets = true
```
Bryce-Soghigian, line 62: If we ever can't remove a candidate due to a budget constraint, we should fire an event?

njtran: We did something similar for NodePool here https://github.com/kubernetes-sigs/karpenter/pull/960. I think we can similarly add this for a candidate in the future too. One thing to consider is let's say a budget allows 3 disruptions, but we have 5 candidates. Do we emit an event for all 5 of the candidates, or just a random 2? This is probably trivial to do for Expiration and Drift, and we can just probably emit, but for multiNodeConsolidation, where the correctness of the command depends on how it's batched, this may make less sense.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -461,6 +461,337 @@ var _ = Describe("Consolidation", func() {
...
+				})
+				Expect(found).To(BeTrue())
+				Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 0))
+			}
+
+			// Execute the command in the queue, deleting all node claims
+			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
+			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(30))
+		})
+		It("should not mark empty node consolidated if candidates can't be disrupted due to budgets", func() {
```
jonathan-innis, line 528: Can you make this a bit more clear what the difference between this test and the one after it is? Maybe a context block or something to scope things appropriately if that's helpful. It was hard for me to quickly grok that these tests were relatively similar but had a one vs. many difference for the checks

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -461,6 +461,337 @@ var _ = Describe("Consolidation", func() {
...
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			cmd, err := emptyConsolidation.ComputeCommand(ctx, budgets, candidates...)
+			Expect(err).To(Succeed())
+			Expect(cmd)
+			wg.Wait()
+
+			Expect(emptyConsolidation.IsConsolidated()).To(BeFalse())
+		})
+		It("should not mark multi node consolidated if candidates can't be disrupted due to budgets", func() {
```
jonathan-innis, line 617:
```suggestion
		It("should not mark multi node consolidated if candidates can't be disrupted due to budgets with one nodepool", func() {
```

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -461,6 +461,337 @@ var _ = Describe("Consolidation", func() {
...
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			cmd, err := multiConsolidation.ComputeCommand(ctx, budgets, candidates...)
+			Expect(err).To(Succeed())
+			Expect(cmd)
+			wg.Wait()
+
+			Expect(multiConsolidation.IsConsolidated()).To(BeFalse())
+		})
+		It("should not mark multi node consolidated if all nodepools are blocked by budgets", func() {
```
jonathan-innis, line 643:
```suggestion
		It("should not mark multi node consolidated if all nodepools are blocked by budgets with many nodepools", func() {
```

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -461,6 +461,337 @@ var _ = Describe("Consolidation", func() {
...
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			cmd, err := multiConsolidation.ComputeCommand(ctx, budgets, candidates...)
+			Expect(err).To(Succeed())
+			Expect(cmd)
+			wg.Wait()
+
+			Expect(multiConsolidation.IsConsolidated()).To(BeFalse())
+		})
+		It("should not mark single node consolidated if candidates can't be disrupted due to budgets", func() {
```
jonathan-innis, line 706:
```suggestion
		It("should not mark single node consolidated if candidates can't be disrupted due to budgets with one nodepool", func() {
```

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -461,6 +461,337 @@ var _ = Describe("Consolidation", func() {
...
+			var wg sync.WaitGroup
+			ExpectTriggerVerifyAction(&wg)
+			cmd, err := singleConsolidation.ComputeCommand(ctx, budgets, candidates...)
+			Expect(err).To(Succeed())
+			Expect(cmd)
+			wg.Wait()
+
+			Expect(singleConsolidation.IsConsolidated()).To(BeFalse())
+		})
+		It("should not mark single node consolidated if all nodepools are blocked by budgets", func() {
```
jonathan-innis, line 732:
```suggestion
		It("should not mark single node consolidated if all nodepools are blocked by budgets with many nodepools", func() {
```

jonathan-innis (approved):
/lgtm
/approve

PR #962, opened 2024-01-25
update benchmark to have anti-affinity pods

tzneal:
Fixes #N/A 

**Description**

Updates the benchmark to have anti-affinity pods.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -35,6 +35,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.17.0
 )
 
+require github.com/evanphx/json-patch v4.12.0+incompatible // indirect
+
 require (
 	contrib.go.opencensus.io/exporter/ocagent v0.7.1-0.20200907061046-05415f1de66d // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -35,6 +35,8 @@ require (
 	sigs.k8s.io/controller-runtime v0.17.0
 )
 
+require github.com/evanphx/json-patch v4.12.0+incompatible // indirect
```
njtran, line 38: Did you add this directly into the go.mod yourself? Curious why it's in a different block than the rest

tzneal: go mod tidy added it. I imported sigs.k8s.io/controller-runtime/pkg/client/fake and its used there.

jonathan-innis: Can we just manually shift it down? I think if you shift it into the lower block, `go mod tidy` should just respect where you put it IIRC.

jonathan-innis (approved):
/lgtm
/approve

PR #963, opened 2024-01-25
feat: Add support for minValues in requirement for scheduling, consolidation layer and corresponding API changes.

nikmohan123:
**Description**
Add support for minValues in requirement for scheduling and consolidation layer. This allows users to enforce higher flexibility constraints on Karpenter so that Karpenter adheres to spot best practices. By enforcing minValues on Karpenter NodePools with node.kubernetes.io/instance-type or karpenter.k8s.aws/instance-family , customers can ensure that launches will have a better chance of providing a highly available instance type by giving PCO more options.

`make presubmit`
Will add functional tests once the API structure has been finalized as the minValues is extracted from the requirement of the API.

Also manually tested by injecting the `minValues` into requirement and tested the scheduling and consolidation path.


pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -233,6 +241,8 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
 		requirementsAndOffering: false,
 		fitsAndOffering:         false,
 	}
+	cumulativeMinRequirementsFromInstanceTypes := make(map[string]sets.Set[string])
```
jonathan-innis, line 244: How expensive is it to rebuild this map every time vs. keeping a running total of key/values with counts and subtracting off based on an instance type's key if it's incompatible? 

Might be worth doing some benchmarking here with `testing.b`: https://gobyexample.com/testing-and-benchmarking to see which one is preferred here. From a Big-O complexity, I think they are pretty similar since you have to iterate through all the keys and then call `--` on each value that you find for the instance type but it's probably worth checking if the map instantiations combined with building new sets each time is expensive.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -254,8 +264,19 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
+					if _, ok := cumulativeMinRequirementsFromInstanceTypes[req.Key]; !ok {
+						cumulativeMinRequirementsFromInstanceTypes[req.Key] = sets.Set[string]{}
+					}
+					cumulativeMinRequirementsFromInstanceTypes[req.Key] =
+						cumulativeMinRequirementsFromInstanceTypes[req.Key].Insert(it.Requirements.Get(req.Key).Values()...)
+				}
+			}
 		}
 	}
+	results.requirementInCompatibleWithMinValues = FindRequirementKeyInCompatibleWithMinValues(cumulativeMinRequirementsFromInstanceTypes, requirements)
```
jonathan-innis, line 279: Is it worth returning back no instance types when you don't meet this condition and then leverage the `FailureReason()` function logic to encapsulate if the reason we get no instance types here is because of `minValues`.

From an interface perspective, I think it's kinda weird that we would return back a response of instance types if we can't fulfill the requirements across all of the instance types that are passed in

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -101,7 +103,13 @@ func (n *NodeClaim) Add(pod *v1.Pod) error {
 
 	// Check instance type combinations
 	requests := resources.Merge(n.Spec.Resources.Requests, resources.RequestsForPods(pod))
+
 	filtered := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, requests)
+
+	if len(filtered.requirementInCompatibleWithMinValues) > 0 {
```
jonathan-innis, line 109:
```suggestion
	if len(filtered.requirementIncompatibleWithMinValues) > 0 {
```

nit: Incompatible is one word so we shouldn't camel case inside of it

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -276,3 +297,13 @@ func hasOffering(instanceType *cloudprovider.InstanceType, requirements scheduli
 	}
 	return false
 }
+
+func FindRequirementKeyInCompatibleWithMinValues(cumulativeMinRequirementsFromInstanceTypes map[string]sets.Set[string], requirements scheduling.Requirements) string {
```
jonathan-innis, line 301:
```suggestion
func FindRequirementKeyIncompatibleWithMinValues(cumulativeMinRequirementsFromInstanceTypes map[string]sets.Set[string], requirements scheduling.Requirements) string {
```

nit: Same comment here on incompatible

pkg/controllers/disruption/consolidation.go
```diff
@@ -45,7 +45,7 @@ import (
 const consolidationTTL = 15 * time.Second
 
 // MinInstanceTypesForSpotToSpotConsolidation is the minimum number of instanceTypes in a NodeClaim needed to trigger spot-to-spot single-node consolidation
-const MinInstanceTypesForSpotToSpotConsolidation = 15
+var MinInstanceTypesForSpotToSpotConsolidation = 15
```
jonathan-innis, line 48: What's the need for us to change this from a constant?

pkg/controllers/disruption/consolidation.go
```diff
@@ -175,18 +175,24 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
...
+	var inCompatibleRequirementKey string
+	// FilterByPriceAndFindIncompatibleRequirementWithMinValues returns the instanceTypes that are lower priced than the current candidate and the requirement for the NodeClaim that does not meet minValues.
+	// If we use this directly for spot-to-spot consolidation, we are bound to get repeated consolidations because the strategy that chooses to launch the spot instance from the list does
+	// it based on availability and price which could result in selection/launch of non-lowest priced instance in the list. So, we would keep repeating this loop till we get to lowest priced instance
 	// causing churns and landing onto lower available spot instance ultimately resulting in higher interruptions.
-	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions = filterByPrice(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
+	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions, inCompatibleRequirementKey =
+		FilterByPriceAndFindIncompatibleRequirementWithMinValues(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
```
jonathan-innis, line 184:
```suggestion
		filterByPriceWithMinValues(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
```

We can probably just shorten the name to describe it filtering by price while respecting minValues

pkg/controllers/disruption/consolidation.go
```diff
@@ -246,9 +258,16 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
 	}
 
 	// For single-node consolidation:
-	// We check whether we have 15 cheaper instances than the current candidate instance. If this is the case, we know the following things:
-	//   1) The current candidate is not in the set of the 15 cheapest instance types and
-	//   2) There were at least 15 options cheaper than the current candidate.
+
+	// If a user has minValues set in their NodePool that require less than 15 instance types, then the default 15 instance type minimum will continue to be enforced to enable a spot-to-spot consolidation.
+	// If minValues for the number of instance types is beyond 15, then number of instance types from minValues would be maintained to enable spot-to-spot consolidation.
+	if results.NewNodeClaims[0].Requirements.Has(v1.LabelInstanceTypeStable) && results.NewNodeClaims[0].Requirements.Get(v1.LabelInstanceTypeStable).MinValues != nil {
```
jonathan-innis, line 264: What's the need for this check? Shouldn't the check that we have at least 15 instance types be enough for this to succeed on its own?

jonathan-innis: I'm assuming that you added this to support truncating, but given the fact that no one is required to set the instance type requirement directly, I think we just have to throw-out the truncation of the instance types that we use for launch during consolidation. I think we simply have to accept that we will most likely get the bottom 15 instance types when we make a consolidation call and, if we don't, we'll just walk down a bit until we get to the bottom 15

pkg/controllers/disruption/helpers.go
```diff
@@ -157,15 +157,61 @@ func GetPodEvictionCost(ctx context.Context, p *v1.Pod) float64 {
 	return clamp(-10.0, cost, 10.0)
 }
 
-func filterByPrice(options []*cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) []*cloudprovider.InstanceType {
+// FilterByPriceAndFindIncompatibleRequirementWithMinValues returns the instanceTypes that are lower priced than the current candidate and iterates over the cumulative minimum requirement of the InstanceTypeOptions to see if it meets the minValues of requirements.
+// The minValues requirement is checked again after filterByPrice as it may result in more constrained InstanceTypeOptions for a NodeClaim
+func FilterByPriceAndFindIncompatibleRequirementWithMinValues(options []*cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) ([]*cloudprovider.InstanceType, string) {
```
jonathan-innis, line 162: Do we take the same stance that we did with the other filter function where we say that the interface says that we don't return instance types unless we meet the minValues requirements?

jonathan-innis:
Consider using the `scheduling_benchmark_test.go` to see what the performance impact is on this change. Would be worth expanding that suite to add some checks for what happens when you pass requirements with `minValues` in

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Addressed.

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Yes, we can do that. I thought it would be important to differentiate the situation where we would have gotten the instanceTypes without the `minValues`. Changed this to return no instanceTypes if it does not meet min requirements.

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Do not want to block rest of the PR. I will follow-up on this one separately since it is tricky to inject the `minvalues` into scheduler logic without API help for simulations.

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Addressed.

pkg/controllers/disruption/helpers.go
nikmohan123: Yes, changed this behavior to return only when minValues requirement is met.

pkg/controllers/disruption/consolidation.go
nikmohan123: Addressed.

pkg/controllers/disruption/consolidation.go
nikmohan123: Reverted. This was dynamically changed previously to avoid churns but is no more relevant.

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Here is the performance/benchmarking analysis for various scenarios:

**Without my changes**
```
scheduled 12610 against 2108 nodes in total in 22.999827136s 548.2649902295335 pods/sec
400 instances 10 pods    2 nodes    6.910374ms per scheduling    691.037µs per pod
400 instances 100 pods   17 nodes   74.308797ms per scheduling   743.087µs per pod
400 instances 500 pods   84 nodes   382.199278ms per scheduling  764.398µs per pod
400 instances 1000 pods  167 nodes  787.569292ms per scheduling  787.569µs per pod
400 instances 1500 pods  253 nodes  12.382227s per scheduling    8.254818ms per pod
400 instances 2000 pods  334 nodes  1.591589125s per scheduling  795.794µs per pod
400 instances 2500 pods  417 nodes  2.111523708s per scheduling  844.609µs per pod
400 instances 5000 pods  834 nodes  4.63844175s per scheduling   927.688µs per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	29.812s
```

**With my changes and without minValues in requirement**
_( a slight increase which is expected due to additional loop to accumulate the requirements across InstanceTypeOptions which has `minValues` in it)_
```
scheduled 12610 against 2106 nodes in total in 23.767375254s 530.559216793523 pods/sec
400 instances 10 pods    2 nodes    7.280119ms per scheduling     728.011µs per pod
400 instances 100 pods   17 nodes   78.313497ms per scheduling    783.134µs per pod
400 instances 500 pods   84 nodes   407.551264ms per scheduling   815.102µs per pod
400 instances 1000 pods  167 nodes  812.94202ms per scheduling    812.942µs per pod
400 instances 1500 pods  251 nodes  12.858819459s per scheduling  8.572546ms per pod
400 instances 2000 pods  334 nodes  1.64878525s per scheduling    824.392µs per pod
400 instances 2500 pods  417 nodes  2.106910541s per scheduling   842.764µs per pod
400 instances 5000 pods  834 nodes  4.767597208s per scheduling   953.519µs per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	30.899s
```

**With minValue = 3 for a requirement (I have taken InstanceType in all cases)**
_( little more increase in time to fetch the key that violates the `minValues` and this wouldn’t change no matter how much is the value of minValue as shown below for few other values)_
```
scheduled 12610 against 2105 nodes in total in 26.640151567s 473.345655270986 pods/sec
400 instances 10 pods    2 nodes    9.5229ms per scheduling       952.29µs per pod
400 instances 100 pods   17 nodes   100.883319ms per scheduling   1.008833ms per pod
400 instances 500 pods   84 nodes   506.510021ms per scheduling   1.01302ms per pod
400 instances 1000 pods  167 nodes  1.025684042s per scheduling   1.025684ms per pod
400 instances 1500 pods  251 nodes  13.503289875s per scheduling  9.002193ms per pod
400 instances 2000 pods  333 nodes  2.128573333s per scheduling   1.064286ms per pod
400 instances 2500 pods  417 nodes  2.723903459s per scheduling   1.089561ms per pod
400 instances 5000 pods  834 nodes  6.021298125s per scheduling   1.204259ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	32.774s
```

**With minValue = 5 for a requirement**
```
scheduled 12610 against 2105 nodes in total in 26.527231966s 475.36056593323644 pods/sec
400 instances 10 pods    2 nodes    9.680388ms per scheduling     968.038µs per pod
400 instances 100 pods   17 nodes   103.119927ms per scheduling   1.031199ms per pod
400 instances 500 pods   84 nodes   506.872854ms per scheduling   1.013745ms per pod
400 instances 1000 pods  167 nodes  1.028840041s per scheduling   1.02884ms per pod
400 instances 1500 pods  251 nodes  13.454834917s per scheduling  8.969889ms per pod
400 instances 2000 pods  333 nodes  2.144536625s per scheduling   1.072268ms per pod
400 instances 2500 pods  417 nodes  2.686135083s per scheduling   1.074454ms per pod
400 instances 5000 pods  834 nodes  5.957702458s per scheduling   1.19154ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	32.776s
```

**With minValue = 10 for a requirement**
```
scheduled 12610 against 2104 nodes in total in 26.174312078s 481.7700638099651 pods/sec
400 instances 10 pods    2 nodes    9.627454ms per scheduling    962.745µs per pod
400 instances 100 pods   17 nodes   100.234663ms per scheduling  1.002346ms per pod
400 instances 500 pods   83 nodes   513.745583ms per scheduling  1.027491ms per pod
400 instances 1000 pods  166 nodes  1.035767916s per scheduling  1.035767ms per pod
400 instances 1500 pods  251 nodes  13.0232s per scheduling      8.682133ms per pod
400 instances 2000 pods  334 nodes  2.174360208s per scheduling  1.08718ms per pod
400 instances 2500 pods  417 nodes  2.797213459s per scheduling  1.118885ms per pod
400 instances 5000 pods  834 nodes  5.968021292s per scheduling  1.193604ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	31.468s
```

**With minValue = 15 for a requirement**
```
scheduled 12610 against 2104 nodes in total in 26.262151959s 480.1586716765064 pods/sec
400 instances 10 pods    2 nodes    9.596ms per scheduling        959.6µs per pod
400 instances 100 pods   17 nodes   100.269287ms per scheduling   1.002692ms per pod
400 instances 500 pods   83 nodes   509.614771ms per scheduling   1.019229ms per pod
400 instances 1000 pods  166 nodes  1.030548292s per scheduling   1.030548ms per pod
400 instances 1500 pods  251 nodes  13.134777458s per scheduling  8.756518ms per pod
400 instances 2000 pods  334 nodes  2.112007375s per scheduling   1.056003ms per pod
400 instances 2500 pods  417 nodes  2.796082833s per scheduling   1.118433ms per pod
400 instances 5000 pods  834 nodes  6.032787959s per scheduling   1.206557ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	31.351s
```

**With minValue = 20 for a requirement**
```
scheduled 12610 against 2104 nodes in total in 26.592401554s 474.19560713211393 pods/sec
400 instances 10 pods    2 nodes    9.572666ms per scheduling     957.266µs per pod
400 instances 100 pods   17 nodes   98.358816ms per scheduling    983.588µs per pod
400 instances 500 pods   83 nodes   505.555083ms per scheduling   1.01111ms per pod
400 instances 1000 pods  166 nodes  1.035924208s per scheduling   1.035924ms per pod
400 instances 1500 pods  251 nodes  13.335883375s per scheduling  8.890588ms per pod
400 instances 2000 pods  334 nodes  2.157725417s per scheduling   1.078862ms per pod
400 instances 2500 pods  417 nodes  2.821344459s per scheduling   1.128537ms per pod
400 instances 5000 pods  834 nodes  6.098517833s per scheduling   1.219703ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	31.832s
```

**With minValue = 30 for a requirement**
```
scheduled 12610 against 2104 nodes in total in 26.064046588s 483.80822054721534 pods/sec
400 instances 10 pods    2 nodes    9.512452ms per scheduling     951.245µs per pod
400 instances 100 pods   17 nodes   99.078468ms per scheduling    990.784µs per pod
400 instances 500 pods   83 nodes   506.811166ms per scheduling   1.013622ms per pod
400 instances 1000 pods  166 nodes  1.029381542s per scheduling   1.029381ms per pod
400 instances 1500 pods  251 nodes  13.102935959s per scheduling  8.73529ms per pod
400 instances 2000 pods  334 nodes  2.115130167s per scheduling   1.057565ms per pod
400 instances 2500 pods  417 nodes  2.698365667s per scheduling   1.079346ms per pod
400 instances 5000 pods  834 nodes  5.966055166s per scheduling   1.193211ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	31.313s
```

**With minValue = 60 for a requirement**
```
scheduled 12610 against 2105 nodes in total in 26.556147719s 474.8429679421456 pods/sec
400 instances 10 pods    2 nodes    9.573855ms per scheduling     957.385µs per pod
400 instances 100 pods   17 nodes   100.061652ms per scheduling   1.000616ms per pod
400 instances 500 pods   84 nodes   504.558083ms per scheduling   1.009116ms per pod
400 instances 1000 pods  167 nodes  1.06675975s per scheduling    1.066759ms per pod
400 instances 1500 pods  251 nodes  13.458783708s per scheduling  8.972522ms per pod
400 instances 2000 pods  333 nodes  2.160281542s per scheduling   1.08014ms per pod
400 instances 2500 pods  417 nodes  2.699132708s per scheduling   1.079653ms per pod
400 instances 5000 pods  834 nodes  5.937499875s per scheduling   1.187499ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	32.652s
```

**With minValue = 100 for a requirement (Cannot be a real usecase since the max is capped at 60 today to CreateFleet)**
```
scheduled 12610 against 2104 nodes in total in 26.30848654s 479.3130148641382 pods/sec
400 instances 10 pods    2 nodes    9.66196ms per scheduling      966.196µs per pod
400 instances 100 pods   17 nodes   99.390145ms per scheduling    993.901µs per pod
400 instances 500 pods   83 nodes   511.4355ms per scheduling     1.022871ms per pod
400 instances 1000 pods  166 nodes  1.141197958s per scheduling   1.141197ms per pod
400 instances 1500 pods  251 nodes  13.184101875s per scheduling  8.789401ms per pod
400 instances 2000 pods  334 nodes  2.158486542s per scheduling   1.079243ms per pod
400 instances 2500 pods  417 nodes  2.704895917s per scheduling   1.081958ms per pod
400 instances 5000 pods  834 nodes  5.958880209s per scheduling   1.191776ms per pod
PASS
ok  	sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling	31.518s
```

_**There is an approx 1 second increase for the worst case scenario with 5000 pods and most other cases are in milliseconds increase in value. I think this is fine and expected. Let me know your thoughts.**_

pkg/scheduling/requirement.go
```diff
@@ -36,8 +36,10 @@ type Requirement struct {
 	values      sets.Set[string]
 	greaterThan *int
 	lessThan    *int
+	MinValues   *int
 }
 
+// TODO: Make changes to accept MinValues from API
```
jonathan-innis, line 42: nit: Should we drop this TODO before we plan on merging this change?

pkg/controllers/disruption/consolidation.go
```diff
@@ -170,11 +170,18 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
 		return c.computeSpotToSpotConsolidation(ctx, candidates, results, candidatePrice)
 	}
 
-	// filterByPrice returns the instanceTypes that are lower priced than the current candidate. If we use this directly for spot-to-spot consolidation
-	// we are bound to get repeated consolidations because the strategy that chooses to launch the spot instance from the list does it based on availability and price which could
-	// result in selection/launch of non-lowest priced instance in the list. So, we would keep repeating this loop till we get to lowest priced instance
+	var inCompatibleRequirementKey string
```
jonathan-innis, line 173:
```suggestion
	var incompatibleRequirementKey string
```

nit

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -216,6 +220,10 @@ func (r filterResults) FailureReason() string {
 		return "no instance type which met the scheduling requirements and the required offering had the required resources"
 	}
 
+	if len(r.requirementIncompatibleWithMinValues) > 0 {
+		return "minimum requirement is not met for " + r.requirementIncompatibleWithMinValues
```
jonathan-innis, line 224:
```suggestion
		return "minValues requirement is not met for " + r.requirementIncompatibleWithMinValues
```

nit

pkg/controllers/disruption/helpers.go
```diff
@@ -145,15 +145,63 @@ func GetPodEvictionCost(ctx context.Context, p *v1.Pod) float64 {
 	return clamp(-10.0, cost, 10.0)
 }
 
-func filterByPrice(options []*cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) []*cloudprovider.InstanceType {
+// filterByPriceWithMinValues returns the instanceTypes that are lower priced than the current candidate and iterates over the cumulative minimum requirement of the InstanceTypeOptions to see if it meets the minValues of requirements.
+// The minValues requirement is checked again after filterByPrice as it may result in more constrained InstanceTypeOptions for a NodeClaim
+func filterByPriceWithMinValues(options []*cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) ([]*cloudprovider.InstanceType, string) {
 	var result []*cloudprovider.InstanceType
+	// cumulativeMinRequirementsFromInstanceTypes is a map for the requirement key with the cumulative values that has minValues supported across InstanceTypeOptions
```
jonathan-innis, line 152: You've given a super thorough explanation of what this variable is tracking here, which I think is great! I wonder if you should do the same thing in the `provisioning/scheduling/nodeclaim.go` where we are basically performing the same logic

pkg/controllers/disruption/helpers.go
```diff
@@ -145,15 +145,63 @@ func GetPodEvictionCost(ctx context.Context, p *v1.Pod) float64 {
...
+
 	for _, it := range options {
 		launchPrice := worstLaunchPrice(it.Offerings.Available(), reqs)
 		if launchPrice < price {
 			result = append(result, it)
+			// Check for minValues in the requirements for the InstanceType which are lower priced than the current candidate
+			for _, req := range reqs {
+				// Check if the scheduling requirement has MinValues
+				if req.MinValues != nil {
+					if _, ok := cumulativeMinRequirementsFromInstanceTypes[req.Key]; !ok {
```
jonathan-innis, line 192: Is it worth building this set of requirements in a pscheduling function as well? I'm curious because we don't lose any performance from doing it in both places and I think we can just build a function that takes a set of filtered instance types and determines whether we meet the requirements. Actually, come to think of it, I wonder if this just means that we check this outside of the filter function entirely. Do we filter out the instance types and then after we filter, we validate our minRequirements in a separate function in both places?

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -208,7 +208,8 @@ func filterOutSameType(newNodeClaim *scheduling.NodeClaim, consolidate []*Candid
 		}
 	}
 
-	return filterByPrice(newNodeClaim.InstanceTypeOptions, newNodeClaim.Requirements, maxPrice)
+	filterByPrice, _ := filterByPriceWithMinValues(newNodeClaim.InstanceTypeOptions, newNodeClaim.Requirements, maxPrice)
```
jonathan-innis, line 211: If we're ignoring a value here, I'm imagining that we need to comment why we're ignoring it. Even more than that though, I'm thinking that if we don't respect the minValues here, we could wind out in another situation where we don't respect the minValues that have been passed through. Particularly, because the `filterOutSameType` function has the potential to further filter down more instance types

jonathan-innis (commented):
Nice work 🎉

pkg/scheduling/requirement.go
nikmohan123: This is actually removed in the API changes PR. So, once this PR and the API PR is approved, this will automatically go away. And I plan to merge these 2 PRs together so that we do not block other releases.

pkg/controllers/disruption/consolidation.go
nikmohan123: Addressed.

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Addressed.

pkg/controllers/disruption/helpers.go
nikmohan123: Initially I thought of adding this at both the places but given the verbosity, I thought a single place was sufficient. Added it at the other place as well.

pkg/controllers/disruption/helpers.go
nikmohan123: >Actually, come to think of it, I wonder if this just means that we check this outside of the filter function entirely. Do we filter out the instance types and then after we filter, we validate our minRequirements in a separate function in both places?

Yes, this statement is right. The conditions to filter the instanceTypes are different at these 2 places. One checks the compatibility of each instanceType against the requirement and the other checks the prices of the instanceTypes to filter. 

And once we filter instanceTypes which already exists today, we are just storing it in the map if their requirement has `minValues` in it and then like you said we validate our minRequirements in a separate function in both places. This will save us one additional loop of InstanceTypeOptions.

pkg/controllers/disruption/multinodeconsolidation.go
nikmohan123: >we could wind out in another situation where we don't respect the minValues that have been passed through.

This case is already handled in the `computeConsolidation` where we try to compute `incompatibleRequirementKey` if minValues is not met before we get to `filterOutSameType`. And if the minValues is not met after `filterOutSameType`, we return empty InstanceTypeOptions and the check [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/multinodeconsolidation.go#L151) is going to prevent from consolidation.

I will add these details in the code.

pkg/controllers/disruption/consolidation.go
```diff
@@ -219,8 +226,14 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
...
 
-	// Possible replacements that are lower priced than the current candidate
-	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions = filterByPrice(instanceTypeOptionsWithSpotOfferings, results.NewNodeClaims[0].Requirements, candidatePrice)
+	var inCompatibleRequirementKey string
+	// Possible replacements that are lower priced than the current candidate and the requirement that is not compatible with minValues
+	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions, inCompatibleRequirementKey =
+		filterByPriceWithMinValues(instanceTypeOptionsWithSpotOfferings, results.NewNodeClaims[0].Requirements, candidatePrice)
+
+	if len(inCompatibleRequirementKey) > 0 {
+		return Command{}, pscheduling.Results{}, fmt.Errorf("minimum requirement is not met for %s", inCompatibleRequirementKey)
```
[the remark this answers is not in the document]
nikmohan123: Addressed.

pkg/controllers/disruption/consolidation.go
```diff
@@ -170,11 +170,18 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
...
+	// If we use this directly for spot-to-spot consolidation, we are bound to get repeated consolidations because the strategy that chooses to launch the spot instance from the list does
+	// it based on availability and price which could result in selection/launch of non-lowest priced instance in the list. So, we would keep repeating this loop till we get to lowest priced instance
 	// causing churns and landing onto lower available spot instance ultimately resulting in higher interruptions.
-	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions = filterByPrice(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
+	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions, inCompatibleRequirementKey =
+		filterByPriceWithMinValues(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
+
+	if len(inCompatibleRequirementKey) > 0 {
+		return Command{}, pscheduling.Results{}, fmt.Errorf("minimum requirement is not met for %s", inCompatibleRequirementKey)
```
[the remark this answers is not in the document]
nikmohan123: Addressed.

pkg/controllers/disruption/consolidation.go
```diff
@@ -219,8 +226,14 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
...
 
-	// Possible replacements that are lower priced than the current candidate
-	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions = filterByPrice(instanceTypeOptionsWithSpotOfferings, results.NewNodeClaims[0].Requirements, candidatePrice)
+	var incompatibleRequirementKey string
+	// Possible replacements that are lower priced than the current candidate and the requirement that is not compatible with minValues
+	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions, incompatibleRequirementKey =
+		filterByPriceWithMinValues(instanceTypeOptionsWithSpotOfferings, results.NewNodeClaims[0].Requirements, candidatePrice)
+
+	if len(incompatibleRequirementKey) > 0 {
+		return Command{}, pscheduling.Results{}, fmt.Errorf("minValues requirement is not met for %s", incompatibleRequirementKey)
```
jonathan-innis, line 235: Is this an error or should this be handled in the same way that we handle other times that we don't have instance type options (see L238)

pkg/controllers/provisioning/provisioner.go
```diff
@@ -343,7 +343,11 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 	if err := latest.Spec.Limits.ExceededBy(latest.Status.Resources); err != nil {
 		return "", err
 	}
-	nodeClaim := n.ToNodeClaim(latest)
+	nodeClaim, err := n.ToNodeClaim(latest)
```
jonathan-innis, line 346: I'll let it be known that I'm not a huge fan of this call being error-able, but I'm willing to take incremental improvement here if need-be

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -77,5 +86,22 @@ func (i *NodeClaimTemplate) ToNodeClaim(nodePool *v1beta1.NodePool) *v1beta1.Nod
 		Spec: i.Spec,
 	}
 	nc.Spec.Requirements = i.Requirements.NodeSelectorRequirements()
-	return nc
+	return nc, nil
+}
+
+func FindIncompatibleRequirementKeyAcrossInstanceTypeOptions(requirements scheduling.Requirements, instanceTypes cloudprovider.InstanceTypes) string {
```
jonathan-innis, line 92: Is there a way to shorten this function name? I get that the function name is highly descriptive, but it feels like an immediate code smell when we start getting function names this long.

pkg/controllers/disruption/consolidation.go
nikmohan123: We can merge these 2. I was thinking if this would bring more clarity since `minValues` is something that is controlled by the customer and would inform them to relax the condition ?

For the sake of decreasing the noise in the logs, we can skip this altogether and just rely on L238 to bring a closure and probably publish an event specifically for this ?

pkg/controllers/provisioning/provisioner.go
nikmohan123: I have now restructured the code to proactively identify this instead of identifying it here in the NodeClaim creation. It is not as good as introducing this into the scheduler but very close. The catch is we will introduce an additional sorting time complexity into the scheduler path.

pkg/controllers/disruption/consolidation.go
nikmohan123: Merged this and created a separate event for not meeting the minValues requirement.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -276,3 +292,68 @@ func hasOffering(instanceType *cloudprovider.InstanceType, requirements scheduli
...
+	//		node.kubernetes.io/instance-type:  ["c4.large","c4.xlarge","c5.2xlarge"],
+	//		karpenter.k8s.aws/instance-family: ["c4","c5"] // minimum requirement failed for this.
+	//	}
+	//  so, returns "karpenter.k8s.aws/instance-family"
+	// Key -> requirement key supporting MinValues
+	// value -> cumulative set of values for the key from all the instanceTypes
+	cumulativeMinRequirementsFromInstanceTypes := make(map[string]sets.Set[string])
+	// We validate if sorting by price and truncating the number of instance types to 100 breaks the minValue requirement since we pass the same bounded request to the Launch API.
+	// If minValue requirement fails, we return the incompatible key.
+	its := lo.Slice(instanceTypes.OrderByPrice(requirements), 0, MaxInstanceTypes)
```
jonathan-innis, line 330: Coupling this to checking the actual validity of the minValues concerns me, specifically when we are making this call in the context of the scheduling loop where it shouldn't matter that we are truncating the instance types at the end of the loop. Maybe I'm missing something here, but I would have expected that we would just be validating the check for minValues within the context of the scheduling loop and then only validating that it works after truncation at the end of the scheduling loop after all of the decisions have been made. Also, we are going to call a sortByPrice each time that we run this function, which is definitely going to impact performance on the scheduling loop if we call this when we are filtering out instance types

pkg/controllers/disruption/helpers.go
```diff
@@ -145,15 +145,33 @@ func GetPodEvictionCost(ctx context.Context, p *v1.Pod) float64 {
 	return clamp(-10.0, cost, 10.0)
 }
 
-func filterByPrice(options []*cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) []*cloudprovider.InstanceType {
+// MinimumRequirementsInfo provides info on incompatible minValue requirement key if exists and minimum number of
+// instanceTypes required to meet all the minValue requirements.
+type MinimumRequirementsInfo struct {
```
jonathan-innis, line 150: Is there a way to surface this data that doesn't require the creation of an entirely new struct?

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -256,6 +265,13 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
 			results.remaining = append(results.remaining, it)
 		}
 	}
+	if requirements.HasMinValues() {
+		// We do not need minimum number of instanceTypes to satisfy all the minimum requirements here since we always send
```
jonathan-innis, line 269: Should this comment talk about the 100 instance types or does it just need to say something like, "We don't care about the minimum number of instance types that meet our requirements here, we only care if they meet our requirements"

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -137,6 +137,39 @@ func (r Results) NonPendingPodSchedulingErrors() string {
...
+		// The InstanceTypeOptions are truncated due to limitations in sending the number of instances to launch API which is capped to 100 today.
+		newNodeClaim.InstanceTypeOptions = lo.Slice(newNodeClaim.InstanceTypeOptions.OrderByPrice(newNodeClaim.NodeClaimTemplate.Requirements), 0, MaxInstanceTypes)
+		// Only check for a validity of NodeClaim if its requirement has minValues in it.
+		if newNodeClaim.NodeClaimTemplate.Requirements.HasMinValues() {
+			// Check if the truncated InstanceTypeOptions in each NewNodeClaim from the results still satisfy the minimum requirements
+			incompatibleKey, _ := IncompatibleReqAcrossInstanceTypes(newNodeClaim.NodeClaimTemplate.Requirements, newNodeClaim.InstanceTypeOptions)
+			// If number of instancetypes in the nodeclaim cannot satisfy the minimum requirements, add its Pods to error map with reason.
+			if len(incompatibleKey) > 0 {
+				for _, pod := range newNodeClaim.Pods {
+					errorPods[pod] = fmt.Errorf("pod didn’t schedule because NodePool %q couldn’t meet minValues requirements after truncating to 100 instance types", newNodeClaim.NodeClaimTemplate.NodePoolName)
```
jonathan-innis, line 153: nit: Can you do a direct assignment to `PodErrors` here?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -137,6 +137,39 @@ func (r Results) NonPendingPodSchedulingErrors() string {
 	return msg.String()
 }
 
+func (r Results) NodeClaimsMeetingMinimumRequirements() Results {
```
jonathan-innis, line 140: nit: Consider describing what this function does above it

jonathan-innis, line 140:
```suggestion
func (r Results) TruncateInstanceTypes(int maxInstanceTypes) Results {
```

nit: It's a tad odd to have a name here that indicates that you are just returning NodeClaims from reading it but actually returns an entire results object. Consider allowing the truncation point be configurable to generalize it.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -256,6 +265,13 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
+		// 100 instancetypes to launch path
+		results.requirementIncompatibleWithMinValues, _ = IncompatibleReqAcrossInstanceTypes(requirements, results.remaining)
+	}
+	// If minValues is NOT met for any of the requirement across InstanceTypes, then return empty InstanceTypeOptions as we cannot launch with the remaining InstanceTypes.
+	results.remaining = lo.Ternary(len(results.requirementIncompatibleWithMinValues) > 0, []*cloudprovider.InstanceType{}, results.remaining)
```
jonathan-innis, line 274: Sanity check: When we make the `ToNodeClaim()` call and we make the instance type requirement do we carry the minValues requirement for instance types with us? If not, we need to add that or else we are going to be generating an instance type requirement without the minValues that have been specified

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Changed it.

pkg/controllers/provisioning/scheduling/scheduler.go
nikmohan123: Done

pkg/controllers/provisioning/scheduling/scheduler.go
nikmohan123: Done.

pkg/controllers/disruption/helpers.go
nikmohan123: Removed this.

pkg/controllers/provisioning/scheduling/nodeclaim.go
nikmohan123: Restructured the code and validated the min requirements against the result of the scheduler now. This is no more a concern.

jonathan-innis (approved):
Really, really nice work. These changes are absolutely non-trivial

/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

Really really awesome work here! 🎉

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
/remove-hold

Nice work! Ship it!

nikmohan123:
Created an issue to fix the corner case of the topology spread not behaving as expected with `minValues` : https://github.com/kubernetes-sigs/karpenter/issues/1034.

jonathan-innis (approved):
/lgtm

PR #966, opened 2024-01-25
RFC: Propose `minValues` to enforce Flexibility on NodeClaim Requests

jonathan-innis:
Fixes #N/A 

**Description**

This RFC proposes a change where users can pass through `minValues` into their different requirement block keys to ensure that Karpenter maintains their request minimum flexibility requirements as it is scheduling pods to nodes. This is useful to maintain spot flexibility when Karpenter is launching new nodes.

### DIFF
--- a/designs/nodepool-requirement-flexibility.md
+++ b/designs/nodepool-requirement-flexibility.md
@@ -19,7 +19,7 @@ spec:
   template:
     spec:
       requirements:
-        - key: karpenter.kwok.sh.instance-family
+        - key: karpenter.kwok.sh/instance-family
           operator: In
           values: ["c", "m", "r"]
           minValues: 2
@@ -45,7 +45,7 @@ spec:
   template:
     spec:
       requirements:
-      - key: karpenter.kwok.sh.instance-family
+      - key: karpenter.kwok.sh/instance-family
         operator: In
         values: ["c", "m", "r"]
         minValues: 2
@@ -69,7 +69,7 @@ kind: EC2NodeClass
 spec:
   minValues:
     node.kubernetes.io/instance-type: 10
-    karpenter.kwok.sh.instance-family: 2
+    karpenter.kwok.sh/instance-family: 2
 ---
 apiVersion: karpenter.sh/v1beta1
 kind: NodePool
@@ -77,7 +77,7 @@ spec:
   template:
     spec:
       requirements:
-      - key: karpenter.kwok.sh.instance-family
+      - key: karpenter.kwok.sh/instance-family
         operator: In
         values: ["c", "m", "r"]
       - key: node.kubernetes.io/instance-type
--- a/designs/nodepool-requirement-flexibility.md
+++ b/designs/nodepool-requirement-flexibility.md
@@ -6,7 +6,7 @@
 
 ## Motivation
 
-Karpenter’s scheduling algorithm uses a well-known bin-packing algorithm ([First Fit Descending](https://en.wikipedia.org/wiki/First-fit-decreasing_bin_packing)) to pack as many pods onto a set of instance type options as possible. This algorithm will necessarily continue packing pods onto Preflight Nodes so long as you have at least one instance type option that can fulfill all the pod requests.
+Karpenter’s scheduling algorithm uses a well-known bin-packing algorithm ([First Fit Decreasing](https://en.wikipedia.org/wiki/First-fit-decreasing_bin_packing)) to pack as many pods onto a set of instance type options as possible. This algorithm will necessarily continue packing pods onto Preflight Nodes so long as you have at least one instance type option that can fulfill all the pod requests.
 
 This methodology works fine for on-demand instance types, but starts to break down when you are requesting spot capacity. If you are launching spot capacity and want to ensure that you will not launch an instance that will immediately get interrupted after launch, you need to make sure that you have enough launch options in your launch request that your likelihood for launching a low-availability instance will be slim. Karpenter currently has no mechanism to ensure that today.
 
@@ -31,8 +31,7 @@ spec:
 ## Use-Cases
 
 1. I want to configure Karpenter to enforce more flexibility when launching spot nodes, ensuring that I am constantly using enough instance types that I get a high-availability instance type
-2. I want to increase the number of nodes that Karpenter creates by ensuring that Karpenter doesn’t over-pack pods onto nodes when it’s scheduling
-3. I want to ensure that application pods scheduling to this NodePool *must* be flexible to different instance types, architectures, etc.
+2. want to ensure that application pods scheduling to this NodePool *must* be flexible to different instance types, architectures, etc.
 
 ## Solutions
 
@@ -84,14 +83,16 @@ spec:
         operator: Exists
 ```
 
-Enabling this solution allows specific Cloud Providers to make this change without pushing this new field down into the neutral API; however, this means we need to define a neutral way where a cloudprovider can define its own flexibility requirements and pass them back to the scheduler. One option here is to allow CloudProviders to layer on their own flexibility requirements for NodePools through the [CloudProvider interface](https://github.com/kubernetes-sigs/karpenter/blob/4e85912c81ed9a51fc8ebd107db8e88e7828fae7/pkg/cloudprovider/types.go). This function in the interface might look like `func GetFlexibility(np *v1beta1.NodePool) scheduling.Flexibility` where `scheduling.Flexibility` is a `map[string]int`.
+In this solution, the onus is on the Cloud Provider to implement and pass-down some form of flexibility or minValues down into the scheduler. Cloud Providers could choose to selectively enable or disable this feature by surfacing this API through their NodeClass. 
+
+Enabling this allows specific Cloud Providers to make this change without pushing this new field down into the neutral API; however, this means we need to define a neutral way where a Cloud Provider can define its own flexibility requirements and pass them back to the scheduler. One option here is to allow CloudProviders to layer on their own flexibility requirements for NodePools through the [CloudProvider interface](https://github.com/kubernetes-sigs/karpenter/blob/4e85912c81ed9a51fc8ebd107db8e88e7828fae7/pkg/cloudprovider/types.go). This function in the interface might look like `func GetFlexibility(np *v1beta1.NodePool) scheduling.Flexibility` where `scheduling.Flexibility` is a `map[string]int`.
 
 #### Pros/Cons
 
 1. 👍👍 Isolates the minValues feature to whether a Cloud Provider wants to implement it or not
 2. 👍 Enables Cloud Providers to have the concept of “default” minValues that is controlled at runtime, allowing cloud providers to enforce certain minimums from their users for things like spot best practices without having to have the user specify it directly (e.g. “don’t allow a spot instance to launch without at least 10 spot instances in the request”).
 3. 👎 Leaks more scheduling behavior into the NodeClass (What if a Cloud Provider doesn’t implement a NodeClass directly? They would have to design one specifically for this feature). As a guiding design principle, we have generally chosen to avoid scheduler-based decision making driven through the NodeClass API
-4. 👎 Requires an additional method on the cloud providers OR requires piggy-backing off of `GetInstanceTypes()` to surface additional minValues information into the scheduler which may be awkward for cloud providers that don’t care about this requirement
+4. 👎 Requires an additional method on the cloud providers which expands our existing interface OR requires piggy-backing off of `GetInstanceTypes()` to surface additional minValues information into the scheduler which may be awkward for cloud providers that don’t care about this requirement
 
 ### Solution 3: `flexibility` field in the `spec.requirements` section of the NodeClaim Template (becomes part of the template on the NodePool)
 
@@ -149,6 +150,8 @@ A check to `minValues` should be done in the context of the pod `[Add()](https:/
 The scheduling simulation should return a result from consolidation that respects `minValues` consistent with the launch behavior. Karpenter will then reduce the set of instance types here to be strictly cheaper than the combined price of all instance types that we plan to replace (this is our `filterByPrice` function).
 
 It’s possible that after Karpenter reduces these instance types down to only cheaper ones, it no longer fulfills the `minValues` criteria of the NodePool requirements. As a result, consolidation will need to re-check whether the `minValues` is met after the instance types are filtered down by price to ensure that we can proceed with the launch. If the `minValues` is not satisfied, we should reject the consolidation decision.
+
+It's important to note here that by setting high `minValues` requirements, you are necessarily constraining the consolidation decisions that Karpenter can make. If Karpenter is forced to maintain a `minValues` of 10 for instance types and you only have 11 instance type options specified, you will only be able to consolidate from the 11th instance type down to one of the bottom 10. At the point that you consolidated down to one of the bottom 10 instance types, you would no longer be able to continue consolidating.
 
 ### How does `minValues` interact with spot-to-spot consolidation?
 
--- a/designs/nodepool-requirement-flexibility.md
+++ b/designs/nodepool-requirement-flexibility.md
@@ -1,0 +1,185 @@
+# NodePool Requirement Flexibility 
+
+**NodeClaim**: Karpenter’s representation of a Node request. A NodeClaim declaratively describes how the node should look, including instance type requirements that constrain the options that the Cloud Provider can launch with. For the AWS Cloud Provider, the instance requirements of the NodeClaim translate to the requirements of the CreateFleet request. More details can be found [here](https://github.com/aws/karpenter-provider-aws/blob/main/designs/node-ownership.md).
+
+**Preflight NodeClaim:** Karpenter’s in-memory representation of a NodeClaim that is computed during a provisioning loop. This is a NodeClaim that Karpenter hasn’t created yet CD, but is planning to launch at the end of the scheduling loop. This contains the full set of instance type possibilities which change (and become more constrained) as more pods are scheduled to the NodeClaim.
+
+## Motivation
+
+Karpenter’s scheduling algorithm uses a well-known bin-packing algorithm ([First Fit Descending](https://en.wikipedia.org/wiki/First-fit-decreasing_bin_packing)) to pack as many pods onto a set of instance type options as possible. This algorithm will necessarily continue packing pods onto Preflight Nodes so long as you have at least one instance type option that can fulfill all the pod requests.
+
+This methodology works fine for on-demand instance types, but starts to break down when you are requesting spot capacity. If you are launching spot capacity and want to ensure that you will not launch an instance that will immediately get interrupted after launch, you need to make sure that you have enough launch options in your launch request that your likelihood for launching a low-availability instance will be slim. Karpenter currently has no mechanism to ensure that today.
+
+This RFC proposes an additional key `minValues` in the NodePool requirements block, allowing Karpenter to be aware of user-specified flexibility minimums while scheduling pods to a cluster. If Karpenter cannot meet this minimum flexibility when scheduling a pod, it will fail the scheduling loop for that NodePool, either falling back to another NodePool which meets the pod requirements or failing scheduling the pod altogether.
+
+```
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+spec:
+  template:
+    spec:
+      requirements:
+        - key: karpenter.kwok.sh.instance-family
+          operator: In
+          values: ["c", "m", "r"]
+          minValues: 2
+        - key: node.kubernetes.io/instance-type
+          operator: Exists
+          minValues: 10
+```
+
+## Use-Cases
+
+1. I want to configure Karpenter to enforce more flexibility when launching spot nodes, ensuring that I am constantly using enough instance types that I get a high-availability instance type
+2. I want to increase the number of nodes that Karpenter creates by ensuring that Karpenter doesn’t over-pack pods onto nodes when it’s scheduling
+3. I want to ensure that application pods scheduling to this NodePool *must* be flexible to different instance types, architectures, etc.
+
+## Solutions
+
+### [Recommended] Solution 1: `minValues` field in the `spec.requirements`  section of NodeClaim Template (becomes part of the template on the NodePool)
+
+```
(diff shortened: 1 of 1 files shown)


designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+spec:
+  template:
+    spec:
+      requirements:
+        - key: karpenter.kwok.sh.instance-family
+          operator: In
+          values: ["c", "m", "r"]
+          minValues: 2
```
sftim, line 25:
```suggestion
          preferDistinctValues:
            - count: 2
              operator: Gt
```

or

```suggestion
          minDistinctValues: 2
```

I would use the field name to emphasize that what matters is how many distinct values there are.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+      - key: karpenter.kwok.sh.instance-family
+        operator: In
+        values: ["c", "m", "r"]
+        minValues: 2
+      - key: node.kubernetes.io/instance-type
+        operator: Exists
+        minValues: 10
+```
+
+We would add a new `minValues` field to the requirements section of the NodeClaim template. By adding this section to the NodeClaim template, we inform the scheduler that it should stop continuing to bin-pack pods onto Preflight NodeClaims and should, instead, create another Preflight NodeClaim that will schedule that pod.
```
sftim, line 57: If I have zero nodes and try to launch a new node, does that new node satisfy `minValues: 2`? If not: does Karpenter avoid using the NodePool at all (maybe there needs to be a fallback NodePool to handle a mass node failure?)

&mdash; or: does Karpenter launch a second node to satisfy the NodePool `.spec`? If not, why not? “Minimum values” implies this.

(I think the proposed field name is not quite right)

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+spec:
+  template:
+    spec:
+      requirements:
+      - key: karpenter.kwok.sh/instance-family
+        operator: In
+        values: ["c5", "r3", "m5"]
+        flexibility: 2
```
sftim, line 108: or:
```diff
-        flexibility: 2
+        distinctValueFlexibility:
+          - count: 2
+            operator: Eq
```

`operator` gets used twice, but for the same meaning in a different context.

sftim (commented):
My thoughts

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+### How does `minValues` interact with spot-to-spot consolidation?
+
+[Spot-to-spot consolidation already requires a spot instance flexibility of 15](https://github.com/kubernetes-sigs/karpenter/blob/main/designs/spot-consolidation.md). If you do not specify any `minValues` across any of your requirements in your NodePool, that 15 instance type minimum will still be required and fulfilled.
+
+If a user has `minValues` set in their NodePool that extend the number of instance types that would be needed for a spot-to-spot consolidation beyond 15, that number of instance types would be maintained. For instance, if a user had the following requirements:
+
+```
+- key: node.kubernetes.io/instance-type
+  operator: Exists
+  minValues: 30
```
Bryce-Soghigian, line 162: Seems like this behavior is cumbersome to define for each individual nodepool. I am not going to want spot thrashing on any nodepool and will just copy paste the configuration everywhere most likely.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Solution 2: `minValues` in NodeClass API
+
+```
+apiVersion: karpenter.sh/v1beta1
+kind: EC2NodeClass
+spec:
+  minValues:
+    node.kubernetes.io/instance-type: 10
+    karpenter.kwok.sh.instance-family: 2
```
Bryce-Soghigian, line 72: how do defaults come into play for these values? requirements are defined by each individual cloud provider so an implementation outside of nodeclass becomes akward when talking about defaulting. The reason being you would need karpenter to be aware of the requirements keys. 

Alertnatively you could only allow specifying "core" requirements selectors which ALL Providers would need to implement, we all can share instance type, and instance family and cpu count and gpu count etc. 

Then you can default the casing for spot that for spot pools we want to default a certain number of instance types to avoid your thrashing case. 

I also like defaulting because we dont have to specify values everywhere, and the idea behind AKS's approach for nap is we are attempting to set smart defaults for the user. Ideally one `OOB` would get a great experience and wouldn't have to configure these values themselves. 

Karpenter should provide the flexibility to be more restrictive or have better control without burdening users with that same control.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+          values: ["c", "m", "r"]
+          minValues: 2
+        - key: node.kubernetes.io/instance-type
+          operator: Exists
+          minValues: 10
+```
+
+## Use-Cases
+
+1. I want to configure Karpenter to enforce more flexibility when launching spot nodes, ensuring that I am constantly using enough instance types that I get a high-availability instance type
```
Bryce-Soghigian, line 33: Thinking about flexibility in launching nodes, or replacing nodes is there a place for minimum discount for consolidation? User may value the nodes just existing and not cycling up and down and paying a bit more, but might have a threshold where the cost makes the disruption worth it.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Pros/Cons
+
+1. 👍👍 Isolates the minValues feature to whether a Cloud Provider wants to implement it or not
+2. 👍 Enables Cloud Providers to have the concept of “default” minValues that is controlled at runtime, allowing cloud providers to enforce certain minimums from their users for things like spot best practices without having to have the user specify it directly (e.g. “don’t allow a spot instance to launch without at least 10 spot instances in the request”).
+3. 👎 Leaks more scheduling behavior into the NodeClass (What if a Cloud Provider doesn’t implement a NodeClass directly? They would have to design one specifically for this feature). As a guiding design principle, we have generally chosen to avoid scheduler-based decision making driven through the NodeClass API
```
Bryce-Soghigian, line 93: i would do a double thumbs down on this

designs/nodepool-requirement-flexibility.md
Bryce-Soghigian: See now you talked about it in the cons

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+Enabling this solution allows specific Cloud Providers to make this change without pushing this new field down into the neutral API; however, this means we need to define a neutral way where a cloudprovider can define its own flexibility requirements and pass them back to the scheduler. One option here is to allow CloudProviders to layer on their own flexibility requirements for NodePools through the [CloudProvider interface](https://github.com/kubernetes-sigs/karpenter/blob/4e85912c81ed9a51fc8ebd107db8e88e7828fae7/pkg/cloudprovider/types.go). This function in the interface might look like `func GetFlexibility(np *v1beta1.NodePool) scheduling.Flexibility` where `scheduling.Flexibility` is a `map[string]int`.
+
+#### Pros/Cons
+
+1. 👍👍 Isolates the minValues feature to whether a Cloud Provider wants to implement it or not
+2. 👍 Enables Cloud Providers to have the concept of “default” minValues that is controlled at runtime, allowing cloud providers to enforce certain minimums from their users for things like spot best practices without having to have the user specify it directly (e.g. “don’t allow a spot instance to launch without at least 10 spot instances in the request”).
```
Bryce-Soghigian, line 92: also sharing a nodeclass across all nodepools makes it easy to have defaults just applied. Configure it once and share that nodeclass everywhere if you aren't too picky about some of the other settings. 

The downside is nodeclass also provides a lot of settings that make sense to be different. I want multiple image families for example. So sharing one default value across in many places with nodeclass becomes tricky

designs/nodepool-requirement-flexibility.md
jonathan-innis: > I am not going to want spot thrashing on any nodepool and will just copy paste the configuration everywhere most likely

It seems like there is utility here in defining different constraints between on-demand NodePools and spot NodePools. On top of that, I think there's value here in allowing differences in minValues requirements when it comes to how tolerant your workloads are to interruption and how aggressive you are willing to be with bin-packing a bunch of nodes on a single, really large instance.

designs/nodepool-requirement-flexibility.md
jonathan-innis: I think this breaks down for the reasons described in the doc. It's either: Impossible to compute, Too hard to compute, or A bad decision to compute strict flexibility vs. a minimum flexibility constraint.

designs/nodepool-requirement-flexibility.md
jonathan-innis: > “Minimum values” implies this.

I'm not I quite follow. This field name is talking about the minimum values that are required for a request of a NodeClaim, not the number of values that need to be satisfied across nodes within a NodePool. For that, it seems like we need a concept of `topologySpreadConstraints` on the NodePool to tell Karpenter, "give me something from this set of instance types, make sure that my request has at least two instance types in it, and I also want to make sure that you spread me across instance types with a minDomains of 3"

designs/nodepool-requirement-flexibility.md
jonathan-innis: Isn't this implied? If my values are: "a", "b", "c", then I feel like asking for a minimum 2 of those values implies that they are distinct.

designs/nodepool-requirement-flexibility.md
jonathan-innis: I could see this potentially being hierarchical here. If we ever felt a need for global configuration we could always say that that could be configured and if you configure anything more granular then that overrides the global configuration.

I think we've generally pushed back on global configuration, though. Because you can achieve the same thing by just templating out your manifests through IaC and it isn't truly that cumbersome to have to respecify the same data across manifests if you are using something like helm or kustomize.

designs/nodepool-requirement-flexibility.md
jonathan-innis: >  there a place for minimum discount for consolidation

Yes! I think this is something that we were considering as part of the spot-to-spot consolidation design: https://github.com/kubernetes-sigs/karpenter/blob/main/designs/spot-consolidation.md?plain=1#L32. We ended up seeing this as orthogonal, but I still think that it's something that we want to do so that users can say I only want to be interrupted if it's going to give me _this_ much cost savings.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+### How does `minValues` interact with consolidation?
+
+The scheduling simulation should return a result from consolidation that respects `minValues` consistent with the launch behavior. Karpenter will then reduce the set of instance types here to be strictly cheaper than the combined price of all instance types that we plan to replace (this is our `filterByPrice` function).
+
+It’s possible that after Karpenter reduces these instance types down to only cheaper ones, it no longer fulfills the `minValues` criteria of the NodePool requirements. As a result, consolidation will need to re-check whether the `minValues` is met after the instance types are filtered down by price to ensure that we can proceed with the launch. If the `minValues` is not satisfied, we should reject the consolidation decision.
```
njtran, line 151: Can we call out that with MinValues and an Operator == In, users should define a set of values with a cardinality larger than the MinValues? They essentially can never get replace actions if they have strict equality here.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+If a user has `minValues` set in their NodePool that require less than 15 instance types, then the 15 instance type minimum will continue to be enforced to enable a spot-to-spot consolidation to occur. For instance, if a user had the following requirements:
+
+```
+- key: node.kubernetes.io/instance-type
+  operator: Exists
+  minValues: 3
+```
+
+In this case, we would schedule with at least 3 instance types and then validate that the result allowed for 15 cheaper instance types that we could consolidate to. If there were not 15 cheaper instance types, then we would not perform spot-to-spot consolidation.
```
njtran, line 175: Just sanity check here: Are you saying that we'd check if we can consolidate first and see if there are 15 instance types, and then after the fact, check to make sure that at least 3 of those instance types are within the requirements of the nodePool? Sounds like we'd have to add a flexibility check during the scheduling simulation and a flexibility check after the results parsing

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+### How does `minValues` interact with Drift?
+
+Karpenter currently performs dynamic checking on drift against the NodeClaim labels. Optionally, we could check `minValues` against the `node.kubernetes.io/instance-type` values that are stored in the NodeClaim. Practically, this is probably not necessary as users that want us to respect `minValues` really only care about launch criteria and the nodes on the cluster have already been launched.
+
+**This is also a fast-follow decision that we could implement if we got an ask for supporting drift with this new `minValues` field.**
```
njtran, line 181: Agreed here, I don't think we need to drift on it, but also makes me think if that means this is actually the best place for the configuration to live. It's really a behavioral field, but is super tied to scheduling, which is why we're putting this in the scheduling loop.

njtran (commented):
Overall, good design, we've already synced on most of my thoughts in the doc, but just posting them here for posterity

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Pros/Cons
+
+1. 👍👍 Isolates the minValues feature to whether a Cloud Provider wants to implement it or not
+2. 👍 Enables Cloud Providers to have the concept of “default” minValues that is controlled at runtime, allowing cloud providers to enforce certain minimums from their users for things like spot best practices without having to have the user specify it directly (e.g. “don’t allow a spot instance to launch without at least 10 spot instances in the request”).
+3. 👎 Leaks more scheduling behavior into the NodeClass (What if a Cloud Provider doesn’t implement a NodeClass directly? They would have to design one specifically for this feature). As a guiding design principle, we have generally chosen to avoid scheduler-based decision making driven through the NodeClass API
```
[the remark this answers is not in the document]
jonathan-innis: I could see us relaxing this requirement today. Realistically, if the CloudProvider just has a single way that nodes can be provisioned, they may not need any NodeClass with any configuration at all. We don't have any tangible examples of this beyond Kwok but I could see a use-case for defaulting all of the configuration and only having the NodePool around to configure what types of instance types that you want; Karpenter figures out the rest.

designs/nodepool-requirement-flexibility.md
jonathan-innis: > They essentially can never get replace actions if they have strict equality here

That's assuming that they are setting `minValues` on the instance type, but this `minValues` requirement could apply to other things like architecture or capacity type.

In either case, I can agree that it may be worth calling out that `minValues` shouldn't be so high that it unnecessarily  constrains consolidation in a way that it can't actually do anything on the cluster for cost savings.

designs/nodepool-requirement-flexibility.md
jonathan-innis: I know we disagree here, but I still think that this is the best place for the scheduling behavior to live, despite the fact that it's in the requirements block, which we drift on for labels. I could see us drifting on the minValues requirements if the instance types in the NodeClaim that we used for the original launch no longer meet the requirements that we are setting.

Realistically, though, this is really of a continuity thing vs. real value for users.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+          minValues: 2
+        - key: node.kubernetes.io/instance-type
+          operator: Exists
+          minValues: 10
+```
+
+## Use-Cases
+
+1. I want to configure Karpenter to enforce more flexibility when launching spot nodes, ensuring that I am constantly using enough instance types that I get a high-availability instance type
+2. I want to increase the number of nodes that Karpenter creates by ensuring that Karpenter doesn’t over-pack pods onto nodes when it’s scheduling
```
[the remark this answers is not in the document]
jonathan-innis: I agree, it's definitely not a strong benefit. I can drop the use-case from the doc.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+```
+
+We would add a new `minValues` field to the requirements section of the NodeClaim template. By adding this section to the NodeClaim template, we inform the scheduler that it should stop continuing to bin-pack pods onto Preflight NodeClaims and should, instead, create another Preflight NodeClaim that will schedule that pod.
+
+#### Pros/Cons
+
+1. 👍👍 Puts all core scheduling behavior together in the same block, allowing users to easily reason about what their requirements and minimums mean
+2. 👎 Piggy-backs off of a well-known core API. There’s risk that if this API changes in a way that is incompatible with this value in the future, we may have to break our `minValues` API
+
+### Solution 2: `minValues` in NodeClass API
```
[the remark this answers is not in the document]
jonathan-innis: Called it out.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+
+#### Pros/Cons
+
+1. 👍👍 Isolates the minValues feature to whether a Cloud Provider wants to implement it or not
+2. 👍 Enables Cloud Providers to have the concept of “default” minValues that is controlled at runtime, allowing cloud providers to enforce certain minimums from their users for things like spot best practices without having to have the user specify it directly (e.g. “don’t allow a spot instance to launch without at least 10 spot instances in the request”).
+3. 👎 Leaks more scheduling behavior into the NodeClass (What if a Cloud Provider doesn’t implement a NodeClass directly? They would have to design one specifically for this feature). As a guiding design principle, we have generally chosen to avoid scheduler-based decision making driven through the NodeClass API
+4. 👎 Requires an additional method on the cloud providers OR requires piggy-backing off of `GetInstanceTypes()` to surface additional minValues information into the scheduler which may be awkward for cloud providers that don’t care about this requirement
```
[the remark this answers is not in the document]
jonathan-innis: It expands our interface and makes cloud providers have to think about one more thing that they may or may not care about. I think having to reason about a new concept that you don't care about is sufficient enough to call this a con here.

designs/nodepool-requirement-flexibility.md
```diff
@@ -0,0 +1,185 @@
...
+        - key: node.kubernetes.io/instance-type
+          operator: Exists
+          minValues: 10
+```
+
+## Use-Cases
+
+1. I want to configure Karpenter to enforce more flexibility when launching spot nodes, ensuring that I am constantly using enough instance types that I get a high-availability instance type
+2. I want to increase the number of nodes that Karpenter creates by ensuring that Karpenter doesn’t over-pack pods onto nodes when it’s scheduling
+3. I want to ensure that application pods scheduling to this NodePool *must* be flexible to different instance types, architectures, etc.
```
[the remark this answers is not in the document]
njtran: This mostly really just is for architecture, right? I guess maybe instance types too, but realistically there shouldn't be any compatible difference in terms of getting pods healthy with capacity type.

designs/nodepool-requirement-flexibility.md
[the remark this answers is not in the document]
njtran: Sure, I'm just thinking that it's not user-facing. It's a cp implementation detail, which when I have a function that has some unknown behavior that I don't care about, if I know I can just return nil, it's an almost negligent con

designs/nodepool-requirement-flexibility.md
njtran: Sure thang. I'll just make sure it's known I disagree here!

designs/nodepool-requirement-flexibility.md
jonathan-innis: That's what I'm saying, yes. For consolidation, because we filter out cheaper instance types, we have to perform the scheduling simulation, then filter out the cheaper instance types from the result, then check that we still have 15 instance types and that we are maintaining the instance type flexibility here.

designs/nodepool-requirement-flexibility.md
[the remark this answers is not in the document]
jonathan-innis: Not sure that I agree. There's a burden here for a cloudprovider that is onboarding to Karpenter to have to think about what this thing means and determine whether I want or don't want to implement it. That's part of the reason that we keep the interface small -- keeps the burden for implementation lower and helps other cloud providers come an implementation faster.

If the method is part of the interface, I'm going to necessarily assume that I need to implement it. Ideally, I don't want to have to go deep into the code to understand what it's doing and to understand whether I do or don't need it.

designs/nodepool-requirement-flexibility.md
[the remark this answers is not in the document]
jonathan-innis: I think the use-case is definitely weaker but seems worthwhile to callout here.

designs/nodepool-requirement-flexibility.md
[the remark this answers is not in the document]
Bryce-Soghigian: > There's a burden here for a cloudprovider that is onboarding to Karpenter to have to think about what this thing means and determine whether I want or don't want to implement it.
Big +1. Its not obvious that a method can be optional. When implementing cloud provider without a good understanding of core, one will probably look at existing examples like the golden standard for karpenter cloudproviders(AWS provider). 

If the team is paid to spend lots of time doing research and figuring out the specifics like azure it won't be a problem, but if Its a community project/implementation reducing the things  one has to do to get started will make the completion of the project more likely. 

Most non GKE, AWS, Azure providers in the Cluster Autoscaler space are driven by the community and their free time. If the goal is to give space to these cloud providers we need to make it as easy as possible for them to get started.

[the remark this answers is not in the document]
Bryce-Soghigian: Maybe there is something we can do to make it easier to know which bits are optional though.

njtran (commented):
/lgtm
/approve

PR #975, opened 2024-01-28
feat: add nodeclaim name into orchestration initialization failures

njtran:
Fixes #N/A 

**Description**
This adds the nodeclaim name into the log. Previously this error could return something like the following, which gives no information about which node claims it was expecting to be initialized.

```
failed to disrupt nodes, command reached timeout after {duration}; waiting for replacement initialization, node claim not initialized; node claim not initialized; node claim not initialized; node claim not initialized
```

make presubmit


jonathan-innis (approved):
/lgtm
/approve

PR #980, opened 2024-01-30
feat: add uid for orchestration queue for logs

njtran:
Fixes #N/A 

**Description**
Adds a types.UID that's attached to a command in memory. This allows us to log this uid on command creation, and track it when the command eventually fails or succeeds.

make apply

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -185,8 +185,8 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command) 
 		methodLabel:            m.Type(),
 		consolidationTypeLabel: m.ConsolidationType(),
 	}).Inc()
-	uid := uuid.NewUUID()
-	logging.FromContext(ctx).With("command-uid", uid).Infof("disrupting via %s %s", m.Type(), cmd)
+	commandID := uuid.NewUUID()
+	logging.FromContext(ctx).With("command-id", commandID).Infof("disrupting via %s %s", m.Type(), cmd)
 
 	stateNodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode {
 		return c.StateNode
@@ -211,7 +211,7 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command) 
 	c.cluster.MarkForDeletion(providerIDs...)
 
 	if err := c.queue.Add(orchestration.NewCommand(nodeClaimNames,
-		lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode }), uid, m.Type(), m.ConsolidationType())); err != nil {
+		lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode }), commandID, m.Type(), m.ConsolidationType())); err != nil {
 		c.cluster.UnmarkForDeletion(providerIDs...)
 		return fmt.Errorf("adding command to queue, %w", multierr.Append(err, state.RequireNoScheduleTaint(ctx, c.kubeClient, false, stateNodes...)))
 	}
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/samber/lo"
 	"go.uber.org/multierr"
+	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/utils/clock"
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -184,7 +185,8 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command) 
 		methodLabel:            m.Type(),
 		consolidationTypeLabel: m.ConsolidationType(),
 	}).Inc()
-	logging.FromContext(ctx).Infof("disrupting via %s %s", m.Type(), cmd)
+	uid := uuid.NewUUID()
+	logging.FromContext(ctx).With("uid", uid).Infof("disrupting via %s %s", m.Type(), cmd)
 
 	stateNodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode {
 		return c.StateNode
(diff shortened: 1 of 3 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -184,7 +185,8 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command)
 		methodLabel:            m.Type(),
 		consolidationTypeLabel: m.ConsolidationType(),
 	}).Inc()
-	logging.FromContext(ctx).Infof("disrupting via %s %s", m.Type(), cmd)
+	uid := uuid.NewUUID()
```
jonathan-innis, line 188: Should we just inject this into the logger generally at the start of the loop and then just pass it through any command execution that we get. That also means that you get any associative logging that happens in scheduling for free

jonathan-innis, line 188: Did we have a chance to look at controller-runtime to see if we could leverage their reconcile id rather than having to create our own concept?

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -202,10 +204,11 @@ func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.R
 		// Log the error
 		logging.FromContext(ctx).With("nodes", strings.Join(lo.Map(cmd.candidates, func(s *state.StateNode, _ int) string {
 			return s.Name()
-		}), ",")).Errorf("failed to disrupt nodes, %s", multiErr)
+		}), ","), "uid", cmd.uid).Errorf("failed to disrupt nodes, %s", multiErr)
```
jonathan-innis, line 207:
```suggestion
		}), ","), "reconcile-id", cmd.uid).Errorf("failed to disrupt nodes, %s", multiErr)
```

Is this value really just a reconciliation id or we could even brand it as the disruption-id?

pkg/controllers/disruption/controller.go
njtran: If you pass it through the context, you don't get it in the orchestration queue. So you could, but it would just be redundant anyway, since this is the only log in this function.

pkg/controllers/disruption/orchestration/queue.go
njtran: I'm not a fan of reconcile-id, as it insinuates it's the same reconcile. Which do you prefer?

pkg/controllers/disruption/controller.go
njtran: synced offline, going to punt this for a future refactor. At first look, I couldn't find an easy reference to the controller-runtime reconcile id, but for within disruption/orchestration, we're thinking it makes less sense for the Singleton controller abstraction.

pkg/controllers/disruption/orchestration/queue.go
njtran: synced offline, went with command-uid

pkg/controllers/disruption/controller.go
```diff
@@ -184,7 +185,8 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command)
...
+	logging.FromContext(ctx).With("command-uid", uid).Infof("disrupting via %s %s", m.Type(), cmd)
```
jonathan-innis, line 189:
```suggestion
	logging.FromContext(ctx).With("command-id", uid).Infof("disrupting via %s %s", m.Type(), cmd)
```

nit: I would just call it `id`, that's typically more common because the fact that it's unique is implied

pkg/controllers/disruption/controller.go
jonathan-innis, line 188:
```suggestion
	commandID := uuid.NewUUID()
```

nit: Call it commandID and reference it as that throughout?

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -57,6 +57,7 @@ type Command struct {
 	Replacements      []Replacement
 	candidates        []*state.StateNode
 	timeAdded         time.Time // timeAdded is used to track timeouts
+	uid               types.UID // used for log tracking
```
jonathan-innis, line 60:
```suggestion
	id               types.UID // used for log tracking
```

Is this just an `id` or `commandID` and the fact that it's a UID is an implementation detail?

njtran: I thought commandID would stutter, since it would be command.commandID. I'm happy to do id though

pkg/controllers/disruption/controller.go
```diff
@@ -184,7 +185,8 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command)
 		methodLabel:            m.Type(),
 		consolidationTypeLabel: m.ConsolidationType(),
 	}).Inc()
-	logging.FromContext(ctx).Infof("disrupting via %s %s", m.Type(), cmd)
+	commandID := uuid.NewUUID()
+	logging.FromContext(ctx).With("command-id", commandID).Infof("disrupting via %s %s", m.Type(), cmd)
 
 	stateNodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode {
 		return c.StateNode
```
jonathan-innis, line 192: Do any of the errors that are fired in this block also need to have the `command-id` here?

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -202,10 +204,11 @@ func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.R
 		// Log the error
 		logging.FromContext(ctx).With("nodes", strings.Join(lo.Map(cmd.candidates, func(s *state.StateNode, _ int) string {
 			return s.Name()
-		}), ",")).Errorf("failed to disrupt nodes, %s", multiErr)
+		}), ","), "command-id", cmd.id).Errorf("failed to disrupt nodes, %s", multiErr)
```
jonathan-innis, line 207: Add this command-id to the context by injecting it into the logger at the top of the reconciliation loop?

jonathan-innis: Might make it easier so that we don't forget adding it to each call that we have to log

pkg/controllers/disruption/controller.go
njtran: I added it in the error log lines. Lmk what you think

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -177,6 +179,7 @@ func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.R
 		panic("unexpected failure, disruption queue has shut down")
 	}
 	cmd := item.(*Command)
+	ctx = context.WithValue(ctx, "command-id", cmd.id)
```
jonathan-innis, line 182:
```suggestion
	ctx = logging.WithLogger(logging.FromContext(ctx).With("command-id", cmd.id))
```

jonathan-innis (approved):
/lgtm
/approve

PR #982, opened 2024-01-30
perf: increase scheduler performance

tzneal:
Fixes #N/A 

**Description**

A few changes that overall increase performance 10-60% depending on the number of pods/nodes. Its around a ~34% overall improvement:
```
name               old pods/sec  new pods/sec  delta
Scheduling1-12       3.89k ± 2%    4.36k ± 4%  +12.10%  (p=0.000 n=10+9)
Scheduling50-12      1.44k ±11%    1.70k ± 5%  +18.29%  (p=0.000 n=9+9)
Scheduling100-12     1.30k ±10%    2.09k ± 3%  +60.63%  (p=0.000 n=9+9)
Scheduling500-12     1.47k ± 3%    2.06k ± 1%  +40.65%  (p=0.000 n=8+9)
Scheduling1000-12    1.42k ± 5%    1.99k ± 1%  +39.32%  (p=0.000 n=10+10)
Scheduling2000-12    1.43k ± 1%    1.83k ± 2%  +28.36%  (p=0.000 n=9+9)
Scheduling5000-12    1.09k ± 6%    1.67k ± 2%  +53.75%  (p=0.000 n=9+10)
```

Benchmarking and unit testing/

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
+++ b/pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
@@ -56,10 +56,22 @@ const PrintStats = false
 //nolint:gosec
 var r = rand.New(rand.NewSource(42))
 
+// To run the benchmarks use:
+// `go test -tags=test_performance -run=XXX -bench=.`
+//
+// to get something statistically significant for comparison we need to run them several times and then
+// compare the results between the old performance and the new performance.
+// ```sh
+//
+//	go test -tags=test_performance -run=XXX -bench=. -count=10 | tee /tmp/old
+//	# make your changes to the code
+//	go test -tags=test_performance -run=XXX -bench=. -count=10 | tee /tmp/new
+//	benchstat /tmp/old /tmp/new
+//
+// ```
 func BenchmarkScheduling1(b *testing.B) {
 	benchmarkScheduler(b, 400, 1)
 }
-
 func BenchmarkScheduling50(b *testing.B) {
 	benchmarkScheduler(b, 400, 50)
 }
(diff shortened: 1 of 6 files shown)


pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -56,10 +56,22 @@ const PrintStats = false
 //nolint:gosec
 var r = rand.New(rand.NewSource(42))
 
+// To run the benchmarks use:
```
jonathan-innis, line 59: Nice 🎉

pkg/scheduling/requirements.go
```diff
@@ -237,9 +237,38 @@ func labelHint(r Requirements, key string, allowedUndefined sets.Set[string]) st
 	return ""
 }
 
+// badKeyError allows lazily generating the error string in the case of a bad key error. When requirements fail
+// to match, we are most often interested in the failure and not why it fails.
+type badKeyError struct {
```
jonathan-innis, line 242: It's surprising to me that templating out those error values into a string is so expensive. What's the performance hit that we take from not taking this change in and just returning errors like normal?

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -256,10 +256,7 @@ func (t *TopologyGroup) nextDomainAntiAffinity(domains *scheduling.Requirement)
 	// list of domains.  The use case where this optimization is really great is when we are launching nodes for
 	// a deployment of pods with self anti-affinity.  The domains map here continues to grow, and we continue to
 	// fully scan it each iteration.
-	if len(t.emptyDomains) == 0 {
-		return options
-	}
-	for domain := range t.domains {
+	for domain := range t.emptyDomains {
```
jonathan-innis, line 259: Nice 🎉

jonathan-innis:
Worth sharing the CPU profile that you generated from pprof as part of the PR just for posterity?

pkg/scheduling/requirements.go
tzneal: Changing the errors is a large part of the performance increase for anti-affinity pods in particular due to the repeated failures to match topologies. String interpolation is expensive and it happens a lot in these cases.

tzneal:
> Worth sharing the CPU profile that you generated from pprof as part of the PR just for posterity?

I don't think it would be that helpful. We should instead set some minimum benchmarks and enable the CI based tests for performance.

pkg/scheduling/requirements.go
jonathan-innis: Got it. This is good to know 👍🏼

jonathan-innis (approved):
/lgtm
/approve

PR #988, opened 2024-02-01
fix: allow disruption when considering past disruption commands

njtran:
Fixes #N/A 

**Description**
When considering deleting pods for scheduling simulations, we won't fail to disrupt if the pods that scheduled to uninitialized nodes were from an already deleting node. This is because they'll either be a node from a previous disruption command or a manually deleted node, either one of which we should assume is being disrupted and getting a replacement node as needed from the disruption or provisioning controller.

make presubmit

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -45,7 +45,7 @@ import (
 )
 
 //nolint:gocyclo
-func simulateScheduling(ctx context.Context, kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner,
+func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner,
 	candidates ...*Candidate,
 ) (*pscheduling.Results, error) {
 	candidateNames := sets.NewString(lo.Map(candidates, func(t *Candidate, i int) string { return t.Name() })...)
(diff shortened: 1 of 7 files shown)


pkg/controllers/disruption/helpers.go
```diff
@@ -119,7 +119,18 @@ func simulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
...
 			for _, p := range n.Pods {
-				results.PodErrors[p] = fmt.Errorf("would schedule against a non-initialized node %s", n.Name())
+				// Only add a pod scheduling error if it isn't on an already deleting node.
+				// If the pod is on a deleting node, we assume one of two things has already happened:
+				// 1. The node was manually terminated, at which the provisioning controller has scheduled or is scheduling a node
+				//    for the pod.
+				// 2. The node was chosen for a previous disruption command, we assume that the uninitialized node will come up
+				//    for this command, and we assume it will be successful. If it is not successful, the node will become
+				//    not terminating, and we will no longer need to consider these pods.
+				if _, ok := lo.Find(deletingNodePods, func(deleting *v1.Pod) bool {
```
jonathan-innis, line 129: How slow is this if you have a lot of deleting pods? Is it worth making this check faster, by just quickly storing the keys once and then just checking if the keys exists or not?

pkg/controllers/disruption/suite_test.go
```diff
@@ -150,6 +150,82 @@ var _ = AfterEach(func() {
 	ExpectCleanedUp(ctx, env.Client)
 })
 
+var _ = Describe("Simulate Scheduling", func() {
```
jonathan-innis, line 153: I think this test is definitely good. Is there room for ensuring that we can execute concurrent replace requests with our budgets by checking that we don't block on deleting pods scheduling to uninitialized nodes as well?

njtran: Yeah that's a good idea

pkg/controllers/disruption/suite_test.go
```diff
@@ -150,6 +150,187 @@ var _ = AfterEach(func() {
...
+		Expect(len(ncs)).To(Equal(13))
+		nc, new = lo.Find(ncs, func(nc *v1beta1.NodeClaim) bool {
+			_, ok := nodeClaimNames[nc.Name]
+			return !ok
+		})
+		Expect(new).To(BeTrue())
+		ExpectNodeClaimDeployed(ctx, env.Client, cluster, cloudProvider, nc)
+		nodeClaimNames[nc.Name] = struct{}{}
+
+		// Try one more time, but fail since the budgets only allow 3 disruptions.
```
jonathan-innis, line 324: Nice 🎉

pkg/test/expectations/expectations.go
```diff
@@ -361,6 +361,16 @@ func ExpectMakeNodeClaimsInitialized(ctx context.Context, c client.Client, nodeC
 	}
 }
 
+func ExpectMakeNodeClaimsLaunchedAndStateUpdated(ctx context.Context, c client.Client, nodeClaimStateController controller.Controller, nodeClaims ...*v1beta1.NodeClaim) {
```
jonathan-innis, line 364: I don't see this being used anywhere. Do you still need it in this PR?

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

PR #998, opened 2024-02-09
fix: Use UUID as a precondition when calling the eviction API

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that when we attempt to evict a pod, we don't accidentally evict a pod with a namespaced name that was recently re-created and shifted onto another node. This is particularly critical for StatefulSets, where pods can be re-created quickly with the same name. 

`make presubmit`

### DIFF
--- a/pkg/controllers/node/termination/terminator/eviction.go
+++ b/pkg/controllers/node/termination/terminator/eviction.go
@@ -22,13 +22,13 @@ import (
 	"fmt"
 	"time"
 
-	set "github.com/deckarep/golang-set"
+	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/workqueue"
 	"knative.dev/pkg/logging"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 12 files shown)


pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -102,53 +115,61 @@ func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.R
...
 }
 
 // Evict returns true if successful eviction call, and false if not an eviction-related error
-func (q *Queue) Evict(ctx context.Context, nn types.NamespacedName) bool {
-	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", nn))
-	if err := q.coreV1Client.Pods(nn.Namespace).EvictV1(ctx, &policyv1.Eviction{
-		ObjectMeta: metav1.ObjectMeta{Name: nn.Name, Namespace: nn.Namespace},
-	}); err != nil {
+func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", key.NamespacedName))
```
njtran, line 141: Thoughts on adding the UUID here too since that's the bug we're fixing?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -102,53 +115,61 @@ func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.R
...
+					UID: lo.ToPtr(key.UID),
+				},
+			},
+		}); err != nil {
 		// status codes for the eviction API are defined here:
 		// https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works
-		if apierrors.IsNotFound(err) { // 404
+		if apierrors.IsNotFound(err) || apierrors.IsConflict(err) {
+			// 404 - The pod no longer exists
+			// 409 - The pod exists, but it is not the same pod that we initiated the eviction on
```
njtran, line 155: I'm assuming the conflict error here is when the Preconditions aren't met?

njtran (commented):
Super clean. Nice work. Just a couple of comments

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: Not sure that I get you. You're thinking that we should add the UUID to the log? If we get a conflict response since the pod was already evicted, then we aren't doing anything on our side anyways, so I'm not sure that we should log here

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: Yep, you got it. Would it be helpful to link out where the docs state that?

jonathan-innis: Just added the link so we can see where in the api-initiated eviction this is happening.

njtran (approved):
/lgtm
/approve

PR #999, opened 2024-02-09
chore: Drop simulation mode in favor of NopLogger and Recording scheduling results

jonathan-innis:
Fixes #N/A 

**Description**

This change drops `scheduling.SimulationMode` from the code in favor of using a NopLogger for ignoring log messages and recording the scheduling results outside the context of the `Solve` method. Ideally, solve is just a common library that any component can use without having to reason about options.

`make presubmit`

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -41,6 +41,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/events"
 	"sigs.k8s.io/karpenter/pkg/metrics"
+	operatorlogging "sigs.k8s.io/karpenter/pkg/operator/logging"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
@@ -78,9 +79,7 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 		pods = append(pods, n.reschedulablePods...)
 	}
 	pods = append(pods, deletingNodePods...)
-	scheduler, err := provisioner.NewScheduler(ctx, pods, stateNodes, pscheduling.SchedulerOptions{
-		SimulationMode: true,
-	})
+	scheduler, err := provisioner.NewScheduler(logging.WithLogger(ctx, operatorlogging.NopLogger), pods, stateNodes)
 	if err != nil {
 		return pscheduling.Results{}, fmt.Errorf("creating scheduler, %w", err)
 	}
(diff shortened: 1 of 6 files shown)


pkg/controllers/disruption/helpers.go
```diff
@@ -78,9 +79,7 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 		pods = append(pods, n.reschedulablePods...)
 	}
 	pods = append(pods, deletingNodePods...)
-	scheduler, err := provisioner.NewScheduler(ctx, pods, stateNodes, pscheduling.SchedulerOptions{
-		SimulationMode: true,
-	})
+	scheduler, err := provisioner.NewScheduler(logging.WithLogger(ctx, operatorlogging.NopLogger), pods, stateNodes)
```
njtran, line 82: Seems like you don't need the context elsewhere in this function for logging, so you could just set the loggin.WithLogger() for the context and pass it in here for L82 and L91 without having to do it two times

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -105,6 +95,45 @@ type Results struct {
...
 }
 
+// Record sends eventing and log messages back for the results that were produced from a scheduling run
+// It also nominates nodes in the cluster state based on the scheduling run to signal to other components
+// leveraging the cluster state that a previous scheduling run that was recorded is relying on these nodes
+func (r Results) Record(ctx context.Context, recorder events.Recorder, cluster *state.Cluster) {
+	// Report failures and nominations
+	for p, err := range r.PodErrors {
+		logging.FromContext(ctx).With("pod", client.ObjectKeyFromObject(p)).Errorf("Could not schedule pod, %s", err)
+		recorder.Publish(PodFailedToScheduleEvent(p, err))
```
njtran, line 105: Looks like now we'll be emitting these events on scheduling simulations in disruption, which doesn't really make sense since we'd be emitting these events for pods that are already scheduled. We're really trying to just reference the re-created replica, so I don't think we want this event here.

njtran (commented):
Looks good, just have one comment on a pod event that we're now emitting

pkg/controllers/disruption/helpers.go
jonathan-innis: I'd personally rather set it directly for now. If we were to add changes in later that started to log things, I could see it being potentially confusing for someone trying to reason about why they aren't getting any log output from the function call.

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: Good callout. I was thinking that we would never call this for any pods during the disruption loop, but I'm realizing that it's possible for us to make a call to this function with pod errors for pending pods that will never actually schedule to nodes.

jonathan-innis: > which doesn't really make sense since we'd be emitting these events for pods that are already scheduled

Why would we emit failure to schedule if we made a scheduling decision? If we are making a disruption decision, it's because we didn't get any errors for scheduling currently scheduled pods. The only errors that we would get in here would be for pending pods that didn't schedule during the last scheduling round either, so I think we are fine to fire events and logs here.

njtran: We'd be firing events here for any mirror or daemonset pods https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/utils/pod/scheduling.go#L87-L88, which I don't think we need to.

jonathan-innis: We don't schedule either of these pods. They aren't part of our reschedulable pod set

njtran (commented):
/lgtm
/approve

PR #1006, opened 2024-02-13
fix: Fix budget validation through CEL

### LINKED ISSUE #1003: NodePool - only partly schedule and duration rule
### Description

**Observed Behavior**:

This x-validation-rule https://github.com/kubernetes-sigs/karpenter/blob/0ac0ff583d940681c205d21617f4443035e293ac/pkg/apis/crds/karpenter.sh_nodepools.yaml#L112
seems to only work if there is only one budget with duration or schedule. As soon as there is at least one budget/entry with a correct schedule+duration set, other budgets/entries can be set with only one of the fields.

I would propose: `rule: 'self.all(x, has(x.schedule) == has(x.duration))'`

**Expected Behavior** + **Reproduction Steps** (Please include YAML):
this example should fail, but it doesn't:
```yaml
kind: NodePool
metadata:
  name: test
spec:
  # ...
  disruption:
    budgets:
    - nodes: "5"
      duration: 100m
    - nodes: "0"
      schedule: "@daily"
      duration: 10m
```

**Versions**:
- Chart Version:
- Kubernetes Version (`kubectl version`): 1.28

jonathan-innis:
Fixes #1003

**Description**

Budget validation for CEL wasn't properly catching the case where one budget had properly set the schedule and duration together but another hadn't. This change ensures that _every_ budget has set a valid combination of schedule and duration

`make presubmit`

### DIFF
--- a/pkg/apis/crds/karpenter.sh_nodepools.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodepools.yaml
@@ -109,7 +109,7 @@ spec:
                       type: array
                       x-kubernetes-validations:
                         - message: '''schedule'' must be set with ''duration'''
-                          rule: '!self.all(x, (has(x.schedule) && !has(x.duration)) || (!has(x.schedule) && has(x.duration)))'
+                          rule: self.all(x, has(x.schedule) == has(x.duration))
                     consolidateAfter:
                       description: |-
                         ConsolidateAfter is the duration the controller will wait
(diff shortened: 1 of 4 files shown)


pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -109,7 +109,7 @@ spec:
                       type: array
                       x-kubernetes-validations:
                         - message: '''schedule'' must be set with ''duration'''
-                          rule: '!self.all(x, (has(x.schedule) && !has(x.duration)) || (!has(x.schedule) && has(x.duration)))'
+                          rule: self.all(x, has(x.schedule) == has(x.duration))
```
engedaam, line 112: Shouldn't this be an `&&` and not an equal?

jonathan-innis: `==` means they match. That's all that we require. Either they are both set or they are both not set.

PR #1008, opened 2024-02-14
ci: Drop `--ginkgo.timeout` and add `--ginkgo.randomize-all` CLI arg for Makefile target

jonathan-innis:
Fixes #N/A 

**Description**

From reviewing the code and testing, it isn't necessary to specify both of `--ginkgo.timeout` and `--timeout`. This drops the former to prefer the suite-based timeout with `go test`

This change also adds the `--ginkgo.randomize-all` CLI flag to ensure that we aren't missing any test flakiness due to implicit ordering in our testing.

`make presubmit`


engedaam (commented):
/lgtm

PR #1011, opened 2024-02-15
fix: Ensure that cluster_state_synced metric gets set if function exits early

jonathan-innis:
Fixes #N/A 

**Description**

Ensure that we return the right value for `cluster_state_synced` even if the function exits early

`make presubmit`


Bryce-Soghigian (approved):

engedaam (commented):
/lgtm

PR #1012, opened 2024-02-15
fix: Fix an issue where the instanceTypes aren't sorted before truncating it for spot-to-spot consolidation.

nikmohan123:
Fix an issue where the instanceTypes aren't sorted before truncating it for spot-to-spot consolidation.

**Description**

As part of spot-to-spot consolidation algorithm (ref: [here](https://github.com/kubernetes-sigs/karpenter/pull/768/files)), we check for minimum flexibility of the instance types and if it is greater than 15, we allow spot-to-spot consolidation. To avoid continual consolidation, we were supposed to sort the spot instance types by prices and truncate the list to 15 instance types to be sent to launch API.

The sorting of spot instance types by prices was missed and this fixes the issue.

Added a functional test. Tested with and without the fix. It fails when the sorting is not place.

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -1391,6 +1391,18 @@ var _ = Describe("Consolidation", func() {
 			// Add these spot instance with this special condition to cloud provider instancetypes
 			cloudProvider.InstanceTypes = spotInstances
 
+			expectedInstanceTypesForConsolidation := make([]*cloudprovider.InstanceType, len(spotInstances))
+			copy(expectedInstanceTypesForConsolidation, spotInstances)
+			// Sort the spot instances by pricing from low to high
+			sort.Slice(expectedInstanceTypesForConsolidation, func(i, j int) bool {
+				return expectedInstanceTypesForConsolidation[i].Offerings[0].Price < expectedInstanceTypesForConsolidation[j].Offerings[0].Price
+			})
+			// These 15 cheapest instance types should eventually be considered for consolidation.
+			var expectedInstanceTypesNames []string
+			for i := 0; i < 15; i++ {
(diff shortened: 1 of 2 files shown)


pkg/controllers/disruption/consolidation_test.go
```diff
@@ -1367,6 +1367,101 @@ var _ = Describe("Consolidation", func() {
...
+			Expect(nodeClaims).To(HaveLen(1))
+			Expect(nodes).To(HaveLen(1))
+
+			// Expect that the new nodeclaim does not request the most expensive instance type
+			Expect(nodeClaims[0].Name).ToNot(Equal(spotNodeClaim.Name))
+			Expect(scheduling.NewNodeSelectorRequirements(nodeClaims[0].Spec.Requirements...).Has(v1.LabelInstanceTypeStable)).To(BeTrue())
+			Expect(scheduling.NewNodeSelectorRequirements(nodeClaims[0].Spec.Requirements...).Get(v1.LabelInstanceTypeStable).Has(mostExpensiveInstanceType.Name)).To(BeFalse())
+
+			// Make sure that the cheapest instance that was outside the bound of 15 instance types is considered for consolidation.
+			Expect(scheduling.NewNodeSelectorRequirements(nodeClaims[0].Spec.Requirements...).Get(v1.LabelInstanceTypeStable).Has(cheapestSpotInstanceType.Name)).To(BeTrue())
```
jonathan-innis, line 1460: Is it worth checking that we just send the 15 instance types overall to ensure a higher bar here? Might also be worth ensuring that the length of this request is also 15 long, which is what we expect

nikmohan123: Addressed. And also made sure to check all the 15 elements in the consolidation list to be the expected first cheapest 15 instance types.

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

Issue #957, opened 2024-01-24
Drift Hash Versioning


engedaam:
### Description

**What problem are you trying to solve?**

The current implementation of the hashing mechanism that is utilized for drift does not make any consideration when there are updates to the NodePool CRDs during karpenter upgrades. The problem that is present here is that all nodes on a customers cluster will get drifted without any changes to the node configuration, but when upgrading the Karpenter CRDs.

More generally, Karpenter will need guard against the following cases during a karpenter upgrade:

1.  NodePool adds a default value to existing fields that are included in hash calculations
2. NodePool  field added to the hash calculation with an already-set values
3. NodePool field, with and without defaults values, is removed from the hash calculations

Without safeguards, users that upgrade to new versions of Karpenter with changes to Drift hashing mechanism will see their whole fleet rolled over and recycled. 

This would required to support in solving the following issues: #909, #337, #493, and https://github.com/aws/karpenter-provider-aws/issues/5447

**How important is this feature to you?**
This will need to be implemented before v1

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1015, opened 2024-02-15
RFC: Propose Drift Hash Versioning

engedaam:
Fixes #N/A 

**Description**
The purpose of this RFC is to Propose option around ways to preform CRD changes to Karpenter and have all nodes in a cluster considered drifted. Related issue: https://github.com/kubernetes-sigs/karpenter/issues/957

### DIFF
--- a/designs/drift-hash-versioning.md
+++ b/designs/drift-hash-versioning.md
@@ -1,0 +1,110 @@
+# Karpenter Drift Hash Versioning with CRD
+
+## Summary of Static Drift Hashing Mechanism
+
+Drift utilizes a hashing mechanism to identify when a `NodeClaim` has been drifted from its owning `NodePool`. Karpenter will hash fields that are within the `NodePool.Spec.Template`, excluding [special cases of drift](https://karpenter.sh/docs/concepts/disruption/#special-cases-on-drift) or [behavior fields](https://karpenter.sh/docs/concepts/disruption/#behavioral-fields). The `NodePool` will be annotated with the hash. When a `NodeClaim` is created from this NodePool, the same `NodePool` hash gets propagated down to the `NodeClaim`. More details can be found in the [drift documentation](https://karpenter.sh/docs/concepts/disruption/#drift). 
+
+When a `NodePool` changes, the hash value of the `NodePool` also changes. The `NodeClaim` disruption controller will periodically check if the `NodePool` hash matches the `NodeClaim`'s hash to determine when a `NodeClaim` is drifted.
+
+## Background
+
+Karpenter's static drift mechanism works well when adding new fields to the CRD that don't have defaults, but breaks down if there is a breaking change to the `Nodepool` hashing mechanism. If a cloud provider makes one of these changes to their CRDs, the hash annotation value would change, causing all nodes to roll when the CRD is upgraded. This causes problems for cloud providers when maintainers need to make these kinds of changes (due to requirements changing or bugs) but don't want to impact users by rolling nodes. There are currently two issues that would force our drift hashing value to change:
+
+1. Setting cloud provider defaults for the: `nodeClassRef.apiVersion` and `nodeClassRef.kind` fields [#909](https://github.com/kubernetes-sigs/karpenter/issues/909)
+2. Fixing `volumeSize` to be correctly considered for hashing [#5447](https://github.com/aws/karpenter-provider-aws/issues/5447)
+
+More generally, these are the types of breaking changes to the hashing function that should not induce drift:
+
+1. A `NodePool` adds default values to an existing field
+2. A `NodePool` field is added with a default
+3. A `NodePool` field is removed
+
+## Solutions
+
+### Option 1: Add a hash version annotation and fallback to rehashing on the NodeClaim 
+
+As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will add a `nodepool-hash-version` into the karpenter controller. The `nodepool-hash-version` will be bumped when a breaking change is made to the hash function. The version bump will be done by the maintainers. The `nodepool-hash-version` will look as such:
+
+```
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: default
+  annotations: 
+    karpenter.sh/nodepool-hash: abcdef
+    karpenter.sh/nodepool-hash-version: v1
+    ...
+---
+apiVersion: karpenter.sh/v1beta1
+kind: NodeClaim 
+metadata:
+  name: default
+  annotations: 
+    karpenter.sh/nodepool-hash: abcdef
+    karpenter.sh/nodepool-hash-version: v1
+    ...
+```
+
+When Karpenter is spun up, the controller will check the `nodepool-hash-version`, and if the version on the NodePool doesn't match the version in the controller, re-calculate the `NodePool` hash and propagate the annotation down to the `NodeClaim`. Karpenter will assert that this `nodepool-hash-version` annotation is equivalent on both the `NodePool` and `NodeClaim` before doing a drift evaluation. 
+
+As a result, two annotations will be used analyze and orchestrate drift: 
+
+* nodepool-hash: (hash of the nodeclaim template on the nodepool)
+* nodepool-hash-version: (the NodePool hashing version defined by Karpenter)
+
+Option 1 has two cases for drift:
+
+1. When the CRD `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter would recalculate the hash of the `NodePool` and back-propagate the updated value for the `karpenter.sh/nodepool-hash` annotation to the NodeClaims. Any `NodeClaims` that are already considered drifted will remain drifted if the `karpenter.sh/nodepool-hash-version` doesn't match
+
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎 Users updating the NodePool while the Karpenter controller is down, during a hash version change, will find NodeClaims not to have nodes drift.
+
+### Option 2: Add a hash version annotation and fallback to individual field evaluation for drift
+
+Similar to Option 1, there are two cases for drift when using `nodepool-hash-version`:
+
+1. When the `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `nodepool-hash` will be compared to evaluate for drift 
+2. When the `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
+    3. The cloudprovider will need to add the `NodeClass` details on the `NodeClaim` for individual field evaluation at NodeClaim creation. Cloudprovider will need to add a JSON annotation of the NodeClass on to the `NodeClaim`.  
+
+Pros/Cons
+
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👍 Remove the race condition of having a new hash being propagating down to the `NodeClaims`
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+### Option 3: Add a hash version annotation, bumping the NodePool API Version to `v1beta2`, utilizing conversion webhooks to update the `nodepool-hash`
+
+The Karpenter APIVersion of the CRDs will be bumped when there is a breaking change to the hashing mechanism described above. As part of retrieving the `NodePool` from the API Server, a conversion webhook would re-calculate the hash, allowing Karpenter will propagate the new hash to the `NodeClaims`.
+
+As a result, we will be coupling the `karpenter.sh/nodepool-hash` hashing mechanism to the apiVersion and we will treat any breaking change to the hash versioning as a breaking change to the CRD. Similar to Option 1, once the conversion webhooks will return an updated `nodepool-hash-version`. This will trigger a hash propagation for `karpenter.sh/nodepool-hash` to all the `NodeClaims`.
+
+Pros/Cons 
+- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👎 Maintain conversion webhooks 
+- 👎👎 In the case of adding a new field with defaults, the API version would be bumped for non-breaking changes. Users running Karpenter would have to bump their manifests to use the new apiVersion to work more frequently
+
+### Option 4: Switch to individual field evaluation for drift
+
+Another option Karpenter to switch completely to checking each field on the `NodePool` against each field on the `NodeClaim`. This option will assume that all fields that used to validate drift will be found on the `NodeClaim`.
+
+Pros/Cons 
+
+- 👍 Consistent drift behavior for upgrading the Karpenter CRDs and standard drift evaluation   
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+## Recommendation
+
+The main goal we are trying to achieve for drift hash versioning is stability on node churn for clusters during a Karpenter upgrade with updates to the `NodePool` CRDs. All four options will prevent Karpenter from automatically drifting nodes during a Karpenter upgrade. Option 1 is recommended as it will give the maintainers more control over when the the drift hash will need to be updated. The advantage of Option 1 is that it would give Karpenter the ability to correctly identify drifted nodes when `NodePool` are changed during a Karpenter upgrade. However, customers that update their `NodePools` resources could run into a race condition, where Karpenter would consider `NodeClaims` not drifted, if they update their `NodePools` while Karpenter is propagating new hashes to `NodeClaims`. The team will mitigate this issue by warning and documenting this side effect to users.


designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Cover general cases of updates to the NodePool/EC2NodeClass CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎 Does not handle cases where users change NodePool/EC2NodeClass resources as Karpenter is updating
+
+### Option 2: Fallback to individual field evaluation for drift
+
+There will be two cases for struct-based hash evaluation:
```
tallaxes, line 80:
```suggestion
Similar to Option 1, there will be two cases for drift:
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Option 1(Recommended): Fallback to rehashing on the NodeClaim 
+
+Similar to option A, there will be two case for drift:
+
+1. When the CRD `nodepool-crd-version/ec2nodeclass-crd-version` matches
+    1. Drift will work as normal and the `NodePool/EC2NodeClass` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-crd-version/ec2nodeclass-crd-version` does not match
+    1. Karpenter would recalculate the hash of the `Nodepool/EC2NodeClass` and back-propagate the annotation to the NodeClaims
+    2. Karpenter will need to lock drift on `NodeClaims` that were considered drifted prior to the Karpenter upgrade.
```
tallaxes, line 69: Can you please elaborate on locking drift?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Option 2: Fallback to individual field evaluation for drift
+
+There will be two cases for struct-based hash evaluation:
+
+1. When the `nodepool-crd-version/ec2nodeclass-crd-version` matches
+    1. Drift will work as normal and the `NodePool/EC2NodeClass` hash will be compared to evaluate for drift 
+2. When the `nodepool-crd-version/ec2nodeclass-crd-version` does not match
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need populate the `NodeClaim` with values that was used in launching the instances. Drift evaluation will need a point of reference on manual checking.
```
tallaxes, line 86:
```suggestion
    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+    karpenter.k8s.aws/ec2nodeclass-hash-v3: stuvwx
+    ...
+
+```
+
+On a Karpenter upgrade, if the hashing mechanism has been updated to contain a new version of the CRDs, a new annotation hash will be added on to the `NodePool/EC2NodeClass`. The new hash will then be back-propagated to the `NodeClaims` that are owned by `NodePool/EC2NodeClass`. During drift evaluation, Karpenter will make sure that all hashes on the `NodeClaims` match the hash present on `NodePool/EC2NodeClass`. If the values do not match, the `NodeClaims` will be drifted. 
+
+Prior to every release, the Karpenter team will validate to see if a new version of the drift hash will be needed based on changes to the CRDs. Each `NodePool/EC2NodeClass` hash annotation will be versioned. We will need to implement proper testing to catch CRD changes that would be breaking and validate that the drift works as expected.
+
+As a result, multiple annotation can potentially be used analyze and orchestrate drift:
```
tallaxes, line 140:
```suggestion
As a result, multiple annotations can potentially be used to analyze and orchestrate drift: 
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+## Future Implications
+
+By the addition of the drift versioning hash, several issues will the have solution paths. With the addition of versioned hashes, defaults of `Kind` and `apiVersion` on the `nodeClassRef` (#909)[https://github.com/kubernetes-sigs/karpenter/issues/909] can be added. **We will need to coordinate with the Azure team to make sure that `Kind` and `apiVersion` are also set on the (karpetner-provider-azure)[https://github.com/Azure/karpenter]**.This additions also allow the team to add improvements on the `NodeClass` drift checking (#377)[https://github.com/kubernetes-sigs/karpenter/issues/337] and a path to surfacing health conditions on resources (#493)[https://github.com/kubernetes-sigs/karpenter/issues/493].
```
tallaxes, line 162:
```suggestion
By the addition of the drift versioning hash, several issues will have solution paths. With the addition of versioned hashes, defaults of `Kind` and `apiVersion` on the `nodeClassRef` (#909)[https://github.com/kubernetes-sigs/karpenter/issues/909] can be added. **We will need to coordinate with the Azure team to make sure that `Kind` and `apiVersion` are also set on the (karpetner-provider-azure)[https://github.com/Azure/karpenter-provider-azure]**. These additions also allow the team to add improvements on the `NodeClass` drift checking (#377)[https://github.com/kubernetes-sigs/karpenter/issues/337] and a path to surfacing health conditions on resources (#493)[https://github.com/kubernetes-sigs/karpenter/issues/493]. 
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+kind: Nodeclaim 
+metadata:
+  name: default
+  annotations: 
+    karpenter.sh/nodepool-hash: abcdef-ghijkl
+    karpenter.k8s.aws/ec2nodeclass-hash: mnopqr-stuvwx
+    ...
+```
+
+As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will prepend a hash of the CRD to the customer resource spec hash. Karpenter will assert that this struct version is equivalent before checking the customer resource spec hash.
```
tallaxes, line 54: > hash of the CRD

This probably deserves more detail. Also, there is likely a class of CRD changes that do not affect hash calculations. Is it proposed to ignore this for the sake of simplicity?

tallaxes (commented):
Minor nits / suggested cleanup from the first reading

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+
+More generally, Karpenter will need to guard against the following cases during a Karpenter upgrade:
+
+1. A `NodePool/EC2NodeClass` adds default values to existing fields that are included in the hash calculations
```
garvinp-stripe, line 15: Slightly ambiguous here. If a new default is added and user previously didn't set an value, doesn't that mean the behavior is changed? Shouldn't it drift? Are there cases where new default values shouldn't be propagated?

garvinp-stripe: Unless this means that the value is set but underlying default is changed. In which case that should be explicitly mentioned to clarify

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+2. A `NodePool/EC2NodeClass` field is added to the hash calculation with an already-set value
```
garvinp-stripe, line 16: Sorry my English isn't the best but I don't think I understand `with an already-set value`. I wonder if rather than saying `guard against the following cases during a Karpenter upgrade` you say `Karpenter won't drift when Karpenter upgrade introduces these changes`. And then here say add a new field into hash calculation without `with an already-set value`.

designs/drift-hash-versioning.md
garvinp-stripe: 👍 agreed. Is this CRD has just the api version + kind? There should be details on what this CRD hash looks like

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will prepend a hash of the CRD to the customer resource spec hash. Karpenter will assert that this struct version is equivalent before checking the customer resource spec hash. 
+
+As a result, two annotations will be used analyze and orchestrate drift: 
+
+* nodepool-hash: (nodepool-crd-version)-(nodepool-hash)
+* ec2nodeclass-hash: (ec2nodeclass-crd-version)-(ec2nodeclass-hash)
+
+### Option 1(Recommended): Fallback to rehashing on the NodeClaim 
+
+Similar to option A, there will be two case for drift:
```
garvinp-stripe, line 63: What is Option A? I think I missed something

designs/drift-hash-versioning.md
engedaam: The idea here is that if a karpenter CRD adds in new default value, it would be unexpected for customers to have all of their nodes be drifted, since they didn't make any changes to the resource itself. I would think customers would have an expectation that the drift takes place once they have made a changes to their resource not update the CRD scheme. Do you disagree that customers would find it unexpected to have all of their nodes drifted on a CRD change? I can make an update to make that point more clear

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+2. A `NodePool/EC2NodeClass` field is added to the hash calculation with an already-set value
+3. A `NodePool/EC2NodeClass` field is removed from the hash calculations
+
+## Consideration
+
+Karpenter currently has special cases of drift for certain fields and behavior fields. These fields do not rely on the hashing mechanism, so hash versioning will not make any changes to these fields.
+
+## Proposed changes
+
+#### Add a CRD hash in the existing drift hash
```
garvinp-stripe, line 25: Is there a reason to prepend vs a different annotation for CRD? Then you don't have to do any string splitting + its more clear what each hash means

designs/drift-hash-versioning.md
engedaam: Typo from a previous version. Fixed!

designs/drift-hash-versioning.md
engedaam: The idea here is to not expand the API surface here, but I can go both ways

designs/drift-hash-versioning.md
garvinp-stripe: I partially agree but at the same time have some concerns. Let's say you change the default value in an CRD update and you don't drift. Now users will have their resources "stuck" in an older CRD behavior for certain fields until they update. What if they never update? What if you continue to update the CRD? How many back versions do you need to track?

In addition, if I am this user and I read the docs to understand something and I read that the current version of the CRD default field to N and I know I deployed this version of the CRD shouldn't I expect N behavior? I didn't realize updating CRD doesn't change this behavior or I need to update my resource in order to pick up this change? What would I update if my resource works the way it is?

I guess this is where the ambiguous behavior shows up for me. I don't think I fully agree that when I update the CRD, I don't expect drift. I do actually because something changed however I agree that changing a CRD shouldn't cause my enter cluster to drift without control. 

I think what I need is clarity in behavior. 
- I don't want my default behavior to differ from expectation (matching the CRD defaults) just because I didn't update the resource.
- I should know when behavior will change during a CRD update. If you add a default I should be able to know ahead of time that updating my CRD will drift my cluster or a set of nodes.
- I need to control how drift happens if a majority of my nodes are going to drift because of a CRD update

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+    1. Drift will work as normal and the `NodePool/EC2NodeClass` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-crd-version/ec2nodeclass-crd-version` does not match
+    1. Karpenter would recalculate the hash of the `Nodepool/EC2NodeClass` and back-propagate the annotation to the NodeClaims
+    2. Karpenter will need to lock drift on `NodeClaims` that were considered drifted prior to the Karpenter upgrade. 
+
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Cover general cases of updates to the NodePool/EC2NodeClass CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade
```
garvinp-stripe, line 75: I dont think I know why this is the case. Can you elaborate?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+2. When the CRD `nodepool-crd-version/ec2nodeclass-crd-version` does not match
+    1. Karpenter would recalculate the hash of the `Nodepool/EC2NodeClass` and back-propagate the annotation to the NodeClaims
+    2. Karpenter will need to lock drift on `NodeClaims` that were considered drifted prior to the Karpenter upgrade. 
+
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Cover general cases of updates to the NodePool/EC2NodeClass CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎 Does not handle cases where users change NodePool/EC2NodeClass resources as Karpenter is updating
```
garvinp-stripe, line 76: Also not completely sure I understand what this means? You can add more details

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,162 @@
...
+    1. Drift will work as normal and the `NodePool/EC2NodeClass` hash will be compared to evaluate for drift 
+2. When the `nodepool-crd-version/ec2nodeclass-crd-version` does not match
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
+
+Pros/Cons
+
+- 👍 Cover general cases of updates to the `NodePool/EC2NodeClass` CRD or the hashing mechanism 
+- 👍 Handles cases where users change `NodePool/EC2NodeClass` resources as Karpenter is updating
+- 👎👎 A lot of implementation and maintenance effort required in supporting individual field evaluation
```
garvinp-stripe, line 92: I would cautious against this as the main reason to not do this. I get this is problematic to maintain but if this solution is more accurate or correct there is value in doing this. 

Its worth explaining if the other solution doesn't cover all the bases why those may not matter or address them in the solution. You should add to what is the risk of having the burden of maintaining this behavior.

designs/drift-hash-versioning.md
engedaam: > I guess this is where the ambiguous behavior shows up for me. I don't think I fully agree that when I update the CRD, I don't expect drift. I do actually because something changed however I agree that changing a CRD shouldn't cause my entire cluster to drift without control.

> I need to control how drift happens if a majority of my nodes are going to drift because of a CRD update 

Trying to understand this point here. What do you mean by drift without control? Currently, customer have ways to control disruption. In the case of a CRD update, disruption will only occur to the extent a customer has allowed for it to happen. The main problem addressed is should karpenter consider all the nodes it has launched to be drifted due to a CRD change? Karpenter will not violate any of the disruption controls that customer provide

> I partially agree but at the same time have some concerns. Let's say you change the default value in an CRD update and you don't drift. Now users will have their resources "stuck" in an older CRD behavior for certain fields until they update. What if they never update their resource?

> I don't want my default behavior to differ from expectation (matching the CRD defaults) just because I didn't update the resource. My nodes should drift because there is a change in behavior defined in the CRD.

This is a good callout. We will need to make sure that customers are aware will need to roll their clusters to match the defaults. This would be through a docs update on the upgrade path. One example would be for them to expire their nodes. Customer should have control over that as opposed to Karpenter automatically drifting IMO. I will make sure to add that here.

> I should know when behavior will change before an CRD update. If you add a new default I should be able to know ahead of time that updating my CRD will drift my cluster or a set of nodes and change its behavior.

We will defiantly need to document to customer when a change that will cause `nodepool-hash` to be re-calculated and propagated down. This change will prevent the CRD update from drifting clusters.

designs/drift-hash-versioning.md
engedaam: I will change to have a separate hash

designs/drift-hash-versioning.md
engedaam: Changed the option to use a version number instead of a hash, that will be updated by users.

designs/drift-hash-versioning.md
engedaam: Defiantly agree here, the main concern is a performance of karpenter when karpenter has to do the individual field evaluation for large clusters.

designs/drift-hash-versioning.md
engedaam: I removed this point. After thinking more about, it didn't make much sense

designs/drift-hash-versioning.md
Bryce-Soghigian: does this refer to when karpenter itself is being upgraded?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+### Option 4: Switch to individual field evaluation for drift
+
+Another option Karpenter to switch completely to checking each field on the `NodePool` against each field on the `NodeClaim`. This option will assume that all fields that used to validate drift will be found on the `NodeClaim`.
+
+Pros/Cons 
+
+- 👍 Consistent drift behavior for upgrading the Karpenter CRDs and standard drift evaluation   
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+- 👎👎👎 Karpenter will experience a performance decrees as we will need to compare each field of the `NodePool` To each field of the `NodeClaim`
```
Bryce-Soghigian, line 167: nit: decrease

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+    karpenter.sh/nodepool-hash-v1: ghijkl
+    karpenter.sh/nodepool-hash-v2: mnopqr
+    karpenter.sh/nodepool-hash-v3: stuvwx
+    ...
+
+```
+
+On a Karpenter upgrade, if the hashing mechanism has been updated to contain a new version of the CRDs, a new annotation hash will be added on to the `NodePool`. The new hash will then be back-propagated to the `NodeClaims` that are owned by `NodePool`. During drift evaluation, Karpenter will make sure that all hashes on the `NodeClaims` match the hash present on `NodePool`. If the values do not match, the `NodeClaims` will be drifted. 
+
+Prior to every release, the Karpenter team will validate to see if a new version of the drift hash will be needed based on changes to the CRDs. Each `NodePool` hash annotation will be versioned. We will need to implement proper testing to catch CRD changes that would be breaking and validate that the drift works as expected.
```
Bryce-Soghigian, line 146: > Prior to every release, the Karpenter team will validate to see if a new version of the drift hash will be needed based on changes to the CRDs.

Should we need to validate and see if a new version of the drift hash will be needed manually or this is proposed code change?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+- 👎👎👎 Karpenter will experience a performance decrees as we will need to compare each field of the `NodePool` To each field of the `NodeClaim` 
+
+## Recommended
+
+The main goal we are trying to achieve for drift hash versioning is stability on node churn for clusters during a Karpenter upgrade with updates to the `NodePool` CRDs. All three options will prevent Karpenter from automatically drifting nodes during a Karpenter upgrade. Option 1A is recommended. The advantage of Option 1A is that it would give Karpenter the ability to correctly identify drifted nodes when `NodePool` are changed during a Karpenter upgrade. This option also allows us to make updates to Karpenter for future cases without any additional API surface changes. However, customers using deployment tools such as Terraform could run into a race condition, where Karpenter would consider `NodeClaims` not drifted, if they update their `NodePools` while also upgrading Karpenter. The team will mitigate this issue by warning and documenting this side effect to users.
```
Bryce-Soghigian, line 171: Nit update all three options to 4 now that there are 4 options.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Rollback Concerns
+
+During rollback to older version of v1beta1 Karpenter that does not contain a method of drift hash versioning, we will need to back-port changes that will prevent drift from not working as expected. Since the proposed changes require altering the format of the current drift hash, older versions of Karpenter do not support any updates to the drift hash on the NodeClaim. This might require customers to roll some or all their nodes on their clusters. The approach going forward will be to back-port the hash-versioning changes to prior release of Karpenter, and document to customers that this is version of Karpenter that they can rollback to have drift work as expected.
```
Bryce-Soghigian, line 175: > This might require customers to roll some or all their nodes on their cluster

Is this feasible for customers?

Bryce-Soghigian (commented):
Overall some solid approaches, others already left good review on sensitive areas. Thanks for including me by asking for my review :) 

WIll take another pass when all comments on other areas are addressed 🫡

designs/drift-hash-versioning.md
engedaam: While the nodes are being drifted, Karpenter will respect all the disruption controls set by users. I'm not sure if Karpenter will have that feature available for customers to use on an older version, but that can be a potential option. We still want to make sure that nodes are not considered drifted, by patching in changes that will update the drift hashes.

designs/drift-hash-versioning.md
garvinp-stripe: > Trying to understand this point here. What do you mean by drift without control? Currently, customer have ways to control disruption. In the case of a CRD update, disruption will only occur to the extent a customer has allowed for it to happen. The main problem addressed is should karpenter consider all the nodes it has launched to be drifted due to a CRD change? Karpenter will not violate any of the disruption controls that customer provide

True I think this is likely less of an issue.

> This is a good callout. We will need to make sure that customers are aware will need to roll their clusters to match the defaults. This would be through a docs update on the upgrade path. One example would be for them to expire their nodes. Customer should have control over that as opposed to Karpenter automatically drifting IMO. I will make sure to add that here.

Exactly. I guess the question I have here is we could recommend what users do but my only worry is will users know? Do users actively read their docs and are the docs consistently up to date? I mainly want to point out that from a user of Karpenter I don't always do these things and in my mind the ideal solution is one that assumes users don't read but yet provide a consistent and/or predictable behavior.

designs/drift-hash-versioning.md
engedaam: This will be something that is needed to be updated manually. Whenever a user makes a change to the NodePool under these condition: 

1. A `NodePool` adds default values to existing fields that are included in the hash calculations
2. A `NodePool` field is added to the hash calculation with an already-set value
3. A `NodePool` field is removed from the hash calculations

We will need to bump the version annotation on the CRD.

designs/drift-hash-versioning.md
jonathan-innis: > that will be updated by users

I think you mean that it will be updated by cloud provider maintainers. I doubt that users will interact with this field.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+```
+...
+- apiGroups: ["apiextensions.k8s.io"]
+  resources: ["customresourcedefinitions"]
+  verbs: ["get", "watch"]
+  resourceNames: ["nodepools.karpenter.sh"]
+...
+```
+
+### Option 1A(Recommended): Fallback to rehashing on the NodeClaim
```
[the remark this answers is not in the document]
jonathan-innis: I'm not quite sure what the best way to re-work it, but I do think having a/b-style options does make the doc harder to track and the pro/cons harder to reason about. I'd consider seeing if you can just flatten everything into a single set of options with pro/cons under each

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎 Does not handle cases where users change `NodePool` resources as Karpenter is updating
+
+### Option 1B: Fallback to individual field evaluation for drift
+
+Similar to Option 1, there will be two cases for drift:
```
jonathan-innis, line 85:
```suggestion
Similar to Option 1a, there will be two cases for drift:
```

Do you mean, option 1a? This is another place where I think flattening the options down would make it easier to read

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+1. When the `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
+
+Pros/Cons
+
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👍 Handles cases where users change `NodePool` resources as Karpenter is updating
```
jonathan-innis, line 96: This one is less clear to the reader since we haven't called out this concern yet anywhere in this doc. I definitely agree that this one is nice because it completely eliminates race issues with the hash annotation options, but it's probably worth giving a little more detail on what this means as a pro

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 Negatively impact on performance for Karpenter until all NodeClaims match the `nodepool-hash-version`
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+### Option 2: Update the NodePool API Version and utilize Conversion webhooks to update the `nodepool-hash`    
+
+The Karpenter APIVersion of the CRDs will be changed each time their is a change as such:
```
jonathan-innis, line 103:
```suggestion
The Karpenter APIVersion of the CRDs will be changed each time there is a change as such: 
```

nit

jonathan-innis, line 103: This criteria was already called out above, do we need to re-list it? I wonder if the last option is really just "Decouple the `karpenter.sh/nodepool-hash` hashing mechanism from the apiVersion" and this one is really just "Couple the `karpenter.sh/nodepool-hash` hashing mechanism to the apiVersion".

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+- 👍 Handles cases where users change `NodePool` resources as Karpenter is updating
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 Negatively impact on performance for Karpenter until all NodeClaims match the `nodepool-hash-version`
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+### Option 2: Update the NodePool API Version and utilize Conversion webhooks to update the `nodepool-hash`    
+
+The Karpenter APIVersion of the CRDs will be changed each time their is a change as such: 
+
+1. A `NodePool` adds default values to existing fields that are included in the hash calculations
```
jonathan-innis, line 105: You could probably distill this criteria down to, "We treat any breaking change to the hash versioning as a breaking change to the CRD, necessitating an apiVersion bump."

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+2. A `NodePool` field is added to the hash calculation with an already-set value
+3. A `NodePool` field is removed from the hash calculations
+
+As part of retrieving the `NodePool` from the API Server, a converion webhook would recaluclate the hash and propagte the new hash to the `NodeClaims`.
```
jonathan-innis, line 109: I don't think the conversion webhook would be able to re-propoagate it down to the NodeClaims, I think you mean that the conversion webhook would convert into the new hash scheme on the NodePool and NodeClasses. The other piece that I think is missing here is that you _still_ need a `karpenter.sh/nodepool-hash-version` annotation (or something similar to it) in this solution. Otherwise, how does the controller know that it needs to re-propagate the updated hash down to the NodeClaims after it converts?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+Pros/Cons 
+- 👍 No Additional permission required on the Karpenter `clusterrole`
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism
```
jonathan-innis, line 113:
```suggestion
- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism 
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+2. A `NodePool` field is added to the hash calculation with an already-set value
+3. A `NodePool` field is removed from the hash calculations
+
+As part of retrieving the `NodePool` from the API Server, a converion webhook would recaluclate the hash and propagte the new hash to the `NodeClaims`.
+
+Pros/Cons 
+- 👍 No Additional permission required on the Karpenter `clusterrole`
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👎 Maintain conversion webhooks 
+- 👎👎 In the case of adding a new field with defaults, the API version would be bumped for non breaking changes
```
jonathan-innis, line 115: We should callout why this is bad. The answer is that it impacts users -- everyone who is running Karpenter would have to bump their manifests to use the new apiVersion to work with this

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+* karpenter.sh/nodepool-hash-v...: (nodepool-hash)
+
+Pros/Cons
+
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎👎 Does not handle cases where users change `NodePool` while Karpenter is updating
+
+### Option 4: Switch to individual field evaluation for drift
```
jonathan-innis, line 158: How is this different from 1b? If we did this, how would the new NodeClaim look with the proposed changes?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+- 👎👎👎 Karpenter will experience a performance decrease as we will need to compare each field of the `NodePool` To each field of the `NodeClaim` 
+
+## Recommended
+
+The main goal we are trying to achieve for drift hash versioning is stability on node churn for clusters during a Karpenter upgrade with updates to the `NodePool` CRDs. All four options will prevent Karpenter from automatically drifting nodes during a Karpenter upgrade. Option 1A is recommended. The advantage of Option 1A is that it would give Karpenter the ability to correctly identify drifted nodes when `NodePool` are changed during a Karpenter upgrade. This option also allows us to make updates to Karpenter for future cases without any additional API surface changes. However, customers using deployment tools such as Terraform could run into a race condition, where Karpenter would consider `NodeClaims` not drifted, if they update their `NodePools` while also upgrading Karpenter. The team will mitigate this issue by warning and documenting this side effect to users.
```
jonathan-innis, line 171: This statement needs to be updated, since it isn't quite accurate anymore. There's a tiny race during the CRD upgrade that needs to be called-out, but other than that, there's not a risk of this race happening.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+## Rollback Concerns
+
+During rollback to older version of v1beta1 Karpenter that does not contain a method of drift hash versioning, we will need to back-port changes that will prevent drift from not working as expected. Since the proposed changes require altering the format of the current drift hash, older versions of Karpenter do not support any updates to the drift hash on the NodeClaim. Once users rollback to an older version of Karpenter, all of their nodes will be considered drifted. The approach going forward will be to back-port the hash-versioning changes to prior release of Karpenter, and document to customers that this is version of Karpenter that they can rollback to have drift work as expected.
```
jonathan-innis, line 175: I think we need to update this rollback section now that we aren't doing prefixing. Agreed though, that we still need some work done ahead-of-time to make sure that is a rollback is possible with the proposed implementation

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+
+## Future Implications
```
jonathan-innis, line 178: I'm 50/50 on whether this section should be included. The main reason that I think that it shouldn't is because some of this was called out in the background while you were talking about the reason why we are trying to solve this problem now.

The piece on Azure _is_ relevant. There's probably a place in the doc to call this out, or maybe just make it a note at the bottom of the doc, rather than its own section.

jonathan-innis (commented):
Nice work. Bunch of nits, structural comments, and high-level feedback.

designs/drift-hash-versioning.md
engedaam: This option would complete switch our drift checking to use individual field evaluation for all drift checking. For 1b, we would only use individual field evaluation if the `NodePool-hash-version` was different.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+As part of retrieving the `NodePool` from the API Server, a converion webhook would recaluclate the hash and propagte the new hash to the `NodeClaims`.
+
+Pros/Cons 
+- 👍 No Additional permission required on the Karpenter `clusterrole`
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👎 Maintain conversion webhooks 
+- 👎👎 In the case of adding a new field with defaults, the API version would be bumped for non breaking changes  
+
+### Option 3: Add Versioned Hash Annotations
```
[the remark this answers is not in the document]
engedaam: Discussed offline, this should not be an issue as the serialization path will be different.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+2. When the `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
+
+Pros/Cons
+
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👍 Handles cases where users change `NodePool` resources as Karpenter is updating
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 Negatively impact on performance for Karpenter until all NodeClaims match the `nodepool-hash-version`
```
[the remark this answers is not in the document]
engedaam: I'm thinking that in large clusters we will see an increases on the time in disruption and drift detection.

designs/drift-hash-versioning.md
engedaam: We should have any rollback issue with new recommended.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+Option 1 will have two case for drift:
+
+1. When the CRD `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter would recalculate the hash of the `Nodepool` and back-propagate the annotation to the NodeClaims
+      - Karpenter will need not back-propagate the `nodepool-crd-hash` annotation `NodeClaims` that are considered drifted.
+    2. Drift will work as normal once a `NodePool` field is updated
+
+Pros/Cons
```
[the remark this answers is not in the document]
engedaam: No need to have extended permission anymore

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Option 1B: Fallback to individual field evaluation for drift
+
+Similar to Option 1, there will be two cases for drift:
+
+1. When the `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
```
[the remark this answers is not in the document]
engedaam: I plan to leave that up to the cloud provider as how they will like to add that in the NodeClass validation on to the NodeClaim. I will mention a few options

designs/drift-hash-versioning.md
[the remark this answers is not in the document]
garvinp-stripe: I am not sure about disruption being slower but I think drift detection can be slower but it would be better to perform some test to see what the delta is. I honestly don't think we are talking about that much slower and disruption controller only trigger reconcile per nodeclaim update / create so worse case we would increase the number of workers on drift controller.

I could be wrong but unless we are changing how disruption controller works for drift, all it does is check `isDrifted` status. I don't see why this change would make that slower

designs/drift-hash-versioning.md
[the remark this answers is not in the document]
engedaam: I believe Jonathan concern is how we will do individual field validation for NodeClasses as those details are not on the NodeClaim. 

Cloudprovider implementation will populate the static fields of the NodeClass on the NodeClaim at NodeClaim creation. This would be the same process as the `ec2nodeclass-hash`, so there is nothing re-populating on to the NodeClaim.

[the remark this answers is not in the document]
garvinp-stripe: ah ok I got it

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
+# Karpenter Drift Hash Versioning with CRD
+
+## Summary of Drift Hashing Mechanism
+
+Drift utilizes a hashing mechanism to identify when a `NodeClaim` has been drifted from its owning `NodePool`. Karpenter will hash fields that are within the `NodePool.Spec.Template`, excluding [special cases of drift](https://karpenter.sh/docs/concepts/disruption/#special-cases-on-drift) or [behavior fields](https://karpenter.sh/docs/concepts/disruption/#behavioral-fields). The `NodePool` will be annotated with the hash. When a `NodeClaim` is created, the `NodePool` hash will be annotated on the `NodeClaim`. More details can be found in the [drift documentation](https://karpenter.sh/docs/concepts/disruption/#drift).
```
jonathan-innis, line 5:
```suggestion
Drift utilizes a hashing mechanism to identify when a `NodeClaim` has been drifted from its owning `NodePool`. Karpenter will hash fields that are within the `NodePool.Spec.Template`, excluding [special cases of drift](https://karpenter.sh/docs/concepts/disruption/#special-cases-on-drift) or [behavior fields](https://karpenter.sh/docs/concepts/disruption/#behavioral-fields). The `NodePool` will be annotated with the hash. When a `NodeClaim` is created from this NodePool, the same `NodePool` hash gets propagated down to the `NodeClaim`. More details can be found in the [drift documentation](https://karpenter.sh/docs/concepts/disruption/#drift). 
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
+# Karpenter Drift Hash Versioning with CRD
+
+## Summary of Drift Hashing Mechanism
+
+Drift utilizes a hashing mechanism to identify when a `NodeClaim` has been drifted from its owning `NodePool`. Karpenter will hash fields that are within the `NodePool.Spec.Template`, excluding [special cases of drift](https://karpenter.sh/docs/concepts/disruption/#special-cases-on-drift) or [behavior fields](https://karpenter.sh/docs/concepts/disruption/#behavioral-fields). The `NodePool` will be annotated with the hash. When a `NodeClaim` is created, the `NodePool` hash will be annotated on the `NodeClaim`. More details can be found in the [drift documentation](https://karpenter.sh/docs/concepts/disruption/#drift). 
+
+When a `NodePool` is updated, a new hash will be annotated on the `NodePool`. The `NodeClaim` disruption controller will periodically check if the `NodePool` hash match to determine when a `NodeClaim` is drifted.
```
jonathan-innis, line 7:
```suggestion
When a `NodePool` changes, the hash value of the `NodePool` also changes. The `NodeClaim` disruption controller will periodically check if the `NodePool` hash matches the `NodeClaim`'s hash to determine when a `NodeClaim` is drifted.
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
+# Karpenter Drift Hash Versioning with CRD
+
+## Summary of Drift Hashing Mechanism
```
jonathan-innis, line 3:
```suggestion
## Summary of Static Drift Hashing Mechanism
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+When a `NodePool` is updated, a new hash will be annotated on the `NodePool`. The `NodeClaim` disruption controller will periodically check if the `NodePool` hash match to determine when a `NodeClaim` is drifted.
+
+## Background
+
+Karpenter's static drift mechanism works well when adding new fields to the CRD that don't have defaults, but breaks down if we add a field that has a set default, remove a field from the CRD, or set a new default for an existing field. If a cloud provider makes one of these changes to their CRDs, the static drift would change, causing all nodes to roll when the CRD is upgraded. This causes problems for cloud providers when maintainers need to make these kinds of changes. We currently have three issue that need a drift hash versioning improvement:
```
jonathan-innis, line 11: I have a recommendation up above to call that section "Summary of Static Drift Hashing Mechanism" so that when you come down here to talk about "Karpenter's static drift mechanism" it's immediately clear what you are referring to. That gives you some nice parallelism in your writing

jonathan-innis, line 11:
```suggestion
Karpenter's static drift mechanism works well when adding new fields to the CRD that don't have defaults, but breaks down if we add a field that has a set default, remove a field from the CRD, or set a new default for an existing field. If a cloud provider makes one of these changes to their CRDs, the hash annotation value would change, causing all nodes to roll when the CRD is upgraded. This causes problems for cloud providers when maintainers need to make these kinds of changes. We currently have three issue that need a drift hash versioning improvement: 
```

I'm not sure if saying that the "static drift" would change is accurate here

jonathan-innis, line 11:
```suggestion
Karpenter's static drift mechanism works well when adding new fields to the CRD that don't have defaults, but breaks down if we add a field that has a set default, remove a field from the CRD, or set a new default for an existing field. If a cloud provider makes one of these changes to their CRDs, the static drift would change, causing all nodes to roll when the CRD is upgraded. This causes problems for cloud providers when maintainers need to make these kinds of changes (due to requirements changing or bugs) but don't want to impact users by rolling nodes. There are currently three issues that would force our drift hashing value to change: 
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+When a `NodePool` is updated, a new hash will be annotated on the `NodePool`. The `NodeClaim` disruption controller will periodically check if the `NodePool` hash match to determine when a `NodeClaim` is drifted.
+
+## Background
+
+Karpenter's static drift mechanism works well when adding new fields to the CRD that don't have defaults, but breaks down if we add a field that has a set default, remove a field from the CRD, or set a new default for an existing field. If a cloud provider makes one of these changes to their CRDs, the static drift would change, causing all nodes to roll when the CRD is upgraded. This causes problems for cloud providers when maintainers need to make these kinds of changes. We currently have three issue that need a drift hash versioning improvement: 
+
+1. The team is looking to set cloud provider defaults for the: `nodeClassRef.apiVersion` and `nodeClassRef.kind` fields [#909](https://github.com/kubernetes-sigs/karpenter/issues/909)
```
jonathan-innis, line 13:
```suggestion
1. Setting cloud provider defaults for the: `nodeClassRef.apiVersion` and `nodeClassRef.kind` fields [#909](https://github.com/kubernetes-sigs/karpenter/issues/909)
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+ -  **We will need to coordinate with the Azure team to make sure that `Kind` and `apiVersion` are also set on the (karpenter-provider-azure)[https://github.com/Azure/karpenter-provider-azure]**
+2. The aws team is blocked from fixing issue [#5447](https://github.com/aws/karpenter-provider-aws/issues/5447)
```
jonathan-innis, line 15:
```suggestion
2. Fixing `volumeSize` to be correctly considered for hashing [#5447](https://github.com/aws/karpenter-provider-aws/issues/5447)
```

jonathan-innis, line 15: Aren't 2 and 3 the same?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+3. The fix to the issue is adding volumeSize in the hashing mechanism of the `EC2NodeClass` (#5454)[https://github.com/aws/karpenter-provider-aws/pull/5454]
+
+However, this will cause all `NodeClaims` to be drifted as soon as a customer upgrades their Karpenter CRDs in their cluster. These two issues raise the need for Karpenter to understand different versions of the CRDs during a drift evaluation. The problem here is that a change on the Karpenter CRDs will cause all nodes on a customer’s cluster will be drifted without any changes to the node configuration.
```
jonathan-innis, line 18: IMO, I think this statement has been said a few times, I think that you could cut this paragraph entirely

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+More generally, these are breaking changes to the hashing function that should not induce drift:
+
+1. A `NodePool` adds default values to an existing field
```
jonathan-innis, line 22: You stated these above (and I really like the way that you have categorized them and put them into a bulleted list). I wonder if there is a way to shift things around so that you don't end up stating the same set of tenets twice 🤔

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+2. A `NodePool` field is added with a default
+3. A `NodePool` field is removed
+
+## Solutions
+
+```
```
jonathan-innis, line 28: This example spec feels out of place? Did you mean for this to be here?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+  name: default
+  annotations: 
+    karpenter.sh/nodepool-hash: abcdef
+    karpenter.sh/nodepool-hash-version: v1
+    ...
+```
+
+#### Option 1: Add a hash version annotation and fallback to rehashing on the NodeClaim 
+
+As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will add a `nodepool-hash-version` into the karpenter controller. The `nodepool-hash-version` be bumped when a breaking change is made to the hash function. The version bump will be done by the maintainers. When Karpenter is spun up, the controller will check the `nodepool-hash-version`, and if needed, re-calculate the `NodePool` hash propagate the annotation down to the `NodeClaim`. Karpenter will assert that this `nodepool-hash-version` annotation is equivalent on both the `NodePool` and `NodeClaim` before doing a drift validation.
```
jonathan-innis, line 50:
```suggestion
As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will add a `nodepool-hash-version` into the karpenter controller. The `nodepool-hash-version` will be bumped when a breaking change is made to the hash function. The version bump will be done by the maintainers. 

When Karpenter is spun up, the controller will check the `nodepool-hash-version`, and if needed, re-calculate the `NodePool` hash propagate the annotation down to the `NodeClaim`. Karpenter will assert that this `nodepool-hash-version` annotation is equivalent on both the `NodePool` and `NodeClaim` before doing a drift validation. 
```
 
small nit on a grammar miss and consider breaking this into two paragraphs

jonathan-innis, line 50: What is a "drift validation"? Is there a way to make that lingo more clear?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+* nodepool-hash: (hash of the nodeclaim template on the nodepool)
+* nodepool-hash-version: (the NodePool hashing version defined by Karpenter)
+
+Option 1 will have two case for drift:
+
+1. When the CRD `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter would recalculate the hash of the `NodePool` and back-propagate the updated value for the `karpenter.sh/nodepool-hash` annotation to the NodeClaims
+      - Any `NodeClaims` that are already considered drifted will remain drifted if the `karpenter.sh/nodepool-hash-version` doesn't match
```
jonathan-innis, line 63: nit: Consider collapsing this bullet into the statement above so you don't get too deep in hierarchy

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+    ...
+```
+
+#### Option 1: Add a hash version annotation and fallback to rehashing on the NodeClaim 
+
+As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will add a `nodepool-hash-version` into the karpenter controller. The `nodepool-hash-version` be bumped when a breaking change is made to the hash function. The version bump will be done by the maintainers. When Karpenter is spun up, the controller will check the `nodepool-hash-version`, and if needed, re-calculate the `NodePool` hash propagate the annotation down to the `NodeClaim`. Karpenter will assert that this `nodepool-hash-version` annotation is equivalent on both the `NodePool` and `NodeClaim` before doing a drift validation. 
+
+As a result, two annotations will be used analyze and orchestrate drift: 
+
+* nodepool-hash: (hash of the nodeclaim template on the nodepool)
```
jonathan-innis, line 54: Have you considered collapsing the two annotations into one with a prefix?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+### Option 2: Add a hash version annotation and Fallback to individual field evaluation for drift
+
+Similar to Option 1, there will be two cases for drift:
+
+1. When the `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter will individually check each field for drift. *This would be an adaptation of the (Deployment Controller)[https://github.com/kubernetes/kubernetes/blob/83e6636096548f4b49a94ece6bc4a26fd67e4449/pkg/controller/deployment/util/deployment_util.go#L603] for podTemplateSpec relationship to pods.*
+    2. Karpenter will need to populate the `NodeClaim` with values that were used in launching the instances. Drift evaluation will need a point of reference on manual checking.
+    3. The cloudprovider will need to add the NodeClass details on the NodeClaim for individual field evaluation at NodeClaim creation. Possible option could be an labels, annotations, JSON raw extension object on the status, etc.
```
jonathan-innis, line 82: I think we should come at this with a recommendation if we are going to propose this as an option. If we need to introduce a new field in the NodeClaim status, let's propose that, if we are going to stick with annotations and JSON, let's propose that

designs/drift-hash-versioning.md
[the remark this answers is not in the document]
jonathan-innis: To me, we should almost completely remove this bullet. The biggest issue with this option (IMO) is the implementation complexity and the maintenance of having to constantly keep up with it

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,155 @@
...
+
+The Karpenter APIVersion of the CRDs will be when their is a breaking change to the hashing mechanism described above. As part of retrieving the `NodePool` from the API Server, a conversion webhook would re-calculate the hash and Karpenter will propagate the new hash to the `NodeClaims`. We will be coupling the `karpenter.sh/nodepool-hash` hashing mechanism to the apiVersion. With this option, the team will treat any breaking change to the hash versioning as a breaking change to the CRD, necessitating an apiVersion bump. Similar to Option 1, once the conversion webhooks will return an updated `nodepool-hash-version`. This will trigger a hash propagation for `karpenter.sh/nodepool-hash` to all the `NodeClaims`.
+
+Pros/Cons 
+- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👎 Maintain conversion webhooks 
+- 👎👎 In the case of adding a new field with defaults, the API version would be bumped for non-breaking changes. Users running Karpenter would have to bump their manifests to use the new apiVersion to work more frequently
+
+### Option 4: Add Versioned Hash Annotations
```
jonathan-innis, line 101: Should we still leave this out here as an option? I'm fine with it, but I also don't see how this option is better in any way than Option 1. It has all of the same behaviors but just has more complexity.

jonathan-innis (commented):
Thanks for breaking all of this out into a streamlined set of options. This is much clearer now that it doesn't have any branching! This is looking good! One more round of comments

designs/drift-hash-versioning.md
engedaam: Changed to drift evaluation

designs/drift-hash-versioning.md
engedaam: The example spec is meant for options 1 and option 2. Added a title to reflect that

designs/drift-hash-versioning.md
engedaam: The statement from above

designs/drift-hash-versioning.md
engedaam: Yeah I did. I was thinking that would require us to do string manipulation, which I think would not be better in the long run

designs/drift-hash-versioning.md
engedaam: Added a proposal to add a JSON annotation on the `NodeClaim`

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,111 @@
...
+
+* nodepool-hash: (hash of the nodeclaim template on the nodepool)
+* nodepool-hash-version: (the NodePool hashing version defined by Karpenter)
+
+Option 1 will have two case for drift:
+
+1. When the CRD `nodepool-hash-version` matches between the `NodePool` and `NodeClaim`:
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter would recalculate the hash of the `NodePool` and back-propagate the updated value for the `karpenter.sh/nodepool-hash` annotation to the NodeClaims. Any `NodeClaims` that are already considered drifted will remain drifted if the `karpenter.sh/nodepool-hash-version` doesn't match
```
njtran, line 62: If we're saying that nodeclaims that are already drifted will remain drifted if the hash version doesn't match, does that mean we don't need to recalculate the hash?

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,111 @@
...
+
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism
+- 👎 Karpenter will need to expanded cluster
```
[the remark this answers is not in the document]
engedaam: Removed. Missed from update

designs/drift-hash-versioning.md
engedaam: Yeah, for those we will not update the hash

designs/drift-hash-versioning.md
jonathan-innis: I personally wouldn't start with an example because you don't get the context for why that example is there. I'd generally give some background information for the solution and then lead that into the example spec. Happy to keep it as-is since we have been through a few iterations at this point, but that's just my two cents.

designs/drift-hash-versioning.md
jonathan-innis: I think @njtran is saying that you don't need to update either of them, it doesn't really matter what you do. You can just leave the NodeClaim around entirely without doing any updates.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+As the Karpenter CRDs change, Karpenter’s underlying hashing function may also change. To handle the upgrade path, Karpenter will add a `nodepool-hash-version` into the karpenter controller. The `nodepool-hash-version` will be bumped when a breaking change is made to the hash function. The version bump will be done by the maintainers.
+
+ When Karpenter is spun up, the controller will check the `nodepool-hash-version`, and if needed, re-calculate the `NodePool` hash propagate the annotation down to the `NodeClaim`. Karpenter will assert that this `nodepool-hash-version` annotation is equivalent on both the `NodePool` and `NodeClaim` before doing a drift evaluation. 
+
+As a result, two annotations will be used analyze and orchestrate drift: 
+
+* nodepool-hash: (hash of the nodeclaim template on the nodepool)
+* nodepool-hash-version: (the NodePool hashing version defined by Karpenter)
+
+Option 1 will have two case for drift:
```
jonathan-innis, line 57:
```suggestion
Option 1 has two cases for drift:
```

nit: present tense is more readable

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter would recalculate the hash of the `NodePool` and back-propagate the updated value for the `karpenter.sh/nodepool-hash` annotation to the NodeClaims. Any `NodeClaims` that are already considered drifted will remain drifted if the `karpenter.sh/nodepool-hash-version` doesn't match
+
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎 Users that update that update the `NodePool` resource while karpenter is propagating to the NodeClaims may not have nodes drift.
```
jonathan-innis, line 69:
```suggestion
- 👎 Users that update the `NodePool` resource while karpenter is back-propagating the hash to the NodeClaims may not have nodes drift.
```

nit

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Option 2: Add a hash version annotation and Fallback to individual field evaluation for drift
```
jonathan-innis, line 71:
```suggestion
### Option 2: Add a hash version annotation and fallback to individual field evaluation for drift
```

nit: for parity with the previous option

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+
+Similar to Option 1, there will be two cases for drift:
```
jonathan-innis, line 73:
```suggestion
Similar to Option 1, there are two cases for drift:
```

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+    3. The cloudprovider will need to add the `NodeClass` details on the `NodeClaim` for individual field evaluation at NodeClaim creation. Cloudprovider will need to add a JSON annotation of the NodeClass on to the `NodeClaim`.  
+
+Pros/Cons
+
+- 👍 Cover general cases of updates to the `NodePool` CRD or the hashing mechanism 
+- 👍 Remove the race condition of having a new hash being propagating down to the `NodeClaims`
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+### Option 3: Add a hash version annotation and Update the NodePool API Version and utilize Conversion webhooks to update the `nodepool-hash`
```
jonathan-innis, line 89:
```suggestion
### Option 3: Add a hash version annotation, bumping the NodePool API Version to `v1beta2`, utilizing conversion webhooks to update the `nodepool-hash`    
```

nit

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+- 👍 Remove the race condition of having a new hash being propagating down to the `NodeClaims`
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+### Option 3: Add a hash version annotation and Update the NodePool API Version and utilize Conversion webhooks to update the `nodepool-hash`    
+
+The Karpenter APIVersion of the CRDs will be when their is a breaking change to the hashing mechanism described above. As part of retrieving the `NodePool` from the API Server, a conversion webhook would re-calculate the hash and Karpenter will propagate the new hash to the `NodeClaims`. We will be coupling the `karpenter.sh/nodepool-hash` hashing mechanism to the apiVersion. With this option, the team will treat any breaking change to the hash versioning as a breaking change to the CRD, necessitating an apiVersion bump. Similar to Option 1, once the conversion webhooks will return an updated `nodepool-hash-version`. This will trigger a hash propagation for `karpenter.sh/nodepool-hash` to all the `NodeClaims`.
```
jonathan-innis, line 91:
```suggestion
The Karpenter APIVersion of the CRDs will be bumped when there is a breaking change to the hashing mechanism described above. As part of retrieving the `NodePool` from the API Server, a conversion webhook would re-calculate the hash, allowing Karpenter will propagate the new hash to the `NodeClaims`.

As a result, we will be coupling the `karpenter.sh/nodepool-hash` hashing mechanism to the apiVersion and we will treat any breaking change to the hash versioning as a breaking change to the CRD. Similar to Option 1, once the conversion webhooks will return an updated `nodepool-hash-version`. This will trigger a hash propagation for `karpenter.sh/nodepool-hash` to all the `NodeClaims`.
```

nit: wording and reducing the number of sentences

designs/drift-hash-versioning.md
engedaam: We will need to disable drift checking if the NodePool versions don't match, so we will need make to update the NodePool version and not update the hash for drifted NodeClaims

jonathan-innis: > not update the hash for drifted NodeClaims

Got it, but you still want to update the version? I don't really think that it makes a major difference either way, I guess to some degree it's nice to have the drifted NodeClaims to be aligned with everything else on version

engedaam: Yeah, that was my thinking

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,111 @@
...
+
+Another option Karpenter to switch completely to checking each field on the `NodePool` against each field on the `NodeClaim`. This option will assume that all fields that used to validate drift will be found on the `NodeClaim`.
+
+Pros/Cons 
+
+- 👍 Consistent drift behavior for upgrading the Karpenter CRDs and standard drift evaluation   
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+## Recommended
```
jonathan-innis, line 109: nit: Consider calling this section `Recommendation`. Not a huge deal, but I feel like you have to pick between "Recommended Option" and "Recommendation" when it comes to the section title.

jonathan-innis (commented):
/lgtm
/approve

jonathan-innis (commented):
/hold

Wait for a second review from @njtran

designs/drift-hash-versioning.md
engedaam: I didn't want to duplicate the example, that was my major concern. Do you have any suggestion on a better way to structure the example?

jonathan-innis: Personally, I would just put the example under Option 1 and then reference back to it with Option 2 if you need. That strikes me as a lot better than throwing it at the beginning without any context to start

engedaam: Updated to include the change

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+
+- 👍 Consistent drift behavior for upgrading the Karpenter CRDs and standard drift evaluation   
+- 👎 Cloud-provider would need to propagate on to the NodeClaim to adopt this method for static drift checking 
+- 👎👎 A high maintenance effort required in supporting individual field evaluation 
+
+## Recommendation
+
+The main goal we are trying to achieve for drift hash versioning is stability on node churn for clusters during a Karpenter upgrade with updates to the `NodePool` CRDs. All four options will prevent Karpenter from automatically drifting nodes during a Karpenter upgrade. Option 1 is recommended as it will give the maintainers when karpenter will need to re-configure the drift hash. The advantage of Option 1 is that it would give Karpenter the ability to correctly identify drifted nodes when `NodePool` are changed during a Karpenter upgrade. However, customers that update their `NodePools` resources could run into a race condition, where Karpenter would consider `NodeClaims` not drifted, if they update their `NodePools` while Karpenter is propagating new hashes to `NodeClaims`. The team will mitigate this issue by warning and documenting this side effect to users.
```
garvinp-stripe, line 110: > Option 1 is recommended as it will give the maintainers when karpenter will need to re-configure the drift hash.

I am not sure I follow this sentence.

designs/drift-hash-versioning.md
```diff
@@ -0,0 +1,110 @@
...
+    1. Drift will work as normal and the `NodePool` hash will be compared to evaluate for drift 
+2. When the CRD `nodepool-hash-version` does not match between the `NodePool` and `NodeClaim`:
+    1. Karpenter would recalculate the hash of the `NodePool` and back-propagate the updated value for the `karpenter.sh/nodepool-hash` annotation to the NodeClaims. Any `NodeClaims` that are already considered drifted will remain drifted if the `karpenter.sh/nodepool-hash-version` doesn't match
+
+Pros/Cons
+
+- 👍👍 Simple implementation effort 
+- 👍 Covers general cases of updates to the `NodePool` CRD or the hashing mechanism
+- 👎 Nodes that were considered drifted before the Karpenter upgrade are not able to be un-drifted after a Karpenter upgrade 
+- 👎 Users that update the `NodePool` resource while karpenter is back-propagating the hash to the NodeClaims may not have nodes drift.
```
garvinp-stripe, line 67: Sorry but can you lay out the sequence of action for this to occur? If this will be addressed in the doc then it can wait. I just can't fully grasp why this may be the case.

designs/drift-hash-versioning.md
engedaam: Updated for clarity.

PR #1016, opened 2024-02-15
feat: Add Versioned for NodePool Hash to Prevent Drift on NodePool CRD Upgrade 

engedaam:
Fixes #N/A 

**Description**
Add versioned NodePool Hash to prevent Drift from occurring under these scenario:
1. A `NodePool` adds default values to existing fields that are included in the hash calculations
2. A `NodePool` field is added to the hash calculation with an already-set value
3. A `NodePool` field is removed from the hash calculations

RFC: https://github.com/kubernetes-sigs/karpenter/pull/1015
Related Issue: https://github.com/kubernetes-sigs/karpenter/issues/957

- `make presubmit`

### DIFF
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -46,6 +46,7 @@ const (
 	ProviderCompatabilityAnnotationKey = CompatabilityGroup + "/provider"
 	ManagedByAnnotationKey             = Group + "/managed-by"
 	NodePoolHashAnnotationKey          = Group + "/nodepool-hash"
+	NodePoolHashVersionAnnotationKey   = Group + "/nodepool-hash-version"
 )
 
 // Karpenter specific finalizers
--- a/pkg/apis/v1beta1/nodepool.go
+++ b/pkg/apis/v1beta1/nodepool.go
@@ -190,6 +190,12 @@ type NodePool struct {
 	Status NodePoolStatus `json:"status,omitempty"`
 }
 
+// We need to bump the NodePoolHashVersion when we make an update to the NodePool CRD under these conditions:
+// 1. A field changes its default value for an existing field that is already hashed
+// 2. A field is added to the hash calculation with an already-set value
+// 3. A field is removed from the hash calculations
+const NodePoolHashVersion = "v1"
+
 func (in *NodePool) Hash() string {
 	return fmt.Sprint(lo.Must(hashstructure.Hash(in.Spec.Template, hashstructure.FormatV2, &hashstructure.HashOptions{
 		SlicesAsSets:    true,
--- a/pkg/controllers/nodeclaim/disruption/drift.go
+++ b/pkg/controllers/nodeclaim/disruption/drift.go
@@ -115,12 +115,19 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1beta1.NodePool, nodeC
 	return driftedReason, nil
 }
 
-// Eligible fields for static drift are described in the docs
+// Eligible fields for drift are described in the docs
 // https://karpenter.sh/docs/concepts/deprovisioning/#drift
 func areStaticFieldsDrifted(nodePool *v1beta1.NodePool, nodeClaim *v1beta1.NodeClaim) cloudprovider.DriftReason {
 	nodePoolHash, foundHashNodePool := nodePool.Annotations[v1beta1.NodePoolHashAnnotationKey]
+	nodePoolVersionHash, foundVersionHashNodePool := nodePool.Annotations[v1beta1.NodePoolHashVersionAnnotationKey]
 	nodeClaimHash, foundHashNodeClaim := nodeClaim.Annotations[v1beta1.NodePoolHashAnnotationKey]
-	if !foundHashNodePool || !foundHashNodeClaim {
+	nodeClaimVersionHash, foundVersionHashNodeClaim := nodeClaim.Annotations[v1beta1.NodePoolHashVersionAnnotationKey]
+
+	if !foundHashNodePool || !foundHashNodeClaim || !foundVersionHashNodePool || !foundVersionHashNodeClaim {
+		return ""
+	}
+	// validate that the version of the crd is the same
+	if nodePoolVersionHash != nodeClaimVersionHash {
 		return ""
 	}
 	return lo.Ternary(nodePoolHash != nodeClaimHash, NodePoolDrifted, "")
--- a/pkg/controllers/nodeclaim/disruption/drift_test.go
+++ b/pkg/controllers/nodeclaim/disruption/drift_test.go
@@ -124,7 +124,12 @@ var _ = Describe("Drift", func() {
 	It("should detect static drift before cloud provider drift", func() {
 		cp.Drifted = "drifted"
 		nodePool.Annotations = lo.Assign(nodePool.Annotations, map[string]string{
-			v1beta1.NodePoolHashAnnotationKey: "123456789",
+			v1beta1.NodePoolHashAnnotationKey:        "test-123456789",
+			v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+		})
+		nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, map[string]string{
+			v1beta1.NodePoolHashAnnotationKey:        "test-123",
+			v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
 		})
 		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
 		ExpectReconcileSucceeded(ctx, nodeClaimDisruptionController, client.ObjectKeyFromObject(nodeClaim))
@@ -431,8 +436,9 @@ var _ = Describe("Drift", func() {
 					Template: v1beta1.NodeClaimTemplate{
 						ObjectMeta: v1beta1.ObjectMeta{
 							Annotations: map[string]string{
-								"keyAnnotation":  "valueAnnotation",
-								"keyAnnotation2": "valueAnnotation2",
+								"keyAnnotation":                          "valueAnnotation",
+								"keyAnnotation2":                         "valueAnnotation2",
+								v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
 							},
 							Labels: map[string]string{
 								"keyLabel":  "valueLabel",
@@ -494,6 +500,20 @@ var _ = Describe("Drift", func() {
 		})
 		It("should not return drifted if karpenter.sh/nodePool-hash annotation is not present on the nodeClaim", func() {
 			nodeClaim.ObjectMeta.Annotations = map[string]string{}
+			ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+			ExpectReconcileSucceeded(ctx, nodeClaimDisruptionController, client.ObjectKeyFromObject(nodeClaim))
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+			Expect(nodeClaim.StatusConditions().GetCondition(v1beta1.Drifted)).To(BeNil())
+		})
+		It("should not return drifted if the NodeClaim's karpenter.sh/nodepool-hash-version annotation does not match the NodePool's", func() {
+			nodePool.ObjectMeta.Annotations = map[string]string{
+				v1beta1.NodePoolHashAnnotationKey:        "test-hash-1",
+				v1beta1.NodePoolHashVersionAnnotationKey: "test-version-1",
+			}
+			nodeClaim.ObjectMeta.Annotations = map[string]string{
+				v1beta1.NodePoolHashAnnotationKey:        "test-hash-2",
+				v1beta1.NodePoolHashVersionAnnotationKey: "test-version-2",
+			}
 			ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
 			ExpectReconcileSucceeded(ctx, nodeClaimDisruptionController, client.ObjectKeyFromObject(nodeClaim))
 			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
--- a/pkg/controllers/nodepool/hash/controller.go
+++ b/pkg/controllers/nodepool/hash/controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 
 	"github.com/samber/lo"
+	"go.uber.org/multierr"
 	"k8s.io/apimachinery/pkg/api/equality"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -48,7 +49,16 @@ func NewController(kubeClient client.Client) operatorcontroller.Controller {
 // Reconcile the resource
 func (c *Controller) Reconcile(ctx context.Context, np *v1beta1.NodePool) (reconcile.Result, error) {
 	stored := np.DeepCopy()
-	np.Annotations = lo.Assign(np.Annotations, map[string]string{v1beta1.NodePoolHashAnnotationKey: np.Hash()})
+
+	if np.Annotations[v1beta1.NodePoolHashVersionAnnotationKey] != v1beta1.NodePoolHashVersion {
+		if err := c.updateNodeClaimHash(ctx, np); err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+	np.Annotations = lo.Assign(np.Annotations, map[string]string{
+		v1beta1.NodePoolHashAnnotationKey:        np.Hash(),
+		v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+	})
 
 	if !equality.Semantic.DeepEqual(stored, np) {
 		if err := c.kubeClient.Patch(ctx, np, client.MergeFrom(stored)); err != nil {
@@ -68,4 +78,43 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 		For(&v1beta1.NodePool{}).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}),
 	)
+}
+
+// Updating `nodepool-hash-version` annotation inside the controller means a breaking change has made to the hash function calculating
+// `nodepool-hash` on both the NodePool and NodeClaim, automatically making the `nodepool-hash` on the NodeClaim different from
+// NodePool. Since we can not rely on the hash on the NodeClaims, we will need to re-calculate the hash and update the annotation.
+// Look at designs/drift-hash-version.md for more information.
+func (c *Controller) updateNodeClaimHash(ctx context.Context, np *v1beta1.NodePool) error {
+	ncList := &v1beta1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, ncList, client.MatchingLabels(map[string]string{v1beta1.NodePoolLabelKey: np.Name})); err != nil {
+		return err
+	}
+
+	errs := make([]error, len(ncList.Items))
+	for i := range ncList.Items {
+		nc := ncList.Items[i]
+		stored := nc.DeepCopy()
+
+		if nc.Annotations[v1beta1.NodePoolHashVersionAnnotationKey] != v1beta1.NodePoolHashVersion {
+			nc.Annotations = lo.Assign(nc.Annotations, map[string]string{
+				v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+			})
+
+			// Any NodeClaim that is already drifted will remain drifted if the karpenter.sh/nodepool-hash-version doesn't match
+			// Since the hashing mechanism has changed we will not be able to determine if the drifted status of the node has changed
+			if nc.StatusConditions().GetCondition(v1beta1.Drifted) == nil {
+				nc.Annotations = lo.Assign(nc.Annotations, map[string]string{
+					v1beta1.NodePoolHashAnnotationKey: np.Hash(),
+				})
+			}
+
+			if !equality.Semantic.DeepEqual(stored, nc) {
+				if err := c.kubeClient.Patch(ctx, &nc, client.MergeFrom(stored)); err != nil {
+					errs[i] = client.IgnoreNotFound(err)
+				}
+			}
+		}
+	}
+
+	return multierr.Combine(errs...)
 }
--- a/pkg/controllers/nodepool/hash/suite_test.go
+++ b/pkg/controllers/nodepool/hash/suite_test.go
@@ -30,9 +30,11 @@ import (
 	"knative.dev/pkg/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	. "sigs.k8s.io/karpenter/pkg/test/expectations"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"sigs.k8s.io/karpenter/pkg/apis"
+	. "sigs.k8s.io/karpenter/pkg/test/expectations"
+
 	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 	"sigs.k8s.io/karpenter/pkg/controllers/nodepool/hash"
 	"sigs.k8s.io/karpenter/pkg/operator/controller"
@@ -95,7 +97,7 @@ var _ = Describe("Static Drift Hash", func() {
 			},
 		})
 	})
-	It("should update the static drift hash when NodePool static field is updated", func() {
+	It("should update the drift hash when NodePool static field is updated", func() {
 		ExpectApplied(ctx, env.Client, nodePool)
 		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
 		nodePool = ExpectExists(ctx, env.Client, nodePool)
@@ -111,7 +113,7 @@ var _ = Describe("Static Drift Hash", func() {
 		expectedHashTwo := nodePool.Hash()
 		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHashTwo))
 	})
-	It("should not update the static drift hash when NodePool behavior field is updated", func() {
+	It("should not update the drift hash when NodePool behavior field is updated", func() {
 		ExpectApplied(ctx, env.Client, nodePool)
 		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
 		nodePool = ExpectExists(ctx, env.Client, nodePool)
@@ -133,5 +135,108 @@ var _ = Describe("Static Drift Hash", func() {
 		nodePool = ExpectExists(ctx, env.Client, nodePool)
 
 		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+	})
+	It("should update nodepool hash version when the nodepool hash version is out of sync with the controller hash version", func() {
+		nodePool.Annotations = map[string]string{
+			v1beta1.NodePoolHashAnnotationKey:        "abceduefed",
+			v1beta1.NodePoolHashVersionAnnotationKey: "test",
+		}
+		ExpectApplied(ctx, env.Client, nodePool)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+
+		expectedHash := nodePool.Hash()
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should update nodepool hash versions on all nodeclaims when the hash versions don't match the controller hash version", func() {
+		nodePool.Annotations = map[string]string{
+			v1beta1.NodePoolHashAnnotationKey:        "abceduefed",
+			v1beta1.NodePoolHashVersionAnnotationKey: "test",
+		}
+		nodeClaimOne := test.NodeClaim(v1beta1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{v1beta1.NodePoolLabelKey: nodePool.Name},
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "123456",
+					v1beta1.NodePoolHashVersionAnnotationKey: "test",
+				},
+			},
+		})
+		nodeClaimTwo := test.NodeClaim(v1beta1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{v1beta1.NodePoolLabelKey: nodePool.Name},
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "123456",
+					v1beta1.NodePoolHashVersionAnnotationKey: "test",
+				},
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaimOne, nodeClaimTwo)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		nodeClaimOne = ExpectExists(ctx, env.Client, nodeClaimOne)
+		nodeClaimTwo = ExpectExists(ctx, env.Client, nodeClaimTwo)
+
+		expectedHash := nodePool.Hash()
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on all nodeclaims when the hash versions match the controller hash version", func() {
+		nodePool.Annotations = map[string]string{
+			v1beta1.NodePoolHashAnnotationKey:        "abceduefed",
+			v1beta1.NodePoolHashVersionAnnotationKey: "test-version",
+		}
+		nodeClaim := test.NodeClaim(v1beta1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{v1beta1.NodePoolLabelKey: nodePool.Name},
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "1234564654",
+					v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+				},
+			},
+		})
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		expectedHash := nodePool.Hash()
+
+		// Expect NodeClaims to have been updated to the original hash
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "1234564654"))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on the nodeclaim if it's drifted", func() {
+		nodePool.Annotations = map[string]string{
+			v1beta1.NodePoolHashAnnotationKey:        "abceduefed",
+			v1beta1.NodePoolHashVersionAnnotationKey: "test",
+		}
+		nodeClaim := test.NodeClaim(v1beta1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{v1beta1.NodePoolLabelKey: nodePool.Name},
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "123456",
+					v1beta1.NodePoolHashVersionAnnotationKey: "test",
+				},
+			},
+		})
+		nodeClaim.StatusConditions().MarkTrue(v1beta1.Drifted)
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		// Expect NodeClaims hash to not have been updated
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "123456"))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
 	})
 })
--- a/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
@@ -66,8 +66,11 @@ func (i *NodeClaimTemplate) ToNodeClaim(nodePool *v1beta1.NodePool) *v1beta1.Nod
 	nc := &v1beta1.NodeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", i.NodePoolName),
-			Annotations:  lo.Assign(i.Annotations, map[string]string{v1beta1.NodePoolHashAnnotationKey: nodePool.Hash()}),
-			Labels:       i.Labels,
+			Annotations: lo.Assign(i.Annotations, map[string]string{
+				v1beta1.NodePoolHashAnnotationKey:        nodePool.Hash(),
+				v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+			}),
+			Labels: i.Labels,
 			OwnerReferences: []metav1.OwnerReference{
 				{
 					APIVersion:         v1beta1.SchemeGroupVersion.String(),


pkg/operator/operator.go
```diff
@@ -127,6 +129,7 @@ func NewOperator() (context.Context, *Operator) {
 
 	// Client
 	kubernetesInterface := kubernetes.NewForConfigOrDie(config)
+	apiextensionsInterface := clientset.NewForConfigOrDie(config)
```
jonathan-innis, line 132: You shouldn't need to create a new interface just for the apiextensions, you could probably re-use the `kubernetesInterface` here

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -36,19 +40,35 @@ var _ operatorcontroller.TypedController[*v1beta1.NodePool] = (*Controller)(nil)
...
+	kubeClient             client.Client
+	apiextensionsInterface clientset.Interface
 }
 
-func NewController(kubeClient client.Client) operatorcontroller.Controller {
+func NewController(kubeClient client.Client, apiextensionsInterface clientset.Interface) operatorcontroller.Controller {
 	return operatorcontroller.Typed[*v1beta1.NodePool](kubeClient, &Controller{
-		kubeClient: kubeClient,
+		kubeClient:             kubeClient,
+		apiextensionsInterface: apiextensionsInterface,
```
jonathan-innis, line 50: Can you look into seeing if you can access the CRD through the `kubeClient` (controller-runtime interface). There's nothing in our code that's actually using the `kubernetesInteface` anymore from the operator, so it would be really nice if we just called out to the apiserver every time with the same controller-runtime interface

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -36,19 +40,35 @@ var _ operatorcontroller.TypedController[*v1beta1.NodePool] = (*Controller)(nil)
...
 	})
 }
 
 // Reconcile the resource
 func (c *Controller) Reconcile(ctx context.Context, np *v1beta1.NodePool) (reconcile.Result, error) {
 	stored := np.DeepCopy()
-	np.Annotations = lo.Assign(np.Annotations, map[string]string{v1beta1.NodePoolHashAnnotationKey: np.Hash()})
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("nodepool", np.Name))
```
jonathan-innis, line 57: This is already auto-injected by the reconciler

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +89,30 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}),
 	)
 }
+
+func (c *Controller) updateNodeClaimHash(ctx context.Context, np *v1beta1.NodePool, nphash string) error {
+	ncList := &v1beta1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, ncList); err != nil {
```
jonathan-innis, line 95: Can we consider creating an indexer here for the hash version and then using a field selector to get all the NodeClaims that don't match the current version. This is going to save on memory and performance

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +89,30 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
+		return client.IgnoreNotFound(err)
```
jonathan-innis, line 96: I don't think that you can get a NotFound error for a List operation

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +89,30 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
+	}
+
+	for i := range ncList.Items {
+		stored := ncList.Items[i].DeepCopy()
+		if ncList.Items[i].Labels[v1beta1.NodePoolLabelKey] != np.Name {
```
jonathan-innis, line 101: You should be able to label select on this, so that you don't have to do this filtering after the fact

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +89,30 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
 	)
 }
+
+func (c *Controller) updateNodeClaimHash(ctx context.Context, np *v1beta1.NodePool, nphash string) error {
+	ncList := &v1beta1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, ncList); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+
+	for i := range ncList.Items {
```
jonathan-innis, line 99:
```suggestion
	for i, nc := range ncList.Items {
```

Is there something preventing us from assigning the NodeClaim to a variable in the range loop here?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -51,8 +55,11 @@ func TestAPIs(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
-	env = test.NewEnvironment(scheme.Scheme, test.WithCRDs(apis.CRDs...))
-	nodePoolController = hash.NewController(env.Client)
+	s := scheme.Scheme
```
jonathan-innis, line 58: Why did we need to create a new scheme and add all of these things to it?

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -62,7 +62,7 @@ func (i *NodeClaimTemplate) ToNodeClaim(nodePool *v1beta1.NodePool) *v1beta1.Nod
 	nc := &v1beta1.NodeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", i.NodePoolName),
-			Annotations:  lo.Assign(i.Annotations, map[string]string{v1beta1.NodePoolHashAnnotationKey: nodePool.Hash()}),
+			Annotations:  lo.Assign(i.Annotations, map[string]string{v1beta1.NodePoolHashAnnotationKey: nodePool.Annotations[v1beta1.NodePoolHashAnnotationKey]}),
```
jonathan-innis, line 65: What happens here if the NodePool hash hasn't been reconciled yet from this NodePool? I think that we grab these NodePools really early on in the scheduling loop so this may be possible. 

The other thing: What's the necessity of changing this logic? Shouldn't new NodeClaims just use whatever the hashing algorithm is for the current in-memory schema?

pkg/operator/operator.go
engedaam: Removed.

pkg/apis/v1beta1/labels.go
```diff
@@ -46,6 +46,7 @@ const (
 	ProviderCompatabilityAnnotationKey = CompatabilityGroup + "/provider"
 	ManagedByAnnotationKey             = Group + "/managed-by"
 	NodePoolHashAnnotationKey          = Group + "/nodepool-hash"
+	NodePoolHashVersionAnnotationKey   = Group + "/nodepool-hash-version"
```
jonathan-innis, line 49: Do we want to do a separate annotation or do a version prefix here?

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -115,15 +116,20 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1beta1.NodePool, nodeC
...
 // https://karpenter.sh/docs/concepts/deprovisioning/#drift
 func areStaticFieldsDrifted(nodePool *v1beta1.NodePool, nodeClaim *v1beta1.NodeClaim) cloudprovider.DriftReason {
 	nodePoolHash, foundHashNodePool := nodePool.Annotations[v1beta1.NodePoolHashAnnotationKey]
 	nodeClaimHash, foundHashNodeClaim := nodeClaim.Annotations[v1beta1.NodePoolHashAnnotationKey]
 	if !foundHashNodePool || !foundHashNodeClaim {
 		return ""
 	}
-	return lo.Ternary(nodePoolHash != nodeClaimHash, NodePoolDrifted, "")
+	// validate that the version of the crd is the same
+	nodePoolHashSlice, nodeClaimHashSlice := strings.Split(nodePoolHash, "-"), strings.Split(nodeClaimHash, "-")
```
jonathan-innis, line 128: Are we using the version annotation or are we prefixing with the version? This indicates that we are prefixing with the version, but from looking at the annotations that you have listed in the other file, it looks like we have a separate annotation for it

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -499,5 +505,12 @@ var _ = Describe("Drift", func() {
 			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 			Expect(nodeClaim.StatusConditions().GetCondition(v1beta1.Drifted)).To(BeNil())
 		})
+		It("should not return drifted if karpenter.sh/nodePool-hash annotation version does not match the nodeClaim", func() {
```
jonathan-innis, line 508:
```suggestion
		It("should not return drifted if karpenter.sh/nodepool-hash-version annotation does not match the NodePool", func() {
```

jonathan-innis, line 508: Do we need a test to validate that we stay drifted in the scenario where we have already drifted and we get a change in the hash annotation version?

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +78,32 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}),
 	)
 }
+
+func (c *Controller) updateNodeClaimHash(ctx context.Context, np *v1beta1.NodePool) error {
+	ncList := &v1beta1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, ncList); err != nil {
```
jonathan-innis, line 84: I think we still need a field selector here to only select NodeClaims that are owned by this NodePool

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +78,32 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
+		return err
+	}
+
+	for i := range ncList.Items {
+		stored := ncList.Items[i].DeepCopy()
+		if ncList.Items[i].Labels[v1beta1.NodePoolLabelKey] != np.Name {
+			continue
+		}
+
+		if ncList.Items[i].Annotations[v1beta1.NodePoolHashVersionAnnotationKey] != v1beta1.NodePoolHashVersion && ncList.Items[i].StatusConditions().GetCondition(v1beta1.Drifted) == nil {
```
jonathan-innis, line 94: Can we comment over this? Otherwise it's going to be a little unclear why we decide to do this check on the drifted status condition

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +78,32 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
+			ncList.Items[i].Annotations = lo.Assign(ncList.Items[i].Annotations, map[string]string{
+				v1beta1.NodePoolHashAnnotationKey:        np.Hash(),
+				v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+			})
+			if !equality.Semantic.DeepEqual(stored, ncList.Items[i]) {
+				if err := c.kubeClient.Patch(ctx, &ncList.Items[i], client.MergeFrom(stored)); err != nil {
+					return client.IgnoreNotFound(err)
```
jonathan-innis, line 101: This means that we will exit out of this loop early returning a nil error in the case that one of the NodeClaims that we went to grab is NotFound, which isn't what we want here

jonathan-innis: I'm wondering if we should just best-effort try to update everything and then create multi-err and return that. The other option is to just shift the `err != nil` check in the `if` statement to a `client.IgnoreNotFound(err) != nil` check

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
 
 		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
 	})
+	It("should update nodepool hash version when the nodepool hash version is out of sync with the controller", func() {
+		nodePool.Annotations[v1beta1.NodePoolHashVersionAnnotationKey] = "test"
+		ExpectApplied(ctx, env.Client, nodePool)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "123456"))
```
jonathan-innis, line 147: Do we need these expectation functions to assert startup conditions for our test case?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+					v1beta1.NodePoolHashVersionAnnotationKey: "test",
+				},
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaimOne, nodeClaimTwo)
+		nodeClaimOne = ExpectExists(ctx, env.Client, nodeClaimOne)
+		nodeClaimTwo = ExpectExists(ctx, env.Client, nodeClaimTwo)
+
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "123456"))
```
jonathan-innis, line 181: Same question here: Do we need to assert that the thing that we just applied is present and has the data that we just applied? Or can we just assume that and create less expectation functions here?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+					v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+				},
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "123456"))
```
jonathan-innis, line 213: Same comment here

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+		nodeClaimOne = ExpectExists(ctx, env.Client, nodeClaimOne)
+		nodeClaimTwo = ExpectExists(ctx, env.Client, nodeClaimTwo)
+
+		expectedHash := nodePool.Hash()
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on all nodeclaims when the hash versions match", func() {
```
jonathan-innis, line 197: Does this test actually validate anything? You're calling reconcile but asserting that the data that you applied at the beginning of the test hasn't changed? I'm not sure that tests any of the functionality of the things that you added to the controller.

Just from reading this test case: I would have expected this test to validate that when the hash version has been propagated down to the NodeClaim already (because the version is updated), that if the NodePool hash is different from the NodeClaim, the NodeClaim hash doesn't get updated

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+
+		expectedHash := nodePool.Hash()
+
+		// Expect NodeClaims to have been updated to the original hash
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "1234564654"))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on for the nodeclaim if its drifted", func() {
```
jonathan-innis, line 230:
```suggestion
	It("should not update nodepool hash on the nodeclaim if it's drifted", func() {
```

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+					v1beta1.NodePoolHashAnnotationKey:        "123456",
+					v1beta1.NodePoolHashVersionAnnotationKey: "test",
+				},
+			},
+		})
+
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "123456"))
```
jonathan-innis, line 244: Same comment, I'd drop, don't think that it's neccesary

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, "test"))
+
+		nodeClaim.StatusConditions().SetCondition(knativeapis.Condition{
```
jonathan-innis, line 247: Can we just do `nodeClaim.StatusConditions().MarkTrue()`?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+			Type:     v1beta1.Drifted,
+			Status:   v1.ConditionTrue,
+			Severity: knativeapis.ConditionSeverityWarning,
+			Reason:   string("Test Reason"),
+		})
+		ExpectApplied(ctx, env.Client, nodeClaim)
```
jonathan-innis, line 253: Do you need to apply the NodeClaim twice? Once before updating the condition and once after?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +139,124 @@ var _ = Describe("Static Drift Hash", func() {
...
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "1234564654"))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on for the nodeclaim if its drifted", func() {
+		nodeClaim := test.NodeClaim(v1beta1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{v1beta1.NodePoolLabelKey: nodePool.Name},
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "123456",
```
jonathan-innis, line 235: It might be nice if the NodeClaim hash value and the NodePool hash value didn't match in tests where you are trying to validate that this controller isn't updating data. It's going to give us a lot stronger confidence that the data isn't being touched if all of the data isn't matching.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -66,8 +66,11 @@ func (i *NodeClaimTemplate) ToNodeClaim(nodePool *v1beta1.NodePool) *v1beta1.Nod
 	nc := &v1beta1.NodeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", i.NodePoolName),
-			Annotations:  lo.Assign(i.Annotations, map[string]string{v1beta1.NodePoolHashAnnotationKey: nodePool.Hash()}),
-			Labels:       i.Labels,
+			Annotations: lo.Assign(i.Annotations, map[string]string{
+				v1beta1.NodePoolHashAnnotationKey:        nodePool.Annotations[v1beta1.NodePoolHashAnnotationKey],
```
jonathan-innis, line 70: I think this is slightly problematic. What happens if we create a NodeClaim before we have updated the version and NodePool hash on controller startup? This NodeClaim will have the old hash and version. Further, it's possible that when we do the List operation, we don't get this NodeClaim (due to a race condition) and then we don't update the hash and version on this NodeClaim. That leaves some NodeClaims on the cluster with the old hash and version when we didn't mean to.

pkg/operator/operator.go
```diff
@@ -89,9 +90,10 @@ func init() {
 type Operator struct {
 	manager.Manager
 
-	KubernetesInterface kubernetes.Interface
-	EventRecorder       events.Recorder
-	Clock               clock.Clock
+	KubernetesInterface   kubernetes.Interface
+	ApiextensionInterface clientset.Interface
```
jonathan-innis, line 94: This isn't needed anymore, right? Further, I don't think that we even need the KubernetesInterface in the operator anymore, but we can take that change separately

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
engedaam: Removed for this design.

pkg/controllers/nodepool/hash/controller.go
engedaam: Removed for this design.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
engedaam: This was from an older iteration, changed back to include the `nodePool.Hash()`

pkg/controllers/nodeclaim/disruption/drift.go
engedaam: The goal will be to use a version annotation, to avoid a string splitting.

pkg/apis/v1beta1/labels.go
engedaam: The goal will be to use a version annotation, to avoid a string splitting.

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -115,12 +115,19 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1beta1.NodePool, nodeC
...
+// Eligible fields for drift are described in the docs
 // https://karpenter.sh/docs/concepts/deprovisioning/#drift
 func areStaticFieldsDrifted(nodePool *v1beta1.NodePool, nodeClaim *v1beta1.NodeClaim) cloudprovider.DriftReason {
 	nodePoolHash, foundHashNodePool := nodePool.Annotations[v1beta1.NodePoolHashAnnotationKey]
+	nodePoolVersionHash, foundVersionHashNodePool := nodePool.Annotations[v1beta1.NodePoolHashVersionAnnotationKey]
 	nodeClaimHash, foundHashNodeClaim := nodeClaim.Annotations[v1beta1.NodePoolHashAnnotationKey]
-	if !foundHashNodePool || !foundHashNodeClaim {
+	nodeClaimVersionHash, foundVersionHashNodeClaim := nodeClaim.Annotations[v1beta1.NodePoolHashVersionAnnotationKey]
+
+	if !foundHashNodePool || !foundHashNodeClaim || !foundVersionHashNodePool || !foundVersionHashNodeClaim {
```
jonathan-innis, line 126: Just to confirm: This logic means that we have a version where we propagate the hash version on but don't change any hash logic. That will allow the rollback scenario to work for users that decide later after they upgrade that they now want to go back to a previous version with a different hash version

engedaam: Yes, that will be exactly how the rollback scenario will work

pkg/controllers/nodepool/hash/controller.go
engedaam: I added a best-effort multi-err approach

pkg/apis/v1beta1/nodepool.go
```diff
@@ -190,6 +190,12 @@ type NodePool struct {
 	Status NodePoolStatus `json:"status,omitempty"`
 }
 
+// We need to bump the NodePoolHashVersion under these condition when we make an update the NodePool CRD
```
njtran, line 193:
```suggestion
// We need to bump the NodePoolHashVersion under these condition when we make an update to the NodePool CRD
```

pkg/apis/v1beta1/nodepool.go
```diff
@@ -190,6 +190,12 @@ type NodePool struct {
...
+// 1. A NodePool adds default values to existing fields that are included in the hash calculations
+// 2. A NodePool field is added to the hash calculation with an already-set value
+// 3. A NodePool field is removed from the hash calculations
```
njtran, line 196: nit: Since it's called NodePool..... you can probably just remove the NodePool bit
```suggestion
// 1. A change to the default value to existing fields that are already hashed 
// 2. A field is added to the hash calculation with an already-set value
// 3. A field is removed from the hash calculations
```

pkg/controllers/nodeclaim/disruption/drift.go
njtran: Before this commit, NodeClaims and NodePools won't have this hash version key, which means we won't be considering these nodes. Can we just change it to only check for the hash here? 
```suggestion
	if !foundHashNodePool || !foundHashNodeClaim {
```

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +79,37 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
+				v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
+			})
+
+			// Any NodeClaim that is already drifted will remain drifted if the karpenter.sh/nodepool-hash-version doesn't match
+			// Since the hashing mechanism has changed we will not be able to determine if the drifted status of the node has changed
+			if ncList.Items[i].StatusConditions().GetCondition(v1beta1.Drifted) == nil {
+				ncList.Items[i].Annotations = lo.Assign(ncList.Items[i].Annotations, map[string]string{
+					v1beta1.NodePoolHashAnnotationKey: np.Hash(),
+				})
+			}
```
njtran, line 104: Don't we want to check this first, and not update the hash version of the nodeclaim if it's drifted? If we just update the nodepool hash version in the first iteration, we'll just update it on the second iteration.

engedaam: We will not update the `nodepool-hash` on the NodeClaim, if the `nodepool-hash-version`  match the NodePool

pkg/controllers/nodeclaim/disruption/drift.go
engedaam: The hash controller will update the both NodePool and NodeClaim to add the `nodepool-hash-version`

pkg/apis/v1beta1/nodepool.go
```diff
@@ -190,6 +190,12 @@ type NodePool struct {
 	Status NodePoolStatus `json:"status,omitempty"`
 }
 
+// We need to bump the NodePoolHashVersion under these condition when we make an update to the NodePool CRD
```
jonathan-innis, line 193:
```suggestion
// We need to bump the NodePoolHashVersion when we make an update to the NodePool CRD under these conditions:
```

nit: consider re-ordering this wording

pkg/apis/v1beta1/nodepool.go
```diff
@@ -190,6 +190,12 @@ type NodePool struct {
...
+// 1. A change to the default value to existing fields that are already hashed
```
jonathan-innis, line 194:
```suggestion
// 1. A field changes its default value for an existing field that is already hashed
```

nit: wording for parallelism with the other conditions here

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -431,8 +436,9 @@ var _ = Describe("Drift", func() {
 					Template: v1beta1.NodeClaimTemplate{
 						ObjectMeta: v1beta1.ObjectMeta{
 							Annotations: map[string]string{
-								"keyAnnotation":  "valueAnnotation",
-								"keyAnnotation2": "valueAnnotation2",
+								"keyAnnotation":                          "valueAnnotation",
+								"keyAnnotation2":                         "valueAnnotation2",
+								v1beta1.NodePoolHashVersionAnnotationKey: "test-version",
```
jonathan-innis, line 441: Just curious: Why are we giving the NodePool the wrong hash version in the general case?

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -499,5 +505,19 @@ var _ = Describe("Drift", func() {
 			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 			Expect(nodeClaim.StatusConditions().GetCondition(v1beta1.Drifted)).To(BeNil())
 		})
+		It("should not return drifted if karpenter.sh/nodepool-hash-version annotation version does not match the NodePool on the NodeClaim", func() {
```
jonathan-innis, line 508:
```suggestion
		It("should not return drifted if the NodeClaim's karpenter.sh/nodepool-hash-version annotation does not match the NodePool's", func() {
```

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +79,37 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
 }
+
+func (c *Controller) updateNodeClaimHash(ctx context.Context, np *v1beta1.NodePool) error {
+	ncList := &v1beta1.NodeClaimList{}
+	if err := c.kubeClient.List(ctx, ncList, client.MatchingLabels(map[string]string{v1beta1.NodePoolLabelKey: np.Name})); err != nil {
+		return err
+	}
+
+	errs := make([]error, len(ncList.Items))
+	for i := range ncList.Items {
```
jonathan-innis, line 90: Can we either assign to the a variable within the range loop like `for i, nc :=` or (if that's not possible because that would cause addressing issues) can we just assign to a value at the top of the loop like `nc := ncList.Items[i]`? That's going to reduce the verbosity for a lot of the checks that are happening within this range loop

pkg/controllers/nodepool/hash/controller.go
jonathan-innis: I could go either way on this. I don't think that it tangibly matters one way or the other here. If you don't update the hash but update the version, this just means nodes that are already drifted are still going to stay drifted because their hashes _still_ won't match the NodePool's hash.

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -63,13 +65,15 @@ var _ = Describe("Static Drift Hash", func() {
 	var nodePool *v1beta1.NodePool
 	BeforeEach(func() {
 		nodePool = test.NodePool(v1beta1.NodePool{
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "abceduefed",
```
jonathan-innis, line 70: Should we add this in the generic case or just in the cases where we know that we want to validate that the hash annotation and the hash version are already on the NodePool

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +138,100 @@ var _ = Describe("Static Drift Hash", func() {
...
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on all nodeclaims when the hash versions match", func() {
+		nodeClaim := test.NodeClaim(v1beta1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{v1beta1.NodePoolLabelKey: nodePool.Name},
+				Annotations: map[string]string{
+					v1beta1.NodePoolHashAnnotationKey:        "1234564654",
```
jonathan-innis, line 190: Can we put a comment over this to mark that this value is different from the set value up above. I think even better would be not setting this value on the NodePool in the BeforeEach and just doing it within the context of the testcase itself

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +138,100 @@ var _ = Describe("Static Drift Hash", func() {
...
+		})
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		expectedHash := nodePool.Hash()
+
+		// Expect NodeClaims to have been updated to the original hash
```
jonathan-innis, line 203: Do you need to update this NodeClaim twice or can you just set this NodeClaim to match the existing version at the beginning of the test so that you don't have to go through two rounds of reconciling and updating the NodePools and NodeClaims

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +138,100 @@ var _ = Describe("Static Drift Hash", func() {
...
+				},
+			},
+		})
+		nodeClaim.StatusConditions().MarkTrue(v1beta1.Drifted)
+		ExpectApplied(ctx, env.Client, nodeClaim)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		// Expect NodeClaims to have been updated to the original hash
```
jonathan-innis, line 233: Is this comment accurate?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +138,100 @@ var _ = Describe("Static Drift Hash", func() {
...
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, "123456"))
+		Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, "test"))
```
jonathan-innis, line 235: Given what you've written in the NodePool, I would have expected this version to update to match `testNodePool` from the NodePool in the BeforeEach block

jonathan-innis (commented):
Some test comments. Other implementation looks good to me

pkg/controllers/nodepool/hash/suite_test.go
engedaam: Added the annotation based on each test case

pkg/controllers/nodepool/hash/suite_test.go
engedaam: We expect that the hash version will be updated, but the hash itself

pkg/controllers/nodepool/hash/controller.go
engedaam: Yup! That was my thinking

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -124,7 +124,12 @@ var _ = Describe("Drift", func() {
 	It("should detect static drift before cloud provider drift", func() {
 		cp.Drifted = "drifted"
 		nodePool.Annotations = lo.Assign(nodePool.Annotations, map[string]string{
-			v1beta1.NodePoolHashAnnotationKey: "123456789",
+			v1beta1.NodePoolHashAnnotationKey:        "test-123456789",
+			v1beta1.NodePoolHashVersionAnnotationKey: "v1",
```
jonathan-innis, line 128:
```suggestion
			v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion
```

We should match this to whatever the constant is currently set to; otherwise, we are going to have to keep updating this value whenever there is a change in this version

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -499,5 +505,19 @@ var _ = Describe("Drift", func() {
 			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 			Expect(nodeClaim.StatusConditions().GetCondition(v1beta1.Drifted)).To(BeNil())
 		})
+		It("should not return drifted if the NodeClaim's karpenter.sh/nodepool-hash-version annotation does not match the NodePool's", func() {
+			nodePool.ObjectMeta.Annotations = map[string]string{
+				v1beta1.NodePoolHashAnnotationKey:        "test-hash-1",
```
jonathan-innis, line 510: This won't return drifted even if these two objects had the same version because their hashes match. When we are creating test cases like this, we need to make sure that when the opposite thing is true that the test will fail, that way we ensure that we are holding the bar for our code here.

Without these things being differentiated, we aren't really testing what we are intending to test.

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -66,10 +68,6 @@ var _ = Describe("Static Drift Hash", func() {
 			Spec: v1beta1.NodePoolSpec{
 				Template: v1beta1.NodeClaimTemplate{
 					ObjectMeta: v1beta1.ObjectMeta{
-						Annotations: map[string]string{
```
jonathan-innis, line 69: Reason that this got dropped from here?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +132,107 @@ var _ = Describe("Static Drift Hash", func() {
...
+		expectedHash := nodePool.Hash()
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on all nodeclaims when the hash versions match", func() {
+		nodePool.Annotations = map[string]string{
+			v1beta1.NodePoolHashAnnotationKey:        "abceduefed",
+			v1beta1.NodePoolHashVersionAnnotationKey: v1beta1.NodePoolHashVersion,
```
jonathan-innis, line 189: You should use a version on the NodePool that doesn't match so that you are still trying to do the update of all of the NodeClaims, but you don't because all of the NodeClaim's versions match.

pkg/controllers/nodepool/hash/suite_test.go
engedaam: Updated the hash to reflect the value NodePoolHashVersion

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +136,107 @@ var _ = Describe("Static Drift Hash", func() {
...
+		nodeClaimOne = ExpectExists(ctx, env.Client, nodeClaimOne)
+		nodeClaimTwo = ExpectExists(ctx, env.Client, nodeClaimTwo)
+
+		expectedHash := nodePool.Hash()
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimOne.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodeClaimTwo.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should not update nodepool hash on all nodeclaims when the hash versions match the controller", func() {
```
jonathan-innis, line 190:
```suggestion
	It("should not update nodepool hash on all nodeclaims when the hash versions match the controller hash version", func() {
```

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +136,107 @@ var _ = Describe("Static Drift Hash", func() {
...
+		ExpectApplied(ctx, env.Client, nodePool)
+
+		ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+
+		expectedHash := nodePool.Hash()
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
+		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashVersionAnnotationKey, v1beta1.NodePoolHashVersion))
+	})
+	It("should update nodepool hash versions on all nodeclaims when the hash versions don't match", func() {
```
jonathan-innis, line 153:
```suggestion
	It("should update nodepool hash versions on all nodeclaims when the hash versions don't match the controller hash version", func() {
```

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +79,38 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}),
 	)
 }
+
+func (c *Controller) updateNodeClaimHash(ctx context.Context, np *v1beta1.NodePool) error {
```
jonathan-innis, line 83: nit: consider putting a comment around this function to describe what it is doing and why. Coming back to this later, `updateNodeClaimHash` isn't going to be descriptive enough for us to understand (or some other reader to understand) the rationale around why we are having to propagate the new hash value down to all NodeClaims who don't have an updated hash version.

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -69,3 +79,42 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}),
 	)
 }
+
+// Updating `nodepool-hash-version` annotation inside the controller means a breaking change has made to the hash function calculating
+// `nodepool-hash` on both the NodePool and NodeClaim, automatically making the `nodepool-hash` on the NodeClaim different from
+// NodePool. Since we can not relay on the hash on the NodeClaims, we will need to re-calculate the hash and update the annotation.
```
jonathan-innis, line 85:
```suggestion
// NodePool. Since we can not rely on the hash on the NodeClaims, we will need to re-calculate the hash and update the annotation.
```

nit

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -134,4 +136,107 @@ var _ = Describe("Static Drift Hash", func() {
 
 		Expect(nodePool.Annotations).To(HaveKeyWithValue(v1beta1.NodePoolHashAnnotationKey, expectedHash))
 	})
+	It("should update nodepool hash version when the nodepool hash version is out of sync with the controller", func() {
```
jonathan-innis, line 139:
```suggestion
	It("should update nodepool hash version when the nodepool hash version is out of sync with the controller hash version", func() {
```

I would say controller "version" because it's not that it's out-of-sync with the controller, it's that it's out-of-sync with the controller version

jonathan-innis (approved):
/lgtm
/approve

PR #1017, opened 2024-02-15
test: Fix data race while calling `Reset()` in CI testing

jonathan-innis:
Fixes #N/A 

**Description**

Because we were running an async flow to try to force a DATA RACE if the locking/unlocking wasn't working properly for the set, we were now causing a Reconcile() to happen while we were calling `Reset()` since we weren't waiting for the goroutine to finish 

`make presubmit`


engedaam (commented):
/lgtm

PR #1023, opened 2024-02-17
test: Fix flake in requeueAfter returned from expiration disruption

jonathan-innis:
Fixes #N/A 

**Description**

This fixes a flake in the `requeueAfter` returned from a test in the NodeClaim Disruption controller. This ensures that the expiry time is always 100 seconds away from the current time.

`make presubmit`


njtran (commented):
/lgtm
/approve

PR #1025, opened 2024-02-19
chore: Add pod/node count to metrics and logs for disruption

jonathan-innis:
Fixes #N/A 

**Description**

This adds a disrupted reschedulable pod count metric along with a disrupted node count metric. This also adds a pod count to the log that is printed when we disrupt nodes. This would help in debugging issues like in #1014 where it's tough to tell if a node is terminated with nodes bound or not. 

`make presubmit`

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -142,7 +142,7 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 }
 
 func (c *Controller) disrupt(ctx context.Context, disruption Method) (bool, error) {
-	defer metrics.Measure(disruptionEvaluationDurationHistogram.With(map[string]string{
+	defer metrics.Measure(EvaluationDurationHistogram.With(map[string]string{
 		methodLabel:            disruption.Type(),
 		consolidationTypeLabel: disruption.ConsolidationType(),
 	}))()
@@ -181,11 +181,6 @@ func (c *Controller) disrupt(ctx context.Context, disruption Method) (bool, erro
 // 2. Spin up replacement nodes
 // 3. Add Command to orchestration.Queue to wait to delete the candiates.
 func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command, schedulingResults scheduling.Results) error {
-	disruptionActionsPerformedCounter.With(map[string]string{
-		actionLabel:            string(cmd.Action()),
-		methodLabel:            m.Type(),
-		consolidationTypeLabel: m.ConsolidationType(),
-	}).Inc()
 	commandID := uuid.NewUUID()
 	logging.FromContext(ctx).With("command-id", commandID).Infof("disrupting via %s %s", m.Type(), cmd)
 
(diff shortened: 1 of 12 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -234,6 +229,27 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+			methodLabel:            m.Type(),
+			consolidationTypeLabel: m.ConsolidationType(),
+		}).Inc()
+		PodsDisruptedCounter.With(map[string]string{
+			metrics.NodePoolLabel:  cd.nodePool.Name,
+			actionLabel:            string(cmd.Action()),
+			methodLabel:            m.Type(),
+			consolidationTypeLabel: m.ConsolidationType(),
+		}).Add(float64(len(cd.reschedulablePods)))
+	}
```
Bryce-Soghigian, line 252: **Shallow Comment**

Seems like iterating through candidates here is expensive for metric. Is there somewhere we can do this inside of the waitOrTerminate function in the orchestration queue?

pkg/controllers/disruption/suite_test.go
```diff
@@ -1361,6 +1366,358 @@ var _ = Describe("Candidate Filtering", func() {
...
+		var wg sync.WaitGroup
+		ExpectTriggerVerifyAction(&wg)
+		ExpectReconcileSucceeded(ctx, disruptionController, client.ObjectKey{})
+		wg.Wait()
+
+		ExpectMetricCounterValue("karpenter_disruption_actions_performed_total", 1, map[string]string{
+			"action":             "replace",
+			"method":             "consolidation",
+			"consolidation_type": "multi",
+		})
```
Bryce-Soghigian, line 1707: nit: can we use the proper constants for these and other labels + value?

pkg/controllers/disruption/suite_test.go
```diff
@@ -1361,6 +1366,358 @@ var _ = Describe("Candidate Filtering", func() {
...
+	}
+	BeforeEach(func() {
+		nodePool = test.NodePool(v1beta1.NodePool{
+			Spec: v1beta1.NodePoolSpec{
+				Disruption: v1beta1.Disruption{
+					ConsolidationPolicy: v1beta1.ConsolidationPolicyWhenUnderutilized,
+					// Disrupt away!
+					Budgets: []v1beta1.Budget{{
+						Nodes: "100%",
+					}},
```
Bryce-Soghigian, line 1382: nit: this seems to be a common Nodepool configuration in our tests. Thoughts on adding another test package helper for greating highDisruptingNodepool? this way we don't keep rewriting this every test?

Bryce-Soghigian:
Very useful metrics, excited to see this!

pkg/controllers/disruption/controller.go
Bryce-Soghigian: After checking the code, looks cleaner to separate this out after the fact.

jonathan-innis: Yeah, I think this is a reasonable place to do it, personally

pkg/controllers/disruption/suite_test.go
jonathan-innis: I tend to avoid using constants in testing. The reason: it can mask weird mistakes where you mess with your constant or variable and you didn't intend to. If you test against a know value that you always expect to be there, then you can be assured that if you constant accidentally changes, you will catch it in your testing.

Bryce-Soghigian: fair principle

pkg/controllers/disruption/suite_test.go
jonathan-innis: For now, I think this is fine. This is really us just wanting to override our default that is set for production (10%). It's not clear to me if setting a value other than 10% in our default NodePool is a good idea or not, so it seems fine to me to just override it specifically when we need it

pkg/controllers/disruption/metrics.go
```diff
@@ -46,16 +53,34 @@ var (
...
 		[]string{methodLabel, consolidationTypeLabel},
 	)
-	disruptionActionsPerformedCounter = prometheus.NewCounterVec(
+	ActionsPerformedCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Namespace: metrics.Namespace,
 			Subsystem: disruptionSubsystem,
 			Name:      "actions_performed_total",
-			Help:      "Number of disruption actions performed. Labeled by disruption method.",
+			Help:      "Number of disruption actions performed. Labeled by disruption action, method, and consolidation type.",
```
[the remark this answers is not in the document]
jonathan-innis: Yeah, I agree. Seems like good scope for some of the observability work that we are about to talk about here soon.

pkg/controllers/disruption/metrics.go
```diff
@@ -46,16 +53,34 @@ var (
...
 		},
 		[]string{actionLabel, methodLabel, consolidationTypeLabel},
 	)
-	disruptionEligibleNodesGauge = prometheus.NewGaugeVec(
+	NodesDisruptedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: disruptionSubsystem,
+			Name:      "nodes_disrupted_total",
+			Help:      "Total number of nodes disrupted. Labeled by disruption action, method, and consolidation type.",
```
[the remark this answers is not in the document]
jonathan-innis: I 100% agree. That's actually captured here: https://github.com/aws/karpenter-provider-aws/issues/5276 but agree that when the docs shift to a neutral site, this also becomes a `kubernetes-sigs/karpenter` issue

pkg/controllers/disruption/metrics.go
```diff
@@ -46,16 +53,34 @@ var (
...
+		},
+		[]string{metrics.NodePoolLabel, actionLabel, methodLabel, consolidationTypeLabel},
```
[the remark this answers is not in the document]
jonathan-innis: Agree. Seems like we should do cleanup of these metrics wholistically when we look at the whole observability story here. Right now, this is the convention that we have gone with, so I'm going to stick to it for the existing metrics.

jonathan-innis:
@Bryce-Soghigian Thanks for the feedback.

engedaam (approved):
/lgtm

Issue #903, opened 2023-12-20
Karpenter is slow to deprovision workers after a significant scale-in


domgoodwin:
### Description

**What problem are you trying to solve?**
We commonly have large scale out events where we might add 2-5k cores onto our cluster and then reverse the scaling once the event has past. Once this happens Karpenter can take 6-12 hours (depending on the size of the drop) to consolidate the workers so that we're back to a normal level of unused space on them. This means we have a significant amount of unused compute whilst it's scaling down and this happens multiple times a week.

In our other, older, cluster, which uses the cluster-autoscaler, the scale in of workers happens relatively quickly - within a couple of hours we see the nodes terminated and drained. We configure the scale down percentage on the cluster-autoscaler to be 80% so any node with it's request utilisation less then that is drained and terminated.

I'm not certain on the best way to address this, it seems like the node consolidation is happening one node at a time as we see a pretty linear drop in our worker node capacity. Maybe if Karpenter could consolidate in parallel or not as wait so long between consolidations this would be improved?

**How important is this feature to you?**
Relatively important given the amount of wasted compute, we're exploring replicating the cluster-autoscaler behaviour where any node with < a % utilisation is drained and deleted but need to see how this might interact with Karpenter's behaviour.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

tzneal:
Can you provide:
- Karpenter version
- Karpenter controller cpu/memory config
- full controller logs
- size of cluster in terms of pods and nodes both before and after scaling

engedaam:
@domgoodwin do you have any updates here? Can you provide the information asked above?

domgoodwin:
> Can you provide:
> - Karpenter version
> - Karpenter controller cpu/memory config
> - full controller logs
> - size of cluster in terms of pods and nodes both before and after scaling
> 

0.32.2

We don't enforce CPU limits and give it 8 cores in it's request, we don't see any signs it's CPU starved in the metric.

Memory 6gb request, 12gb limit

Cluster size drops by about 3k pods from peak, from 16k to 13k. Most of our resizing is scaling down the CPU requests per pod which causes the large drop.

[...]

PR #1031, opened 2024-02-20
feat: Add options to configure consolidation timeouts

domgoodwin:
Issue: https://github.com/kubernetes-sigs/karpenter/issues/903

**Description**
Adds option values to configure the timeout for multi and single node consolidations. Depending on the size of a cluster and your various affinities/topology spreads this can take longer then these hard coded values. Allowing them to be configured means large complex clusters can have longer timeouts if wanted.

I added this version as the Karpenter version for the aws-provider repo, built and image and deployed it on a test cluster. Working with the value both set and not set (defaulting to existing values) things were fine.
ie.
```
replace sigs.k8s.io/karpenter => github.com/domgoodwin/karpenter v0.0.0-20240220105243-cf5336018872
```

The unit test changes also cover setting the value to a non-default and then timing out after that time.

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -59,6 +59,8 @@ var _ = Describe("Consolidation", func() {
 	var labels = map[string]string{
 		"app": "test",
 	}
+	singleNodeConsolidationTimeout := 5 * time.Minute
+	multiNodeConsolidationTimeout := 1 * time.Minute
 	BeforeEach(func() {
 		nodePool = test.NodePool(v1beta1.NodePool{
 			Spec: v1beta1.NodePoolSpec{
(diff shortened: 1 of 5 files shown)


jonathan-innis:
> Depending on the size of a cluster and your various affinities/topology spreads this can take longer then these hard coded values

This features starts to look similar to our batch duration parameters, which we were a little iffy on whether or not we should have surfaced in the first-place. Do you also set the batch duration parameters to some custom values overriding the defaults?

The other question that I have: What would you set the values to if you could override them as is proposed here? And what's the size of your cluster that doesn't work well with the current timeout values?

domgoodwin:
> > Depending on the size of a cluster and your various affinities/topology spreads this can take longer then these hard coded values
> 
> This features starts to look similar to our batch duration parameters, which we were a little iffy on whether or not we should have surfaced in the first-place. Do you also set the batch duration parameters to some custom values overriding the defaults?
> 
> The other question that I have: What would you set the values to if you could override them as is proposed here? And what's the size of your cluster that doesn't work well with the current timeout values?

I haven't changed the batch duration parameters currently as I figured this was more around scale out, which seems to work ok. We do get a bit of rubber banding where we add more nodes then we need and consolidate down but honestly the speed trade off by not having to run any headroom pods seems worth it.

We actually are running this code as our cluster just wasn't scaling in at all with these timeouts. We've currently set them to 10m, although based on metrics 5m would be fine.

We initially just threw more CPU at Karpenter hoping it would help but it only ever uses 2 CPU cores seemingly

njtran:
I'm hesitant to make this configurable, as changing these knobs can have unknown impacts on the performance of Consolidation. Can you share what cluster sizes you're running at? Are you using any complex scheduling constraints like anti affinity? If anything, I can see making this scale to the size of the cluster in the long run, but I'd be concerned with giving users free reign over this.

domgoodwin:
The cluster is anywhere between 25k and 15k pods from peak to overnight, 250-500 nodes too. 

We used to have anti-affinities but removed them across the board and saw a significant scheduling timing performance increase.
We do have podTopologySpread's setup basically on every pod to spread across AZs and hosts (we don't want 15 of a the same deployment on a node) but both of these are mostly "ScheduledAnyway" so soft constraints.

njtran:
How have you come to these numbers? Did you test them out yourself? I can understand that these timeouts shouldn't be one size fits all, and it sounds like the numbers you're proposing would work for your situation better. Especially when adding an API surface (even though it's a feature gate), this is a type of feature that would still require an RFC. Do you mind writing one up? Feel free to reach out to the maintainers on the kubernetes slack to figure out how to write one, or check out the existing RFCs in the repo.

PR #1041, opened 2024-02-22
chore: Reject NodePools and NodeClaims if they don't have spec

jonathan-innis:
Fixes #N/A 

**Description**

We've seen some Karpenter users running into issues with NodePool or NodeClaim applies, specifically when the spec isn't present. If you apply the NodePool or NodeClaim without a spec, you can circumvent the other checking, we should just require the spec.

One example issue that we hit in the AWS provider repo is here: https://github.com/aws/karpenter-provider-aws/issues/5689

`make presubmit`
`kubectl apply -f pkg/apis/crds && kubectl apply -f nodepool.yaml` and validate failure with no `spec` specified


engedaam (commented):
/lgtm

PR #1043, opened 2024-02-23
perf: don't inject new loggers in GetVolumes

jmdeal:
Fixes #N/A 

**Description**
This PR addresses a performance issue brought forward [here](https://github.com/aws/karpenter-provider-aws/issues/5505). Based on the provided profile image in this [comment](https://github.com/aws/karpenter-provider-aws/issues/5505#issuecomment-1944034348) we can see that `GetVolumes` is using a disproportional amount of CPU time (204 / 586 seconds). 

The culprit line:
```
ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", pod.Name))
```

At most scales this is negligible, however, for each scheduling loop this new logger with be injected `n*p` times, where `n` is the number of existing nodes on the cluster and `p` is the number of pods being considered in a given scheduling loop. Deferring the `With` calls to the log site rather than injecting a new logger into the context does reduce code maintainability, but is worth the performance tradeoff in this case.

### DIFF
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -80,16 +80,14 @@ func (u Volumes) Insert(volumes Volumes) {
 
 //nolint:gocyclo
 func GetVolumes(ctx context.Context, kubeClient client.Client, pod *v1.Pod) (Volumes, error) {
-	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("pod", pod.Name))
 	podPVCs := Volumes{}
 	for _, volume := range pod.Spec.Volumes {
-		ctx = logging.WithLogger(ctx, logging.FromContext(ctx).With("volume", volume.Name))
 		pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, pod, volume)
 		// If the PVC is not found it was manually deleted and its finalizer removed. We should ignore this volume when
 		// computing limits, otherwise Karpenter may never be able to update its cluster state.
 		if err != nil {
 			if errors.IsNotFound(err) {
-				logging.FromContext(ctx).Errorf("failed updating volume limits for volume, %w", err)
+				logging.FromContext(ctx).With("pod", pod.Name, "volume", volume.Name).Errorf("failed updating volume limits for volume, %w", err)
 				continue
 			}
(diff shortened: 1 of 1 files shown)


pkg/scheduling/volumeusage.go
```diff
@@ -135,7 +132,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
 	// In either of these cases, a PV must have been previously bound to the PVC and has since been removed. We can
 	// ignore this PVC while computing limits and continue.
 	if storageClassName == "" {
-		logging.FromContext(ctx).Errorf("failed updating volume limits for volume with unbound PVC, no storage class specified")
+		logging.FromContext(ctx).With("pod", pod.Name).With("pvc", pvc.Name).Errorf("failed updating volume limits for volume with unbound PVC, no storage class specified")
```
jonathan-innis, line 135:
```suggestion
		logging.FromContext(ctx).With("pod", pod.Name, "pvc", pvc.Name).Errorf("failed updating volume limits for volume with unbound PVC, no storage class specified")
```

Prefer this over calling `With()` twice

pkg/scheduling/volumeusage.go
```diff
@@ -146,7 +143,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, volumeName str
 		//  2. The StorageClass never existed and was used to bind the PVC to an existing PV, but that PV was removed
 		// In either of these cases, we should ignore the PVC while computing limits and continue.
 		if errors.IsNotFound(err) {
-			logging.FromContext(ctx).With("storageclass", storageClassName).Errorf("failed updating volume limits for volume with unbound PVC, %w", err)
+			logging.FromContext(ctx).With("pod", pod.Name).With("pvc", pvc.Name).With("storageclass", storageClassName).Errorf("failed updating volume limits for volume with unbound PVC, %w", err)
```
jonathan-innis, line 146:
```suggestion
			logging.FromContext(ctx).With("pod", pod.Name, "pvc", pvc.Name).With("storageclass", storageClassName).Errorf("failed updating volume limits for volume with unbound PVC, %w", err)
```

Prefer this over calling `With()` twice

jonathan-innis (approved):
/lgtm
/approve

PR #1048, opened 2024-02-27
feat: support custom instance types with the KWOK provider

### LINKED ISSUE #987: Non-hard-coded list of instances for the KWOK provider
### Description

I'm the maintainer of the [SimKube](https://github.com/acrlabs/simkube) project which is designed to record-and-replay traces or event streams in a simulated Kubernetes cluster to identify misconfigurations or possible improvements in the k8s control plane.  One of the things I'd like to be able to do is a simulated comparison between the Kubernetes Cluster Autoscaler and Karpenter -- in order to do this, I need Karpenter to be able to launch simulated nodes (via KWOK) that match the real Node objects in a cluster.

In the [Cluster Autoscaler kwok cloudprovider](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/kwok), users can specify a list of [templated node objects as a ConfigMap](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/templates/configmap.yaml#L42)that the KWOK provider uses to build autoscaler NodeGroups.  As a first pass, this seems like it could be an OK solution for the Karpenter KWOK provider too.

drmorr0:
Fixes #987 

**Description**

This extends the KWOK provider for Karpenter by allowing users to specify custom lists of instance types that will be launched by Karpenter/KWOK.

Manual testing in a local kind+kwok cluster; I confirmed that, with the provided `kwok/examples/instance_types.json`, that an nginx deployment with zonal pod topology spread enabled worked correctly:

```
> cat test.k8s.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: test
  name: test-depl
  namespace: simkube
spec:
  replicas: 3
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
        - env:
            - name: POD_OWNER
              value: test-depl
          image: nginx:latest
          name: nginx
          resources:
            requests:
              cpu: "1"
      nodeSelector:
        type: virtual
      tolerations:
        - effect: NoSchedule
          key: kwok-provider
          value: "true"
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: "topology.kubernetes.io/zone"
          labelSelector:
            matchLabels:
              app: test
> kubectl get nodeclaims
NAME                 TYPE         ZONE         NODE                        READY   AGE
kwok-default-9szxv   g5.4xlarge   us-east-1c   strange-swirles-833325525   True    4m41s
kwok-default-twk5m   g5.4xlarge   us-east-1b   strange-joliot-1822046751   True    4m41s
kwok-default-vrcqw   g5.4xlarge   us-east-1a   kind-ganguly-3882149666     True    4m41s
> kubectl get nodes
NAME                        STATUS   ROLES           AGE     VERSION
karpenter-control-plane     Ready    control-plane   15d     v1.29.0
karpenter-worker            Ready    <none>          15d     v1.29.0
kind-ganguly-3882149666     Ready    <none>          5m18s   kwok-v0.4.0
strange-joliot-1822046751   Ready    <none>          5m18s   kwok-v0.4.0
strange-swirles-833325525   Ready    <none>          5m18s   kwok-v0.4.0
> kubectl get pods
NAMESPACE   NAME                             READY   STATUS    RESTARTS   AGE
simkube     test-depl-c5f8ddd84-h5tzw        1/1     Running   0          5m58s
simkube     test-depl-c5f8ddd84-nnz9x        1/1     Running   0          5m58s
simkube     test-depl-c5f8ddd84-xshgx        1/1     Running   0          5m58s
```

Here are the relevant logs from karpenter:

```
{"level":"INFO","time":"2024-02-27T23:00:02.648Z","logger":"controller.provisioner","message":"computed new nodeclaim(s) to fit pod(s)","commit":"7b5bd17-dirty","nodeclaims":3,"pods":3}
{"level":"INFO","time":"2024-02-27T23:00:02.666Z","logger":"controller.provisioner","message":"created nodeclaim","commit":"7b5bd17-dirty","nodepool":"kwok-default","nodeclaim":"kwok-default-9szxv","requests":{"cpu":"1","pods":"2"},"instance-types":"g5.4xlarge"}
{"level":"INFO","time":"2024-02-27T23:00:02.667Z","logger":"controller.provisioner","message":"created nodeclaim","commit":"7b5bd17-dirty","nodepool":"kwok-default","nodeclaim":"kwok-default-vrcqw","requests":{"cpu":"1","pods":"2"},"instance-types":"g5.4xlarge"}
{"level":"INFO","time":"2024-02-27T23:00:02.667Z","logger":"controller.provisioner","message":"created nodeclaim","commit":"7b5bd17-dirty","nodepool":"kwok-default","nodeclaim":"kwok-default-twk5m","requests":{"cpu":"1","pods":"2"},"instance-types":"g5.4xlarge"}
{"level":"INFO","time":"2024-02-27T23:00:02.683Z","logger":"controller.nodeclaim.lifecycle","message":"launched nodeclaim","commit":"7b5bd17-dirty","nodeclaim":"kwok-default-vrcqw","provider-id":"kwok://kind-ganguly-3882149666","instance-type":"g5.4xlarge","zone":"us-east-1a","capacity-type":"on-demand","allocatable":{"cpu":"15900m","ephemeral-storage":"600Gi","m
emory":"65526Mi","nvidia.com/gpu":"1","pods":"110"}}
{"level":"INFO","time":"2024-02-27T23:00:02.685Z","logger":"controller.nodeclaim.lifecycle","message":"launched nodeclaim","commit":"7b5bd17-dirty","nodeclaim":"kwok-default-twk5m","provider-id":"kwok://strange-joliot-1822046751","instance-type":"g5.4xlarge","zone":"us-east-1b","capacity-type":"on-demand","allocatable":{"cpu":"15900m","ephemeral-storage":"600Gi",
"memory":"65526Mi","nvidia.com/gpu":"1","pods":"110"}}
{"level":"INFO","time":"2024-02-27T23:00:02.687Z","logger":"controller.nodeclaim.lifecycle","message":"launched nodeclaim","commit":"7b5bd17-dirty","nodeclaim":"kwok-default-9szxv","provider-id":"kwok://strange-swirles-833325525","instance-type":"g5.4xlarge","zone":"us-east-1c","capacity-type":"on-demand","allocatable":{"cpu":"15900m","ephemeral-storage":"600Gi",
"memory":"65526Mi","nvidia.com/gpu":"1","pods":"110"}}
```


kwok/README.md
```diff
@@ -47,12 +47,54 @@ spec:
 EOF
 ```
 
+## Specifying Instance Types
```
njtran, line 50: Would also be nice if we could just generate the existing hard-coded instance types and put that in the filesystem, so that it always takes in a file, but you don't have to go fishing for how you might want to define it.  The only complaint there would be if the number of zones changed, or became variable, a static file would need some extra propagation work to succeed.

drmorr0: Oh, sure, that should be easy to do.

drmorr0: @njtran Ok, I created a `kwok/tools/gen_instances.go` file and moved the code that previously generated the static/hardcoded list in here.  `make gen_instances` will re-run and update the auto-generated list, and the helm charts will create the configmap and wire it all up together.

You can see these changes in the second commit of the stack.  I tested it out locally and it still seems to work.

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,91 @@ import (
...
-		family = "s" // standard
-	case 8:
-		family = "m" // memory
-	default:
-		family = "e" // exotic
+// ConstructInstanceTypes create many unique instance types with varying CPU/memory/architecture/OS/zone/capacity type.
+func ConstructInstanceTypes() ([]*cloudprovider.InstanceType, error) {
+	instanceTypeFile := os.Getenv(kwokInstanceTypeFileKey)
+	if instanceTypeFile == "" {
+		return nil, fmt.Errorf("no instance type file specified")
```
njtran, line 57: Just checking, this is defaulted to the currently hard-coded file, right?

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,91 @@ import (
...
+	for res, q := range opts.Resources {
+		switch res {
+		case "cpu":
+			cpu = q.String()
+		case "memory":
+			memory = q.String()
 		}
 	}
-	return instanceTypes
+	family := opts.Name[0:1] // all of the three major compute clouds (AWS, Azure, GCP) use the first letter of the instance name as the family
```
njtran, line 100: In AWS, we call what you're talking about as category, where family encompasses the category + generation + other details. 

https://karpenter.sh/docs/concepts/scheduling/#:~:text=an%20instance%20category-,karpenter.k8s.aws/instance%2Dfamily,-g4dn

Is there a common string delimiter that we can use to infer what the family should be? If not, it's probably fine to just keep as is, and understand that we can't fully emulate all the CP instance types.

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,91 @@ import (
...
+		InstanceTypeLabelKey:   opts.Name,
+		InstanceSizeLabelKey:   parseSizeFromType(opts.Name, cpu),
+		InstanceFamilyLabelKey: family,
+		InstanceCPULabelKey:    cpu,
+		InstanceMemoryLabelKey: memory,
+	}
+
+	// if the user specified a different pod limit, override the default
+	opts.Resources = lo.Assign(v1.ResourceList{
+		v1.ResourcePods: resource.MustParse("110"), // Default number of pods on a node in Kubernetes
```
njtran, line 111: Curious why you want to set 110 for all nodes, especially if users can't modify this.

Previously, I had it scale by the CPU so that it generally trends higher for larger instances:
https://github.com/kubernetes-sigs/karpenter/pull/1048/files#diff-ad69f1eb7049340837b1afd0668ed8e66a7ed6e0d4632923776c6584c7598241L72

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,91 @@ import (
...
-								Zone:         zone,
-								Price:        lo.Ternary(ct == v1beta1.CapacityTypeSpot, price*.7, price),
-								Available:    true,
-							})
-						}
-					}
-					instanceTypes = append(instanceTypes, newInstanceType(opts))
-				}
-			}
+func parseSizeFromType(ty, cpu string) string {
```
njtran, line 79: can you add a comment to the top of the function rather than inside the function?
```suggestion
// parseSizeFromType will attempt to discover the instance 
// size if it matches a special AWS format.
// If not, fall back to the cpu value.
// This works for both Azure and GCP (and the "generic" instances
// generated by tools/gen_instances.go)
func parseSizeFromType(ty, cpu string) string {
```

kwok/main.go
```diff
@@ -35,9 +39,14 @@ func init() {
 }
 
 func main() {
-	ctx, op := operator.NewOperator()
+	instanceTypes, err := kwok.ConstructInstanceTypes()
+	if err != nil {
+		fmt.Printf("kwok provider failed: %v\n", err)
```
njtran, line 44: Thoughts on this? 
```suggestion
		fmt.Printf("failed to construct instance types: %v\n", err)
```

njtran (commented):
Super nice work! Love the simplifications you've made.

Makefile
```diff
@@ -95,4 +95,7 @@ download: ## Recursively "go mod download" on all directories where go.mod exist
 toolchain: ## Install developer toolchain
 	./hack/toolchain.sh
 
-.PHONY: help presubmit install-kwok uninstall-kwok build apply delete test deflake vulncheck licenses verify download toolchain
\ No newline at end of file
+gen_instances:
```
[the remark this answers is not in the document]
drmorr0: Done.

kwok/README.md
```diff
@@ -47,12 +47,54 @@ spec:
...
+
+```
+{
+    "capacityType": "(on-demand|spot)",
+    "zone": "<zone or region name>",
+    "price": <instance type price>,
+}
+```
+
+There is an example instance types file in [examples/instance\_types.json](examples/instance_types.json).
```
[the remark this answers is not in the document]
drmorr0: There is another (small, easily readable) file in kwok/examples/instance_types.json.  Since the link is relative to the README location I don't think the kwok prefix is needed?

kwok/charts/templates/config.yaml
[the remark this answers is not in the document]
drmorr0: TIL, this is nice, I will do this.

kwok/cloudprovider/const.go
```diff
@@ -28,15 +28,18 @@ const (
...
 
 	// Labels that can be selected on and are propagated to the node
+	InstanceTypeLabelKey   = Group + "/instance-type"
 	InstanceSizeLabelKey   = Group + "/instance-size"
 	InstanceFamilyLabelKey = Group + "/instance-family"
 	InstanceMemoryLabelKey = Group + "/instance-memory"
 	InstanceCPULabelKey    = Group + "/instance-cpu"
+
+	// Environment variables for configuration
+	kwokInstanceTypeFileKey = "KWOK_INSTANCE_TYPES_FILE"
```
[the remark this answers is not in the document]
drmorr0: no longer relevant now that we're embedding things.

kwok/cloudprovider/helpers.go
drmorr0: no longer relevant since we're embedding

kwok/cloudprovider/helpers.go
drmorr0: So, AWS uses `(family).(size)`, GCP uses `(machine series)-(configuration)-(cpu count)`, and Azure (annoyingly) uses `[Family] + [Sub-family]* + [# of vCPUs] + [Constrained vCPUs]* + [Additive Features] + [Accelerator Type]* + [Version]`.  So we have things like

AWS: c5d.9xlarge
GCP: c3d-standard-4
Azure: M416ms_v2

I think I'm going to split on `[.-]` and if that fails, fall back to "first character", which will also work for the generic instance types you created.

kwok/cloudprovider/helpers.go
drmorr0: Kubernetes defaults to a max of 110 pods per node if you don't override it, so I'm just using the same default here.  See the maxPods argument to kubelet:  https://kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

Users can override this by providing a different value in the resources section of the instance type options.  For example, in the `generic_instance_types.json` file, a `c-1x-amd64-linux` node sets the following:

```json
"resources": {
    "cpu": "1",
    "ephemeral-storage": "20G",
    "memory": "2Gi",
    "pods": "16"
}
```

So nodes of this type will only come up with 16 pods per node.

kwok/cloudprovider/cloudprovider.go
```diff
@@ -133,49 +132,39 @@ func (c CloudProvider) toNode(nodeClaim *v1beta1.NodeClaim) (*v1.Node, error) {
 	//nolint
 	newName = fmt.Sprintf("%s-%d", newName, rand.Uint32())
 
-	capacityType := v1beta1.CapacityTypeOnDemand
-	requirements := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.
-		Spec.Requirements...)
-	if requirements.Get(v1beta1.CapacityTypeLabelKey).Has(v1beta1.CapacityTypeSpot) {
-		capacityType = v1beta1.CapacityTypeSpot
-	}
+	requirements := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...)
```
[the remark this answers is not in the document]
drmorr0: The capacity type is specified inside the instance type offering.  It's not guaranteed that people are going to even create a set of offerings for their cluster that have spot instances in them, so we can't just assume that the capacity type will be spot.

(For example, in the kwok/examples/instance_types.json file, the only offerings that are specified are on-demand)

kwok/cloudprovider/cloudprovider.go
```diff
@@ -189,7 +178,19 @@ func (c CloudProvider) toNode(nodeClaim *v1beta1.NodeClaim) (*v1.Node, error) {
 	}, nil
 }
 
-func addInstanceLabels(labels map[string]string, instanceType *cloudprovider.InstanceType, nodeClaim *v1beta1.NodeClaim, capacityType, price string) map[string]string {
+func offeringMatchesRequirements(of *cloudprovider.Offering, requirements *scheduling.Requirements) bool {
```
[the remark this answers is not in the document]
drmorr0: Note here we're not checking if the offering is valid, we're checking if the offering under consideration matches the requirements given in the nodeclaim.  I've updated the function name and arguments to make this a bit clearer.

kwok/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
drmorr0: @njtran In (old) L139, the check is looking to see if the _nodeClaim_ allows/requests spot, not if there are spot instances available from the _offering_.  If the embedded list of instance types only has on-demand instances in it, the check in (old) L163 will fail because there is no offering that supports spot, however we've unconditionally said we will return a spot instance.

This is relevant even if the generated list of instance types includes spot, because not every zone has spot instances in it.  For instance, in us-east-1, r7a.xlarge is only available as spot in 5 of the 6 AZs, which means if we unconditionally declare up front that we need a spot instance based on the node claim, and then later select the AZ that doesn't have spot instances, it will fail even though the zone potentially could have launched an on-demand instance.

Does that make sense?

Makefile
```diff
@@ -24,11 +24,11 @@ build: ## Build the Karpenter KWOK controller images using ko build
 	$(eval IMG_REPOSITORY=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 1 | cut -d ":" -f 1))
 	$(eval IMG_TAG=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 1 | cut -d ":" -f 2 -s))
 	$(eval IMG_DIGEST=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 2))
-	
+
 
 # Run make install-kwok to install the kwok controller in your cluster first
 # Webhooks are currently not supported in the kwok provider.
-apply: verify build ## Deploy the kwok controller from the current state of your git repository into your ~/.kube/config cluster
+apply: gen_instance_types verify build ## Deploy the kwok controller from the current state of your git repository into your ~/.kube/config cluster
```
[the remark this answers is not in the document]
drmorr0: @njtran I'm a bit uncertain what to do here.  I was _hoping_ I could create a symlink and then the user could point the symlink somewhere else if they wanted to use a different set of instance types; however, go embed explicitly does not support symlinks.  

I think removing  the target as an `apply` dependency makes sense.  I could also see adding it into `make presubmit`.  On my system at least, this is the timing to run `make presubmit` from "clean" with and without gen_instance_types added:

### `make presubmit` without gen_instance_types:

```
> time make presubmit
...
1579.89 user
142.01 system
3:05.57 elapsed
```

### `make presubmit` with gen_instance_types:

```
> time make presubmit
...
1685.19 user
157.95 system
3:17.38 elapsed
```
### `make gen_instance_types` only:

```
> time make gen_instance_types
...
55.47 user
4.90 system
0:08.96 elapsed
```

So it's not negligible, but it's also not huge?  Idk.  I updated the PR and just left it out of both targets for now, but let me know if you'd like me to add it into `presubmit`.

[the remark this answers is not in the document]
njtran: Thanks for this. We've seen our make presubmit shoot up in time length recently, and I'd want to evaluate our testing to make sure we're doing better here. 55 seconds is quite a bit of time on the dev loop, so I think we can remove it. I'm not sure how often this file should be changing anyway.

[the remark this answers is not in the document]
drmorr0: Sounds good to me.  It's no longer in the `presubmit` or the `apply` dependencies.

drmorr0:
@njtran any other concerns or comments about this PR?  Did my response about the spot/on-demand logic make sense?

kwok/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
njtran: So the NodeClaim represents a list of potential options to the cloud provider for a given list of instance types, already considering any out of capacity errors. The NodeClaim itself shouldn't have the spot requirement here if there are no spot offerings available. This is just the function to translate the nodeclaim into a node, which is a selection of already vetted choices. Let me know if this makes sense.

kwok/cloudprovider/helpers.go
njtran: I think 110 as a default is fine, but I want to make sure that our larger instance types are more similar to real instance types. For instance, this is how AWS calculates max pods: https://github.com/awslabs/amazon-eks-ami/blob/master/files/eni-max-pods.txt

kwok/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
drmorr0: @njtran I don't believe this is true in the kwok provider, at least.  I created a branch off of `main` where the only change I made was commenting out the `v1beta1.CapacityTypeSpot` option on L90 of helpers.go (where we construct the list of static instance types), and then demonstrate that submitting a vanilla nginx pod causes karpenter to fail to scale up, because of the check that we're discussing.  Specifically, it fails with `failed to find offering spot/test-zone-a/c-128x-amd64-linux`.

I filed #1144 as a bug and include steps to reproduce the behaviour.

kwok/cloudprovider/helpers.go
njtran: That's fair enough. Good point!

kwok/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
njtran: Ah, thanks for pushing back on me on this. I got my wires crossed. Let's tag this issue that you opened as part of the PR description and we can close this out. Thanks!

njtran (commented):
/hold

njtran:
/lgtm
/approve
Will wait for @jonathan-innis to take a look

njtran:
/lgtm-cancel

@drmorr0 it looks like you have some gen_instance_types outputs that need to be committed. Can you add that in?

njtran:
/lgtm cancel

Oops got the command wrong

drmorr0:
Added the boilerplate to gen_instance_types.go

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,102 @@ import (
...
-	Name               string
-	Offerings          cloudprovider.Offerings
-	Architecture       string
-	OperatingSystems   sets.Set[string]
-	Resources          v1.ResourceList
-	InstanceTypeLabels map[string]string
+	Name             string                  `json:"name"`
+	Offerings        cloudprovider.Offerings `json:"offerings"`
+	Architecture     string                  `json:"architecture"`
+	OperatingSystems sets.Set[string]        `json:"operatingSystems"`
```
njtran, line 46: Can we change this to a list of strings rather than a set? Should make the Json blob a bit easier to reason about. 
```suggestion
	OperatingSystems []string        `json:"operatingSystems"`
```

Makefile
```diff
@@ -96,4 +96,7 @@ download: ## Recursively "go mod download" on all directories where go.mod exist
 toolchain: ## Install developer toolchain
 	./hack/toolchain.sh
 
-.PHONY: help presubmit install-kwok uninstall-kwok build apply delete test deflake vulncheck licenses verify download toolchain
\ No newline at end of file
+gen_instance_types:
```
jonathan-innis, line 99: I'd love to see this target part of the `presubmit` target so that we make sure that we always re-gen these instance types if anything changes in the gen script.

kwok/cloudprovider/cloudprovider.go
```diff
@@ -133,49 +132,37 @@ func (c CloudProvider) toNode(nodeClaim *v1beta1.NodeClaim) (*v1.Node, error) {
...
-		if !ok {
-			return nil, fmt.Errorf("failed to find offering %s/%s/%s", capacityType, zone, val)
-		}
-		// All offerings of the same capacity type are the same price.
-		price = offering.Price
-		if price < minInstanceTypePrice {
-			minInstanceTypePrice = price
+		availableOfferings := it.Offerings.Available().Compatible(requirements)
+
+		offeringsByPrice := lo.GroupBy(availableOfferings, func(of cloudprovider.Offering) float64 { return of.Price })
```
jonathan-innis, line 154: Do we need this logic to group all of the offerings by price like this? Can we consider just iterating through the available offerings that are compatible one-by-one and just doing the same logic -- e.g. finding the cheapest offering of the lot

kwok/cloudprovider/cloudprovider.go
```diff
@@ -133,49 +132,37 @@ func (c CloudProvider) toNode(nodeClaim *v1beta1.NodeClaim) (*v1.Node, error) {
...
 	req, found := lo.Find(nodeClaim.Spec.Requirements, func(req v1beta1.NodeSelectorRequirementWithMinValues) bool {
 		return req.Key == v1.LabelInstanceTypeStable
 	})
 	if !found {
 		return nil, fmt.Errorf("instance type requirement not found")
 	}
 
-	minInstanceTypePrice := math.MaxFloat64
 	var instanceType *cloudprovider.InstanceType
+	var bestOffering *cloudprovider.Offering
```
jonathan-innis, line 144:
```suggestion
	var cheapestOffering *cloudprovider.Offering
```

Is it "best" or "cheapest"? I could see this being "best" if we were working off something different from price, but since we are using price as an optimization proxy here, it seems like "cheapest" might be the more operative word here

kwok/tools/gen_instance_types.go
```diff
@@ -0,0 +1,121 @@
+/*
```
jonathan-innis, line 1: I'm figuring you're thinking about this, but it would be awesome if we had follow-ups to this PR that auto-gened test data from AWS and Azure cloudproviders as well so you could do a bunch of test simulating with data from those providers!

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,102 @@ import (
...
-							})
-						}
-					}
-					instanceTypes = append(instanceTypes, newInstanceType(opts))
-				}
-			}
+// parseSizeFromType will attempt to discover the instance size if it matches a special AWS format.
+// If not, fall back to the cpu value.  This works for both Azure and GCP (and the "generic" instances
+// generated by tools/gen_instances.go)
+func parseSizeFromType(ty, cpu string) string {
```
jonathan-innis, line 77: musing: What if you could just add these labels yourself? Rather than having to parse out the data from the instance type naming, it seems more flexible for anyone trying to integrate their own special form of instance types into this kwok provider if you just opened up a "labels" or "requirements" block for instance types and just allowed everyone to set their own custom "labels" for nodes that were launched or selected on with that instance type.

This seems good enough to me for now, but might be a cool follow-on if we could make this a bit more flexible.

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,102 @@ import (
...
+
+// parseFamilyFromType will attempt to discover the instance family from the node type.  The three major cloud providers
+// have the following formats:
+//
+// AWS   - [Family].[Size]
+// GCP   - [Family]-[Configuration]-[# of vCPUs])
+// Azure - [Family] + [Sub-family]* + [# of vCPUs] + ... + [Version]
+//
+// So here, we split on [.-], and if that fails, fall back to the first character of the instance type name.
+func parseFamilyFromType(ty string) string {
```
jonathan-innis, line 93:
```suggestion
func parseFamilyFromType(instanceType string) string {
```

nit: I'm assuming that `ty` is short for `type` but on first-read, I assumed that it was an acronym for something. Maybe just worth it to type out the whole `instanceType` to make the signature a bit easier to parse

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,102 @@ import (
...
+		return ty[0:1]
+	}
+	return familySplit[0]
+}
+
+func setDefaultOptions(opts InstanceTypeOptions) InstanceTypeOptions {
+	var cpu, memory string
+	for res, q := range opts.Resources {
+		switch res {
+		case "cpu":
```
jonathan-innis, line 105:
```suggestion
		case v1.ResourceCPU:
```

consider using the upstream names for these resource labels

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,102 @@ import (
...
+			cpu = q.String()
+		case "memory":
```
jonathan-innis, line 107:
```suggestion
		case v1.ResourceMemory:
```

same suggestion here

kwok/cloudprovider/helpers.go
```diff
@@ -30,78 +32,102 @@ import (
...
+	}
+	return familySplit[0]
+}
+
+func setDefaultOptions(opts InstanceTypeOptions) InstanceTypeOptions {
+	var cpu, memory string
+	for res, q := range opts.Resources {
+		switch res {
+		case "cpu":
+			cpu = q.String()
```
jonathan-innis, line 106: Does the string value here give you what you are looking for? I assumed that this would print out the string version of the `resource.Quantity` value, which may be different from the bytes value that we were using previously in the instance type label

kwok/main.go
```diff
@@ -35,9 +39,14 @@ func init() {
 }
 
 func main() {
-	ctx, op := operator.NewOperator()
+	instanceTypes, err := kwok.ConstructInstanceTypes()
+	if err != nil {
+		fmt.Printf("failed to construct instance types: %v\n", err)
```
jonathan-innis, line 44: Rather than a `fmt` statement, could we consider using our existing logger to call `Fatal`? `logging.FromContext(ctx).Fatal()`? This would have to obviously come after initializing the operator components, but I think that's a reasonable trade-off since that's still happening before actually starting the controllers

Makefile
drmorr0: @jonathan-innis I already discussed this with @njtran and he asked me to not include it in the `presubmit` target

kwok/cloudprovider/cloudprovider.go
drmorr0: Previous behaviour picked a random offering if there are multiple offerings of the same price.  I'm using the grouping logic to maintain parity with previous behaviour (which I think is probably correct?)

kwok/cloudprovider/helpers.go
drmorr0: I don't think there are strong reasons to print out the raw bytes value, I feel like this is mostly "informational"?  I think it's generally-speaking easier to read the resource.Quantity string value than trying to parse some very large number of bytes?

kwok/cloudprovider/helpers.go
drmorr0: I converted it to a v1.OSName.

kwok/cloudprovider/helpers.go
drmorr0: @jonathan-innis I thought about this and almost added the functionality, but "my" interpretation was that if users wanted to add their own labels that should be done as part of the NodePool config?

kwok/tools/gen_instance_types.go
drmorr0: I have one for AWS, I've done very little with Azure so don't have a generator for them yet.  Agree it would be nice!

drmorr0:
@jonathan-innis OK I think I've addressed the remaining comments in here?  I tested this again locally and it seems to work, and tests here pass as well.  If there are additional concerns, I'd kinda like to address those in follow-up PRs if that's OK?  There's been quite a bit of back-and-forth on this already.

Makefile
njtran: If we include the gen_instance_types in the `make presubmit` target, we'll end up overwriting anything that's in that file path (which is the path that we use for embedding)

kwok/cloudprovider/helpers.go
```diff
@@ -17,104 +17,131 @@ limitations under the License.
...
-		InstanceSizeLabelKey:   size,
-		InstanceFamilyLabelKey: family,
-		InstanceCPULabelKey:    fmt.Sprintf("%d", cpu),
-		InstanceMemoryLabelKey: fmt.Sprintf("%d", cpu*memFactor*1024),
+
+	return cpu
+}
+
+// parseFamilyFromType will attempt to discover the instance family from the node type.  The three major cloud providers
+// have the following formats:
```
njtran, line 85: I'd like to phrase it like this
```suggestion
// parseFamilyFromType will attempt to discover the instance family from the node type. 
// Some examples of well known cloud provider formats:
```

kwok/cloudprovider/helpers.go
```diff
@@ -17,104 +17,131 @@ limitations under the License.
...
+//
+// AWS   - [Family].[Size]
+// GCP   - [Family]-[Configuration]-[# of vCPUs])
+// Azure - [Family] + [Sub-family]* + [# of vCPUs] + ... + [Version]
+//
+// So here, we split on [.-], and if that fails, fall back to the first character of the instance type name.
+func parseFamilyFromType(instanceType string) string {
```
njtran, line 92: Since this user-input isn't validated, is it worth doing a quick length check so that we're not going to out of bounds panic here, since L97 assumes its non-empty?

kwok/tools/gen_instance_types.go
```diff
@@ -0,0 +1,119 @@
...
+					pods := lo.Clamp(cpu*16, 0, 1024)
+					opts := kwok.InstanceTypeOptions{
+						Name:             name,
+						Architecture:     arch,
+						OperatingSystems: []v1.OSName{os},
+						Resources: v1.ResourceList{
+							v1.ResourceCPU:              resource.MustParse(fmt.Sprintf("%d", cpu)),
+							v1.ResourceMemory:           resource.MustParse(fmt.Sprintf("%dGi", mem)),
+							v1.ResourcePods:             resource.MustParse(fmt.Sprintf("%d", pods)),
+							v1.ResourceEphemeralStorage: resource.MustParse("20G"),
```
njtran, line 87: Does this need to be Gi? 
```suggestion
							v1.ResourceEphemeralStorage: resource.MustParse("20Gi"),
```

njtran: Does this need to be Gi? Looks like you resolved.

drmorr0: Ah, whoops, my bad.

njtran (commented):
/lgtm
/approve

Good work! Thanks for committing this!

Issue #751, opened 2022-06-06
Mega Issue: Karpenter doesnt support custom resources requests/limit


prateekkhera:
## Version

**Karpenter**: v0.10.1

**Kubernetes**: v1.20.15

## Expected Behavior
Karpenter should be able to trigger an autoscale

## Actual Behavior
Karpenter isnt able to trigger an autoscale

## Steps to Reproduce the Problem
We're using Karpenter on EKS. We have pods that has custom resource requests/limits in their spec definition - `smarter-devices/fuse: 1`. Karpenter seems to not respecting this resource and fails to autoscale and the pod remains to be in pending state 

## Resource Specs and Logs

Provisioner spec

```
apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: default
spec:
  limits:
    resources:
      cpu: "100"
  provider:
    launchTemplate: xxxxx
    subnetSelector:
      xxxxx: xxxxx
  requirements:
  - key: karpenter.sh/capacity-type
    operator: In
    values:
    - on-demand
  - key: node.kubernetes.io/instance-type
    operator: In
    values:
    - m5.large
    - m5.2xlarge
    - m5.4xlarge
    - m5.8xlarge
    - m5.12xlarge
  - key: kubernetes.io/arch
    operator: In
    values:
    - amd64
  ttlSecondsAfterEmpty: 30
status:
  resources:
    cpu: "32"
    memory: 128830948Ki
```

pod spec

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: fuse-test
  labels:
    app: fuse-test
spec:
  replicas: 1
  selector:
    matchLabels:
      name: fuse-test
  template:
    metadata:
      labels:
        name: fuse-test
    spec:
      containers:
      - name: fuse-test
        image: ubuntu:latest
        ports:
          - containerPort: 8080
            name: web
            protocol: TCP
        securityContext:
          capabilities:
            add:
              - SYS_ADMIN
        resources:
          limits:
            cpu: 32
            memory: 4Gi
            smarter-devices/fuse: 1  # Custom resource
          requests:
            cpu: 32
            memory: 2Gi
            smarter-devices/fuse: 1  # Custom resource
        env:
        - name: S3_BUCKET
          value: test-s3
        - name: S3_REGION
          value: eu-west-1
```

karpenter controller logs:

[...]

PR #603, opened 2023-10-13
feat:  support unknown resources

universam1:
**Description**

Karpenter **cannot be used** on clusters where custom resources for pods are defined, such as device drivers like `/dev/fuse` used with Podman and many more (see references).

Following error is logged:
```
karpenter-778b9dbc4f-gk88t {"level":"ERROR",..."logger":"controller.provisioner","message":"Could not schedule pod, incompatible with provisioner \"default\", daemonset overhead={\"cpu\":\"562m\",\"memory\":\"758026799\",\"pods\":\"10\"}, no instance type satisfied resources {\"cpu\":\"1562m\",\"memory\":\"1831768623\",\"pods\":\"11\",\"smarter-devices/fuse\":\"1\"} and requirements karpenter.k8s.aws/instance-category In [c m r], karpenter.k8s.aws/instance-generation Exists >2, karpenter.k8s.aws/instance-hypervisor In [nitro], karpenter.k8s.aws/instance-size NotIn [medium micro nano small], karpenter.sh/capacity-type In [on-demand spot], karpenter.sh/provisioner-name In [default], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux], node.kubernetes.io/node-group In [primary] (no instance type has enough resources)"}
```

Here we add a flag to instruct Karpenter to ignore certain defined resources, which will allow the usage of Karpenter for these clusters.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: karpenter-global-settings
data:
  ignoredDeviceRequests: "smarter-devices/fuse,some-other-device"
```

Fixes https://github.com/aws/karpenter-core/issues/751
Fixes https://github.com/aws/karpenter/issues/2390
Fixes https://github.com/aws/karpenter/issues/2899
Fixes https://github.com/navvis-dev/karpenter/pull/3
Fixes https://github.com/aws/karpenter/issues/3535
Fixes https://gitlab.com/gitlab-org/gitlab-runner/-/merge_requests/3717
Fixes https://github.com/aws/karpenter-core/issues/308
Fixes https://github.com/aws/karpenter/issues/3315
Fixes https://github.com/aws/karpenter/issues/3693

**How was this change tested?**

This fork is run in dozen of production clusters.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1056, opened 2024-02-29
feat: add ignore resource requests arg

chomatdam:
### Description

This PR is inspired by #603 and a follow-up regarding mega issue #751 opened a bit less than 2 years ago, its goal is to unblock many Karpenter users interested in the short term to still provision new nodes despite an unknown custom resource being requested by a pending pod.

### Impact
- Function `resources.Fits()` has a new arg to accept the list of resource requests to ignore, being a method signature change it requires a tiny change in the provider stack (aws/azure) to pass this new option resolved through the application context: `options.FromContext(ctx).IgnoredResourceRequests.Keys`

- Unit tests around options updated
- Karpenter controller deployed on 3 different EKS clusters using KubeVirt, it's provisioning as expected new nodes for virtual machines requiring resource requests mentioned in the _new_ environment variable below:
```
# Option 'IgnoredResourceRequests'
IGNORED_RESOURCE_REQUESTS:   devices.kubevirt.io/kvm,devices.kubevirt.io/tun,devices.kubevirt.io/vhost-net
```

### Fixes
- #751
- #603 
- https://github.com/aws/karpenter-provider-aws/pull/2161

---

### DIFF
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -55,7 +55,7 @@ const (
 
 var (
 	// RestrictedLabelDomains are either prohibited by the kubelet or reserved by karpenter
-	RestrictedLabelDomains = sets.New(
+	RestrictedLabelDomains = sets.New[string](
 		"kubernetes.io",
 		"k8s.io",
 		Group,
@@ -63,7 +63,7 @@ var (
 
 	// LabelDomainExceptions are sub-domains of the RestrictedLabelDomains but allowed because
 	// they are not used in a context where they may be passed as argument to kubelet.
-	LabelDomainExceptions = sets.New(
+	LabelDomainExceptions = sets.New[string](
 		"kops.k8s.io",
 		v1.LabelNamespaceSuffixNode,
 		v1.LabelNamespaceNodeRestriction,
@@ -72,7 +72,7 @@ var (
 	// WellKnownLabels are labels that belong to the RestrictedLabelDomains but allowed.
 	// Karpenter is aware of these labels, and they can be used to further narrow down
 	// the range of the corresponding values by either nodepool or pods.
-	WellKnownLabels = sets.New(
+	WellKnownLabels = sets.New[string](
 		NodePoolLabelKey,
 		v1.LabelTopologyZone,
 		v1.LabelTopologyRegion,
--- a/pkg/apis/v1beta1/nodepool_validation_webhook_test.go
+++ b/pkg/apis/v1beta1/nodepool_validation_webhook_test.go
@@ -18,6 +18,7 @@ package v1beta1_test
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"strings"
 	"time"
 
@@ -28,7 +29,6 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"knative.dev/pkg/ptr"
 
 	. "sigs.k8s.io/karpenter/pkg/apis/v1beta1"
--- a/pkg/cloudprovider/fake/cloudprovider.go
+++ b/pkg/cloudprovider/fake/cloudprovider.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 	"sort"
 	"sync"
 
@@ -98,7 +99,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *v1beta1.NodeClaim
 	instanceTypes := lo.Filter(lo.Must(c.GetInstanceTypes(ctx, np)), func(i *cloudprovider.InstanceType, _ int) bool {
 		return reqs.Compatible(i.Requirements, scheduling.AllowUndefinedWellKnownLabels) == nil &&
 			len(i.Offerings.Compatible(reqs).Available()) > 0 &&
-			resources.Fits(nodeClaim.Spec.Resources.Requests, i.Allocatable())
+			resources.Fits(nodeClaim.Spec.Resources.Requests, i.Allocatable(), options.FromContext(ctx).IgnoredResourceRequests.Keys)
 	})
 	// Order instance types so that we get the cheapest instance types of the available offerings
 	sort.Slice(instanceTypes, func(i, j int) bool {
--- a/pkg/controllers/provisioning/scheduling/existingnode.go
+++ b/pkg/controllers/provisioning/scheduling/existingnode.go
@@ -19,6 +19,7 @@ package scheduling
 import (
 	"context"
 	"fmt"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -84,7 +85,7 @@ func (n *ExistingNode) Add(ctx context.Context, kubeClient client.Client, pod *v
 	// node, which at this point can't be increased in size
 	requests := resources.Merge(n.requests, resources.RequestsForPods(pod))
 
-	if !resources.Fits(requests, n.Available()) {
+	if !resources.Fits(requests, n.Available(), options.FromContext(ctx).IgnoredResourceRequests.Keys) {
 		return fmt.Errorf("exceeds node resources")
 	}
 
--- a/pkg/controllers/provisioning/scheduling/nodeclaim.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaim.go
@@ -17,7 +17,9 @@ limitations under the License.
 package scheduling
 
 import (
+	"context"
 	"fmt"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 	"strings"
 	"sync/atomic"
 
@@ -64,7 +66,7 @@ func NewNodeClaim(nodeClaimTemplate *NodeClaimTemplate, topology *Topology, daem
 	}
 }
 
-func (n *NodeClaim) Add(pod *v1.Pod) error {
+func (n *NodeClaim) Add(ctx context.Context, pod *v1.Pod) error {
 	// Check Taints
 	if err := scheduling.Taints(n.Spec.Taints).Tolerates(pod); err != nil {
 		return err
@@ -103,7 +105,7 @@ func (n *NodeClaim) Add(pod *v1.Pod) error {
 	// Check instance type combinations
 	requests := resources.Merge(n.Spec.Resources.Requests, resources.RequestsForPods(pod))
 
-	filtered := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, requests)
+	filtered := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, requests, options.FromContext(ctx).IgnoredResourceRequests.Keys)
 
 	if len(filtered.remaining) == 0 {
 		// log the total resources being requested (daemonset + the pod)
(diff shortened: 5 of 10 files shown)


jonathan-innis:
Our main concern with ignoring resource requests is that none of the scheduling simulations that we are going to do are going to be accurate and we could very well assign too many pods to an instance type that we launch, which would cause you to over-pay for instances on your cluster. We'd discover it as soon as the instance comes up, but every scheduling simulation that we would do wouldn't consider these resources. Most likely what this would result in is us continually consolidating your cluster because we'd keep thinking that we could pack pods more tightly than is actually possible.

@chomatdam Do you have any thoughts about how we could avoid this problem? One idea that's floated around is if we allowed users to specify resource requests assigned to sets of instance types through requirements in some ConfigMap or other CustomResource, that might solve it.

Realistically, if we are going to make informed scheduling decisions, we need to know _exactly_ how many resources we can expect when we launch a given instance type.

The other question: How does this change when [Dynamic Resource Allocation](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/3063-dynamic-resource-allocation) comes around?

chomatdam:
Hi Jonathan, thanks for taking the time and inviting me to the challenges this topic has 🙂 My recommendation would be splitting the problem into 2 parts.

### Short-term
Users are fully aware of the risk taken by ignoring a custom resource request, if they do so they _may_ already know in advance it won't cause any issue.

**Example:** 
If I'm targeting a metal instance size with a `NodePool`, I already know `devices.kubevirt.io/kvm: 1` will be available. By ignoring this custom request, my node auto scaler is not broken anymore, quick win, I get nodes provisioned for KubeVirt VMs.

### Long-term
I think we need more insights about people's use cases or at least being able to list those to understand what they need (KubeVirt/fuse/hugepages/...) in something like a table anyone can participate to:
1. Custom resource request(s) and/or the use case requiring it
2. Is it static data? dynamic / resolved based on API? can only be determined post launch?
3. any extra details / links

If I'm taking KubeVirt as an example:
1. KubeVirt (`devices.kubevirt.io/kvm`, `devices.kubevirt.io/tun`, `devices.kubevirt.io/vhost-net`, and more!)
2. A KubeVirt daemonset (virt-handler) starts on every node with `kubevirt.io/schedulable:true`, then the node gets annotated with its resource capabilities based on node system information / available binaries / etc...
3. [code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-controller/services/renderresources.go#L466) adding resource requirements to a VM pod / [code](https://github.com/kubevirt/kubevirt/tree/main/pkg/virt-handler/node-labeller) for the node labeler in the virt-handler daemonset based on data collected from the node

### Dynamic Resource Allocation
I have to spend more time on it reading the entire KEP but it sounds like the right "long long term solution" where everyone could meet in the middle by adhering to a k8s built-in interface taking care of this aspect - but some time will pass before being implemented/supported by everyone.

jonathan-innis:
> By ignoring this custom request, my node auto scaler is not broken anymore, quick win, I get nodes provisioned for KubeVirt VMs

The only hang-up that I have with this is all of the edge-cases that I've outlined above. Even if you know that you are only going to have a single resource that's going to be registered once the node comes up, Karpenter might churn a lot trying to get all of the pods scheduled, it's going to schedule all pods to that single node, that node is going to be too big for just a single pod, then it's going to launch another large node, schedule one pod, repeat. You find yourself over-spending vastly for the nodes that you have and then Karpenter's consolidation mechanism is going to keep trying to replace all of those nodes into a single node, but it's going to find out that it's unsuccessful in doing so over and over again. What's worse: Consolidation will delete all of the nodes on a multi-consolidation decision, which is going to leave you with more pods in a pending state than you want. You're basically never going to reach a steady state on the cluster with the "unlimited resource" approach.

Are there ways that we could use config to inform Karpenter about the resources that it needs to know about? For the kubevirt use-case, is it always a single resource across every node or is there a good amount of "dynamicness" to the number of resources that are assigned to each resource type.

chomatdam:
Trying to provide a bit more context, so if I'm doing a recap from the pending pod and node capacity perspectives:

### Pending pod - Resource requests a VM can ask:
#### Sample
```
    Requests:
      cpu:                            2
      devices.kubevirt.io/kvm:        1
      devices.kubevirt.io/tun:        1
      devices.kubevirt.io/vhost-net:  1
      ephemeral-storage:              50M
      memory:                         2302Mi
```
#### Full list of resource requests
- `devices.kubevirt.io/kvm` => based on `kubevirt` CR object, field `allowEmulation=false`
- `devices.kubevirt.io/tun` => based on VM spec, present if more than 1 network interface defined or `Spec.Domain.Devices.AutoattachPodInterface=true`
- `devices.kubevirt.io/vhost-net` => based on VM spec, present if network interface has field `model=VirtIO`
- `devices.kubevirt.io/vhost-vsock` => based on VM spec, `.Spec.Domain.Devices.AutoattachVSOCK`

### Node capacities (once the `virt-handler` daemonset is enriching it)
#### Sample
```
  capacity:
    cpu: "72"
    devices.kubevirt.io/kvm: 1k         <-
    devices.kubevirt.io/tun: 1k         <-
    devices.kubevirt.io/vhost-net: 1k   <-
    ephemeral-storage: 20959212Ki
    hugepages-1Gi: "0"
    hugepages-2Mi: "0"
    memory: 197674116Ki
    pods: "737"
```
We can see in the [KubeVirt code](https://github.com/kubevirt/kubevirt/blob/main/pkg/virt-handler/device-manager/device_controller.go#L105), these are 3 capacity values set with `maxDevices=1000`, I didn't narrow down in the code where exactly but for the `kvm` binary I would guess it checks the `/dev/kvm` filepath, if it's not there it defaults to qemu (software emulation) instead of kvm (hw) and put the capacity for it to zero.

As you mentioned, we could map to AWS instance types (23 *.metal existing), or ideally to the instance size ("metal"). What the virt-handler's logic does with its tools, syscalls, file path checks but we would hardcode capacities before provisioning for Karpenter whereas right now it's resolved dynamically post node provisioning. It feels a bit off to hardcode it but may be a good enough way "in the middle" that solves other use cases.

A random idea but already more complex and moving some ownership to your end for this use case (and likely not the priority), could be an AWS~Karpenter-owned pipeline where any open source project can submit a PR to run a node labeler that runs against X instance types and the output is captured and attached to an "extension ID", people can later reference with Karpenter

Everything mentioned is more involved in development time. I also want to mention the fact that Karpenter as a product, the choice is maybe not purely technical but also about having a way to overcome limitations preventing its adoption. A short-term fix can be considered, if the complexity of the problem to solve is consequent, while it's easy to deprecate/unplug it and the long-term fix has been identified. I'll try to look into the approach you proposed if I have time but no guarantees.

Regarding dynamic resource allocation, I still have to read it again I'm unclear on the ownership and communication between the resource driver and scheduler 🙂

uniemimu:
> schedule all pods to that single node, that node is going to be too big for just a single pod, then it's going to launch another large node, schedule one pod, repeat. You find yourself over-spending vastly for the nodes that you have and then Karpenter's consolidation mechanism is going to keep trying to replace all of those nodes into a single node, but it's going to find out that it's unsuccessful in doing so over and over again. What's worse: Consolidation will delete all of the nodes on a multi-consolidation decision, which is going to leave you with more pods in a pending state than you want. You're basically never going to reach a steady state on the cluster with the "unlimited resource" approach.

Chiming in to mention that we have the exact same need for ignored resources. Our controller creates extended resources when a new node is created, but as the pod specs have those resources in use, a new node is not created for a pending pod.

I used a more dirty hack to mitigate, namely I injected a near infinite amount of said resources inside the precompute() function of the cloudprovider types.go into the allocatable of the instance type. For a single resource, all you need is one line of code there, plus perhaps an import. But a fork is not a solution here, so it was more a case of "could it work, somehow" than a solution.

I have never seen what you describe above happening, but having said that, I have only tested with my own workloads that have well-defined cpu and memory requests and limits. Plus the Pods have node affinity based node selectors that help Karpenter to choose a good sized new node. The users can add such node affinity based selectors if necessary. I assume pod affinity would also work.

The well-defined native resources plus the possibility of adding said affinity based node selectors at least in my testing have ensured that the created node is not initially too big, nor have I seen Karpenter consolidating to a single big node. In our case a single bigger node would be sometimes perhaps even ok, but that depends on other factors such as cost etc.

To make it crystal clear, with "ignored or infinite extended resources" I have never seen:
- too big nodes created initially
- oscillating of node creations and deletions
- unnecessary consolidations

But before the use of extended resources, I have witnessed some of the above in certain corner cases when a modified secondary scheduler was used. The lesson learned was that the scheduler should be bone stock to match with how Karpenter simulates things.

DRA is no silver bullet. It is alpha, feature-gated, and the original version is slow due to relying on using the API-server a lot. Future versions may be faster, and perhaps suitable also for use with cluster autoscalers. Perhaps. In the current form, DRA **does not work with cluster autoscalers, including Karpenter**, we have tried it with Google's cluster autoscaler and it was a no-go. Changes are needed to the cluster autoscaler. In EKS you can't even get the feature-gates enabled. Even if you started with VMs, and enabled the feature-gates etc, you would still need to do quite a bit of work to enable DRA scheduling in Karpenter. Without changes it will just stop new node creation the same way as adding an unknown extended resource will do. You see, the original DRA will use a node level entity (basically a resource driver kubelet plugin daemonset) to figure out whether a device claim can be allocated there, but that plugin won't deploy before the node is created, and Karpenter does not create the node because it doesn't understand anything about DRA yet, so you end up in the same limbo. DRA is expected to evolve, but for the time being, forget about it.

I suggest to add a way to support ignored extended resources. If one day DRA comes to rescue, great, you could deprecate the ignorable resources then.

uniemimu:
> 
> Are there ways that we could use config to inform Karpenter about the resources that it needs to know about? For the kubevirt use-case, is it always a single resource across every node or is there a good amount of "dynamicness" to the number of resources that are assigned to each resource type.

In our use-case, which is not related to kubevirt, there is an amount of dynamicness to the amount of resources eventually assigned to nodes, but the number of new resource types is a constant (6). So far the Kubernetes native resources have themselves been sufficient for Karpenter to work well-enough, even if not ideally, while ignoring the 6 types.

In our case, perhaps the amount of these unknown resources could be described as a more or less linear function of the allocatable native resources (size of node). At least an approximation could be created with such. But for sure, it isn't a case of always constant 1 per node or more generically constant N for every node regardless of the instance type. Not like that. Thus I opted to try int64 maximum as my mitigation, and it worked sufficiently well.

Ideally one would somehow give to Karpenter the correct extended resource amounts of new nodes based on the node instance type and any configured labels from the node pool. Meaning a list of extended resource definitions that match to an instance type name and a set of other labels. Generally, a list of extended resource definitions matching a set of node labels. The users could perhaps generate such, according to some new custom resource definition. But given the amount of instance types, it could be a long list to generate. In any case, such a CRD would allow Karpenter then to know how much extended resources that instance type with those labels from the node pool is eventually going to have.

The use-cases where the extended resource amounts are actually constants for every node regardless of node instance type, could perhaps be covered by just a single instance of such an extended resource list, where there just wouldn't be any labels limiting the scope of instance types that it applies to. And using int64 max value as the constant resource amount would be just another way to express "ignoring", potentially.

But lacking that, ignoring the resources as seen implemented in this patch seems like a way to proceed in the short term.

josmo:
for my two cents: Ignoring the custom resources on containers sounds like a reasonable short-term approach which solves the issues for numerous (not all of course) folks. It allows people to opt into the risk until a more sustainable solution is available, and this has been an issue for quite a while. Moreover, this provides a straightforward way to transition and deprecate the workaround once the long-term solution is implemented, minimizing project risks to maintaining extra complexity. Any chance we could get this in sooner vs later?

chomatdam:
@jonathan-innis thinking again about it, solutions don't necessarily have to be exclusive, one or the other, we could see it as "Depending on the use case, do we want to include (based on a mapping) or exclude custom resource requests"

- there are 765 AWS instance types (65 metal), in some cases ignoring a **capability** may be a valid option. As for KubeVirt, having a node pool targeting metal instances for hardware-based virtualization where we know the "kvm" custom resource request would always be there
- in some cases, a bounded scope of instance types where we need to set a specific **capacity** with a quantity for a custom resource request

universam1:
> The only hang-up that I have with this is all of the edge-cases that I've outlined above. 

@jonathan-innis In all respect to the effort put into this project, please lets come to some progress finally. The many users running this type of fork over years (!) proof that the concerns are acceptable and managable. ClusterAutoscaler does it and well. Put it behind a dangerous flag, but let's unblock us finally.

PR #1070, opened 2024-03-06
fix: Changing the MaxInstanceTypes back to original value of 60 in order to…

nikmohan123:
Fixes: https://github.com/aws/karpenter-provider-aws/issues/5776

**Description**
`MaxInstanceTypes` was changed from 60 to 100 in version `0.35.0` in order to send more instance types to launch API but that resulted in the request size to be too large for the cases where the requests are highly flexible with multiple zones in large regions leading to validation error from the launch API. This caused unscheduled pods in the cluster. So, we have reverted this particular value back to 60.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
@@ -31,8 +31,6 @@ import (
 
 // MaxInstanceTypes is a constant that restricts the number of instance types to be sent for launch. Note that this
 // is intentionally changed to var just to help in testing the code.
-// Increasing this value needs caution as it also requires stress testing the cloud provider API to launch so many instances
-// with various other combinatorial parameters such as zones, tags etc. in the largest region.
 var MaxInstanceTypes = 60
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -31,7 +31,9 @@ import (
 
 // MaxInstanceTypes is a constant that restricts the number of instance types to be sent for launch. Note that this
 // is intentionally changed to var just to help in testing the code.
-var MaxInstanceTypes = 100
+// Increasing this value needs caution as it also requires stress testing the cloud provider API to launch so many instances
```
jonathan-innis, line 34: This is relevant only to AWS. I'd just leave the detail in the PR description and not include the comment

nikmohan123: Hmm, I will remove this but this could be true for any cloud provider right ?

jonathan-innis (approved):
/lgtm
/approve

PR #1074, opened 2024-03-06
cherrypick: fix: Changing the MaxInstanceTypes back to original value of 60 in order to reduce the request size to be sent to launch APIs.order t

nikmohan123:
Fixes: https://github.com/aws/karpenter-provider-aws/issues/5776

**Description**
`MaxInstanceTypes` was changed from 60 to 100 in version `0.35.0` in order to send more instance types to launch API but that resulted in the request size to be too large for the cases where the requests are highly flexible with multiple zones in large regions leading to validation error from the launch API. This caused unscheduled pods in the cluster. So, we have reverted this particular value back to 60.

`make presubmit`


jonathan-innis (approved):
/lgtm
/approve

njtran (commented):
/lgtm
/approve

PR #1075, opened 2024-03-06
fix: fix spurious logging for non-CSI volumes

### LINKED ISSUE #1069: v0.34.1 reproducible StorageClass bug: false positive error, misleading error message, and a suggestion
### Description

(Migrated from https://github.com/aws/karpenter-provider-aws/issues/5757), per request of Karpenter Maintainer @jmdeal 

### Versions:
- Chart: https://gallery.ecr.aws/karpenter/karpenter
- Chart Version: v0.34.1 (Currently latest)
- Kubernetes Version: 1.29.0 (probably not relevant)

### Summary of Bug: 
* It's one of many storageclass edge case bugs
* Since it's an edge case, I've put together detailed notes & IaC that should allow easy E2E reproducibility :), within < 2 hours, that's 1 hour to read this, ~20 min of hands on, and ~40 min of waiting for infra to spawn.
* 3 main issues:
   * False Positive Error message
   * Misleading Error message makes it sound like Karpenter is trying to edit a PVC or PV
   * Suggestion: While researching this, I've found multiple similar reports. I believe it's something of a whack a mole issue that keeps reappearing. (a fix gets implemented to fix 1 edge case, but that just results in the creation of a new edge case.) I theorize the reason the issue is so hard to permanently fix, is the approach used to deal with the problem. So further down I'll explain a suggestion for another approach that could be used either in addition to or instead of the current edge case riddled method of dealing with it.

### Observed Behavior:
* If you follow the reproducibility notes to have Terraform provision EKS with Karpenter, and deploy a manually provisioned a PVC & PV (edge case scenario of not using a StorageClass), then scale workload to trigger karpenter to scale up a node, something like the following will show in logs:
  * `"failed updating volume limits for volume with unbound PVC, no storage class specified"`
  * Full Log for Context:  
    `{"level":"ERROR","time":"2024-02-28T03:39:03.801Z","logger":"controller.state.pod","message":"failed updating volume limits for volume with unbound PVC, no storage class specified","commit":"17d6c05","pod":"default/nfs-client-59d9f46ff6-dhhdx","pod":"nfs-client-59d9f46ff6-dhhdx","volume":"pvc","pvc":"nfs-client-pvc"}`
* It seems to be a **False Positive Error message**, because the scale up of a new node will happen correctly, completely ignoring the error like it was never a problem.
* **Misleading Error Message:**
  * The above error message is based on the reproducible example code and at first glance this bug appears to mostly be a cosmetic issue, that's not that important to fix.  
    I'd like to share the short version of a real world story, to achieve 2 goals: 
     1. give some context of why fixing it matters
     2. give context on why the error message should be changed.
  * A team of engineers was adopting/onboarding Karpenter for the first time, and between editing Karpenter and Terraform code an EKS security group got misconfigured, (you know that thing that can break EKS's networking in a way that's hard to figure out and debug, because it can break networking in a way where it's only half broken.) When that bug was encountered Karpenter wasn't working, because Kubernetes networking was broken, because an EKS security group was misconfigured, but we didn't know that at the time. What we saw was that Karpenter wasn't working, and we observed that Karpenter was giving an error message like the above. So the error message showing up when it shouldn't have lead us to misattribute the error, and waste 2 days messing with PVCs and reading up on Karpenter. 
  * If we never saw the false positive error, we wouldn't have gone on a wild goose chase, and would have realized the real error sooner.
  * Myself and other engineers misinterpreted the way that `"failed updating volume limits for volume with unbound PVC, no storage class specified"` is worded as meaning:  
    Karpenter had experienced an error when it failed to update/edit a PVC or PV. (we thought it was trying to add volume limits to a PVC.)
  * That was very concerning, as I was like Karpenter has no business editing a PVC, read I can understand, but edit, no. That prompted me and another engineer to do a deep dive learning session to research / fully understand how karpenter works. I now realize after reviewing the source code that it was only reading the PVC/PV and not attempting to make any edits. I now understand that Karpenter was updating it's understanding of PVC scheduling constraints, but I only came to that conclusion after alot of research of docs, issue tickets, and source code.

**Expected Behavior**:
* The fact that karpenter scales up a node, tells me things worked / there's no actual issue. 
If there's no actual issue, then no error message should appear in logs, a warning message would be fine.
* Ambiguous Error Messages that can be misinterpreted; specifically, "failed updating volume limits for volume with unbound PVC, no storage class specified" should be avoided in favor of clear easy to understand messages. 
* A better alternative would be something like: 'Error: karpenter has failed to detect CSI volume limit, associated with pvc "$NAME", this is used as a scheduling constraint. It's normally found by reading pvc, pv, sc, and csinode objects.'

------------------------------------------------------------------------------------------

### Useful Background Context
* It seems like https://github.com/aws/karpenter-provider-aws/issues/4574 might be related, the error message is a little different, and this has more details than that. I think fixing this will fix that.
* After doing 3 days of deep dive research, I know understand that Karpenter works like this:
  * Karpenter is designed around the concept of layered constraints.
  * All pods follow the pattern of being constrained by configuration and constraints that are embedded in the EC2NodeClass & NodePool custom resources. For example: A NodePool will reference a EC2NodeClass, which might use a subnet tag selector, and if you describe the EC2NodeClass for it's status you'll might see it select 3 subnets/AZs that a Karpenter NodeClaim could be provisioned into.
  * Something I learned from doing a deep dive is that Stateful Pods mounting PVCs are given more than 1 addition layer of constraints.
    * One layer was obvious, imposed by AWS, and is specific to pods using PVCs backed by EBS CSI driver. EC2 instances can't do cross AZ mounts of EBS volumes, so Karpenter needs to know what AZ to provision a new node in. Karpenter dynamically determines the AZ constraint of such pods at runtime:
      * If the PV already exists, karpenter will check the PVs metadata to see what AZ to spin up the node in.
      * If the PV doesn't exist, karpenter will check the storage class, in the case of ebs.csi.aws.com provisioner, that often has a subset of allowed AZs. [Karpenter's Docs](https://karpenter.sh/docs/concepts/scheduling/#persistent-volume-topology) give an example where the storage class further constrains stateful nodes to 2 AZs. 
    * ^-- This part was all review to me and made sense. **I kept asking myself after reading like 20 issue tickets related to storage class edge cases, Why does karpenter care so much about ALL storage classes?** (There are plenty of examples like EFS, NFS, Rook-Ceph, Linstor's-DRBD, etc. that can mount across AZs, so why care so much about detecting all storageclasses?)
    * That's when I learned [PVCs and storage classes have a 2nd layer of constraint imposed by the CSI spec](https://kubernetes-csi.github.io/docs/volume-limits.html). "Some (most), storage providers may have a restriction on the number of volumes that can be used in a Node. ... To support volume limits in a CSI driver, the plugin must fill in max_volumes_per_node" So most storage classes have a CSI driver that imposes an additional scheduling constraint, and that's why karpenter cares about accurately detecting all storageclasses.

### Proposal to Consider when implementing solution
* **Pointing out a problem (for context):** 
  * There's like 20+ issues related to storage class edge case bugs, they've long existed, and keep popping up.
  * From what I can tell there are more than a few cases like issue [#3647](https://github.com/aws/karpenter-provider-aws/issues/3647) where someone would open an issue to report an edge case bug, that was then solved by improving kaprenter's ability to auto-detect storage classes. Thing is there's been a few times where it seems like fixing 1 edge case results in a new edge case that created a new bug [(sig issue 664)](https://github.com/kubernetes-sigs/karpenter/issues/664), that then resulted in new logic for non-existant StorageClasses, but then that fix also fixed one edge case while creating a new edge case. Then logic had to be added for all kinds of edge cases "", (storageclass like "manual", which could be listed in pvc, but a non-existant actual storage class), this was all iteratively figured out over time, and got complex enough to add some volume related unit tests. 
  * The logic for handling all these edge cases is complicated enough that it's no surprize that there's bugs.
* **Suggestion of another way to handle the problem**
  * Karpenter currently dynamically figures out the scheduling constraints of stateful workloads that mount PVCs, by reading pvc, pv, sc, and csinode objects. 
  * Karpenter doesn't seem to have a solid strategy for dealing with edge cases like local storage class, other instances of using no-provisioner, and just manual PVC assignment without a storage class, like shown in the below bug reproduction steps.
     ```yaml
     apiVersion: storage.k8s.io/v1
     kind: StorageClass
     metadata:
       name: local-storage
     provisioner: kubernetes.io/no-provisioner
     ```
  * My suggestion is that rather than continueing to complicate the storage class auto detection logic even further. It might be worth considering a new approach that could be used to compliment the current auto detection methodology, (and it could be used to deal with the remaining edge cases vs try to perfect the auto detection further with more complexity.)
  * Instead of trying to dynamicaly auto detect every possible edge case, what if you just take an approach of surfacing a warning message in karpenter logs like "Error: karpenter detected an edge case when trying to schedule pvc xyz, please see linked doc, on how to configure edge case resolution"
  * Then that linked doc could give tips on how karpenter auto detection works, how to improve karpenter's ability to do auto detection, and offer an [optional configmap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#optional-configmaps) that can be used to pass explicit instructions on how you'd like karpenter to handle this edge case. (like if it detects storage class local-storage, but it get's confused about unspecified volume limit of no-provisioner, the configmap can let the user specify a volume limit to use when dynamic detection of constraints fails / when that edge case is detected)
  * The reason I suggest an optional configmap, is I think it's not uncommon for users to use something like awscdk, or terraform and just say karpenter enabled give me explicitly pinned version of current latest, and not want to need to learn how to override the default helm values. An optional configmap would be easier to patch in some edge case config without having to customize default helm chart install, and thus result in a better UX. 

------------------------------------------------------------------------------------------

### Easy E2E Reproduction Steps (TF, YAML, notes, etc.)
1.) Download TF's EKS Karpenter Example (The git commit 19b79df, has little significance, it was latest as of Feb 29th, just listing it for reproducibility sake)
```bash
cd ~
git clone https://github.com/terraform-aws-modules/terraform-aws-eks.git
cd terraform-aws-eks
git checkout 19b79df
cd examples/karpenter
```
2.) edit line 172 of main.tf, which refers to the version of the karpenter helm chart   
      `version             = "v0.34.1"`
3.) deploy the EKS cluster, takes ~15min
```bash
terraform init
terraform apply
```
4.) Make sure karpenter's running.  
(Note there's a bug in the Terraform example where the karpenter pods won't run, because karpenter namespace is configured to be backed by fargate notes, but karpenter pods don't tolerate that, we'll fix it with a quick patch.)
```bash
aws eks --region eu-west-1 update-kubeconfig --name ex-karpenter

tee toleration.patch << EOF
spec:
  template:
    spec:
      tolerations:
      - effect: NoSchedule
        key: eks.amazonaws.com/compute-type
        value: fargate
EOF

kubectl patch deploy karpenter -n=karpenter --patch-file toleration.patch
kubectl scale deploy karpenter -n=karpenter --replicas=1
kubectl get pods -A
kubectl get deploy karpenter -n=karpenter -o yaml | grep -i image
# image: public.ecr.aws/karpenter/controller:v0.34.1@sha256:94...
```
5.) Deploy nfs-server that has zero dependencies
```bash
tee nfs-server.yaml << EOF
apiVersion: v1
kind: Service
metadata:
  name: nfs-server
spec:
  ports:
  - name: nfs
    port: 2049
  selector:
    app: nfs-server
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nfs-server
spec:
  replicas: 1
  selector:
    matchLabels:
      role: nfs-server
  template:
    metadata:
      labels:
        role: nfs-server
    spec:
      nodeSelector:
        kubernetes.io/arch: amd64 #The image only works on x86_64, not arm64
      containers:
      - name: nfs-server
        image: k8s.gcr.io/volume-nfs:0.8
        resources:
          requests:
            memory: "1Gi"
            cpu: "1"
          limits:
            memory: "1Gi"
        ports:
        - name: nfs
          containerPort: 2049
        securityContext:
          privileged: true
        volumeMounts:
        - name: node-local-storage
          mountPath: /exports
      volumes:
      - name: node-local-storage
        hostPath:
          path: /data/nfs
          type: DirectoryOrCreate
---
EOF

kubectl apply -f nfs-server.yaml
```
6.) get nfs-server's pod IP, and store it into a variable
```bash
export NFS_POD_IP=$(kubectl get pod -l=role=nfs-server -o=jsonpath="{range .items[*]}{.status.podIP}")
echo $NFS_POD_IP
```
7.) After making sure NFS_POD_IP's value seems reasonable, deploy nfs-client
```bash
tee nfs-client.yaml << EOF
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: nfs-client-pvc
  namespace: default
spec:
  accessModes:
  - ReadWriteMany
  resources:
    requests:
      storage: 5Gi
  storageClassName: ""
  volumeMode: Filesystem
  volumeName: nfs-client-pv
---
apiVersion: v1
kind: PersistentVolume
metadata:
  name: nfs-client-pv
spec:
  accessModes:
  - ReadWriteMany
  capacity:
    storage: 5Gi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: nfs-client-pvc
    namespace: default
  mountOptions:
  - nfsvers=4.2
  nfs:
    path: / #Note: kubelet (worker node agent) is what tries to resolve the server. 
    server: $NFS_POD_IP #pod ip of nfs-server, (b/c kubelet can't resolve service of type cluster ip)
  persistentVolumeReclaimPolicy: Retain
  volumeMode: Filesystem
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nfs-client
spec:
  replicas: 1
  selector:
    matchLabels:
      name: nfs-client
  template:
    metadata:
      labels:
        name: nfs-client
    spec:
      containers:
      - image: busybox:latest
        resources:
          requests:
            memory: "512Mi"
            cpu: "500m"
          limits:
            memory: "512Mi"
        command: #based on https://github.com/mr-pascal/medium-k8s-nfs/blob/master/nfs-busybox-deployment.yaml
        - sh
        - -c
        - 'while true; do date > /mnt/index.html; hostname >> /mnt/index.html; sleep $(($RANDOM % 5 + 5)); done'
        imagePullPolicy: IfNotPresent
        name: busybox
        volumeMounts:
        - name: pvc
          mountPath: "/mnt"
      volumes:
      - name: pvc
        persistentVolumeClaim:
          claimName: nfs-client-pvc
---
EOF

kubectl apply -f nfs-client.yaml
```
8.) scale the nfs client to trigger a scale up
```bash
kubectl scale deploy/nfs-client -n=default --replicas=10
#^-- some replicas will be pending but that's enough to trigger a scale up.

kubectl logs deploy/karpenter -n=karpenter
#^-- will show the error
#v-- Then these will show it's a false positive / ignoreable error because they come up
kubectl get pods
kubectl get nodes

kubectl exec -it deploy/nfs-client -n=default -- cat /mnt/index.html
# ^-- side note if you periodically rerun this, you'll see they're all accessing the same volume
```

**Standard Boiler Plate:** 

jmdeal:
Fixes #1069

**Description**
Today, if we fail to resolve a CSI driver from a bound PV, we attempt to resolve it from the storage class on the PVC. This can lead to spurious error logs when the volume is not managed by a CSI driver or supported in-tree equivalent (currently only EBS). What we should do is bail early if the PVC is bound and no volume could be discovered. 

Some notes on logging:
- The bound PVC + no driver case isn't logged to avoid repetitive logging for a non-issue, I'd consider logging it if we had a `trace` log level but it feels too verbose even for debug.
- Logs that were previously errors have been reduced to debug messages. Neither of the cases are fatal errors that disrupt Karpenter's ability to schedule but they should still be surfaced because they could indicate to the customer they have a misconfiguration. 
  - The case where the PVC doesn't exist is still kept as an error since it shouldn't be a valid state (i.e. would have needed to be removed with the finalizer patched off). I'm not certain what this one should be though, definitely would like some other opinions here.
- The log messages have been adjusted to ensure that it is clear Karpenter is not updating values on in-cluster resources, rather it is updating it's own cluster state.

`make test` / manual

### DIFF
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -87,7 +87,7 @@ func GetVolumes(ctx context.Context, kubeClient client.Client, pod *v1.Pod) (Vol
 		// computing limits, otherwise Karpenter may never be able to update its cluster state.
 		if err != nil {
 			if errors.IsNotFound(err) {
-				logging.FromContext(ctx).With("pod", pod.Name, "volume", volume.Name).Errorf("failed updating volume limits for volume, %w", err)
+				logging.FromContext(ctx).With("pod", pod.Name, "volume", volume.Name).Errorf("failed tracking CSI volume limits for volume, %w", err)
 				continue
 			}
 			return nil, fmt.Errorf("failed updating volume limits, %w", err)
@@ -123,6 +123,8 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, pod *v1.Pod, v
 		if driverName != "" {
 			return driverName, nil
 		}
+		// The PVC is bound, but not to a volume managed by a CSI driver or a known in-tree equivalent. This PVC can be ignored for the purposes of volume limit tracking.
+		return "", nil
 	}
 
 	// This can occur in two scenarios:
@@ -131,7 +133,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, pod *v1.Pod, v
 	// In either of these cases, a PV must have been previously bound to the PVC and has since been removed. We can
 	// ignore this PVC while computing limits and continue.
 	if storageClassName == "" {
-		logging.FromContext(ctx).With("pod", pod.Name, "volume", volumeName, "pvc", pvc.Name).Errorf("failed updating volume limits for volume with unbound PVC, no storage class specified")
+		logging.FromContext(ctx).With("pod", pod.Name, "volume", volumeName, "pvc", pvc.Name).Debugf("failed tracking CSI volume limits for volume with unbound PVC, no storage class specified")
 		return "", nil
 	}
 
@@ -142,7 +144,7 @@ func resolveDriver(ctx context.Context, kubeClient client.Client, pod *v1.Pod, v
 		//  2. The StorageClass never existed and was used to bind the PVC to an existing PV, but that PV was removed
 		// In either of these cases, we should ignore the PVC while computing limits and continue.
 		if errors.IsNotFound(err) {
-			logging.FromContext(ctx).With("pod", pod.Name, "volume", volumeName, "pvc", pvc.Name, "storageclass", storageClassName).Errorf("failed updating volume limits for volume with unbound PVC, %w", err)
+			logging.FromContext(ctx).With("pod", pod.Name, "volume", volumeName, "pvc", pvc.Name, "storageclass", storageClassName).Debugf("failed tracking CSI volume limits for volume with unbound PVC, %w", err)
 			return "", nil
 		}
 		return "", err


jonathan-innis:
> The bound PVC + no driver case isn't logged to avoid repetitive logging for a non-issue

I also think that the bound PVC + no driver case isn't really an issue. There are real use-cases for not using CSI and using something like local storage to provision your volumes for your pods.

jonathan-innis (approved):
/lgtm
/approve

jmdeal:
> there are real use-cases for not using CSI and using something like local storage to provision your volumes for your pods.

Agreed, that's what I meant by a non-issue. I'd consider logging it if we had a trace because it could potentially mean something is corrupted but there are too many valid use-cases to justify logging in debug.

PR #1091, opened 2024-03-11
ci: add go version to dep cache key

jmdeal:
Fixes #N/A 

**Description**
Adds the go version to the dependency cache key. This is to address an issue with govulncheck, where the same version of govulncheck built with 1.21 crashes when run against a 1.22 project: https://github.com/golang/go/issues/65590

- Run against github actions


jonathan-innis (approved):
/lgtm
/approve

Issue #909, opened 2023-12-24
How to handle unset `apiVersion` and `kind` in `nodeClassRef`


jonathan-innis:
### Description

**What problem are you trying to solve?**

As part of #337, we need to know the `apiVersion` and the `kind` of the `nodeClassRef`. This is also true when it comes to supporting #493 where we would want to roll-up NodeClass readiness into NodePool readiness so that we can determine if a NodePool can be provisioned with based on both whether the NodePool is in a "Ready" state as well as when the NodeClass is in a "Ready" state. You can imagine something like the example below when it comes to reasoning about NodeClass readiness and NodePool readiness.

##### NodeClass Readiness

```yaml
status:
  conditions:
  - type: Ready
    status: "True"
```

##### NodePool Readiness

```yaml
status:
  conditions:
  - type: Ready
    status: "True"
  - type: NodeClassReady
    status: "True"
```

Right now, there is no hard requirement to set `apiVersion` and `kind` in your `nodeClassRef` (only `name`) since the CloudProviders who are actually grabbing the `nodeClassRef` to pull true type for `GetInstanceTypes` and `Create` calls already know the type under the hood. For AWS, this type is _always_ `EC2NodeClass`; for Azure, this type is _always_ `AKSNodeClass` today. This works fine for now; however, this becomes problematic when we want to grab the underlying resource dynamically from the neutral code using the dynamic client, like we would need to do if we were trying to grab the readiness of the NodeClass to bubble-up that readiness into the NodePool.

We _need_ to have some way to infer the actual type and the `name` field isn't enough. Ideally, we would require that the `apiVersion` and `kind` _always_ be set in the NodePool's `nodeClassRef`; however, doing so out of the gate would be a breaking change and would break existing resources which don't already have this field set on the cluster. There are a couple thoughts that I had here around how to achieve this:

#### Solution 1: Runtime Defaults through `cloudprovider.go`

[...]

PR #1093, opened 2024-03-12
feat: Set default values on the `APIVersion` and `Kind` on the nodeClassRef

engedaam:
Fixes #N/A 

**Description**
- Add `GetNodeClassGVK` to the cloudprovider interface 
- The cloudprovider will give the `APIVersion` and `Kind` of the NodeClass that is supported by the provider. Karpenter will default all NodePools to the `APIVersion` and `Kind`, given by the provider, inside of each `nodeClassRef`. 
- To avoid a breaking change, we will use the hash controller to set the `APIVersion` and `Kind` at runtime.
- The team is considering on making all fields on the `nodeClassRef` required at v1, which will allow us to drop the defaulting behavior.
- Related to issue: https://github.com/kubernetes-sigs/karpenter/issues/909

- `make presubmit`


kwok/cloudprovider/cloudprovider.go
```diff
@@ -118,6 +118,13 @@ func (c CloudProvider) Name() string {
 	return "kwok"
 }
 
+func (c CloudProvider) GetNodeClassGVK() metav1.TypeMeta {
```
jonathan-innis, line 121:
```suggestion
func (c CloudProvider) GetSupportedNodeClasses() []schema.GroupVersionKind {
```

Consider using the following signature

pkg/apis/v1beta1/nodepool.go
```diff
@@ -194,7 +194,7 @@ type NodePool struct {
 // 1. A field changes its default value for an existing field that is already hashed
 // 2. A field is added to the hash calculation with an already-set value
 // 3. A field is removed from the hash calculations
-const NodePoolHashVersion = "v1"
+const NodePoolHashVersion = "v2"
```
jonathan-innis, line 197: This is assuming that we have a release with the previous version, right?

engedaam: Yes, defaulting on the nodeClassRef must come after the hash versioning changes have been released

engedaam:
/hold

We need to make sure that this PR is merged in after drift hash versioning have been released. Issue: https://github.com/kubernetes-sigs/karpenter/issues/957

kwok/cloudprovider/cloudprovider.go
```diff
@@ -118,6 +119,14 @@ func (c CloudProvider) Name() string {
 	return "kwok"
 }
 
+func (c CloudProvider) GetSupportedNodeClasses() schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group:   Group,
+		Version: "v1alpha1",
+		Kind:    "KOWKNodeClass",
```
njtran, line 126:
```suggestion
		Kind:    "KwokNodeClass",
```

kwok/cloudprovider/cloudprovider.go
```diff
@@ -118,6 +119,14 @@ func (c CloudProvider) Name() string {
 	return "kwok"
 }
 
+func (c CloudProvider) GetSupportedNodeClasses() schema.GroupVersionKind {
```
njtran, line 122: I don't think there's any need for us to have a KwokNodeClass. What were you thinking we would have in here? Is this just a placeholder because this information would be required? Do we need to enforce that this is non-zero length?

pkg/apis/v1beta1/nodepool.go
njtran: Can we add this as a `do-not-merge` in the PR description? I want to make sure I don't approve this and accidentally merge it before we have our ducks in a row.

njtran: Also, does this really need to be in two different releases? I can imagine that a user upgrades from version from before the hash versioning changes were released to this version with a v2, do we have that case covered?

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -53,6 +54,9 @@ type CloudProvider struct {
 
 	CreatedNodeClaims map[string]*v1beta1.NodeClaim
 	Drifted           cloudprovider.DriftReason
+	NodeClassGroup    string
+	NodeClassVersion  string
+	NodeClassKind     string
```
njtran, line 59: Is there a reason we're making this three variables rather than one variable of type schema.GroupVersionKind?

kwok/cloudprovider/cloudprovider.go
njtran, line 122: Shouldn't this also return a list to match the language of the function definition?

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -37,19 +39,33 @@ var _ operatorcontroller.TypedController[*v1beta1.NodePool] = (*Controller)(nil)
...
 }
 
 // Reconcile the resource
 func (c *Controller) Reconcile(ctx context.Context, np *v1beta1.NodePool) (reconcile.Result, error) {
 	stored := np.DeepCopy()
 
+	// To avoid a breaking change on the NodePool API, we will be setting a default APIVersion and Kind
+	// defined by the cloudprovider to each nodeClassRef on every NodePool. This will be removed once
+	// the NodePool API requires APIVersion and Kind to be set at NodePool creation.
+	// TODO: remove once APIVersion and Kind are required fields on NodePool
```
njtran, line 60: Is this at v1? Can you put that here if so?
```suggestion
	// TODO: remove at v1 when APIVersion and Kind are required fields on NodePool
```

pkg/apis/v1beta1/nodepool.go
engedaam: Discussed offline. We will be doing a patch release to each version of the v1beta1 to include drift hash versioning changes.

pkg/controllers/nodepool/hash/controller.go
```diff
@@ -37,19 +38,35 @@ var _ operatorcontroller.TypedController[*v1beta1.NodePool] = (*Controller)(nil)
...
 }
 
 // Reconcile the resource
 func (c *Controller) Reconcile(ctx context.Context, np *v1beta1.NodePool) (reconcile.Result, error) {
 	stored := np.DeepCopy()
 
+	// To avoid a breaking change on the NodePool API, we will be setting a default APIVersion and Kind
+	// defined by the cloudprovider to each nodeClassRef on every NodePool. This will be removed once
+	// the NodePool API requires APIVersion and Kind to be set at NodePool creation.
+	// TODO: remove at v1 when APIVersion and Kind are required fields on NodePool
```
njtran, line 59: I was thinking here, should we make it required, or should we make it optional and just require that cloud providers add the defaulting logic to include the GVK into NodePools? If we require it, we can't default it. Is there a point in allowing users to be wrong here?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -239,4 +244,62 @@ var _ = Describe("Static Drift Hash", func() {
...
+			Expect(nodePool.Spec.Template.Spec.NodeClassRef.Kind).To(Equal(cp.NodeClassGroupVersionKind[0].Kind))
+		})
+		It("should not set a cloudprovider default kind on a nodeclassref when kind is set", func() {
+			nodePool.Spec.Template.Spec.NodeClassRef.Kind = "ExistingKind"
+			ExpectApplied(ctx, env.Client, nodePool)
+			ExpectReconcileSucceeded(ctx, nodePoolController, client.ObjectKeyFromObject(nodePool))
+			nodePool = ExpectExists(ctx, env.Client, nodePool)
+			Expect(nodePool.Spec.Template.Spec.NodeClassRef.Kind).To(Equal("ExistingKind"))
+		})
+		It("should not set cloudprovider default for kind and apiversion if there are more then one supported node class", func() {
```
njtran, line 285:
```suggestion
		It("should not set cloudprovider default for kind and apiversion if there is more than one supported node class", func() {
```

jonathan-innis:
@engedaam Given that we are punting on setting the default GVK and we have already merged in the CloudProvider changes to allow for shorter reconciles, do we need to keep this PR open?

engedaam:
Closing for now. We plan on handling defaults for the APIVersion and Kind as part of the conversion webhooks at v1.

PR #1095, opened 2024-03-13
perf: Reduce delete calls by checking DeletionTimestamp

jonathan-innis:
Fixes #N/A 

**Description**

Check DeletionTimestamp when calling `Delete()`. If the DeletionTimestamp is already set, we don't need to call `Delete()` again. I noticed that we were over-calling Delete() when I was looking at kube-apiserver audit logs for the termination flow.

`make presubmit`


fmuyassarov (approved):
Looks good :)

Bryce-Soghigian:
🙌

PR #1096, opened 2024-03-13
perf: fix helm namespace for delete Makefile target

fmuyassarov:
Fixes #N/A

**Description**

The `delete` Makefile target was targeting the wrong namespace or rather expecting it to be passed as environment variable (without defaulting) while `install` target has the namespace hardcoded. This commit re-uses the same hardcoded namespace kube-system for delete target too.

`make delete`


Makefile
```diff
@@ -40,7 +40,7 @@ apply: verify build ## Deploy the kwok controller from the current state of your
 		--set-string controller.env[0].value=true
 
 delete: ## Delete the controller from your ~/.kube/config cluster
-	helm uninstall karpenter --namespace ${KARPENTER_NAMESPACE}
+	helm uninstall karpenter --namespace kube-system
```
jonathan-innis, line 43: Seems like we should just update both the `make apply` and the `make deploy` targets to use `$KARPENTER_NAMESPACE` and we should change `$SYSTEM_NAMESPACE` to `$KARPENTER_NAMESPACE`

fmuyassarov: Yes, we can do that too instead of hardcoding.

jonathan-innis (approved):
/lgtm
/approve

PR #1103, opened 2024-03-15
chore: Drop CloudProvider delete call from Node termination

jonathan-innis:
Fixes #N/A 

**Description**

Removes the call to `cloudProvider.Delete()` from the node termination controller. This hands-off responsibility of deleting the instance solely to the NodeClaim termination controller. This delete call was only in there for legacy reasons during the machine migration that could have occurred prior to v0.32.0

`make presubmit`


engedaam (commented):
/lgtm

PR #1125, opened 2024-03-21
fix: Adjust duration validation to include format for `1h10m`

engedaam:
Fixes https://github.com/aws/karpenter-provider-aws/issues/5913

**Description**
- We should be able to create a budget with a duration format of  `1h10m`

- N/A

### DIFF
--- a/pkg/apis/crds/karpenter.sh_nodepools.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodepools.yaml
@@ -80,7 +80,7 @@ spec:
                               This is required if Schedule is set.
                               This regex has an optional 0s at the end since the duration.String() always adds
                               a 0s at the end.
-                            pattern: ^([0-9]+(m|h)+(0s)?)$
+                            pattern: ^((([0-9]+(h|m))|([0-9]+h[0-9]+m))(0s)?)$
                             type: string
                           nodes:
                             default: 10%
--- a/pkg/apis/v1beta1/nodepool.go
+++ b/pkg/apis/v1beta1/nodepool.go
@@ -124,7 +124,7 @@ type Budget struct {
 	// This is required if Schedule is set.
 	// This regex has an optional 0s at the end since the duration.String() always adds
 	// a 0s at the end.
-	// +kubebuilder:validation:Pattern=`^([0-9]+(m|h)+(0s)?)$`
+	// +kubebuilder:validation:Pattern=`^((([0-9]+(h|m))|([0-9]+h[0-9]+m))(0s)?)$`
 	// +kubebuilder:validation:Type="string"
 	// +optional
 	Duration *metav1.Duration `json:"duration,omitempty" hash:"ignore"`
--- a/pkg/apis/v1beta1/nodepool_validation_cel_test.go
+++ b/pkg/apis/v1beta1/nodepool_validation_cel_test.go
@@ -165,7 +165,7 @@ var _ = Describe("CEL/Validation", func() {
 		It("should fail when creating a budget with a duration but no cron", func() {
 			nodePool.Spec.Disruption.Budgets = []Budget{{
 				Nodes:    "10",
-				Duration: &metav1.Duration{Duration: lo.Must(time.ParseDuration("-20m"))},
+				Duration: &metav1.Duration{Duration: lo.Must(time.ParseDuration("20m"))},
 			}}
 			Expect(env.Client.Create(ctx, nodePool)).ToNot(Succeed())
 		})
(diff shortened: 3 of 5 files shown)


pkg/apis/v1beta1/nodepool.go
```diff
@@ -124,7 +124,7 @@ type Budget struct {
 	// This is required if Schedule is set.
 	// This regex has an optional 0s at the end since the duration.String() always adds
 	// a 0s at the end.
-	// +kubebuilder:validation:Pattern=`^([0-9]+(m|h)+(0s)?)$`
+	// +kubebuilder:validation:Pattern=`^(([0-9]+(h|m)(0s)?)|([0-9]+h+[0-9]+m+(0s)?))$`
```
njtran, line 127: I think there only needs to be one 0s at the end of the whole string, right before the $. It also seems like this means I could do 1h2h3m4m, which I don't think is what we want to do. 
```suggestion
	// +kubebuilder:validation:Pattern=`^(([0-9]+(h|m)(0s)?)|([0-9]+h+[0-9]+m+(0s)?))$`
```

engedaam: This should not allow for that kind of format. We are only allowing 1h/1m or 1h1m

pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -80,7 +80,7 @@ spec:
                               This is required if Schedule is set.
                               This regex has an optional 0s at the end since the duration.String() always adds
                               a 0s at the end.
-                            pattern: ^([0-9]+(m|h)+(0s)?)$
+                            pattern: ^((([0-9]+(h|m))|([0-9]+h[0-9]+m))(0s)?)$
```
jonathan-innis, line 83: Is there a simpler way to write this pattern? I could imagine that you could do something like `([1-9](h|m|s)+)|(0s)`

engedaam: The format you suggested would allow 1h2m1h yo happen

njtran: We also don't want to allow seconds, since this breaches the smallest denominator value of the schedule, which is minutes

njtran: I could understand disallowing a leading 0 on the front of the duration, but I don't think it's a big deal, since they should really be treated the exact same. Disallowing it means that users can't do something like 0h5m or 0h or 0m or 5h0m. These are all weird in that there's better ways to define it, but it's up to us on how we want to allow/disallow this.

pkg/apis/v1beta1/nodepool_validation_cel_test.go
```diff
@@ -177,6 +177,14 @@ var _ = Describe("CEL/Validation", func() {
 			}}
 			Expect(env.Client.Create(ctx, nodePool)).To(Succeed())
 		})
+		It("should succeed when creating a budget with hours and minuets in duration", func() {
```
njtran, line 180:
```suggestion
		It("should succeed when creating a budget with hours and minutes in duration", func() {
```

pkg/apis/v1beta1/nodepool_validation_webhook_test.go
```diff
@@ -131,6 +131,14 @@ var _ = Describe("Webhook/Validation", func() {
 			}}
 			Expect(nodePool.Validate(ctx)).To(Succeed())
 		})
+		It("should succeed when creating a budget with hours and minuets in duration", func() {
```
njtran, line 134:
```suggestion
		It("should succeed when creating a budget with hours and minutes in duration", func() {
```

pkg/apis/v1beta1/nodepool.go
```diff
@@ -124,7 +124,7 @@ type Budget struct {
 	// This is required if Schedule is set.
 	// This regex has an optional 0s at the end since the duration.String() always adds
 	// a 0s at the end.
-	// +kubebuilder:validation:Pattern=`^([0-9]+(m|h)+(0s)?)$`
+	// +kubebuilder:validation:Pattern=`^((([0-9]+(h|m))|([0-9]+h[0-9]+m))(0s)?)$`
```
Bryce-Soghigian, line 127: Devils advocate: can't we just lower the unit we are using? 1h10m is just 70m. Maybe it gets more complex with 12d10m vs 730m. 

I commonly see 
```
  consolidateAfter: 300s
  expireAfter: 720h
```

If we do move to a 1h10m pattern, we should probably do it everywhere to be consistent. [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1beta1/nodepool.go#L85) for example we follow a different pattern. It will be fairly confusing for users to see they can set 1h10m in one place but not everywhere

engedaam: This PR would allow both to happen. We currently promise both format to customer and this is addressing a bug

njtran: It seems in your link @Bryce-Soghigian that the only difference in these two regexes is the order. Seems compound durations are enabled there as well.

njtran (approved):
/lgtm
/approve

pkg/apis/crds/karpenter.sh_nodepools.yaml
jonathan-innis: Yep, makes sense. I just looked over it quick. Current regex makes sense 👍🏼

PR #1086, opened 2024-03-11
chore: Use strategic merge patching for Core APIs

jonathan-innis:
Fixes #N/A 

**Description**

Core APIs support [strategic merge patching](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/). This provides a more advanced way to patch together lists from standard merge patching, where lists can be merged if they have a `strategic-merge-patch-key` to denote them. For instance, this is true for the status conditions API.

[CRDs do not support strategic merge patching out of the box](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility:~:text=Yes-,strategic%2Dmerge%2Dpatch,-The%20new%20endpoints) so we cannot get rid of standard merge patching for them. This will resolve some errors in AWS E2E testing, where we are seeing errors like `patching node, Node \"ip-192-168-49-61.us-east-2.compute.internal\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"foregroundDeletion\"}"}`

**How was this change tested?**

`make presubmit`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1126, opened 2024-03-21
chore: Remove finalizer with `Update()` instead of `Patch()`

jonathan-innis:
Fixes #N/A 

**Description**

`Update()` operations to the finalizer allow us to avoid the errors that were mentioned in #1086 around strategic merge patching. This is the better alternative compared to `Patch()`, since it allows us to do optimistic locking and avoid spurious errors in Karpenter for adding finalizers due to list replacement on JSON merge patching. `removing termination finalizer, NodeClaim.karpenter.sh \"slothboulder-2-etpuq1of8e-bnllc\" is invalid: metadata.finalizers: Forbidden: no new finalizers can be added if the object is being deleted, found new finalizers []string{\"foregroundDeletion\"}"}`

`make presubmit`

### DIFF
--- a/pkg/controllers/nodeclaim/termination/controller.go
+++ b/pkg/controllers/nodeclaim/termination/controller.go
@@ -24,6 +24,7 @@ import (
 	"golang.org/x/time/rate"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/util/workqueue"
 	"knative.dev/pkg/logging"
 	controllerruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


pkg/controllers/nodeclaim/termination/controller.go
```diff
@@ -93,7 +95,10 @@ func (c *Controller) Finalize(ctx context.Context, nodeClaim *v1beta1.NodeClaim)
 	}
 	controllerutil.RemoveFinalizer(nodeClaim, v1beta1.TerminationFinalizer)
 	if !equality.Semantic.DeepEqual(stored, nodeClaim) {
-		if err = c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
+		if err = c.kubeClient.Update(ctx, nodeClaim); err != nil {
+			if errors.IsConflict(err) {
```
engedaam, line 99: Why are we requiring only on conflict errors?

jonathan-innis: We're requeueing on other errors too, see L102. This logic adds handling so a conflict error on an Update isn't treated as a true controller error

engedaam (commented):
/lgtm

PR #1138, opened 2024-03-26
chore: Changes to run e2e for private cluster

jigisha620:
Fixes #N/A 

**Description**
When running E2E tests on private cluster, we can't pull the pod image from the internet. But currently the default pod image is a constant. Changing it to variable so that the value can be configured in case the tests are running on private cluster.


njtran (commented):
/lgtm
/approve

PR #1147, opened 2024-03-31
feat: Detect Drift on NodeClaims on changes to NodeClass

### LINKED ISSUE #337: Karpenter NodeClaim Disruption `spec.nodeClassRef` Drift Detection

Currently, the node controller which is responsible for annotating nodes as expired, drifted, or empty, watches node, provisioner, and pod events, and enqueues reconcile requests for the linked nodes. 

https://github.com/aws/karpenter-core/blob/main/pkg/controllers/node/controller.go#L110-L140

Since karpenter-core isn't aware of provider-specific CRDs, it doesn't track watch events for these CRDs. 

We may be able to just watch `v1beta1.NodeClaimReference` objects.

This only affects the drift annotation. If someone changes their AWSNodeTemplate, they'll have to rely on the node heartbeat or another object event to get the node annotated. Since the heartbeat is 5 minutes, this is still happening at a good pace, but it could definitely be faster.

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

engedaam:
Fixes #337 

**Description**
- Detect drift on NodeClaims from when NodeClass get updated.

- `make presubmit`

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/controller.go
+++ b/pkg/controllers/nodeclaim/disruption/controller.go
@@ -23,6 +23,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/clock"
 	controllerruntime "sigs.k8s.io/controller-runtime"
@@ -48,7 +49,8 @@ type nodeClaimReconciler interface {
 // Controller is a disruption controller that adds StatusConditions to nodeclaims when they meet certain disruption conditions
 // e.g. When the NodeClaim has surpassed its owning provisioner's expirationTTL, then it is marked as "Expired" in the StatusConditions
 type Controller struct {
-	kubeClient client.Client
+	kubeClient    client.Client
+	cloudProvider cloudprovider.CloudProvider
 
 	drift      *Drift
 	expiration *Expiration
@@ -58,10 +60,11 @@ type Controller struct {
 // NewController constructs a nodeclaim disruption controller
 func NewController(clk clock.Clock, kubeClient client.Client, cluster *state.Cluster, cloudProvider cloudprovider.CloudProvider) operatorcontroller.Controller {
 	return operatorcontroller.Typed[*v1beta1.NodeClaim](kubeClient, &Controller{
-		kubeClient: kubeClient,
-		drift:      &Drift{cloudProvider: cloudProvider},
-		expiration: &Expiration{kubeClient: kubeClient, clock: clk},
-		emptiness:  &Emptiness{kubeClient: kubeClient, cluster: cluster, clock: clk},
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
+		drift:         &Drift{cloudProvider: cloudProvider},
+		expiration:    &Expiration{kubeClient: kubeClient, clock: clk},
+		emptiness:     &Emptiness{kubeClient: kubeClient, cluster: cluster, clock: clk},
 	})
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -125,6 +129,15 @@ func (c *Controller) Builder(_ context.Context, m manager.Manager) operatorcontr
...
-		),
-	)
+		)
+	for _, ncGVK := range c.supportedNodeClasses {
+		nodeclass := &unstructured.Unstructured{}
+		nodeclass.SetGroupVersionKind(ncGVK)
+		builder.Watches(
+			nodeclass,
+			nodeclaimutil.NodeClassEventHandler(c.kubeClient),
+		)
```
njtran, line 139: Don't you need to re-set the builder here? like 
builder := builder.Watches()

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -58,10 +61,11 @@ type Controller struct {
 // NewController constructs a nodeclaim disruption controller
 func NewController(clk clock.Clock, kubeClient client.Client, cluster *state.Cluster, cloudProvider cloudprovider.CloudProvider) operatorcontroller.Controller {
 	return operatorcontroller.Typed[*v1beta1.NodeClaim](kubeClient, &Controller{
-		kubeClient: kubeClient,
-		drift:      &Drift{cloudProvider: cloudProvider},
-		expiration: &Expiration{kubeClient: kubeClient, clock: clk},
-		emptiness:  &Emptiness{kubeClient: kubeClient, cluster: cluster, clock: clk},
+		kubeClient:           kubeClient,
+		supportedNodeClasses: cloudProvider.GetSupportedNodeClasses(),
```
njtran, line 65: WDYT about just passing in the cloud provider as a variable? The way we do it in other controllers is to pass in the cloud provider so that we can just call whatever member functions we need. It's also how we create the Controller anyway.

engedaam: Make sense, just added that

pkg/controllers/nodeclaim/disruption/controller.go
engedaam: `builder` is a pointer that is updated by `watches`, since `watches` is updates the pointer the returned values will be the same as the value we have at the top level 
- https://github.com/kubernetes-sigs/controller-runtime/blob/a17fd5870b906b50662fed6658265f5f8cf0863e/pkg/builder/controller.go#L137
- https://github.com/kubernetes-sigs/controller-runtime/blob/a17fd5870b906b50662fed6658265f5f8cf0863e/pkg/builder/controller.go#L179

njtran: Good practice to probably re-set the builder anyway.

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -89,6 +89,26 @@ func NodePoolEventHandler(c client.Client) handler.EventHandler {
 	})
 }
 
+// NodeClassEventHandler is a watcher on v1beta1.NodeClaim that maps NodeClass to NodeClaims based
+// on the nodeClassRef and enqueues reconcile.Requests for the NodeClaim
+func NodeClassEventHandler(c client.Client) handler.EventHandler {
```
jonathan-innis, line 94: How is this going to work when someone doesn't set the apiVersion or the kind today? We need some handling to ensure that we work with the default apiVersion/kind appropriately

jonathan-innis: Today, you will get an empty string back from some of the NodeClaims, and then you aren't going to requeue appropriately on it.

jonathan-innis, line 94: Do we think that there's a way to test any of this logic? May be kind of tough to do in functional testing but we should think about some E2E test case that we can do to ensure that we are requeueing immediately on these changes.

If possible we can think about a way to leverage the `handler.EnqueueRequestsFromMapFunc` to try and conjure up some kind of functional test. I'm not sure how tough that it would be to do so, though

engedaam: I was thinking that we could tell user to set the apiVersion and Kind to get better response for drift detection. Since it's not a bug to have an eventually consistent drift detection on NodeClass, I was thinking this would work really nicely once we default and require all fields in the NodeClassRef

jonathan-innis (approved):
/lgtm
/approve

pkg/utils/nodeclaim/nodeclaim.go
engedaam: For E2Es I was thinking we could have really short SLA on when we see the NodeClaims will get drifted on changes to both NodePool and NodeClass.

jonathan-innis (approved):
/lgtm
/approve

Issue #745, opened 2022-09-08
Global resource limits 


ozahavi:
**Tell us about your request**
The ability to configure resource limits (CPU\GPU) on a global level that will be enforced\applied to all provisioners in the cluster.

**Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?**
In certain use cases, I need to enable a mix of several CPU and GPU instance types that are configured in separate provisioners (because of different taints that are added to the nodes). Without a global limit for resources, I cannot limit the overall capacity of nodes created by Karpenter in the cluster.

**Are you currently working around this issue?**
No.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

stevehipwell:
@njtran have there been any more discussions about this? My understanding is that part of the Karpenter v1 process is to achieve parity with CA and this functionality would be required for that? This is also a significant feature in the context of multi-role cluster operations where there should be central controls in place to support constraining self-service capabilities.

jonathan-innis:
> My understanding is that part of the Karpenter v1 process is to achieve parity with CA 

I'm not sure that this is 100% correct here. Our goal here isn't necessarily to have 100% parity with CA but to provide a complete set of features and stable APIs that we believe are sufficient for all users to handle their node provisioning and lifecycle handling. Whether there is cross-over with CA or not is somewhat immaterial to whether we decide to scope a feature into GA for Karpenter v1.

> This is also a significant feature in the context of multi-role cluster operations where there should be central controls in place to support constraining self-service capabilities

As for multi-role cluster operations, I'm curious how you are thinking about how the self-service aspect of this would work, considering NodePools aren't namespaced at all today. If you have a user that has permission to update a NodePool, then you most likely also have permission to edit or delete other NodePools. Are you thinking that you would have some sort of policy engine in place that would allow you to determine whether someone does or doesn't have permission to do certain things on a certain set of NodePools? 

[...]

PR #1160, opened 2024-04-02
RFC: add NodeClaim limits proposal

jukie:
Fixes #N/A 
WIP PR: https://github.com/kubernetes-sigs/karpenter/pull/1151
**Description**
Adds RFC for discussing limiting number of provisioned NodeClaims

n/a


designs/nodeclaim-limits.md
```diff
@@ -0,0 +1,54 @@
...
+#### Pros/Cons
+
+1. 👍👍 Allows for simple cap on number on NodeClaims created by Karpenter, allowing users to apply this globally and per-NodePool
+2. 👎 May be used as a crutch vs having the cloudprovider providing the relevant resources. An example here could be supporting a limit for IP usage.
+
+### Solution 2: Suggest users request/implement reporting of resource types in CloudProvider
+
+In this solution, the community would need to update Cloud Providers to collect and report on resource usage of the desired types.
+
+#### Pros/Cons
```
Bryce-Soghigian, line 49: could you provide more details with this approach? I don't fully understand it.

jukie: I'll rework this a bit as there's an argument to push for both independently.
As a summary though, for certain use cases such as wanting to limit IP usage it may be possible to instead expose this information available under node resources by working with CNIs. I'm not sure how feasible this specific example is but the point I'm trying to make is that while limiting node count resolves the use case, it may not be addressing support for the actually desired resource to limit on (e.g. IPs).

Still, there are valid use cases for purely limiting node count such as your mention in the PR about Kubenet's limitation on scaling beyond 400 nodes. I think that Karpenter should support setting node count limits but users should be encouraged to open feature requests against the relevant projects they'd like to have more resource information from.

designs/nodeclaim-limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Use-Cases
+1. I'm using a shared network with limited IP pools and need to ensure Karpenter's usage doesn't consume too many IPs. Each Node consumes a minimum number of IPs but low-level resource usage for things like CPU/Memory differs between instance types.
+2. I have a special NodePool (for example a set of high performance nodes or AWS Reserved Instances) with a finite amount of compute available. I need a way to prevent Karpenter from launching too many nodes of this type. The only to do this is by calculating against an avialable resource like number of CPU's and adding this as a limit. This doesn't target the actual limiter which is number of instances rather than resource consumption.
+3. My CNI can't support more than 400 nodes but I'm not concerned about resource usage. I have no way to restrict Karpenter to this limit without severly limiting instance type options.
```
Bryce-Soghigian, line 14: A bit more on the history of this limitation. 

With kubebnet, only the nodes recieve an IP in the subnet. Pods can't communicate directly with each other and instead rely on UDR and IP forwarding handle connectivity between pods across nodesl. 


Azure supports a maximum of 400 routes in a UDR, so you can't have an AKS cluster larger than 400 nodes if you are using kubenet. 

Thats where this limitation comes from

designs/nodeclaim-limits.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+Currently, Karpenter users face a limitation in controlling the maximum number of nodes provisioned by Karpenter. This constraint hinders operations in environments where the CNI in use can't scale beyond a certain number of nodes or in scenarios involving shared networks with a finite IP address allocation. Some additional use cases and discussion on this limitation can be seen in the following issues: (#732, aws/karpenter-provider-aws#4462).
+
+## Solutions
+
+### [Recommended] Solution 1: New controller flag `max-node-claims` and per-NodePool limits in the `spec.limits` section of NodePool template
+
+1. A new controller startup flag `max-node-claims` which would allow for setting a global cap on the number of nodes provisioned via this Karpenter controller.
+2. Allow for setting per-NodePool limits on number of NodeClaims.
```
njtran, line 24: Why would we need both? It seems sufficient that users could set the nodepool level. Global limits are a different and probably are better encapsulated with a different RFC/feature request as shown here: https://github.com/kubernetes-sigs/karpenter/issues/745

Bryce-Soghigian: Nodepool level isn't sufficent as there are hard limits at the cluster level  in some cases, but I agree it is better encapsulated with the proper cluster level limits feature request. 

These limits should be part of that ratehr than its own flag.

jukie: Hi, yes it's desirable to have a global limit for users that have constraints at the cluster level.
Would it be possible to keep the scope of this proposal to cover a mechanism for limiting number of nodes?

designs/nodeclaim-limits.md
jukie: I've removed the second proposal but can add additional options if you'd like. I think no matter what there's still valid use cases for having a configurable limit on number of nodes so don't want to distract discussions.

designs/nodeclaim-limits.md
Bryce-Soghigian: I would say the real design choice here would be deciding if we should add a flag for "max-node-claims", or should we have global limits encapsulated in the design shared by njtran. 

Otherwise, later we might end up adding the node limits to this global mechanism and then later have to change it.

It is better to get the configuration right from the start.

jukie: Agreed on getting it right the first time. Is there a design that's been discussed for global limits or do you just mean the feature request in https://github.com/kubernetes-sigs/karpenter/issues/745?

njtran:
In my head, this RFC should tackle one of the following two: 
1. Global limits
2. Number of nodes/nodeclaims in limits

I think it's possible to do both in one RFC, but I think the review would go much quicker as two separate RFCs. Which one are we tackling here?

sftim (commented):
/ok-to-test

It's just Markdown

designs/nodeclaim-limits.md
```diff
@@ -0,0 +1,44 @@
...
+3. My CNI can't support more than 400 nodes but I'm not concerned about underlying resource usage. I have no way to restrict Karpenter to this limit without severly limiting instance type options.
+
+Currently, Karpenter users face a limitation in controlling the maximum number of nodes provisioned by Karpenter. This constraint hinders operations in environments where the CNI in use can't scale beyond a certain number of nodes or in scenarios involving shared networks with a finite IP address allocation. Some additional use cases and discussion on this limitation can be seen in the following issues: (#732, aws/karpenter-provider-aws#4462).
+
+## Proposal
+
+### New controller flag `max-node-claims` and per-NodePool limits in the `spec.limits` section of NodePool template
```
sftim, line 22: I'd really prefer to be able to reconfigure this without a controller restart, maybe using a ConfigMap?

What I'm imagining: you specify a NodePool selector, and you specify a cap on NodeClaims owned by the selected NodePool(s). Karpenter doesn't make a NodeClaim if doing so would go over the cap.

njtran:
@jukie Any update here on the above? https://github.com/kubernetes-sigs/karpenter/pull/1160#issuecomment-2070842425

designs/nodeclaim-limits.md
ellistarn: We avoid global configuration in Karpenter as a general rule. Is a policy enforcement engine + NodePool limits a better approach?

PR #1161, opened 2024-04-02
feat: In the event of a cost tie during consolidation, prefer to consolidate if the node count decreases.

bguo-figma:
Fixes https://github.com/aws/karpenter-provider-aws/issues/5944#issuecomment-2033227060

**Description**
In the event of a cost-tie, we should consider the consolidation downward if the instance count decreases. While this is "the same" from a cost perspective, it can be more efficient (less space wasted on Daemonsets, less Datadog monitoring costs, etc). 

Minor change

### DIFF
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -180,7 +180,7 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
 	// it based on availability and price which could result in selection/launch of non-lowest priced instance in the list. So, we would keep repeating this loop till we get to lowest priced instance
 	// causing churns and landing onto lower available spot instance ultimately resulting in higher interruptions.
 	results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions, incompatibleMinReqKey, _ =
-		filterByPriceWithMinValues(results.NewNodeClaims[0].InstanceTypeOptions, results.NewNodeClaims[0].Requirements, candidatePrice)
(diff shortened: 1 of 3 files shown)


bguo-figma:
> I can imagine there being users who don't want this, due to the cost of disrupting pods.

I hear what you're saying, but I don't believe this is true in a properly configured system. The change I'm proposing is still monotonically "cheaper" (ie there's no waffling/persistent disruption/constant churn), so it might "increase" disruption, but its scoped/limited/temporary (ie tiebreak only). 

Properly configured, this should have no production impact to users, since consolidation respects PDBs. If you care about limiting pod disruption, use a PDB and this change as-is is probably sufficient, if you don't then you don't care about this change anyways.

> If a cluster has 256 large instances, we could similarly pack them into 4 32xlarge instances, of which, all the pods on those instances would be disrupted.

This is true, IF all those pods are configured to be disruptible via PDB right? Either those pods can tolerate the disruption (throttled by PDB size, so they don't all experience disruption at once), or they can't. As as soon as the "tiebreak" is resolved there should be no ongoing disruption/consolidation/we should be at steady state.

devinsmith911:
Is this something that is still being considered?
We have a case where we would prefer a lower node count in the event of the cost being the same regardless of extra disruptions.

bguo-figma:
IMO it's still desirable behavior, but I'm unfortunately busy with other work to really polish this PR/feature flag it properly. I may take another stab at it later if someone who works more closely with Karpenter is unable to

PR #1164, opened 2024-04-03
chore: Add retryable error to cloud provider

### LINKED ISSUE #947: Wait for underlying instances to be terminated before removing node finalizers in termination
### Description

The termination controller calls `cloudProvider.Delete(nodeClaim)` which deletes the underlying instance. It will remove the finalizer once the cloudProvider.Delete() call succeeds. [code](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/controller.go#L100-L102)

Within AWS, Karpenter currently succeeds and proceeds to remove the finalizer once the ec2.TerminateInstance() call has succeeded, which can result in nodes being deregistered before the instance reaches a `Terminated` state (as opposed to the `Terminating` it first reaches when the API call succeeds). 

(@tallaxes please correct me if i'm wrong) Within Azure, Karpenter models its return value of the `cloudProvider.Delete()` call with a retryable error, where Karpenter will continue to call `cloudProvider.Delete()` which will eventually succeed when the instance is terminated. 

One way this could generically be implemented is by creating a cloudProvider.RetryableError() which will allow Karpenter to parse a state of an instance that has been successfully called terminate, but hasn't been fully terminated yet, re-enqueueing the node into the termination controller reconciliation loop.

Important, as currently any pods that tolerate the Karpenter NoSchedule taint are leaked, and are never given a chance to clean up. In addition, there may be other processes that are interfered with if the node is deregistered preemptively.

jigisha620:
Fixes #947

**Description**
The termination controller calls `cloudProvider.Delete(nodeClaim)` which tries to terminate the underlying instance. It will remove the finalizer once the cloudProvider.Delete() call succeeds. However, the current implementation in AWS does not wait until the instance state is actually `terminated` and proceeds to remove the finalizer once the `ec2.TerminateInstance()` call has succeeded. 

The change in this PR adds a retryable error to cloudprovider which will allow Karpenter to parse instance state that has been successfully called terminate, but wasn't fully terminated. This effectively re-enqueues the node into the reconciliation loop of the termination controller.

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -198,8 +198,8 @@ func IsNodeClaimNotFoundError(err error) bool {
 	if err == nil {
 		return false
 	}
-	var mnfErr *NodeClaimNotFoundError
-	return errors.As(err, &mnfErr)
+	var ncnfErr *NodeClaimNotFoundError
+	return errors.As(err, &ncnfErr)
 }
 
 func IgnoreNodeClaimNotFoundError(err error) error {
@@ -267,4 +267,27 @@ func IgnoreNodeClassNotReadyError(err error) error {
 		return nil
 	}
 	return err
+}
+
+// RetryableError is an error type returned by CloudProviders when the action emitting the error has to be retried
+type RetryableError struct {
+	error
+}
+
+func NewRetryableError(err error) *RetryableError {
+	return &RetryableError{
+		error: err,
+	}
+}
+
+func (e *RetryableError) Error() string {
+	return fmt.Sprintf("retryable error, %s", e.error)
+}
+
+func IsRetryableError(err error) bool {
+	if err == nil {
+		return false
+	}
+	var retryableError *RetryableError
+	return errors.As(err, &retryableError)
 }
--- a/pkg/controllers/nodeclaim/termination/controller.go
+++ b/pkg/controllers/nodeclaim/termination/controller.go
@@ -89,7 +89,7 @@ func (c *Controller) Finalize(ctx context.Context, nodeClaim *v1beta1.NodeClaim)
 		return reconcile.Result{}, nil
 	}
 	if nodeClaim.Status.ProviderID != "" {
-		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
+		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil || cloudprovider.IsRetryableError(err) {
 			return reconcile.Result{}, fmt.Errorf("terminating cloudprovider instance, %w", err)
 		}
 	}


pkg/controllers/nodeclaim/termination/controller.go
```diff
@@ -89,7 +89,7 @@ func (c *Controller) Finalize(ctx context.Context, nodeClaim *v1beta1.NodeClaim)
 		return reconcile.Result{}, nil
 	}
 	if nodeClaim.Status.ProviderID != "" {
-		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
+		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil || cloudprovider.IsRetryableError(err) {
```
jonathan-innis, line 92: What if we just re-leverage the existing NodeClaimNotFound error that we already use? We could just return a NotFound error once the instance is terminated like we do already today with our Get() and List() operations.

pkg/cloudprovider/types.go
```diff
@@ -268,3 +268,26 @@ func IgnoreNodeClassNotReadyError(err error) error {
 	}
 	return err
 }
+
+// RetryableError is an error type returned by CloudProviders when the action emitting the error has to be retried
+type RetryableError struct {
```
jonathan-innis, line 273: Not sure that we have a need for using a `RetryableError` if we can just re-leverage our NodeClaimNotFound error

pkg/controllers/nodeclaim/termination/controller.go
jigisha620: @jonathan-innis If the instance is terminated then we don't want to return an error. It's only when it is not terminated that we want to return some form of error that will re-enqueue the reconciliation. And can't use  NodeClaimNotFound error in that case because we ignore that error here if I understand correctly.

jonathan-innis: > And can't use NodeClaimNotFound error in that case because we ignore that error here if I understand correctly

You could keep trying to Delete() until you got the NodeClaimNotFound error that would indicate that the instance has been correctly terminated. I discussed with @njtran offline. I think it's fine to add this additional error type but I was generally trying to avoid us adding additional error types into our cloudprovider unless they had a specific reason for being there. Also, the other thing to consider here is whether returning an error on a Delete() call that succeeded so that the client can know that is should keep retrying makes a lot of sense. 

These are my thoughts, but I'm fine to take in what you have here and we can see how all the error types pan out over time.

pkg/cloudprovider/types.go
```diff
@@ -268,3 +268,26 @@ func IgnoreNodeClassNotReadyError(err error) error {
...
+}
+
+func NewRetryableError(err error) *RetryableError {
+	return &RetryableError{
+		error: err,
+	}
+}
+
+func (e *RetryableError) Error() string {
+	return fmt.Sprintf("retry error, %s", e.error)
```
njtran, line 284:
```suggestion
	return fmt.Sprintf("retryable error, %s", e.error)
```

jonathan-innis (approved):
/lgtm
/approve

Bryce-Soghigian:
We are also interested in a Retriable cloud provider error. 

We were discussing should we drive retries for creating vms through the azure sdk or through the controller? There are some cases where we see `429` throttling errors, and being able to set a retry after on the nodeclaim would be very interesting

njtran (commented):
/lgtm
/approve

pkg/test/environment.go
```diff
@@ -123,6 +123,13 @@ func NewEnvironment(scheme *runtime.Scheme, options ...functional.Option[Environ
 	} else {
 		c = lo.Must(client.New(environment.Config, client.Options{Scheme: scheme}))
 	}
+
+	namespace := corev1.Namespace{}
```
jonathan-innis, line 127: Does this work all of the time? Can we try doing this with a continual retry for up to say 10s with a 1s back-off? Consider `retry.Do()`? The other thing: I think that we should add a comment here with a `// TODO` to remove this once we upgrade to the latest version of controller-runtime

jonathan-innis (approved):
/lgtm
/approve

raychinov:
👋 Hey! The initially approved fix seems to be reverted, are there plans for the issue to be addressed in a different way?

PR #1169, opened 2024-04-04
chore: adding log for when cluster state has been out of sync for 15 seconds

### LINKED ISSUE #1163: Cluster State Logs take up 2/3s of the total log volume
### Description

There is a large amount of the log volume dedicated to a single log `cluster sync` logs. I see two ways to solve the volume. 

1. As we move to the new controller runtime logging with verbosity levels, I suggest we add it at lower verbosity.
2. Log after a longer period similar to the `abnormal amount of time between consolidation` like seen [here](https://github.com/Azure/karpenter-provider-azure/issues/191). We can log if there has been an abnormal amount of time for cluster sync. 

Bryce-Soghigian:
Fixes #1163 

**Description**
Adding a LastSyncTime to cluster state, along with a log to tell us when cluster state has been out of sync for 15 or more seconds.

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -115,7 +115,6 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
 			logging.FromContext(ctx).Infof("detected abnormal duration since cluster was last synced %v", c.cluster.LastSyncTime())
 		}
-		logging.FromContext(ctx).Debugf("waiting on cluster sync")
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -112,6 +112,9 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
+		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration { 
+			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime()) 
+		}
 		logging.FromContext(ctx).Debugf("waiting on cluster sync")
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -113,7 +113,7 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
 		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
-			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime())
+			logging.FromContext(ctx).Infof("detected abnormal duration since cluster was last synced %v", c.cluster.LastSyncTime())
 		}
 		logging.FromContext(ctx).Debugf("waiting on cluster sync")
 		return reconcile.Result{RequeueAfter: time.Second}, nil
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -112,8 +112,8 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
-		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration { 
-			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime()) 
+		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
+			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime())
 		}
 		logging.FromContext(ctx).Debugf("waiting on cluster sync")
 		return reconcile.Result{RequeueAfter: time.Second}, nil
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -112,9 +112,6 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
-		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
-			logging.FromContext(ctx).Infof("detected abnormal duration since cluster was last synced %v", c.cluster.LastSyncTime())
-		}
 		return reconcile.Result{RequeueAfter: time.Second}, nil
 	}
 
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -70,6 +70,7 @@ func WithReason(reason string) func(LaunchOptions) LaunchOptions {
 		return o
 	}
 }
+
 // Provisioner waits for enqueued pods, batches them, creates capacity and binds the pods to the capacity.
 type Provisioner struct {
 	cloudProvider  cloudprovider.CloudProvider
@@ -80,7 +81,6 @@ type Provisioner struct {
 	recorder       events.Recorder
 	cm             *pretty.ChangeMonitor
 }
-
 
 func NewProvisioner(kubeClient client.Client, recorder events.Recorder,
 	cloudProvider cloudprovider.CloudProvider, cluster *state.Cluster,
@@ -118,8 +118,8 @@ func (p *Provisioner) Reconcile(ctx context.Context, _ reconcile.Request) (resul
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !p.cluster.Synced(ctx) {
-		if time.Since(p.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration { 
-			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", p.cluster.LastSyncTime()) 
+		if time.Since(p.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
+			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", p.cluster.LastSyncTime())
 		}
 		logging.FromContext(ctx).Debugf("waiting on cluster sync")
 		return reconcile.Result{RequeueAfter: time.Second}, nil
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -118,9 +118,6 @@ func (p *Provisioner) Reconcile(ctx context.Context, _ reconcile.Request) (resul
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !p.cluster.Synced(ctx) {
-		if time.Since(p.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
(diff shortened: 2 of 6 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -114,6 +114,9 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
+		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration { 
+			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime())
```
Bryce-Soghigian, line 118: Alternatively we could log how many seconds it has been out of sync. Or we can do some modulo math to log this only every 10 seconds. The behavior here is we will log out of sync since <time> over and over until its synced

pkg/controllers/disruption/controller.go
```diff
@@ -114,6 +114,9 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
+		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration { 
+			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime()) 
+		}
 		logging.FromContext(ctx).Debugf("waiting on cluster sync")
```
Bryce-Soghigian, line 120: Should we just remove this now, or do we want to keep it and put a much lower verbosity when the controller runtime logging change goes in?

pkg/controllers/state/cluster.go
```diff
@@ -90,6 +97,9 @@ func (c *Cluster) Synced(ctx context.Context) (synced bool) {
 	// Set the metric to whatever the result of the Synced() call is
 	defer func() {
 		clusterStateSynced.Set(lo.Ternary[float64](synced, 1, 0))
+		if synced { 
+			c.lastSyncTime = c.clock.Now()
+		}
```
Bryce-Soghigian, line 102: Need to add some tests here

pkg/controllers/state/cluster.go
```diff
@@ -197,7 +212,7 @@ func (c *Cluster) NominateNodeForPod(ctx context.Context, providerID string) {
 	}
 }
 
-// TODO remove this when v1alpha5 APIs are deprecated. With v1beta1 APIs Karpenter relies on the existence
+// TODO remove this when v1alpha5 APIs are eprecated. With v1beta1 APIs Karpenter relies on the existence
```
Bryce-Soghigian, line 215:
```suggestion
// TODO remove this when v1alpha5 APIs are deprecated. With v1beta1 APIs Karpenter relies on the existence
```

pkg/controllers/disruption/controller.go
njtran: We should just be able to remove it.

pkg/controllers/disruption/controller.go
```diff
@@ -114,6 +114,9 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
+		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
+			logging.FromContext(ctx).Infof("cluster state has not been in sync since %v, waiting for sync", c.cluster.LastSyncTime())
+		}
```
njtran, line 119: This is interesting, since you'd have no logs for 15 seconds, but then spam it continuously after.

Additionally, I'd name the log to match the variable if you're naming it like this
```suggestion
		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
			logging.FromContext(ctx).Infof("detected abnormal duration since cluster was last synced %v", c.cluster.LastSyncTime())
		}
```

pkg/controllers/provisioning/provisioner.go
njtran, line 1: same comments here as in disruption.go

Bryce-Soghigian: fixed

pkg/controllers/disruption/controller.go
Bryce-Soghigian: removed

pkg/controllers/disruption/controller.go
```diff
@@ -114,7 +114,9 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !c.cluster.Synced(ctx) {
-		logging.FromContext(ctx).Debugf("waiting on cluster sync")
+		if time.Since(c.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
```
jonathan-innis, line 117: Look at `pretty.ChangeMonitor`. It offers you a chance to leverage a cache that validates if a key is changed and you can use that as a signal to know whether you should log or not here.

jonathan-innis: That should prevent you spamming the log line a bunch of times after you first log it once

pkg/controllers/provisioning/provisioner.go
```diff
@@ -118,7 +118,9 @@ func (p *Provisioner) Reconcile(ctx context.Context, _ reconcile.Request) (resul
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !p.cluster.Synced(ctx) {
-		logging.FromContext(ctx).Debugf("waiting on cluster sync")
+		if time.Since(p.cluster.LastSyncTime()) > state.AbnormalClusterSyncDuration {
```
jonathan-innis, line 121: I wonder if it's possible to encapsulate this behavior so that we don't actually have to handle logging it everywhere. What if it handled like the client-go "waiting 1.5s due to client-side throttling" error message. If we keep calling cluster synced and we stay false for a long enough time-frame, the method just logs that it hasn't synced itself

Bryce-Soghigian: > If we keep calling cluster synced and we stay false for a long enough time-frame, the method just logs that it hasn't synced itself

Yeah I was thinking about this too. Just in cluster state we log in the Synced() call if it hasn't synced. This way we also don't have to expose "lastSyncedTime" as well.

pkg/controllers/state/cluster.go
```diff
@@ -67,7 +71,16 @@ type Cluster struct {
 	antiAffinityPods sync.Map // pod namespaced name -> *v1.Pod of pods that have required anti affinities
 }
 
+const (
+	// AbnormalClusterSyncDuration is the duration after which the cluster state is considered to be out of sync for a long enough time to alarm the customer
+	AbnormalClusterSyncDuration = 15 * time.Second
+	// AbnormalClusterSyncLoggingInterval is the interval at which we will log about the cluster being out of sync if it is out of sync
+	AbnormalClusterSyncLoggingInterval = 5 * time.Minute
```
Bryce-Soghigian, line 78: Is five minute expiry on the cache entries too long? Should I log every minute instead?

pkg/controllers/state/cluster.go
```diff
@@ -90,6 +103,14 @@ func (c *Cluster) Synced(ctx context.Context) (synced bool) {
 	// Set the metric to whatever the result of the Synced() call is
 	defer func() {
 		clusterStateSynced.Set(lo.Ternary[float64](synced, 1, 0))
+		if synced {
+			c.lastSynced = c.clock.Now()
+		}
+		// If we have been out of sync for a while, and we haven't logged about it recently, log it
+		if time.Since(c.lastSynced) > AbnormalClusterSyncDuration && c.cm.HasChanged("ClusterSynced", synced) {
```
Bryce-Soghigian, line 110: Do we add unit tests for testing logging changes in the project?

pkg/controllers/state/cluster.go
```diff
@@ -90,6 +103,14 @@ func (c *Cluster) Synced(ctx context.Context) (synced bool) {
...
+			logging.FromContext(ctx).Warnf("cluster state is out of sync since %v", time.Since(c.lastSynced))
```
Bryce-Soghigian, line 111: https://github.com/kubernetes/client-go/blob/aa7909e7d7c0661792ba21b9e882f3cd6ad0ce53/rest/request.go#L625 is how they log this in client-go for clientside throttling

pkg/controllers/provisioning/provisioner.go
Bryce-Soghigian: Moved it into the Synced() method and am using a change monitor lmk what you think.

Bryce-Soghigian: https://github.com/kubernetes/client-go/blob/aa7909e7d7c0661792ba21b9e882f3cd6ad0ce53/rest/request.go#L625 is how they log this in client-go for clientside throttling. 

@jonathan-innis  did you mean just copy the error message style? "waiting for "x seconds" dur to cluster state being out of sync?

pkg/controllers/state/cluster.go
```diff
@@ -67,11 +71,21 @@ type Cluster struct {
 	antiAffinityPods sync.Map // pod namespaced name -> *v1.Pod of pods that have required anti affinities
 }
 
+const (
+	// AbnormalClusterSyncDuration is the duration after which the cluster state is considered to be out of sync for a long enough time to alarm the customer
```
jonathan-innis, line 75:
```suggestion
	// AbnormalClusterSyncDuration is the duration after which the cluster state is considered to be out of sync for a long enough time to log
```

pkg/controllers/state/cluster.go
jonathan-innis: No, feels a little overkill to add it as well. I think just testing that you get the behavior that you are expecting here should be sufficient

pkg/controllers/state/cluster.go
jonathan-innis: Makes sense. We really don't have a lot of warning logs in the codebase; I almost think that we shouldn't have them at all. What do you think about firing this as an error log and making sure that the sync duration is sufficiently long that a user would be convinced that this is actually an issue? 

Inevitably, the trouble with warning logs is that users end up treating them as Info (ignoring them because they don't actually mean anything) or they end up treating them as Error (because for them, the warning is always indicative of a problem). Warnings are basically an easy-out for a project to say "we don't know whether this can be a problem for you or not". Ideally, we can ship something that's always indicative of a problem

pkg/controllers/state/cluster.go
jonathan-innis: Depending on the interval that we pick above, I think that's going to determine the interval here. If we are saying that 1m of NotSynced is a problem (I'm not sure that this is the right value, but this is probably what I would pick off the top of my head), then logging every 1m or so after it happens is probably sufficient

Bryce-Soghigian:
Did some testing, I am seeing some weirdness. For some reason we get a log saying its been out of sync for 292 years. Will dig into it

~/dev/karp/karpenter-provider-azure (bsoghigian/v0.35*) » k logs -n karpenter karpenter-849b677444-6pn49 | grep "sync"                                                                                                                                                                                                                                      
{"level":"WARN","time":"2024-04-07T00:07:41.871Z","logger":"controller.disruption","message":"cluster state is out of sync since 2562047h47m16.854775807s","commit":"f395e0f-dirty"}
{"level":"WARN","time":"2024-04-07T00:16:51.998Z","logger":"controller.nodepool.counter","message":"cluster state is out of sync since 15.167988727s","commit":"f395e0f-dirty","nodepool":"general-purpose"}

pkg/controllers/state/cluster.go
Bryce-Soghigian: I am thinking 1m instead as well.

pkg/controllers/state/cluster.go
Bryce-Soghigian: yeah that makes sense, changed the severity for now

pkg/controllers/state/cluster.go
```diff
@@ -90,6 +104,14 @@ func (c *Cluster) Synced(ctx context.Context) (synced bool) {
 	// Set the metric to whatever the result of the Synced() call is
 	defer func() {
 		clusterStateSynced.Set(lo.Ternary[float64](synced, 1, 0))
+		if synced {
+			c.lastSynced = time.Now()
+		}
+		// If we have been out of sync for a while, and we haven't logged about it recently, log it
+		if time.Since(c.lastSynced) > AbnormalClusterSyncDuration && c.cm.HasChanged("ClusterSynced", synced) {
```
jonathan-innis, line 111: I'm still surprised that this works with the existing code. I'd imagine that you'd need to set `c.lastSynced` to a value right when we startup; else, you are going to get it set to the zero value if it doesn't sync and it's going to result in the log that you have been seeing where we are reporting that we haven't synced for a _very_ long time :D

jonathan-innis, line 111: musing: I wonder if we could encapsulate this behavior to reduce the complexity of our Synced() cluster function.

```go
// ErrorStateMonitor monitors how long something maintains a "target" value. If that target value is maintained for long enough (`errorThreshold`), it fires an error to represent that the component hasn't recovered from the error state and will fire this error at every `interval`.
type ErrorStateMonitor struct {
  target interface{}
  errorThreshold time.Duration
  interval time.Duration
}
```

Bryce-Soghigian: Yep that makes sense the first synced log would be from 292 years ago. Modified it to set last synced to time.Now() on startup.

PR #1172, opened 2024-04-05
chore: Re-enqueue nodeclaim termination after 10s in case of retryable error

jigisha620:
Fixes #N/A 

**Description**
1. Re-enqueue nodeClaim termination after 10s when we see an error. 
2. Added test for retryable error.

### DIFF
--- a/pkg/controllers/nodeclaim/termination/controller.go
+++ b/pkg/controllers/nodeclaim/termination/controller.go
@@ -89,7 +89,13 @@ func (c *Controller) Finalize(ctx context.Context, nodeClaim *v1beta1.NodeClaim)
 		return reconcile.Result{}, nil
 	}
 	if nodeClaim.Status.ProviderID != "" {
-		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil || cloudprovider.IsRetryableError(err) {
+		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
+			// We expect cloudProvider to emit a Retryable Error when the underlying instance is not terminated and if that
+			// happens, we want to re-enqueue reconciliation until we terminate the underlying instance before removing
+			// finalizer from the nodeClaim.
+			if cloudprovider.IsRetryableError(err) {
+				return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
+			}
 			return reconcile.Result{}, fmt.Errorf("terminating cloudprovider instance, %w", err)
 		}
 	}
--- a/pkg/controllers/nodeclaim/termination/suite_test.go
+++ b/pkg/controllers/nodeclaim/termination/suite_test.go
@@ -18,6 +18,7 @@ package termination_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


pkg/controllers/nodeclaim/termination/suite_test.go
```diff
@@ -216,4 +217,30 @@ var _ = Describe("Termination", func() {
...
+		ExpectApplied(ctx, env.Client, node)
+
+		// Expect the node and the nodeClaim to both be gone
+		Expect(env.Client.Delete(ctx, nodeClaim)).To(Succeed())
+		ExpectReconcileSucceeded(ctx, nodeClaimTerminationController, client.ObjectKeyFromObject(nodeClaim)) // triggers the node deletion
+		ExpectFinalizersRemoved(ctx, env.Client, node)
+		ExpectNotFound(ctx, env.Client, node)
+
+		cloudProvider.NextDeleteErr = cloudprovider.NewRetryableError(fmt.Errorf("underlying instance not terminated"))
+		ExpectReconcileFailed(ctx, nodeClaimTerminationController, client.ObjectKeyFromObject(nodeClaim)) // re-enqueue reconciliation due to retryable error
```
jonathan-innis, line 238: This is good test; however, we should think about what our back-off strategy is going to be here and ensure that we are returning the correct result here with this change.

pkg/controllers/nodeclaim/termination/controller.go
```diff
@@ -90,7 +90,7 @@ func (c *Controller) Finalize(ctx context.Context, nodeClaim *v1beta1.NodeClaim)
 	}
 	if nodeClaim.Status.ProviderID != "" {
 		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil || cloudprovider.IsRetryableError(err) {
-			return reconcile.Result{}, fmt.Errorf("terminating cloudprovider instance, %w", err)
+			return reconcile.Result{RequeueAfter: 10 * time.Second}, fmt.Errorf("terminating cloudprovider instance, %w", err)
```
jonathan-innis, line 93: Can we special case this for the retryable error? For a transient error, I think that standard back-off is fine, but for this special type of error that we have, I'm thinking that's the only time that we should do this 10s back-off

pkg/controllers/nodeclaim/termination/suite_test.go
```diff
@@ -216,4 +217,31 @@ var _ = Describe("Termination", func() {
...
+		// Expect the node and the nodeClaim to both be gone
+		Expect(env.Client.Delete(ctx, nodeClaim)).To(Succeed())
+		ExpectReconcileSucceeded(ctx, nodeClaimTerminationController, client.ObjectKeyFromObject(nodeClaim)) // triggers the node deletion
+		ExpectFinalizersRemoved(ctx, env.Client, node)
+		ExpectNotFound(ctx, env.Client, node)
+
+		cloudProvider.NextDeleteErr = cloudprovider.NewRetryableError(fmt.Errorf("underlying instance not terminated"))
+		err = cloudProvider.Delete(ctx, nodeClaim)
+		Expect(cloudprovider.IsRetryableError(err)).To(BeTrue())
+		ExpectReconcileSucceeded(ctx, nodeClaimTerminationController, client.ObjectKeyFromObject(nodeClaim))
```
jonathan-innis, line 240: I think if the reconcile succeeds, we get back the result. Can we check that result here to see that we are doing the back-off that we expect? e.g. 10s?

pkg/controllers/nodeclaim/termination/controller.go
```diff
@@ -89,7 +89,11 @@ func (c *Controller) Finalize(ctx context.Context, nodeClaim *v1beta1.NodeClaim)
 		return reconcile.Result{}, nil
 	}
 	if nodeClaim.Status.ProviderID != "" {
-		if err = c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil || cloudprovider.IsRetryableError(err) {
+		err = c.cloudProvider.Delete(ctx, nodeClaim)
```
jonathan-innis, line 92: style: consider the following

```go
if err := c.cloudProvider.Delete(ctx, nodeClaim); cloudprovider.IgnoreNodeClaimNotFound(err) != nil {
  if cloudprovider.IsRetryableError(err) {
    return reconcile.Result{RequeueAfter: 10*time.Second}, nil
  }
  return reconcile.Result{}, fmt.Errorf("terminating cloudprovider instance, %v", err)
}
```

pkg/controllers/nodeclaim/termination/suite_test.go
```diff
@@ -216,4 +217,32 @@ var _ = Describe("Termination", func() {
...
+		ExpectApplied(ctx, env.Client, node)
+
+		// Expect the node and the nodeClaim to both be gone
+		Expect(env.Client.Delete(ctx, nodeClaim)).To(Succeed())
+		ExpectReconcileSucceeded(ctx, nodeClaimTerminationController, client.ObjectKeyFromObject(nodeClaim)) // triggers the node deletion
+		ExpectFinalizersRemoved(ctx, env.Client, node)
+		ExpectNotFound(ctx, env.Client, node)
+
+		cloudProvider.NextDeleteErr = cloudprovider.NewRetryableError(fmt.Errorf("underlying instance not terminated"))
+		result := ExpectReconcileSucceeded(ctx, nodeClaimTerminationController, client.ObjectKeyFromObject(nodeClaim))
```
jonathan-innis, line 238: Nice 🎉

pkg/controllers/nodeclaim/termination/controller.go
jonathan-innis: Consider also adding a comment on why we are retrying here with this kind of error, since it isn't immediately apparent to the reader what a "retryable error" is

jonathan-innis (approved):
LGTM 🚀

Issue #651, opened 2023-10-31
Taint nodes with a NoSchedule for Consolidation before Validation begins


njtran:
### Description

**What problem are you trying to solve?**
Karpenter adds a `karpenter.sh/disruption:NoSchedule=disrupting` taint for Consolidation actions after a 15s validation period. There is a narrow interval of time where pods blocking eviction can schedule to nodes when Karpenter taints the nodes, resulting in these pods getting eventually evicted during termination. This was validated to be a race of `< 300ms` [here](https://github.com/aws/karpenter/issues/4310#issuecomment-1758750554). 

While https://github.com/aws/karpenter-core/issues/624 includes PreferNoSchedule as an option here, it may make more sense to only use NoSchedule for Consolidation here to completely remove this race condition.  

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

wmgroot:
I'm working on implementing a patch to deal with this issue at my company. We're running into this constantly when using Karpenter for a cluster with almost nothing but ephemeral CI jobs, which seems like a recipe for constant disruption evaluation because the pod profile is highly volatile.

Can I ask that we prioritize an air-tight fix for the issue instead of another "reduce the chance of it occurring" option that is proposed with the PreferNoSchedule implementation? Right now the do-not-disrupt annotation does not do what it says on tin.

njtran:
@wmgroot Can you share how often you're hitting this? From my understanding the race is super small window, so I'm curious exactly how often this is hitting. Are you seeing a do-not-disrupt pod added/scheduled to a node, only for that node to be soon consolidated after?

wmgroot:
We've seen almost 2300 occurrences in the last week with Gitlab CI job pods. Every pod has the do-not-disrupt annotation. This cluster has about 25 NodeClaims which are primarily used for extremely volatile CI jobs.
![Screenshot from 2024-02-14 11-45-20](https://github.com/kubernetes-sigs/karpenter/assets/3124965/533e354b-d26c-49d2-82b1-2f5e556c9614)

[...]

PR #1180, opened 2024-04-10
fix: address disruption taint race condition

### LINKED ISSUE #1167: Pod gets evicted still even with "karpenter.sh/do-not-disrupt" annotation
### Description

**Observed Behavior**:
Pod with "karpenter.sh/do-not-disrupt" annotation still get evicted - not sure how to reproduce this cause this seems to be intermittent

Some examples of Karpenter evicting pods despite claiming the same pods to be undisruptable just moments ago.

```
Apr 3, 2024 @ 16:10:41.000: Cannot disrupt Node: Pod "mesh-west/runner-zjioycxf-project-14388-concurrent-0lvg45" has "karpenter.sh/do-not-disrupt" annotation
Apr 3, 2024 @ 16:10:41.000: Cannot disrupt NodeClaim: Pod "mesh-west/runner-zjioycxf-project-14388-concurrent-0lvg45" has "karpenter.sh/do-not-disrupt" annotation
...
(Events from Datadog) Apr 3, 4:10:41 pm: Events from the Pod mesh-west/runner-zjioycxf-project-14388-concurrent-0lvg45 - Evicted pod - Events emitted by the karpenter
```

```
Apr 2, 2024 @ 13:09:04.000: Cannot disrupt Node: Pod "mesh-west/runner-su7-c7kj-project-14388-concurrent-0qc8j7" has "karpenter.sh/do-not-disrupt" annotation
Apr 2, 2024 @ 13:09:04.000: Cannot disrupt NodeClaim: Pod "mesh-west/runner-su7-c7kj-project-14388-concurrent-0qc8j7" has "karpenter.sh/do-not-disrupt" annotation
...
(Events from Datadog) Apr 2, 1:09:05 pm: Events from the Pod mesh-west/runner-su7-c7kj-project-14388-concurrent-0qc8j7 - Evicted pod - Events emitted by the karpenter
```

**Expected Behavior**:
Pod with "karpenter.sh/do-not-disrupt" annotation shouldn't be evicted

**Versions**:
- Chart Version: 0.35.2
- Kubernetes Version (`kubectl version`): EKS v1.23.17-eks-508b6b3

jmdeal:
Fixes #1167 

**Description**
Implements a short-term, zero-API mitigation for the consolidation race condition called out in issues #651 and #1167. This PR will not close out issue #651 since we still may change the exact mechanism as part of the holistic taint redesign for #624. 

Manual validation + karpenter-provider-aws E2E suite

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -3788,6 +3788,8 @@ var _ = Describe("Consolidation", func() {
 			Eventually(finished.Load, 10*time.Second).Should(BeTrue())
 			wg.Wait()
 
+			ExpectReconcileSucceeded(ctx, queue, types.NamespacedName{})
+
 			// nothing should be removed since the node is no longer empty
 			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(1))
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
--- a/pkg/controllers/disruption/emptynodeconsolidation.go
+++ b/pkg/controllers/disruption/emptynodeconsolidation.go
@@ -19,8 +19,8 @@ package disruption
 import (
 	"context"
 	"errors"
-	"fmt"
 
+	"github.com/samber/lo"
 	"knative.dev/pkg/logging"
 
 	"sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling"
@@ -88,31 +88,20 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
 		return Command{}, scheduling.Results{}, errors.New("interrupted")
 	case <-c.clock.After(consolidationTTL):
 	}
-	validationCandidates, err := GetCandidates(ctx, c.cluster, c.kubeClient, c.recorder, c.clock, c.cloudProvider, c.ShouldDisrupt, c.queue)
+
+	v := NewValidation(c.clock, c.cluster, c.kubeClient, c.provisioner, c.cloudProvider, c.recorder, c.queue)
+	validatedCandidates, err := v.ValidateCandidates(ctx, cmd.candidates...)
 	if err != nil {
-		logging.FromContext(ctx).Errorf("computing validation candidates %s", err)
 		return Command{}, scheduling.Results{}, err
 	}
-	// Get the current representation of the proposed candidates from before the validation timeout
-	// We do this so that we can re-validate that the candidates that were computed before we made the decision are the same
-	candidatesToDelete := mapCandidates(cmd.candidates, validationCandidates)
-
-	postValidationMapping, err := BuildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient, c.recorder)
-	if err != nil {
-		return Command{}, scheduling.Results{}, fmt.Errorf("building disruption budgets, %w", err)
+	// In addition to the normal candidate validation checks, also check that the nodes are still empty
+	if len(validatedCandidates) == 0 || lo.ContainsBy(validatedCandidates, func(c *Candidate) bool {
+		return len(c.reschedulablePods) != 0
+	}) {
+		logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd)
+		return Command{}, scheduling.Results{}, nil
 	}
 
-	// The deletion of empty NodeClaims is easy to validate, we just ensure that:
-	// 1. All the candidatesToDelete are still empty
-	// 2. The node isn't a target of a recent scheduling simulation
-	// 3. the number of candidates for a given nodepool can no longer be disrupted as it would violate the budget
-	for _, n := range candidatesToDelete {
-		if len(n.reschedulablePods) != 0 || c.cluster.IsNodeNominated(n.ProviderID()) || postValidationMapping[n.nodePool.Name] == 0 {
-			logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd)
-			return Command{}, scheduling.Results{}, nil
-		}
-		postValidationMapping[n.nodePool.Name]--
-	}
 	return cmd, scheduling.Results{}, nil
 }
 
--- a/pkg/controllers/disruption/emptynodeconsolidation.go
+++ b/pkg/controllers/disruption/emptynodeconsolidation.go
@@ -92,10 +92,15 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
 	v := NewValidation(c.clock, c.cluster, c.kubeClient, c.provisioner, c.cloudProvider, c.recorder, c.queue)
 	validatedCandidates, err := v.ValidateCandidates(ctx, cmd.candidates...)
 	if err != nil {
+		if IsValidationError(err) {
+			logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd)
+			return Command{}, scheduling.Results{}, nil
+		}
 		return Command{}, scheduling.Results{}, err
 	}
-	// In addition to the normal candidate validation checks, also check that the nodes are still empty
-	if len(validatedCandidates) == 0 || lo.ContainsBy(validatedCandidates, func(c *Candidate) bool {
+
+	// TODO (jmdeal@): better encapsulate within validation
+	if lo.ContainsBy(validatedCandidates, func(c *Candidate) bool {
 		return len(c.reschedulablePods) != 0
 	}) {
 		logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd)
--- a/pkg/controllers/disruption/validation.go
+++ b/pkg/controllers/disruption/validation.go
@@ -113,7 +113,7 @@ func (v *Validation) IsValid(ctx context.Context, cmd Command, validationPeriod 
 //
 //	a. It must pass the global candidate filtering logic (no blocking PDBs, no do-not-disrupt annotation, etc)
 //	b. It must not have any pods nominated for it
-//	c. It must still be disruptible without violating node disruption budgets
+//	c. It must still be disruptable without violating node disruption budgets
 //
 // If these conditions are met for all candidates, ValidateCandidates returns a slice with the updated representations.
 func (v *Validation) ValidateCandidates(ctx context.Context, candidates ...*Candidate) ([]*Candidate, error) {
--- a/pkg/controllers/disruption/validation.go
+++ b/pkg/controllers/disruption/validation.go
@@ -39,40 +39,37 @@ import (
 // of the commands passed to IsValid were constructed based off of the same consolidation state.  This allows it to
 // skip the validation TTL for all but the first command.
 type Validation struct {
-	validationPeriod time.Duration
-	start            time.Time
-	clock            clock.Clock
-	cluster          *state.Cluster
-	kubeClient       client.Client
-	cloudProvider    cloudprovider.CloudProvider
-	provisioner      *provisioning.Provisioner
-	once             sync.Once
-	recorder         events.Recorder
-	queue            *orchestration.Queue
+	start         time.Time
+	clock         clock.Clock
+	cluster       *state.Cluster
+	kubeClient    client.Client
+	cloudProvider cloudprovider.CloudProvider
+	provisioner   *provisioning.Provisioner
+	once          sync.Once
+	recorder      events.Recorder
+	queue         *orchestration.Queue
 }
 
-func NewValidation(validationPeriod time.Duration, clk clock.Clock, cluster *state.Cluster, kubeClient client.Client, provisioner *provisioning.Provisioner,
+func NewValidation(clk clock.Clock, cluster *state.Cluster, kubeClient client.Client, provisioner *provisioning.Provisioner,
 	cp cloudprovider.CloudProvider, recorder events.Recorder, queue *orchestration.Queue) *Validation {
 	return &Validation{
-		validationPeriod: validationPeriod,
-		clock:            clk,
-		cluster:          cluster,
-		kubeClient:       kubeClient,
-		provisioner:      provisioner,
-		cloudProvider:    cp,
-		recorder:         recorder,
-		queue:            queue,
+		clock:         clk,
+		cluster:       cluster,
+		kubeClient:    kubeClient,
+		provisioner:   provisioner,
+		cloudProvider: cp,
+		recorder:      recorder,
+		queue:         queue,
 	}
 }
 
-//nolint:gocyclo
-func (v *Validation) IsValid(ctx context.Context, cmd Command) (bool, error) {
+func (v *Validation) IsValid(ctx context.Context, cmd Command, validationPeriod time.Duration) (bool, error) {
 	var err error
 	v.once.Do(func() {
 		v.start = v.clock.Now()
 	})
 
-	waitDuration := v.validationPeriod - v.clock.Since(v.start)
+	waitDuration := validationPeriod - v.clock.Since(v.start)
 	if waitDuration > 0 {
 		select {
 		case <-ctx.Done():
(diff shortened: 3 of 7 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -58,8 +58,11 @@ type Controller struct {
 	lastRun       map[string]time.Time
 }
 
-// pollingPeriod that we inspect cluster to look for opportunities to disrupt
-const pollingPeriod = 10 * time.Second
+const (
+	// pollingPeriod that we inspect cluster to look for opportunities to disrupt
+	pollingPeriod         = 10 * time.Second
+	commandExecutionDelay = 10 * time.Second
```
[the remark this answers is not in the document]
jmdeal: I made it longer when testing and missed changing it back, this should be 5 seconds. This is based on cluster autoscaler's default for their equivalent field, `node-delete-delay-after-taint`. This was sufficient in my testing (with 5s) and I haven't been able to find any instances of users needing to change this delay.

pkg/controllers/disruption/controller.go
```diff
@@ -191,11 +196,45 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+			state.RequireNoScheduleTaint(ctx, c.kubeClient, false, stateNodes...),
+		)
+	}
+
+	// After adding the NoSchedule taint to the proposed candidates, revalidate the candidates after a waiting period. This addresses a race condition
+	// where pods with the `karpenter.sh/do-not-disrupt` annotation can be scheduled after the command has been validated but before the nodes have been tainted.
+	// A waiting period is used to ensure Karpenter's internal cluster state has synced and all pods are accounted for.
+	select {
+	case <-ctx.Done():
+		return multierr.Append(fmt.Errorf("interrupted (command-id: %s)", commandID), state.RequireNoScheduleTaint(ctx, c.kubeClient, false, stateNodes...))
```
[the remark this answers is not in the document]
jmdeal: Good point, I threw an error here based on what we were already do for consolidation: https://github.com/kubernetes-sigs/karpenter/blob/43da3604c14097ab1ee3ca360af5a82446a4db4d/pkg/controllers/disruption/emptynodeconsolidation.go#L86-L90
but you're right that there's no value in attempting to untaint. I could go either way with surfacing this as an error at all, but chose to follow what we were already doing.

pkg/controllers/disruption/controller.go
```diff
@@ -191,11 +196,45 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+	case <-c.clock.After(commandExecutionDelay):
+	}
+
+	currentCandidates, err := GetCandidates(ctx, c.cluster, c.kubeClient, c.recorder, c.clock, c.cloudProvider, candidateFilter, c.queue)
```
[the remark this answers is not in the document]
jmdeal: I think I've arrived at a better approach. I've realized the validation behavior ideally should be slightly different for each disruption method so it seems reasonable to extend the method interface. This leaves only the common logic in the disruption controller (untainting / canceling the command). Will probably split this and tainting into a separate function from executeCommand, more for readability than anything else.

pkg/controllers/disruption/controller.go
```diff
@@ -194,10 +203,42 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
 		return multierr.Append(fmt.Errorf("tainting nodes (command-id: %s), %w", commandID, err), state.RequireNoScheduleTaint(ctx, c.kubeClient, false, stateNodes...))
 	}
 
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("interrupted (command-id: %s)", commandID)
```
njtran, line 208:
```suggestion
		return fmt.Errorf("interrupted command (command-id: %s)", commandID)
```

pkg/controllers/disruption/controller.go
```diff
@@ -58,8 +58,13 @@ type Controller struct {
...
 }
 
-// pollingPeriod that we inspect cluster to look for opportunities to disrupt
-const pollingPeriod = 10 * time.Second
+const (
+	// pollingPeriod that we inspect cluster to look for opportunities to disrupt
+	pollingPeriod = 10 * time.Second
+
+	// commandExecutionDelay is the duration spent waiting after tainting candidate nodes before validaing the command
+	commandValidationDelay = 5 * time.Second
```
njtran, line 66: Is this a dart throw? Or are there metrics you're using to drive this decision

pkg/controllers/disruption/controller.go
```diff
@@ -194,10 +203,42 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+	if err != nil {
+		return fmt.Errorf("validating command, determining candidates, %w", err)
+	}
+	disruptionBudgetMapping, err := BuildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient, c.recorder)
+	if err != nil {
+		return fmt.Errorf("validating command, building disruption budgets, %w", err)
+	}
+	validatedCommand := m.ValidateCommand(ctx, cmd, disruptionBudgetMapping, candidates...)
+	if len(validatedCommand.candidates) == 0 {
+		logging.FromContext(ctx).Infof("abandoning disruption attempt, 0 out of %d nodes are still valid candidates", len(cmd.candidates))
```
njtran, line 224: You could alternatively phrase it like this
```suggestion
		logging.FromContext(ctx).Infof("abandoning disruption attempt, %d candidates are no longer valid", len(cmd.candidates))
```

pkg/controllers/disruption/controller.go
```diff
@@ -194,10 +203,42 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+	// This addresses a race conditions where pods can schedule to the candidate nodes after the command has been finalized, but before the candidates were tainted.
+	candidates, err := GetCandidates(ctx, c.cluster, c.kubeClient, c.recorder, c.clock, c.cloudProvider, m.ShouldDisrupt, c.queue)
+	if err != nil {
+		return fmt.Errorf("validating command, determining candidates, %w", err)
+	}
+	disruptionBudgetMapping, err := BuildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient, c.recorder)
+	if err != nil {
+		return fmt.Errorf("validating command, building disruption budgets, %w", err)
+	}
+	validatedCommand := m.ValidateCommand(ctx, cmd, disruptionBudgetMapping, candidates...)
```
njtran, line 222: From looking at the various `ValidateCommand()` calls, it looks like they all implement the following 

1. Map L214 candidates to command.Candidates
2. do `method.ShouldDisrupt`
3. check disruption budgets

I personally don't think we need another interface method for this, as the method specific behavior is already encapsulated in the GetCandidates() call on L214. I could see doing a helper function for the disruption budget mapping check. WDYT?

pkg/controllers/disruption/controller.go
```diff
@@ -194,10 +203,42 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+		invalidatedCandidates := lo.FilterMap(cmd.candidates, func(candidate *Candidate, _ int) (*state.StateNode, bool) {
+			return candidate.StateNode, !lo.ContainsBy(validatedCommand.candidates, func(c *Candidate) bool {
+				return c.Name() == candidate.Name()
+			})
+		})
+		if err = state.RequireNoScheduleTaint(ctx, c.kubeClient, false, invalidatedCandidates...); err != nil {
+			return fmt.Errorf("untainting invalidated candidates, %w", err)
+		}
+		logging.FromContext(ctx).Infof("updated disruption command after validation, disrupting via %s %s", m.Type(), validatedCommand)
+	}
```
njtran, line 237: Should we be continuing if a subset of the candidates are no longer candidates? Let's say we had 100 candidates, and now only 1 candidate is invalid. We need to either re-do a scheduling simulation to know what the replacement should be, or just abandon the command entirely. 

For reference, we abandon the command for consolidation.

pkg/controllers/disruption/controller.go
```diff
@@ -194,10 +203,42 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+	case <-ctx.Done():
+		return fmt.Errorf("interrupted (command-id: %s)", commandID)
+	case <-c.clock.After(commandValidationDelay):
+	}
+
+	// Revalidate the proposed command after the node has been tainted.
+	// This addresses a race conditions where pods can schedule to the candidate nodes after the command has been finalized, but before the candidates were tainted.
+	candidates, err := GetCandidates(ctx, c.cluster, c.kubeClient, c.recorder, c.clock, c.cloudProvider, m.ShouldDisrupt, c.queue)
+	if err != nil {
+		return fmt.Errorf("validating command, determining candidates, %w", err)
```
tzneal, line 216: This leaves nodes tainted, right?

pkg/controllers/disruption/controller.go
```diff
@@ -194,10 +203,42 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+	}
+	disruptionBudgetMapping, err := BuildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient, c.recorder)
+	if err != nil {
+		return fmt.Errorf("validating command, building disruption budgets, %w", err)
```
tzneal, line 220: Same comment

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -3750,7 +3750,7 @@ var _ = Describe("Consolidation", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodes[0])
 		})
-		It("should not consolidate if the action becomes invalid during the node TTL wait", func() {
+		FIt("should not consolidate if the action becomes invalid during the node TTL wait", func() {
```
njtran, line 3753:
```suggestion
		It("should not consolidate if the action becomes invalid during the node TTL wait", func() {
```

pkg/controllers/disruption/emptynodeconsolidation.go
```diff
@@ -96,6 +96,10 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
 	// Get the current representation of the proposed candidates from before the validation timeout
 	// We do this so that we can re-validate that the candidates that were computed before we made the decision are the same
 	candidatesToDelete := mapCandidates(cmd.candidates, validationCandidates)
+	if invalidatedCount := len(cmd.candidates) - len(candidatesToDelete); invalidatedCount != 0 {
```
njtran, line 99: Isn't this the same as len(candidates) != len(candidatesToDelete)

pkg/controllers/disruption/emptynodeconsolidation.go
```diff
@@ -96,6 +96,10 @@ func (c *EmptyNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
...
+		logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt, %d candidates are no longer valid, command is no longer valid, %s", invalidatedCount, cmd)
```
njtran, line 100: Can we add these formatted directives into structured values in the logger?
```suggestion
		logging.FromContext(ctx).Debugf("abandoning empty node consolidation attempt, %d candidates are no longer valid, command is no longer valid, %s", invalidatedCount, cmd)
```

pkg/controllers/disruption/validation.go
```diff
@@ -80,37 +79,54 @@ func (v *Validation) IsValid(ctx context.Context, cmd Command) (bool, error) {
 		case <-v.clock.After(waitDuration):
 		}
 	}
-	// Get the current representation of the proposed candidates from before the validation timeout
-	// We do this so that we can re-validate that the candidates that were computed before we made the decision are the same
-	// We perform filtering here to ensure that none of the proposed candidates have blocking PDBs or do-not-evict/do-not-disrupt pods scheduled to them
-	validationCandidates, err := GetCandidates(ctx, v.cluster, v.kubeClient, v.recorder, v.clock, v.cloudProvider, v.ShouldDisrupt, v.queue)
+	validatedCandidates, err := v.validateCandidates(ctx, cmd.candidates...)
+	if len(validatedCandidates) == 0 || err != nil {
```
njtran, line 83: Why would we want to continue if the number of candidates isn't 0, but isn't the set of candidates we started off with?

jmdeal: `ValidateCandidates` returns an empty slice if the full set of candidates were not valid. If the candidates were all valid, the returned slice is the updated Candidate representations.

pkg/controllers/disruption/controller.go
jmdeal: Resolving, updated approach does not involve taint and wait strategy. This number was a slight dart through that was justified by the default value for the same value in CAS.

pkg/controllers/disruption/controller.go
jmdeal: Resolving, current approach removes the interface function. Realistically we only see this race occur for consolidation where the race window is expanded due to the validating scheduling simulation so I've added a second validation pass after that check and am going to opt to leave it at that for now.

pkg/controllers/disruption/controller.go
jmdeal: Like you called out the decision we make here is going to differ depending on the disruption type. It was handled in the ValidateCommand function for that reason. If we want to cancel the entire command it should return a command with no candidates. Otherwise it should return a command with a reduced number of candidates and we would untaint the invalidated candidates here.

pkg/controllers/disruption/emptynodeconsolidation.go
jmdeal: Resolving since this was removed, but yes. I changed it temporarily to move that math out of the log statement but was going to change it back if I kept this log line.

pkg/controllers/disruption/controller.go
jmdeal: Yes, that was a miss. I've swapped out this PR with an alternate implementation that doesn't involve tainting during validation but I'll fix this in my backup branch in case we need to go back.

pkg/controllers/disruption/validation.go
jonathan-innis: I think @njtran's point is that a failed validation should be an error here. Reading this function, I would assume I either get: 1) No error and the mapped candidates or 2) No mapped candidates and an error. The in-between state is a little confusing. I wonder if a special "FailedValidation" error could handle this better. At a high-level, I wonder if we could replace `isValid` with a `FailedValidation` error generally.

pkg/utils/pod/scheduling.go
```diff
@@ -53,10 +53,12 @@ func IsReschedulable(pod *v1.Pod) bool {
 // - Is an active pod (isn't terminal or actively terminating)
 // - Doesn't tolerate the "karpenter.sh/disruption=disrupting" taint
 // - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
+// - Doesn't have the "karpenter.sh/do-not-disrupt=true" annotation
 func IsEvictable(pod *v1.Pod) bool {
 	return IsActive(pod) &&
 		!ToleratesDisruptionNoScheduleTaint(pod) &&
-		!IsOwnedByNode(pod)
+		!IsOwnedByNode(pod) &&
+		IsDisruptable(pod)
```
jonathan-innis, line 61: We're going to stop evicting `karpenter.sh/do-not-disrupt` pods in this change? That's a major breaking change and not something that I think we can do without the grace period mechanism that we've been talking about. We've told users that they can expect that there is a way to bail out of pods hanging nodes during drain by doing a `kubectl delete node`. I agree that we need to reverse this, but shouldn't do so without giving these same users a mechanism to gracefully delete their nodes with `kubectl delete node`

jonathan-innis: This also strikes me as out of scope of this race condition fix -- something we want to do long-term, but probably not in the context of this bug fix

pkg/controllers/disruption/validation.go
jonathan-innis: The other thing that I don't love but was trying to figure out a good solve for and couldn't think of anything: a `Validate` function returns back retrieved data. Realistically, you want to retrieve the data, map the data, then validate the data. I get this is a little difficult since we've coupled some of our validation in the GetCandidates() logic. More of a musing and to see if you can think of something better that has closer semantic meaning than just `ValidateCandidates`.

jmdeal: We could definitely replace this with a `ValidationError`, basically just needed a mechanism to differentiate between a 'fatal' error that would be propogated back up the call stack and an error reported vs a non-fatal error where we just discard the consolidation attempt.

pkg/utils/pod/scheduling.go
jmdeal: Sounds good, I included it here as a mechanism to ensure that these pods aren't disrupted even if they slip through but I agree it can be decoupled.

jonathan-innis (approved):
/lgtm
/approve

pkg/controllers/disruption/validation.go
```diff
@@ -35,82 +35,114 @@ import (
...
+	}
+	return nil
+}
+
+// ValidateCandidates gets the current representation of the provided candidates and ensures that they are all still valid.
+// For a candidate to still be valid, the following conditions must be met:
+//
+//	a. It must pass the global candidate filtering logic (no blocking PDBs, no do-not-disrupt annotation, etc)
+//	b. It must not have any pods nominated for it
+//	c. It must still be disruptible without violating node disruption budgets
```
njtran, line 116: spelling nit
```suggestion
//	c. It must still be disruptable without violating node disruption budgets
```

njtran (commented):
/lgtm
/approve

PR #1183, opened 2024-04-12
chore: bump registration ttl log

Bryce-Soghigian:
Fixes #N/A 

**Description**
We have a system that allow us to raise alerts based on logs. Liveness probe is a useful signal for greater node registration problems on the cluster.

We log at the Info level verbosity. Registration TTL should be more than just debug. I can figure out the nodeclaim is being deleted at 15 minutes but not why. Having this explicit log would be useful for alerting on logs. 
- it wasn't


jonathan-innis (approved):
/lgtm
/approve

PR #1187, opened 2024-04-16
fix: update metric when there are zero disruption candidates

jmdeal:
Fixes #N/A 

**Description**
Moves the eligible node metric update to the top-level disruption controller from the individual consolidation implementations. This both reduces code duplication and, more importantly, fixes a bug where the karpenter_disruption_eligible_nodes metric is not updated if the number of candidates is zero. This results in the last non-zero number of candidates being reported indefinitely.

Tested in a personal cluster via the Karpenter AWS provider


pkg/controllers/disruption/controller.go
```diff
@@ -152,6 +152,11 @@ func (c *Controller) disrupt(ctx context.Context, disruption Method) (bool, erro
 	if err != nil {
 		return false, fmt.Errorf("determining candidates, %w", err)
 	}
+	EligibleNodesGauge.With(map[string]string{
```
jonathan-innis, line 155: Do we have testing in place for this metric to ensure we don't regress on this issue in the future?

jmdeal: Not at the moment, but I'll go ahead and add some 👍

pkg/controllers/disruption/emptiness_test.go
```diff
@@ -105,6 +105,40 @@ var _ = Describe("Emptiness", func() {
 			Expect(recorder.Calls("Unconsolidatable")).To(Equal(2))
 		})
 	})
+	Context("Metrics", func() {
+		const eligibleNodesMetric = "karpenter_disruption_eligible_nodes"
```
njtran, line 109: same here

pkg/controllers/disruption/expiration_test.go
```diff
@@ -76,6 +76,47 @@ var _ = Describe("Expiration", func() {
 		})
 		nodeClaim.StatusConditions().MarkTrue(v1beta1.Expired)
 	})
+	Context("Metrics", func() {
+		const eligibleNodesMetric = "karpenter_disruption_eligible_nodes"
```
njtran, line 80: same here

pkg/controllers/disruption/expiration_test.go
```diff
@@ -76,6 +76,47 @@ var _ = Describe("Expiration", func() {
...
+
+			fakeClock.Step(10 * time.Minute)
+			wg := sync.WaitGroup{}
+			ExpectTriggerVerifyAction(&wg)
+			ExpectReconcileSucceeded(ctx, disruptionController, types.NamespacedName{})
+			wg.Wait()
+
+			ExpectMetricGaugeValue(eligibleNodesMetric, 0, eligibleNodesLabels)
+
+			// remove the do-not-disturb annotation to make the node eligible for drift and update cluster state
```
njtran, line 107:
```suggestion
			// remove the do-not-disrupt annotation to make the node eligible for drift and update cluster state
```

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -134,6 +134,54 @@ var _ = Describe("Consolidation", func() {
 			Expect(recorder.Calls("Unconsolidatable")).To(Equal(6))
 		})
 	})
+	Context("Metrics", func() {
+		const eligibleNodesMetric = "karpenter_disruption_eligible_nodes"
```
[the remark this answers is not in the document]
jmdeal: Totally, don't know why I didn't 🤦. AFAIK there shouldn't be any cyclic dep considering this is the test suite.

pkg/controllers/disruption/drift_test.go
```diff
@@ -78,6 +78,47 @@ var _ = Describe("Drift", func() {
 		})
 		nodeClaim.StatusConditions().MarkTrue(v1beta1.Drifted)
 	})
+	Context("Metrics", func() {
+		const eligibleNodesMetric = "karpenter_disruption_eligible_nodes"
+		var eligibleNodesLabels = map[string]string{
+			"method":             "drift",
+			"consolidation_type": "",
```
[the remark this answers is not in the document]
jmdeal: IIUC not specifying it would result in us returning the first metric that has the method set to `"drift"`. That's fine in this case since there is only one but I wanted to be explicit. This also tests that we shouldn't set the consolidation_type for this metric.

pkg/controllers/disruption/consolidation_test.go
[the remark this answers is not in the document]
jmdeal: Got around to rechecking, thought there was but there aren't any constants here. Could go either way on making some or just defining one here. Leaning towards just keeping it as is, this is what we do elsewhere in the project.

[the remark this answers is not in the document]
njtran: WDYT about making this more programmatic so that if we update the metric name in the future the test automatically updates? Also makes helps to not mistyping metrics, and get compiler checks. Maybe we could make a helper that gets the namespace, subsystem, and name of the metric and joins it into the metric name?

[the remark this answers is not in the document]
jmdeal: So I was able to figure it out but it is...interesting to say the least. Doesn't seem like there's anyway to just get the qualified name, only the full description which you need to extract the name from. Also, need to use an async API to get this description from all collector types (looking at gauge vecs). Seems to work fine now that it's implemented though 🤷 .

pkg/test/expectations/expectations.go
```diff
@@ -461,14 +463,33 @@ func ExpectOwnerReferenceExists(obj, owner client.Object) metav1.OwnerReference
 	return or
 }
 
+func ExpectFullyQualifiedNameFromCollector(collector prometheus.Collector) string {
```
njtran, line 466: So this also has the added benefit of making sure that if a metric isn't found in the environment, we won't silently fail anymore, right?

pkg/controllers/disruption/expiration_test.go
njtran: You un-did the change!

pkg/controllers/disruption/expiration_test.go
```diff
@@ -76,6 +77,50 @@ var _ = Describe("Expiration", func() {
...
+	Context("Metrics", func() {
+		var eligibleNodesMetric string
+		var eligibleNodesLabels = map[string]string{
+			"method":             "expiration",
+			"consolidation_type": "",
+		}
+		BeforeEach(func() {
+			eligibleNodesMetric = ExpectFullyQualifiedNameFromCollector(disruption.EligibleNodesGauge)
+		})
+		It("should correctly report eligible nodes", func() {
```
njtran, line 89: I understand making this its own separate test is nice to test individual things, but I'd rather not add more tests (when we're already having to increase the timeouts). Can you just add this eligible metric check into the existing disruption tests? If you can do this for each of the disruption tests, it'll make sure that the metric is working properly in all the different ways we're testing the codepaths too.

pkg/test/expectations/expectations.go
jonathan-innis, line 466:
```suggestion
func ExpectMetricName(c prometheus.Collector) string {
```

Someone who is calling this probably doesn't care about the detail at that is the fully qualified name or is coming from a collector. I'd just simplify the name down to "This gets the metric name" and accepts a collector in as input. 

Can we also put some commenting above this function for what it's doing and why it's doing it? I spent a bit of time reading through this function to understand what we were doing and took a bit of time reading through the prometheus code to understand this limitation. Also, is it worth to go to the prometheus client_go folks and open an issue asking them if there's a way to make this less hard

pkg/controllers/state/suite_test.go
```diff
@@ -1117,6 +1117,12 @@ var _ = Describe("Pod Anti-Affinity", func() {
 })
 
 var _ = Describe("Cluster State Sync", func() {
+	var clusterStateSyncedMetric string
+	var clusterStateNodesCountMetric string
+	BeforeEach(func() {
+		clusterStateSyncedMetric = ExpectFullyQualifiedNameFromCollector(state.ClusterStateSynced)
```
jonathan-innis, line 1123: Just curious if this is possible: Rather than storing the name and then passing it through to `ExpectMetricGaugeValue`, is there a way to directly pass the gauge into the function so all of this is encapsulated and we never have to store the metric name?

jonathan-innis:
Nice work tracking this down! This is a great simplifying change and some solid testing added!

pkg/test/expectations/expectations.go
jmdeal: I can definitely add some additional comments, agreed that this function is not super straightforward. I actually originally named this `ExpectMetricName`, I changed it because the fully qualified name != the individual metric name although it does for our use cases. The complexity here is largely because I wanted a common function that would across individual metrics as well as vecs which required using a higher level interface.

jmdeal: I wasn't considering that but you're right, it should.

pkg/controllers/disruption/expiration_test.go
jmdeal: I had a pretty large set of updates for the disruption suite in general in my original consolidation race condition fix PR that I'm going to incorporate into a new PR. The main change was a rework of how we handle faking the clock which significantly sped up the test suite (~5x speed improvement IIRC). I think with that coming as well we can justify a standalone test, but I could also see incorporating this elsewhere.

pkg/controllers/state/suite_test.go
jmdeal: We totally could, I honestly don't remember why I opted to go this route at this point. I'm happy to make the change.

pkg/controllers/disruption/expiration_test.go
jmdeal: I'll redo it...did a rebase and force push but had forgot to pull first.

pkg/controllers/disruption/expiration_test.go
njtran: Synced offline, opting for followups to solve this problem

njtran (commented):
/lgtm
/approve

PR #1188, opened 2024-04-16
cleanup: clean redundant condition checking

jwcesign:
Fixes #N/A 

**Description**
```
func (e *Emptiness) ComputeCommand(_ context.Context, disruptionBudgetMapping map[string]int, candidates ...*Candidate) (Command, scheduling.Results, error) {
	// First check how many nodes are empty so that we can emit a metric on how many nodes are eligible
	emptyCandidates := lo.Filter(candidates, func(cn *Candidate, _ int) bool {
                // After this filtering, all the candidates have zero reschedulablePods
		return cn.NodeClaim.DeletionTimestamp.IsZero() && len(cn.reschedulablePods) == 0
	})

	EligibleNodesGauge.With(map[string]string{
		methodLabel:            e.Type(),
		consolidationTypeLabel: e.ConsolidationType(),
	}).Set(float64(len(candidates)))

	empty := make([]*Candidate, 0, len(emptyCandidates))
	for _, candidate := range emptyCandidates {
                // Here is redundant 
                if len(candidate.reschedulablePods) > 0 {
			continue
		}
		if disruptionBudgetMapping[candidate.nodePool.Name] > 0 {
			empty = append(empty, candidate)
			disruptionBudgetMapping[candidate.nodePool.Name]--
		}
	}
	return Command{
		candidates: empty,
	}, scheduling.Results{}, nil
}

```

Code review

### DIFF
--- a/pkg/controllers/disruption/emptiness.go
+++ b/pkg/controllers/disruption/emptiness.go
@@ -75,15 +75,13 @@ func (e *Emptiness) ComputeCommand(_ context.Context, disruptionBudgetMapping ma
 
 	empty := make([]*Candidate, 0, len(emptyCandidates))
 	for _, candidate := range emptyCandidates {
-		if len(candidate.reschedulablePods) > 0 {
-			continue
-		}
 		// If there's disruptions allowed for the candidate's nodepool,
 		// add it to the list of candidates, and decrement the budget.
-		if disruptionBudgetMapping[candidate.nodePool.Name] > 0 {
-			empty = append(empty, candidate)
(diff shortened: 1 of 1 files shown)


jonathan-innis:
Hey @jwcesign, I think @jmdeal already has a change out in #1187 that proposes this change. Given the scope of that PR and that it was opened first, we'll be taking that change in over this one. We'd still love for you to contribute. Happy to point you to some good-first-issues if you're interested in picking something up.

jonathan-innis:
Closing in favor of #1187

Issue #887, opened 2023-12-19
Pod Metric Testing is Flakey due to the Lack of the `default` namespace


jonathan-innis:
### Description

**Observed Behavior**:

Karpenter testing is flaking often with failures in the pod metric testing like

```bash
[FAILED] Expected success, but got an error:
      <*errors.StatusError | 0xc000138aa0>: 
      namespaces "default" not found
      {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "namespaces \"default\" not found",
              Reason: "NotFound",
              Details: {Name: "default", Group: "", Kind: "namespaces", UID: "", Causes: nil, RetryAfterSeconds: 0},
              Code: 404,
          },
      }
  In [It] at: /home/runner/work/karpenter/karpenter/pkg/controllers/metrics/pod/suite_test.go:58 @ 12/18/23 17:16:47.[69](https://github.com/kubernetes-sigs/karpenter/actions/runs/7251475446/job/19753908191?pr=885#step:5:70)6
```

An example job where this occurred can be seen here: https://github.com/kubernetes-sigs/karpenter/actions/runs/7251475446/job/19753908191?pr=885

This is being tracked upstream to see what their recommendation is e.g. whether we should do a downstream check or whether this is something that can be fixed upstream: https://github.com/kubernetes-sigs/controller-runtime/issues/2626

**Expected Behavior**:

Tests shouldn't be flakey.

**Reproduction Steps** (Please include YAML):

**Versions**:
- Chart Version:
- Kubernetes Version (`kubectl version`):

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
This change was merged in https://github.com/kubernetes-sigs/controller-runtime/pull/2668 in upstream controller-runtime. As soon as they release their next version, the hope is that we can bump and we should stop seeing flakes.

jonathan-innis:
[...]

PR #1193, opened 2024-04-18
chore: Retry retrieving default namespace to reduce flakiness

jonathan-innis:
Fixes #N/A 

**Description**

Retry retrieving the default namespace until it gets propagated into etcd and in the cache. This should reduce flakiness called out in https://github.com/kubernetes-sigs/karpenter/issues/887 until controller-runtime v0.18.0 is available for use.

`make presubmit`

### DIFF
--- a/pkg/test/environment.go
+++ b/pkg/test/environment.go
@@ -22,10 +22,13 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


engedaam (commented):
/lgtm

PR #1214, opened 2024-04-25
fix: Only log found provisionable pods with new nodeclaims

jonathan-innis:
Fixes #N/A 

**Description**

This change fixes a regression where we were logging `found provisionable pod(s)` more frequently than we should have been

`make presubmit`


njtran (commented):
/lgtm
/approve

Issue #894, opened 2023-12-19
Add Karpenter High-Level Architecture Diagram


jonathan-innis:
### Description

There have been a few requests from users to better understand how Karpenter runs under the hood and how the different controllers interact with each other. It might be nice to have a high-level architecture diagram that we could point users to when they ask how Karpenter works.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

sftim:
Some tips:
https://www.k8s.dev/blog/2021/12/01/improve-your-documentation-with-mermaid.js-diagrams/

sftim:
and https://kubernetes.io/docs/contribute/style/diagram-guide/

jonathan-innis:
@srijan-shukla and I discussed a bit offline. This one may be tough if you don't already have deep familiarity with the codebase. I'd recommend picking up some issues that fall under [good-first-issues](https://github.com/kubernetes-sigs/karpenter/issues?q=is%3Aopen+is%3Aissue+label%3Agood-first-issue) first to get better familiarity and then come back to this one after you've gotten your feet wet a bit.

PR #1222, opened 2024-05-01
[RFC] Karpenter v1 API & Roadmap Proposal

jonathan-innis:
Fixes #N/A 

**Description**

This proposes the Karpenter stable v1 API and Roadmap. The Roadmap includes features and cleanup tasks that we need to complete before reaching v1.


designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+
+#### Moving `spec.template.spec.kubelet` into the NodeClass
+
+**Category:** Breaking
+
+When the KubeletConfiguration was first introduced into the NodePool, the assumption was that the kubelet configuration is a common interface and that every Cloud Provider supports the same set of kubelet configuration fields.
+
+This turned out not to be the case in reality. For instance, Cloud Providers like Azure [do not support configuring the kubelet configuration through the NodePool API](https://learn.microsoft.com/en-us/azure/aks/node-autoprovision?tabs=azure-cli#:~:text=Kubelet%20configuration%20through%20Node%20pool%20configuration%20is%20not%20supported). Kwok also has no need for the Kubelet API. Shifting these fields into the NodeClass API allows CloudProvider to pick on a case-by-case basis what kind of configuration they want to support through the Kubernetes API.
```
Bryce-Soghigian, line 495: Should we reference the slack conversation here? https://kubernetes.slack.com/archives/C04JW2J5J5P/p1709226455964629

designs/v1-roadmap.md
```diff
@@ -0,0 +1,208 @@
...
+
+Karpenter currently has no conceptual documentation around NodeClaims. NodeClaims have become a fundamental part of how Karpenter launches and manages nodes. There is critical observability information that is stored inside of the NodeClaim that can help users understand when certain disruption conditions are met (Expired, Empty, Drifted) or why the NodeClaim fails to launch.
+
+For Karpenter’s feature completeness at v1, we need to accurately describe to users what the purpose of Karpenter’s NodeClaims are and how to leverage the information that is stored within the NodeClaim to troubleshoot Karpenter’s decision-making.
+
+#### Tasks
+
+- [ ] Add a NodeClaim doc to the “Concepts” section of the documentation
+
+### Drop Knative References from the Code
```
Bryce-Soghigian, line 151: cc: @comtalyst

Bryce-Soghigian, line 151: cc: @comtalyst

Bryce-Soghigian (commented):
Great work on this proposal sir! 

Documentation: Does it make sense to make the azure + aws documentation unification effort happen alongside v1?

Bryce-Soghigian:
Musing for v1: 

E2E Testing for Core: Would there be an interest in testing core via e2es with kwok? There are some bits of functionality that change a lot of behavior in the v1 proposal like the ConsolidateAfter for the WhenUnderutilized consolidation policy. Would it be something we want to take on to have some of the e2e tests go from providers to also being included in core? We should take some additional testing steps for ensuring stability in core releases since we are moving to v1. 

We still should run and develop e2es for our providers, but some of the functionality like consolidation is driven by core, mostly.  Given our providers are both on the latest version of core so you can bump the version with tags to test things there as well. But having some e2es gate consolidation and scheduling etc would be useful for gating code. 

We can even write them to run on multiple managed k8s services (AKS, EKS) with kwok as the cloud provider in multiple clouds(AWS, Azure)? I haven't played enough with kwok to see if something like that would work, but supposedly we could, right?

designs/v1-roadmap.md
```diff
@@ -0,0 +1,208 @@
...
+
+Karpenter needs to stabalize a set of metrics, status conditions, and events that can be relied-upon for monitoring. The design for these metrics, status conditions and events will be added in a separate RFC.
+
+### Stabilize Karpenter’s Tainting Logic
+
+**Issue Ref:** https://github.com/kubernetes-sigs/karpenter/issues/624, https://github.com/kubernetes-sigs/karpenter/issues/1049
+
+**Category:** Breaking
+
+Karpenter wants to expand the usage of the `karpenter.sh/disruption=disrupting:NoSchedule` taint that it currently leverages to cordon nodes during disruption, to also taint nodes with a `karpenterh.sh/disruption-candidate:NoSchedule` taint. By tainting nodes when they become candidates (past expiry, drifted, etc.), we ensure that we will launch new nodes when we get more pods that join the cluster, reducing the chance that we will continue to get `karpenter.sh/do-not-disrupt` pods that continue to schedule to the same node.
```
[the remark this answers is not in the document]
Bryce-Soghigian: nitlord comment.

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+  weight: 10
+  limits:
+    cpu: "1000"
+    memory: 1000Gi
+status:
+  conditions:
+    # Abnormal True Conditions
+    - lastTransitionTime: "2024-02-02T19:54:34Z"
+      status: "True"
+      type: NodeClassReady
```
sftim, line 70: Is this an abnormal-true condition? I'd expect `false` to indicate abnormality here.

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+    - lastTransitionTime: "2024-02-02T19:54:34Z"
+      status: "True"
+      type: Ready
```
sftim, line 73: Is this an abnormal-true condition? I'd expect `false` to indicate abnormality here.

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+  resources:
+    cpu: "2"
+    memory: "100Mi"
```
sftim, line 76:
```suggestion
    memory: "8192Mi"
```
(more plausible)

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+➜  karpenter git:(main) ✗ kubectl get nodepools -o wide
+NAME      NODECLASS  NODES  READY  AGE   WEIGHT  CPU%  MEMORY%
+default   default    4      True   2d7h  100     5.8%  99%
+fallback  default    100    True   2d7h          20%   20%
+```
+
+**Standard Columns**
+
+1. Name
+2. NodeClass - Allows users to easily see the NodeClass that the NodePool is using
```
sftim, line 114: There's a problem: we don't know what the kind is. With this API, there could be two different node classes with the same name but different kinds.

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+      type: Expired
+    - lastTransitionTime: "2024-02-02T19:54:34Z"
+      status: "True"
+      type: Empty
+    - lastTransitionTime: "2024-02-02T19:54:34Z"
+      status: "True"
+      type: Ready
+  nodeName: ip-192-168-71-87.us-west-2.compute.internal
+  providerID: aws:///us-west-2b/i-053c6b324e29d2275
+  imageID: ami-0b1e393fbe12f411c
```
sftim, line 256: How will we handle:
- providers where the machine image specification is a tuple (eg `{image: "e911d3ce-5a58-4e5f-bab1-e5c5492779df", version: 42}`
- node provisioning where there's no image (for example: there's a pool of provisioned bare metal and someone wants Karpenter to wake up servers when needed)

?

It'd be a shame to need a v2 API for those cases.

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+default-fhzpm   c7gd.large   spot           us-west-2b   ip-192-168-165-122.us-west-2.compute.internal   True    2d7h   aws:///us-west-2c/i-053c6b324e29d2275   default   default
+default-rw4vf   c6gn.large   on-demand      us-west-2b   ip-192-168-91-38.us-west-2.compute.internal     True    2d7h   aws:///us-west-2a/i-053c6b324e29d2275   default   default
+default-v5qfb   c7gd.large   spot           us-west-2a   ip-192-168-58-94.us-west-2.compute.internal     True    2d7h   aws:///us-west-2b/i-053c6b324e29d2275   default   default
+```
+
+**Standard Columns**
+
+1. Name
+2. Instance Type
+3. Capacity - Moved from the wide output to the standard output
```
sftim, line 291: Where in the NodeClaim API does Capacity come from?

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+        volumeSize: 100Gi
+        volumeType: gp3
+        iops: 10000
+        encrypted: true
+        kmsKeyID: "1234abcd-12ab-34cd-56ef-1234567890ab"
+        deleteOnTermination: true
+        throughput: 125
+        snapshotID: snap-0123456789
+  detailedMonitoring: **true**
+status:
```
sftim, line 389: Doesn't block v1, but: could we add the ARN of the instance profile into `.status`?

designs/v1-roadmap.md
```diff
@@ -0,0 +1,208 @@
...
+
+**Category:** Breaking, Stability
+
+Karpenter needs to stabalize a set of metrics, status conditions, and events that can be relied-upon for monitoring. The design for these metrics, status conditions and events will be added in a separate RFC.
+
+### Stabilize Karpenter’s Tainting Logic
+
+**Issue Ref:** https://github.com/kubernetes-sigs/karpenter/issues/624, https://github.com/kubernetes-sigs/karpenter/issues/1049
+
+**Category:** Breaking
```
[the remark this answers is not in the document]
sftim: We shouldn't align on `node-lifecycle.kubernetes.io/do-not-disrupt` until it's registered with Kubernetes (see https://kubernetes.io/docs/reference/labels-annotations-taints/ for the list  - you just need a PR against that page to add a registered thing).

So, an easy prerequisite to meet, but still something to remember to do.

sftim:
If we want to announce v1, the Kubernetes blog team can help with that.

designs/v1-roadmap.md
```diff
@@ -0,0 +1,208 @@
...
+
+### NodeClaim Conceptual Documentation
+
+**Issue Ref:** https://github.com/aws/karpenter-provider-aws/issues/5144
+
+**Category:** Stability
+
+Karpenter currently has no conceptual documentation around NodeClaims. NodeClaims have become a fundamental part of how Karpenter launches and manages nodes. There is critical observability information that is stored inside of the NodeClaim that can help users understand when certain disruption conditions are met (Expired, Empty, Drifted) or why the NodeClaim fails to launch.
+
+For Karpenter’s feature completeness at v1, we need to accurately describe to users what the purpose of Karpenter’s NodeClaims are and how to leverage the information that is stored within the NodeClaim to troubleshoot Karpenter’s decision-making.
```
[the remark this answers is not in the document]
jonathan-innis: Agreed. I wasn't directly scoping in #894 to the v1 work, but @sanjeevrg89 volunteered himself to start adding some architecture diagrams to the upstream docs. This should include the NodeClaim lifecycle controller and should hopefully a lot of the detail to outside contributors about how each of the Karpenter controllers interacts with the system.

designs/v1-roadmap.md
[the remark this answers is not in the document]
jonathan-innis: > do we plan on [aligning](https://docs.google.com/document/d/1rHhltfLV5V1kcnKr_mKRKDC4ZFPYGP4Tde2Zy-LE72w/edit) with cas on some of the taints and annotations for v1 as well? Or is this work still in discussion phase

There is an [open issue on this upstream.](https://github.com/kubernetes/autoscaler/issues/6648) We've been actively having some conversations around what makes sense here, but I imagine that, yes, this will be part of the taint redesign so that we don't break everyone twice.

Since the discussion was still ongoing and nothing was set in stone yet, I didn't want to specifically scope it into this item; though, I can add a little bit to make mention to it.

designs/v1-api.md
```diff
@@ -0,0 +1,529 @@
...
+        - key: example.com/another-taint
+          effect: NoSchedule
+      requirements:
+        - key: "karpenter.k8s.aws/instance-category"
+          operator: In
+          values: ["c", "m", "r"]
+          minValues: 2 # Alpha field, added for spot best practices support
+  disruption:
+    budgets:
+      - nodes: 0
```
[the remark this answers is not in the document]
jonathan-innis: Ack, yep. I'll add an example with some reasons.

designs/v1-api.md
jonathan-innis: Nope, mistype. I'll just drop this from the example.

designs/v1-api.md
jonathan-innis: Nope, mistype. I'll just drop this from the example.

designs/v1-api.md
jonathan-innis: How big of a problem do you think that is in the printer columns? I'm hesitant to want to include both the name and the kind/group in an API when most users may name them differently for clarity anyways. Perhaps the wide columns, but even there I'm not fully convinced that it's needed at this point since no CloudProvider supports multiple kinds for NodeClasses right now.

designs/v1-api.md
jonathan-innis: I'm struggling to see a case where an image doesn't exist assigned to a NodeClaim/Instance. In the pre-provisioned case, the NodeClaim will eventually have an image assigned to it when it wakes up. In the tuple case, you could construct a string that would make sense in the context of "imageID" e.g. "e911d3ce-5a58-4e5f-bab1-e5c5492779df/v42"

designs/v1-api.md
jonathan-innis: It comes from the labels -- `karpenter.sh/capacity-type`

designs/v1-api.md
jonathan-innis: What's the need for the ARN vs. just the name? The instance profile will be in the same partition and is global, so I think the ARN should be easily constructed from the name

jonathan-innis:
> E2E Testing for Core: Would there be an interest in testing core via e2es with kwok

jonathan-innis:
> Does it make sense to make the azure + aws documentation unification effort happen alongside v1

This another one that has unknown timelines in my mind and I don't think has any direct impact to users if it comes after v1 (at least from a breaking change perspective). I'd rather just keep working on this effort and not view it as a blocking action for the v1 release.

designs/v1-api.md
```diff
@@ -0,0 +1,535 @@
...
+
+### Changes to the API
+
+#### NodeClaim Spec Immutability
+
+**Category:** Breaking
+
+Karpenter currently doesn’t enforce immutability on NodeClaims in v1beta1, though we implicitly assume that users should not be acting against these objects after creation, as the NodeClaim lifecycle controller won’t react to any change after the initial instance launch.
+
+Karpenter can make every `spec` field immutable on the NodeClaim after its initial creation. This will be enforced through CEL validation, where you can perform a check like `[self == oldSelf](https://kubernetes.io/docs/reference/using-api/cel/#language-overview)` to enforce that the fields cannot have changed after the initial apply. Users who are not on K8s 1.25+ that supports CEL will get the same validation enforced by validating webhooks.
```
toVersus, line 317:
```suggestion
Karpenter can make every `spec` field immutable on the NodeClaim after its initial creation. This will be enforced through CEL validation, where you can perform a check like [`self == oldSelf`](https://kubernetes.io/docs/reference/using-api/cel/#language-overview) to enforce that the fields cannot have changed after the initial apply. Users who are not on K8s 1.25+ that supports CEL will get the same validation enforced by validating webhooks.
```

designs/v1-api.md
```diff
@@ -0,0 +1,535 @@
...
+```
+
+### Defaults
+
+Defaults are unchanged in the v1 API version but are included in the API proposal for completeness.
+
+`disruption.budgets` : `{"nodes": "10%"}`
+`disruption.expireAfter`: `30d/720h`
+`disruption.consolidationPolicy`: `WhenUnderutilized`
+`disruption.consolidateAfter`: `15s`
```
willthames, line 93: I don't see how these can be the current defaults when `consolidateAfter` doesn't yet work with `WhenUnderutilized` (which needs #1218)

GnatorX: ^ +1

designs/v1-api.md
GnatorX: Agree with Jonathan on this one. I don't know if anyone is mixing providers

designs/v1-api.md
```diff
@@ -0,0 +1,535 @@
...
+1. Name
+2. NodeClass - Allows users to easily see the NodeClass that the NodePool is using
+3. Nodes - Allow users to easily view the number of nodes that are associated with a NodePool
+4. Ready - NodePools now have status conditions and will only be used for scheduling when ready. This readiness should be easily viewable by users
+5. Age - This is a common field for all Kubernetes objects and should be maintained in the standard set of columns
+
+**Wide Columns (-o wide)**
+
+1. Weight - Viewing the NodePools that will be evaluated first should be easily observable but may not be immediately useful to all users, particularly if the NodePools are named in a way that already indicate their ordering e.g. suffixed with fallback
+2. CPU% - The capacity of the CPU for all nodes provisioned by this NodePool as a percentage of its limits
```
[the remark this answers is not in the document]
GnatorX: Would love to see pod utilization eventually

designs/v1-api.md
```diff
@@ -0,0 +1,535 @@
...
+default-97v9h   c6gn.large   us-west-2b   ip-192-168-71-87.us-west-2.compute.internal     True    2d7h   spot       default    default
+default-fhzpm   c7gd.large   us-west-2b   ip-192-168-165-122.us-west-2.compute.internal   True    2d7h   spot       default    default
+default-rw4vf   c6gn.large   us-west-2b   ip-192-168-91-38.us-west-2.compute.internal     True    2d7h   spot       default    default
+default-v5qfb   c7gd.large   us-west-2a   ip-192-168-58-94.us-west-2.compute.internal     True    2d7h   spot       default    default
+```
+
+#### Proposed
+
+```
+➜  karpenter git:(main) ✗ kubectl get nodeclaims -A -o wide
```
GnatorX, line 284: what is the -A? There is no namespace in nodeclaim

designs/v1-api.md
```diff
@@ -0,0 +1,535 @@
...
+
+1. Role - As a best practice, we are recommending that users use a Node role and let Karpenter create a managed instance profile on behalf of the customer. We should easily expose this role.
+
+#### Status Conditions
+
+**Category:** Stability
+
+Defining the complete set of status condition types that we will include on v1 launch is **out of scope** of this document and will be defined with more granularly in Karpenter’s Observability design. Minimally for v1, we will add a `Ready` condition so that we can determine whether a EC2NodeClass can be used by a NodePool during scheduling. More robustly, we will define status conditions that ensure that each required “concept” that’s needed for an instance launch is resolved e.g. InstanceProfile resolved, Subnet resolved, Security Groups resolved, etc.
+
+#### Require AMISelectorTerms and Change `amiFamily` to `bootstrapMode`
```
GnatorX, line 485: 👍 that's much more clear

designs/v1-api.md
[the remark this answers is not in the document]
jonathan-innis: Pod utilization / Total Capacity? Honestly, this strikes me as a more useful thing to show in printer columns than the capacity / limit; however, we have to be cautious around the number of writes that either of these updates would generate. Both the capacity values and the pod utilization values are going to change extremely quickly, so if we go with something like this we're going to have to concede that it can only be updated on a poll rather than on a watch

designs/v1-roadmap.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Drift Hash Breaking Change Handling
+
+**Issue Ref(s):** https://github.com/kubernetes-sigs/karpenter/issues/957
+
+**Category:** Stability
+
+Karpenter currently relies on a hash to determine whether certain fields in Karpenter’s NodeClaims have drifted from their owning NodePool and owning EC2NodeClass. Today, this is determined by hashing a set of fields on the NodePool or EC2NodeClass and then validating that this hash still matches the NodeClaim’s hash.
+
+This hashing mechanism works well for additive changes to the API, but does not work well when adding fields to the hashing function that already have a set value on a customer’s cluster. In particular, we have a need to make breaking changes to this hash scheme from these two issues: https://github.com/kubernetes-sigs/karpenter/issues/909 and https://github.com/aws/karpenter-provider-aws/issues/5447.
```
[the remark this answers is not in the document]
jonathan-innis: I'm not sure I follow your question. Are you asking if a user overriding the default value will work? It will work as it normally should -- it should trigger drift and nodes should roll in response to the update

designs/v1-api.md
jonathan-innis: Good call. I'm dropping this line. Added by mistake.

designs/v1-api.md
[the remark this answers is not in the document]
garvinp-stripe: pod utilization / spec.limit.pod if limit was set there. It would be cool to see a more generic approach where if a limit is set then we can show utilization of that resource limit

sftim:
Should we merge this?

designs/v1-api.md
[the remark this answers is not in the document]
jonathan-innis: We ended up choosing to drop this from the printer columns for now. Messing with printer columns was confirmed by api-machinery to not be a breaking change so we may add these percentages in later when CRDs support the ability to perform math calculations from data in the status.

For now, this change would require us to duplicate data into our status -- which feels unnecessary right now when this data can be captured from metrics.

designs/v1-api.md
```diff
@@ -0,0 +1,558 @@
...
+
+#### NodeClaim Spec Immutability
+
+**Category:** Breaking
+
+Karpenter currently doesn’t enforce immutability on NodeClaims in v1beta1, though we implicitly assume that users should not be acting against these objects after creation, as the NodeClaim lifecycle controller won’t react to any change after the initial instance launch.
+
+Karpenter can make every `spec` field immutable on the NodeClaim after its initial creation. This will be enforced through CEL validation, where you can perform a check like [`self == oldSelf`](https://kubernetes.io/docs/reference/using-api/cel/#language-overview)` to enforce that the fields cannot have changed after the initial apply. Users who are not on K8s 1.25+ that supports CEL will get the same validation enforced by validating webhooks.
+
+## EC2NodeClass API [AWS CloudProvider]
```
engedaam, line 341: Should we be merging this here? Should we split out the RFC and open a new one in the provider?

designs/v1-api.md
```diff
@@ -0,0 +1,558 @@
...
+
+```
+apiVersion: karpenter.k8s.aws/v1
+kind: EC2NodeClass
+metadata:
+  name: default
+spec:
```
engedaam, line 348: We need to add `amiFamily` back in here, no?

designs/v1-roadmap.md
```diff
@@ -0,0 +1,270 @@
...
+
+#### Tasks
+
+- [ ] Remove the knative logger and replace with the controller-runtime logger
+- [ ] Update the status condition schema for Karpenter CustomResources to use the [metav1 status condition schema](https://github.com/kubernetes/apimachinery/blob/f14778da5523847e4c07346e3161a4b4f6c9186e/pkg/apis/meta/v1/types.go#L1523)
+
+### Drop Knative Webhook from Karpenter
```
engedaam, line 196: I think we should call out that this is for validation only. 
```suggestion
### Drop Knative Webhook for Validation from Karpenter
```

designs/v1-roadmap.md
```diff
@@ -0,0 +1,270 @@
...
+
+Karpenter supported drift in its alpha state from v0.21-v0.32. During alpha, we worked to build out features and promoted drift to beta after releasing full drift support for all Karpenter configuration. Since releasing drift in beta, we’ve received no feedback that would lead us to believe the feature is unstable or not the right direction for the declarative state of Karpenter.
+
+Since the feature is such a fundamental part to how the declarative state of Karpenter functions, we will promote drift to stable at Karpenter v1.
+
+#### Tasks
+
+- [ ] Remove `Drift` as a feature flag, always enabling it in Karpenter
+
+### Change default TopologySpreadConstraint policy for Deployment from `ScheduleAnyways` to `DoNotSchedule` [AWS CloudProvider]
```
engedaam, line 224: Should we consider making an RFC for AWS provider in that repo? I'm not sure this makes much sense in the k8s-sigs context

engedaam (commented):
Just some small nits, and questions

designs/v1-api.md
jonathan-innis: I went back and forth on it; however, I thought that coalescing the changes in a single document rather than splitting them out everywhere made more sense. When Azure or other CloudProviders have v1 changes, they are welcome to add these changes to the v1 RFC doc that is included here.

designs/v1-api.md
jonathan-innis: Good call

designs/v1-roadmap.md
jonathan-innis: Plan is to drop this long-term generally so I'm fine to have this here since we are definitely going to be fully dropping this in v1.1

designs/v1-api.md
```diff
@@ -0,0 +1,357 @@
...
+
+For more details on the need for shifting this field from the NodePool to the NodeClass, see [the conversation in the #karpenter-dev chat](https://kubernetes.slack.com/archives/C04JW2J5J5P/p1709226455964629).
+
+#### Add `spec.template.spec.terminationGracePeriod`
+
+**Category:** Stability
+
+[Users have asked for the ability to configure a max drain timeout](https://github.com/kubernetes-sigs/karpenter/issues/743) for their nodes. This avoids nodes staying stuck in a draining state on the cluster due to fully blocking PDBs or `karpenter.sh/do-not-disrupt` pods. Starting in v1, we will support a feature that will allow you to configure this max time (called `terminationGracePeriod`). This field will be part of the NodeClaim object and will be static for the lifetime of the NodeClaim. You can template this field onto NodeClaims that are created by NodePools through the `spec.template.spec.terminationGracePeriod` field. 
+
+#### Moving `spec.disruption.expireAfter` to `spec.template.spec.expireAfter`
```
engedaam, line 153: Should we link the RFC here?

designs/v1-api.md
```diff
@@ -0,0 +1,357 @@
...
+          values: ["c", "m", "r"]
+          minValues: 2 # Alpha field, added for spot best practices support
+      expireAfter: 720h | Never
+      terminationGracePeriod: 1d
+  disruption:
+    budgets:
+      - nodes: 0
+        schedule: "0 10 * * mon-fri"
+        duration: 16h
+        reasons:
```
engedaam, line 62: I didn't see disruption by reason section. Should we add that? I saw we did that for termination grace period

engedaam (commented):
Small nits

designs/v1-api.md
jonathan-innis: Thought about it more: Decided to go ahead and break it out so that the AWS provider repo has its own version

designs/v1-roadmap.md
```diff
@@ -0,0 +1,203 @@
...
+
+- [ ] Design and implement a `spec.consolidateAfter` field for the v1 NodePool API, reworking our synchronous wait to ensure that waiting for nodes that haven’t reached the end of their `consolidateAfter` timeframe doesn’t block other disruption evaluation
+
+### Change Expiration from Graceful to Forceful
+
+**Issue/RFC Ref:** https://github.com/kubernetes-sigs/karpenter/pull/1303
+
+**Category:** Breaking
+
+Users care about the removal of their nodes within a certain timeframe due to security requirements. Budgets, `karpenter.sh/do-not-disrupt` pods and blocking PDBs should not stop us from starting the drain of the node after it exceeds its expireAfter value. Once a node has begun evicting, it will remain on the cluster up to its t`erminationGracePeriod` duration (assuming there are blocking PDBs or pods that are running with the `karpenter.sh/do-not-disrupt` annotation) and then be removed.
```
engedaam, line 122:
```suggestion
Users care about the removal of their nodes within a certain timeframe due to security requirements. Budgets, `karpenter.sh/do-not-disrupt` pods and blocking PDBs should not stop us from starting the drain of the node after it exceeds its expireAfter value. Once a node has begun evicting, it will remain on the cluster up to its `terminationGracePeriod` duration (assuming there are blocking PDBs or pods that are running with the `karpenter.sh/do-not-disrupt` annotation) and then be removed.
```

PR #1170, opened 2024-04-05
chore: Add method to ignore Retryable error

jigisha620:
Fixes #N/A 

**Description**
We may want to ignore retryable error in certain scenarios. Added a method for that in the PR.

**How was this change tested?**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1243, opened 2024-05-10
revert changes made in #1164, #1170 and #1172

jigisha620:
Fixes #N/A 

**Description**
This PR reverts the changes that were made in #1164, #1170 and #1172. We don't want to make use of retryable error to drive nodeClaim termination. When this approach was tested for AWS provider we found that some instances could take too long to delete. We do not want to use retryable error to trigger multiple `cloudProvider.Delete()` calls.

NA


jonathan-innis (approved):
/lgtm
/approve

PR #1246, opened 2024-05-12
chore: Factor minValues requirements checking into the InstanceTypes methods

jonathan-innis:
Fixes #N/A 

**Description**

This refactors the minValues code to be reusable by other places were instance types are used and truncated (like in https://github.com/aws/karpenter-provider-aws/pull/6182). It also clarifies some of the usages and checking of minValues.

`make presubmit`

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -27,6 +27,7 @@ import (
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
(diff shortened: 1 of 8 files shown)


pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -136,8 +136,8 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
 		// required
 		replacementHasValidInstanceTypes := false
 		if cmd.Action() == ReplaceAction {
-			cmd.replacements[0].InstanceTypeOptions = filterOutSameType(cmd.replacements[0], candidatesToConsolidate)
-			replacementHasValidInstanceTypes = len(cmd.replacements[0].InstanceTypeOptions) > 0
+			cmd.replacements[0].InstanceTypeOptions, err = filterOutSameType(cmd.replacements[0], candidatesToConsolidate)
```
jmdeal, line 139: Do we ever expect this error to be non-nil since min values is still already validated in compute consolidation? It doesn't hurt to have this check, but if there is a new non-nil scenario, I'm wondering if we should still silently fail here.

jonathan-innis: It could definitely be non-nil. If we filter out the same type and this breaks the minValues requirement, then we should fail

pkg/cloudprovider/types.go
```diff
@@ -123,6 +124,82 @@ func (its InstanceTypes) Compatible(requirements scheduling.Requirements) Instan
...
+//     values: ["c4","c5","m4"]
+//     minValues: 3
+//
+// InstanceTypes: ["c4.large","c5.xlarge","m4.2xlarge"], it PASSES the requirements
+//
+//		we get the map as : {
+//			node.kubernetes.io/instance-type:  ["c4.large","c5.xlarge","m4.2xlarge"],
+//			karpenter.k8s.aws/instance-family: ["c4","c5","m4"]
+//		}
+//	 so, returns empty key.
```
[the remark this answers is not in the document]
jonathan-innis: Yeah, good call. When I reworked this code, I just copy-pastaed this section

pkg/controllers/disruption/multinodeconsolidation.go
jmdeal: Got it, makes sense to me 👍 When I was reading the original implementation I missed that `filterByPriceWithMinValues` would return an empty slice when violated so I thought this was a new error check.

PR #1248, opened 2024-05-13
chore: Add status conditions on nodepool

jigisha620:
Fixes #N/A 

**Description**
We want to add status conditions to nodePool. One of the condition that drives the readiness of nodePool is readiness of underlying nodeClass which has been implemented in this PR.

Unit tests and tested on my local cluster


pkg/controllers/nodepool/status/controller.go
```diff
@@ -37,61 +35,58 @@ import (
...
-	// counts on startup
-	if !c.cluster.Synced(ctx) {
-		return reconcile.Result{RequeueAfter: time.Second}, nil
-	}
 	stored := nodePool.DeepCopy()
-	// Determine resource usage and update nodepool.status.resources
-	nodePool.Status.Resources = c.resourceCountsFor(v1beta1.NodePoolLabelKey, nodePool.Name)
+	var results []reconcile.Result
+	var errs error
+	for _, reconciler := range []nodePoolStatusReconciler{
```
jonathan-innis, line 72: Do we need to couple these controllers or can they be separate controllers entirely?

pkg/controllers/nodepool/status/readiness.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+type Readiness struct {
+	cloudProvider cloudprovider.CloudProvider
+	kubeClient    client.Client
+}
+
+func (n Readiness) Reconcile(ctx context.Context, nodePool *v1beta1.NodePool) (reconcile.Result, error) {
+	nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodePoolReady, "NodePoolNotReady", "unable to resolve underlying nodeClass")
+	nodeClassList := &unstructured.UnstructuredList{}
```
jonathan-innis, line 40: Should this just be a Get()? I think the NodePool can only reference a single NodeClass

pkg/controllers/nodepool/status/readiness.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	for _, nodeClass := range nodeClassList.Items {
+		conditions, found, err := unstructured.NestedSlice(nodeClass.Object, "status", "conditions")
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("resolving nodeClass, %w", err)
+		}
+		if !found {
+			return reconcile.Result{RequeueAfter: time.Second}, nil
+		}
+		for _, condition := range conditions {
```
jonathan-innis, line 60: Where are we checking that the condition that we are grabbing is the `Ready` type?

pkg/controllers/nodepool/status/readiness.go
```diff
@@ -0,0 +1,70 @@
...
+	for _, ncGVK := range n.cloudProvider.GetSupportedNodeClasses() {
+		nodeClassList.SetGroupVersionKind(ncGVK)
+		if err := n.kubeClient.List(ctx, nodeClassList, client.MatchingFields{
+			"metadata.name": nodePool.Spec.Template.Spec.NodeClassRef.Name,
+		}); err != nil {
+			return reconcile.Result{}, fmt.Errorf("resolving nodeClass, %w", err)
+		}
+	}
+	if len(nodeClassList.Items) == 0 {
+		return reconcile.Result{RequeueAfter: time.Second}, nil
```
jonathan-innis, line 50: If the item isn't found, don't we want to mark the status of the NodePool as NotReady?

pkg/controllers/nodepool/status/readiness.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	for _, nodeClass := range nodeClassList.Items {
+		conditions, found, err := unstructured.NestedSlice(nodeClass.Object, "status", "conditions")
```
jonathan-innis, line 53: Does it make sense to create a helper function to get the ready condition out of any path?

Something like `IsReady(conditionPath string) (metav1.ConditionStatus, error)` that returns the condition status value and an error if the path doesn't exist or the ready condition can't be found

pkg/controllers/nodepool/status/controller.go
```diff
@@ -14,21 +14,19 @@ See the License for the specific language governing permissions and
...
 */
 
-package counter
+package status
 
 import (
 	"context"
-	"time"
 
+	"go.uber.org/multierr"
```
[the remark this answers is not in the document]
fmuyassarov: You can see an example here: https://github.com/kubernetes-sigs/karpenter/pull/1258

pkg/controllers/provisioning/provisioner.go
```diff
@@ -217,6 +217,10 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	domains := map[string]sets.Set[string]{}
 	for _, nodePool := range nodePoolList.Items {
+		nodePoolReady := nodePool.StatusConditions().Get(v1beta1.ConditionTypeNodePoolReady)
+		if !nodePoolReady.IsTrue() {
+			return nil, fmt.Errorf("nodepool not ready, %s", nodePoolReady.Message)
```
engedaam, line 222: Do we want to produce an error here if a NodePool is not ready? I think we should log but skip the NodePool for provisioning

pkg/apis/v1beta1/nodepool_status.go
```diff
@@ -25,4 +26,24 @@ type NodePoolStatus struct {
...
 	// +optional
 	Resources v1.ResourceList `json:"resources,omitempty"`
+	// Conditions contains signals for health and readiness
+	// +optional
+	Conditions []op.Condition `json:"conditions,omitempty"`
+}
+
+var (
+	// 	ConditionTypeNodePoolReady = "Ready" condition indicates that underlying nodeClass was resolved
+	ConditionTypeNodePoolReady = "Ready"
```
engedaam, line 36: What is the thought around adding NodeClassReady in to the NodePool?

pkg/controllers/nodepool/status/controller.go
jigisha620: I think it makes sense to have these controllers coupled because both of them make update to nodepool status. We did something similar for the AWS provider where we coupled the nodeClass controllers that update status. Thoughts?

pkg/controllers/nodepool/status/controller.go
```diff
@@ -0,0 +1,134 @@
...
+	nodeClass.SetKind(nodePool.Spec.Template.Spec.NodeClassRef.Kind)
+	supportedNC := c.cloudProvider.GetSupportedNodeClasses()
+	if len(nodeClass.GetAPIVersion()) == 0 {
+		nodeClass.SetAPIVersion(fmt.Sprintf("%s/%s", supportedNC[0].Group, supportedNC[0].Version))
+	}
+	if len(nodeClass.GetKind()) == 0 {
+		nodeClass.SetKind(supportedNC[0].Kind)
+	}
+	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		return reconcile.Result{}, fmt.Errorf("resolving nodeClass, %w", err)
```
jonathan-innis, line 75: If we don't find the NodeClass that we're referencing, is this a reason to set the readiness to false?

pkg/controllers/nodepool/status/controller.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	ready, err := c.IsReady(status.ConditionReady, nodeClass)
```
jonathan-innis, line 77: What are the cases that this returns an error? Is this possible for us to hit?

pkg/controllers/nodepool/status/controller.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil || !ready {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "NodeClassNotReady", "unable to resolve underlying nodeClass")
```
jonathan-innis, line 79:
```suggestion
		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "FailedNodeClass", "unable to resolve underlying nodeClass")
```

When we're creating a condition reason, I think we should be more specific about the actual _reason_ why we failed to set this to true. Maybe we should take the `Reason` from the NodeClass readiness condition and the message and copy it over here?

jonathan-innis, line 79:
```suggestion
		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "NodeClassNotReady", "Unable to resolve underlying nodeClass")
```

nit: Messages should always start with a capital letter. Same thing with reasons, but you've already got that 👍🏼

pkg/controllers/nodepool/status/controller.go
```diff
@@ -0,0 +1,134 @@
...
+	if ready {
+		nodePool.StatusConditions().SetTrue(v1beta1.ConditionTypeNodeClassReady)
+	}
+	if !equality.Semantic.DeepEqual(stored, nodePool) {
+		if e := c.kubeClient.Status().Update(ctx, nodePool); e != nil {
+			err = multierr.Append(err, client.IgnoreNotFound(e))
+		}
+	}
+	if err != nil {
+		if errors.IsConflict(err) {
```
jonathan-innis, line 90: Should we handle the conflict next to the actual update call?

pkg/controllers/nodepool/status/controller.go
```diff
@@ -0,0 +1,134 @@
...
+			return reconcile.Result{Requeue: true}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	return reconcile.Result{Requeue: true}, nil
```
jonathan-innis, line 95: Why do we want to requeue with exponential backoff here?

pkg/controllers/nodepool/status/controller.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+// IsReady checks if status of condition passed as conditionPath is true or not
+func (c *Controller) IsReady(conditionPath string, nodeClass *unstructured.Unstructured) (bool, error) {
```
jonathan-innis, line 99: Is there anything upstream that already exists upstream for this? If not, is this something that should potentially live in [`operatorpkg`](https://github.com/awslabs/operatorpkg)

pkg/controllers/provisioning/provisioner.go
```diff
@@ -215,6 +217,11 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	domains := map[string]sets.Set[string]{}
 	for _, nodePool := range nodePoolList.Items {
+		nodePoolReady := nodePool.StatusConditions().Get(status.ConditionReady)
```
jonathan-innis, line 220: nit: I'd inline L220 into L221

pkg/controllers/provisioning/provisioner.go
```diff
@@ -215,6 +217,11 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
...
+		if !nodePoolReady.IsTrue() {
+			logging.FromContext(ctx).With("nodepool", nodePool.Name).Errorf("skipping, nodepool not ready")
```
jonathan-innis, line 222: Is this an error or a debug log? I think a debug log, personally. Since this is already surfaced through the readiness of the NodePool, I think we can just add this as additional info, but not something that we should be screaming about in our logs

pkg/utils/nodepool/nodepool.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+)
+
+// NodeClassEventHandler is a watcher on v1beta1.NodePool that maps NodeClass to NodePools based
+// on the nodeClassRef and enqueues reconcile.Requests for the NodePool
+func NodeClassEventHandler(c client.Client) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, o client.Object) (requests []reconcile.Request) {
+		nodePoolList := &v1beta1.NodePoolList{}
+		if err := c.List(ctx, nodePoolList, client.MatchingFields{
```
jonathan-innis, line 35: Why is this a list and not a Get()? If this is done elsewhere, should we consider also moving that other call to a Get()?

go.mod
```diff
@@ -111,3 +111,5 @@ retract (
 	v0.34.4 // accidentally published incomplete patch release
 	v0.27.7 // accidentally published incomplete patch release
 )
+
+replace github.com/awslabs/operatorpkg => github.com/jigisha620/operatorpkg v0.0.0-20240617015706-4398356b409b
```
jigisha620, line 115: This will be removed once the operatorpkg PR is merged.

pkg/utils/nodepool/nodepool.go
jigisha620: If I understand this correctly, we don't have an object key to make a Get() call here.

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -262,6 +266,8 @@ func (c *CloudProvider) Name() string {
 	return "fake"
 }
 
-func (c *CloudProvider) GetSupportedNodeClasses() []schema.GroupVersionKind {
-	return c.NodeClassGroupVersionKind
+func (c *CloudProvider) GetSupportedNodeClasses() []status.Object {
+	nodeClasses := make([]status.Object, 1)
+	nodeClasses[0] = status.Object(&v1test1.TestNodeClass{})
+	return nodeClasses
```
jonathan-innis, line 272:
```suggestion
	return []status.Object{&v1.TestNodeClass{}}
```

I'd just call it v1 or v1alpha1. Best to keep it as close to what an actual version would look like as possible

jonathan-innis: Also, you can simplify all this with a statically declared slice

kwok/cloudprovider/cloudprovider.go
```diff
@@ -122,14 +123,10 @@ func (c CloudProvider) Name() string {
...
-		{
-			Group:   v1alpha1.SchemeGroupVersion.Group,
-			Version: v1alpha1.SchemeGroupVersion.Version,
-			Kind:    "KWOKNodeClass",
-		},
-	}
+func (c CloudProvider) GetSupportedNodeClasses() []status.Object {
+	nodeClasses := make([]status.Object, 1)
+	nodeClasses[0] = status.Object(&v1alpha1.KWOKNodeClass{})
+	return nodeClasses
```
jonathan-innis, line 129:
```suggestion
	return []status.Object{&v1alpha1.KWOKNodeClass{}}
```

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -115,8 +117,9 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim
 
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	builder := controllerruntime.NewControllerManagedBy(m)
-	for _, ncGVK := range c.cloudProvider.GetSupportedNodeClasses() {
+	for _, supportedNC := range c.cloudProvider.GetSupportedNodeClasses() {
 		nodeclass := &unstructured.Unstructured{}
```
jonathan-innis, line 121: Do you even need to do unstructured if you have the `status.Object`? I think you can just directly use the types now

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+	if ready.IsFalse() {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, ready.Reason, ready.Message)
+		return false
+	}
+	nodePool.StatusConditions().SetTrue(v1beta1.ConditionTypeNodeClassReady)
+	return true
+}
+
+func (c *Controller) hasNodeClass(ctx context.Context, supportedNC []status.Object, nodePool *v1beta1.NodePool) (status.Object, error) {
+	if len(supportedNC) == 0 {
```
jonathan-innis, line 105: IMO, this should be a tenant of the CloudProvider that you _must_ have at least one of these. We should just crash if there isn't one defined IMO

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1beta1.NodePool) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("nodepool.status").With("nodepool", nodePool.Name))
+	ctx = injection.WithControllerName(ctx, "nodepool.status")
+	stored := nodePool.DeepCopy()
+	supportedNC := c.cloudProvider.GetSupportedNodeClasses()
+	var errs error
+	var ready bool
+	nodeClass, err := c.hasNodeClass(ctx, supportedNC, nodePool)
```
jonathan-innis, line 69: Do you need a helper for this or can you just call the Get() and check that it exists inline?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "UnresolvedNodeClass", "Unable to resolve nodeClass")
+		errs = multierr.Append(errs, err)
+	} else {
+		ready = c.setReadyCondition(nodePool, nodeClass)
```
jonathan-innis, line 74: Can you call this inline?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1beta1.NodePool) (reconcile.Result, error) {
+	ctx = logging.WithLogger(ctx, logging.FromContext(ctx).Named("nodepool.status").With("nodepool", nodePool.Name))
+	ctx = injection.WithControllerName(ctx, "nodepool.status")
+	stored := nodePool.DeepCopy()
+	supportedNC := c.cloudProvider.GetSupportedNodeClasses()
+	var errs error
+	var ready bool
+	nodeClass, err := c.hasNodeClass(ctx, supportedNC, nodePool)
+	if err != nil {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "UnresolvedNodeClass", "Unable to resolve nodeClass")
```
jonathan-innis, line 71: Should this be set on any error or only on a NotFound?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+		errs = multierr.Append(errs, err)
+	} else {
+		ready = c.setReadyCondition(nodePool, nodeClass)
+	}
+	if !equality.Semantic.DeepEqual(stored, nodePool) {
+		if err := c.kubeClient.Status().Patch(ctx, nodePool, client.MergeFrom(stored)); err != nil {
```
jonathan-innis, line 77:
```suggestion
		if err := c.kubeClient.Status().Update(ctx, nodePool, client.MergeFrom(stored)); err != nil {
```

Status conditions should be updated with `Update()` rather than `Patch()`. When you have async controller calls you risk racing and fighting with each other

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+	}
+	if errs != nil {
```
jonathan-innis, line 84: Do you need to capture all of the errors together or can you just call the requeue/reconcile right when the error is raised? I'm not sure that you need a multierr like you have here

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+		return reconcile.Result{}, errs
+	}
+	return reconcile.Result{Requeue: !ready}, nil
```
jonathan-innis, line 87: Why do we need to requeue until it's ready? I think we should just requeue on updates to the NodeClass readiness status.

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (c *Controller) setReadyCondition(nodePool *v1beta1.NodePool, nodeClass status.Object) bool {
+	ready := nodeClass.StatusConditions().Get(status.ConditionReady)
+	if ready.IsUnknown() {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "NodeClassNotReady", "Node Class Not Ready")
```
jonathan-innis, line 93: Should be more descriptive with our reasons and message?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, supportedNC[0]); err != nil {
+		return nil, fmt.Errorf("resolving nodeClass, %w", err)
+	}
+	return supportedNC[0], nil
+}
+
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	builder := controllerruntime.NewControllerManagedBy(m)
+	for _, supportedNC := range c.cloudProvider.GetSupportedNodeClasses() {
```
jonathan-innis, line 116: Like with the other controller, I don't think that we need to use unstructured here. I think we can just directly use the types

pkg/test/nodeclass.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/awslabs/operatorpkg/status"
+
+	"sigs.k8s.io/karpenter/pkg/test/v1test1"
+
+	"github.com/imdario/mergo"
+)
+
+// NodeClass creates a test NodeClass with defaults that can be overridden by overrides.
+// Overrides are applied in order, with a last write wins semantic.
+func NodeClass(overrides ...v1test1.TestNodeClass) *v1test1.TestNodeClass {
```
jonathan-innis, line 31:
```suggestion
func TestNodeClass(overrides ...v1test1.TestNodeClass) *v1test1.TestNodeClass {
```

jonathan-innis: Only to match the name

pkg/test/nodepool.go
```diff
@@ -51,6 +51,9 @@ func NodePool(overrides ...v1beta1.NodePool) *v1beta1.NodePool {
 	if override.Spec.Template.Spec.Requirements == nil {
 		override.Spec.Template.Spec.Requirements = []v1beta1.NodeSelectorRequirementWithMinValues{}
 	}
+	if override.Status.Conditions == nil {
```
jonathan-innis, line 54: Is there a reason we would not want to set the status conditions in this way in the helper? I guess we can always force it to false later if we need?

pkg/test/v1test1/doc.go
```diff
@@ -0,0 +1,38 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	Group = "karpenter.test.sh"
+)
+
+func init() {
+	gv := schema.GroupVersion{Group: Group, Version: "v1test1"}
+	v1.AddToGroupVersion(scheme.Scheme, gv)
```
jonathan-innis, line 34: I don't think that you need this line anymore

pkg/test/v1test1/doc.go
```diff
@@ -0,0 +1,38 @@
...
+	scheme.Scheme.AddKnownTypes(gv,
+		&TestNodeClass{},
```
jonathan-innis, line 36: Do you have to add the `&TestNodeClassList{}` to the set of known types as well?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -225,6 +227,10 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	domains := map[string]sets.Set[string]{}
 	for _, nodePool := range nodePoolList.Items {
+		if !nodePool.StatusConditions().IsTrue(status.ConditionReady) {
```
jonathan-innis, line 230: Do we have a provisioning test somewhere to ensure that we don't use NotReady NodePools?

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,119 @@
...
+		cloudProvider.InstanceTypes = fake.InstanceTypesAssorted()
+		nodePool = test.NodePool()
+		nodeClass = test.NodeClass(v1test1.TestNodeClass{
+			ObjectMeta: metav1.ObjectMeta{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name},
+		})
+	})
+	AfterEach(func() { cloudProvider.NodeClassGroupVersionKind = nil })
+	It("should have status condition on nodePool as not ready due to error when getting nodeClass", func() {
+		ExpectApplied(ctx, env.Client, nodePool)
+		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodePoolReadinessController, nodePool)
```
jonathan-innis, line 81: If the error is a NotFound error, should we return an error? In actuality, requeueing is going to be pointless and there isn't really an error here outside of the status condition. I could see us logging once, but I'm not sure it's worth throwing it back

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,119 @@
...
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		Expect(nodePool.StatusConditions().IsTrue(status.ConditionReady)).To(BeFalse())
```
jonathan-innis, line 83: Should it be the reverse of this? We shouldn't just validate that it's not True, but we should validate that it's False()?

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,119 @@
...
+	})
+	It("should have status condition on nodePool as ready if nodeClass is ready", func() {
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		_ = ExpectObjectReconciled(ctx, env.Client, nodePoolReadinessController, nodePool)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		Expect(nodePool.StatusConditions().IsTrue(status.ConditionReady)).To(BeTrue())
```
jonathan-innis, line 89: nice!

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,119 @@
...
+				},
+			},
+		}
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		result := ExpectObjectReconciled(ctx, env.Client, nodePoolReadinessController, nodePool)
+		Expect(result.Requeue).To(BeTrue())
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		Expect(nodePool.StatusConditions().IsTrue(status.ConditionReady)).To(BeFalse())
+	})
+	It("should have status condition on nodePool as not ready if nodeClass does not have status conditions", func() {
```
jonathan-innis, line 109: Awesome testing coverage here! Great work!

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,119 @@
...
+		nodeClass.Status = v1test1.TestNodeClassStatus{
+			Conditions: []status.Condition{},
+		}
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		result := ExpectObjectReconciled(ctx, env.Client, nodePoolReadinessController, nodePool)
+		Expect(result.Requeue).To(BeTrue())
```
jonathan-innis, line 115: Do we requeue in this case? Or just wait for the update from the NodeClass?

pkg/test/nodeclass.go
jigisha620: Kept it as NodeClass because of the warning that function name should not start with package name.

pkg/test/v1test1/doc.go
jigisha620: I think we still need it. I see errors if we don't have this.
`v1.GetOptions is not suitable for converting to "karpenter.test.sh/v1alpha1" in scheme "pkg/runtime/scheme.go:100"`

pkg/controllers/nodepool/readiness/controller.go
jigisha620: Did this to reduce the cyclomatic complexity of the function.

pkg/controllers/nodepool/readiness/suite_test.go
jigisha620: Thank you.

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,121 @@
...
+		panic("no supported node classes found for the cloud provider")
+	}
+	nodeClass := supportedNC[0]
+	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		// If nodeClass is not found then we set status condition on nodePool as not ready
+		if errors.IsNotFound(err) {
+			log.FromContext(ctx).WithValues("NodeClass", nodePool.Spec.Template.Spec.NodeClassRef.Name).Info("nodeClass not found")
+			nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "UnresolvedNodeClass", "Unable to resolve nodeClass")
+		}
+		if !equality.Semantic.DeepEqual(stored, nodePool) {
```
njtran, line 73: One way to de-dupe this from the same code block on L84 is to make the Get() call in a separate function and return it. If it's nil, set the condition to false, if it's not, you set it to true, and only make the one patch call.

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,121 @@
...
+	ready := nodeClass.StatusConditions().Get(status.ConditionReady)
+	if ready.IsUnknown() {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, "NodeClassReadinessUnknown", "Node Class Readiness Unknown")
+	}
+	if ready.IsFalse() {
+		nodePool.StatusConditions().SetFalse(v1beta1.ConditionTypeNodeClassReady, ready.Reason, ready.Message)
+	}
+	if ready.IsTrue() {
+		nodePool.StatusConditions().SetTrue(v1beta1.ConditionTypeNodeClassReady)
+	}
```
njtran, line 105: nit: consider doing 1) rather than 2) to portray that they're mutually exclusive
1)
```
if
elif
else 
```
2)
```
if 
if
if
```

njtran (commented):
Nice work!

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,120 @@
...
+// NewController is a constructor
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
+	return &Controller{
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1beta1.NodePool) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "nodepool.status")
```
njtran, line 61: should match the name of the package
```suggestion
	ctx = injection.WithControllerName(ctx, "nodepool.readiness")
```

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,120 @@
...
+
+	if len(supportedNC) == 0 {
+		logger.Fatal("no supported node classes found for the cloud provider")
+	}
+	nodeClass, err := c.getNodeClass(ctx, nodePool, supportedNC)
+	if err != nil {
+		if !errors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		log.FromContext(ctx).WithValues("NodeClass", nodePool.Spec.Template.Spec.NodeClassRef.Name).Info("nodeClass not found")
```
njtran, line 73: Do we need to emit a log here? Seems like an status condition should be sufficient.  I'd say an event on the nodepool maybe too

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,120 @@
...
+			}
+			return reconcile.Result{}, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+func (c *Controller) getNodeClass(ctx context.Context, nodePool *v1beta1.NodePool, supportedNC []status.Object) (status.Object, error) {
+	nodeClass := supportedNC[0]
+	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		return nil, err
```
njtran, line 91: you could do the ignore not found error here, and then when you consume the results of this you can just check with 

```
if err != nil {
    return
} 
if nodeclass != nil {
    setNotReady()
} else {
    setReady()
}
```

Personally, i think that's cleaner, but up to you

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,120 @@
...
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	builder := controllerruntime.NewControllerManagedBy(m)
+	for _, nodeClass := range c.cloudProvider.GetSupportedNodeClasses() {
+		builder = builder.Watches(
+			nodeClass,
+			nodepool.NodeClassEventHandler(c.kubeClient),
+		)
+	}
+	return builder.
+		Named("nodepool.status").
```
njtran, line 116: same here on matching file name
```suggestion
		Named("nodepool.readiness").
```

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/awslabs/operatorpkg/status"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/karpenter/pkg/apis"
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider/fake"
+	nodepoolreadiness "sigs.k8s.io/karpenter/pkg/controllers/nodepool/readiness"
```
njtran, line 36: is there a reason this has to be aliased?

pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"sigs.k8s.io/karpenter/pkg/test"
+	. "sigs.k8s.io/karpenter/pkg/test/expectations"
+	. "sigs.k8s.io/karpenter/pkg/utils/testing"
+)
+
+var (
+	nodePoolReadinessController *nodepoolreadiness.Controller
```
njtran, line 43: it'd be nice if we could just call this controller for simplicity
```suggestion
	controller *readiness.Controller
```

pkg/controllers/provisioning/provisioner.go
```diff
@@ -288,7 +295,9 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*v1.Pod, stateNod
 			}
 		}
 	}
-
+	if len(notReadyNodePools) > 0 {
+		log.FromContext(ctx).WithValues("nodePools", nodePoolList).Info("skipped nodePool(s) due to not ready")
```
njtran, line 299: nit on this, up to you. Do whatever you think looks best. Personally don't think you need the (s)
```suggestion
		log.FromContext(ctx).WithValues("nodePools", nodePoolList).Info("skipped nodePools, not ready")
```

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -0,0 +1,117 @@
...
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
```
njtran, line 41: one nit i just noticed. If you put lines in between imports, the linter won't sort them. Best to do 3 groups here

1. golang native libraries
2. non sigs/karpenter libraries
3. sigs/karpenter libraries

njtran (commented):
/lgtm
/approve

njtran (commented):
/lgtm
/approve

pkg/controllers/nodepool/readiness/controller.go
jigisha620: Realized that nodeClass will never will be nil in this case, that's because cloudProvider returns []status.Object which is populates with nodeClass object. For example EC2NodeClass in case of aws provider.

pkg/test/nodepool.go
```diff
@@ -53,6 +53,7 @@ func NodePool(overrides ...v1beta1.NodePool) *v1beta1.NodePool {
 	}
 	if override.Status.Conditions == nil {
 		override.StatusConditions().SetTrue(v1beta1.ConditionTypeValidationSucceeded)
+		override.StatusConditions().SetTrue(v1beta1.ConditionTypeNodeClassReady)
```
njtran, line 56: Shouldn't this be done automatically by operatorpkg?

pkg/controllers/nodepool/readiness/controller.go
njtran: If the NodeClass doesn't exist, we should return it as nil, no?

pkg/test/nodepool.go
jigisha620: `operatorpkg` would take care of the base condition which is `Ready` for nodePool. This is `NodeClassReady` condition which is one of the conditions that will affect nodePool's readiness.

pkg/controllers/nodepool/readiness/controller.go
jigisha620: If we check for `notFounderror` like this, then we don't return nil. This is how I had it earlier -
```
if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass); client.ignoreNotFound(err)!=nil {
      return nil, err
}
return nodeClass,nil
```
I have updated it now so that it checks for notFound and then returns nil.

njtran (commented):
/lgtm
/approve

PR #1254, opened 2024-05-13
chore: Don't require message or reason to be set for NodeClaim status conditions

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that our use of `metav1.Condition` doesn't cause validation failures when making updates to the NodeClaim. We can drop this injected removal of these required fields when we get to v1, where we can use conversion to automatically set these values.

`make presubmit`


njtran (commented):
/lgtm
/approve

PR #1258, opened 2024-05-15
chore: Migrate to Go's native multierror handling

fmuyassarov:
Fixes #N/A

**Description**
Now that Go supports handling multiple errors directly from version 1.20 onwards, this commit removes the need for an external library and instead uses native methods of multi-error handling.

Unit tests.

### DIFF
--- a/pkg/apis/v1beta1/nodeclaim_validation.go
+++ b/pkg/apis/v1beta1/nodeclaim_validation.go
@@ -18,12 +18,12 @@ package v1beta1
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
 
 	"github.com/samber/lo"
-	"go.uber.org/multierr"
 	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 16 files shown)


pkg/apis/v1beta1/nodeclaim_validation.go
```diff
@@ -142,43 +142,43 @@ func (in *NodeClaimSpec) validateRequirements() (errs *apis.FieldError) {
...
 
 func ValidateRequirement(requirement NodeSelectorRequirementWithMinValues) error { //nolint:gocyclo
-	var errs error
+	errs := []error{}
 	if normalized, ok := NormalizedLabels[requirement.Key]; ok {
 		requirement.Key = normalized
 	}
 	if !SupportedNodeSelectorOps.Has(string(requirement.Operator)) {
-		errs = multierr.Append(errs, fmt.Errorf("key %s has an unsupported operator %s not in %s", requirement.Key, requirement.Operator, SupportedNodeSelectorOps.UnsortedList()))
+		errs = append(errs, fmt.Errorf("key %s has an unsupported operator %s not in %s", requirement.Key, requirement.Operator, SupportedNodeSelectorOps.UnsortedList()))
```
jonathan-innis, line 150: I'd be interested in taking another look at this. Do you know how this changes the formatting of errors? When we first approached this, we opted not to do it because we saw it injected newlines and created some additional weirdness.

fmuyassarov: I think if I fix the how unit tests are expecting the error messages then it should be fine. But I don't see any weirdness apart from a new line based on the example above.

fmuyassarov: But to be honest, I personally didn't like that a newline is added. Because it kind of breaks idea embedding context information to the previous error. Apparently we can still construct something similar to `key x has an unsupported operator y not in z; key x has an unsupported controller y not in z` by Go itself, using `%w` format specifiers. 
For example:
```Go
var errs error
operatorError := fmt.Errorf("key %s has an unsupported operator %s not in %s", "x", "y", "z")
errs = fmt.Errorf("%w ; %w", errs, operatorError)
```

But the problem with this approach is that, initial value of the errs will be `nil`, and when constructing a wrapped error with `fmt.Errorf("%w....)`, the end result will have nil value added, because `Errorf` can't discard nil error values unfortunately. 

So, if we use `Errorf` with `%w` format specifier then we have the issue of `nil` value appearing and on the other hand if we use the `errors.Join()` then we end up with multine error in case when there is more than a single error.

Taking all these into account, I'm fine to close this PR for the timing being since I don't see a nice way around.

PR #1262, opened 2024-05-16
feat: create constraints on offerings

rschalo:
Fixes 
https://github.com/aws/karpenter-provider-aws/issues/3878
https://github.com/aws/karpenter-provider-aws/issues/4612

**Description**
This change allows cloudprovider specific constraints to be added to Offering. Zone and capacityType are special labels that are used across the Karpenter codebase.

There shouldn't be the need for new tests as long as the current test suite passes. There will be additional tests required in karpenter-aws-provider in order to validate additional constraints. Starting this draft PR for discussion.


pkg/cloudprovider/fake/instancetype.go
```diff
@@ -194,3 +197,10 @@ func PriceFromResources(resources v1.ResourceList) float64 {
 	}
 	return price
 }
+
+func constraints(capacityType, zone string) map[string]string {
+	return map[string]string{
+		v1beta1.CapacityTypeLabelKey: capacityType,
+		v1.LabelTopologyZone:         zone,
+	}
+}
```
Bryce-Soghigian, line 206: you define this same helper function in multiple places maybe we can put it in our helpers?

pkg/cloudprovider/types.go
```diff
@@ -231,7 +231,7 @@ type Offerings []Offering
 // passed zone and capacity type
 func (ofs Offerings) Get(ct, zone string) (Offering, bool) {
```
jmdeal, line 232: I know we had discussed leaving this function as is previously, but having thought about it some more I think we probably should make this generic across offering constraints for consistency. I think this would be simple enough to handle by constructing a set of requirements from the disruption candidate labels and passing them rather than the capacity type and zone.

pkg/cloudprovider/types.go
```diff
@@ -215,11 +215,11 @@ func (i InstanceTypeOverhead) Total() v1.ResourceList {
...
 
 // An Offering describes where an InstanceType is available to be used, with the expectation that its properties
-// may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type)
+// may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type) and
+// these properties are captured with labels in the Constraints map
 type Offering struct {
-	CapacityType string
-	Zone         string
-	Price        float64
+	Constraints map[string]string
```
jmdeal, line 221: Can we add a comment calling out which constraints are explicitly required? For example, we require a capacity-type constraint for disruption.

rschalo: Yep, I think that we keep Zone and CapacityType as required.

pkg/cloudprovider/types.go
```diff
@@ -245,8 +245,8 @@ func (ofs Offerings) Available() Offerings {
 // Compatible returns the offerings based on the passed requirements
 func (ofs Offerings) Compatible(reqs scheduling.Requirements) Offerings {
```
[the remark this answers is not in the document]
rschalo: It could be but then the concepts of Zone and CapacityType are no longer special and I think we want to keep them that way. This also forces other cloudproviders to think about constraints besides those two.

What this does is that if there are additional constraints specific to a cloudprovider then they can handle it within their implementation when initially creating offerings like at https://github.com/aws/karpenter-provider-aws/pull/6216/files#diff-b0143481b3c25f93f1ad9de2b0a7b67dbbe61c01c4187ade18c26759a0ff5c04R343

WDYT?

pkg/cloudprovider/types.go
rschalo: I think we should consider leaving Zone and CapacityType as the options in Get() and Compatible(). More reasoning above.

pkg/cloudprovider/types.go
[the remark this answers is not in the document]
jmdeal: The problem is that we do need to consider these additional constraints in the upstream code. Consider a case where we have a `NotIn` requirement for a zone-id. When we're computing consolidation decisions and filtering for offerings, we would still include the offering with that zone-id since we're only filtering on zone. This could lead to us making incorrect decisions around pricing.

pkg/cloudprovider/types.go
rschalo: Discussed offline - generalized in next commit.

pkg/test/helpers.go
```diff
@@ -0,0 +1,30 @@
...
+
+package test
+
+import (
+	v1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+)
+
+func Constraints(capacityType, zone string) map[string]string {
```
jmdeal, line 25: nit: we tend to lean away from global test helpers, personally I'd be happy with constructing the constraint map in the tests. I can go either way here though.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -284,8 +284,8 @@ func fits(instanceType *cloudprovider.InstanceType, requests v1.ResourceList) bo
 
 func hasOffering(instanceType *cloudprovider.InstanceType, requirements scheduling.Requirements) bool {
```
jmdeal, line 285: IMO this should also be made generic, or maybe we can leverage `offerings.Compatible` directly and then further filter on available.

pkg/controllers/disruption/helpers.go
```diff
@@ -317,7 +317,14 @@ func worstLaunchPrice(ofs []cloudprovider.Offering, reqs scheduling.Requirements
 	// We prefer to launch spot offerings, so we will get the worst price based on the node requirements
 	if reqs.Get(v1beta1.CapacityTypeLabelKey).Has(v1beta1.CapacityTypeSpot) {
 		spotOfferings := lo.Filter(ofs, func(of cloudprovider.Offering, _ int) bool {
```
jmdeal, line 319: Can we leverage `ofs.Compatible` here? Something along the lines of:
```go
spotOfferings := ofs.Compatible(scheduling.NewRequirements(scheduling.NewRequirement(v1beta1.CapacityTypeLabelKey, v1.NodeSelectorOpIn, v1beta1.CapacityTypeSpot)))
```
Same for the on-demand case.

pkg/controllers/provisioning/scheduling/instance_selection_test.go
```diff
@@ -43,8 +43,8 @@ var _ = Describe("Instance Type Selection", func() {
 	var minPrice float64
 	var instanceTypeMap map[string]*cloudprovider.InstanceType
 	nodePrice := func(n *v1.Node) float64 {
-		of, _ := instanceTypeMap[n.Labels[v1.LabelInstanceTypeStable]].Offerings.Get(n.Labels[v1beta1.CapacityTypeLabelKey], n.Labels[v1.LabelTopologyZone])
-		return of.Price
+		return instanceTypeMap[n.Labels[v1.LabelInstanceTypeStable]].Offerings.Compatible(
```
jmdeal, line 46: Any reason to not use `Get` here, and use `Cheapest`? Only one offering should match since well known labels have already been resolved on the node right?

pkg/controllers/provisioning/scheduling/instance_selection_test.go
```diff
@@ -485,7 +485,7 @@ var _ = Describe("Instance Type Selection", func() {
 		// remove all Arm instance types in zone-2
 		cloudProvider.InstanceTypes = filterInstanceTypes(cloudProvider.InstanceTypes, func(i *cloudprovider.InstanceType) bool {
 			for _, off := range i.Offerings {
-				if off.Zone == "test-zone-2" {
+				if off.Requirements.Get(v1.LabelTopologyZone).Has("test-zone-2") {
```
jmdeal, line 488: nit:
```suggestion
				if off.Zone() == "test-zone-2" {
```

pkg/controllers/provisioning/scheduling/instance_selection_test.go
```diff
@@ -514,7 +514,7 @@ var _ = Describe("Instance Type Selection", func() {
...
```
jmdeal, line 517: nit:
```suggestion
				if off.Zone() == "test-zone-2" {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3795,7 +3801,8 @@ func ExpectInstancesWithOffering(instanceTypes []*cloudprovider.InstanceType, ca
 	for _, it := range instanceTypes {
 		matched := false
 		for _, offering := range it.Offerings {
-			if offering.CapacityType == capacityType && offering.Zone == zone {
+			if offering.Requirements.Get(v1beta1.CapacityTypeLabelKey).Values()[0] == capacityType &&
+				offering.Requirements.Get(v1.LabelTopologyZone).Values()[0] == zone {
```
jmdeal, line 3805: nit
```suggestion
			if offering.CapacityType() == capacityType && offering.Zone() == zone {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3814,7 +3821,7 @@ func ExpectInstancesWithLabel(instanceTypes []*cloudprovider.InstanceType, label
 			{
 				matched := false
 				for _, offering := range it.Offerings {
-					if offering.Zone == value {
+					if offering.Requirements.Get(v1.LabelTopologyZone).Has(value) {
```
jmdeal, line 3824:
```suggestion
					if offering.Zone() == value {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3825,12 +3832,12 @@ func ExpectInstancesWithLabel(instanceTypes []*cloudprovider.InstanceType, label
 			{
 				matched := false
 				for _, offering := range it.Offerings {
-					if offering.CapacityType == value {
+					if offering.Requirements.Get(v1beta1.CapacityTypeLabelKey).Has(value) {
```
jmdeal, line 3835:
```suggestion
					if offering.CapacityType() == value {
```

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -108,7 +108,7 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *v1beta1.NodeClaim
 	np := &v1beta1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: nodeClaim.Labels[v1beta1.NodePoolLabelKey]}}
 	instanceTypes := lo.Filter(lo.Must(c.GetInstanceTypes(ctx, np)), func(i *cloudprovider.InstanceType, _ int) bool {
 		return reqs.Compatible(i.Requirements, scheduling.AllowUndefinedWellKnownLabels) == nil &&
-			len(i.Offerings.Compatible(reqs).Available()) > 0 &&
+			len(i.Offerings.Available().Compatible(reqs)) > 0 &&
```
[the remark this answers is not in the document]
rschalo: No reason - I actually have this diff removed in the next coming commit

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -2636,26 +2631,22 @@ var _ = Describe("Consolidation", func() {
...
-						CapacityType: v1beta1.CapacityTypeOnDemand,
-						Zone:         "test-zone-1b",
+						Requirements: scheduling.NewLabelRequirements(map[string]string{v1beta1.CapacityTypeLabelKey: v1beta1.CapacityTypeOnDemand, v1.LabelTopologyZone: "test-zone-1b"}),
 						Price:        0.6,
 						Available:    true,
 					},
 					{
-						CapacityType: v1beta1.CapacityTypeSpot,
-						Zone:         "test-zone-1b",
+						Requirements: scheduling.NewLabelRequirements(map[string]string{v1beta1.CapacityTypeLabelKey: v1beta1.CapacityTypeOnDemand, v1.LabelTopologyZone: "test-zone-1c"}),
```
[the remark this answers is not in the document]
rschalo: No and also the capacity type is fixed in the next commit.

pkg/cloudprovider/types.go
```diff
@@ -257,6 +257,31 @@ func (ofs Offerings) Cheapest() Offering {
...
+		return a.Price > b.Price
+	})
+}
+
+// CapacityType returns the capacity type requirement for an offering, and panics if a capacity type is not set
+func (of Offering) CapacityType() string {
+	req, ok := of.Requirements[v1beta1.CapacityTypeLabelKey]
+	if !ok {
+		panic("failed to get capacity type for offering, capacity type must be definied on all offerings")
+	}
```
njtran, line 272: Does this need to be a panic? This should also be an invariant enforced across our code, right? Can this ever happen? 

Can we do a fatal logging instead of a panic? Maybe even just an error log?

njtran, line 272: also nit on spelling
```suggestion
	if !ok {
		panic("failed to get capacity type for offering, capacity type must be defined on all offerings")
	}
```

pkg/cloudprovider/types.go
```diff
@@ -257,6 +257,31 @@ func (ofs Offerings) Cheapest() Offering {
...
+	return req.Values()[0]
+}
+
+// Zone returns the zonal requirement for an offering, and panics if a zone is not set
+func (of Offering) Zone() string {
+	req, ok := of.Requirements[v1.LabelTopologyZone]
+	if !ok {
+		panic("failed to get zone for offering, zone must be definied on all offerings")
```
njtran, line 280: same here on panic and nit
```suggestion
		panic("failed to get zone for offering, zone must be defined on all offerings")
```

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -1223,10 +1223,11 @@ var _ = Describe("Consolidation", func() {
 			// Forcefully assign lowest possible instancePrice to make sure we have atleast one instance
 			// that is lower than the current node.
 			cloudProvider.InstanceTypes[0].Offerings[0].Price = 0.001
-			cloudProvider.InstanceTypes[0].Offerings[0].CapacityType = v1beta1.CapacityTypeSpot
+			cloudProvider.InstanceTypes[0].Offerings[0].Requirements[v1beta1.CapacityTypeLabelKey] = scheduling.NewRequirement(
+				v1beta1.CapacityTypeLabelKey, v1.NodeSelectorOpIn, v1beta1.CapacityTypeSpot)
```
njtran, line 1227: Wondering how this doesn't error if we ensure that zone and capacity type must be set in these requirements.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -1345,14 +1346,15 @@ var _ = Describe("Consolidation", func() {
 			// Forcefully assign lowest possible instancePrice to make sure we have atleast one instance
 			// that is lower than the current node.
 			cloudProvider.InstanceTypes[0].Offerings[0].Price = 0.001
-			cloudProvider.InstanceTypes[0].Offerings[0].CapacityType = v1beta1.CapacityTypeSpot
+			cloudProvider.InstanceTypes[0].Offerings[0].Requirements[v1beta1.CapacityTypeLabelKey] = scheduling.NewRequirement(
```
njtran, line 1349: same here

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -283,11 +282,6 @@ func fits(instanceType *cloudprovider.InstanceType, requests v1.ResourceList) bo
 }
 
 func hasOffering(instanceType *cloudprovider.InstanceType, requirements scheduling.Requirements) bool {
```
njtran, line 284: nit: im thinking since it's only the two lines here, it may be fine to remove this

pkg/cloudprovider/types.go
jmdeal: This should be an invariant, maybe a fatal log is the better approach. Could also make it an error, but it was a programming mistake if this isn't present, not a runtime error.

jonathan-innis: How much do we need these helper functions? Is it a lot to grab the scheduling requirement back from the offering directly?

pkg/cloudprovider/types.go
```diff
@@ -228,10 +229,10 @@ type Offering struct {
 type Offerings []Offering
 
 // Get gets the offering from an offering slice that matches the
-// passed zone and capacity type
-func (ofs Offerings) Get(ct, zone string) (Offering, bool) {
+// passed zone, capacityType, and other constraints
+func (ofs Offerings) Get(reqs scheduling.Requirements) (Offering, bool) {
```
jonathan-innis, line 233: Is this `GetCompatible()` now?

pkg/cloudprovider/types.go
```diff
@@ -245,8 +246,7 @@ func (ofs Offerings) Available() Offerings {
 // Compatible returns the offerings based on the passed requirements
 func (ofs Offerings) Compatible(reqs scheduling.Requirements) Offerings {
```
jonathan-innis, line 247: This looks the same as the Get() function now. Do we need both?

pkg/cloudprovider/types.go
```diff
@@ -215,10 +215,11 @@ func (i InstanceTypeOverhead) Total() v1.ResourceList {
 }
 
 // An Offering describes where an InstanceType is available to be used, with the expectation that its properties
-// may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type)
+// may be tightly coupled (e.g. the availability of an instance type in some zone is scoped to a capacity type) and
+// these properties are captured with labels in the Constraints map. v1beta1.CapacityTypeLabelKey and
```
jonathan-innis, line 219: I think we should update this comment now that we changed this from a constraints map over into `scheduling.Requirements`

pkg/cloudprovider/types.go
```diff
@@ -215,10 +215,11 @@ func (i InstanceTypeOverhead) Total() v1.ResourceList {
...
+// v1.LabelTopologyZone are required to be set.
 type Offering struct {
-	CapacityType string
-	Zone         string
+	Requirements scheduling.Requirements
```
jonathan-innis, line 222: Does it make sense to put the invariants that are enforced by these Offerings in terms of requirement labels over this Requirements block here?

jonathan-innis: Something like `Requirement is required to contain the keys..."

pkg/cloudprovider/types.go
jonathan-innis: Interesting enough: It seems like these functions are really only used in testing and I think there is one usage in the kwok cloudprovider. Given that they aren't used through the critical path of code, is this something that we think that we need?

pkg/controllers/provisioning/scheduling/nodeclaim.go
jonathan-innis: +1, we can probably inline this function now 🎉

pkg/cloudprovider/types.go
rschalo: This returns the first matched whereas compatible returns all matches, we could update the function names to better capture that behavior.

pkg/cloudprovider/types.go
rschalo: FWIW it is something that we use in getOverrides for creating the launch template in karpenter-provider-aws. I think this looks a lot better than getting the zero indexed value from a slice on requirements every time.

Also agree that it doesn't have to be a panic and could be an error/fatal log.

pkg/cloudprovider/types.go
jmdeal: We could also just take the result from compatible anywhere we use `Get`. The only reason we get the first result in `Get` is because we know there will only be a single offering that meets the constraints passed. There's definitely an argument to just drop `Get` in favor of using compatible.

pkg/cloudprovider/types.go
jmdeal: Still leaning toward fatal or panic myself, the main benefit IMO is enforcing the invariant rather than failing silently by returning an empty string. Since the offerings we generate are relatively static (they will always have the same requirements), this should never panic at runtime unless there's a programming error, in which case a panic will help ensure we catch it before release. Granted, since it's really only used in the provider and tests the benefit here is a little bit tenuous, but I'd rather have it than not IMO.

pkg/controllers/disruption/consolidation_test.go
jmdeal: This is just overwriting the existing capacity type, it is set by default by the fake cloudprovider

jonathan-innis (commented):
One comment, otherwise lgtm

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -245,7 +244,7 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
 		// about why scheduling failed
 		itCompat := compatible(it, requirements)
 		itFits := fits(it, requests)
-		itHasOffering := hasOffering(it, requirements)
+		_, itHasOffering := it.Offerings.Available().Get(requirements)
```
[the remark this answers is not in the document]
rschalo: Done.

jonathan-innis (approved):
/lgtm
/approve

njtran (commented):
/approve

PR #1263, opened 2024-05-17
feat: RFC Implementation Supporting AWS On-Demand Capacity Reservations

tvonhacht-apple:
Needed for RFC https://github.com/aws/karpenter-provider-aws/pull/5716
Helps RFC implementation for karpenter-provider-aws https://github.com/aws/karpenter-provider-aws/pull/6198

**Description**

RFC: https://github.com/aws/karpenter-provider-aws/pull/5716

Running in personal EKS cluster

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -294,6 +294,16 @@ func mapCandidates(proposed, current []*Candidate) []*Candidate {
 // on an instance type. If the instance type has a spot offering available, then it uses the spot offering
 // to get the launch price; else, it uses the on-demand launch price
 func worstLaunchPrice(ofs []cloudprovider.Offering, reqs scheduling.Requirements) float64 {
+	if reqs.Get(v1beta1.CapacityTypeLabelKey).Has("capacity-reservation") {
+		capacityReservationsOfferings := lo.Filter(ofs, func(of cloudprovider.Offering, _ int) bool {
+			return of.CapacityType == "capacity-reservation" && reqs.Get(v1.LabelTopologyZone).Has(of.Zone)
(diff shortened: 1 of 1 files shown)


sftim:
In the PR description, I recommend mentioning AWS On-Demand Capacity Reservations (not just ODCR).

pkg/controllers/disruption/helpers.go
```diff
@@ -294,6 +294,16 @@ func mapCandidates(proposed, current []*Candidate) []*Candidate {
 // on an instance type. If the instance type has a spot offering available, then it uses the spot offering
 // to get the launch price; else, it uses the on-demand launch price
 func worstLaunchPrice(ofs []cloudprovider.Offering, reqs scheduling.Requirements) float64 {
```
jonathan-innis, line 296: We should also think about the interactions that we have today with our current scheduling behavior and capacity reservation prioritization. Today, we are going to continually pack pods as tightly as we can on bigger and bigger instance types. This can cause us to rule out capacity reservations that would be cheaper to use if we supported a combination of spot, on-demand, and ODCR capacity. This probably works fine if we are just trying to build ODCR prioritized with on-demand/spot fallback to the same instance type sets. The second you extend the set of instance types to a combination of those that have ODCR offerings and others that don't, you start weird modeling during scheduling -- because you can overpack pods, causing us to rule out ODCR capacity and only leave behind spot and on-demand. 

Considering the capacity is free -- we could choose to build affordances so that we always choose a free offering if it's available and don't continue to pack pods if it would cause us to work outside of this free ODCR offering e.g. create a new node instead of creating a bigger instance type.

Something to think about as we are considering the design options here.

PR #1277, opened 2024-05-22
feat: add kwoknodeclass

njtran:
Fixes #N/A 

**Description**
This adds a dummy nodeclass for the kwok provider. This is essential when NodePools rely on a nodeclass, Karpenter must be able to discover the supported nodeclass and get its ready condition. For Kwok, we assume the ready condition is defaulted to true.

testing here


kwok/apis/v1alpha1/kwoknodeclass_status.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package v1alpha1
+
+import (
+	"github.com/awslabs/operatorpkg/status"
+)
+
+// KwokNodeClassStatus contains the resolved state of the KwokNodeClass
+type KwokNodeClassStatus struct {
```
Bryce-Soghigian, line 24: nit: should this be `KWOK` since it's an acronym?

njtran: That makes sense, but for some reason KwokNodeClass looks better to me from a camelcased perspective. From a titling perspective, you could think it should be acronym'd like KwoK, which is how I've sometimes referenced it, since `wo` is an "unimportant" word. WDYT?

kwok/README.md
```diff
@@ -38,12 +38,17 @@ spec:
           operator: In
           values: ["spot"]
       nodeClassRef:
-        name: nil
+        name: default
```
jonathan-innis, line 41: We're going to be wanting this to have all the detail at v1. Better to just add it in there now so that we are adhering to best practices?

kwok/apis/v1alpha1/kwoknodeclass.go
```diff
@@ -0,0 +1,47 @@
...
+
+// KwokNodeClass is the Schema for the KwokNodeClass API
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=kwoknodeclasses,scope=Cluster,categories=karpenter,shortName={kwoknc,kwokncs}
+// +kubebuilder:subresource:status
+type KwokNodeClass struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec KwokNodeClassSpec `json:"spec,omitempty"`
```
jonathan-innis, line 36: Do we even need a spec at all if there's nothing in it or does controller-runtime complain if you try to do that?

kwok/apis/v1alpha1/kwoknodeclass.go
```diff
@@ -0,0 +1,47 @@
...
+	// +kubebuilder:default:={conditions: {{type: "Ready", status: "True", reason:"NodeClassReady", lastTransitionTime: "2024-01-01T01:01:01Z", message: ""}}}
```
jonathan-innis, line 37: This feels strange to me. We do this because we don't want to write a controller to actually mark this as ready?

jonathan-innis, line 37: IIRC, the reason matches the type on a MarkTrue

kwok/apis/v1alpha1/kwoknodeclass_status.go
jonathan-innis: I think we generally want to just capitalize acronyms everywhere as a style thing if we can. See: https://go.dev/wiki/CodeReviewComments#initialisms

kwok/apis/v1alpha1/register.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+const (
+	Group              = "kwok.karpenter.sh"
+	CompatabilityGroup = "compatibility." + Group
```
jonathan-innis, line 27: Do we need a compatibility group? This was used for migration for Karpenter but probably isn't relevant to KWOK?

kwok/cloudprovider/cloudprovider.go
```diff
@@ -119,7 +120,13 @@ func (c CloudProvider) Name() string {
 }
 
 func (c CloudProvider) GetSupportedNodeClasses() []schema.GroupVersionKind {
-	return []schema.GroupVersionKind{}
+	return []schema.GroupVersionKind{
+		{
+			Group:   v1alpha1.SchemeGroupVersion.Group,
+			Version: v1alpha1.SchemeGroupVersion.Version,
+			Kind:    "KwokNodeClass",
```
jonathan-innis, line 127:
```suggestion
			Kind:    "KWOKNodeClass",
```
Given that we do `EC2NodeClass` and not `Ec2NodeClass`

jonathan-innis (commented):
Only real comment is on the capitalization of KWOK

kwok/apis/v1alpha1/kwoknodeclass.go
njtran: Exactly. What's the point of writing a fake controller here? I think it's fine to default the condition since we're treating this as a dummy nodeclass anyway.

kwok/apis/v1alpha1/kwoknodeclass_status.go
njtran: okkkkkkkkkk

kwok/apis/v1alpha1/register.go
njtran: Good call, didn't think too deeply about it.

kwok/apis/v1alpha1/kwoknodeclass.go
njtran: Good call, removed it

jonathan-innis (approved):
/lgtm
/approve

Issue #1133, opened 2024-03-25
karpenter evicts DaemonSet pods before all service pods are terminated


robertoriv:
### Description

**Observed Behavior**:

When draining a node, karpenter will evict all service pods (managed by Deployments) and then evict the DaemonSet pods running on the node. However, it starts evicting DaemonSet pods before the service pods are entirely gone. 

We use a `preStopHook` to flush metrics and/or drain the pods out of the ALBs. The observed behavior will cause a pod to run in `Terminating` status for a while, without the DaemonSets required to support it -- in our case: metrics, logging. This results in loss of metrics and logs.

Example:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: my-service
  namespace: my-service
  labels:
    app: my-service
spec:
  template:
    metadata:
      labels:
        app: my-service
    spec:
      containers:
      - name: my-service
        image: ...
        command: ...
        env: ...

        lifecycle:
            preStop:
              exec:
                command: ["/bin/sh","-c","curl -X GET \"$POD_IP:3000/flush?delay=120\" > /proc/1/fd/1"]
      terminationGracePeriodSeconds: 135
```

The events look like this:
```
❯ k get events -A | grep -E "my-service-d977cd6fd-w494k|datadog-agent-jv5nz|vector-logging-246l7"
my-service                34s         Normal    Evicted                     pod/my-service-d977cd6fd-w494k                                  Evicted pod
my-service                34s         Normal    Killing                     pod/my-service-d977cd6fd-w494k                                  Stopping container my-service
monitoring                33s         Normal    Evicted                     pod/datadog-agent-jv5nz                                           Evicted pod
monitoring                33s         Normal    Killing                     pod/datadog-agent-jv5nz                                           Stopping container agent
monitoring                33s         Normal    Evicted                     pod/vector-logging-246l7                                          Evicted pod
monitoring                33s         Normal    Killing                     pod/vector-logging-246l7                                          Stopping container vector
```
(There's a 1s gap between the "request" to evict `my-service`, and the request to evict `datadog-agent` and `vector-logging`. `my-service` will delay termination by `120s`, and we would expect datadog and vector to hand around until then.)

The `my-service` pod is still `Terminating` 97s after the replacement pod was scheduled:
```
❯ k get pods -n my-service
NAME                           READY   STATUS        RESTARTS   AGE
my-service-d977cd6fd-t8vn8   1/1     Running       0          97s          # replacement pod
[...]

PR #1280, opened 2024-05-23
feat: delay draining of higher order pods while pods are still terminating

robertoriv:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1133
Fixes https://github.com/aws/karpenter-provider-aws/issues/5858

**Description**

`preStop` hooks are widely used to allow for the [graceful termination of pods](https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-terminating-with-grace). Karpenter currently runs through the list of pods in "priority groups", but moves on to the next group once all evictions in the current group are acknowledged, without waiting for the pods to actually terminate. This can lead to DaemonSets -- like metrics or logging -- terminating before the pods they are serving have had a chance to do so.

To address this issue, this PR introduces a check in the `Drain` function to identify terminating pods. When such pods are found, Karpenter will queue only the pods belonging to the same or a lower group, ensuring the desired eviction order is maintained. When no terminating pods are found, the queueing continues as normal.

Test files: https://gist.github.com/robertoriv/7537e77844d12154437dee0116f79603

I tried using `kwok` for testing, but it doesn't seem to execute the `preStop` hooks properly. Terminating a node is nearly instantaneous when using a `kind` cluster.

I then tried building the controller locally, and deploying it to a live cluster. That worked well enough to test the queueing behavior, but the locally-built controller -- modified or unmodified -- was unable to create new instances to satisfy the node claims due to a bunch of `"message":"waiting on cluster sync"` messages and `launching nodeclaim, resolving ec2nodeclass, object is awaiting reconciliation` erros. I'm guessing that's a problem with my setup.

Here are the logs for the feature, printed out using a more verbose version of the `Drain` function that is included in the gist:
```
karpenter-58ffc9fcf8-sl7q9 {... "message":"Started with 18 evictable pods.","commit":"d39fdb5-dirty","controller":"node.termination", ...}
karpenter-58ffc9fcf8-sl7q9 {... "message":"Found 3 terminating pods: ","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-gfx9t, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-qsh4b, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-src47, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"Highest order pod: non-critical-pod-5f4f67fd94-gfx9t, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ...}
karpenter-58ffc9fcf8-sl7q9 {... "message":"Narrowed the list of evictablePods to be queued to 4 pods.","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-5hjfr, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-8j5hm, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-d78pq, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination", ... }
karpenter-58ffc9fcf8-sl7q9 {... "message":"  - Pod: non-critical-pod-5f4f67fd94-rtnqz, Group: 1","commit":"d39fdb5-dirty","controller":"node.termination",  ... }
```

### DIFF
--- a/pkg/controllers/node/termination/terminator/terminator.go
+++ b/pkg/controllers/node/termination/terminator/terminator.go
@@ -84,8 +84,24 @@ func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
 	if err != nil {
 		return fmt.Errorf("listing pods on node, %w", err)
 	}
+
 	// evictablePods are pods that aren't yet terminating are eligible to have the eviction API called against them
 	evictablePods := lo.Filter(pods, func(p *v1.Pod, _ int) bool { return podutil.IsEvictable(p) })
+
+	// check if there are any pods that are still terminating and haven't exceeded their termination grace period
+	// if there are any, only queue pods from the same eviction group or lower
+	terminatingPods := lo.Filter(pods, func(p *v1.Pod, _ int) bool {
+		return podutil.IsTerminating(p) && !podutil.IsStuckTerminating(p, t.clock)
+	})
+	if len(terminatingPods) > 0 {
+		highestOrderPod := lo.MaxBy(terminatingPods, func(p *v1.Pod, max *v1.Pod) bool {
+			return podutil.GetPodEvictionGroup(p) > podutil.GetPodEvictionGroup(max)
+		})
+		evictablePods = lo.Filter(evictablePods, func(p *v1.Pod, _ int) bool {
+			return podutil.GetPodEvictionGroup(p) <= podutil.GetPodEvictionGroup(highestOrderPod)
+		})
+	}
+
 	t.Evict(evictablePods)
 
 	// podsWaitingEvictionCount are  the number of pods that either haven't had eviction called against them yet
@@ -98,28 +114,26 @@ func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
 }
 
 func (t *Terminator) Evict(pods []*v1.Pod) {
-	// 1. Prioritize noncritical pods, non-daemon pods https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
+	// 1. Group pods based on criticality and daemonset ownership
 	var criticalNonDaemon, criticalDaemon, nonCriticalNonDaemon, nonCriticalDaemon []*v1.Pod
 	for _, pod := range pods {
-		if pod.Spec.PriorityClassName == "system-cluster-critical" || pod.Spec.PriorityClassName == "system-node-critical" {
-			if podutil.IsOwnedByDaemonSet(pod) {
-				criticalDaemon = append(criticalDaemon, pod)
-			} else {
-				criticalNonDaemon = append(criticalNonDaemon, pod)
-			}
-		} else {
-			if podutil.IsOwnedByDaemonSet(pod) {
-				nonCriticalDaemon = append(nonCriticalDaemon, pod)
-			} else {
-				nonCriticalNonDaemon = append(nonCriticalNonDaemon, pod)
-			}
+		switch podutil.GetPodEvictionGroup(pod) {
+		case podutil.NonCriticalNonDaemonsets:
+			nonCriticalNonDaemon = append(nonCriticalNonDaemon, pod)
+		case podutil.NonCriticalDaemonsets:
+			nonCriticalDaemon = append(nonCriticalDaemon, pod)
+		case podutil.CriticalNonDaemonsets:
+			criticalNonDaemon = append(criticalNonDaemon, pod)
+		case podutil.CriticalDaemonsets:
+			criticalDaemon = append(criticalDaemon, pod)
 		}
 	}
-	// 2. Evict in order:
-	// a. non-critical non-daemonsets
-	// b. non-critical daemonsets
-	// c. critical non-daemonsets
-	// d. critical daemonsets
+
+	// 2. Evict in order: https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/
+	//   a. non-critical non-daemonsets
+	//   b. non-critical daemonsets
+	//   c. critical non-daemonsets
+	//   d. critical daemonsets
 	if len(nonCriticalNonDaemon) != 0 {
 		t.evictionQueue.Add(nonCriticalNonDaemon...)
 	} else if len(nonCriticalDaemon) != 0 {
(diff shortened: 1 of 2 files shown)


pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -98,28 +114,26 @@ func (t *Terminator) Drain(ctx context.Context, node *v1.Node) error {
 }
 
 func (t *Terminator) Evict(pods []*v1.Pod) {
```
jonathan-innis, line 116: What about just updating our Evict() function to pull in all pods waiting termination, grouping them, and then checking if the group is empty. If the group isn't empty, then we'd just pass the pods that haven't started termination yet to the eviction queue. 

I think realistically, this allows us to drop the check on `podsWaitingEvictionCount > 0` and should reduce the amount of change required to get this functionality

robertoriv: @jonathan-innis thanks for taking a look at this. Happy to work on some improvements.

I could use some clarification on a few points.

>What about just updating our Evict() function to pull in all pods waiting termination, grouping them, and then checking if the group is empty. If the group isn't empty, then we'd just pass the pods that haven't started termination yet to the eviction queue.

That would break with the current behavior where `Drain(...)` selects the pods that need to be evicted, and `Evict(...)` groups them and queues them up. (Not necessarily a bad thing, but I was trying to color within the existing lines when I wrote this code.)

In order to pull/query for not-yet-evicted pods in `Evict(...)`, we would need to call `nodeutil.GetPods(ctx, t.kubeClient, node)`. That would require a function signature change (`Evict(ctx context.Context, node *v1.Node)`) and would create a slight misalignment with how the k8s API works. We'd be calling for an eviction on a node, instead of pods.. so to speak. Is that something we'd be ok with?

>I think realistically, this allows us to drop the check on podsWaitingEvictionCount > 0 and should reduce the amount of change required to get this functionality

Is that check not required? I'm not particularly familiar with the `finalize` function that calls `Drain`, but it looks like the error returned is used to halt progress on the termination: https://github.com/kubernetes-sigs/karpenter/blob/f1a2d42b5051585503035aa0883f4cb2cd8cf9d6/pkg/controllers/node/termination/controller.go#L90-L93

jonathan-innis: > That would break with the current behavior where `Drain(...)` selects the pods that need to be evicted, and `Evict(...)` groups them and queues them up. (Not necessarily a bad thing, but I was trying to color within the existing lines when I wrote this code.)

Totally, though I think this is a reasonable change to make, given what we are trying to achieve here. If anything, I think this allows you to actually just remove the `Evict` function entirely and just collapse all of the logic into the `Drain()` function. Inside of the drain, you can just try evicting a set of pods until all of the pods are fully terminated. The way I had thought about doing it was something like this

```
	pods, err := nodeutil.GetPods(ctx, t.kubeClient, node)
	if err != nil {
		return fmt.Errorf("listing pods on node, %w", err)
	}
        podGroups := t.groupPodsByPriority(pods)
        waitingOnEviction := false
        for _, group := range podGroups {
           if len(group) > 0 {
               t.evictionQueue.Add(lo.Filter(pods, func(p *v1.Pod, _ int) bool { return podutil.IsEvictable(p) }))
               waitingOnEviction = true
           }
       }
      if waitingOnEviction {
        return NewNodeDrainError(fmt.Errorf("%d pods are waiting to be evicted", len(pods)))
     }
}
        

func (t *Terminator) groupPodsByPriority(pods []*v1.Pod) [][]*v1.Pod {
  // 1. Prioritize noncritical pods, non-daemon pods https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown
	var criticalNonDaemon, criticalDaemon, nonCriticalNonDaemon, nonCriticalDaemon []*v1.Pod
	for _, pod := range pods {
		if pod.Spec.PriorityClassName == "system-cluster-critical" || pod.Spec.PriorityClassName == "system-node-critical" {
			if podutil.IsOwnedByDaemonSet(pod) {
				criticalDaemon = append(criticalDaemon, pod)
			} else {
				criticalNonDaemon = append(criticalNonDaemon, pod)
			}
		} else {
			if podutil.IsOwnedByDaemonSet(pod) {
				nonCriticalDaemon = append(nonCriticalDaemon, pod)
			} else {
				nonCriticalNonDaemon = append(nonCriticalNonDaemon, pod)
			}
		}
	}
	return [][]*v1.Pod{criticalNonDaemon, criticalDaemon, nonCriticalNonDaemon, nonCriticalDaemon}
}
```

jonathan-innis:
@robertoriv Just noticed that it's been a few weeks and there's been no movement on this PR. Happy to help in any way that we can. Feel free to ping me or any of the maintainers on slack.

jonathan-innis:
@robertoriv Do you expect to come back to this one? If not, we may take in a similar change to the one that's proposed above (see https://github.com/kubernetes-sigs/karpenter/pull/1280#discussion_r1661796076)

robertoriv:
> @robertoriv Do you expect to come back to this one? If not, we may take in a similar change to the one that's proposed above (see [#1280 (comment)](https://github.com/kubernetes-sigs/karpenter/pull/1280#discussion_r1661796076))

Hi @jonathan-innis 👋🏻

I’m currently on leave and won’t be able to get back to this for a couple of months. Feel free to close this PR and/or adopt any pieces that are useful from it.

PR #1285, opened 2024-05-28
fix: allow empty string for condition reason

jmdeal:
Fixes #N/A 

**Description**
Allows empty strings for status condition reasons in the nodeclaims. This is a gap in the #1254 where we still fail validation when the reason is set to an empty string.

Manual validation: installed v0.36.2, spun up nodes, upgraded to HEAD, expired existing nodes


njtran (commented):
/lgtm
/approve

PR #1289, opened 2024-06-03
chore: Convert singleton reconcilers to operatorpkg

jonathan-innis:
Fixes #N/A 

**Description**

This updates our singleton controllers to use the singleton constructs in the `operatorpkg` package. This enables us to drop the implementation of the singleton builder that we currently have in Karpenter

`make presubmit`
`/karpenter snapshot` in the AWS provider repo


pkg/operator/controller/controller.go
```diff
@@ -41,157 +30,3 @@ const (
...
-		name:        name,
-		rateLimiter: workqueue.DefaultItemBasedRateLimiter(),
-	}
-	s.initMetrics()
-	return s
-}
-
-// initMetrics is effectively the same metrics initialization function used by controller-runtime
-// https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/internal/controller/controller.go
-func (s *Singleton) initMetrics() {
```
jmdeal, line 83: I assume all of these metrics are injected since we're now using the controller-runtime controller directly?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -103,23 +107,26 @@ func (p *Provisioner) Trigger() {
...
-		return reconcile.Result{}, nil
+		return reconcile.Result{RequeueAfter: controller.Immediately}, nil
 	}
 	// We need to ensure that our internal cluster state mechanism is synced before we proceed
 	// with making any scheduling decision off of our state nodes. Otherwise, we have the potential to make
 	// a scheduling decision based on a smaller subset of nodes in our cluster state than actually exist.
 	if !p.cluster.Synced(ctx) {
 		log.FromContext(ctx).V(1).Info("waiting on cluster sync")
-		return reconcile.Result{RequeueAfter: time.Second}, nil
+		return reconcile.Result{RequeueAfter: controller.Immediately}, nil
```
jmdeal, line 129: Any reason we're bumping this from a 1 second delay to Immediately?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -103,23 +107,26 @@ func (p *Provisioner) Trigger() {
...
 }
 
-func (p *Provisioner) Reconcile(ctx context.Context, _ reconcile.Request) (result reconcile.Result, err error) {
+func (p *Provisioner) Reconcile(ctx context.Context) (result reconcile.Result, err error) {
+	ctx = injection.WithControllerName(ctx, "provisioner")
+
 	// Batch pods
 	if triggered := p.batcher.Wait(ctx); !triggered {
-		return reconcile.Result{}, nil
+		return reconcile.Result{RequeueAfter: controller.Immediately}, nil
```
jmdeal, line 122: Double-checking, but this should be the same behavior as what we had before, right? Our singleton controller would requeue immediately on an empty result, whereas controller runtime will not.

pkg/operator/controller/controller.go
jonathan-innis: Yep, these are added for free by controller-runtime. I validated this through checking prometheus after the change,

pkg/controllers/provisioning/provisioner.go
jonathan-innis: Correct.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -154,12 +158,15 @@ func NewCommand(replacements []string, candidates []*state.StateNode, id types.U
...
+	return controllerruntime.NewControllerManagedBy(m).
 		Named("disruption.queue").
-		Complete(q)
+		WatchesRawSource(singleton.Source()).
+		Complete(singleton.AsReconciler(q))
 }
 
-func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "disruption.queue")
```
njtran, line 168: should we rename the folder? Or match the injection to the package? 
```suggestion
	ctx = injection.WithControllerName(ctx, "orchestration.queue")
```

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -123,7 +128,9 @@ func (q *Queue) Has(pod *v1.Pod) bool {
 	return q.set.Has(NewQueueKey(pod))
 }
 
-func (q *Queue) Reconcile(ctx context.Context, _ reconcile.Request) (reconcile.Result, error) {
+func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "eviction-queue")
```
njtran, line 132: Can we align on if we're doing kabob case or doing `.` across all controllers?

pkg/operator/controller/controller.go
```diff
@@ -41,157 +30,3 @@ const (
 type Controller interface {
 	Register(context.Context, manager.Manager) error
 }
```
njtran, line 32: I'm wondering why this can't be done in operatorpkg or why we still need it? I'd imagine this whole file gets removed.

pkg/operator/controller/controller.go
```diff
@@ -41,157 +30,3 @@ const (
...
-	}, []string{"controller"})
-	workerCount = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "controller_runtime_max_concurrent_reconciles",
-		Help: "Maximum number of concurrent reconciles per controller",
-	}, []string{"controller"})
-	activeWorkers = prometheus.NewGaugeVec(prometheus.GaugeOpts{
-		Name: "controller_runtime_active_workers",
-		Help: "Number of currently used workers per controller",
-	}, []string{"controller"})
-)
```
njtran, line 180: Can we make sure that these metrics are covered by operatorpkg? Since we're relying on operatorpkg to give us these metrics now, I'm worried about unexpectedly introducing metrics changes to users. thoughts?

pkg/test/expectations/expectations.go
```diff
@@ -168,15 +169,27 @@ func ExpectDeleted(ctx context.Context, c client.Client, objects ...client.Objec
 	}
 }
 
-// TODO: Consider removing converting this into a standard reconciler; however, objects have to be properly updated if we don't rely on the Rconcile() method to get the object for us
+func ExpectSingletonReconciled(ctx context.Context, reconciler singleton.Reconciler) reconcile.Result {
```
njtran, line 172: Can't this be done for us in operatorpkg/

jonathan-innis: Yes and no. We're going to have to take a separate PR to migrate over all of these functions into operatorpkg. As it is, that's going to introduce a lot of change into the PR and we can't only do half of the changes at once because you're not allowed to define the same functions from two different packages when you are using "dot" importing

pkg/operator/controller/controller.go
jonathan-innis: This was validated, yes. Operatorpkg isn't providing these metrics; it's controller-runtime that's doing it. We had to add these before because we were creating our own form of a "runnable" that wasn't using the standard controller framework with a `NewControllerManagedBy` call. Now that all of our controllers are making calls to this same function and are using the controller framework, there's no longer any need to statically define these metrics.

pkg/operator/controller/controller.go
jonathan-innis: It's a good question. I'll follow-up with @ellistarn offline to validate why this interface doesn't exist.

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: I'd agree we should align on this and ideally have this match the file directory structure; however, this isn't a change that's being introduced in this PR (it's only maintaining the current functionality), so I'd like to decouple things so that we make that change separately

pkg/controllers/disruption/orchestration/queue.go
jonathan-innis: Same comment here. I think we should decouple this into a separate PR so that we aren't making additional changes in the context of this refactor.

pkg/controllers/node/termination/terminator/eviction.go
njtran: Fine with this, just wondering what the reasonable difference is here? Is it that logging would be slightly different for those who are looking at the eviction queue? Seems like maybe it's just this file, which we could just change it to be . instead of -.

pkg/operator/controller/controller.go
njtran: Just so I understand, the set of metrics in operatorpkg will use the same controller-runtime metrics, and potentially add more of its own. Is that right?

jonathan-innis: Just the controller-runtime ones. Doesn't add any more.

pkg/operator/controller/controller.go
jonathan-innis: Closed, we've now added this to operatorpkg: https://github.com/awslabs/operatorpkg/pull/51

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: > Seems like maybe it's just this file, which we could just change it to be . instead of -.

We can decouple and follow-up on this in a follow-up PR IMO

pkg/operator/controller/controller.go
njtran, line 1: fully deleted!!!

njtran (commented):
/lgtm
/approve

PR #916, opened 2023-12-27
feat: disruption.terminationGracePeriod

wmgroot:
Fixes #743 

**Description**
This is a PR to add support for an equivalent to CAPI's `nodeDrainTimeout` feature for Karpenter.
I've taken the feedback provided in https://github.com/kubernetes-sigs/karpenter/pull/834.

This implementation relies on K8s 1.26 for default support of the non-graceful node shutdown Taint.
https://kubernetes.io/docs/concepts/architecture/nodes/#non-graceful-node-shutdown

Additionally, this PR implements the behavior described in https://github.com/jmdeal/karpenter/blob/disruption-grace-design/designs/termination-grace-period-extension.md. This allows the feature to be a comprehensive solution that addresses issues with disruption logic that previously prevented cluster administrators from enforcing drain timeout behavior for eventual disruption cases (such as node expiration and drift, but not consolidation).

Pods are also deleted preemptively at time T = node.expirationTime - pod.terminationGracePeriodSeconds. This attempts to ensure that a pod always has its full terminationGracePeriodSeconds worth of time to terminate gracefully before the node expires.

**How was this change tested?**
The PR includes unit tests for all added logic.

I've also tested this by building a custom Karpenter image and running it in our live clusters with various test cases by deleting nodes and nodeclaims directly to verify pods are are successfully deleted and the node is reaped when it reaches the desired expiration time.
1. Pods at their PDB limit.
2. Pods still running a prestop hook.
3. Pods with do-not-disrupt annotations.

### Example
Initial Resources
1. A deployment with a strict PDB (maxUnavailable: 1)
4. A preStop hook sleep of 120 seconds to delay pod termination.
5. A pod terminationGracePod of 300 seconds. Pods should be deleted on the node 5m after node termination starts (10m - 5m).
6. A startupProbe delay of 180s to delay pod readiness.
```
$ kubectl get nodepool qa -o yaml
apiVersion: karpenter.sh/v1beta1
kind: NodePool
spec:
  template:
    terminationGracePeriod: 10m

$ kubectl get deploy -n default        
NAME                READY   UP-TO-DATE   AVAILABLE   AGE
hello-world-nginx   3/4     4            3           70d

$ kubectl get pdb -n default                                     
NAME                MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
hello-world-nginx   N/A             1                 1                     64d

$ kubectl get pod -n default -o yaml hello-world-nginx-6965d88b66-7ks7t
...
[...]

PR #1294, opened 2024-06-04
feat: Ensure persistent volumes are detached before deleting node

AndrewSirenko:
Fixes #N/A 

**Description**

Fixes 6+ minute delays for disrupted EBS-Backed stateful workloads when starting on their new node. 

For more context see [RFC for solving 6+ minute delays for disrupted stateful workloads](https://github.com/aws/karpenter-provider-aws/pull/6336)

TLDR:

In order for a stateful pod to smoothly migrate from terminating node to new node...

0. Consolidation event starts
1. Stateful pods must terminate
2. EBS CSI Node pod must unmount all filesystems (NodeUnpublish & NodeUnstage RPCs)
3. EBS CSI Controller pod must detach all volumes from instance
4. Karpenter terminates EC2 Instance
6. Karpenter ensures Node object deleted from Kubernetes

Problems:
A. If 2 doesn't happen, today there's a 6+ minute delay in stateful pod migration because Kubernetes is afraid volume still attached and mounted to instance (6+ min delay)
B. If 3 doesn't happen, the new stateful pod can't start until consolidated instance is terminated which auto-detaches volumes (1+ min delay)

Solution:
Wait for volumeattachment objects associated with drainable pods & non-multi-attach volumes before deleting the node. 

Manual: Create statefulset + nodepool. Have nodes expire every 3 minutes. Check that stateful pods migrate to new node and start running in under a minute. 

Also tested that we do not block deletion when there are stateful workloads that tolerate all taints, or Node terminationGracePeriod elapsed. 

**Additional Notes**

Note 1: Must add read permissions for volumeattachments to clusterrole-core.yaml in `karpenter-provider-aws`

```
  - apiGroups: ["storage.k8s.io"]
    resources: ["storageclasses", "csinodes", "volumeattachments"]
    verbs: ["get", "watch", "list"]
```

Note 2: Separate PR to add e2e tests in karpenter-provider-aws: https://github.com/aws/karpenter-provider-aws/pull/6484

Note 3: It was decided in the RFC we will block node deletion for all volumeattachments, regardless of CSI Driver. In the future, we may decide to inject a list of CSI Drivers via the cloud provider instead. 

Note 4: There might be some rare cases where EBS CSI Node pod can get killed before it unmounts volumes. The solution would be karpenter (or some reliable automation) tainting node with `nodeshutdown:NoExecute` once node is terminated, as discussed in RFC. In design meeting consensus was that this could be added later if customers run into it. 

### DIFF
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -21,15 +21,16 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/api/errors"
-
-	"sigs.k8s.io/karpenter/pkg/utils/termination"
-
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/samber/lo"
 	"golang.org/x/time/rate"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/utils/clock"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
@@ -38,19 +39,22 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
-	"sigs.k8s.io/karpenter/pkg/operator/injection"
-
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator"
 	terminatorevents "sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator/events"
 	"sigs.k8s.io/karpenter/pkg/events"
 	"sigs.k8s.io/karpenter/pkg/metrics"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
 	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
+	"sigs.k8s.io/karpenter/pkg/utils/pod"
+	"sigs.k8s.io/karpenter/pkg/utils/termination"
+	volumeutil "sigs.k8s.io/karpenter/pkg/utils/volume"
 )
 
 // Controller for the resource
 type Controller struct {
+	clock         clock.Clock
 	kubeClient    client.Client
 	cloudProvider cloudprovider.CloudProvider
 	terminator    *terminator.Terminator
@@ -58,8 +62,9 @@ type Controller struct {
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, terminator *terminator.Terminator, recorder events.Recorder) *Controller {
+func NewController(clk clock.Clock, kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, terminator *terminator.Terminator, recorder events.Recorder) *Controller {
 	return &Controller{
+		clock:         clk,
 		kubeClient:    kubeClient,
 		cloudProvider: cloudProvider,
 		terminator:    terminator,
@@ -119,6 +124,18 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+	// In order for Pods associated with PersistentVolumes to smoothly migrate from the terminating Node, we wait
+	// for VolumeAttachments of drain-able Pods to be cleaned up before terminating Node and removing its finalizer.
+	// However, if TerminationGracePeriod is configured for Node, and we are past that period, we will skip waiting.
+	if nodeTerminationTime == nil || c.clock.Now().Before(*nodeTerminationTime) {
+		areVolumesDetached, err := c.ensureVolumesDetached(ctx, node)
+		if err != nil {
+			return reconcile.Result{}, fmt.Errorf("ensuring no volume attachments, %w", err)
+		}
+		if !areVolumesDetached {
+			return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+	}
 	nodeClaims, err = nodeutils.GetNodeClaims(ctx, node, c.kubeClient)
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("deleting nodeclaims, %w", err)
@@ -156,6 +173,58 @@ func (c *Controller) deleteAllNodeClaims(ctx context.Context, nodeClaims ...*v1.
 		}
 	}
 	return nil
+}
+
+func (c *Controller) ensureVolumesDetached(ctx context.Context, node *corev1.Node) (volumesDetached bool, err error) {
+	volumeAttachments, err := nodeutils.GetVolumeAttachments(ctx, c.kubeClient, node)
+	if err != nil {
+		return false, err
+	}
+	// Filter out VolumeAttachments associated with not drain-able Pods
+	filteredVolumeAttachments, err := filterVolumeAttachments(ctx, c.kubeClient, node, volumeAttachments, c.clock)
+	if err != nil {
+		return false, err
+	}
+	return len(filteredVolumeAttachments) == 0, nil
+}
+
+// filterVolumeAttachments filters out storagev1.VolumeAttachments that should not block the termination
+// of the passed corev1.Node
+func filterVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node, volumeAttachments []*storagev1.VolumeAttachment, clk clock.Clock) ([]*storagev1.VolumeAttachment, error) {
+	// No need to filter empty VolumeAttachments list
+	if len(volumeAttachments) == 0 {
+		return volumeAttachments, nil
+	}
+	// Create list of non-drain-able Pods associated with Node
+	pods, err := nodeutils.GetPods(ctx, kubeClient, node)
+	if err != nil {
+		return nil, err
+	}
+	unDrainablePods := lo.Reject(pods, func(p *corev1.Pod, _ int) bool {
+		return pod.IsDrainable(p, clk)
+	})
+	// Filter out VolumeAttachments associated with non-drain-able Pods
+	// Match on Pod -> PersistentVolumeClaim -> PersistentVolume Name <- VolumeAttachment
+	shouldFilterOutVolume := sets.New[string]()
+	for _, p := range unDrainablePods {
+		for _, v := range p.Spec.Volumes {
+			pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, p, v)
+			if errors.IsNotFound(err) {
+				continue
+			}
+			if err != nil {
+				return nil, err
+			}
+			if pvc != nil {
+				shouldFilterOutVolume.Insert(pvc.Spec.VolumeName)
+			}
+		}
+	}
+	filteredVolumeAttachments := lo.Reject(volumeAttachments, func(v *storagev1.VolumeAttachment, _ int) bool {
+		pvName := v.Spec.Source.PersistentVolumeName
+		return pvName == nil || shouldFilterOutVolume.Has(*pvName)
+	})
+	return filteredVolumeAttachments, nil
 }
 
 func (c *Controller) removeFinalizer(ctx context.Context, n *corev1.Node) error {
--- a/pkg/utils/node/node.go
+++ b/pkg/utils/node/node.go
@@ -148,11 +148,9 @@ func GetProvisionablePods(ctx context.Context, kubeClient client.Client) ([]*cor
 func GetVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node) ([]*storagev1.VolumeAttachment, error) {
 	var volumeAttachmentList storagev1.VolumeAttachmentList
 	if err := kubeClient.List(ctx, &volumeAttachmentList, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
-		return nil, fmt.Errorf("listing volumeattachments, %w", err)
+		return nil, fmt.Errorf("listing volumeAttachments, %w", err)
 	}
-	return lo.FilterMap(volumeAttachmentList.Items, func(v storagev1.VolumeAttachment, _ int) (*storagev1.VolumeAttachment, bool) {
-		return &v, v.Spec.NodeName == node.Name
-	}), nil
+	return lo.ToSlicePtr(volumeAttachmentList.Items), nil
 }
 
 func GetCondition(n *corev1.Node, match corev1.NodeConditionType) corev1.NodeCondition {
--- a/pkg/utils/node/node.go
+++ b/pkg/utils/node/node.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 
@@ -140,6 +141,17 @@ func GetProvisionablePods(ctx context.Context, kubeClient client.Client) ([]*cor
 	}
 	return lo.FilterMap(podList.Items, func(p corev1.Pod, _ int) (*corev1.Pod, bool) {
 		return &p, pod.IsProvisionable(&p)
+	}), nil
+}
+
+// GetVolumeAttachments grabs all volumeAttachments associated with the passed node
+func GetVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node) ([]*storagev1.VolumeAttachment, error) {
+	var volumeAttachmentList storagev1.VolumeAttachmentList
+	if err := kubeClient.List(ctx, &volumeAttachmentList, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+		return nil, fmt.Errorf("listing volumeattachments, %w", err)
+	}
+	return lo.FilterMap(volumeAttachmentList.Items, func(v storagev1.VolumeAttachment, _ int) (*storagev1.VolumeAttachment, bool) {
+		return &v, v.Spec.NodeName == node.Name
 	}), nil
 }
 
--- a/pkg/utils/pod/scheduling.go
+++ b/pkg/utils/pod/scheduling.go
@@ -62,14 +62,20 @@ func IsEvictable(pod *corev1.Pod) bool {
 
 // IsWaitingEviction checks if this is a pod that we are waiting to be removed from the node by ensuring that the pod:
 // - Isn't a terminal pod (Failed or Succeeded)
-// - Isn't a pod that has been terminating past its terminationGracePeriodSeconds
-// - Doesn't tolerate the "karpenter.sh/disruption=disrupting" taint
-// - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
-// Note: pods with the `karpenter.sh/do-not-disrupt` annotation are included since node drain should stall until these pods are evicted or become terminal, even though Karpenter won't orchestrate the eviction.
+// - Can be drained by Karpenter (See IsDrainable)
 func IsWaitingEviction(pod *corev1.Pod, clk clock.Clock) bool {
 	return !IsTerminal(pod) &&
+		IsDrainable(pod, clk)
+}
+
+// IsDrainable checks if a pod can be drained by Karpenter by ensuring that the pod:
+// - Doesn't tolerate the "karpenter.sh/disruption=disrupting" taint
+// - Isn't a pod that has been terminating past its terminationGracePeriodSeconds
+// - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
+// Note: pods with the `karpenter.sh/do-not-disrupt` annotation are included since node drain should stall until these pods are evicted or become terminal, even though Karpenter won't orchestrate the eviction.
+func IsDrainable(pod *corev1.Pod, clk clock.Clock) bool {
+	return !ToleratesDisruptedNoScheduleTaint(pod) &&
 		!IsStuckTerminating(pod, clk) &&
-		!ToleratesDisruptedNoScheduleTaint(pod) &&
 		// Mirror pods cannot be deleted through the API server since they are created and managed by kubelet
 		// This means they are effectively read-only and can't be controlled by API server calls
 		// https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#drain
--- a/kwok/charts/templates/clusterrole.yaml
+++ b/kwok/charts/templates/clusterrole.yaml
@@ -36,7 +36,7 @@ rules:
     resources: ["pods", "nodes", "persistentvolumes", "persistentvolumeclaims", "replicationcontrollers", "namespaces"]
     verbs: ["get", "list", "watch"]
   - apiGroups: ["storage.k8s.io"]
-    resources: ["storageclasses", "csinodes"]
+    resources: ["storageclasses", "csinodes", "volumeattachments"]
     verbs: ["get", "watch", "list"]
   - apiGroups: ["apps"]
     resources: ["daemonsets", "deployments", "replicasets", "statefulsets"]
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -77,7 +77,7 @@ func NewControllers(
 		informer.NewPodController(kubeClient, cluster),
 		informer.NewNodePoolController(kubeClient, cluster),
 		informer.NewNodeClaimController(kubeClient, cluster),
-		termination.NewController(kubeClient, cloudProvider, terminator.NewTerminator(clock, kubeClient, evictionQueue, recorder), recorder),
+		termination.NewController(clock, kubeClient, cloudProvider, terminator.NewTerminator(clock, kubeClient, evictionQueue, recorder), recorder),
 		metricspod.NewController(kubeClient),
 		metricsnodepool.NewController(kubeClient),
 		metricsnode.NewController(cluster),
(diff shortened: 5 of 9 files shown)


pkg/controllers/node/termination/controller.go
```diff
@@ -132,6 +160,23 @@ func (c *Controller) deleteAllNodeClaims(ctx context.Context, node *v1.Node) err
 	return nil
 }
 
+func (c *Controller) waitForVolumeDetachments(ctx context.Context, node *v1.Node, backoff wait.Backoff) error {
+	err := wait.ExponentialBackoffWithContext(ctx, backoff, func(ctx context.Context) (bool, error) {
```
AndrewSirenko, line 164: Is there a better polling mechanism to use than wait.ExponentialBackoff? I don't think a reconciler requeue works here because we WANT to give up on waiting for volumeattachment objects to be deleted if it takes too long.

This is because adding the out of service taint is enough to prevent the 6+ delay. (albeit not the 1+ minute delay)

pkg/controllers/node/termination/controller.go
```diff
@@ -46,6 +48,20 @@ import (
 	nodeclaimutil "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
 )
 
+var (
```
AndrewSirenko, line 51: If we go with the backoff approach, where should I put these?

pkg/controllers/node/termination/controller.go
```diff
@@ -104,12 +120,24 @@ func (c *Controller) finalize(ctx context.Context, node *v1.Node) (reconcile.Res
...
 	// until the full instance shutdown has taken place
 	if err := c.cloudProvider.Delete(ctx, nodeclaimutil.NewFromNode(node)); cloudprovider.IgnoreNodeClaimNotFoundError(err) != nil {
 		return reconcile.Result{}, fmt.Errorf("terminating cloudprovider instance, %w", err)
 	}
+	// This taint confirms node termination so leftover VolumeAttachments are marked as unmounted and can be detached
+	if err := c.terminator.TaintOutOfService(ctx, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("tainting node as out-of-service, %w", err)
+	}
+	// Wait to make sure AttachDetach controller notices node is out-of-service.
+	_ = c.waitForVolumeDetachments(ctx, node, outOfServiceTaintBackoff)
```
AndrewSirenko, line 139: FYI on a small cluster it usually takes 3-8 seconds between applying out-of-service taint and deleting the node finalizer for the AttachDetach controller to see that the node had the taint before node is deleted. 

We can either wait on volumeDetachments with shorter backoff here or just wait until all volumeattachments have DeletionTimestamp if we're in a rush to remove node finalizer. 

Today the node object has to stay up long enough for attachdetach controller to see that the node has this out-of-service taint so that it can apply deletion timestamp to volumeattachment.

pkg/controllers/node/termination/controller.go
```diff
@@ -104,12 +120,24 @@ func (c *Controller) finalize(ctx context.Context, node *v1.Node) (reconcile.Res
 		}
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+	// In order for stateful pods to smoothly migrate from the terminating Node, we wait for VolumeAttachments
+	// to be cleaned up before terminating the node and removing it from the cluster.
+	if err := c.waitForVolumeDetachments(ctx, node, volumeDetachmentsBackoff); err != nil {
+		c.recorder.Publish(terminatorevents.NodeVolumeAttachmentsRemaining(node, err))
```
AndrewSirenko, line 126: Do we prefer logging a warning and resolving errors in waitForVolumeDetachments instead of publishing an event?

AndrewSirenko:
In the future we should consider adding an e2e test to `karpenter-provider-aws` that tests that a stateful workload on a consolidating node is migrated and starts running on second node < 6 minutes. @jmdeal mentioned that EBS CSI Driver is already installed on e2e environment. Thoughts?

pkg/controllers/node/termination/controller.go
jmdeal: If possible we should continue to use the reconciler. IMO there are a couple of options for how we can achieve the same results:
- Add an annotation to the node to indicate when the drain attempt begin. Continue to reconcile until an upper time limit has been hit.
- Do the same thing but with an in-memory map from node to timestamp.
Either option would work, the advantage to an annotation is that we would persist that information through a controller restart, though I'm not sure how valuable that really is here.

pkg/controllers/node/termination/controller.go
jmdeal: Are we convinced we should be using the `out-of-service` taint here? I'm not aware of what all it does, other than detaching volumes and taint driven pod eviction, but everywhere it's referenced it's emphasized to only use this on nodes which have already been shut off. Do we know why that is? Is it because of potential data corruption caused by detaching volumes from running processes or something else? 

I'm also wondering if we want to be detaching volumes if pods are still running and using them (e.g. if they tolerated the disruption taint). It seems like there's still an opportunity for them to be cleaned up during graceful shutdown.

pkg/utils/node/node.go
```diff
@@ -64,6 +66,23 @@ func GetProvisionablePods(ctx context.Context, kubeClient client.Client) ([]*v1.
 	}), nil
 }
 
+// GetVolumeAttachments grabs all attached volumeAttachments to passed node
+func GetVolumeAttachments(ctx context.Context, kubeClient client.Client, node *v1.Node) ([]*storagev1.VolumeAttachment, error) {
```
jmdeal, line 70: I'm wondering if rather than getting all `VolumeAttachment` objects associated with the node, if we should only get `VolumeAttachment` objects associated with the drainable pods on the nodes. We know that any others won't be able to detach successfully until the node is terminating, and they could potentially be detached during graceful shutdown (if configured). Seems like it should be reasonable to look at all PVCs on drainable pods and map them to the correct `VolumeAttachment`, not sure about inline volumes though.

pkg/controllers/node/termination/controller.go
jmdeal: I'm thinking we should handle this through the reconciler natively, in which case inlining it is fine.

pkg/utils/node/node.go
AndrewSirenko: Great call!

Regarding Inline: Not sure if we have to worry about inline volumes because they should only be used when lifecycle of volume matches that of pod:

```
A CSI driver is not suitable for inline use when:
- The volume needs to persist longer than the lifecycle of a pod
```

src: https://kubernetes.io/blog/2022/08/29/csi-inline-volumes-ga/

pkg/controllers/node/termination/controller.go
```diff
@@ -107,6 +108,15 @@ func (c *Controller) finalize(ctx context.Context, node *v1.Node) (reconcile.Res
 		}
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+	// In order for stateful pods to smoothly migrate from the terminating Node, we wait for VolumeAttachments
+	// of drain-able pods to be cleaned up before terminating the node and removing it from the cluster.
+	areVolumesDetached, err := c.ensureVolumesDetached(ctx, node)
```
jmdeal, line 113: Now that Karpenter supports TerminationGracePeriod (#916), we'll need to support a mechanism that ensures we don't stall here indefinitely if detachments fail. This would only kick in if TGP is configured, otherwise we would wait indefinitely. I don't think this should be too complicated, really we just need to check if currentTime - deletionTimestamp > TGP and proceed if it is.

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,80 @@
...
+	if len(volumeAttachments) == 0 {
+		return volumeAttachments, nil
+	}
+	// Filter out non-drain-able pods
+	pods, err := nodeutil.GetPods(ctx, kubeClient, node)
+	if err != nil {
+		return nil, err
+	}
+	drainablePods := lo.Reject(pods, func(p *v1.Pod, _ int) bool {
+		return pod.ToleratesDisruptionNoScheduleTaint(p)
```
jmdeal, line 47: I think we also need to exclude [static pods](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) here. I'm also thinking we should exclude volume attachments from pods which are [stuck terminating](https://github.com/kubernetes-sigs/karpenter/blob/a122e25df3decf77daf0930d5fb11f7c91c6b7ca/pkg/utils/pod/scheduling.go#L132-L138). If a pod gets stuck terminating past it's TGP, we want to be able to proceed and terminate the EC2 instance.

pkg/utils/node/node.go
jmdeal: Just to confirm, since the lifetime is tied together does that mean we know the VolumeAttachment will be deleted once the pod is terminated? Or is it still deleted asynchronously after?

AndrewSirenko: I am pretty sure pods with inline volumes never even have volume attachment objects that karpenter waitForVolumeDetaches could get stuck on. 

I'll double check this tomorrow, but I came to this belief from reading https://kubernetes.io/blog/2020/01/21/csi-ephemeral-inline-volumes/ and the https://kubernetes-csi.github.io/docs/ephemeral-local-volumes.html#implementing-csi-ephemeral-inline-support mentioning that "ephemeral volumes will not be created using the Controller Service, but the Node Service [and therefore kubelet], instead."

AndrewSirenko:
/hold

Need to Confirm the following through Manual tests:
- multi attach volume doesn’t block node deletion
- In line volume doesn’t block node deletion (because no volumeattachment objects)
- static pod doesn't block node deletion (because no volumeattachment objects?)

Feel free to still lgtm or have other teammates review if code looks fine @jmdeal @tallaxes @jackfrancis

pkg/utils/volumeattachment/volumeattachment.go
AndrewSirenko: > I think we also need to exclude static pods here.

I'm not sure if static pods can even have volumeattachments because "The spec of a static Pod cannot refer to other API objects," but will double check manually. 

> If a pod gets stuck terminating past it's TGP, we want to be able to proceed and terminate the EC2 instance.

Does the current PR address this implicitly by skipping waitForVolumeDetachments if after TGP? 

I believe the 1 second reconciler requeue if we hit TGP time during waitForVolumeDetachments is a good feature (1 sec is palatable delay for customers, but 1 sec is enough time for a few more volumes to get unmounted such that we avoid 6+ min delay)

AndrewSirenko: Yeah static pods are not allowed to have non host-path and empty dir volumes. Therefore no volume attachment objects. 

"K/K #103587: Kubelet should enforce that static pods do not reference other API objects" mentions that the mirrored pod will fail to create because of API validation, even if kubelet doesn't enforce it.

pkg/controllers/node/termination/controller.go
jmdeal: That's a good point, we don't currently differentiate between critical and non-critical pods when deleting for TGP. Not something I think we need to consider for this PR, but I think you're right that that would be the more correct behavior.

pkg/utils/volumeattachment/volumeattachment.go
jmdeal: Sounds good, thanks for looking into the static pod case. 

> Does the current PR address this implicitly by skipping waitForVolumeDetachments if after TGP?

Not necessarily, since TGP isn't a required field. If a user doesn't configure TGP, they could potentially be stuck here indefinitely, whereas before Karpenter would proceed to terminate the instance.

AndrewSirenko:
/hold 

Re-doing manual tests now that Karpenter is v1.
- No 1 minute EC2 delay
- multi attach volume doesn’t block node deletion
- Node TGP Works properly
- workload tolerating Disruption taint does not block node deletion

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,80 @@
...
+	// No need to filter empty volumeAttachments list
+	if len(volumeAttachments) == 0 {
+		return volumeAttachments, nil
+	}
+	// Filter out non-drain-able pods
+	pods, err := nodeutil.GetPods(ctx, kubeClient, node)
+	if err != nil {
+		return nil, err
+	}
+	drainablePods := lo.Filter(pods, func(p *v1.Pod, _ int) bool {
```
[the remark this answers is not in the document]
AndrewSirenko: Alas, because volume attachment lifecycle is separate from pod lifecycle, these pods can technically still have volume attachments associated with them. Therefore `IsWaitingEviction` or `IsEvictable` are not able to be used.

This is similar to why terminated pods can still have volume attachments, and why we have this whole 6+ minute delay issue in the first place. 

To be fair, this would be pretty rare, because it would mean the pod succeeded/failed right as node was terminating.

pkg/utils/pod/scheduling.go
```diff
@@ -137,6 +137,17 @@ func IsStuckTerminating(pod *corev1.Pod, clk clock.Clock) bool {
 	return IsTerminating(pod) && clk.Since(pod.DeletionTimestamp.Time) > time.Minute
 }
 
+// IsDrainable checks if a pod can be drained by Karpenter
+// It checks whether the following is true for the pod:
+// - Doesn't tolerate the "karpenter.sh/disruption=disrupting" taint
+// - Is not stuck terminating
+// - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
+func IsDrainable(pod *corev1.Pod, clk clock.Clock) bool {
```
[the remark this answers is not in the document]
AndrewSirenko: See https://github.com/kubernetes-sigs/karpenter/pull/1294#discussion_r1690396271

pkg/utils/node/node.go
```diff
@@ -76,6 +77,24 @@ func GetProvisionablePods(ctx context.Context, kubeClient client.Client) ([]*cor
 	}), nil
 }
 
+// GetVolumeAttachments grabs all volumeAttachments of passed node
+func GetVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node) ([]*storagev1.VolumeAttachment, error) {
+	var volumeAttachments []*storagev1.VolumeAttachment
+	var volumeAttachmentList storagev1.VolumeAttachmentList
+	if err := kubeClient.List(ctx, &volumeAttachmentList, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
+		// If there have not been any volumeAttachments, index may not exist. Therefore, fall back to default List
+		if err = kubeClient.List(ctx, &volumeAttachmentList); err != nil {
```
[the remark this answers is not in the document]
AndrewSirenko: If cluster never had a volumeattachment, `kubeClient.List(ctx, &volumeAttachmentList, client.MatchingFields{"spec.nodeName": node.Name})` would fail due to error like "index doesn't exist." But the regular List call works fine in this case, which is why we fall back to it. 

My hunch was that this is because it is not a natural API Server index, and there is some just-in-time index creation being done?

pkg/utils/node/node.go
```diff
@@ -76,6 +77,24 @@ func GetProvisionablePods(ctx context.Context, kubeClient client.Client) ([]*cor
 	}), nil
 }
 
+// GetVolumeAttachments grabs all volumeAttachments of passed node
+func GetVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node) ([]*storagev1.VolumeAttachment, error) {
```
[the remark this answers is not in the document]
AndrewSirenko: My understanding was that kubeClient.List() is already fetching from some hydrated cache? But I'm not an expert here. 

Or is this just a second thought connected to your previous comment https://github.com/kubernetes-sigs/karpenter/pull/1294#discussion_r1690282131.

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,80 @@
...
+		pvName := volumeAttachments[i].Spec.Source.PersistentVolumeName
+		if pvName != nil && shouldNotFilterOutVolume[*pvName] {
+			filteredVolumeAttachments = append(filteredVolumeAttachments, volumeAttachments[i])
+		}
+	}
+	return filteredVolumeAttachments, nil
+}
+
+// CannotMultiAttach returns true if the persistentVolumeClaim's underlying volume cannot be attached to multiple nodes
+// i.e. its access mode is not ReadWriteOnce/ReadWriteOncePod
```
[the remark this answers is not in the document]
AndrewSirenko: TLDR: Deleting this code because the faster node termination is not worth orphaned volume attachments. This implicitly resolves this comment

---

Context: Stateful workloads necessitating multi-attach volumes DO NOT run into 6+ minute delays. This is because the rescheduled workload never gets the `FailedAttachVolume, Multi-attach error" because that volume IS allowed to be attached to multiple nodes at once (and therefore have more than 1 volume attachment object in kubernetes at a time) 

```
FailedAttachVolume  108s  attachdetach-controller  Multi-Attach error for volume "pvc-01998971-b963-4ea8-9170-53cb49837082" Volume is already exclusively attached to one node and can't be attached to another
```

The point of filtering out volumes that are not ReadWriteOnce/ReadWriteOncePod access mode (and therefore are node multi-attach) was so that we can terminate node quicker. 

However, the more I think about it, it's probably best to err on the side of waiting for volume attachment object to be cleaned up before deleting node object because we don't want orphaned volume attachment objects at all. Especially when Kubernetes v1.32 might NEVER force detach volume attachments. 

We can always add this optimization later.

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,80 @@
...
+	// Filter out non-drain-able pods
+	pods, err := nodeutil.GetPods(ctx, kubeClient, node)
+	if err != nil {
+		return nil, err
+	}
+	drainablePods := lo.Filter(pods, func(p *v1.Pod, _ int) bool {
+		return pod.IsDrainable(p, clk)
+	})
+	// Filter out Multi-Attach volumes
+	shouldNotFilterOutVolume := make(map[string]bool)
```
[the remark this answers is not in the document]
AndrewSirenko: I don't think we can replace this map with a filter without having an inner loop over unDrainablePods for each VolumeAttachment (VA), right? 

Or are you saying we prefer the inner loop over allocating the map due to wanting to minimize GC? 

We'd have to go UndrainablePods -> PVCs -> PV (name) for each VA within the lo.Reject like so, right?

```
filteredVolumeAttachments := lo.Reject(volumeAttachments, func(v *storagev1.VolumeAttachment, _ int) bool {
	pvName := v.Spec.Source.PersistentVolumeName
	for _, p := range unDrainablePods {
		for _, v := range p.Spec.Volumes {
			pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, p, v)
			if err != nil {
				continue
			}
			if pvc != nil && pvc.Spec.VolumeName == pvName {
				return true
			}
		}
	}
	return false
})
```

Let me know if you'd prefer this.

pkg/controllers/node/termination/controller.go
```diff
@@ -119,6 +123,18 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+	// In order for stateful pods to smoothly migrate from the terminating Node, we wait for VolumeAttachments
+	// of drain-able pods to be cleaned up before terminating the node and removing it from the cluster.
+	// However, if TerminationGracePeriod is configured for node, and we are past that period, we will skip waiting.
+	if nodeTerminationTime == nil || time.Now().Before(*nodeTerminationTime) {
```
jmdeal, line 129: We should use the injected clock here as well rather than the time package

pkg/utils/node/node.go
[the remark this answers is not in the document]
AndrewSirenko: Jason prefers handling the exact error, handing back an empty list instead of making second list call.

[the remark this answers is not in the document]
AndrewSirenko: Welp perhaps my grep-fu is not good enough, but looks like a wrapped version of this error isn't explicitly exported anywhere. Seems like every package just uses [client-go](https://github.com/kubernetes/client-go/blob/93c6a5bf507f28f7d004b26a6f9f65fa2dc9f69c/tools/cache/thread_safe_store.go#L76) and `fmt.Errorf("Index with name %s does not exist", indexName)`

See my k8s.hound query: https://cs.k8s.io/?q=Index%20with%20name%20.*%20does%20not%20exist&i=nope&files=&excludeFiles=&repos=

And the CI run which shows error on my test PR: https://github.com/kubernetes-sigs/karpenter/actions/runs/10100097019/job/27930677893?pr=1471

I think string matching on the error message here in production code is a non-starter... Even though the message ever changing is extremely unlikely. 

---

Any cleaner alternatives @jmdeal @jonathan-innis? I see two alternatives to the backup un-indexed List:
1. If this indexed list call fails, we always return empty list of VolumeAttachments instead of returning error. 
2. Attempting a string match on error message.

pkg/utils/pod/scheduling.go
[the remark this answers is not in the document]
AndrewSirenko: Resolving to move conversation over to other comment

pkg/utils/node/node.go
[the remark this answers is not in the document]
jmdeal: I agree, string matching on the error absolutely doesn't sound like the way to go. I also don't think returning an empty list would be a good approach since this would capture transient errors (e.g. timeout). Another option we could consider is tracking VolumeAttachments through cluster state, but that really seems overkill. I'm ok with taking the current solution unless @jonathan-innis has any objections.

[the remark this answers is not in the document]
AndrewSirenko: I was very wrong, tests weren't failing because a volumeattachment was never created, they were failing because functional tests never created the volumeattachment custom index.

Thanks @jmdeal for figuring this out...

Will take duplicate list out of PR because there shouldn't be a case where index does not exist.

pkg/utils/volumeattachment/volumeattachment.go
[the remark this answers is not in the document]
jonathan-innis: Got it -- agree that we should catch this case even if the pod is no longer active so makes sense 👍🏼

pkg/utils/pod/scheduling.go
```diff
@@ -137,6 +137,17 @@ func IsStuckTerminating(pod *corev1.Pod, clk clock.Clock) bool {
...
+	return !ToleratesDisruptionNoScheduleTaint(pod) &&
```
jonathan-innis, line 146: Do we want to modify `IsWaitingEviction` and put this function next to it since this is now a subset of that other function? So then `IsWaitingEviction` is just looking at the pods that exist that are drainable + removing the pods that are already terminated

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,68 @@
...
+	pods, err := nodeutil.GetPods(ctx, kubeClient, node)
+	if err != nil {
+		return nil, err
+	}
+	unDrainablePods := lo.Reject(pods, func(p *corev1.Pod, _ int) bool {
+		return pod.IsDrainable(p, clk)
+	})
+	// Filter out VolumeAttachments associated with non-drain-able Pods
+	// Match on Pod -> PersistentVolumeClaim -> PersistentVolume Name <- VolumeAttachment
+	shouldFilterOutVolume := make(map[string]bool)
```
jonathan-innis, line 51: nit: consider using `sets.New[string]()` instead of a bool map

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,68 @@
...
+	for _, p := range unDrainablePods {
+		for _, v := range p.Spec.Volumes {
+			pvc, err := volumeutil.GetPersistentVolumeClaim(ctx, kubeClient, p, v)
+			if err != nil {
```
jonathan-innis, line 55: What should we do here if the `Get()` fails with a transient error?

pkg/utils/volumeattachment/volumeattachment.go
```diff
@@ -0,0 +1,68 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	nodeutil "sigs.k8s.io/karpenter/pkg/utils/node"
+	"sigs.k8s.io/karpenter/pkg/utils/pod"
+	volumeutil "sigs.k8s.io/karpenter/pkg/utils/volume"
+)
+
+// FilterVolumeAttachments filters out storagev1.VolumeAttachments that should not block the termination
+// of the passed corev1.Node
+func FilterVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node, volumeAttachments []*storagev1.VolumeAttachment, clk clock.Clock) ([]*storagev1.VolumeAttachment, error) {
```
jonathan-innis, line 36: nit (style): I'd either move this to the node termination controller itself since it's so tightly scoped to that use-case OR I would rename this function in a way that indicates what this is filtering the volume attachments on. Because this is really specific to filtering out volume attachments that are attached to pods that are not drainable by Karpenter I'd indicate that through the naming

pkg/controllers/node/termination/controller.go
```diff
@@ -158,6 +174,19 @@ func (c *Controller) deleteAllNodeClaims(ctx context.Context, nodeClaims ...*v1.
 	return nil
 }
 
+func (c *Controller) ensureVolumesDetached(ctx context.Context, node *corev1.Node) (volumesDetached bool, err error) {
+	volumeAttachments, err := nodeutils.GetVolumeAttachments(ctx, c.kubeClient, node)
+	if err != nil {
+		return false, err
+	}
+	// Filter out VolumeAttachments associated with not drain-able Pods
+	filteredVolumeAttachments, err := volumeattachmentutils.FilterVolumeAttachments(ctx, c.kubeClient, node, volumeAttachments, c.clock)
```
jonathan-innis, line 183: I could see you shifting this helper logic directly into this file or putting this inline in this function here

pkg/utils/volumeattachment/volumeattachment.go
AndrewSirenko: Shall we skip fetching that pod's PVCs in that case? Therefore we would not filter OUT those volumes associated with that undrainable pod. At worst, that blocks termination for one reconciliation loop (1 second delay). Perhaps log warning.

On next reconciliation, get() should not fail again with transient error. 

Issue here is that if error not transient, we are reconciling forever. 

What do you think?

jonathan-innis: > Issue here is that if error not transient, we are reconciling forever

We'll only reconcile forever if we don't hit our terminationGracePeriod which I think is a reasonable trade-off for safety.

jonathan-innis: > Shall we skip fetching that pod's PVCs in that case

Seems like we should throw the error back upward if it's not a NotFound error and just let the reconciler handle the error and the requeue

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (commented):
/lgtm

AndrewSirenko:
/unhold 

^^

PR #1295, opened 2024-06-04
chore: fix up kwok cloud provider

njtran:
Fixes #N/A 

**Description**
This adds necessary RBAC permissions, copies the CRDs to a kwok specific folder so kwok yq'd requirements don't affect the real CRDs, and applies the crds only within the kwok folder for the kwok cloud provider.

This also changes the replicas field to 1, since usually kwok is run on kind nodes, which most of the time have 1 node. 

make apply && apply KWOKNodeClass and NodePool


Makefile
```diff
@@ -31,9 +31,8 @@ build: ## Build the Karpenter KWOK controller images using ko build
 	$(eval IMG_DIGEST=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 2))
 	
 apply-with-kind: verify build-with-kind ## Deploy the kwok controller from the current state of your git repository into your ~/.kube/config cluster
-	hack/validation/kwok-requirements.sh
-	kubectl apply -f pkg/apis/crds
-	helm upgrade --install karpenter kwok/charts --namespace $(KARPENTER_NAMESPACE) --skip-crds \
+	kubectl apply -f kwok/charts/crds
+	kubectl apply -f kwok/apis/crds	helm upgrade --install karpenter kwok/charts --namespace $(KARPENTER_NAMESPACE) --skip-crds \
```
jonathan-innis, line 35: Is this line correct? Does a kubectl apply with a helm upgrade work?

Makefile
```diff
@@ -31,9 +31,8 @@ build: ## Build the Karpenter KWOK controller images using ko build
 	$(eval IMG_DIGEST=$(shell echo $(CONTROLLER_IMG) | cut -d "@" -f 2))
 	
 apply-with-kind: verify build-with-kind ## Deploy the kwok controller from the current state of your git repository into your ~/.kube/config cluster
-	hack/validation/kwok-requirements.sh
-	kubectl apply -f pkg/apis/crds
-	helm upgrade --install karpenter kwok/charts --namespace $(KARPENTER_NAMESPACE) --skip-crds \
+	kubectl apply -f kwok/charts/crds
```
jonathan-innis, line 34: Why are we applying both a kubectl apply on the chart CRDs and a kubectl apply on kwok API CRDs

Makefile
```diff
@@ -91,6 +89,8 @@ verify: ## Verify code. Includes codegen, docgen, dependencies, linting, formatt
 	hack/validation/labels.sh
 	hack/validation/resources.sh
 	hack/validation/status.sh
+	cp -r pkg/apis/crds kwok/charts
```
jonathan-innis, line 92: Consider doing a symlink like we do elsewhere with our CRDs

kwok/apis/apis.go
```diff
@@ -20,10 +20,11 @@ import (
 	_ "embed"
 
 	"github.com/samber/lo"
-	"k8s.io/api/node/v1alpha1"
```
jonathan-innis, line 23: Was this scheme not correct before?

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -1 +1,459 @@
-../../../pkg/apis/crds/karpenter.sh_nodeclaims.yaml
\ No newline at end of file
+---
```
jonathan-innis, line 1: Why are you removing the symlink?

kwok/charts/crds/karpenter.sh_nodepools.yaml
```diff
@@ -1 +1,526 @@
-../../../pkg/apis/crds/karpenter.sh_nodepools.yaml
\ No newline at end of file
+---
```
jonathan-innis, line 1: Same question here

kwok/charts/values.yaml
```diff
@@ -30,7 +30,7 @@ serviceMonitor:
   # -- Endpoint configuration for the ServiceMonitor.
   endpointConfig: {}
 # -- Number of replicas.
-replicas: 2
+replicas: 1
```
jmdeal, line 33: Wonder if we should disable leader election when setting replicas to 1. Doesn't appear to be necessary, and if you've tested this without issues we don't need to do this. If we wanted to though, something along the lines of this should work:
```yaml
- name: LEADER_ELECT
  value: {{ int .Values.replicas | eq 1 | ternary "false" "true" | quote }}

```

jmdeal (commented):
One optional comment, otherwise looks good

Makefile
jmdeal: I think the point is we don't want to modify the original CRDs when applying the additional kwok validation requirements

Makefile
njtran: definitely not...... sad.... didn't test this target......

Makefile
njtran: Are you asking why they're not in the same directory? Or are you wondering why we care about doing two separate CRD applys?

Makefile
njtran: That's exactly the point. i've had too many PRs where `make apply` made the change, but then I had to restore the change so it wouldn't be included. I'd liken this model to the same way that we do it in the AWS Provider. We essentially copy over the upstream nodepool/nodeclaim apis and add our own validations then apply it. KWOK is doing the same but just in the same repo.

kwok/apis/apis.go
njtran: Correct

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
njtran: I wanted to copy the yamls over so we could add our own validation. Am I missing something by removing these?

kwok/charts/values.yaml
njtran: my gut feeling is we can just opt for simplicity here and keep it as is. Leader election usually runs into problems when there's more vying to be the leader, not less.

Makefile
jonathan-innis: I'm trying to understand why we are doing two separate CRD applies if we have the CRDs in the charts directory

jonathan-innis: Closed after adding a symlink to the charts directory to pull in the CRD

jonathan-innis (approved):
/lgtm
/approve

PR #1296, opened 2024-06-04
chore: move disruption and pdb utils and factor isDisruptable into cluster state

njtran:
Fixes #N/A 

**Description**
Moves disruption and pdb utils and factor isDisruptable into cluster state to stage in some changes for ConsolidateAfter.

make apply 


pkg/utils/disruption/disruption.go
```diff
@@ -0,0 +1,123 @@
...
+		ageInSeconds := clock.Since(node.CreationTimestamp.Time).Seconds()
+		totalLifetimeSeconds := nodePool.Spec.Disruption.ExpireAfter.Duration.Seconds()
+		lifetimeRemainingSeconds := totalLifetimeSeconds - ageInSeconds
+		remaining = lo.Clamp(lifetimeRemainingSeconds/totalLifetimeSeconds, 0.0, 1.0)
+	}
+	return remaining
+}
+
+// GetPodEvictionCost returns the disruption cost computed for evicting the given pod.
+func GetPodEvictionCost(ctx context.Context, p *v1.Pod) float64 {
```
rschalo, line 51: Wdyt about dropping `Get` and `Pod` like
```suggestion
func EvictionCost(ctx context.Context, p *v1.Pod) float64 {
```

pkg/controllers/state/statenode.go
```diff
@@ -150,6 +158,54 @@ func (in *StateNode) Pods(ctx context.Context, c client.Client) ([]*v1.Pod, erro
 	return nodeutils.GetPods(ctx, c, in.Node)
 }
 
+// IsDisruptable returns an error if the StateNode cannot be disrupted
+// This checks all associated StateNode internals, node labels, and do-not-disrupt annotations
+// on both the pod and the node.
+// IsDisruptable takes in a recorder to emit events on the nodeclaims when the state node is not a candidate
+//
+//nolint:gocyclo
+func (in *StateNode) IsDisruptable(ctx context.Context, kubeClient client.Client) error {
```
rschalo, line 167: nit: I'd expect a bool from this func name

pkg/controllers/disruption/types.go
```diff
@@ -36,6 +36,8 @@ import (
 	"sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/events"
+	disruptionutils "sigs.k8s.io/karpenter/pkg/utils/disruption"
```
rschalo, line 39: Thoughts about just calling this utils since we are already in the disruption package?

pkg/controllers/state/statenode.go
njtran: We've used the convention of an error to be a bool + string throughout the repo. What would the name need to be for you to think that error should be the return ?

pkg/controllers/disruption/types.go
njtran: This isn't in the same overall package, but yes it's mostly to follow convention. I think it's more clear (especially if there are more util packages being used) that we prepend/append it with the type of util it is. Thoughts?

pkg/utils/disruption/disruption.go
njtran: Good call!

pkg/utils/disruption/disruption.go
```diff
@@ -0,0 +1,123 @@
...
+	for _, p := range pods {
+		cost += EvictionCost(ctx, p)
+	}
+	return cost
+}
+
+// FilterByPrice returns the instanceTypes that are lower priced than the current candidate.
+// The minValues requirement is checked again after FilterByPrice as it may result in more constrained InstanceTypeOptions for a NodeClaim.
+// If the result of the filtering means that minValues can't be satisfied, we return an error.
+func FilterByPrice(options []*cloudprovider.InstanceType, reqs scheduling.Requirements, price float64) ([]*cloudprovider.InstanceType, error) {
```
ellistarn, line 85: We have a cloudprovider.InstanceTypes type that we hang a bunch of methods off of. Do we think this might be better encapsulated there?

pkg/utils/disruption/disruption.go
```diff
@@ -0,0 +1,123 @@
...
+	// the scheduling priority is in [-2147483648, 1000000000]
+	if p.Spec.Priority != nil {
+		cost += float64(*p.Spec.Priority) / math.Pow(2, 25)
+	}
+
+	// overall we clamp the pod cost to the range [-10.0, 10.0] with the default being 1.0
+	return lo.Clamp(cost, -10.0, 10.0)
+}
+
+func DisruptionCost(ctx context.Context, pods []*v1.Pod) float64 {
```
ellistarn, line 74: ctx is unused. 

This is only used in one place -- do we think it might be better homed in controllers/disruption until we have more callers?

njtran: context is used in the EvictionCost() and logging bit. It'll be used in the underutilization controller in the consolidateAfter PR, so this is staging in that change to reduce the scope of the other PR. Happy to not move it if you think that's easier to review as part of ConsolidateAfter.

pkg/controllers/state/statenode.go
```diff
@@ -70,6 +71,13 @@ func (n StateNodes) Pods(ctx context.Context, c client.Client) ([]*v1.Pod, error
 	return pods, nil
 }
 
+// Disruptable filters StateNodes that are meet the IsDisruptable condition
+func (n StateNodes) Disruptable(ctx context.Context, kubeClient client.Client, recorder events.Recorder) StateNodes {
```
ellistarn, line 75: I'm noticing a lot of these helper methods pass dependencies around (e.g. kubeclient, recorder). Instead, you could encapsulate them inside of our StateNodes struct to minimize parameter wiring

pkg/controllers/disruption/events/events.go
```diff
@@ -102,23 +102,26 @@ func Unconsolidatable(node *v1.Node, nodeClaim *v1beta1.NodeClaim, reason string
 
 // Blocked is an event that informs the user that a NodeClaim/Node combination is blocked on deprovisioning
 // due to the state of the NodeClaim/Node or due to some state of the pods that are scheduled to the NodeClaim/Node
-func Blocked(node *v1.Node, nodeClaim *v1beta1.NodeClaim, reason string) []events.Event {
-	return []events.Event{
-		{
+func Blocked(node *v1.Node, nodeClaim *v1beta1.NodeClaim, reason string) (evs []events.Event) {
+	if node != nil {
```
ellistarn, line 106: Curious why we're reasoning about both types of objects here.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -202,7 +203,7 @@ func filterOutSameType(newNodeClaim *scheduling.NodeClaim, consolidate []*Candid
 			}
 		}
 	}
-	return filterByPrice(newNodeClaim.InstanceTypeOptions, newNodeClaim.Requirements, maxPrice)
+	return disruptionutils.FilterByPrice(newNodeClaim.InstanceTypeOptions, newNodeClaim.Requirements, maxPrice)
```
ellistarn, line 206: sugg: `newNodeClaim.InstanceTypeOptions.FilterByPrice(nodeNodeClaim.Requiremenets, maxPrice)`

pkg/controllers/disruption/events/events.go
njtran: We emit an event for both the node and nodeclaim. If there's no node or nodeclaim object associated with the computation we're doing, we won't emit. Is that what you're asking?

pkg/controllers/disruption/suite_test.go
```diff
@@ -1066,24 +1068,24 @@ var _ = Describe("Candidate Filtering", func() {
 				},
 			},
 		})
-		pdb := test.PodDisruptionBudget(test.PDBOptions{
+		budget := test.PodDisruptionBudget(test.PDBOptions{
```
ellistarn, line 1071: curious why we renamed these

pkg/controllers/state/statenode.go
ellistarn: (string, bool) is a pretty common go idiom for this, though both `*string` and `error` represent almost the same thing

pkg/controllers/state/statenode.go
ellistarn: You might even want to place it inside StateNode itself

pkg/utils/disruption/disruption.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+// lifetimeRemaining calculates the fraction of node lifetime remaining in the range [0.0, 1.0].  If the TTLSecondsUntilExpired
+// is non-zero, we use it to scale down the disruption costs of candidates that are going to expire.  Just after creation, the
+// disruption cost is highest, and it approaches zero as the node ages towards its expiration time.
+func LifetimeRemaining(clock clock.Clock, nodePool *v1beta1.NodePool, node *v1.Node) float64 {
```
ellistarn, line 39: As an alternative encapsulation: nodePool.lifetimeRemaining(nodeClaim).

I think we probably want to define time w.r.t. NodeClaim and not node.

pkg/utils/disruption/disruption.go
```diff
@@ -0,0 +1,123 @@
...
+
+// EvictionCost returns the disruption cost computed for evicting the given pod.
+func EvictionCost(ctx context.Context, p *v1.Pod) float64 {
+	cost := 1.0
+	podDeletionCostStr, ok := p.Annotations[v1.PodDeletionCost]
+	if ok {
+		podDeletionCost, err := strconv.ParseFloat(podDeletionCostStr, 64)
+		if err != nil {
+			logging.FromContext(ctx).Errorf("parsing %s=%s from pod %s, %s",
+				v1.PodDeletionCost, podDeletionCostStr, client.ObjectKeyFromObject(p), err)
```
ellistarn, line 58: Should we return max float value or something here?

pkg/utils/disruption/disruption.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// WorstLaunchPrice gets the worst-case launch price from the offerings that are offered
+// on an instance type. If the instance type has a spot offering available, then it uses the spot offering
+// to get the launch price; else, it uses the on-demand launch price
+func WorstLaunchPrice(ofs cloudprovider.Offerings, reqs scheduling.Requirements) float64 {
+	// We prefer to launch spot offerings, so we will get the worst price based on the node requirements
+	if reqs.Get(v1beta1.CapacityTypeLabelKey).Has(v1beta1.CapacityTypeSpot) {
+		spotOfferings := ofs.Compatible(reqs).Compatible(
+			scheduling.NewRequirements(scheduling.NewRequirement(v1beta1.CapacityTypeLabelKey, v1.NodeSelectorOpIn, v1beta1.CapacityTypeSpot)))
```
ellistarn, line 110: We might gain some scheduling performance by taking defining these static requirements statically. 

```
var (
  SpotRequirement = ...
  OnDemandRequirement = ...
) 
```

pkg/utils/pdb/pdb.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package disruption
+package pdb
```
ellistarn, line 17: should this be in utils or in scheduling?

ellistarn (approved):

pkg/controllers/state/statenode.go
jonathan-innis: My take: Seems odd that a date object would have something that's more global and unrelated to it like the `client.Client`. Is there precedent for doing things like this in upstream K8s or elsewhere in go? I've generally seen this as an anti-pattern.

pkg/utils/pdb/pdb.go
```diff
@@ -29,16 +29,16 @@ import (
 	podutil "sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
-// PDBLimits is used to evaluate if evicting a list of pods is possible.
-type PDBLimits struct {
+// Limits is used to evaluate if evicting a list of pods is possible.
+type Limits struct {
 	ctx        context.Context
 	clk        clock.Clock
 	kubeClient client.Client
```
jonathan-innis, line 36: I have a general question about why we need to store the kubeClient, context, and clock in this data structure when it doesn't appear to be a global construct but more of a data holder? Are there places where we are utilizing this client ad-hoc?

pkg/controllers/state/statenode.go
```diff
@@ -150,6 +158,54 @@ func (in *StateNode) Pods(ctx context.Context, c client.Client) ([]*v1.Pod, erro
...
+		v1beta1.CapacityTypeLabelKey,
+		v1.LabelTopologyZone,
+		v1.LabelInstanceTypeStable,
+		v1beta1.NodePoolLabelKey,
+	} {
+		if _, ok := in.Labels()[label]; !ok {
+			return fmt.Errorf("state node doesn't have required label %q", label)
+		}
+	}
+	pods, err := in.Pods(ctx, kubeClient)
```
jonathan-innis, line 195: Something to keep in mind: If we use this call here and don't store the pods: The pods that we are going to pull to build the candidate are going to be different from the pods that we are using to evaluate whether the StateNode is disruptable. This may not cause a problem, but we should think through race conditions here

pkg/controllers/state/statenode.go
```diff
@@ -150,6 +158,54 @@ func (in *StateNode) Pods(ctx context.Context, c client.Client) ([]*v1.Pod, erro
...
+	if err != nil {
+		return fmt.Errorf("getting pods from state node, %w", err)
+	}
+	for _, po := range pods {
+		// We only consider pods that are actively running for "karpenter.sh/do-not-disrupt"
+		// This means that we will allow Mirror Pods and DaemonSets to block disruption using this annotation
+		if !podutils.IsDisruptable(po) {
+			return fmt.Errorf(`pod %q has "karpenter.sh/do-not-disrupt" annotation`, client.ObjectKeyFromObject(po))
+		}
+	}
```
jonathan-innis, line 205: This doesn't include evaluating pdbs?

pkg/controllers/state/statenode.go
jonathan-innis: I'd almost advise that we try and avoid passing through these kinds of global constructs all together and pre-fetch the _data_ and then pass the _data_ through. I think this is what @ellistarn  is suggesting too to some degree. I agree that we want to be careful about how we are passing these things around

pkg/utils/disruption/disruption.go
jonathan-innis:
```suggestion
func Cost(ctx context.Context, pods []*v1.Pod) float64 {
```

Should this just be `Cost` since this is already in the disruption package?

pkg/controllers/disruption/types.go
```diff
@@ -60,60 +62,31 @@ type Candidate struct {
...
-		return nil, fmt.Errorf("state node doesn't contain both a node and a nodeclaim")
-	}
-	// skip any candidates that are already marked for deletion and being handled
-	if node.MarkedForDeletion() {
-		return nil, fmt.Errorf("state node is marked for deletion")
-	}
-	// skip candidates that aren't initialized
-	if !node.Initialized() {
-		return nil, fmt.Errorf("state node isn't initialized")
+	if err := node.IsDisruptable(ctx, kubeClient); err != nil {
```
jonathan-innis, line 67: Nice! Love that we are returning this error back now and just using that to fire the event!

pkg/controllers/state/statenode.go
jonathan-innis, line 75: I don't think that you even need the recorder here anymore given that we are now returning the event back from the caller

pkg/controllers/disruption/types.go
```diff
@@ -60,60 +62,31 @@ type Candidate struct {
...
 	if nodePool == nil || instanceTypeMap == nil {
-		recorder.Publish(disruptionevents.Blocked(node.Node, node.NodeClaim, fmt.Sprintf("Owning nodepool %q not found", nodePoolName))...)
+		recorder.Publish(disruptionevents.Blocked(node.Node, node.NodeClaim, fmt.Sprintf("NodePool %q not found", nodePoolName))...)
 		return nil, fmt.Errorf("nodepool %q can't be resolved for state node", nodePoolName)
 	}
 	instanceType := instanceTypeMap[node.Labels()[v1.LabelInstanceTypeStable]]
 	// skip any candidates that we can't determine the instance of
 	if instanceType == nil {
-		recorder.Publish(disruptionevents.Blocked(node.Node, node.NodeClaim, fmt.Sprintf("Instance type %q not found", node.Labels()[v1.LabelInstanceTypeStable]))...)
+		recorder.Publish(disruptionevents.Blocked(node.Node, node.NodeClaim, fmt.Sprintf("Instance Type %q not found", node.Labels()[v1.LabelInstanceTypeStable]))...)
```
jonathan-innis, line 87: Why do we think that some of these checks belong at the top-level codepath and others belong in statenode. Is there a way to push everything into a common place or are there specific limitations here that we have to consider?

pkg/controllers/disruption/suite_test.go
njtran: pdbutil import conflict

pkg/cloudprovider/types.go
```diff
@@ -83,6 +88,20 @@ type InstanceType struct {
 
 type InstanceTypes []*InstanceType
 
+func (its InstanceTypes) FilterByMinValues(reqs scheduling.Requirements) (InstanceTypes, error) {
```
ellistarn, line 91: thoughts on making cloudprovider/instancetypes.go to separate out some of this logic?

pkg/cloudprovider/types.go
```diff
@@ -264,6 +283,26 @@ func (ofs Offerings) MostExpensive() Offering {
 	})
 }
 
+// WorstLaunchPrice gets the worst-case launch price from the offerings that are offered
+// on an instance type. If the instance type has a spot offering available, then it uses the spot offering
+// to get the launch price; else, it uses the on-demand launch price
+func (ofs Offerings) WorstLaunchPrice(reqs scheduling.Requirements) float64 {
```
ellistarn, line 289: Same. Does offerings deserve its own file?

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -126,6 +127,18 @@ func (n *NodeClaim) FinalizeScheduling() {
 	delete(n.Requirements, v1.LabelHostname)
 }
 
+func (n *NodeClaim) FilterInstanceTypeOptionsByPriceAndMinValues(reqs scheduling.Requirements, maxPrice float64) (*NodeClaim, error) {
```
ellistarn, line 130: I would expand a Filter function to not mutate. Consider calling RemoveXXX

pkg/controllers/state/cluster.go
```diff
@@ -173,9 +173,13 @@ func (c *Cluster) Nodes() StateNodes {
...
 	defer c.mu.RUnlock()
 
-	return lo.Map(lo.Values(c.nodes), func(n *StateNode, _ int) *StateNode {
-		return n.DeepCopy()
-	})
+	return StateNodes{
+		Nodes: lo.Map(lo.Values(c.nodes), func(n *StateNode, _ int) *StateNode {
+			return n.DeepCopy()
+		}),
+		kubeClient: c.kubeClient,
```
ellistarn, line 180: I was wondering if kubeClient should live in state node itself.

pkg/cloudprovider/types.go
```diff
@@ -83,6 +88,20 @@ type InstanceType struct {
...
 type InstanceTypes []*InstanceType
 
+func (its InstanceTypes) FilterByMinValues(reqs scheduling.Requirements) (InstanceTypes, error) {
+	if !reqs.HasMinValues() {
+		return its, nil
+	}
+
+	// We would have already filtered the invalid NodeClaim not meeting the minimum requirements in simulated scheduling results.
+	// Here the instanceTypeOptions changed again based on the price and requires re-validation.
+	if _, err := its.SatisfiesMinValues(reqs); err != nil {
```
jonathan-innis, line 98: This function is so short, do we event need the helper? Also, it's just validating that minValues are satisfied if they exist based on the SatisfiesMinValues function. there's no actual filtering that's occurring here from what I can tell

pkg/cloudprovider/types.go
jonathan-innis: +1

pkg/controllers/disruption/helpers.go
```diff
@@ -55,21 +54,21 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 	candidateNames := sets.NewString(lo.Map(candidates, func(t *Candidate, i int) string { return t.Name() })...)
 	nodes := cluster.Nodes()
 	deletingNodes := nodes.Deleting()
-	stateNodes := lo.Filter(nodes.Active(), func(n *state.StateNode, _ int) bool {
+	stateNodes := lo.Filter(nodes.Active().Nodes, func(n *state.StateNode, _ int) bool {
```
jonathan-innis, line 57: Why do we have to call `Nodes` again now. This looks pretty awkward to me at first glance

pkg/controllers/disruption/helpers.go
```diff
@@ -55,21 +54,21 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
...
 		return !candidateNames.Has(n.Name())
 	})
 
 	// We do one final check to ensure that the node that we are attempting to consolidate isn't
 	// already handled for deletion by some other controller. This could happen if the node was markedForDeletion
 	// between returning the candidates and getting the stateNodes above
-	if _, ok := lo.Find(deletingNodes, func(n *state.StateNode) bool {
+	if _, ok := lo.Find(deletingNodes.Nodes, func(n *state.StateNode) bool {
```
jonathan-innis, line 64: Same question here. This stutters a bunch and feels potentially uneccesary

pkg/utils/disruption/disruption.go
jonathan-innis: Maybe I'm forgetting that conversation, but I actually think that reasoning about it in terms of the instance lifetime is probably the right thing to do here, rather than the Node. Don't we reason about expiration in terms of the NodeClaim for NodeClaim disruption status conditions?

pkg/controllers/state/statenode.go
jonathan-innis:
```suggestion
func (in *StateNode) ValidateDisruptable(ctx context.Context, kubeClient client.Client) error {
```

Validate gives more indication that you are returning an error rather than an `Is` statement. We definitely do this in calls like in our disruption validation, so that might be more appropriate here if we are looking for a better function name

pkg/controllers/state/cluster.go
jonathan-innis: > I was wondering if kubeClient should live in state node itself.

I'm still really strongly against this pattern of storing kubeClients in data objects. If we need some kind of data accessor, we can build a separate client for that and instantiate it, but storing that kubeClient many times across all StateNodes feels wrong to me

pkg/controllers/state/cluster.go
```diff
@@ -173,9 +173,13 @@ func (c *Cluster) Nodes() StateNodes {
 	c.mu.RLock()
 	defer c.mu.RUnlock()
 
-	return lo.Map(lo.Values(c.nodes), func(n *StateNode, _ int) *StateNode {
-		return n.DeepCopy()
-	})
+	return StateNodes{
+		Nodes: lo.Map(lo.Values(c.nodes), func(n *StateNode, _ int) *StateNode {
```
jonathan-innis, line 177: Do we need to create a new structure like this that causes us have to call into `Nodes` to get our nodes back? This also doesn't really indicate there is a client attached to it. This is back to my thoughts on `StateNode` being a data holder and not a client in and of itself. StateNodes is just a set of data holders.

pkg/controllers/state/cluster.go
jonathan-innis: I'd be more in favor of something like `state.NewClient(ctx, kubeClient)` or similar that allows us to build the client from whatever thing we are injecting our dependencies in from and then building methods off of that.

pkg/utils/disruption/disruption.go
njtran: I didn't want to name it cost, since it implies something about pricing. I can understand the stuttering concern, so I've renamed it to ReschedulingCost

pkg/utils/disruption/disruption.go
njtran: Yeah you're absolutely right, we do calculate it on the nodeclaim, not on the node. I'll change it here.

njtran: On the encapsulation, I'm not sure the juice is worth the squeeze on making it a nodepool member function. It's pretty specific to disruption. I could maybe see an encapsulation making it a nodeclaim member function, but considering it's just 3 parameters, I'm thinking it's fine as it is.

pkg/utils/pdb/pdb.go
njtran: Good call. It really only needs to be a list of PDBItems. Changed it to be a type of list of pdbs.

pkg/controllers/state/statenode.go
njtran: I guess we'd need to return the list of pods in ValidateDisruptable if we think that this is a problem. Realistically, I'm not sure it would change that much, considering that it would only modify the disruption cost.

pkg/controllers/state/statenode.go
njtran: reorg'd

pkg/controllers/state/cluster.go
jonathan-innis: > I'm still really strongly against this pattern of storing kubeClients in data objects

Actually, after thinking about this a bit deeper, I may be completely eating my own words here 😆 . I'm coming around to the idea and think the pattern may actually be reasonable to think about conceptually, "The StateNode stores state about a NodeClaim/Node and has mechanism to grab pods directly from the cluster (because the StateNode doesn't directly store pods from informers)." I'm fine either way. Definitely worth a try if we think that it's going to improve the code readabilitiy

pkg/cloudprovider/types.go
njtran: Eh, Im personally a fan of having the types.go file have the cloudprovider interface + instance types + respective member functions + enums

pkg/controllers/disruption/helpers.go
njtran: Yeah this is because we're wrapping it with kubeClient now. I thought it was cleaner at first, but after having to make these changes, I'm going to revert it to the original.

pkg/controllers/state/cluster.go
ellistarn: > storing kubeClients in data objects

This is an old idea from the RAII pattern from C++

pkg/controllers/state/statenode.go
```diff
@@ -45,23 +47,25 @@ type StateNodes []*StateNode
 
 // Active filters StateNodes that are not in a MarkedForDeletion state
 func (n StateNodes) Active() StateNodes {
-	return lo.Filter(n, func(node *StateNode, _ int) bool {
+	n = lo.Filter(n, func(node *StateNode, _ int) bool {
```
jonathan-innis, line 50: Sorry, why do this?

pkg/controllers/state/statenode.go
```diff
@@ -45,23 +47,25 @@ type StateNodes []*StateNode
...
 		return node.MarkedForDeletion()
 	})
+	return n
 }
 
 // Pods gets the pods assigned to all StateNodes based on the kubernetes api-server bindings
-func (n StateNodes) Pods(ctx context.Context, c client.Client) ([]*v1.Pod, error) {
+func (n StateNodes) Pods(ctx context.Context, kubeClient client.Client) ([]*v1.Pod, error) {
```
jonathan-innis, line 65: This is fine, but I think this may have been updated from moving things around and the rename may not be required here

pkg/controllers/state/statenode.go
```diff
@@ -70,10 +74,23 @@ func (n StateNodes) Pods(ctx context.Context, c client.Client) ([]*v1.Pod, error
...
 
-func (n StateNodes) ReschedulablePods(ctx context.Context, c client.Client) ([]*v1.Pod, error) {
+// Disruptable filters StateNodes that are meet the IsDisruptable condition
+func (n StateNodes) Disruptable(ctx context.Context, clk clock.Clock, kubeClient client.Client) (StateNodes, error) {
+	pdbs, err := pdb.NewLimits(ctx, clk, kubeClient)
+	if err != nil {
+		return StateNodes{}, fmt.Errorf("constructing pdbs, %w", err)
+	}
+	n = lo.Filter(n, func(node *StateNode, _ int) bool {
+		_, err := node.ValidateDisruptable(ctx, kubeClient, pdbs)
```
jonathan-innis, line 84: nit: One liner?

pkg/controllers/state/statenode.go
njtran: This was an artifact, lemme remove.

pkg/controllers/state/statenode.go
njtran: Fair, just renamed them all the kubeClient anyway to make them the same across all functions

jonathan-innis: Sounds fine 👍🏼

pkg/controllers/state/statenode.go
jonathan-innis: Mainly thinking about differences between candidacy and the pods that we actually use for scheduling; though, I think you're right that it probably makes literally no difference, because having to re-grab the pods just means that we will be more correct with our data, it doesn't make any difference from the perspective of whether we choose to work with the node or not

jonathan-innis (approved):
/lgtm
/approve

Issue #1275, opened 2024-05-22
Replace util functions that can be replaced with those from k8s.io/kubernetes


rschalo:
### Description

**What problem are you trying to solve?**
Example is https://github.com/kubernetes-sigs/karpenter/blob/75826eb51589e546fffb594bfefa91f3850e6c82/pkg/utils/pod/scheduling.go#L122 and https://github.com/kubernetes/kubernetes/blob/84a7f1ea15e4fcabe88e06ca27de611fa3b74437/pkg/api/v1/pod/util.go#L312

**How important is this feature to you?**

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #1298, opened 2024-06-05
chore: replace custom utils functions with k8s.io/kubernetes library

fmuyassarov:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1275

**Description**
Replace custom utils function `IsTerminal()` with k8s.io/kubernetes library function `IsPodTerminal()`.

```
make install-kwok
make apply
```


fmuyassarov:
@jonathan-innis @njtran I noticed that there is a common issue with the govulncheck failing due to some vulnerabilities found in the netip and I fixed it [here](https://github.com/kubernetes-sigs/karpenter/pull/1298/commits/2e8d3a0ee4088e657b97dfa9989ba623e3fc31d0) so that the patch doesn't fail required checks.

njtran:
@fmuyassarov go vuln should be fixed now

njtran (commented):
/lgtm
/approve

njtran:
Thanks for the commit!

jonathan-innis:
@fmuyassarov I think this change may have broken dependency resolution when trying to build with this latest version now. We can't directly use the upstream `kubernetes/kubernetes` go module because it references a bunch of local modules with pseudo-versions (v0.0.0) and then does a bunch of [replace statements at the bottom](https://github.com/kubernetes/kubernetes/blob/9e2075b3c87061d25759b0ad112266f03601afd8/go.mod#L227). Any downstream packages that consume this directly would also have to do replace statements, which is the reason why k8s went out and created a bunch of distinct packages that they publish separately (conroller-runtime, apimachinery, etc)

I did a bit of searching around but I'm not sure there's a utils package that exists in upstream right now that contains this function. Maybe @dims or @sftim has more insight if there's some work that we could do here to move these kinds of utils functions into a distinct package.

fmuyassarov:
Sorry that I didn't notice the possible issue with the dependency resolution. Probably because I'm using vscode.
If we need this to be fixed asap for developers, I'm fine to revert that commit for the time being.

dims:
+1 to revert! please avoid depending on `k8s.io/kubernetes` ( documented here - https://github.com/kubernetes/kubernetes?tab=readme-ov-file#to-start-using-k8s )

If you do need some things from there, please open an issue and propose a good landing spot in one of the staging repositories:
https://github.com/kubernetes/kubernetes/tree/master/staging#external-repository-staging-area

fmuyassarov:
> +1 to revert! please avoid depending on `k8s.io/kubernetes` ( documented here - https://github.com/kubernetes/kubernetes?tab=readme-ov-file#to-start-using-k8s )
> 
@jonathan-innis I have filed a revert patch in https://github.com/kubernetes-sigs/karpenter/pull/1302. PTAL

PR #1301, opened 2024-06-05
test: Try pinning a lower version in go.mod

jonathan-innis:
Fixes #N/A 

**Description**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/karpenter
 
-go 1.22.4
+go 1.22.2
 
 require (
 	github.com/Pallinder/go-randomdata v1.2.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/karpenter
 
-go 1.22.4
+go 1.22.2
```
fmuyassarov, line 3: Would you mind to share why?

jonathan-innis: Just a test attempt. Some downstream users of this package were running into some painpoints around upgrading to recently released go versions, so we were trying to see if this is possible. I'm not fully convinced that we should do this yet, though.

jonathan-innis:
Closing since this was just to validate that this works with our vulnchecks.

go.mod
fmuyassarov: thanks for sharing

Issue #1047, opened 2024-02-27
Allow Forced Disruption of Nodes after TTL (`disruptionTolerationDuration`)


levinedaniel:
### Description

**What problem are you trying to solve?**
Currently we are advising teams that submit workloads to our cluster to use the 'do-not-drift' label within pods to allow for ML workloads to finish uninterrupted. Some of these workloads can stay indefinitely on the cluster. As cluster admins, It would be ideal if we could force remove nodes that pass a certain threshold of time past the time the nodeclaim is drifted. The longer the node stays out, we more chance we can accumulate vulnerabilities from the infrastructure.

We can implement someone similar i.e. deploying a cronjob that takes the action specified above, however I believe it would make sense to have this functionality native to Karpenter as it is something that any cluster admin could run into.

I would be interested in working on this issue should the idea gets approved.

**How important is this feature to you?**
This feature would result in a decrease of the SRE team it takes to maintain clusters. As such, it is fairly important to use before we wide-spread adopt karpenter.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Bryce-Soghigian:
Could you leverage [these controls](https://github.com/kubernetes-sigs/karpenter/blob/59c79ee552198a78e4f7133eebba21c0426a6631/designs/disruption-controls-by-reason.md#approach-a-extending-the-budget-api-to-specify-a-reason) to have windows where you allow drift, and windows where you dont? Every weekend for example you allow drift and during weekdays you dont?

levinedaniel:
@Bryce-Soghigian the primary issue with just using disruption controls is that we would ideally like to start the timer as the workload comes in or the node is drifted to give our downstream users an expectation for how long they will have for their deployments. 

Given this scenario:
- User deploys workload Thursday 
- We update amis Friday

we wouldn’t want their nodes to re-provision that soon if we want to give them an allocation of a week for their model to run. 

We could delay updating the ami but in our situation we would like to just specify ami tags and let Ami’s stream in as they are created. 

If we specify disruption gates to just re-provision on weekends we also risk leaving instances running for the better part of a week when they are not in use.

[...]

PR #1303, opened 2024-06-06
docs: add RFC on making expiration forceful

njtran:
Fixes #N/A 

**Description**
This proposes making Expiration forceful, allowing cluster admins to have a guaranteed way to understand the maximum node lifetime of the node. 


njtran:
Hey all, I figure this is a long shot, but as you all were up-votes on https://github.com/aws/karpenter-provider-aws/issues/1884, I wanted to see if any of you had thoughts on this RFC, and know the use-cases for Graceful Expiration.

@andrebikeev @agrrh @karma-git @funes79 @FernandoMiguel @duboisf @diranged @creosonic @gbvanrenswoude @universam1 @philslab-ninja @round-corgi @mmukarram @grandich

FernandoMiguel:
Thanks for the ping.
I'll do my best to take a look tomorrow.

designs/forceful-expiration.md
```diff
@@ -0,0 +1,64 @@
...
+
+**T3 (S3 → S4):** Karpenter waits for necessary capacity in the cluster to be healthy and initialized. Once capacity is available, Karpenter will `kubectl delete` the node(s), tainting the node, beginning graceful termination.
+
+**S4:** Karpenter adds a `NoSchedule` taint and adds an Termination Grace Period Timestamp on the NodeClaim, indicating when Karpenter should transition to Forceful Termination of the node. Karpenter begins draining the node, _respecting PDBs and `do-not-disrupt` pods_. 
+
+***S4** is a point of no return. The node will be terminated within its TGP.* 
+
+**T4 (S4 → S5):** Once the node has no running drainable pods or the TGP timeout is reached, Karpenter directly terminates the underlying instance, initiating forceful termination and cleanup of daemons and networking applications.
```
GnatorX, line 28: This will be the same mechanism as https://github.com/kubernetes-sigs/karpenter/issues/1047?

designs/forceful-expiration.md
```diff
@@ -0,0 +1,64 @@
...
+
+## 🔑 Adding Forceful Expiration
+
+### Option 1 - Make Expiration Forceful - Recommended
+
+Revert Expiration’s semantic to be **Forceful**, bypassing past NDBs and `do-not-disrupt`, starting graceful termination as soon as nodes are detected as expired. This will still respect PDBs and `do-not-disrupt` pods while draining, and can be timed out when the user sets a `nodePool.TerminationGracePeriod`.
+
+**Pro:** This natively addresses the cluster administrator's use-case without any user configuration. Making this change is easier to reason about before v1, and harder to do after v1. Adding in Graceful Expiration after v1 is easier, as it's less disruptive. 
+
+**Con:** This removes waiting for pre-spun capacity, increasing the chances that pods aren’t able to reschedule when nodes expire, and is a breaking change.
```
GnatorX, line 48: I believe you mentioned this during working group but during termination Karpenter would remove the terminating node capacity when calculating how much capacity is needed and possibly start creating nodes?

I think the difference now is there is less time for new node to come up and we don't guarantee the new node is ready before we terminate?

GnatorX: One other issue is if you have a broken EC2NodeClass, such as bad AMI or bad config, you will lose capacity as node expires where before it didn't allow this scenario.

designs/forceful-expiration.md
```diff
@@ -0,0 +1,64 @@
+# Forceful Expiration - v0.37+
```
GnatorX, line 1: Would calling Graceful "Launch before Terminate" and Forceful as "Terminate and launch" make more sense? Both flow is still graceful they just order the node creation differently

designs/forceful-expiration.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Option 1 - Make Expiration Forceful - Recommended
+
+Revert Expiration’s semantic to be **Forceful**, bypassing past NDBs and `do-not-disrupt`, starting graceful termination as soon as nodes are detected as expired. This will still respect PDBs and `do-not-disrupt` pods while draining, and can be timed out when the user sets a `nodePool.TerminationGracePeriod`.
+
+**Pro:** This natively addresses the cluster administrator's use-case without any user configuration. Making this change is easier to reason about before v1, and harder to do after v1. Adding in Graceful Expiration after v1 is easier, as it's less disruptive. 
+
+**Con:** This removes waiting for pre-spun capacity, increasing the chances that pods aren’t able to reschedule when nodes expire, and is a breaking change.
+
+### Option 2 - Allow Forceful and Graceful Expiration
```
GnatorX, line 50: Sidenote: I think this is a useful feature for users for all disruption methods because some users may prioritize saving vs maintaining high level of availability.

designs/forceful-expiration.md
FernandoMiguel, line 28: there seems to be a design issue here.
when using controllers like ebs csi plugin, which work as daemonset, if you drain it before removing the EBS volumes, the volumes wont be detached.

designs/forceful-expiration.md
```diff
@@ -0,0 +1,64 @@
...
+
+Karpenter’s original Expiration semantic was **Forceful**. It was changed to **Graceful** in [this PR](https://github.com/kubernetes-sigs/karpenter/pull/59) so that Karpenter could orchestrate replacement capacity for expired nodes, grouping Expiration with Drift, Underutilization, and Emptiness in **T1**. Users with blocking NDBs, PDBs, or `do-not-disrupt` can prevent expiration from starting, preventing Karpenter from kicking off its TerminationGracePeriod. 
+
+The only difference in the diagram below is Expiration is moved to **T1 (S1 → S2)**.
+
+![](./images/graceful-expiration.png)
+
+## 🔑 Adding Forceful Expiration
+
+### Option 1 - Make Expiration Forceful - Recommended
```
FernandoMiguel, line 42: I like option 1

designs/forceful-expiration.md
GnatorX: Wouldn't deletion of pods that asked for the volumes kick off the CSI del to remove those volumes?

designs/forceful-expiration.md
GnatorX: However likely not worth opening up if no one is asking for this

designs/forceful-expiration.md
njtran: Nope, this is just Termination Grace Period, or drain timeout.

designs/forceful-expiration.md
njtran: As soon as the node is terminated, the provisioning controller will trigger and start provisioning nodes for the pods that will be deleted soon. You're right we won't wait to guarantee that the node is ready before we terminate. This is the inherent tradeoff here. Availability (of the pods) vs Compliance.

GnatorX: Ah terminated. I was thinking terminat`ing`. Thanks

designs/forceful-expiration.md
njtran: I think there's value in having one-word terms to map to the phrase, but I do agree those are pretty clear. There's another mechanism in between called Eventual that's not called out here that is harder to explain in those terms as well.

designs/forceful-expiration.md
njtran: Would you expect users to want this for Consolidation?

GnatorX: I was thinking certain Drift conditions. This is just a thought I am not sure if anyone is looking for this.

designs/forceful-expiration.md
GnatorX: Sorry if I am missing something the Termination Grace Period I know about, https://github.com/kubernetes-sigs/karpenter/pull/916, isn't merged yet right? Is there some other termination or drain timeout?

GnatorX (approved):
Overall lgtm. I only have questions for myself

designs/forceful-expiration.md
njtran: This is talking aspirationally, not about the current behavior, so yes, it's talking about #916

njtran:
For those I've pinged, hoping to close out the RFC this week. Please let me know if you have questions.

designs/forceful-expiration.md
jmdeal: If the `ebs-csi-node` pod tolerates Karpenter's disruption taint and is drained early, this could result in volumes still being attached. This shouldn't happen in this specific case since it has a preStop hook that waits for all volumes to be detached, though this can still be problematic if its terminationGracePeriodSeconds is exceeded before all volumes are detached. This should also apply once the instance is terminating if graceful node termination has been configured (though there's currently an issue since Karpenter removes the Node object before the instance is terminated).

designs/forceful-expiration.md
camaeel: For me this option is better. If I could define per NodePool either forcefulExpiration, gracefullExpiration, none of those or both would be ideal situation. 
In some cases (business critical applications, stateful applications, databases, etc) cluster admin can have strong limitations on when they could be patched - for those gracefulExpiration should be useful. 

Ideal connection would be to have:
* gracefulExpiration + NodeDisruptionBudgets set up for business critical applications as mentioned before. 
* for some specific use cases - none of those could be defined so the cluster administrator will have to provide on its own means of patching those nodes
* gracefulExpiration + NodeDisruptionBudgets + forcefulExpiration - would make sense to try rotating such node after gracefulExpiration, and let it run to maximum of forcefulExpiration. 
* only forcefulExpiration - for environments with strong requirements for periodic rotation of notes

designs/forceful-expiration.md
FernandoMiguel: this is a current issue. blocking many Orgs.
https://github.com/kubernetes-sigs/aws-ebs-csi-driver/issues/1955#issue-2165787055
So whatever we design here needs to take in consideration who currently Karpenter works and plan for the future, or I fear we may make the issue worst by assuming certain things

designs/forceful-expiration.md
njtran: > gracefulExpiration + NodeDisruptionBudgets set up for business critical applications as mentioned before.

I want to understand this use case more @camaeel. For forceful expiration, these business critical applications wouldn't be disrupted if you had PDBs and do-not-disrupt properly set-up. The only case where you would see these applications disrupted with those safeguards in is if you had set TerminationGracePeriod (drain timeout) as implemented [here](https://github.com/kubernetes-sigs/karpenter/pull/916), which is in the control of the administrator.

camaeel: @njtran In a perfect world PDB and proper configuration should solve it. 
But let's say we have not so perfect applications. It is always better to rotate nodes/pods off business hours and not in the highest demand. Application even if gracefully restarted might have degraded performance just after the start (caches need to be filled up or if 1 pod out of 5 is down we loose 20% of capacity in crucial moment). Also some statefull workloads like databases or kafka, even if available, might cause some hiccups during switching (we observed that with one or two applications with high traffic utilizing kafka). Or you can have a legacy application that was barely moved to the "cloud age" and has low tolerance for disruptions or has simply single points of failure. Or your organization can simply have strict rules that all the patchings have to be taken out of business hours. 
IMHO it would be best if administrators are given possibility to set any combination of gracefulExpiration and forcefulExpiration for each node pool, so they will have full flexibility to configure according to what is really needed for each of them. Since there are still some cases to be handled gracefully (drifted, udnerutilzied,empty) and some for forcefully (interrupted, unhealthy) both branches in the code have to be supported so not that much extra code/effort should be required to implement both.

njtran: Got it. If there was a config for both graceful and forceful, what would you use in your NodePool? @camaeel

camaeel: @njtran At the end I think for most of the pools I would use graceful expiration.

njtran: Do you mind me asking what you're currently using expiration for?

camaeel: @njtran My k8s clusters are not yet using Karpenter. I did some POCs/tests, but it is still not yet deployed and in use.
My main use case for expiration is to "rotate nodes that are older than X". This is a security requirement to patch OS every few days. And it is easier to achieve by replacing the nodes with fresh ones than by letting them run very long and patch and restart everything (immutable infrastructure). I should run those node replacements durimg specufic times of day/week.
When I was working on POC I ended up by implementing a custom automation job to replace nodes. I think back then (~1y ago) expiration was limited/working different.

garvinp-stripe: I am sorry but I don't fully grok how this use case requires "graceful" expiration. All termination of nodes are still graceful in that pods are not forced killed unless specified by the user, which is a feature outside of this RFC. The difference is two folds between "graceful" and "forceful" described in this RFC :
1) Nodes aren't created before expired nodes are terminated. This means a longer time to reschedule pods, because new nodes need be provisioned before terminated pods land on it.
2) No guarantee that new nodes can be created. This means there is a possibility that pods are stuck in pending.

Nothing mentioned by @camaeel talks about these two points, which I think is partially why the naming should be changed. 

One last thing. The mention of controlling when expiration or disruption happens is in another RFC https://github.com/kubernetes-sigs/karpenter/pull/991

garvinp-stripe: @njtran I think the doc goes in depth on explaining the mechanism of what happens but we need a high level summary of what "forceful" expiration actually means for users and what the trade offs are at a much higher level

camaeel: > This is what I'm confused about. Is this security requirement really a requirement to you? If you have graceful expiration and fail to set your node disruption budgets properly, you may not meet security requirements (security becomes a preference). If you have forceful expiration and fail to set your pod disruption budgets properly, you may get unwanted app disruptions (uptime becomes a preference).

I think it is still requirement for my infrastructure. I'd still use graceful termination and handle misconfigurations/lack of resources/any other in the expiration procedure by using monitoring that will alert clsuter adminsitrator that something is wrong and cluster can't do it automatically. 
I still have another requirement for doing planned node rotations out of business hours. That's why I would rather handle it this way. 

> 
> With Graceful Termination, the cluster admin has to jump through multiple hoops to guarantee a security requirement. You need to orchestrate and ensure:
> 
>     1. No blocking PDBs or `do-not-disrupt` pods are scheduled
PDB's should be fine, but those are managed by development teams, so this is possible source of problems. Just as the pods. 

> 
>     2. No blocking NDBs or `do-not-disrupt` nodes are set

Those are fully under cluster admin control so shouldn't happen without my knowledge.

> 
>     3. Replacement capacity can be spun-up for the pods on the node
IMHO this can be monitored if my clusters have less than 10-20 % space for extension, and a 

> 
> 
> With Forceful Termination, the cluster admin can guarantee the requirement just by setting `expireAfter + terminationGracePeriod = maxNodeLifetime`, and the app team communicates to you, the cluster admin, how long to set the terminationGracePeriod to be compatible with their required cleanup times, which more closely models how I think about the ownership model of nodes/pods.

If I understand correctly this is how "gracefulTermination" works, right?

njtran: > PDB's should be fine, but those are managed by development teams, so this is possible source of problems. Just as the pods.

I've heard feedback that especially in large clusters, ensuring that these are set properly is tough. One option is to apply back-pressure and enforce to the app teams to properly model their allowed disruption.

> IMHO this can be monitored if my clusters have less than 10-20 % space for extension, and a

More than just limits, this also applies to insufficient capacity failures, image bootstrapping issues, and launch failures, which can be ephemeral in nature. 

> If I understand correctly this is how "gracefulTermination" works, right?

This may be where the details might have gotten lost. Making Expiration forceful would initiate graceful termination (draining) of a node (state 4 in forceful-expiration.png) as soon as the node is expired.  If expiration were to be graceful, any of the three reasons I mentioned above could indefinitely delay the initiation of graceful termination. This is meant to reduce the operational load and burden for the cluster administrator by reducing the number of exceptions that they have to consider.

garvinp-stripe: > If I understand correctly this is how "gracefulTermination" works, right?

This is how it works for both graceful and forceful termination. The difference is if a replacement node is created first or after.

njtran: After our discussion here, I'm not convinced we need to allow set a graceful termination at v1. This is a two-way door, so we can always add this in later.

pkit:
Forceful expiration that ignores all budgets **will** lead to downtime.
No matter how you design your HA it may disrupt **all** the nodes your HA is using right now.
None of the options karpenter has right now will save you from this.
Moreover, things like `do-not-disrupt` annotation will just exacerbate the problem: it will wait as long as possible (for example 720h+48h grace) and then still kill everything leading to downtime.
It's even worse, if gateway API is used. It will disconnect the LB from the "shutting down" node, and then you have a node with NoSchedule taint where services run happily (`do-not-disrupt`), but everything is down because they don't receive any traffic anymore. 
Nodes that scale all at the same time cannot be avoided too, due to peak traffic usually happening as "point-in-time" and all services getting scaled when the spike happens.

Essentially at the current state karpenter is unusable for any use case where `Deployment` and `StatefulSet` are needed:
 - if `<1.0` it has the EBS attach problem, where EBS volumes will wait for 6 minutes after pod is moved to reattach
 - if `>=1.0` it has this forced expiration which will lead to downtime no matter what

camaeel:
@pkit I also don't like the way expiry after X hours works. This will ignore at the end both PDB, do-not-disrupt or disruption budgets of node pools. So I can't reliably control when it happens. My idea is to give up this feature and simply build new AMI and then use drift detection to rotate/replace/patch nodes.

PR #1304, opened 2024-06-06
fix: allow single character status condition reason

jmdeal:
Fixes #N/A 

**Description**
This updates the regex validation for the status condition reason on `NodeClaims`. A previous PR (https://github.com/kubernetes-sigs/karpenter/pull/1285) updated this regex to allow empty strings and unblock migration from previous versions to v0.37. It also inadvertently changed the regex such that if a string is provided it must be at least two characters. This doesn't have any immediate impact since none of the status conditions are a single character, but this PR corrects the regex.

Manual validation

### DIFF
--- a/hack/validation/status.sh
+++ b/hack/validation/status.sh
@@ -1,5 +1,5 @@
 # Updating the set of required items in our status conditions so that we support older versions of the condition
 # TODO: This can be removed once we upgrade to v1 and can do conversion to automatically set values for old versions of the condition
 yq eval 'del(.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.conditions.items.properties.reason.minLength)' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
-yq eval '.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.conditions.items.properties.reason.pattern = "(^([A-Za-z][A-Za-z0-9_,:]*[A-Za-z0-9_])?$)"' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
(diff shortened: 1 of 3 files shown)


hack/validation/status.sh
```diff
@@ -1,5 +1,5 @@
 # Updating the set of required items in our status conditions so that we support older versions of the condition
 # TODO: This can be removed once we upgrade to v1 and can do conversion to automatically set values for old versions of the condition
 yq eval 'del(.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.conditions.items.properties.reason.minLength)' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
-yq eval '.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.conditions.items.properties.reason.pattern = "(^([A-Za-z][A-Za-z0-9_,:]*[A-Za-z0-9_])?$)"' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+yq eval '.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.conditions.items.properties.reason.pattern = "^([A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?|)$"' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
```
jonathan-innis, line 4:
```suggestion
yq eval '.spec.versions[0].schema.openAPIV3Schema.properties.status.properties.conditions.items.properties.reason.pattern = "^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?|$"' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
```

Is this valid regex to be able to drop the the matching group parenthesis?

jmdeal: No, that scopes the start and end line anchors to either side of the or. For example, a value with a colon at the end would pass the regex since there's no end of line anchor.

jonathan-innis (approved):
/lgtm
/approve

Issue #729, opened 2023-02-09
karpenter doesn't scaleout when using time-slicing for GPUs


vickywh:
### Tell us about your request

Karpenter should respect time-slicing configuration when provisioning gpu instances 

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

When you use time slicing and the NVIDIA/k8s-device-plugin, Karpenter can fail to provision new instances 

As per the documentation here: https://github.com/NVIDIA/k8s-device-plugin#shared-access-to-gpus-with-cuda-time-slicing

To implement time-slicing you would configure: 

version: v1
sharing:
  timeSlicing:
    renameByDefault: <bool>
    failRequestsGreaterThanOne: <bool>
    resources:
    - name: < resource-name >
      replicas: < num-replicas >
 
the outcome of this being that if you did $ kubectl describe node you’d see

Capacity:
  nvidia.com/gpu: < num-replicas >

Karpenter respects that when scheduling pods, for example 6 pods which each request 4 GPU will be placed onto a node when num-replicas is 24. 
 
However, when it comes to scaling, the following error occurs:

 {"memory":"4Gi","[nvidia.com/gpu](http://nvidia.com/gpu)":"4","pods":"1"}, 

because the deployment is requesting
 
          resources:
            limits:
              [nvidia.com/gpu:](http://nvidia.com/gpu:) '4'
            requests:
              memory: 4Gi
              [nvidia.com/gpu:](http://nvidia.com/gpu:) '4'
 
but the instance types in the Provisioner only have 1 GPU:
 
spec:
  requirements:
  - key: [node.kubernetes.io/instance-type](http://node.kubernetes.io/instance-type)
    operator: In
    values:
    - g5.2xlarge
    - g5.4xlarge
 
When evaluating which instance to provision, Karpenter should evaluate the number of GPUs specified in < num-replicas >, not the number of GPUs that instance type/size has, and therefore provision a new instance (in this example)

### Are you currently working around this issue?

No

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

atulghodmare777:
[...]

PR #1305, opened 2024-06-06
RFC: NodeOverlay

ellistarn:
Fixes:
* https://github.com/aws/karpenter-provider-aws/issues/3860
* https://github.com/aws/karpenter-provider-aws/pull/4697
* https://github.com/kubernetes-sigs/karpenter/issues/751
* https://github.com/kubernetes-sigs/karpenter/issues/729
* https://github.com/aws/karpenter-provider-aws/issues/5161

**Description**
```
---
# Reduce on-demand prices to 90%
# https://github.com/aws/karpenter-provider-aws/issues/3860
# https://github.com/aws/karpenter-provider-aws/pull/4697
kind: NodeOverlay
metadata:
  name: discount
spec:
  selector:
    matchLabels:
      karpenter.sh/capacity-type: on-demand
  pricePercent: 90
---
# Support for extended resource types (e.g. smarter-devices/fuse)
# https://github.com/kubernetes-sigs/karpenter/issues/751
# https://github.com/kubernetes-sigs/karpenter/issues/729
kind: NodeOverlay
metadata:
  name: discount
spec:
  selector:
    matchLabels:
      karpenter.sh/capacity-type: on-demand
    matchExpressions:
      operator: In
      values:
      - m5.large
      - m5.2xlarge
      - m5.4xlarge
      - m5.8xlarge
      - m5.12xlarge
  capacity:
    smarter-devices/fuse: 1
---
# Add memory overhead of 10Mi to all instances with 2Gi memory
# https://github.com/aws/karpenter-provider-aws/issues/5161
kind: NodeOverlay
metadata:
  name: discount
spec:
  selector:
    matchLabels:
      karpenter.k8s.aws/instance-memory: 2048
  overhead:
    memory: 10Mi
---
```


sftim:
(nit) In the PR description, the NodeOverlays all have the same `.metadata.name`

sftim:
Is this an overlay for Nodes, or an overlay for NodePools (maybe more than one NodePool, if we support matching via a selector)?

daimaxiaxie:
@ellistarn We redefined the value of `topology.kubernetes.io/zone`. However, the current zone value is provided by the CloudProvider. Is it possible to support custom `topology.kubernetes.io/zone`  and other WellKnownLabel values?

sftim:
> @ellistarn We redefined the value of `topology.kubernetes.io/zone`. However, the current zone value is provided by the CloudProvider. Is it possible to support custom `topology.kubernetes.io/zone` and other WellKnownLabel values?

It sounds like you don't agree with the zone naming used in an existing integration. You can:
- use a different label
  - this is the easy and recommended option
- replace the cloud provider integration code with your own
  (Karpenter, cloud controller manager, etc)

I don't think this PR would need to change to accommodate the ask.

GnatorX:
Thanks for trying to tackle this!

I wonder if there is a cleaner way to represent the overlays to reduce sprawl.

Thinking more generally, each dimension can be additive, adding a new custom resource, and/or adjustment, price changes or resource modification. (Assuming we aren't supporting deleting properties at this time).

Price is a slight outlier from resources so we can separate out. Would it make sense to have at the highest level price and resources. Then subsection of adjustment and additions. Adjustment would support +/- percent or values and additions would just be a dict of new custom resources and values?

daimaxiaxie:
@sftim I mean nodeoverlay can provide more overlay options, not just resources. For example, Lables.

ellistarn:
Hi folks. Please comment inline in the docs w/ threads, so I can respond directly.

ellistarn:
For those watching. I just pushed a rev with an RFC document.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+  instanceTypes:
+    - m5.large
+	- ...
+```
+
+Exposing the full details of the mutations caused by NodeOverlays cannot be feasibly stored in ETCD. Further, we do not want to require that customers read the Karpenter logs to debug this feature. When GetInstanceTypes is called per NodePool, we will emit an event on the NodePool to surface deeper semantic information about the effects of the NodeOverlays. Our implementation will need to be conservative on the volume and frequency of information, but provide sufficient insight for customers to verify that their configurations are working as expected.
+
+### Per-Field Semantics
+
+The semantics of each field carry individual nuance. Price could be modeled directly via a `price` field, but we can significantly reduce the number of NodeOverlays that need to be defined if we instead model the semantic as a function of the currently defined price (e.g. `pricePercent` or `priceAdjustment`). Similarly, `capacity` and `overhead` make sense to be modeled directly, but `capacityPercent` makes much less sense than `overheadPercent`.
```
GnatorX, line 180: Is it possible to lay out what you are thinking the API looks like for each field? I am not sure I grok what we will support. 
Are we supporting something like:
```
overhead:
  - memory: 10%
```

Second I think the names can be confusing at the moment, maybe its just me.

For example, if I look just a pricePercent does that mean percent of discount or percent of price? 
For overhead its a similar story, is this overhead the amount I am adding to a resource or overhead that I need to subtract from the resource itself.

Lastly, if I specify a capacity that exist within the scheduling simulation for an instance type but its incorrect, such as huge pages, do I override that value or can I even do that? If not can I specify a negative overhead to add value to an existing resource?

GnatorX: Got it. I realize now I totally missed that section right under.

I think I understand the pricePercent a bit better now so I am happy with keeping that name. Other names I was thinking about was `priceAdjustmentPercent` which indicates you are adjusting the price based on input % rather than a percent of price. Happy to either way

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,72 @@
...
+	// +optional
+	PricePercent *int `json:"pricePercent,omitempty"`
+
+	//
+	// The following fields are not yet implemented
+	//
+
+	// PriceAdjustment modifies the price of the simulated node (PriceAdjustment + (Price * PricePercent / 100)).
+	// +optional
+	// PriceAdjustment *int `json:"priceAdjustment,omitempty"`
```
jwcesign, line 56: In which situation we need this field?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+Each instance offering comes with a price, which is used as a minimization function in Karpenter's scheduling algorithm. However, Karpenter cannot perfectly understand all factors that influence price, such as an external vendor's fee, a pricing discount, or an external motiviation such as carbon-offset.
+
+* https://github.com/aws/karpenter-provider-aws/pull/4686
+* https://github.com/aws/karpenter-provider-aws/issues/3860
+* https://github.com/aws/karpenter-provider-aws/pull/4697
+
+### Extended Resources
+
+InstanceTypes have a set of well known resource capacities, such as CPU, memory, or GPUs. However, Kubernetes supports arbitrary [extended resources](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/), which cannot be known prior to Node launch. Karpenter lacks a mechanism to teach the simulation about extended resource capacities.
```
jonathan-innis, line 17: I'm assuming that we would still support mechanisms for doing first-class support of a bunch of extended resources so that we don't have to be asking users to specify NodeOverlays for all of them?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+* https://github.com/kubernetes-sigs/karpenter/issues/751
+* https://github.com/kubernetes-sigs/karpenter/issues/729
+
+### Resource Overhead
```
jonathan-innis, line 22: I'd like to see if we can do a better job being more accurate on these. Like extended resource support, I'd like it if we didn't have to require users to have to set all of these but could just do the right thing here where possible

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+* https://github.com/aws/karpenter-provider-aws/issues/5161
+
+System software such as containerd, kubelet, and the host operating system come with resource overhead like CPU and memory, which is substracted from a Node's capacity. Karpenter attempts to calculate this overhead where it is known, but cannot predict the overhead required by custom operating systems. To complicate this, the overhead often changes for each version of an operating system.
+
+## Proposal
+
+NodeOverlays are a new Karpenter API that enable customers to fine-tune the scheduling simulation for advanced use cases. NodeOverlays can be configured with a `metav1.Selector` which flexibly constrain when they are applied to a simulation. Selectors can match well-known labels like `node.kubernetes.io/instance-type` or `karpenter.sh/node-pool` and custom labels defined in a NodePool's `.spec.labels` or `.spec.requirements`. When multiple NodeOverlays match an offering for a given simulation, they are merged together using `spec.weight` to resolve conflicts.
```
njtran, line 30: As they're NodeOverlays, does it make sense to say it just references the same as a pod NodeSelector? 

Additionally, if I'm a user, and I'm wondering if my NodeOverlays are right, will these match the node labels or the nodeclaim labels? Realistically it shouldn't matter

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+After two years of additional customer feedback in this problem space, this proposal asserts that the flexibility of a selector-based approach is a signficant simplification for customers compared to a 1:1 approach and avoids the scalability challenges associated with persisting large numbers of CRs in the Kubernetes API. This decision will require us to work through the associated observability challenges.
+
+### Merging
+
+Our decision to pursue Selectors that allow for an n:m semantic presents a challenge where multiple NodeOverlays could match a simulation, but with different values. To address this, we will support an optional `.spec.weight` parameter (discussed in the [previous proposal](https://github.com/aws/karpenter-provider-aws/pull/2390/files#diff-b7121b2d822e57b70203794f3b6367e7173b16d84070c3712cd2f15b8dbd11b2R153)).
+
+We have a few choices for dealing with conflicting NodeOverlays
+
+1. Multiple NodeOverlays are merged in order of weight
+2. Multiple NodeOverlays are applied sequentially in order of weight
```
jonathan-innis, line 104: This option strikes me as more confusing in a lot of ways and doesn't really adhere to an "overlay" construct, where a user is probably thinking about this in terms of layers

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+```yaml
+kind: NodeOverlay
+metadata:
+  name: extended-resource
+spec:
+  selector:
+    matchLabels:
+      karpenter.sh/capacity-type: on-demand
+    matchExpressions:
```
njtran, line 55: Do we want to do validation on well known labels? Or should this also apply to custom labels? I can't think of any reason it shouldn't apply to all custom labels.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+    - key: node.kubernetes.io/instance-type
+      operator: In
+      values: ["m5.large", "m5.2xlarge", "m5.4xlarge", "m5.8xlarge", "m5.12xlarge"]
+  capacity:
+    smarter-devices/fuse: 1
+```
+
+Add memory overhead of of 10Mi and 50Mi for all instances with more than 2Gi memory
```
njtran, line 63: Does this replace `vmOverheadMemoryPercent` or just lay on top of it?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+      karpenter.sh/capacity-type: on-demand
+    matchExpressions:
+    - key: node.kubernetes.io/instance-type
+      operator: In
+      values: ["m5.large", "m5.2xlarge", "m5.4xlarge", "m5.8xlarge", "m5.12xlarge"]
+  capacity:
+    smarter-devices/fuse: 1
+```
+
+Option 1 enables us to be strictly more flexible than option 3. However, it introduces ambiguity when multiple NodeOverlays have the same weight. Option 3 would only allow one to apply, so any conflict must be a misconfiguration. However, our example demonstrates that Option 1 has valid use cases where multiple NodeOverlays share the same weight, as long as they don't define the same value. For Option 1, NodeOverlays with the same weight that both attempt to configure the same field are considered to be a misconfiguration.
```
jonathan-innis, line 149: Does it just apply to the things that are being selected on or the whole set of instance types? Sounds like the former. So basically -- it only applies to misconfiguration errors on the intersection of an invalid set of selectors

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Misconfiguration
+
+Given our merge semantics, it's possible for customers to define multiple NodeOverlays with undefined behavior. In both cases, we will identify and communicate these misconfigurations, but should we fail open or fail closed? Failing closed means that a misconfigured would halt Karpenter from launching new Nodes until the misconfiguration is resolved. Failing open means that Karpenter might launch capacity that differs from what an operator expects, and result in thrashing scale in/outs.
+
+We choose to fail closed to prevent undefined behavior that could incur costs. However, we have the option to fail closed for the entire provisioning loop, or for a single NodePool. For example, if multiple NodeOverlays match NodePoolA with the same weight, but only a single NodeOverlay matches NodePoolB, Karpenter could exclude NodePoolA as invalid, but proceed with NodePoolB for simulation. These semantics are very similar to how Karpenter deals with `.spec.limits` or `InsufficientCapacityErrors`. See: https://karpenter.sh/docs/concepts/scheduling/#fallback.
+
+### Observability
```
jonathan-innis, line 157: What are our thoughts around performance and memory impact of this change?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+After two years of additional customer feedback in this problem space, this proposal asserts that the flexibility of a selector-based approach is a signficant simplification for customers compared to a 1:1 approach and avoids the scalability challenges associated with persisting large numbers of CRs in the Kubernetes API. This decision will require us to work through the associated observability challenges.
+
+### Merging
+
+Our decision to pursue Selectors that allow for an n:m semantic presents a challenge where multiple NodeOverlays could match a simulation, but with different values. To address this, we will support an optional `.spec.weight` parameter (discussed in the [previous proposal](https://github.com/aws/karpenter-provider-aws/pull/2390/files#diff-b7121b2d822e57b70203794f3b6367e7173b16d84070c3712cd2f15b8dbd11b2R153)).
```
njtran, line 99: I'm thinking about these overlays as a combination of mathematical operations on a subset of instance types for a subset of attributes. Is it possible to just arrange these mathematical operations in a way like PEMDAS such that we don't have to think about weight and can just merge them?  I think this would be option 2.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+How can users be aware of how their NodeOverlays have applied to their NodePools? Following from our decision above, we will model the application of NodeOverlays at the NodePool level. This leads us to observe the behavior of NodeOverlays in the status of the NodePool.
+
+For example:
+```
+kind: NodePool
+status:
+  conditions:
+    - type: InstanceTypesReady
+      status: True
+  nodeOverlays:
```
jonathan-innis, line 168: How convinced are we that we want to surface the applied nodeOverlays in the NodePool in this way?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+    - default
+    - extended-resources
+  instanceTypes:
```
jonathan-innis, line 171: I think the instance types is a good add and something that I would be interested in us adding

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+Our decision to pursue Selectors that allow for an n:m semantic presents a challenge where multiple NodeOverlays could match a simulation, but with different values. To address this, we will support an optional `.spec.weight` parameter (discussed in the [previous proposal](https://github.com/aws/karpenter-provider-aws/pull/2390/files#diff-b7121b2d822e57b70203794f3b6367e7173b16d84070c3712cd2f15b8dbd11b2R153)).
+
+We have a few choices for dealing with conflicting NodeOverlays
+
+1. Multiple NodeOverlays are merged in order of weight
+2. Multiple NodeOverlays are applied sequentially in order of weight
+3. One NodeOverlay is selected by weight
+
+We recommend option 1, as it supports the common use case of default + override semantic. In the memory overhead example, both NodeOverlays attempt to define the same parameter, but one is more specific than the other. Option 1 and 3 allow us to support this use case, where option 2 would result in 50Mi+10Mi. We are not aware of use cases that would benefit from sequential application.
```
njtran, line 107: I'm wondering how common we're thinking overlapping nodeoverlays will occur. I could see a sequential merging if users wanted to model multiple pricing discounts as multiple overlays, rather than having to merge them all themselves into a CR.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+      status: True
+  nodeOverlays:
+    - default
+    - extended-resources
+  instanceTypes:
+    - m5.large
+	- ...
+```
+
+Exposing the full details of the mutations caused by NodeOverlays cannot be feasibly stored in ETCD. Further, we do not want to require that customers read the Karpenter logs to debug this feature. When GetInstanceTypes is called per NodePool, we will emit an event on the NodePool to surface deeper semantic information about the effects of the NodeOverlays. Our implementation will need to be conservative on the volume and frequency of information, but provide sufficient insight for customers to verify that their configurations are working as expected.
```
jonathan-innis, line 176: Do we have an example of that event? Would that event just say that we are applying the overlay or would it be more specific about how it's mutating each thing?

jonathan-innis, line 176: Is there anything that we could surface through metrics here that would allow us to also show this information? How does this affect our current instance types metrics and the values that they carry?

designs/node-overlays.md
jonathan-innis, line 180: Is there any prior art for doing arbitrary "equation-based" things? I don't know if we quite need the flexibiliy of CEL here but giving some level of arbitrary flexibility here that we could reason about might be nice

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+We choose option 1 over option 3 to support NodeOverlays that define multiple values. For example, we could collapse the example NodeOverlays using a merge semantic.
+
+```yaml
+kind: NodeOverlay
+metadata:
+  name: default
+spec:
+  pricePercent: 90
+  overhead:
+    memory: 10Mi
```
njtran, line 118: My guess is that price discounts vary by a different dimension than by how overhead will vary. Overhead could vary by instance size, especially if modeled as a percent. If it's flat, it may be properly modeled like this.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+---
+kind: NodeOverlay
+metadata:
+  name: memory
+spec:
+  weight: 1 # required to resolve conflicts with default
+  selector:
+    matchExpressions:
+      key: karpenter.k8s.aws/instance-memory
```
njtran, line 127: nit: using kwok labels in our upstream to not use AWS labels
```suggestion
      key: karpenter.kwok.sh/instance-memory
```

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+  # weight is not required since default does not conflict with extended resources
+  selector:
+    matchLabels:
+      karpenter.sh/capacity-type: on-demand
+    matchExpressions:
+    - key: node.kubernetes.io/instance-type
+      operator: In
+      values: ["m5.large", "m5.2xlarge", "m5.4xlarge", "m5.8xlarge", "m5.12xlarge"]
+  capacity:
+    smarter-devices/fuse: 1
```
njtran, line 146: If there was an overhead here, does it apply to this capacity?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Misconfiguration
+
+Given our merge semantics, it's possible for customers to define multiple NodeOverlays with undefined behavior. In both cases, we will identify and communicate these misconfigurations, but should we fail open or fail closed? Failing closed means that a misconfigured would halt Karpenter from launching new Nodes until the misconfiguration is resolved. Failing open means that Karpenter might launch capacity that differs from what an operator expects, and result in thrashing scale in/outs.
+
+We choose to fail closed to prevent undefined behavior that could incur costs. However, we have the option to fail closed for the entire provisioning loop, or for a single NodePool. For example, if multiple NodeOverlays match NodePoolA with the same weight, but only a single NodeOverlay matches NodePoolB, Karpenter could exclude NodePoolA as invalid, but proceed with NodePoolB for simulation. These semantics are very similar to how Karpenter deals with `.spec.limits` or `InsufficientCapacityErrors`. See: https://karpenter.sh/docs/concepts/scheduling/#fallback.
```
njtran, line 155: Since we're selecting across an arbitrary set of labels, not a set of nodepools, we can't guarantee a NodePool partitioned set of NodeOverlays. If I have an overlay that points to one ICEd instance type of my allowed nodepool, but it's misconfigured, should I prevent that NodePool from being scheduled to?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+kind: NodePool
+status:
+  conditions:
+    - type: InstanceTypesReady
+      status: True
+  nodeOverlays:
+    - default
+    - extended-resources
+  instanceTypes:
+    - m5.large
```
njtran, line 172: Are we adding instance types into the status? I wasn't aware of that.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+### Observability
+
+How can users be aware of how their NodeOverlays have applied to their NodePools? Following from our decision above, we will model the application of NodeOverlays at the NodePool level. This leads us to observe the behavior of NodeOverlays in the status of the NodePool.
+
+For example:
+```
+kind: NodePool
+status:
+  conditions:
+    - type: InstanceTypesReady
```
njtran, line 166: maybe a condition for NodeOverlaysRegistered?

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+      status: True
+  nodeOverlays:
+    - default
+    - extended-resources
```
njtran, line 170: If we're merging nodeOverlays (I realize it's not your recommendation), then we could consider having the merged nodeOverlay here.

designs/node-overlays.md
jmdeal, line 155: Since NodeOverlays don't necessarily apply to all offerings available in a NodePool, do we want to exclude the entire NodePool if, for example, only a single offering had a conflict? Or do we just treat that single offering as unavailable?

kwok/charts/crds/karpenter.sh_nodeoverlays.yaml
```diff
@@ -0,0 +1,85 @@
...
+              More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds
+            type: string
+          metadata:
+            type: object
+          spec:
+            properties:
+              pricePercent:
+                description: PricePercent modifies the price of the simulated node
+                  (PriceAdjustment + (Price * PricePercent / 100)).
+                type: integer
```
sftim, line 45: If we can use a [quantity](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) here, it'll allow adjustments finer than 0.01
Maybe `priceMultiplier` (default value `1`)?

(if we don't, it could end up hard to change that later)

designs/node-overlays.md
njtran: This event strikes me as something that would be extremely verbose to get the right details across. You have to list out all the offerings that are compatible, and what the notion of each overlay is

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,72 @@
...
+	Spec NodeOverlaySpec `json:"spec"`
+}
+
+type NodeOverlaySpec struct {
+	// Selector matches against simulated nodes and modifies their scheduling properties. Matches all if empty.
+	// +optional
+	Selector []v1.NodeSelectorRequirement `json:"selector,omitempty"`
+	// PricePercent modifies the price of the simulated node (PriceAdjustment + (Price * PricePercent / 100)).
+	// +optional
+	PricePercent *int `json:"pricePercent,omitempty"`
```
sftim, line 48: We could make this mandatory with a default - and I recommend that we do. Defaulting to 100% is roughly equivalent to making the field optional, and feels easier to teach.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,72 @@
...
+
+	//
+	// The following fields are not yet implemented
+	//
+
+	// PriceAdjustment modifies the price of the simulated node (PriceAdjustment + (Price * PricePercent / 100)).
```
sftim, line 54: What unit does an integer value of 1 represent here?

sftim: Maybe we need a way to express the unit, eg as `priceCurrency` (borrowing a name from https://schema.org/priceCurrency)

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+NodeOverlays are a new Karpenter API that enable customers to fine-tune the scheduling simulation for advanced use cases. NodeOverlays can be configured with a `metav1.Selector` which flexibly constrain when they are applied to a simulation. Selectors can match well-known labels like `node.kubernetes.io/instance-type` or `karpenter.sh/node-pool` and custom labels defined in a NodePool's `.spec.labels` or `.spec.requirements`. When multiple NodeOverlays match an offering for a given simulation, they are merged together using `spec.weight` to resolve conflicts.
+
+### Examples
+
+Configure price to match a compute savings plan
+```yaml
+kind: NodeOverlay
+metadata:
+  name: default-discount
+spec:
```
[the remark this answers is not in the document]
ellistarn: Yep. See the section below

pkg/apis/v1alpha1/nodeoverlay.go
ellistarn: I thought about this, but intentionally wanted to make all fields of the overlay explicitly optional. The way I think of it is that the lack of an overlay == not change in behavior. Think about how this would operate w/ extended resources -- we wouldn't know what to zero out, so we must instead choose the nil semantic.

kwok/charts/crds/karpenter.sh_nodeoverlays.yaml
ellistarn: I thought about using quantities and a multiplier, but the ergonomics were really confusing. e.g., what does it mean to have a multiplier of 100m. It's 1/10, but a milli multiplier is very unintuitive to me. Do you think we will see use cases of less than 1% increments?

designs/node-overlays.md
ellistarn: Yeah, but the merged overlay is on a per instance type basis, as each overlay could have a different selector. It gets challenging to model =/.

designs/node-overlays.md
ellistarn: I'll clean this up. Not something for launch.

designs/node-overlays.md
ellistarn: My current position is we shouldn't do any observability.

designs/node-overlays.md
ellistarn: We will need to decide these semantics. I think yes. But it also makes me want to revisit how our cloudprovider models base+overhead.

designs/node-overlays.md
ellistarn: This is a cool idea, but I'm not sure about the use cases it unlocks.

designs/node-overlays.md
ellistarn: Good question. Punted to a future design :)

designs/node-overlays.md
ellistarn: It will apply to all

designs/node-overlays.md
ellistarn: Let's punt to a future rev. I don't think using CEL is crazy.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// +required
+	Spec NodeOverlaySpec `json:"spec"`
+}
+
+type NodeOverlaySpec struct {
+	// Selector matches against simulated nodes and modifies their scheduling properties. Matches all if empty.
+	// +optional
+	Selector []v1.NodeSelectorRequirement `json:"selector,omitempty"`
+	// PricePercent modifies the price of the simulated node (PriceAdjustment + (Price * PricePercent / 100)).
```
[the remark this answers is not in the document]
ellistarn: The more complex the pricing constructs are, the more complex our algorithm gets -- at some point, we can't solve everything. We also need to start somewhere and make progress. One option would be to try to write some automation that removes the NodeOverlay when the savings plan is exhausted.

pkg/apis/v1alpha1/nodeoverlay.go
ellistarn: e.g. per-instance price increases, like data dog or windows licensing.

designs/node-overlays.md
ellistarn: Yes. We would prefer first class support for all of this.

designs/node-overlays.md
ellistarn: 👍

designs/node-overlays.md
ellistarn: I'm not grokking this.

designs/node-overlays.md
ellistarn: I'm gonna cut this for now :)

designs/node-overlays.md
ellistarn: Going to cut this for now :)

designs/node-overlays.md
ellistarn: Going to cut this for now :)

designs/node-overlays.md
ellistarn: I'm not grokking this

designs/node-overlays.md
ellistarn: Can you elaborate a little bit about how a user might model something with sequential applies?

designs/node-overlays.md
ellistarn: Yeah strongly agree. The merge semantic is a solve for this.

designs/node-overlays.md
ellistarn: I'd rather be clear with a concrete provider, than less clear with a fake provider.

designs/node-overlays.md
ellistarn: Logically, I think the price bits of overlays are actually operating against the offerings. From other discussions though, I think this one is moot because we can avoid conflicts with alphabetizing the way we do with nodepool.weight.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+A [previous proposal](https://github.com/aws/karpenter-provider-aws/pull/2390/files#diff-b7121b2d822e57b70203794f3b6367e7173b16d84070c3712cd2f15b8dbd11b2R133) explored the tradeoffs between a 1:1 semantic vs more flexible alternatives. It raised valid observability challenges around communicating to customers exactly how the CR was modifying the scheduling simulation. It recommended a 1:1 approach, with the assumption that some system would likely dynamically codegenerate the CRs for the relevant use case. We received [direct feedback](https://github.com/aws/karpenter-provider-aws/pull/2404#issuecomment-1250688882) from some customers that the number of variations would be combinatorial, and thus unmanagable for their use case. Ultimately, we paused this analysis in favor of other priorities.
```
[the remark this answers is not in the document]
ellistarn: Custom resource. An instance of a CRD (the definition)

pkg/apis/v1alpha1/nodeoverlay.go
jwcesign: Got it, but the price could be queried from the aws price API. Why not put it into the InstanceType struct?

pkg/apis/v1alpha1/nodeoverlay.go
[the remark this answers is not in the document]
jwcesign: > I don't know how you could detect this.

It will query the Compute Savings Plan and calculate how much it costs for the current existing nodes. If the total hourly cost is over the commitment, it should be the original on-demand price. It may not be very accurate, but it could solve part of the problem.

> At some point, we can't solve everything.

Absolutely correct. I am thinking about whether we can write a component to handle this kind of situation. Karpenter will query the real price from the component service.

This is what I am thinking.

```yaml
# ec2 sp node overlay
---
kind: NodeOverlay
metadata:
  name: discount
spec:
  selector:
    matchLabels:
      karpenter.sh/capacity-type: on-demand
  priceAPIEndpoint: http://price.karpenter.svc:8080
```

```yaml
# ec2 sp node pool
kind: NodePool
metadata:
  name: default
spec:
  template:
    spec:
      requirements:
        - key: kubernetes.io/arch
          operator: In
          values: ["amd64"]
        - key: kubernetes.io/os
          operator: In
          values: ["linux"]
          operator: In
          values: ["t"]
          operator: Gt
          values: ["3"]
      nodeClassRef:
        apiVersion: karpenter.k8s.aws/v1beta1
        kind: EC2NodeClass
        name: default
  limits:
    limitsAPIEndpoint: http://limits.karpenter.svc:8080
  weight: 100 # highest
---
# other node pool
```

So, using the EC2 SP node pool, it will query the limits and price from another component, and the complex logic will be calculated in that component. When the EC2 SP is finished, we will use another node pool.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
+# Node Overlays
+
+Node launch decisions are the output a scheduling simulation algorithm that involves satisfying pod scheduling constraints with available instance type offerings. Cloud provider implementations are responsible for surfacing these offerings to the Karpenter scheduler through the [cloudprovider.GetInstanceTypes()](https://github.com/kubernetes-sigs/karpenter/blob/37db06f4742eada19934f76e616f2b1860aca57b/pkg/cloudprovider/types.go#L59) API. This separation of responsibilities enables cloud providers to make simplifying assumptions, but limits extensibility for more advanced use cases. NodeOverlays enable customers to inject alternative assumptions into the scheduling simulation for more accurate scheduling simulations.
+
+## Use Cases
+
+### Price
+
+Each instance offering comes with a price, which is used as a minimization function in Karpenter's scheduling algorithm. However, Karpenter cannot perfectly understand all factors that influence price, such as an external vendor's fee, a pricing discount, or an external motiviation such as carbon-offset.
```
tallaxes, line 9: It may be worth describing the potential interaction of client-side use (and customization) of pricing with provider-specific server-side price-sensitive allocation strategies (like those supported by EC2 Fleet).

rlindsberg:
Hi all!
Do we have an estimation of when this feature will be publicly available? We have been waiting for it for over a year now..

Our use case is we run Circle CI runner jobs on a EKS cluster. And we have this limit for [buildah](https://buildah.io) to work:

```
resources:
  limits:
    github.com/fuse: 2
```

But we got the following error:
```
{
    "level": "ERROR",
    "logger": "controller.provisioner",
    "message": "Could not schedule pod, incompatible with nodepool \"default\", daemonset overhead={\"cpu\":\"180m\",\"memory\":\"120Mi\",\"pods\":\"8\"}, no instance type satisfied resources {\"cpu\":\"2180m\",\"github.com/fuse\":\"1\",\"memory\":\"8312Mi\",\"pods\":\"9\"} and requirements karpenter.k8s.aws/instance-category In [m t], karpenter.k8s.aws/instance-generation Exists >2, karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [default], kubernetes.io/arch In [amd64], kubernetes.io/os In [linux] (no instance type has enough resources); incompatible with nodepool \"arm64\", daemonset overhead={\"cpu\":\"180m\",\"memory\":\"120Mi\",\"pods\":\"8\"}, no instance type satisfied resources {\"cpu\":\"2180m\",\"github.com/fuse\":\"1\",\"memory\":\"8312Mi\",\"pods\":\"9\"} and requirements karpenter.k8s.aws/instance-category In [c m t], karpenter.k8s.aws/instance-generation Exists >2, karpenter.sh/capacity-type In [on-demand], karpenter.sh/nodepool In [arm64], kubernetes.io/arch In [arm64], kubernetes.io/os In [linux] (no instance type has enough resources)"
}
```

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+spec:
+  selector:
+    matchLabels:
+      karpenter.sh/capacity-type: on-demand
+    matchExpressions:
+    - key: node.kubernetes.io/instance-type
+      operator: In
+      values: ["m5.large", "m5.2xlarge", "m5.4xlarge", "m5.8xlarge", "m5.12xlarge"]
+  capacity:
+    smarter-devices/fuse: 1
```
wmgroot, line 60: To be clear, the "extended resource support" in this design would only support statically applied extended resources on selected nodes, correct? This would not cover cases where dynamic extended resources are needed, for example setting an extended resource for available network bandwidth which varies by EC2 instance type.

Additionally, karpenter still needs to have its provisioning and deprovisioning logic updated to account for arbitrary extended resources other than the gpu resource, right?

designs/node-overlays.md
ellistarn: Strongly agree that we should provide first class support for extended resources wherever possible. In my mind it's whether or not these can be reasonably known by a cloud provider -- some simply can't.

designs/node-overlays.md
ellistarn: > This would not cover cases where dynamic extended resources are needed, for example setting an extended resource for available network bandwidth which varies by EC2 instance type. 

You could achieve through multiple NodeOverlays, that match different selectors.

e.g. m5.large has 2, m5.2xlarge has 4.

designs/node-overlays.md
```diff
@@ -0,0 +1,186 @@
...
+
+### Merging
+
+Our decision to pursue Selectors that allow for an n:m semantic presents a challenge where multiple NodeOverlays could match a simulation, but with different values. To address this, we will support an optional `.spec.weight` parameter (discussed in the [previous proposal](https://github.com/aws/karpenter-provider-aws/pull/2390/files#diff-b7121b2d822e57b70203794f3b6367e7173b16d84070c3712cd2f15b8dbd11b2R153)).
+
+We have a few choices for dealing with conflicting NodeOverlays
+
+1. Multiple NodeOverlays are merged in order of weight
+2. Multiple NodeOverlays are applied sequentially in order of weight
+3. One NodeOverlay is selected by weight
```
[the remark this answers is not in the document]
ellistarn: Baking these semantics into selector is quite a bit more complicated, and requires customers to deeply understand this. There may be cases where they disagree as well. Weight is a common idiom from other scheduling APIs, which is why I chose this here.

designs/node-overlays.md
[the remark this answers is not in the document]
ellistarn: Agree. Updating this to always make progress by tie breaking alphabetically, the same way we do tie breaking for node pools.

designs/node-overlays.md
wmgroot: I don't think defining a NodeOverlay for each network bandwidth amount is a reasonable solution for my use case. We use a wide variety of instance types and sizes.

Do you think it could be feasible to incorporate some level of templating for values that can be easily retrieved from the EC2 API?

ellistarn: For NetworkBandwidth, I'd love to just support this one first class. 

We could always try to do something that attempts to model this like: 
```
resourcesPerCPU: 
  network: 1Gi
```
But it's hard to capture all potential relationships without devolving into something like https://github.com/google/cel-spec

daverin:
Hi All

I'm a bit confused about the relationship between [kubernetes-sigs/karpenter](https://github.com/kubernetes-sigs/karpenter/) and [aws/karpenter-provider-aws](https://github.com/aws/karpenter-provider-aws).

The lack of support for custom resource requests/limits is blocking me from autocalling with [xilinx](https://xilinx.github.io/video-sdk/v3.0/deploying_with_kubernetes.html?highlight=eks#installing-the-xilinx-device-plugin-for-kubernetes) devices.

I want to understand how to deploy this branch to EKS.

Any help or understanding would be appreciated.

sftim:
> I want to understand how to deploy this branch to EKS.

This branch adds a “request for comments“. @ellistarn should the RFC and implementation land as one commit? Seems unusual.

zaafar:
i think this shouldn't be closed.

zaafar:
commenting to make it not stale.

designs/node-overlays.md
chomatdam: Similar use case with KubeVirt, VM pods are getting "dynamic" label keys as node selectors, e.g. `scheduling.node.kubevirt.io/tsc-frequency-${cpuTimerValue}`

In my Karpenter fork, just providing an "ignored list" through environment variable where I'm doing pattern matching against `scheduling.node.kubevirt.io/tsc-frequency-*`

abbas-square:
Is there any update on this?

devusb:
Is this still planned for a future Karpenter release? Working on deploying in some clusters where I use hugepages, bumping up against this issue -- just trying to get a feel for whether I'll need a short-term workaround or if this is planned to be a long-term limitation.

zarko-a:
Also wondering if this is still planned to be merged.

likeabbas:
@ellistarn could you provide and update on when this PR will be merged?

ssanders1449:
commenting to make it not stale

LeszekBlazewski:
Just bumping the PR. My team would be highly interested in this feature as well for similar reasons that other commenters already have mentioned.

Keep up the awesome work on karpenter

pkg/apis/v1alpha1/nodeoverlay.go
MarkIannucci: @jwcesign , how would I query datadog's monthly price per host from the AWS price API?

rschalo:
Discussed in triage meeting, what's left for this RFC is that we still need to find agreement for implementation as well as what is being implemented first. This is likely the price adjustment.

ellistarn:
Would this be better named NodeProperties, or something? Overlay implies to me that it's happening after/on top of the node, where it's really a simulation concept.

jonathan-innis:
> Overlay implies to me that it's happening after/on top of the node, where it's really a simulation concept

But it's a merge semantic with the existing data -- I'm actually still convinced by the current `NodeOverlay` naming :)

sftim:
Slightly devil's advocate: is it more of NodeClaimOverlay?

designs/node-overlays.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Observability
+
+Exposing the full details of the mutations caused by NodeOverlays cannot be feasibly stored in ETCD. Further, we do not want to require that customers read the Karpenter logs to debug this feature. When GetInstanceTypes is called per NodePool, we will emit an event on the NodePool to surface deeper semantic information about the effects of the NodeOverlays. Our implementation will need to be conservative on the volume and frequency of information, but provide sufficient insight for customers to verify that their configurations are working as expected.
+
+### Per-Field Semantics
+
+The semantics of each field carry individual nuance. Price could be modeled directly via a `price` field, but we can significantly reduce the number of NodeOverlays that need to be defined if we instead model the semantic as a function of the currently defined price (e.g. `pricePercent` or `priceAdjustment`). Similarly, `capacity` and `overhead` make sense to be modeled directly, but `capacityPercent` makes much less sense than `overheadPercent`.
```
adammw, line 165: As an alternative to CEL here - it could work with having each modifier be an object (ie, with a chosen operator, like selectors) rather than explicitly having each field make decisions of what resource to scale on.

e.g. for the default use case a static operator
```
spec:
  overhead:
    memory:
      operator: Static
      value: 2MiB
```

Or a percentage operator, to have percentage of existing
```
spec:
  overhead:
    memory:
      operator: Percentage
      value: 80%
```

Or could get fancy with things like PerCPU
```
spec:
  overhead:
    memory:
      operator: PerCPU
      value: 20MiB
```

Or point to another resource (e.g. for the GPU splitting use-case)
```
spec:
  capacity:
    nvidia.com/gpu:
      operator: PerResource
      resource: nvidia.com/gpu
      value: 2 
```

engedaam:
Closing this old PR in favor of https://github.com/kubernetes-sigs/karpenter/pull/2166

/close

PR #1311, opened 2024-06-09
feat: promote drift to stable

rschalo:
Fixes #N/A 

**Description**
Removes the feature gating of `Drift`. Will need to update `values.yaml` in cloudprovider specific karpenter.

Deployed to local karpenter and unit testing.


pkg/controllers/nodeclaim/disruption/suite_test.go
```diff
@@ -126,12 +126,12 @@ var _ = Describe("Disruption", func() {
 		ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
 		ExpectMakeNodeClaimsInitialized(ctx, env.Client, nodeClaim)
 
-		// Drift, Expiration, and Emptiness are disabled through configuration
-		ctx = options.ToContext(ctx, test.Options(test.OptionsFields{FeatureGates: test.FeatureGates{Drift: lo.ToPtr(false)}}))
+		// Expiration, and Emptiness are disabled through configuration
```
jonathan-innis, line 129: Do we even need this comment anymore? Or the following line? This was originally here from what I can remember to make sure that when we change the setting that it re-propagates. Is this test relevant anymore?

rschalo: Good catch - I think we can take out the comment and line. The test does still test for the simultaneous removal of Empty and Expired conditions. I'm not sure if this is tested anywhere else.

jonathan-innis (approved):
/lgtm
/approve

PR #1314, opened 2024-06-10
[WIP] feat: add disruption candidate taints

jmdeal:
Fixes #N/A 

**Description**
This PR adds a new controller to taint disruption candidates with `karpenter.sh/candidate:PreferNoSchedule`.

`make test`

### DIFF
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -26,6 +26,7 @@ import (
 const (
 	DisruptionTaintKey             = apis.Group + "/disruption"
 	DisruptingNoScheduleTaintValue = "disrupting"
+	DisruptionCandidateTaintKey    = Group + "/candidate"
 )
 
 var (
@@ -35,6 +36,11 @@ var (
 		Key:    DisruptionTaintKey,
 		Effect: v1.TaintEffectNoSchedule,
 		Value:  DisruptingNoScheduleTaintValue,
+	}
+
+	DisruptionCandidatePreferNoScheduleTaint = v1.Taint{
+		Key:    DisruptionCandidateTaintKey,
+		Effect: v1.TaintEffectPreferNoSchedule,
 	}
 )
 
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -42,6 +42,16 @@ var (
 		Key:    DisruptionCandidateTaintKey,
 		Effect: v1.TaintEffectPreferNoSchedule,
 	}
+
+	// DisruptionCandidateNoScheduleTaint is applied in-memory to drifted and expired nodes.
+	// This, along with the PreferNoSchedule taint applied to all disruption candidates, ensures
+	// that we don't reschedule nodes on disruption candidates to other disruption candidates.
+	// Note: this taint is **only** applied in-memory. It does not get applied to nodes on the
+	// API server.
+	DisruptionCandidateNoScheduleTaint = v1.Taint{
+		Key:    DisruptionCandidateTaintKey,
+		Effect: v1.TaintEffectNoSchedule,
+	}
 )
 
 func IsDisruptingTaint(taint v1.Taint) bool {
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -26,7 +26,7 @@ import (
 const (
 	DisruptionTaintKey             = apis.Group + "/disruption"
 	DisruptingNoScheduleTaintValue = "disrupting"
-	DisruptionCandidateTaintKey    = Group + "/candidate"
+	DisruptionCandidateTaintKey    = apis.Group + "/candidate"
 )
 
 var (
--- a/pkg/apis/v1beta1/taints.go
+++ b/pkg/apis/v1beta1/taints.go
@@ -38,19 +38,10 @@ var (
 		Value:  DisruptingNoScheduleTaintValue,
 	}
 
+	// DisruptionCandidatePreferNoScheduleTaint is applied to any disruption candidates (e.g. drifted, empty, etc).
 	DisruptionCandidatePreferNoScheduleTaint = v1.Taint{
 		Key:    DisruptionCandidateTaintKey,
 		Effect: v1.TaintEffectPreferNoSchedule,
-	}
-
-	// DisruptionCandidateNoScheduleTaint is applied in-memory to drifted and expired nodes.
-	// This, along with the PreferNoSchedule taint applied to all disruption candidates, ensures
-	// that we don't reschedule nodes on disruption candidates to other disruption candidates.
-	// Note: this taint is **only** applied in-memory. It does not get applied to nodes on the
-	// API server.
-	DisruptionCandidateNoScheduleTaint = v1.Taint{
-		Key:    DisruptionCandidateTaintKey,
-		Effect: v1.TaintEffectNoSchedule,
 	}
 )
 
--- a/pkg/controllers/node/tainting/controller.go
+++ b/pkg/controllers/node/tainting/controller.go
@@ -1,0 +1,139 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tainting
+
+import (
+	"context"
+	"fmt"
+	golog "log"
+
+	"github.com/awslabs/operatorpkg/singleton"
+	"github.com/awslabs/operatorpkg/status"
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+
+	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
+)
+
+var (
+	// conditionToTaintsMap maps condition values which can be present on the node to a set of taints that should be added in response
+	conditionToTaintsMap = map[string]v1.Taint{
+		v1beta1.ConditionTypeDrifted: v1beta1.DisruptionCandidatePreferNoScheduleTaint,
+		v1beta1.ConditionTypeExpired: v1beta1.DisruptionCandidatePreferNoScheduleTaint,
+		v1beta1.ConditionTypeEmpty:   v1beta1.DisruptionCandidatePreferNoScheduleTaint,
+	}
+
+	// managedTaints is the string representation of all taints managed by the tainting controller
+	managedTaints = sets.New(lo.Uniq(lo.MapToSlice(conditionToTaintsMap, func(_ string, taint v1.Taint) string {
+		return taint.ToString()
+	}))...)
+)
+
+func init() {
+	// Assert that no managed taint has a key-value collision. The tainting controller relies on this invariant, change with caution.
+	taintsPerKeyEffect := lo.GroupBy(lo.UniqBy(lo.Values(conditionToTaintsMap), func(taint v1.Taint) string {
+		return taint.ToString()
+	}), func(taint v1.Taint) string {
+		// Group all taints with the same key + effect together
+		taint.Value = ""
+		return taint.ToString()
+	})
+	for _, taints := range taintsPerKeyEffect {
+		if len(taints) != 1 {
+			golog.Fatalf("taints managed by the tainting controller must only have one value per key-effect pair")
+		}
+	}
+}
+
+type Controller struct {
+	kubeClient client.Client
+}
+
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		kubeClient: kubeClient,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	node := &v1.Node{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClaim.Status.NodeName}, node); err != nil {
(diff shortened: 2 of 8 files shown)


njtran (commented):
I'm wondering why we need the additional layer of direction here and can't just add the taint when we add the status condition in the nodeclaim lifecycle controller?

jmdeal:
We might be able to get away with that if all conditions are set by the NodeclaimLifecycleController but I didn't think that was going to be the case for Underutilized, or has that changed? This is factoring for the changes in your ConsolidateAfter PR

pkg/apis/v1beta1/taints.go
```diff
@@ -32,6 +33,21 @@ var (
...
+		Key:    DisruptionCandidateTaintKey,
+		Effect: v1.TaintEffectPreferNoSchedule,
+	}
+
+	// DisruptionCandidateNoScheduleTaint is applied in-memory to drifted and expired nodes.
+	// This, along with the PreferNoSchedule taint applied to all disruption candidates, ensures
+	// that we don't reschedule nodes on disruption candidates to other disruption candidates.
+	// Note: this taint is **only** applied in-memory. It does not get applied to nodes on the
+	// API server.
+	DisruptionCandidateNoScheduleTaint = v1.Taint{
```
jonathan-innis, line 47: I'm still not convinced that we want to store this in the APIs package. This isn't used externally by anything. Why not in scheduling?

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+
+	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
+)
+
+var (
+	// conditionToTaintsMap maps condition values which can be present on the node to a set of taints that should be added in response
+	conditionToTaintsMap = map[string]v1.Taint{
```
jonathan-innis, line 45: If every condition is added the same taint, what's the need of this map

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// managedTaints is the string representation of all taints managed by the tainting controller
+	managedTaints = sets.New(lo.Uniq(lo.MapToSlice(conditionToTaintsMap, func(_ string, taint v1.Taint) string {
+		return taint.ToString()
+	}))...)
+)
+
+func init() {
+	// Assert that no managed taint has a key-value collision. The tainting controller relies on this invariant, change with caution.
+	taintsPerKeyEffect := lo.GroupBy(lo.UniqBy(lo.Values(conditionToTaintsMap), func(taint v1.Taint) string {
```
jonathan-innis, line 59: Do we need this init()? I'm a little unclear why we can't enforce this invariant just be knowledge of how this controller works

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+func NewController(kubeClient client.Client) *Controller {
+	return &Controller{
+		kubeClient: kubeClient,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeClaim) (reconcile.Result, error) {
+	node := &v1.Node{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClaim.Status.NodeName}, node); err != nil {
+		if client.IgnoreNotFound(err) == nil {
```
jonathan-innis, line 86: nit: Typical handling is `errors.IsNotFound()` if you are doing checking like this

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("getting node, %w", err)
+	}
+	stored := node.DeepCopy()
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("node", node.Name))
```
jonathan-innis, line 92:
```suggestion
	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef("", node.Name)))
```

See how we are referring to objects elsewhere in the code, we're using `klog.Kref` for referencing objects and capitalizing the object name

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+
+	taintsForConditions := lo.SliceToMap(lo.FilterMap(nodeClaim.GetConditions(), func(cond status.Condition, _ int) (v1.Taint, bool) {
```
jonathan-innis, line 94: Can this just be a check on the condition type and then know that it should be a PreferNoSchedule? Do we expect to expand the taint that we are using here? If so, can we change it then rather than now?

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+		if taint, ok := conditionToTaintsMap[cond.Type]; ok {
+			return taint, cond.IsTrue()
+		}
+		return v1.Taint{}, false
+	}), func(taint v1.Taint) (string, v1.Taint) {
+		return taint.ToString(), taint
+	})
+	currentTaints := sets.New(lo.Map(node.Spec.Taints, func(taint v1.Taint, _ int) string {
```
jonathan-innis, line 102: Is there a helper that we can use here to do an idempotent add to the taints since this isn't the only place that we add taints like this

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+	if toAdd.Len() == 0 && toRemove.Len() == 0 {
+		return reconcile.Result{}, nil
+	}
+
+	node.Spec.Taints = lo.Reject(node.Spec.Taints, func(taint v1.Taint, _ int) bool {
+		return toRemove.Has(taint.ToString())
+	})
+	node.Spec.Taints = append(node.Spec.Taints, lo.Values(lo.PickByKeys(taintsForConditions, toAdd.UnsortedList()))...)
+
+	if len(toAdd) != 0 {
```
jonathan-innis, line 118: Just high-level: This strikes me as complicated and a bit hard to reason about. Are there helpers that we can make here for adding the taint and removing the taint to make this easier for someone new coming in to read?

pkg/controllers/node/tainting/controller.go
```diff
@@ -0,0 +1,139 @@
...
+		log.FromContext(ctx).V(1).Info("adding taints for status conditions", "taints", toAdd.UnsortedList())
+	}
+	if len(toRemove) != 0 {
+		log.FromContext(ctx).V(1).Info("removing taints for status conditions", "taints", toRemove.UnsortedList())
+	}
+	if err := nodeutils.PatchTaints(ctx, c.kubeClient, stored, node); err != nil {
```
jonathan-innis, line 124: Should this be `PatchTaints` or is this about PatchingWithConflict? I wonder if there's a good way to generalize this into a util? Since this is not very well known, I wonder if we just go with Update() since everyone is familiar that this path can lead to a conflict and just deal with the fact that this may take some retries

pkg/controllers/state/statenode.go
```diff
@@ -291,6 +290,13 @@ func (in *StateNode) Taints() []v1.Taint {
 			return false
 		})
 	}
+	// Add an in-memory NoSchedule taint for all eventual disruption candidates. This ensures that we don't consider
+	// these nodes as candidates for pending pods being rescheduled from other disruption candidates.
+	if in.EventualDisruptionCandidate() && !lo.ContainsBy(taints, func(t v1.Taint) bool {
+		return t.MatchTaint(&v1beta1.DisruptionCandidateNoScheduleTaint)
+	}) {
+		taints = append(taints, v1beta1.DisruptionCandidateNoScheduleTaint)
```
jonathan-innis, line 298: Should we be doing an append or a mutation of the existing taint?

pkg/controllers/state/statenode.go
```diff
@@ -312,6 +318,18 @@ func (in *StateNode) Initialized() bool {
 	return true
 }
 
+func (in *StateNode) EventualDisruptionCandidate() bool {
+	if !in.Managed() {
+		return false
+	}
+	return lo.ContainsBy(in.NodeClaim.GetConditions(), func(cond status.Condition) bool {
+		if cond.Type != v1beta1.ConditionTypeDrifted && cond.Type != v1beta1.ConditionTypeExpired {
```
jonathan-innis, line 326: Does empty not come into scope here?

pkg/controllers/state/statenode.go
```diff
@@ -312,6 +318,18 @@ func (in *StateNode) Initialized() bool {
 	return true
 }
 
+func (in *StateNode) EventualDisruptionCandidate() bool {
```
jonathan-innis, line 321: Can we add a comment over this to describe what we are going for here?

pkg/utils/node/node.go
```diff
@@ -83,3 +83,17 @@ func GetCondition(n *v1.Node, match v1.NodeConditionType) v1.NodeCondition {
...
+	// had been updated between the original GET and the PATCH. This is necessary because .spec.taints does not specify
+	// a patchMergeKey and patchStrategy. Ref: https://github.com/kubernetes/kubernetes/pull/113136
+	oldNodeStripped := oldNode.DeepCopy()
+	oldNodeStripped.ResourceVersion = ""
+
+	newTaints := newNode.Spec.Taints
+	newNodeCloned := oldNode.DeepCopy()
+	newNodeCloned.Spec.Taints = newTaints
+
+	return kubeClient.Patch(ctx, newNodeCloned, client.StrategicMergeFrom(oldNodeStripped))
```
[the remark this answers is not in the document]
jonathan-innis: I question whether we should take in this hack since it takes someone understanding what's going on here and why this works. What's the issue that we think that we might encounter here by simplifying with an Update() call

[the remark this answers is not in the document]
jonathan-innis: Is there some option that we can ask controller-runtime to support? Or is there a wrapper around our client that we should add that would support `PatchWithConflict()`

pkg/controllers/state/statenode.go
jmdeal: No, we only want to weigh away from drifted or expired nodes. If we can reschedule to an empty node, that is the indication that we **shouldn't** consolidate an empty node.

pkg/controllers/state/statenode.go
jmdeal: This could be added before the tainting controller adds the `PreferNoSchedule` taint, or before cluster state is aware of the change. This eliminates a case where we may run a scheduling simulation and nominate a pod for a drifted node before we become aware of the taint. I don't think this is strictly necessary, but is an improvement.

pkg/controllers/node/tainting/controller.go
jmdeal: I could see calling this `PatchWithConflicts` and generalizing. I also think I'm fine with just using `Update` and defer.

pkg/controllers/node/tainting/controller.go
jmdeal: Since the values in this map are known at compile time, I wanted to create a compile time assertion to validate this behavior. I was sad to learn there isn't a way to do this in go, so I went with this approach. This could totally be a comment, but an assertion like this will give a loud signal for a programmer error.

pkg/controllers/node/tainting/controller.go
jmdeal: We probably don't **need** this, I wrote the initial implementation when we were still considering different taints. This does have the advantage of being easily expandable without adding significant complexity, but this could be reduced to just a set of status conditions in the short term.

pkg/utils/node/node.go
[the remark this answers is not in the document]
jmdeal: That was the intention of wrapping it here, and calling out the hack in the comment. IIUC the main advantage is to reduce potential retries. I think we can probably start with Update and add this as an optimization down the road if necessary.

pkg/apis/v1beta1/taints.go
jmdeal: I'm good with removing it and constructing the taint directly in statenode.

pkg/controllers/node/tainting/controller.go
jmdeal: This was originally implemented when it was up in the air if we were doing a single or multiple taints. I was going to leave the general approach since it makes it easy for us to expand this in the future without adding much complexity.

pkg/controllers/node/tainting/controller.go
njtran: I'm on the same page. Could consider adding a utils taint package that gets utilized in the other places where we're also orchestrating tainting on the node

pkg/controllers/state/statenode.go
njtran: The way this is orchestrated means that someone could theoretically tolerate the PreferNoSchedule taint with their pods, but their pods wouldn't tolerate the NoSchedule, right? We might want to also allow the pods that tolerate PreferNoSchedule to also tolerate this in memory by default.

pkg/controllers/node/tainting/controller.go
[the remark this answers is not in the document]
jmdeal: I'm removing this as part of "degeneralizing" the controller, but it actually wouldn't be possible to hit in the suite test. The idea is that this is a compile time invariant; it would fail at startup every time, or it would never fail at all. You'd have to change the data it's checking the invariant against then recompile.

pkg/apis/v1beta1/taints.go
```diff
@@ -32,6 +33,21 @@ var (
...
+	DisruptionCandidatePreferNoScheduleTaint = v1.Taint{
+		Key:    DisruptionCandidateTaintKey,
+		Effect: v1.TaintEffectPreferNoSchedule,
+	}
+
+	// DisruptionCandidateNoScheduleTaint is applied in-memory to drifted and expired nodes.
+	// This, along with the PreferNoSchedule taint applied to all disruption candidates, ensures
+	// that we don't reschedule nodes on disruption candidates to other disruption candidates.
+	// Note: this taint is **only** applied in-memory. It does not get applied to nodes on the
+	// API server.
```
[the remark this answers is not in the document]
jmdeal: Going to take @jonathan-innis's suggestion and just construct this in place in cluster state that way it's not even here.

pkg/controllers/state/statenode.go
jmdeal: That's a good point, it's currently in a weird state where if you tolerate the taint we apply you still get the weigh-away behavior, but not if you tolerate all effects. I'm wondering if a toleration is the right mechanism to opt out of the weigh away though (vs a separate field we could add down the line). FWIW I think it probably is since, by tolerating the `PreferNoSchedule` taint, there's nothing stopping a pod from scheduling to a drifted node even if we spin up new capacity.

jmdeal:
Realized there's a race condition with the current implementation for the eventual candidate weigh away. It should work fine for **voluntary** disruption since we wait for the replacement nodes to register before proceeding with draining the disrupted nodes. However, this isn't the case with non-voluntary disruption. Since we include pods from deleting nodes in scheduling simulations, we can run into a situation where we determine we a new NodeClaim since we won't schedule pods to drifted nodes. However, the pod drains and the replacement schedules to the drifted node (since it's only a `NoSchedule` taint in-memory all before the new NodeClaim is created. Since this is going to introduce unnecessary node churn, I'm going to drop the weigh-away behavior for now and just implement the `PreferNoSchedule` taint in this PR.

PR #1319, opened 2024-06-13
fix: kwok offerings to use correct requirements

njtran:
Fixes #N/A 

**Description**
The kwok instance types use a differently typed requirements. The instance types have the right requirements, but the offerings don't have any requirements. This fixes that bug and adds them in

make apply


jonathan-innis (approved):
/lgtm
/approve

PR #1320, opened 2024-06-14
test: update node termination test to model eventual removal

jmdeal:
Fixes #N/A 

**Description**
Updates the node/termination test suite to properly model eventual removal now that it may take multiple reconciliations.

`make test`

### DIFF
--- a/pkg/controllers/node/termination/suite_test.go
+++ b/pkg/controllers/node/termination/suite_test.go
@@ -78,6 +78,9 @@ var _ = AfterSuite(func() {
 })
 
 var _ = Describe("Termination", func() {
+	// expectedReconciliations is the number of reconciliations the termination controller should need to remove a node.
+	// One reconciliation to start terminating the nodeclaim, and another to get the instance, see that it's gone, and remove the node.
+	const expectedReconciliations = 2
 	var node *v1.Node
 	var nodeClaim *v1beta1.NodeClaim
 
@@ -104,12 +107,10 @@ var _ = Describe("Termination", func() {
 			ExpectApplied(ctx, env.Client, node, nodeClaim)
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
-			Eventually(func(g Gomega) {
-				_, err := reconcile.AsReconciler(env.Client, terminationController).Reconcile(ctx, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(node)})
-				g.Expect(err).ToNot(HaveOccurred())
-				tmpNode := &v1.Node{}
-				g.Expect(errors.IsNotFound(env.Client.Get(ctx, client.ObjectKeyFromObject(node), tmpNode))).To(BeTrue())
-			}, ReconcilerPropagationTime, RequestInterval).Should(Succeed())
+			for range expectedReconciliations {
+				ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			}
+			ExpectNotFound(ctx, env.Client, node)
 		})
 		It("should delete nodeclaims associated with nodes", func() {
 			ExpectApplied(ctx, env.Client, node, nodeClaim)
@@ -143,7 +144,7 @@ var _ = Describe("Termination", func() {
 				nodes = append(nodes, node)
 			}
 
-			Eventually(func(g Gomega) {
+			for range expectedReconciliations {
 				var wg sync.WaitGroup
 				// this is enough to trip the race detector
 				for i := 0; i < numNodes; i++ {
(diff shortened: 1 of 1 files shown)


jonathan-innis:
Can you talk with @jigisha620 on this one? I'm surprised that the changes that were made in the other PR didn't cause failures in these tests and we're having to make these changes after the fact now. 

Can you clarify what is failing that wasn't properly captured through testing?

jmdeal:
Already spoke with her offline about it, the gist of the problem is that the node termination controller **can** now take multiple reconciliations to delete the node. It didn't before because there was never a nodeclaim associated with the node, so it immediately deleted. IMO since it can now take multiple reconciliations this is the correct way to model it and unblocks some issues upon rebase in #916 since it applied the nodeclaim.

pkg/controllers/node/termination/suite_test.go
```diff
@@ -129,34 +143,43 @@ var _ = Describe("Termination", func() {
...
+						_, err := reconcile.AsReconciler(env.Client, terminationController).Reconcile(ctx, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(node)})
+						g.Expect(err).ToNot(HaveOccurred())
+					}(nodes[i])
+				}
+				wg.Wait()
+				for _, n := range nodes {
+					tmpNode := &v1.Node{}
+					g.Expect(errors.IsNotFound(env.Client.Get(ctx, client.ObjectKeyFromObject(n), tmpNode))).To(BeTrue())
+				}
+				ExpectNotFound(ctx, env.Client, lo.Map(nodes, func(n *v1.Node, _ int) client.Object { return n })...)
```
jigisha620, line 163: Aren't we validating the same thing on line 161?

jigisha620:
> Can you clarify what is failing that wasn't properly captured through testing?

There wasn't a nodeClaim associated with a node in the tests that we had and that's why there was no failure. This wasn't captured until #916 applied nodeClaim and started seeing the failures.

pkg/controllers/node/termination/suite_test.go
jmdeal: Oh that's a miss on my part, will remove. This is actually a no-op since we can't use these expectations inside `Eventually`.

pkg/controllers/node/termination/suite_test.go
```diff
@@ -99,25 +101,37 @@ var _ = Describe("Termination", func() {
 
 	Context("Reconciliation", func() {
 		It("should delete nodes", func() {
-			ExpectApplied(ctx, env.Client, node)
+			ExpectApplied(ctx, env.Client, node, nodeClaim)
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
-			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
-			ExpectNotFound(ctx, env.Client, node)
+			Eventually(func(g Gomega) {
```
jonathan-innis, line 107: I'd like us to avoid doing eventual calls when we control what the backend looks like. Can we just make the number of calls that we need to make to ensure that the instance is removed on the backend so that we know a specific ExpectObjectReconciled call will succeed to terminate?

jigisha620 (approved):
LGTM  :rocket:

pkg/controllers/node/termination/suite_test.go
jonathan-innis: These kinds of eventually calls also lead to the kind of state that we are seeing over in the disruption code where we don't have any kind of guarantees that operations will be completed if we are relying on a time construct

pkg/controllers/node/termination/suite_test.go
```diff
@@ -186,8 +209,12 @@ var _ = Describe("Termination", func() {
 
 			// Reconcile to delete node
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
-			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
-			ExpectNotFound(ctx, env.Client, node)
+			Eventually(func(g Gomega) {
```
jonathan-innis, line 212: Same comment here about Eventually. Do we need to do this with a time-based check or can we get away with doing this in a way that has more determinism around it?

pkg/controllers/node/termination/suite_test.go
jmdeal: We could, currently we only need two iterations. I had opted for eventually since I didn't see a massive increase in test duration and this decoupled the test from the fake cloudprovider implementation. I'm fine with hardcoding this to two iterations and adjusting if we need to later down the road.

jonathan-innis: I think we should opt for determinism any time that we can. Let's go with two reconciliations for now

pkg/controllers/node/termination/suite_test.go
```diff
@@ -76,6 +78,9 @@ var _ = AfterSuite(func() {
 })
 
 var _ = Describe("Termination", func() {
+	// expectedReconciliations is the number of reconciliations the termination controller should need to remove a node.
+	// One reconciliation to start terminating the nodeclaim, and another to get the instance, see that it's gone, and remove the node.
+	const expectedReconciliations = 2
```
jonathan-innis, line 83: IMO: I think abstracting in this way may actually make things more confusing than just calling the ExpectObjectReconciled function multiple times in code. This isn't that strange of a pattern to do (we do it in the NodeClaim termination loop and in some other termination flows) and, in my opinion, creates a lot less indirection when it comes to figuring out what's happening and _why_ it's happening in a particular test case.

What do you think about just calling ExpectObjectReconciled twice and then just commenting above each call what the call is there for?

jmdeal: Sure, I don't have a strong preference either way. My original thought on eventually was so we could decouple the number of iterations, but if we're doing a set number either way it doesn't make a big difference. I think I agree that it's more readable if we just inline things.

jonathan-innis (approved):
/lgtm
/approve

PR #1324, opened 2024-06-17
test: add test timestamp collector

njtran:
Fixes #N/A 

**Description**
Adds a Collector that automatically collects and prints timestamps of the tests to a temporary csv directory. 

make apply && make e2etests

### DIFF
--- a/test/pkg/environment/common/environment.go
+++ b/test/pkg/environment/common/environment.go
@@ -133,13 +133,16 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
 		})
 		return []string{t.Value}
 	}))
-	// drifted type
+	// disruption status conditions
 	lo.Must0(cache.IndexField(ctx, &v1beta1.NodeClaim{}, "status.conditions[*].type", func(o client.Object) []string {
 		nodeClaim := o.(*v1beta1.NodeClaim)
-		t, _ := lo.Find(nodeClaim.Status.Conditions, func(c status.Condition) bool {
-			return c.Type == v1beta1.ConditionTypeDrifted
+		conditionTypes := lo.FilterMap(nodeClaim.Status.Conditions, func(c status.Condition, _ int) (string, bool) {
+			if c.Type == v1beta1.ConditionTypeDrifted || c.Type == v1beta1.ConditionTypeEmpty || c.Type == v1beta1.ConditionTypeExpired {
+				return c.Type, true
+			}
+			return "", false
 		})
-		return []string{t.Type}
+		return conditionTypes
 	}))
 
 	c := lo.Must(client.New(config, client.Options{Scheme: scheme, Cache: &client.CacheOptions{Reader: cache}}))
--- a/test/pkg/environment/common/environment.go
+++ b/test/pkg/environment/common/environment.go
@@ -133,16 +133,11 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
 		})
 		return []string{t.Value}
 	}))
-	// disruption status conditions
 	lo.Must0(cache.IndexField(ctx, &v1beta1.NodeClaim{}, "status.conditions[*].type", func(o client.Object) []string {
 		nodeClaim := o.(*v1beta1.NodeClaim)
-		conditionTypes := lo.FilterMap(nodeClaim.Status.Conditions, func(c status.Condition, _ int) (string, bool) {
-			if c.Type == v1beta1.ConditionTypeDrifted || c.Type == v1beta1.ConditionTypeEmpty || c.Type == v1beta1.ConditionTypeExpired {
-				return c.Type, true
-			}
-			return "", false
+		return lo.Map(nodeClaim.Status.Conditions, func(c status.Condition, _ int) string {
+			return c.Type
 		})
-		return conditionTypes
 	}))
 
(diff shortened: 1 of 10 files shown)


test/suites/perf/suite_test.go
```diff
@@ -55,7 +65,17 @@ var _ = BeforeEach(func() {
 	env.BeforeEach()
 	nodeClass = env.DefaultNodeClass()
 	nodePool = env.DefaultNodePool(nodeClass)
+	test.ReplaceRequirements(nodePool, v1beta1.NodeSelectorRequirementWithMinValues{
+		NodeSelectorRequirement: v1.NodeSelectorRequirement{
+			Key:      v1alpha1.InstanceSizeLabelKey,
+			Operator: v1.NodeSelectorOpLt,
+			Values:   []string{"32"},
+		},
+	})
```
njtran, line 74: Added this so that we just have a lower instance size and get multiple nodes on simple scale ups.

test/suites/perf/scheduling_test.go
```diff
@@ -44,18 +51,81 @@ var _ = Describe("Performance Benchmark", func() {
...
+					},
+				}})
+			env.ExpectCreated(deployment)
+			env.ExpectCreated(nodePool, nodeClass)
+			env.EventuallyExpectHealthyPodCount(labelSelector, replicas)
 
-			fmt.Println("--------- RESULTS ---------")
-			fmt.Printf("This is the duration: %s\n", duration)
+			env.TimeIntervalCollector.Start("Drift")
+			env.TimeIntervalCollector.End("Drift")
```
jonathan-innis, line 75: Seems like we probably don't want to call this here. I'm imagining that this was supposed to be a defer statement or should be dropped entirely because we already call it later in test

test/pkg/environment/common/environment.go
```diff
@@ -129,6 +133,14 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
 		})
 		return []string{t.Value}
 	}))
+	// drifted type
+	lo.Must0(cache.IndexField(ctx, &v1beta1.NodeClaim{}, "status.conditions[*].type", func(o client.Object) []string {
```
jonathan-innis, line 137: Should we name this something that indicates that this is matching on the drift condition? It seems odd to me that we would name the indexer like this when it is only matching on the Drift type

test/pkg/environment/common/expectations.go
```diff
@@ -623,75 +623,64 @@ func (env *Environment) EventuallyExpectInitializedNodeCount(comparator string,
...
 		g.Expect(env.Client.List(env.Context, nodeClaimList)).To(Succeed())
 		g.Expect(len(nodeClaimList.Items)).To(BeNumerically(comparator, count))
 	}).Should(Succeed())
-	return lo.Map(nodeClaimList.Items, func(nc corev1beta1.NodeClaim, _ int) *corev1beta1.NodeClaim {
+	return lo.Map(nodeClaimList.Items, func(nc v1beta1.NodeClaim, _ int) *v1beta1.NodeClaim {
 		return &nc
 	})
 }
 
-func (env *Environment) EventuallyExpectNodeClaimsReady(nodeClaims ...*corev1beta1.NodeClaim) {
```
jonathan-innis, line 646: Why are we dropping this?

test/pkg/environment/common/setup.go
```diff
@@ -59,16 +61,17 @@ var (
...
 	}
 )
 
 // nolint:gocyclo
 func (env *Environment) BeforeEach() {
 	debug.BeforeEach(env.Context, env.Config, env.Client)
-
 	// Expect this cluster to be clean for test runs to execute successfully
 	env.ExpectCleanCluster()
+	env.TimeIntervalCollector.Reset()
```
jonathan-innis, line 73: Should the reset be at the BeforeEach or in the AfterEach?

njtran: We actually call it in the Record() call in the AfterEach, so we don't need this at all.

test/pkg/environment/common/expectations.go
njtran: Removed it since we're technically not using it, but since we're keeping everything else around, i'll add it back in.

test/pkg/environment/common/environment.go
```diff
@@ -129,6 +133,17 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
 		})
 		return []string{t.Value}
 	}))
+	// disruption status conditions
+	lo.Must0(cache.IndexField(ctx, &v1beta1.NodeClaim{}, "status.conditions[*].type", func(o client.Object) []string {
+		nodeClaim := o.(*v1beta1.NodeClaim)
+		conditionTypes := lo.FilterMap(nodeClaim.Status.Conditions, func(c status.Condition, _ int) (string, bool) {
+			if c.Type == v1beta1.ConditionTypeDrifted || c.Type == v1beta1.ConditionTypeEmpty || c.Type == v1beta1.ConditionTypeExpired {
```
jonathan-innis, line 140: Why not all types?

test/pkg/environment/common/expectations.go
jonathan-innis: Got it. Yeah, could consider dropping it in a separate PR

jonathan-innis (approved):
/lgtm
/approve

Issue #1323, opened 2024-06-17
Karpenter does not disrupt node when changing annotation from "true" to "false"


iamhritik:
### Description

**Observed Behavior**: Karpenter does not disrupt the node and logs the following event:
`Cannot disrupt Node: Disruption is blocked with the "karpenter.sh/do-not-disrupt" annotation
`

**Expected Behavior**: Karpenter should allow disrupting the node when the annotation is set to karpenter.sh/do-not-disrupt: "false".

**Reproduction Steps** (Please include YAML):

- Annotate a nodepool with karpenter.sh/do-not-disrupt: "true":
- Change the annotation to karpenter.sh/do-not-disrupt: "false" in the nodepool, nodeclaim and the node:

**Versions**:
- Chart Version: 0.36.0
- Kubernetes Version (`kubectl version`): 1.28

jigisha620:
Do you mind sharing what your nodepool looks like?

iamhritik:
This is the NodePool YAML.
Initially, We've set the annotation to true, but now we have set it to false and also changed this annotation on the existing NodeClaims and Nodes but still getting this in the events :

`Cannot disrupt Node: Disruption is blocked with the "karpenter.sh/do-not-disrupt" annotation 
`

```
apiVersion: karpenter.sh/v1beta1
kind: NodePool
metadata:
  annotations:
    karpenter.sh/nodepool-hash: "9991694099770600129"
    karpenter.sh/nodepool-hash-version: v2
  labels:
    argocd.argoproj.io/instance: om-karpenter-deployment
  name: oa-nifi-nodes-prod
spec:
  disruption:
    budgets:
    - nodes: 20%
    - nodes: "1"
    consolidationPolicy: WhenUnderutilized
    expireAfter: 720h
  limits:
    cpu: 420
  template:
    metadata:
      annotations:
        karpenter.sh/do-not-disrupt: "false" -> initially it was true
        prod-ec1-eks: owned
      labels:
        Component: k8s
        Environment: prod
        Managed: karpenter
        Name: prod-ec1-eks-karpenter
        Project: Onemind
        Resource: eks
        Vertical: operations
    spec:
      nodeClassRef:
        name: om-oa-nifi-nodeclass
      requirements:
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
      - key: kubernetes.io/os
        operator: In
        values:
        - linux
      - key: karpenter.sh/capacity-type
        operator: In
        values:
[...]

PR #1331, opened 2024-06-18
chore: Validate do not disrupt annotation on node is true

jigisha620:
Fixes #1323

**Description**
Currently we don't validate if `do-not-disrupt` annotation on a node is set to `true` and only check if the annotation exists. This PR adds a change to add the validation.

Tested on local cluster. Added a functional test for the change.

### DIFF
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -185,7 +185,7 @@ func (in *StateNode) ValidateDisruptable(ctx context.Context, kubeClient client.
 	if in.Nominated() {
 		return nil, fmt.Errorf("state node is nominated for a pending pod")
 	}
-	if _, ok := in.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok {
+	if in.Annotations()[v1beta1.DoNotDisruptAnnotationKey] == "true" {
 		return nil, fmt.Errorf("disruption is blocked through the %q annotation", v1beta1.DoNotDisruptAnnotationKey)
 	}
 	// check whether the node has all the labels we need
--- a/pkg/controllers/disruption/drift_test.go
+++ b/pkg/controllers/disruption/drift_test.go
@@ -932,5 +932,29 @@ var _ = Describe("Drift", func() {
 			ExpectExists(ctx, env.Client, nodeClaim)
 			ExpectExists(ctx, env.Client, node)
 		})
+		It("should delete nodes with the karpenter.sh/do-not-disrupt annotation set to false", func() {
+			node.Annotations = lo.Assign(node.Annotations, map[string]string{v1beta1.DoNotDisruptAnnotationKey: "false"})
+			ExpectApplied(ctx, env.Client, nodeClaim, node, nodePool)
+
+			// inform cluster state about nodes and nodeclaims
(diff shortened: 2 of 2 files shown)


pkg/controllers/state/statenode.go
```diff
@@ -185,8 +185,10 @@ func (in *StateNode) ValidateDisruptable(ctx context.Context, kubeClient client.
 	if in.Nominated() {
 		return nil, fmt.Errorf("state node is nominated for a pending pod")
 	}
-	if _, ok := in.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok {
-		return nil, fmt.Errorf("disruption is blocked through the %q annotation", v1beta1.DoNotDisruptAnnotationKey)
+	if val, ok := in.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok {
+		if val == "true" {
+			return nil, fmt.Errorf("disruption is blocked through the %q annotation", v1beta1.DoNotDisruptAnnotationKey)
+		}
```
jmdeal, line 191: nit
```suggestion
	if val, ok := in.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok && val == "true" {
		return nil, fmt.Errorf("disruption is blocked through the %q annotation", v1beta1.DoNotDisruptAnnotationKey)
```

pkg/controllers/state/statenode.go
```diff
@@ -185,7 +185,7 @@ func (in *StateNode) ValidateDisruptable(ctx context.Context, kubeClient client.
 	if in.Nominated() {
 		return nil, fmt.Errorf("state node is nominated for a pending pod")
 	}
-	if _, ok := in.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok {
```
jonathan-innis, line 188: I think we just want to say that this is a presence or lack of presence thing. We do the same thing with `karpenter.sh/do-not-disrupt` where we just check for the presence of the annotation. 

Do we need users to be able to set this value to false? IMO, we just not let that mean anything. Let's update our docs to say that this is a presence/lack-of-presence kind of thing

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (commented):
/hold 

Add a test to validate this change

pkg/utils/pod/scheduling.go
```diff
@@ -168,7 +168,8 @@ func HasDoNotDisrupt(pod *v1.Pod) bool {
 	if pod.Annotations == nil {
 		return false
 	}
-	return pod.Annotations[v1beta1.DoNotDisruptAnnotationKey] == "true"
+	_, found := pod.Annotations[v1beta1.DoNotDisruptAnnotationKey]
+	return found
```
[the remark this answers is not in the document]
jigisha620: Yes, working on adding the tests.

pkg/controllers/state/statenode.go
```diff
@@ -185,7 +185,7 @@ func (in *StateNode) ValidateDisruptable(ctx context.Context, kubeClient client.
...
+	if val, ok := in.Annotations()[v1beta1.DoNotDisruptAnnotationKey]; ok && val == "true" {
```
ellistarn, line 188: Do you even need to check ok? If you access the map and it's empty, it should be "", so the comparison will still work

ellistarn (approved):

pkg/controllers/state/statenode.go
jigisha620: Agreed. Updated.

ellistarn (approved):

PR #1333, opened 2024-06-19
BREAKING: revert back to forceful expiration by default

Bryce-Soghigian:
Fixes #N/A 

**Description**
This PR is an implementation of the preferred implementation of the [Forceful Expiration Design](https://github.com/kubernetes-sigs/karpenter/pull/1303). 

We remove status condition 'expired', node disruption budgets for 'expired' and change the behavior to immediately delete the nodeclaim when we see a nodeclaim is past its expiry ttl.
- make presubmit


pkg/controllers/disruption/suite_test.go
```diff
@@ -281,7 +281,7 @@ var _ = Describe("Simulate Scheduling", func() {
 
 		// Mark all nodeclaims as expired
```
njtran, line 282: probably want to adjust this comment too

pkg/controllers/nodeclaim/disruption/expiration.go
njtran, line 1: We should probably remove the status condition bits, and just simply delete the node if it's expired, and exit if it's not.

pkg/controllers/disruption/suite_test.go
Bryce-Soghigian, line 282:
```suggestion
		// Mark all nodeclaims as drifted
```

pkg/controllers/nodeclaim/disruption/expiration.go
```diff
@@ -36,36 +36,28 @@ type Expiration struct {
...
 	if nodePool.Spec.Disruption.ExpireAfter.Duration == nil {
-		_ = nodeClaim.StatusConditions().Clear(v1beta1.ConditionTypeExpired)
-		if hasExpiredCondition {
-			log.FromContext(ctx).V(1).Info("removing expiration status condition, expiration has been disabled")
-		}
 		return reconcile.Result{}, nil
 	}
 	expirationTime := nodeClaim.CreationTimestamp.Add(*nodePool.Spec.Disruption.ExpireAfter.Duration)
-	// 2. If the NodeClaim isn't expired, remove the status condition.
+	// 2. If the NodeClaim isn't expired, remove the status condition and leave the reconcile loop
```
Bryce-Soghigian, line 45:
```suggestion
	// 2. If the NodeClaim isn't expire leave the reconcile loop.
```

pkg/controllers/nodeclaim/disruption/expiration.go
```diff
@@ -36,36 +36,28 @@ type Expiration struct {
...
 	expirationTime := nodeClaim.CreationTimestamp.Add(*nodePool.Spec.Disruption.ExpireAfter.Duration)
-	// 2. If the NodeClaim isn't expired, remove the status condition.
+	// 2. If the NodeClaim isn't expire leave the reconcile loop.
 	if e.clock.Now().Before(expirationTime) {
-		_ = nodeClaim.StatusConditions().Clear(v1beta1.ConditionTypeExpired)
-		if hasExpiredCondition {
-			log.FromContext(ctx).V(1).Info("removing expired status condition, not expired")
-		}
-		// If the NodeClaim isn't expired and doesn't have the status condition, return.
+		// If the NodeClaim isn't expired, return.
```
Bryce-Soghigian, line 47:
```suggestion
```

Bryce-Soghigian:
I removed the status conditions all together for expiration, and refactored a couple of the tests so we don't lose coverage(some of them were just using expiry as a replace action)

pkg/controllers/nodeclaim/disruption/expiration_test.go
```diff
@@ -64,38 +63,30 @@ var _ = Describe("Expiration", func() {
...
-		nodeClaim.StatusConditions().SetTrue(v1beta1.ConditionTypeExpired)
 		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
-
 		ExpectObjectReconciled(ctx, env.Client, nodeClaimDisruptionController, nodeClaim)
-
 		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
-		Expect(nodeClaim.StatusConditions().Get(v1beta1.ConditionTypeExpired)).To(BeNil())
 	})
-	It("should mark NodeClaims as expired if the nodeClaim is expired but the node isn't", func() {
+	It("should delete NodeClaims as expired if the nodeClaim is expired but the node isn't", func() {
```
njtran, line 89: nit:
```suggestion
	It("should delete NodeClaims if the nodeClaim is expired but the node isn't", func() {
```

pkg/controllers/nodeclaim/disruption/expiration.go
```diff
@@ -36,36 +36,26 @@ type Expiration struct {
...
-		metrics.NodeClaimsDisruptedCounter.With(prometheus.Labels{
-			metrics.TypeLabel:     metrics.ExpirationReason,
-			metrics.NodePoolLabel: nodeClaim.Labels[v1beta1.NodePoolLabelKey],
-		}).Inc()
+	// 3. Otherwise, if the NodeClaim is expired we can forcefully expire the nodeclaim (by deleting it)
+	if err := e.kubeClient.Delete(ctx, nodeClaim); err != nil {
+		return reconcile.Result{}, err
 	}
+	// At this point we know we would have returned
+	log.FromContext(ctx).V(1).Info("deleting expired nodeclaim")
```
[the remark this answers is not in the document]
Bryce-Soghigian: Agreed it would make more sense in general for us to have the termination metrics only after successful termination.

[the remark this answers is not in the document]
Bryce-Soghigian: Added the behavior and a test validating that same behavior

pkg/controllers/nodeclaim/disruption/expiration.go
```diff
@@ -36,36 +36,31 @@ type Expiration struct {
...
+	// 4. The deletion timestamp has successfully been set for the NodeClaim, update relevant metrics.
+	log.FromContext(ctx).V(1).Info("deleting expired nodeclaim")
+	metrics.NodeClaimsDisruptedCounter.With(prometheus.Labels{
+		metrics.TypeLabel:     metrics.ExpirationReason,
+		metrics.NodePoolLabel: nodeClaim.Labels[v1beta1.NodePoolLabelKey],
+	}).Inc()
+	metrics.NodeClaimsTerminatedCounter.With(prometheus.Labels{
+		metrics.ReasonLabel:       metrics.ExpirationReason,
+		metrics.NodePoolLabel:     nodeClaim.Labels[v1beta1.NodePoolLabelKey],
+		metrics.CapacityTypeLabel: nodeClaim.Labels[v1beta1.CapacityTypeLabelKey],
```
njtran, line 63: This is super small, but we don't check the existence of this label in our controller code. I don't think the lack of capacity type label should block us from deleting this, but we should have some good null value. I was thinking something like `n/a` or `-`, so that we can aggregate on the empty value.

njtran: I can see your point, but I think it's probably easiest for this PR to add the metric in here, and refactoring this to the termination controller later. I think we should add in the null value label though.

njtran (commented):
/hold
/lgtm
/approve
approving but holding so @Bryce-Soghigian can unhold when he's ready

PR #1335, opened 2024-06-19
test: add output_dir env variable for timestamp exports

njtran:
Fixes #N/A 

**Description**
Adds the ability to set OUTPUT_DIR with testing framework for timestamp exports

OUTPUT_DIR=$(mktemp -d) make e2etests

### DIFF
--- a/test/pkg/debug/collector.go
+++ b/test/pkg/debug/collector.go
@@ -142,10 +142,16 @@ func PrintTestTimes(times map[string][]TimeInterval) map[string][][]string {
 }
 
 // WriteTimestamps will create a temp directory and a .csv file for each suite test
+// If the OUTPUT_DIR environment variable is set, we'll print the csvs to that directory.
 func WriteTimestamps(path string, timestamps *TimeIntervalCollector) error {
-	directory, err := os.MkdirTemp("/tmp", "")
-	if err != nil {
-		return err
+	var directory string
+	var err error
+	directory, ok := os.LookupEnv("OUTPUT_DIR")
+	if !ok {
+		directory, err = os.MkdirTemp("/tmp", "")
+		if err != nil {
+			return err
+		}
 	}
 	for name, table := range PrintTestTimes(timestamps.suiteTimeIntervals) {
 		file, err := os.CreateTemp(directory, fmt.Sprintf("*-%s.csv", name))
--- a/test/pkg/debug/collector.go
+++ b/test/pkg/debug/collector.go
@@ -143,18 +143,16 @@ func PrintTestTimes(times map[string][]TimeInterval) map[string][][]string {
 
 // WriteTimestamps will create a temp directory and a .csv file for each suite test
 // If the OUTPUT_DIR environment variable is set, we'll print the csvs to that directory.
-func WriteTimestamps(path string, timestamps *TimeIntervalCollector) error {
-	var directory string
+func WriteTimestamps(outputDir string, timestamps *TimeIntervalCollector) error {
 	var err error
-	directory, ok := os.LookupEnv("OUTPUT_DIR")
-	if !ok {
-		directory, err = os.MkdirTemp("/tmp", "")
+	if outputDir == "" {
+		outputDir, err = os.MkdirTemp("/tmp", "")
 		if err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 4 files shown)


test/suites/perf/scheduling_test.go
```diff
@@ -33,7 +33,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/test"
 )
 
-var replicas int = 100
+var replicas int = 10
```
njtran, line 36: scaling this back down to account for some kwok scalabilty issues at the moment.

test/pkg/debug/collector.go
```diff
@@ -142,10 +142,16 @@ func PrintTestTimes(times map[string][]TimeInterval) map[string][][]string {
...
 
 // WriteTimestamps will create a temp directory and a .csv file for each suite test
+// If the OUTPUT_DIR environment variable is set, we'll print the csvs to that directory.
 func WriteTimestamps(path string, timestamps *TimeIntervalCollector) error {
-	directory, err := os.MkdirTemp("/tmp", "")
-	if err != nil {
-		return err
+	var directory string
+	var err error
+	directory, ok := os.LookupEnv("OUTPUT_DIR")
```
jonathan-innis, line 149: Generally, it's a good idea to ensure that environment variables are set at the top-level in startup and then pass the config down through DI or through contexts.

If we scale-out to add more environment variables, it's probably preferable that we would grab them all in the same place rather than grabbing them at runtime

test/pkg/debug/collector.go
```diff
@@ -142,13 +142,17 @@ func PrintTestTimes(times map[string][]TimeInterval) map[string][][]string {
...
+	var err error
+	if outputDir == "" {
+		outputDir, err = os.MkdirTemp("/tmp", "")
+		if err != nil {
+			return err
+		}
 	}
 	for name, table := range PrintTestTimes(timestamps.suiteTimeIntervals) {
-		file, err := os.CreateTemp(directory, fmt.Sprintf("*-%s.csv", name))
+		file, err := os.CreateTemp(outputDir, fmt.Sprintf("*-%s.csv", name))
```
jonathan-innis, line 155: Should you be using createtmp if someone specifies a directory that they want to be keeping around? I'm not sure what the impact of adding a temp file is like this if someone expects the file to live for a longer period of time. I believe that the kernel can come by later and cleanup the file if it's marked as temporary

test/pkg/debug/collector.go
```diff
@@ -142,13 +142,17 @@ func PrintTestTimes(times map[string][]TimeInterval) map[string][][]string {
...
 
 // WriteTimestamps will create a temp directory and a .csv file for each suite test
-func WriteTimestamps(path string, timestamps *TimeIntervalCollector) error {
-	directory, err := os.MkdirTemp("/tmp", "")
-	if err != nil {
-		return err
+// If the OUTPUT_DIR environment variable is set, we'll print the csvs to that directory.
+func WriteTimestamps(outputDir string, timestamps *TimeIntervalCollector) error {
+	var err error
+	if outputDir == "" {
```
jonathan-innis, line 148: If the output dir doesn't exist, do we just want to create the whole path on behalf of the user, regardless of whether it's using the default or not?

jonathan-innis (approved):
/lgtm
/approve

test/pkg/debug/collector.go
njtran: I think the best we could do here is inferring if the string starts with `/tmp` that it's a temporary directory, and otherwise isn't. I'm expecting this to be containerized and ephemeral, and I'd expect automation to copy these files elsewhere for permanent storage. I also think files in /tmp/ are cleaned up after 10d anyway.

If the /tmp check is good enough for you, I don't mind adding that in

test/pkg/debug/collector.go
njtran: I think we should expect that a user has preprovisioned the directory, and wouldn't expect us to recursively create the directories for them.

test/pkg/debug/collector.go
njtran: Going to resolve this and remove the hold since I think the expectation is that we'll be only using these files temporarily for now. We can always add the functionality in later.

Issue #1049, opened 2024-02-28
Karpenter should taint nodes on startup with `karpenter.sh/unregistered:NoSchedule` and remove once the node is registered


jonathan-innis:
### Description

**What problem are you trying to solve?**

Karpenter currently doesn't taint nodes in its userData, but it chooses to propagate some labels, annotations, etc. down to nodes when they join the cluster. There is currently a race that is present between the Karpenter controller's apply of the labels to the node and the kube-scheduler being able to bind pods against the node. It's possible that Karpenter may not have applied labels yet when the kube-scheduler determines that it can start to schedule pods against the node.

This problem is more prevalent with DaemonSets, where they can tolerate the [`node.kubernetes.io/not-ready`](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-not-ready) taint, meaning that they can be scheduled against the node as soon as it joins. This is definitely less of a problem for application pods that have to wait for the node to go ready, but can still be an issue if the Karpenter controller is down.

This also isn't a large issue, in general, since the Karpenter cloudproviders will generally pass all of the labels down from a NodeClaim into the relevant startup script that is passed to the kubelet. This means that those labels will be there atomically on kubelet start, which, again, removes the chance that this race can occur.

Realistically, we shoudn't have to rely on this behavior and should just take a stance, similar to the [external cloudprovider implementation with its taint](https://kubernetes.io/docs/tasks/administer-cluster/running-cloud-controller/#running-cloud-controller-manager) that a node gets tainted on startup with a known taint and Karpenter removes the taint from the node once it has finished its registration process.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
There is some relevant conversation around this topic that I found [here](https://github.com/kubernetes/kubeadm/issues/2509) as I was exploring what it would look like if Karpenter were to allow users to specify the `node-role.kubernetes.io` label in the NodePool configuration.

sftim:
- I think it's reasonable to taint with the `NoExecute` form as well. Pods that want to tolerate it can do.
[...]

PR #1336, opened 2024-06-19
feat: require and then remove karpenter startup taint

rschalo:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1049 

**Description**
Based on implementation discussion, it is required that userdata scripts for starting kubelet include `karpenter.sh/unregistered=true:NoExecute` as part of the `--register-with-taints` flag.

Unit testing and local `aws/karpenter-provider-aws` deployment. This will be ship with https://github.com/aws/karpenter-provider-aws/pull/6388/


pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -80,6 +80,10 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeCla
 
 func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1beta1.NodeClaim, node *v1.Node) error {
 	stored := node.DeepCopy()
+	// Strip the resource version off the original node. This ensures the patch will fail with a conflict if the node
+	// had been updated between the original GET and the PATCH. This is necessary because .spec.taints does not specify
+	// a patchMergeKey and patchStrategy. Ref: https://github.com/kubernetes/kubernetes/pull/113136
+	stored.ResourceVersion = ""
```
jonathan-innis, line 86: Can we consider just doing an Update here rather than leveraging this "hack" on the PATCH API in upstream? Though I think it is a cool thing to use, I question whether it makes more sense for us to just use the UPDATE API since this behavior is well-known, including it's well-known how conflicts get handled with these calls

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -88,13 +92,35 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1beta1.NodeClai
 	// Sync all taints inside NodeClaim into the Node taints
 	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.Taints)
 	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.StartupTaints)
+	// Remove karpenter.sh/unregistered taint
+	taints, err := isolateKarpenterStartupTaint(node.Spec.Taints)
+	if err != nil {
+		return fmt.Errorf("missing required startup taint: %w", err)
```
jonathan-innis, line 98:
```suggestion
		return fmt.Errorf("missing required startup taint, %w", err)
```

nit: style, prefer using a comma on error wrapping

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -88,13 +92,35 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1beta1.NodeClai
 	// Sync all taints inside NodeClaim into the Node taints
 	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.Taints)
 	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.StartupTaints)
+	// Remove karpenter.sh/unregistered taint
+	taints, err := isolateKarpenterStartupTaint(node.Spec.Taints)
```
jonathan-innis, line 96: lo.Reject?

pkg/controllers/nodeclaim/lifecycle/registration.go
jonathan-innis, line 98: Should we do the check on this taint at the top of the sub-reconcile (right after we do the condition check)? This would allow us to short-circuit so all we have to do is remove the taint in the `syncNode` call.

pkg/test/nodeclaim.go
```diff
@@ -48,6 +48,7 @@ func NodeClaim(overrides ...v1beta1.NodeClaim) *v1beta1.NodeClaim {
 	if override.Spec.Requirements == nil {
 		override.Spec.Requirements = []v1beta1.NodeSelectorRequirementWithMinValues{}
 	}
+	override.Spec.StartupTaints = append(override.Spec.StartupTaints, v1beta1.UnregisteredNoExecuteTaint)
```
jonathan-innis, line 51: This seems unexpected to me. Since we are artificially adding this taint from the CloudProvider side, is it possible that the node just registers with this taint, even if the NodeClaim didn't store it?

pkg/test/nodes.go
```diff
@@ -68,14 +69,17 @@ func Node(overrides ...NodeOptions) *v1.Node {
 }
 
 func NodeClaimLinkedNode(nodeClaim *v1beta1.NodeClaim) *v1.Node {
+	taints := lo.Filter(nodeClaim.Spec.StartupTaints, func(t v1.Taint, _ int) bool {
```
jonathan-innis, line 72: This filters out the taint from the NodeClaim startup taints, but I'm curious if we should have this taint be added to nodes always to properly mock the true behavior of what happens from a CloudProvider and then orchestrate the removal from the controller.

This also prevents you having to add it in the StartupTaints section of the NodeClaim

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -133,6 +133,7 @@ var _ = Context("Scheduling", func() {
 								},
 							},
 						},
+						StartupTaints: []v1.Taint{v1beta1.UnregisteredNoExecuteTaint},
```
jonathan-innis, line 136: I don't think that we should have to manually add this as a startup taint on the NodePool; however, we should make sure that we properly consider the taints correctly using the cluster state node before the node has been marked as registered

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -82,6 +82,7 @@ func (i *NodeClaimTemplate) ToNodeClaim(nodePool *v1beta1.NodePool) *v1beta1.Nod
 		},
 		Spec: i.Spec,
 	}
+	nc.Spec.StartupTaints = append(nc.Spec.StartupTaints, v1beta1.UnregisteredNoExecuteTaint)
```
jonathan-innis, line 85: Same comment here, I think we should avoid adding this as an actual startup taint on the NodeClaim, and this being injected into the Node userData by the CloudProvider should be implicit to the implementation

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -80,6 +80,10 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeCla
 
 func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1beta1.NodeClaim, node *v1.Node) error {
```
jonathan-innis, line 81: Are there any updates to the registration_test.go file that need to happen with this change?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -88,13 +92,35 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1beta1.NodeClai
...
+	node.Spec.Taints = taints
 	if !equality.Semantic.DeepEqual(stored, node) {
 		if err := r.kubeClient.Patch(ctx, node, client.StrategicMergeFrom(stored)); err != nil {
 			return fmt.Errorf("syncing node labels, %w", err)
 		}
 	}
 	return nil
 }
+
+func isolateKarpenterStartupTaint(ts []v1.Taint) ([]v1.Taint, error) {
```
njtran, line 112: Is this "return all taints that have the same key"? or is it "return the taint that exactly matches this node"? 

If it's the latter, we should just be able to replace this with a `lo.Find()`, since there shouldn't be any collisions. If it's the former, we should use `lo.Filter()`

njtran: I also don't think it needs to be an error, right? If the taint was already removed/DNE, we just succeed, so this can just be a boolean check + short circuit

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
njtran, line 85: Do we need to add this into the spec if it's in our known ephemeral taints?

pkg/test/nodes.go
```diff
@@ -68,14 +69,17 @@ func Node(overrides ...NodeOptions) *v1.Node {
...
+		return !t.MatchTaint(&v1beta1.UnregisteredNoExecuteTaint)
+	})
```
njtran, line 74: I'm wondering why we need this?

pkg/controllers/nodeclaim/lifecycle/registration.go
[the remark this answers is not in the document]
rschalo: Correct, unsetting it for a patch forces apiserver to get the latest resourceVersion when performing the patch. However, we will move away from using patch and will use update. If a conflict is found then we can requeue and try again.

pkg/controllers/nodeclaim/lifecycle/registration.go
rschalo: Yea, moving towards `lo.Reject()` and then using a short circuit after getting the node post status condition checks.

pkg/test/nodes.go
rschalo: Removed.

pkg/test/nodeclaim.go
rschalo: No longer going this route.

pkg/controllers/provisioning/scheduling/suite_test.go
rschalo: Done.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
rschalo: Done.

rschalo: No longer being added here.

pkg/controllers/nodeclaim/lifecycle/registration.go
rschalo: Checking after getting node from nodeclaim and before syncing the node.

pkg/controllers/nodeclaim/lifecycle/registration.go
rschalo: Yup

pkg/controllers/nodeclaim/lifecycle/registration.go
rschalo: What's our strategy for requeue limits? Until it hits the registration TTL and is then deleted?

pkg/controllers/nodeclaim/lifecycle/registration.go
rschalo: Added.

pkg/apis/v1beta1/taints.go
```diff
@@ -20,8 +20,10 @@ import v1 "k8s.io/api/core/v1"
 
 // Karpenter specific taints
 const (
-	DisruptionTaintKey             = Group + "/disruption"
-	DisruptingNoScheduleTaintValue = "disrupting"
+	DisruptionTaintKey              = Group + "/disruption"
+	DisruptingNoScheduleTaintValue  = "disrupting"
+	UnregisteredTaintKey            = Group + "/unregistered"
+	UnregisteredNoExecuteTaintValue = "true"
```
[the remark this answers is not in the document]
rschalo: Would it make more sense for it to be `preregistered`?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -61,9 +61,14 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeCla
...
 	}
+	if _, hasStartupTaint := lo.Find(node.Spec.Taints, func(t v1.Taint) bool {
+		return t.MatchTaint(&v1beta1.UnregisteredNoExecuteTaint)
+	}); !hasStartupTaint {
+		return reconcile.Result{}, fmt.Errorf("missing required startup taint, %s", v1beta1.UnregisteredTaintKey)
+	}
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef("", node.Name)))
 	if err = r.syncNode(ctx, nodeClaim, node); err != nil {
-		return reconcile.Result{}, fmt.Errorf("syncing node, %w", err)
+		return reconcile.Result{Requeue: true}, err
```
jmdeal, line 71: We should automatically requeue with backoff if the error is not true, is there a reason we're setting `Requeue` explicitly here?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -2069,7 +2069,8 @@ var _ = Context("Scheduling", func() {
 				// delete the pod so that the node is empty
 				ExpectDeleted(ctx, env.Client, initialPod)
 				// startup taint + node not ready taint = 2
```
jmdeal, line 2071:
```suggestion
				// startup taint + node not ready taint + unregistered taint = 3
```

pkg/controllers/nodeclaim/lifecycle/registration.go
rschalo: In this case we're returning the error though so don't we have to set requeue?

jmdeal: Yep, if the error is non-nil the result is ignored and it enters expoenential backoff: [ref](https://github.com/kubernetes-sigs/controller-runtime/blob/12cc8d59fabe3a1dbbd927ae6aea077cf7962267/pkg/reconcile/reconcile.go#L95-L97)

pkg/controllers/nodeclaim/disruption/suite_test.go
```diff
@@ -113,6 +113,7 @@ var _ = Describe("Disruption", func() {
 		Expect(nodeClaim.StatusConditions().Get(v1beta1.ConditionTypeEmpty).IsTrue()).To(BeTrue())
 	})
 	It("should remove multiple disruption conditions simultaneously", func() {
+		cp.Drifted = ""
```
jmdeal, line 116: I'm wondering if this was an intentional "feature" of the test, to verify that we could remove multiple disruption conditions **without** removing all conditions? Or was this change related to your other changes and I'm not realizing it?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -88,12 +93,17 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1beta1.NodeClai
 	// Sync all taints inside NodeClaim into the Node taints
 	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.Taints)
 	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.StartupTaints)
+	// Remove karpenter.sh/unregistered taint
+	taints := lo.Reject(node.Spec.Taints, func(t v1.Taint, _ int) bool {
```
jmdeal, line 97: nit: any reason we can't just assign directly?
```suggestion
	node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t v1.Taint, _ int) bool {
```

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -85,12 +85,32 @@ var _ = Describe("Registration", func() {
...
+				},
+			},
+		})
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		node := test.Node(test.NodeOptions{ProviderID: nodeClaim.Status.ProviderID, Taints: []v1.Taint{v1beta1.UnregisteredNoExecuteTaint}})
+		ExpectApplied(ctx, env.Client, node)
+		Expect(node.Spec.Taints).To(ContainElement(v1beta1.UnregisteredNoExecuteTaint))
```
jmdeal, line 108: I think we can drop this expectation, this really only checks that the `test.Node` constructor properly adds the taint to the object. We'd have to requery the node via `ExpectExists` to actually see if the taint was applied to the API server. I don't think it's necessary either way though.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -2069,6 +2069,7 @@ var _ = Context("Scheduling", func() {
...
+				// note that this node is not post registration so the unregistered taint is present
```
jmdeal, line 2072: Is the unregistered taint still present or are you no longer applying it? Looks like the length is now still 2.

pkg/test/nodes.go
```diff
@@ -67,15 +68,23 @@ func Node(overrides ...NodeOptions) *v1.Node {
 	}
 }
 
+// this check is done for tests that assume the nodeclaim is past registration but will not go through reconciliation
```
jmdeal, line 71: Can we specify that it won't be reconclied by the NodeClaim Lifecycle Controller?

pkg/test/nodes.go
```diff
@@ -67,15 +68,23 @@ func Node(overrides ...NodeOptions) *v1.Node {
...
 func NodeClaimLinkedNode(nodeClaim *v1beta1.NodeClaim) *v1.Node {
+	taints := lo.Reject(nodeClaim.Spec.StartupTaints, func(t v1.Taint, _ int) bool {
```
jmdeal, line 73: Do we expect the unregistered taint to be in the startup taints anymore?

rschalo: Nope.

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -61,9 +61,14 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1beta1.NodeCla
 		}
 		return reconcile.Result{}, fmt.Errorf("getting node for nodeclaim, %w", err)
 	}
+	if _, hasStartupTaint := lo.Find(node.Spec.Taints, func(t v1.Taint) bool {
```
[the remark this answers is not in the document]
jonathan-innis: I think you need to rely on the `karpenter.sh/registered` label to ensure that you are properly doing the check and you aren't relying on a non-atomic operation for doing the fail-closed check

pkg/test/nodes.go
```diff
@@ -67,15 +67,21 @@ func Node(overrides ...NodeOptions) *v1.Node {
 	}
 }
 
+// this check is done for tests that assume the nodeclaim is past registration but will not be reconciled by the NodeClaim Lifecycle controller
 func NodeClaimLinkedNode(nodeClaim *v1beta1.NodeClaim) *v1.Node {
+	taints := append(nodeClaim.Spec.StartupTaints, nodeClaim.Spec.Taints...)
```
jonathan-innis, line 72: nit: Have we thought about keeping the old code, assigning Node to `n` and then appending the taints after creating the node object all in one go

pkg/test/expectations/expectations.go
```diff
@@ -453,7 +453,7 @@ func ExpectMakeNodesReady(ctx context.Context, c client.Client, nodes ...*v1.Nod
 		// Remove any of the known ephemeral taints to make the Node ready
 		nodes[i].Spec.Taints = lo.Reject(nodes[i].Spec.Taints, func(taint v1.Taint, _ int) bool {
 			_, found := lo.Find(pscheduling.KnownEphemeralTaints, func(t v1.Taint) bool {
-				return t.MatchTaint(&taint)
+				return t.MatchTaint(&taint) && !t.MatchTaint(&v1beta1.UnregisteredNoExecuteTaint)
```
jonathan-innis, line 456: When we make the node ready, you're saying here that we don't want to remove the unregistered taint?

pkg/controllers/provisioning/suite_test.go
```diff
@@ -259,6 +259,7 @@ var _ = Describe("Provisioning", func() {
 				},
 				Finalizers: []string{v1beta1.TerminationFinalizer},
 			},
+			Taints: []v1.Taint{v1beta1.UnregisteredNoExecuteTaint},
```
jonathan-innis, line 262: Just curious: Why is this the only place in the scheduler code that requires the addition of this taint?

pkg/test/expectations/expectations.go
rschalo: Refactored so that the tests are running reconciliation more often.

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -85,11 +85,46 @@ var _ = Describe("Registration", func() {
...
 		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 
-		node := test.Node(test.NodeOptions{ProviderID: nodeClaim.Status.ProviderID})
+		node := test.Node(test.NodeOptions{ProviderID: nodeClaim.Status.ProviderID, Taints: []v1.Taint{v1beta1.UnregisteredNoExecuteTaint}})
+		ExpectApplied(ctx, env.Client, node)
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+		node = ExpectExists(ctx, env.Client, node)
+		Expect(node.Labels).To(HaveKeyWithValue(v1beta1.NodeRegisteredLabelKey, "true"))
+	})
+	It("should remove the karpenter.sh/unregistered taint from the Node when the Node comes online", func() {
```
njtran, line 94: If you want to reduce total number of tests, you could just also assert that the taint doesn't exist in the test before this, since it's effectively the same test, but with a different end assertion which the two aren't mutually exclusive assertions.

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -85,11 +85,46 @@ var _ = Describe("Registration", func() {
...
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		node := test.Node(test.NodeOptions{ProviderID: nodeClaim.Status.ProviderID, Taints: []v1.Taint{v1beta1.UnregisteredNoExecuteTaint}})
 		ExpectApplied(ctx, env.Client, node)
 		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
 		node = ExpectExists(ctx, env.Client, node)
 		Expect(node.Labels).To(HaveKeyWithValue(v1beta1.NodeRegisteredLabelKey, "true"))
+		Expect(node.Spec.Taints).To(Not(ContainElement(v1beta1.UnregisteredNoExecuteTaint)))
+	})
+	It("should fail registration if the karpenter.sh/unregistered taint is not present on the node", func() {
```
njtran, line 113:
```suggestion
	It("should fail registration if the karpenter.sh/unregistered taint is not present on the node and the node isn't labeled as registered", func() {
```

pkg/test/nodes.go
```diff
@@ -67,18 +67,22 @@ func Node(overrides ...NodeOptions) *v1.Node {
...
 				Finalizers:  nodeClaim.Finalizers,
 			},
-			Taints:      append(nodeClaim.Spec.Taints, nodeClaim.Spec.StartupTaints...),
+			Taints:      taints,
 			Capacity:    nodeClaim.Status.Capacity,
 			Allocatable: nodeClaim.Status.Allocatable,
 			ProviderID:  nodeClaim.Status.ProviderID,
 		},
 	)
+	n.Spec.Taints = append(n.Spec.Taints, v1beta1.UnregisteredNoExecuteTaint)
```
njtran, line 86: does it matter if the unregistered taint is the last one or not? Seems that's the only reason why I'd want to have two different taint append calls here.

rschalo: Just in response to comment about reducing diff in this file from @jonathan-innis  above

pkg/controllers/nodeclaim/lifecycle/registration_test.go
rschalo: Great point, making the change.

njtran (commented):
/lgtm
/approve

PR #1343, opened 2024-06-20
fix: add nil check for required nodeclassref field

jmdeal:
Fixes #N/A 

**Description**
Adds explicit `nil` checks when we set up the indexers for nodeclaim nodeclassrefs. This fields is required, and that requirement should be enforced through validation, but it is possible to circumvent navigation and remove this field (e.g. editing the CEL or removing the validating webhook). This change makes Karpenter fail open rather than crash in this scenario.

Manual validation

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -19,6 +19,7 @@ package operator
 import (
 	"context"
 	"fmt"
+	golog "log"
 	"net/http"
 	"net/http/pprof"
 	"runtime"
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -194,13 +195,25 @@ func NewOperator() (context.Context, *Operator) {
 		return []string{o.(*v1beta1.NodeClaim).Status.ProviderID}
 	}), "failed to setup nodeclaim provider id indexer")
 	lo.Must0(mgr.GetFieldIndexer().IndexField(ctx, &v1beta1.NodeClaim{}, "spec.nodeClassRef.apiVersion", func(o client.Object) []string {
-		return []string{o.(*v1beta1.NodeClaim).Spec.NodeClassRef.APIVersion}
+		nc := o.(*v1beta1.NodeClaim)
+		if nc.Spec.NodeClassRef == nil {
```
engedaam, line 199: Should we validate that Spec is defined?

engedaam: Should we expanded this check for all the indexers?

jmdeal: Spec isn't a pointer, this is really just meant to ensure the error we get when this isn't set isn't just a stack trace. For the required, non pointer types if they're unset because you circumvented validation I think it's fine to just have UB. Really I don't think there's anything strictly wrong with how we crash currently, this is just a free QoL upgrade.

pkg/operator/operator.go
```diff
@@ -194,13 +194,25 @@ func NewOperator() (context.Context, *Operator) {
 		return []string{o.(*v1beta1.NodeClaim).Status.ProviderID}
 	}), "failed to setup nodeclaim provider id indexer")
 	lo.Must0(mgr.GetFieldIndexer().IndexField(ctx, &v1beta1.NodeClaim{}, "spec.nodeClassRef.apiVersion", func(o client.Object) []string {
-		return []string{o.(*v1beta1.NodeClaim).Spec.NodeClassRef.APIVersion}
+		nc := o.(*v1beta1.NodeClaim)
```
jonathan-innis, line 197: Can we validate what the result of this happening is? I'm assuming that we don't crash, but do we clean things up and throw errors properly?

jonathan-innis: Can we add a test into Kwok or AWS provider (or both) for validating that we don't crash here?

jmdeal: In this case, there's no error to throw since we're failing open. This is consistent with other places in the project that we check the `NodeClassRef`. I can add a test in the AWS provider since we don't have e2e testing wired fully up with KWOK yet.

jonathan-innis:
I question whether this is needed. Can we just assume that the user has not messed with the validations on the CRD and assume that this is an invalid state that we expect to cause a crash in Karpenter.

jmdeal:
Discussed further offline, I'm going to close this out. This failure mode should only occur when validation is intentionally altered (you would need to edit the CEL). In that case, crashing and giving a loud signal to the user that they've broke an invariant is probably the best course of action.

That does leave us with the option of continuing to panic vs a `log.Fatal`, like originally implemented in this PR. Opting to keep the panic since it provides a stack trace, and it's pretty easy to track down the context since the source code is available.

PR #1344, opened 2024-06-21
chore: Increase nodeClaim spec.requirements to 100

jigisha620:
Fixes #N/A 

**Description**
In response to the issue - https://github.com/aws/karpenter-provider-aws/issues/6387
Currently nodeclaim `spec.requirements` is restricted to 30. This PR increases the limit to 100.

Tested on local cluster and also added a functional test for this.


pkg/apis/v1beta1/nodeclaim.go
```diff
@@ -37,7 +37,7 @@ type NodeClaimSpec struct {
 	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
 	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt' or 'Lt' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
 	// +kubebuilder:validation:XValidation:message="requirements with 'minValues' must have at least that many values specified in the 'values' field",rule="self.all(x, (x.operator == 'In' && has(x.minValues)) ? x.values.size() >= x.minValues : true)"
-	// +kubebuilder:validation:MaxItems:=30
+	// +kubebuilder:validation:MaxItems:=100
```
jonathan-innis, line 40: Do we have this same validation for the webhook or is this just specific to CEL?

jonathan-innis: Are we going to take this change into the v1 API since this change won't get propagated over if we don't add it there now. You may need to create at PR against the staging branch to add this change

jonathan-innis (commented):
Seems like this works across all K8s versions? If so, seems completely reasonable.

pkg/apis/v1beta1/nodeclaim.go
engedaam: Should we make the same changes to the NodePool?

engedaam: Nvm, this will do that

jigisha620: > Do we have this same validation for the webhook or is this just specific to CEL?

Nope, we don't.

>Are we going to take this change into the v1 API since this change won't get propagated over if we don't add it there now. You may need to create at PR against the staging branch to add this change

I will create another PR against staging as well.

jonathan-innis (approved):
/lgtm
/approve

PR #1351, opened 2024-06-26
feat: drop webhook for v1

rschalo:
Fixes #N/A 

**Description**
This is part of dropping knative from the project.

Unit tests.

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -166,34 +166,48 @@ func (p *Provisioner) GetPendingPods(ctx context.Context) ([]*v1.Pod, error) {
 	if err != nil {
 		return nil, fmt.Errorf("listing pods, %w", err)
 	}
-	return lo.Reject(pods, func(po *v1.Pod, _ int) bool {
+	pods = lo.Reject(pods, func(po *v1.Pod, _ int) bool {
 		if err := p.Validate(ctx, po); err != nil {
 			log.FromContext(ctx).WithValues("Pod", klog.KRef(po.Namespace, po.Name)).V(1).Info(fmt.Sprintf("ignoring pod, %s", err))
 			return true
 		}
-		p.consolidationWarnings(ctx, po)
 		return false
-	}), nil
+	})
+	p.consolidationWarnings(ctx, pods)
+	return pods, nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/apis/v1/nodeclaim_validation.go
```diff
@@ -68,74 +67,75 @@ func (in *NodeClaim) SupportedVerbs() []admissionregistrationv1.OperationType {
 }
 
 // Validate the NodeClaim
-func (in *NodeClaim) Validate(_ context.Context) (errs *apis.FieldError) {
-	return errs.Also(
-		apis.ValidateObjectMetadata(in).ViaField("metadata"),
-		in.Spec.validate().ViaField("spec"),
-	)
+func (in *NodeClaim) Validate(_ context.Context) error {
```
engedaam, line 70: Do we still need this function? AFIK this was only needed to be a Knative interface

pkg/apis/v1/nodepool_validation.go
```diff
@@ -34,93 +34,96 @@ func (in *NodePool) SupportedVerbs() []admissionregistrationv1.OperationType {
 	}
 }
 
-func (in *NodePool) Validate(_ context.Context) (errs *apis.FieldError) {
-	return errs.Also(
-		apis.ValidateObjectMetadata(in).ViaField("metadata"),
-		in.Spec.validate().ViaField("spec"),
-	)
+func (in *NodePool) Validate(_ context.Context) error {
```
engedaam, line 37: Same question here

pkg/apis/v1/nodeclaim_validation.go
engedaam, line 70: Since, we don't do any runtime validation on the NodeClaim, do we still need this validations?

rschalo: Good point, do we need it and if so should it be added in this PR?

rschalo: So does that mean that all of the tests in validate are also no longer valid or useful until the CEL validation is added?

engedaam: So the CEL validation tests are currently in the package. Yeah, I don't think we need this validation or webhook tests anymore since they were only used for Webhook validation

PR #1356, opened 2024-06-26
feat: Add Conversion Webhooks for V1 APIs 

engedaam:
Fixes #N/A 

**Description**
- Add Conversion webhook APIs for NodePool and NodeClaims
- Removed Kubelet Configuration for V1 NodePool and NodeClaim
- Removed resource field for NodeClaim Template Spec on the NodePool
- Change NodeClassRef.APIVersion -> NodeClassRef.Group
- Required All the fields on the NodeClassRef
- Made NodeClaim.Spec Immutable


kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -361,6 +247,9 @@ spec:
                 - nodeClassRef
                 - requirements
               type: object
+              x-kubernetes-validations:
```
jonathan-innis, line 250: Does this automatically work for every field in the spec?

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -454,7 +343,7 @@ spec:
             - spec
           type: object
       served: true
-      storage: true
+      storage: false
```
jonathan-innis, line 346: Should we be changing the storage version here if we're only wanting to add the conversion webhook support here

kwok/charts/crds/karpenter.sh_nodepools.yaml
```diff
@@ -521,7 +393,7 @@ spec:
             - spec
           type: object
       served: true
-      storage: true
+      storage: false
```
jonathan-innis, line 396: Same question here. Do we want to change the storage version in this PR

kwok/main.go
```diff
@@ -40,5 +40,5 @@ func main() {
 			state.NewCluster(op.Clock, op.GetClient(), cloudProvider),
 			op.EventRecorder,
 			cloudProvider,
-		)...).Start(ctx)
+		)...).Start(ctx, cloudProvider)
```
jonathan-innis, line 43: Need to make this change here?

pkg/apis/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -894,6 +783,6 @@ spec:
             - spec
           type: object
       served: true
-      storage: false
+      storage: true
```
jonathan-innis, line 786: Same question

pkg/apis/v1/labels.go
```diff
@@ -47,6 +47,7 @@ const (
 	ManagedByAnnotationKey             = Group + "/managed-by"
 	NodePoolHashAnnotationKey          = Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey   = Group + "/nodepool-hash-version"
+	V1Beta1KubeletConfiguration        = Group + "/v1beta1-kubelet"
```
jonathan-innis, line 50: Should this label be similar to the compatibility annotation that we were using for a while?

pkg/apis/v1/nodeclaim.go
```diff
@@ -40,21 +40,19 @@ type NodeClaimSpec struct {
 	// +kubebuilder:validation:MaxItems:=100
 	// +required
 	Requirements []NodeSelectorRequirementWithMinValues `json:"requirements" hash:"ignore"`
-	// Resources models the resource requirements for the NodeClaim to launch
-	// +optional
-	Resources ResourceRequirements `json:"resources,omitempty" hash:"ignore"`
```
jonathan-innis, line 45: We don't want to drop resources from the NodeClaim. This is still relevant for us understanding when certain resources register. What we wanted to drop was this stanza from the public NodePool OpenAPI coinfig

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,189 @@
...
+	v1OwnerNodePool, err := in.getV1NodePoolDriftHash(ctx)
+	if err != nil {
+		return err
+	}
+	v1beta1NC.Annotations = lo.Assign(v1beta1NC.Annotations, map[string]string{
+		NodePoolHashAnnotationKey:        v1OwnerNodePool.Annotations[NodePoolHashAnnotationKey],
+		NodePoolHashVersionAnnotationKey: v1OwnerNodePool.Annotations[NodePoolHashVersionAnnotationKey],
+	})
+
+	err = in.Spec.convertTo(ctx, &v1beta1NC.Spec, in.Annotations[V1Beta1KubeletConfiguration])
```
jonathan-innis, line 51: How does the kubelet configurationt annotation work when someone changes their reference to the EC2NodeClass that they are using on the v1 API? When we convert back, aren't we going to be converting back to the old kubelet config? Do we have a good solve for this?

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,189 @@
...
+				Operator: v1Requirements.Operator,
+				Values:   v1Requirements.Values,
+			},
+			MinValues: v1Requirements.MinValues,
+		}
+	})
+
+	if in.NodeClassRef != nil {
+		nodeclass, found := lo.Find(injection.NodeClassFromContext(ctx), func(nc schema.GroupVersionKind) bool {
+			return nc.Kind == in.NodeClassRef.Kind && nc.Group == in.NodeClassRef.Group
```
jonathan-innis, line 76: Can you explain what this is doing? You are looking for the NodeClass that the user set to be equal to the current setting on their NodePool, but I'm not sure that this is the behavior that you want. Don't you want to auto-set the kind if it isn't set _as well as_ auto-set the APIVersion?

pkg/apis/v1/nodeclaim_validation.go
```diff
@@ -75,11 +72,10 @@ func (in *NodeClaim) Validate(_ context.Context) (errs *apis.FieldError) {
 	)
 }
 
-func (in *NodeClaimSpec) validate() (errs *apis.FieldError) {
+func (in *NodeClaimTemplateSpec) validate() (errs *apis.FieldError) {
```
jonathan-innis, line 75: Why did we differentiate between NodeClaimSpec and NodeClaimTemplateSpec now?

pkg/apis/v1/nodepool.go
```diff
@@ -156,7 +156,7 @@ func (l Limits) ExceededBy(resources v1.ResourceList) error {
 type NodeClaimTemplate struct {
 	ObjectMeta `json:"metadata,omitempty"`
 	// +required
-	Spec NodeClaimSpec `json:"spec"`
+	Spec NodeClaimTemplateSpec `json:"spec"`
```
jonathan-innis, line 159: Why did this change? Why not just use the same NodeClaimSpec? What's the restriction on us doing that here?

pkg/apis/v1beta1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,27 @@
...
+
+package v1beta1
+
+import (
+	"context"
+
+	"knative.dev/pkg/apis"
+)
+
+func (in *NodeClaim) ConvertTo(_ context.Context, _ apis.Convertible) error { return nil }
```
jonathan-innis, line 25: We don't need to implement this on the v1beta1 API?

pkg/operator/injection/injection.go
```diff
@@ -57,3 +63,29 @@ func WithOptionsOrDie(ctx context.Context, opts ...options.Injectable) context.C
...
+func GetClient(ctx context.Context) client.Client {
+	c := ctx.Value(clientKey)
+	if c == nil {
+		// This is a developer error if this happens, so we should panic
+		panic("client doesn't exist in context")
+	}
+	return c.(client.Client)
+}
+
+func NodeClassToContext(ctx context.Context, opts []schema.GroupVersionKind) context.Context {
```
jonathan-innis, line 80:
```suggestion
func WithDefaultNodeClass(ctx context.Context, opts []schema.GroupVersionKind) context.Context {
```

This is DefaultNodeClass, right?

jonathan-innis: Can you put a comment over this on why we need this at all

pkg/operator/injection/injection.go
```diff
@@ -57,3 +63,29 @@ func WithOptionsOrDie(ctx context.Context, opts ...options.Injectable) context.C
...
+	return context.WithValue(ctx, nodeClassKey, opts)
+}
+
+func NodeClassFromContext(ctx context.Context) []schema.GroupVersionKind {
```
jonathan-innis, line 84:
```suggestion
func GetDefaultNodeClass(ctx context.Context) []schema.GroupVersionKind {
```

pkg/operator/injection/injection.go
```diff
@@ -57,3 +63,29 @@ func WithOptionsOrDie(ctx context.Context, opts ...options.Injectable) context.C
 	}
 	return ctx
 }
+
+func WithClient(ctx context.Context, client client.Client) context.Context {
```
jonathan-innis, line 67: This feels tenuous to me since it enforces an anti-pattern IMO of injecting the client into the context. Does knative already directly support this? Are there no other ways around doing something like this?

pkg/operator/operator.go
```diff
@@ -246,6 +247,9 @@ func (o *Operator) Start(ctx context.Context) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+			// Taking the first supported NodeClass to be the default NodeClass
+			ctx = injection.NodeClassToContext(ctx, cp.GetSupportedNodeClasses())
```
jonathan-innis, line 251: We shouldn't inject all of the NodeClasses, right? We should just inject the first one

pkg/webhooks/webhooks.go
```diff
@@ -54,11 +57,31 @@ var (
 		v1beta1.SchemeGroupVersion.WithKind("NodePool"):  &v1beta1.NodePool{},
 		v1beta1.SchemeGroupVersion.WithKind("NodeClaim"): &v1beta1.NodeClaim{},
 	}
+	ConversionResource = map[schema.GroupKind]conversion.GroupKindConversion{
```
jonathan-innis, line 60: Should we note down that we want to pull this out in v1.1.0

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
engedaam: Yes, it does

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
engedaam: Moved to a separate PR. https://github.com/kubernetes-sigs/karpenter/pull/1357

pkg/apis/v1/nodeclaim.go
engedaam: The idea here is to make the NodeClaim only be the CRD that contains the resource field. This is done by creating type `NodeClaimTemplateSpec` as the struct that will hold all shared components between the `NodeClamTemplate` and `NodeClaimSpec`. In my mind, I think this would be the way to go, since the other option is having a script to remove the field. I think scripting should be the last resort, as that introduces additional complexity that will need to reason about

pkg/apis/v1/nodeclaim_conversion.go
engedaam: This for the case when user that are going from v1 -> v1beta1, containing Group to APIVersion. We will match the Group and Kind to the supported NodeClasses and use the data found there to get the version.

pkg/apis/v1beta1/nodeclaim_conversion.go
engedaam: No, since the conversion is happening from v1. We only need to implement the hub version. https://github.com/knative/pkg/blob/3f6a546ac3a42c8237ea0a8b00f5f1fd80260aa8/webhook/resourcesemantics/conversion/controller.go#L64

pkg/apis/v1/nodeclaim_validation.go
engedaam: This would be to keep all the fields that are shared between NodeClaim and NodePool

pkg/operator/injection/injection.go
engedaam: No, we want to use the NodeClass schemes as a way to find the APIVersion for going between v1 and v1beta1

pkg/operator/injection/injection.go
engedaam: I personally don't love this solution my self. Knative does not provide support for accessing the client or accessing other resource inside of the cluster. To minimize the context injection, we are only going to be containing the client on the webhooks side.

pkg/operator/operator.go
engedaam: I'm planning on using the nodeClass information to fill the gap of APIVersion for when converting between v1 -> v1beta1

pkg/apis/v1/labels.go
engedaam: Will be added in a separate PR

pkg/apis/v1/nodeclaim_conversion.go
engedaam: Will be added in a separate PR

kwok/main.go
engedaam: Yes, since the Kwok provider will need to be updated as the operator in upstream is updated, no?

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
engedaam: Removed from this PR

pkg/apis/v1/nodeclaim_validation.go
engedaam: Removed from this PR

pkg/apis/v1/nodepool.go
engedaam: Removed from this PR

pkg/webhooks/webhooks.go
```diff
@@ -54,11 +57,32 @@ var (
 		{Group: "karpenter.sh", Version: "v1beta1", Kind: "NodePool"}:  &v1beta1.NodePool{},
 		{Group: "karpenter.sh", Version: "v1beta1", Kind: "NodeClaim"}: &v1beta1.NodeClaim{},
 	}
+	// Remove conversion webhooks once v1.1.0, and v1beta1 APIs are droped
+	ConversionResource = map[schema.GroupKind]conversion.GroupKindConversion{
+		{Group: "karpenter.sh", Kind: "NodePool"}: {
```
jonathan-innis, line 62: Can we use operatorpkg's `GVK` function to get the GroupKind here?

pkg/webhooks/webhooks.go
```diff
@@ -54,11 +57,32 @@ var (
 		{Group: "karpenter.sh", Version: "v1beta1", Kind: "NodePool"}:  &v1beta1.NodePool{},
 		{Group: "karpenter.sh", Version: "v1beta1", Kind: "NodeClaim"}: &v1beta1.NodeClaim{},
 	}
+	// Remove conversion webhooks once v1.1.0, and v1beta1 APIs are droped
```
jonathan-innis, line 60:
```suggestion
	// Remove conversion webhooks once v1.1.0, and v1beta1 APIs are dropped
```

pkg/webhooks/webhooks.go
```diff
@@ -54,11 +57,32 @@ var (
 		{Group: "karpenter.sh", Version: "v1beta1", Kind: "NodePool"}:  &v1beta1.NodePool{},
```
jonathan-innis, line 57: I wonder if we can use operatorpkg's `GVK` function here as well

pkg/operator/injection/injection.go
jonathan-innis: Do we need this injection anymore now that we don't have cross-resource access in our conversion webhooks?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+	"knative.dev/pkg/apis"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+)
+
+// Convert v1 NodePool to v1beta1 NodePool
+func (in *NodePool) ConvertTo(ctx context.Context, to apis.Convertible) error {
+	v1beta1NP := to.(*v1beta1.NodePool)
+	v1beta1NP.Name = in.Name
```
jonathan-innis, line 35: Do you need to handle these fields individually or can you just copy over all of the meta? For instance, I notice that this causes you to lose the ownerReferences in the translation

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (in *NodePoolSpec) convertTo(ctx context.Context, v1beta1np *v1beta1.NodePoolSpec) {
+	v1beta1np.Weight = in.Weight
+	v1beta1np.Limits = v1beta1.Limits(in.Limits)
+	in.Disruption.convertTo(&v1beta1np.Disruption)
+	in.Template.convertTo(ctx, &v1beta1np.Template)
+}
+
+func (in *Disruption) convertTo(v1beta1np *v1beta1.Disruption) {
```
jonathan-innis, line 59: What happens here if the disruption block is `nil`?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+			NodeSelectorRequirement: v1.NodeSelectorRequirement{
+				Key:      v1Requirements.Key,
+				Values:   v1Requirements.Values,
+				Operator: v1Requirements.Operator,
+			},
+			MinValues: v1Requirements.MinValues,
+		}
+	})
+
+	if in.Spec.NodeClassRef != nil {
```
jonathan-innis, line 88: Is there ever a case where it's valid for the NodeClassRef to be nil?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+
+	// Need to implement Kubelet Conversion
+}
+
+// Convert v1beta1 NodePool to V1 NodePool
+func (in *NodePool) ConvertFrom(ctx context.Context, v1beta1np apis.Convertible) error {
+	v1beta1NP := v1beta1np.(*v1beta1.NodePool)
+	in.Name = v1beta1NP.Name
```
jonathan-innis, line 105: Same comment here. How do we make sure that we don't lose our ObejctMeta that isn't included here?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+
+func (in *NodePoolSpec) convertFrom(ctx context.Context, v1beta1np *v1beta1.NodePoolSpec) {
+	in.Weight = v1beta1np.Weight
+	in.Limits = Limits(v1beta1np.Limits)
+	in.Disruption.convertFrom(&v1beta1np.Disruption)
+	in.Template.convertFrom(ctx, &v1beta1np.Template)
+}
+
+func (in *Disruption) convertFrom(v1beta1np *v1beta1.Disruption) {
+	in.ConsolidateAfter = (*NillableDuration)(v1beta1np.ConsolidateAfter)
```
jonathan-innis, line 129: Same question: what happens if the disruption block is nil?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+				Key:      v1beta1Requirements.Key,
+				Values:   v1beta1Requirements.Values,
+				Operator: v1beta1Requirements.Operator,
+			},
+			MinValues: v1beta1Requirements.MinValues,
+		}
+	})
+
+	nodeclasses := injection.GetNodeClasses(ctx)
+	if v1beta1np.Spec.NodeClassRef != nil {
```
jonathan-innis, line 158: same question: Is there a case where it's valid for this to be `nil`?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,167 @@
...
+		in.Spec.NodeClassRef = &NodeClassReference{
+			Name:  v1beta1np.Spec.NodeClassRef.Name,
+			Kind:  lo.Ternary(v1beta1np.Spec.NodeClassRef.Kind == "", nodeclasses[0].Kind, v1beta1np.Spec.NodeClassRef.Kind),
+			Group: lo.Ternary(v1beta1np.Spec.NodeClassRef.APIVersion == "", nodeclasses[0].Group, strings.Split(v1beta1np.Spec.NodeClassRef.APIVersion, "/")[0]),
```
jonathan-innis, line 162: I think there's an idiomatic function to parse the APIVersion in `apimachinery`. I think it basically boils down to the same thing that you have here, but just a callout if you want to use that instead

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,172 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"knative.dev/pkg/apis"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+)
+
+func (in *NodeClaim) ConvertTo(ctx context.Context, to apis.Convertible) error {
+	v1beta1NC := to.(*v1beta1.NodeClaim)
+	v1beta1NC.Name = in.Name
```
jonathan-innis, line 35: A lot of the same questions I had on the NodePool conversion apply here. I'd check those and apply the same thinking to this resource

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+							Name:  "test",
+							Kind:  "test-kind",
+							Group: "test-group",
+						},
+						Requirements: []NodeSelectorRequirementWithMinValues{},
+					},
+				},
+			},
+		}
+		v1NodeClaim = &NodeClaim{}
```
jonathan-innis, line 61: nit: It's not standard convention to use capitalization like this in golang. Should we lowercase our variables throughout this file?

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+				Group:   "fake-cloudprovider-group",
+				Version: "fake-cloudprovider-version",
+				Kind:    "fake-cloudprovider-kind",
+			},
+		}
+		ctx = injection.WithNodeClasses(ctx, cloudProvider.GetSupportedNodeClasses())
+		ctx = injection.WithClient(ctx, env.Client)
+	})
+
+	Context("MetaData", func() {
```
jonathan-innis, line 81:
```suggestion
	Context("Metadata", func() {
```

nit: It's one word, not two

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+							Kind:  "test-kind",
+							Group: "test-group",
+						},
+						Requirements: []NodeSelectorRequirementWithMinValues{},
+					},
+				},
+			},
+		}
+		v1NodeClaim = &NodeClaim{}
+		v1NodeClaim.Labels = map[string]string{
```
jonathan-innis, line 62: nit: Why not define these inline with the original object definition?

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+					Effect: v1.TaintEffectNoExecute,
+				},
+				{
+					Key:    "test-key-startup-2",
+					Value:  "test-value-startup-2",
+					Effect: v1.TaintEffectNoSchedule,
+				},
+			}
+			Expect(len(v1beta1NodeClaim.Spec.StartupTaints)).To(BeNumerically("==", 0))
+			Expect(v1NodeClaim.ConvertTo(ctx, v1beta1NodeClaim)).To(BeNil())
```
jonathan-innis, line 146:
```suggestion
			Expect(v1NodeClaim.ConvertTo(ctx, v1beta1NodeClaim)).To(Succeed())
```

I wonder if we should use this throughout since it's a bit more idiomatic to write it this way

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+			}
+		})
+		Context("NodeClassRef", func() {
+			It("should convert v1 nodeclaim template nodeClassRef", func() {
+				v1NodeClaim.Spec.NodeClassRef = &NodeClassReference{
+					Kind:  "fake-cloudprovider-kind",
+					Name:  "nodeclass-test",
+					Group: "fake-cloudprovider-group",
+				}
+				Expect(v1beta1NodeClaim.Spec.NodeClassRef).To(BeNil())
```
jonathan-innis, line 199: Why is this here? Can we just assume that it's not set to start?

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+				Group:   "fake-cloudprovider-group",
+				Version: "fake-cloudprovider-version",
+				Kind:    "fake-cloudprovider-kind",
+			},
+		}
+		ctx = injection.WithNodeClasses(ctx, cloudProvider.GetSupportedNodeClasses())
+		ctx = injection.WithClient(ctx, env.Client)
+	})
+
+	Context("MetaData", func() {
```
jonathan-innis, line 314:
```suggestion
	Context("Metadata", func() {
```

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,539 @@
...
+		It("should convert v1beta1 nodeclaim Name", func() {
```
jonathan-innis, line 315: I wonder if you check all ObjectMeta at once. There's a lot of components of the ObjectMeta and checking that each one was copied one by one isn't probably worth the time as long as we're just doing a straight ObjectMeta mapping

pkg/operator/operator.go
```diff
@@ -247,6 +248,9 @@ func (o *Operator) Start(ctx context.Context) {
 		wg.Add(1)
 		go func() {
 			defer wg.Done()
+			// Taking the first supported NodeClass to be the default NodeClass
+			ctx = injection.WithNodeClasses(ctx, cp.GetSupportedNodeClasses())
+			ctx = injection.WithClient(ctx, o.GetClient())
```
jonathan-innis, line 253: Do we need GetClient() anymore?

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -0,0 +1,532 @@
+/*
```
jonathan-innis, line 1: I'm not planning to look at this file since I expect it to be fairly similar to the NodeClaim conversion test. I'd mainly just apply the same comments that I added over there into this test file

pkg/apis/v1/nodepool_conversion.go
engedaam: How can you have a `nil` disruption block? We have default values set, no?

pkg/apis/v1/nodepool_conversion.go
engedaam: I didn't realize that NodePool had owner reference. What would own the NodePools? I can update to use meta

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,172 @@
...
+			MinValues: v1beta1Requirements.MinValues,
+		}
+	})
+
+	nodeclasses := injection.GetNodeClasses(ctx)
+	if v1beta1nc.NodeClassRef != nil {
+		in.NodeClassRef = &NodeClassReference{
+			Name:  v1beta1nc.NodeClassRef.Name,
+			Kind:  lo.Ternary(v1beta1nc.NodeClassRef.Kind == "", nodeclasses[0].Kind, v1beta1nc.NodeClassRef.Kind),
+			Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", nodeclasses[0].Group, strings.Split(v1beta1nc.NodeClassRef.APIVersion, "/")[0]),
```
njtran, line 136: Don't we want to change it to the new api version? If i'm misunderstanding, can you show me how the nodeClassRef.APIVersion is set?

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,172 @@
...
+	in.Allocatable = v1beta1nc.Allocatable
+	in.Conditions = v1beta1nc.Conditions
+}
+
+func (in *NodeClaim) getV1NodePoolDriftHash(ctx context.Context) (*NodePool, error) {
+	client := injection.GetClient(ctx)
+
+	nodepool := &NodePool{}
+	err := client.Get(ctx, types.NamespacedName{Name: in.Labels[NodePoolLabelKey]}, nodepool)
+	if err != nil {
```
njtran, line 157: nit: any reason not to inline?

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return nodepool, nil
+}
+
+func getV1Beta1NodePoolDriftHash(ctx context.Context, v1beta1nc *v1beta1.NodeClaim) (*v1beta1.NodePool, error) {
+	client := injection.GetClient(ctx)
+
+	nodepool := &v1beta1.NodePool{}
+	err := client.Get(ctx, types.NamespacedName{Name: v1beta1nc.Labels[NodePoolLabelKey]}, nodepool)
+	if err != nil {
```
njtran, line 168: nit: same on inline

pkg/apis/v1/nodepool_conversion.go
njtran: Nothing owns a nodepool

pkg/operator/injection/injection.go
engedaam: I was initially thinking we might have drift issue and needed to pull v1 annotations, however after I tested the hash controller can handle the changes that users make to cause drift

pkg/apis/v1/nodepool_conversion.go
njtran: Is it fine to just set in.Meta to v1beta1NP.Meta?

pkg/apis/v1beta1/nodeclaim_conversion.go
njtran, line 1: worth adding as a comment why we don't need conversion webhooks here, and only in v1

pkg/operator/injection/injection.go
njtran: +1

pkg/operator/operator.go
engedaam: This was initially in to pull the nodepool hash that would need to be calculated for each version. Tested with the new approach if it would be needed, I will remove it.

pkg/apis/v1/nodeclaim_conversion_test.go
engedaam: We can assume this. I was trying explicitly call that out

pkg/apis/v1/nodepool_conversion.go
engedaam: I ran into this when I was testing. But after running a few test, I saw it might have been due to a misconfiguration

pkg/apis/v1/nodepool_conversion.go
engedaam: Yes, it is. I was trying to be careful with how we handled drift, so I has just moving one component at a time. With our new approach, I will be just copying over the fields

pkg/apis/v1/nodeclaim_conversion.go
engedaam: APIVersion is set above in the `ConvertTo` https://github.com/kubernetes-sigs/karpenter/pull/1356/files#diff-b8d4e10f084dfecb98330acb7dc0b082455afea1837c74bd59ab11d29b8518adR96

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (in *NodeClaim) ConvertFrom(ctx context.Context, from apis.Convertible) error {
+	v1beta1NC := from.(*v1beta1.NodeClaim)
+	in.Name = v1beta1NC.Name
+	in.UID = v1beta1NC.UID
+	in.Labels = v1beta1NC.Labels
+	in.Annotations = v1beta1NC.Annotations
+
+	in.Spec.convertFrom(ctx, &v1beta1NC.Spec)
```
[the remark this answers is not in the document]
engedaam: Removed the hash. We will not need ordering anymore

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -0,0 +1,536 @@
...
+			v1nodepool.Spec.Limits = Limits{
+				v1.ResourceCPU:    resource.MustParse("5"),
+				v1.ResourceMemory: resource.MustParse("14145G"),
+			}
+			Expect(v1nodepool.ConvertTo(ctx, v1beta1nodepool)).To(Succeed())
+			for _, resource := range lo.Keys(v1nodepool.Spec.Limits) {
+				Expect(v1beta1nodepool.Spec.Limits[resource]).To(Equal(v1nodepool.Spec.Limits[resource]))
+			}
+		})
+		Context("NodeClaimTemplate", func() {
```
njtran, line 99: Any reason we can't just compare the meta here too?

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -0,0 +1,536 @@
...
+				for i := range v1nodepool.Spec.Template.Spec.Taints {
+					Expect(v1beta1nodepool.Spec.Template.Spec.Taints[i].Key).To(Equal(v1nodepool.Spec.Template.Spec.Taints[i].Key))
+					Expect(v1beta1nodepool.Spec.Template.Spec.Taints[i].Value).To(Equal(v1nodepool.Spec.Template.Spec.Taints[i].Value))
+					Expect(v1beta1nodepool.Spec.Template.Spec.Taints[i].Effect).To(Equal(v1nodepool.Spec.Template.Spec.Taints[i].Effect))
+				}
+			})
+			It("should convert v1 nodepool template startup taints", func() {
+				v1nodepool.Spec.Template.Spec.StartupTaints = []v1.Taint{
+					{
+						Key:    "test-key--startup-1",
```
njtran, line 143: nit doesn't really matter though
```suggestion
						Key:    "test-key-startup-1",
```

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -0,0 +1,536 @@
...
+						NodeSelectorRequirement: v1.NodeSelectorRequirement{
+							Key:      CapacityTypeLabelKey,
+							Operator: v1.NodeSelectorOpIn,
+							Values:   []string{CapacityTypeSpot},
+						},
+						MinValues: lo.ToPtr(0),
+					},
+				}
+				Expect(v1nodepool.ConvertFrom(ctx, v1beta1nodepool)).To(Succeed())
+				for i := range v1beta1nodepool.Spec.Template.Spec.Requirements {
```
njtran, line 414: add minvalues here

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,168 @@
...
+	in.Limits = Limits(v1beta1np.Limits)
+	in.Disruption.convertFrom(&v1beta1np.Disruption)
+	return in.Template.convertFrom(ctx, &v1beta1np.Template)
+}
+
+func (in *Disruption) convertFrom(v1beta1np *v1beta1.Disruption) {
+	in.ConsolidateAfter = (*NillableDuration)(v1beta1np.ConsolidateAfter)
+	in.ConsolidationPolicy = ConsolidationPolicy(v1beta1np.ConsolidationPolicy)
+	in.ExpireAfter = NillableDuration(v1beta1np.ExpireAfter)
+	in.Budgets = lo.Map(v1beta1np.Budgets, func(v1beta1Budget v1beta1.Budget, _ int) Budget {
```
njtran, line 124: What about reasons?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,168 @@
...
+				Key:      v1Requirements.Key,
+				Values:   v1Requirements.Values,
+				Operator: v1Requirements.Operator,
+			},
+			MinValues: v1Requirements.MinValues,
+		}
+	})
+
+	nodeClasses := injection.GetNodeClasses(ctx)
+	sort.Slice(nodeClasses, func(i int, j int) bool {
```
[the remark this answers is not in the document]
engedaam: That is only when converting from v1beta1 -> v1. Since users might not have the defined

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -0,0 +1,168 @@
...
+	in.Budgets = lo.Map(v1beta1np.Budgets, func(v1beta1Budget v1beta1.Budget, _ int) Budget {
+		return Budget{
+			Nodes:    v1beta1Budget.Nodes,
+			Schedule: v1beta1Budget.Schedule,
+			Duration: v1beta1Budget.Duration,
+		}
+	})
+}
+
+func (in *NodeClaimTemplate) convertFrom(ctx context.Context, v1beta1np *v1beta1.NodeClaimTemplate) error {
```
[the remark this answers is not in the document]
engedaam: We can, I did label and annotation, since that the only values we care about

pkg/apis/v1/nodeclaim_conversion_test.go
```diff
@@ -0,0 +1,513 @@
...
+				Kind:    "fake-cloudprovider-kind",
+			},
+		}
+		ctx = injection.WithNodeClasses(ctx, cloudProvider.GetSupportedNodeClasses())
+	})
+
+	It("should convert v1 nodeclaim metadata", func() {
+		v1nodeclaim.ObjectMeta = test.ObjectMeta()
+		Expect(v1nodeclaim.ConvertFrom(ctx, v1beta1nodeclaim)).To(Succeed())
+		Expect(v1beta1nodeclaim.ObjectMeta).To(BeEquivalentTo(v1nodeclaim.ObjectMeta))
```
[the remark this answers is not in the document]
engedaam: Yes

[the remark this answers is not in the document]
engedaam: https://pkg.go.dev/github.com/onsi/gomega@v1.33.1#BeEquivalentTo

pkg/apis/v1/nodepool_conversion.go
engedaam: Good callout. Added that in

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -0,0 +1,549 @@
...
+						Duration: &metav1.Duration{Duration: time.Second * 2121},
+					})
+					Expect(v1nodepool.ConvertTo(ctx, v1beta1nodepool)).To(Succeed())
+					for i := range v1nodepool.Spec.Disruption.Budgets {
+						Expect(v1beta1nodepool.Spec.Disruption.Budgets[i].Duration.Duration).To(Equal(v1nodepool.Spec.Disruption.Budgets[i].Duration.Duration))
+					}
+				})
+				It("should convert v1 nodepool reason", func() {
+					v1nodepool.Spec.Disruption.Budgets = append(v1nodepool.Spec.Disruption.Budgets, Budget{
+						Reasons: []DisruptionReason{DisruptionReasonDrifted, DisruptionReasonUnderutilized},
```
njtran, line 251: Can you add in emptiness for coverage too? Think you missed that one.

njtran (commented):
/lgtm
/approve

njtran (commented):
/lgtm
/approve

PR #1359, opened 2024-06-26
chore: Make NodeClaim Spec Immutable on v1 API

engedaam:
Fixes #N/A 

**Description**
- Make NodeClaim Immutable on v1 API

- `make presubmit`

### DIFF
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -361,6 +361,9 @@ spec:
                 - nodeClassRef
                 - requirements
               type: object
+              x-kubernetes-validations:
+                - message: immutable field changed
+                  rule: self == oldSelf
             status:
               description: NodeClaimStatus defines the observed state of NodeClaim
               properties:
--- a/pkg/apis/v1/nodeclaim.go
+++ b/pkg/apis/v1/nodeclaim.go
@@ -176,6 +176,7 @@ type NodeClaim struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="immutable field changed"
 	// +required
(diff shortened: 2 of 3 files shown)


kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -361,6 +361,9 @@ spec:
                 - nodeClassRef
                 - requirements
               type: object
+              x-kubernetes-validations:
+                - message: immutable field changed
+                  rule: self == oldSelf
```
njtran, line 366: Are we saying this doesn't apply to metadata? I'm seeing the title of the PR say spec, so I'm wondering what the decision is. I'm thinking it has to be mutable, since we have to add these labels in ourselves at nodeclaim launch

pkg/apis/v1/nodeclaim.go
```diff
@@ -175,6 +175,7 @@ type NodeClaim struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="immutable field changed"
```
njtran, line 178:
```suggestion
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="spec is immutable"
```

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
engedaam: Yes, we currently want to make sure the metadata is mutable. The goal here is to only make node claim spec immutable. v1 RFC: https://github.com/kubernetes-sigs/karpenter/pull/1222/files#diff-5721182d9716c35869521e76985d0593716478602bf2a8e4862303ef8f3d5f06R310

engedaam: Since, the Karpenter controller makes updates to the annotation and labels after node claim creation we need to make sure the metadata object is mutable.

jonathan-innis (approved):
/lgtm
/approve

PR #1360, opened 2024-06-26
chore: Update `nodeClassRef` to be Group instead of APIVersion

### LINKED ISSUE #909: How to handle unset `apiVersion` and `kind` in `nodeClassRef`

engedaam:
Fixes #909 

**Description**
- Change apiVersion -> Group 
- Require Kind and Group for NodeClassRef

- `make presubmit`

### DIFF
--- a/pkg/apis/v1/nodeclaim.go
+++ b/pkg/apis/v1/nodeclaim.go
@@ -146,14 +146,14 @@ type KubeletConfiguration struct {
 
 type NodeClassReference struct {
 	// Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"
-	// +optional
-	Kind string `json:"kind,omitempty"`
+	// +required
+	Kind string `json:"kind"`
 	// Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names
 	// +required
 	Name string `json:"name"`
 	// API version of the referent
-	// +optional
-	APIVersion string `json:"apiVersion,omitempty"`
+	// +required
+	Group string `json:"group"`
 }
 
 // +kubebuilder:object:generate=false
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -186,7 +186,7 @@ spec:
                 nodeClassRef:
                   description: NodeClassRef is a reference to an object that defines provider specific configuration
                   properties:
-                    apiVersion:
+                    group:
                       description: API version of the referent
                       type: string
                     kind:
@@ -196,6 +196,8 @@ spec:
                       description: 'Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names'
                       type: string
                   required:
+                    - group
+                    - kind
                     - name
                   type: object
                 requirements:
--- a/kwok/charts/crds/karpenter.sh_nodepools.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodepools.yaml
@@ -310,7 +310,7 @@ spec:
                         nodeClassRef:
                           description: NodeClassRef is a reference to an object that defines provider specific configuration
                           properties:
-                            apiVersion:
+                            group:
                               description: API version of the referent
                               type: string
                             kind:
@@ -320,6 +320,8 @@ spec:
                               description: 'Name of the referent; More info: http://kubernetes.io/docs/user-guide/identifiers#names'
                               type: string
                           required:
+                            - group
+                            - kind
                             - name
                           type: object
                         requirements:
--- a/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
@@ -186,7 +186,7 @@ spec:
                 nodeClassRef:
                   description: NodeClassRef is a reference to an object that defines provider specific configuration
                   properties:
-                    apiVersion:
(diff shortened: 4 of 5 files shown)


njtran (commented):
Looks good to me. Do we need to change anything for the kwoknodeclass?

engedaam:
No, this will only impact v1 NodePools. KOWK currenlty uses v1beta1. We will need to update once switch to use v1

pkg/apis/v1/nodeclaim.go
```diff
@@ -146,14 +146,14 @@ type KubeletConfiguration struct {
 
 type NodeClassReference struct {
 	// Kind of the referent; More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#types-kinds"
-	// +optional
-	Kind string `json:"kind,omitempty"`
+	// +required
```
jonathan-innis, line 149: Should we also require that these strings have a minimum length so that they also can't be an empty string?

engedaam: Can we leave that to CodeClass readiness on the NodePool to validate if the nodeClassRef is valid?

jonathan-innis: Sure, I'm more just thinking that required fields shouldn't be empty anyways so it may just be better to validate that they aren't empty earlier

jonathan-innis (approved):
/lgtm
/approve

PR #1361, opened 2024-06-26
chore: Remove Resource field from the NodePool API

engedaam:
Fixes #N/A 

**Description**
- Remove resources field from the NodePool v1 API
- The idea here is to make the NodeClaim only be the CRD that contains the resource field. This is done by creating type NodeClaimTemplateSpec as the struct that will hold all shared components between the NodeClamTemplate and NodeClaimSpec. In my mind, I think this would be the way to go, since the other option is having a script to remove the field. I think scripting should be the last resort, as that introduces additional complexity that will need to reason about

- `make presubmit`


hack/validation/resources.sh
```diff
@@ -1,7 +1,4 @@
...
-# Version = 1 // v1beta1 API
-for Version in  $(seq 0 1); do
-    # Adding validation for nodepool.spec.template.spec.resources 
-    yqVersion="$Version" yq eval ".spec.versions[env(yqVersion)].schema.openAPIV3Schema.properties.spec.properties.template.properties.spec.properties.resources.maxProperties = 0" -i pkg/apis/crds/karpenter.sh_nodepools.yaml 
-done
\ No newline at end of file
+# Adding validation v1beta1 NodePool API
+
+# Adding validation for nodepool.spec.template.spec.resources 
+yqVersion="$Version" yq eval ".spec.versions[1].schema.openAPIV3Schema.properties.spec.properties.template.properties.spec.properties.resources.maxProperties = 0" -i pkg/apis/crds/karpenter.sh_nodepools.yaml
```
njtran, line 4: I don't see the yqVersion envVar referenced in this line. Am I missing something?

pkg/apis/v1/nodeclaim.go
```diff
@@ -55,6 +52,14 @@ type NodeClaimSpec struct {
 	NodeClassRef *NodeClassReference `json:"nodeClassRef"`
 }
 
+// NodeClaimSpec describes the desired state of the NodeClaim
+type NodeClaimSpec struct {
+	NodeClaimTemplateSpec `json:",inline"`
```
njtran, line 57: This naming feels odd to me. What if we named this NodeClaimTemplate?
```suggestion
// NodeClaimSpec describes the desired state of the NodeClaim
type NodeClaimTemplate struct {
	NodeClaimTemplateSpec `json:",inline"`
```

kwok/charts/crds/karpenter.sh_nodepools.yaml
```diff
@@ -191,7 +191,7 @@ spec:
                               rule: self.all(x, x != "kubernetes.io/hostname")
                       type: object
                     spec:
-                      description: NodeClaimSpec describes the desired state of the NodeClaim
+                      description: NodeClaimTemplateSpec describes the desired shared fields between NodeClaimTemplate and NodeClaimSpec
```
njtran, line 194: nit: 
```suggestion
                      description: NodeClaimTemplateSpec describes the shared fields between NodeClaimTemplate and NodeClaimSpec
```

hack/validation/resources.sh
engedaam: This was included to indicate the version. Removed it now since we only need to reason about v1beta1

pkg/apis/v1/nodeclaim.go
engedaam: Update the name of NodeClaimTemplateSpec to be NodeClaimCoreSpec

pkg/apis/v1/nodeclaim.go
```diff
@@ -22,8 +22,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-// NodeClaimSpec describes the desired state of the NodeClaim
-type NodeClaimSpec struct {
+// NodeClaimCoreSpec describes the shared fields between NodeClaimTemplate and NodeClaimSpec
```
jonathan-innis, line 25: Can we just keep the same go struct types and then dynamically remove the resources field from the openAPI spec?

engedaam: Okay, I updated the PR to reflect that

jonathan-innis (approved):
/lgtm
/approve

PR #1370, opened 2024-06-28
feat: expose nodeclaim disruption through new disruption condition, improves pod eviction event message

cnmcavoy:
Fixes #N/A 

**Description**

Add's a new nodeclaim condition `DisruptionCandidate` which is set when a nodeclaim is being disrupted, and is applied after the disruption taint is set. The `DisruptionCandidate` nodeclaim condition contains the reason why the nodeclaim is being terminated (e.g `node worker-mgn6n/ip-10-115-200-242.us-east-2.compute.internal  was single node consolidated`).

The motivation for this new nodeclaim condition is so that when evicting pods, we can look up this condition and use the condition's message in the pod event.

Example of what the pod events look like now from testing in our clusters:
```
keda                                          85s         Normal    Evicted                           pod/keda-admission-webhooks-5bd6b554ff-tn25h                                                Evicted pod: node worker-qa-czn28/ip-10-115-195-50.us-east-2.compute.internal drifted
keda                                          86s         Normal    Evicted                           pod/keda-operator-688bc9b887-7t4gd                                                          Evicted pod: node worker-qa-czn28/ip-10-115-195-50.us-east-2.compute.internal drifted
```

Built Karpenter with this change locally and tested in our clusters. Also `make presubmit`

### DIFF
--- a/pkg/apis/v1/nodeclaim_status.go
+++ b/pkg/apis/v1/nodeclaim_status.go
@@ -30,6 +30,7 @@ const (
 	ConditionTypeDrifted              = "Drifted"
 	ConditionTypeInstanceTerminating  = "InstanceTerminating"
 	ConditionTypeConsistentStateFound = "ConsistentStateFound"
+	ConditionTypeDisruptionReason     = "DisruptionReason"
 )
 
 // NodeClaimStatus defines the observed state of NodeClaim
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -27,6 +27,7 @@ import (
 
 	"github.com/awslabs/operatorpkg/singleton"
 	"github.com/samber/lo"
+	"go.uber.org/multierr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/uuid"
 	"k8s.io/utils/clock"
@@ -35,9 +36,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-
-	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
-	"sigs.k8s.io/karpenter/pkg/utils/pretty"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
@@ -49,6 +47,8 @@ import (
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
 	operatorlogging "sigs.k8s.io/karpenter/pkg/operator/logging"
+	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
+	"sigs.k8s.io/karpenter/pkg/utils/pretty"
 )
 
 type Controller struct {
@@ -124,13 +124,20 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	// Karpenter taints nodes with a karpenter.sh/disruption taint as part of the disruption process while it progresses in memory.
 	// If Karpenter restarts or fails with an error during a disruption action, some nodes can be left tainted.
 	// Idempotently remove this taint from candidates that are not in the orchestration queue before continuing.
-	if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, false, lo.Filter(c.cluster.Nodes(), func(s *state.StateNode, _ int) bool {
-		return !c.queue.HasAny(s.ProviderID())
-	})...); err != nil {
+	outdatedNodes := lo.Filter(c.cluster.Nodes(), func(s *state.StateNode, _ int) bool {
+		return !c.queue.HasAny(s.ProviderID()) && !s.Deleted()
+	})
+	if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, false, outdatedNodes...); err != nil {
 		if errors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
 		}
 		return reconcile.Result{}, fmt.Errorf("removing taint %s from nodes, %w", pretty.Taint(v1.DisruptedNoScheduleTaint), err)
+	}
+	if err := state.ClearNodeClaimsCondition(ctx, c.kubeClient, v1.ConditionTypeDisruptionReason, outdatedNodes...); err != nil {
+		if errors.IsConflict(err) {
+			return reconcile.Result{Requeue: true}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("removing %s condition from nodeclaims, %w", v1.ConditionTypeDisruptionReason, err)
 	}
 
 	// Attempt different disruption methods. We'll only let one method perform an action
@@ -197,12 +204,9 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command, 
 	commandID := uuid.NewUUID()
 	log.FromContext(ctx).WithValues("command-id", commandID, "reason", strings.ToLower(string(m.Reason()))).Info(fmt.Sprintf("disrupting nodeclaim(s) via %s", cmd))
 
-	stateNodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode {
-		return c.StateNode
-	})
 	// Cordon the old nodes before we launch the replacements to prevent new pods from scheduling to the old nodes
-	if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, true, stateNodes...); err != nil {
-		return fmt.Errorf("tainting nodes with %s (command-id: %s), %w", pretty.Taint(v1.DisruptedNoScheduleTaint), commandID, err)
+	if err := c.MarkDisrupted(ctx, m, cmd.candidates...); err != nil {
+		return fmt.Errorf("marking disrupted (command-id: %s), %w", commandID, err)
 	}
 
 	var nodeClaimNames []string
@@ -226,12 +230,9 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command, 
 	// the node is cleaned up.
 	schedulingResults.Record(log.IntoContext(ctx, operatorlogging.NopLogger), c.recorder, c.cluster)
 
-	providerIDs := lo.Map(cmd.candidates, func(c *Candidate, _ int) string { return c.ProviderID() })
-	// We have the new NodeClaims created at the API server so mark the old NodeClaims for deletion
-	c.cluster.MarkForDeletion(providerIDs...)
-
-	if err = c.queue.Add(orchestration.NewCommand(nodeClaimNames,
-		lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode }), commandID, m.Reason(), m.ConsolidationType())); err != nil {
+	statenodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode })
+	if err := c.queue.Add(orchestration.NewCommand(nodeClaimNames, statenodes, commandID, m.Reason(), m.ConsolidationType())); err != nil {
+		providerIDs := lo.Map(cmd.candidates, func(c *Candidate, _ int) string { return c.ProviderID() })
 		c.cluster.UnmarkForDeletion(providerIDs...)
 		return fmt.Errorf("adding command to queue (command-id: %s), %w", commandID, err)
 	}
@@ -256,6 +257,30 @@ func (c *Controller) createReplacementNodeClaims(ctx context.Context, m Method, 
 		return nil, fmt.Errorf("expected %d replacements, got %d", len(cmd.replacements), len(nodeClaimNames))
 	}
 	return nodeClaimNames, nil
+}
+
+func (c *Controller) MarkDisrupted(ctx context.Context, m Method, candidates ...*Candidate) error {
+	stateNodes := lo.Map(candidates, func(c *Candidate, _ int) *state.StateNode {
+		return c.StateNode
+	})
+	if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, true, stateNodes...); err != nil {
+		return fmt.Errorf("tainting nodes with %s: %w", pretty.Taint(v1.DisruptedNoScheduleTaint), err)
+	}
+
+	providerIDs := lo.Map(candidates, func(c *Candidate, _ int) string { return c.ProviderID() })
+	c.cluster.MarkForDeletion(providerIDs...)
+
+	return multierr.Combine(lo.Map(candidates, func(candidate *Candidate, _ int) error {
+		// refresh nodeclaim before updating status
+		nodeClaim := &v1.NodeClaim{}
+
+		if err := c.kubeClient.Get(ctx, client.ObjectKeyFromObject(candidate.NodeClaim), nodeClaim); err != nil {
+			return client.IgnoreNotFound(err)
+		}
+		stored := nodeClaim.DeepCopy()
+		nodeClaim.StatusConditions().SetTrueWithReason(v1.ConditionTypeDisruptionReason, v1.ConditionTypeDisruptionReason, string(m.Reason()))
+		return client.IgnoreNotFound(c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFrom(stored)))
+	})...)
 }
 
 func (c *Controller) recordRun(s string) {
--- a/pkg/controllers/node/termination/terminator/events/events.go
+++ b/pkg/controllers/node/termination/terminator/events/events.go
@@ -26,12 +26,12 @@ import (
 	"sigs.k8s.io/karpenter/pkg/events"
 )
 
-func EvictPod(pod *corev1.Pod) events.Event {
+func EvictPod(pod *corev1.Pod, message string) events.Event {
 	return events.Event{
 		InvolvedObject: pod,
 		Type:           corev1.EventTypeNormal,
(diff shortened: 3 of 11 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -251,6 +263,21 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
 	return nil
 }
 
+func evictionReason(m Method, nodeClaim *v1beta1.NodeClaim) string {
```
Bryce-Soghigian, line 266: Is there a reason we prefer passing in a method vs just including this in the methods' pointer receivers

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -68,13 +69,15 @@ func IsNodeDrainError(err error) bool {
 
 type QueueKey struct {
 	types.NamespacedName
-	UID types.UID
+	UID      types.UID
+	nodeName string
```
njtran, line 73: I think we should probably do providerID instead of node name. We already have the indexes on providerID for both nodeclaim and node, so it shouldn't be that much bigger of a deal.

pkg/controllers/node/termination/terminator/events/events.go
njtran, line 1: We probably want something else in the nodeclaim termination controller that detects if the DisruptionReason() status condition hasn't been set and sets it to "Forcefully terminated" to represent anything like Expiration, kubectl delete, or nodepool cascade deletion

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -164,6 +167,10 @@ func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
 // Evict returns true if successful eviction call, and false if there was an eviction-related error
 func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(key.Namespace, key.Name)))
+	evictionMessage, err := evictionReason(ctx, key, q.kubeClient)
+	if err != nil {
+		log.FromContext(ctx).Error(err, "failed looking up pod eviction reason")
```
njtran, line 172: This would really just fail if the nodeclaim doesn't exist, right?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -192,10 +199,35 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
+	if err != nil {
+		return "", err
+	}
+	for _, nodeClaim := range nodeClaims {
+		terminationCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeDisruptionCandidate)
+		if terminationCondition.IsTrue() {
+			return terminationCondition.Message, nil
+		}
+		if !nodeClaim.DeletionTimestamp.IsZero() {
+			return fmt.Sprintf("node %s/%s is marked for deletion", nodeClaim.Name, nodeClaim.Status.NodeName), nil
```
njtran, line 225: probably would want to pick one or the other here, including just the nodeclaim name or node name
```suggestion
			return fmt.Sprintf("nodeclaim %s/%s is marked for deletion", nodeClaim.Name, nodeClaim.Status.NodeName), nil
```

pkg/operator/operator.go
```diff
@@ -179,6 +179,9 @@ func NewOperator() (context.Context, *Operator) {
 	lo.Must0(mgr.GetFieldIndexer().IndexField(ctx, &v1.NodeClaim{}, "status.providerID", func(o client.Object) []string {
 		return []string{o.(*v1.NodeClaim).Status.ProviderID}
 	}), "failed to setup nodeclaim provider id indexer")
+	lo.Must0(mgr.GetFieldIndexer().IndexField(ctx, &v1.NodeClaim{}, "status.nodeName", func(o client.Object) []string {
```
njtran, line 182: probably don't need this and can rely on providerid

pkg/controllers/disruption/controller.go
```diff
@@ -230,9 +230,21 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+	})...)
+	if err != nil {
+		return multierr.Append(fmt.Errorf("updating nodeclaim status: %w", err), state.RequireNoScheduleTaint(ctx, c.kubeClient, false, stateNodes...))
+	}
+
 	if err := c.queue.Add(orchestration.NewCommand(nodeClaimNames,
 		lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode }), commandID, m.Reason(), m.ConsolidationType())); err != nil {
 		c.cluster.UnmarkForDeletion(providerIDs...)
+		err = multierr.Combine(err, multierr.Combine(lo.Map(cmd.candidates, func(candidate *Candidate, _ int) error {
+			return multierr.Append(candidate.NodeClaim.StatusConditions().Clear(v1.ConditionTypeDisruptionCandidate), c.kubeClient.Status().Update(ctx, candidate.NodeClaim))
```
njtran, line 246: We try to do this at the top of the disruption loop now. You should be able to clear the condition for any nodeclaim that doesn't exist in the orchestration queue

njtran (commented):
Nice work!

pkg/apis/v1/nodeclaim_status.go
```diff
@@ -30,6 +30,7 @@ const (
 	ConditionTypeDrifted              = "Drifted"
 	ConditionTypeInstanceTerminating  = "InstanceTerminating"
 	ConditionTypeConsistentStateFound = "ConsistentStateFound"
+	ConditionTypeDisruptionCandidate  = "DisruptionCandidate"
```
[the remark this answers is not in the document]
cnmcavoy: sgtm

pkg/controllers/disruption/controller.go
```diff
@@ -230,9 +230,21 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
 	// We have the new NodeClaims created at the API server so mark the old NodeClaims for deletion
 	c.cluster.MarkForDeletion(providerIDs...)
 
+	// Set the status of the nodeclaims to reflect that they are disruption candidates
+	err = multierr.Combine(lo.Map(cmd.candidates, func(candidate *Candidate, _ int) error {
+		candidate.NodeClaim.StatusConditions().SetTrueWithReason(v1.ConditionTypeDisruptionCandidate, v1.ConditionTypeDisruptionCandidate, m.EvictionReason(candidate.NodeClaim))
```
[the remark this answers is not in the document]
cnmcavoy: Funny enough, it didn't exist when I first wrote this. Refactored to use it.

pkg/controllers/node/termination/terminator/eviction.go
cnmcavoy: I agree, `providerID` would be better... How can we achieve this? The Pod resource does not have access to the providerID, except indirectly with the `nodeName`.

Is it safe to look up the node when we add a pod to the queue? Should the interface be modified to include the nodeClaim so we can take it off that resource? Any preferences?

pkg/controllers/node/termination/terminator/events/events.go
cnmcavoy: Added

pkg/operator/operator.go
cnmcavoy: I don't think we can use providerID here, see above comment.

pkg/controllers/disruption/controller.go
```diff
@@ -237,10 +241,26 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
+		}
+		nodeClaim.StatusConditions().SetTrueWithReason(v1.ConditionTypeDisruptionReason, v1.ConditionTypeDisruptionReason, string(m.Reason()))
+		return c.kubeClient.Status().Update(ctx, nodeClaim)
+	})...)
+	if err != nil {
+		c.cluster.UnmarkForDeletion(providerIDs...)
+		return fmt.Errorf("updating nodeclaim status: %w", err)
+	}
+
+	statenodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode })
```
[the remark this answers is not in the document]
cnmcavoy: Done.

pkg/controllers/state/statenode.go
```diff
@@ -517,3 +517,24 @@ func RequireNoScheduleTaint(ctx context.Context, kubeClient client.Client, addTa
...
+		if err != nil {
+			return err
+		}
+		if !nodeClaim.StatusConditions().IsTrue(conditionType) {
+			return nil
+		}
+		if err = nodeClaim.StatusConditions().Clear(conditionType); err != nil {
+			return err
+		}
+		return kubeClient.Status().Update(ctx, nodeClaim)
```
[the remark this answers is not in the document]
cnmcavoy: Sure, I replaced all the Update's with Patch.

pkg/controllers/state/statenode.go
```diff
@@ -517,3 +518,31 @@ func RequireNoScheduleTaint(ctx context.Context, kubeClient client.Client, addTa
 	}
 	return multiErr
 }
+
+// ClearNodeClaimsCondition will remove the conditionType from the NodeClaim status of the provided statenodes
+func ClearNodeClaimsCondition(ctx context.Context, kubeClient client.Client, conditionType string, nodes ...*StateNode) (err error, requeue bool) {
+	return multierr.Combine(lo.Map(nodes, func(s *StateNode, _ int) error {
+		if !s.Initialized() || s.NodeClaim == nil {
```
njtran, line 525: initialized should capture if the nodeclaim isn't set
```suggestion
		if !s.Initialized() {
```
https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/state/statenode.go#L320

pkg/controllers/state/statenode.go
```diff
@@ -517,3 +518,31 @@ func RequireNoScheduleTaint(ctx context.Context, kubeClient client.Client, addTa
...
+			return nil
+		}
+		nodeClaim := &v1.NodeClaim{}
+		err := kubeClient.Get(ctx, client.ObjectKeyFromObject(s.NodeClaim), nodeClaim)
+		if err != nil {
```
njtran, line 530: i think not found errors would be noisy here
```suggestion
		if client.IgnoreNotFound(err) != nil {
```

pkg/controllers/state/statenode.go
```diff
@@ -517,3 +518,31 @@ func RequireNoScheduleTaint(ctx context.Context, kubeClient client.Client, addTa
...
+			return err
+		}
+		if !nodeClaim.StatusConditions().IsTrue(conditionType) {
```
njtran, line 533: Do we want to keep the status condition around if it's false? Probably easier just to clear then do a deep equals diff?

pkg/controllers/state/statenode.go
```diff
@@ -517,3 +518,31 @@ func RequireNoScheduleTaint(ctx context.Context, kubeClient client.Client, addTa
...
+			return nil
+		}
+		stored := nodeClaim.DeepCopy()
+		_ = nodeClaim.StatusConditions().Clear(conditionType)
+
+		if err := kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if errors.IsConflict(err) {
+				requeue = true
+				return nil
+			}
```
njtran, line 543: I think personally I'd rather we not include reconcile.Result as a return value and then do a `errors.IsConflict()` in the call in the disruption controller, so that we leave the error handling up to the controller. Should make the code easier too

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -174,6 +174,19 @@ func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (rec
 		return reconcile.Result{}, fmt.Errorf("adding nodeclaim terminationGracePeriod annotation, %w", err)
 	}
 
+	// set a disruption reason if the nodeclaim was deleted manually
+	if !nodeClaim.StatusConditions().IsTrue(v1.ConditionTypeDisruptionReason) {
```
njtran, line 178: So I think we have a race here currently in the code with disruption and this controller fighting to add/remove the condition. We probably want to do the following:
1. When draining, assume the non-existence of the status condition indicates that it's forceful termination. 
2. In the disruption controller, check if something is terminating (has deletion timestamp set) before clearing the status condition, and only clear it if it's a condition that we own. The IsConflict() errors will prevent any races here if a node is terminated the same time it's removed from the disruption queue internally.

Ideally, we'd be able to extend this so that another forceful termination controller could choose to add its own status condition here, but since the status patch and nodeclaim delete calls can't be batched together, there's no way to guarantee the status condition for the forceful method will exist. I think doing "Forceful Termination" with correlating to other metrics are good enough here.

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -205,6 +213,21 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
+func evictionReason(ctx context.Context, key QueueKey, kubeClient client.Client) (string, error) {
+	nodeClaim, err := node.NodeClaimForNode(ctx, kubeClient, &corev1.Node{Spec: corev1.NodeSpec{ProviderID: key.providerID}})
+	if err != nil {
+		return "", err
+	}
+	terminationCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeDisruptionReason)
+	if terminationCondition.IsTrue() {
+		return terminationCondition.Message, nil
+	}
+	if !nodeClaim.DeletionTimestamp.IsZero() {
```
njtran, line 229: I think we don't need this check, we can just return "Forceful Deletion", nil if the condition doesn't exist.

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -171,6 +175,10 @@ func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
 // Evict returns true if successful eviction call, and false if there was an eviction-related error
 func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(key.Namespace, key.Name)))
+	evictionMessage, err := evictionReason(ctx, key, q.kubeClient)
+	if err != nil {
```
njtran, line 179: I personally think we should just silently fail here. I'm worried about super noisy logs in the case where a nodeclaim accidentally goes away, or if there are other issues with the apiserver

pkg/controllers/disruption/controller.go
```diff
@@ -204,12 +210,9 @@ func (c *Controller) executeCommand(ctx context.Context, m Method, cmd Command,
...
 	log.FromContext(ctx).WithValues("command-id", commandID, "reason", strings.ToLower(string(m.Reason()))).Info(fmt.Sprintf("disrupting nodeclaim(s) via %s", cmd))
 
-	stateNodes := lo.Map(cmd.candidates, func(c *Candidate, _ int) *state.StateNode {
-		return c.StateNode
-	})
 	// Cordon the old nodes before we launch the replacements to prevent new pods from scheduling to the old nodes
-	if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, true, stateNodes...); err != nil {
-		return fmt.Errorf("tainting nodes with %s (command-id: %s), %w", pretty.Taint(v1.DisruptedNoScheduleTaint), commandID, err)
+	if err := c.MarkDisrupted(ctx, m, cmd.candidates...); err != nil {
+		return fmt.Errorf("marking disrupted (command-id: %s), %w", commandID, err)
```
njtran, line 215: I think we probably want to mask any notfound errors here.

pkg/controllers/disruption/controller.go
```diff
@@ -265,6 +265,34 @@ func (c *Controller) createReplacementNodeClaims(ctx context.Context, m Method,
...
+		if err != nil {
+			return err
+		}
+		nodeClaim.StatusConditions().SetTrueWithReason(v1.ConditionTypeDisruptionReason, v1.ConditionTypeDisruptionReason, string(m.Reason()))
+		return c.kubeClient.Status().Update(ctx, nodeClaim)
+	})...)
+	if err != nil {
+		c.cluster.UnmarkForDeletion(providerIDs...)
+		return fmt.Errorf("updating nodeclaim status: %w", err)
+	}
```
njtran, line 292: I think you can remove this as part of the beginning of the disruption reconcile loop

njtran (commented):
Nice work, think we're getting close!

pkg/controllers/state/statenode.go
cnmcavoy: Removing this check results in test panics - I don't think your assertions is correct. Here is `Initialized()`:
```
func (in *StateNode) Initialized() bool {
	// Node is managed by Karpenter, so we can check for the Initialized label
	if in.Managed() {
		return in.Node != nil && in.Node.Labels[v1.NodeInitializedLabelKey] == "true"
	}
	// Nodes not managed by Karpenter are always considered Initialized
	return true
}
```
`in.Managed()` return `false` because `in.NodeClaim` is nil. So initialized actually returns `true` when the nodeclaim is `nil`.

pkg/controllers/disruption/controller.go
cnmcavoy: I masked the errors in `MarkDisrupted`, so this code didn't change.

pkg/controllers/node/termination/terminator/eviction.go
cnmcavoy: I hate silently failing, bc when debugging I often end up having to re-add all the logs for suppressed errors. So I compromised and moved it to `V(1)`.

pkg/controllers/state/statenode.go
njtran: Most of the time, the nodeclaim should exist if the node has the initialized label, but it's definitely possible for it to go away.

njtran (commented):
/lgtm
/approve

thanks for the contribution!

PR #1379, opened 2024-07-01
chore: Add an instancetype controller to encapsulate caching logic

ellistarn:
Fixes #N/A 

**Description**
This is a first step in a multi-PR change that causes the Karpenter codebase to get instance types from an abstraction layer. In the future, we will hang Ice caching and #1305 on top of this provider.

* `make presubmit`
* Manually

### DIFF
--- a/pkg/controllers/instancetype/controller.go
+++ b/pkg/controllers/instancetype/controller.go
@@ -1,0 +1,54 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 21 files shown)


pkg/controllers/instancetype/controller.go
sftim, line 1: (nit) We might call this a machinetype controller or a nodetype controller. “Instance” isn't a term we use much within Kubernetes.

ellistarn: Karpenter, thus far, has followed the convention defined in the upstream well known label `node.kubernetes.io/instance-type`. I'm not aware of the term `machinetype` or `nodetype` being used upstream in the project. I know CAPI uses the term `machine`, but I don't see this as a stronger "standard" then `instance-type`, which has been around a lot longer.

PR #1380, opened 2024-07-01
chore: update kwok to v0.5.2 and scale back replicas

njtran:
Fixes #N/A 

**Description**
Upgrades kwok to the latest version, and scales back the number of partitions to 1, so that multiple kwok controllers aren't competing for a kind cluster's limited control plane resources.

Note: if you were using kwok before this PR, you'll need to `make uninstall-kwok` on a commit before this to fully clean up all kwok controllers/resources.

make install-kwok && make apply 


jonathan-innis (approved):
/lgtm
/approve

PR #1382, opened 2024-07-02
chore: Remove usage of `knative.dev` logger

jonathan-innis:
Fixes #N/A 

**Description**

Remove spurious log line for `terminated nodeclaim` that was causing the following log line with the fallback logger

```
{"level":"info","ts":1719908125.4508243,"logger":"fallback","caller":"termination/controller.go:139","msg":"deleted nodeclaim"}
```

This also replaces the usage of the `logging.FromContext(ctx).Error()` call that was being made in the disruption controller

`make presubmit`


njtran (commented):
/lgtm
/approve

PR #1393, opened 2024-07-05
chore: Add Kubelet Annotation for the Conversion webhooks

engedaam:
Fixes #N/A 

**Description**
- Adding Kubelet annotation on on v1 NodePool to allow for conversion between v1beta1 and v1 NodePool
- Since v1 NodePool does not have the Kubelet field, the conversion webhooks will need to store v1beta1 Kubelet field as an annotation to avoid data loss when v1 API will be the storage version of the NodePool CRD. 

- `make presubmit`


pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -63,7 +63,15 @@ func (in *NodeClaimSpec) convertTo(ctx context.Context, v1beta1nc *v1beta1.NodeC
 		}
 	}
 
-	// Need to implement Kubelet Conversion
+	if kubeletAnnotation != "" {
+		v1beta1kubelet := &v1beta1.KubeletConfiguration{}
+		err := json.Unmarshal([]byte(kubeletAnnotation), v1beta1kubelet)
+		if err != nil {
+			return err
```
njtran, line 70: add a wrapper so we can know where this is coming from
```suggestion
			return fmt.Errorf("unmarshaling kubelet config annotation, %w", err)
```

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -101,16 +115,19 @@ func (in *NodeClaimSpec) convertFrom(ctx context.Context, v1beta1nc *v1beta1.Nod
...
 		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, nodeclassGroupVersion.Group),
 	}
 
-	// Need to implement Kubelet Conversion
-	return nil
+	kubelet, err := json.Marshal(v1beta1nc.Kubelet)
+	if err != nil {
+		return "", err
+	}
+	return string(kubelet), nil
```
njtran, line 130: Is there any key validation we need to do to make sure it can be inputted as an annotation?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -100,7 +101,15 @@ func (in *NodeClaimTemplate) convertTo(ctx context.Context, v1beta1np *v1beta1.N
 		APIVersion: lo.Ternary(found, matchingNodeClass.GroupVersion().String(), ""),
 	}
 
-	// Need to implement Kubelet Conversion
+	if kubeletAnnotation != "" {
+		v1beta1kubelet := &v1beta1.KubeletConfiguration{}
+		err := json.Unmarshal([]byte(kubeletAnnotation), v1beta1kubelet)
+		if err != nil {
+			return err
```
njtran, line 108: same here on error wrapping

pkg/apis/v1/nodeclaim_conversion.go
engedaam: CEL validation will catch that, and deny the update

njtran: I'm wondering if the set of characters that we can marshal to are mutually exclusive from the disallowed characters in a kubernetes annotation. If they're not mutually exclusive, then we could potentially run into a case where a kubelet config never can be converted to an annotation.

engedaam: The values of annotation do not contain validation. This means that the characters that we can marshal to are mutually exclusive and should not cause a problem. https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

njtran: Ah gotcha. Just tried everything myself. Thanks!

njtran (commented):
/lgtm
/approve

PR #1399, opened 2024-07-08
test: fix kind cluster test cleanup flakes

njtran:
Fixes #N/A 

**Description**
Fix test flakes that were preventing kind clusters from cleaning up. Resources were stuck in a foreground deleting, where these resources are all fake, so we don't need to wait for all resources to be cleaned up. 

make presubmit

### DIFF
--- a/test/pkg/environment/common/setup.go
+++ b/test/pkg/environment/common/setup.go
@@ -56,12 +56,12 @@ var (
 		&v1.PersistentVolumeClaim{},
 		&v1.PersistentVolume{},
 		&storagev1.StorageClass{},
-		&v1beta1.NodePool{},
-		&v1.LimitRange{},
-		&schedulingv1.PriorityClass{},
 		&v1.Node{},
 		&v1beta1.NodeClaim{},
+		&v1beta1.NodePool{},
 		&v1alpha1.KWOKNodeClass{},
+		&v1.LimitRange{},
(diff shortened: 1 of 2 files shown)


test/pkg/environment/common/setup.go
```diff
@@ -135,7 +135,6 @@ func (env *Environment) CleanupObjects(cleanableObjects ...client.Object) {
 					defer GinkgoRecover()
 					g.Expect(env.ExpectTestingFinalizerRemoved(&metaList.Items[i])).To(Succeed())
 					g.Expect(client.IgnoreNotFound(env.Client.Delete(env, &metaList.Items[i],
```
ellistarn, line 137: Can we use https://github.com/awslabs/operatorpkg/blob/main/test/expectations/expectations.go#L156

njtran: Just tried this. We need some way to only delete objects with labels that we care about, which we use test.DiscoveryLabel right now to only delete nodes that were created in the test. Happy to include in a followup when that capability is added in

ellistarn (approved):

test/pkg/environment/common/setup.go
njtran: Tried removing nodes from the list, the cleanup also cleans up all namespaces for resources, which we want to keep around across tests

PR #1405, opened 2024-07-10
fix: incorrect early return location blocking multi-node consolidation

cnmcavoy:
Fixes #N/A 

**Description**
 Incorrect early return location blocking multi-node consolidation


njtran:
This should be correct. We won't consolidate a node if it would require two nodes to be created. The second len check is just how we've decided to emit logs in reference to a single node consolidation or a multi node consolidation decision. If we emit on multi node consolidation, the event doesn't make sense without the awareness of the other nodes that were included in the event.

PR #1426, opened 2024-07-15
chore: move expireAfter to NodeClaimTemplate

njtran:
Fixes #N/A 

**Description**
Moves ExpireAfter to the NodeClaimTemplate in the NodePool. This makes it so that the NodeClaim doesn't require a NodePool to enforce a maximum node lifetime.

Also adds in some fixes for the conversion webhooks for the kwok provider. 

make presubmit && make apply 


hack/mutation/nodepool.sh
```diff
@@ -1,3 +1,8 @@
...
 
-yq eval 'del(.spec.versions[0].schema.openAPIV3Schema.properties.spec.properties.template.properties.spec.properties.resources)' -i pkg/apis/crds/karpenter.sh_nodepools.yaml
\ No newline at end of file
+# Add the conversion stanza to the CRD spec to enable conversion via webhook
+yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i kwok/apis/crds/karpenter.kwok.sh_kwoknodeclasses.yaml
+yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+yq eval '.spec.conversion = {"strategy": "Webhook", "webhook": {"conversionReviewVersions": ["v1beta1", "v1"], "clientConfig": {"service": {"name": "karpenter", "namespace": "kube-system", "port": 8443}}}}' -i pkg/apis/crds/karpenter.sh_nodepools.yaml
```
ellistarn, line 8: Why are we bothering with conversion for KWOK? Who needs auto upgrade for it?

pkg/operator/injection/injection.go
```diff
@@ -63,6 +65,19 @@ func WithOptionsOrDie(ctx context.Context, opts ...options.Injectable) context.C
 	return ctx
 }
 
+func WithClient(ctx context.Context, client client.Client) context.Context {
+	return context.WithValue(ctx, clientKey, client)
+}
+
+func GetClient(ctx context.Context) client.Client {
```
ellistarn, line 72: IIRC @jonathan-innis intentionally was avoiding this pattern, though I don't know his rationale. I don't see a great alternative though.

pkg/operator/injection/injection.go
```diff
@@ -71,7 +86,7 @@ func GetNodeClasses(ctx context.Context) []schema.GroupVersionKind {
 	retval := ctx.Value(nodeClassKey)
 	if retval == nil {
 		// This is a developer error if this happens, so we should panic
-		logging.FromContext(ctx).Fatal("nodeclass doesn't exist in context")
+		log.Fatal("nodeclass doesn't exist in context")
```
ellistarn, line 89:
```suggestion
		log.FromContext("nodeclass doesn't exist in context")
```

pkg/operator/injection/injection.go
```diff
@@ -71,7 +86,7 @@ func GetNodeClasses(ctx context.Context) []schema.GroupVersionKind {
 	retval := ctx.Value(nodeClassKey)
 	if retval == nil {
```
ellistarn, line 87: You could just skip this check. It'll be clear from the trace where/why this crashed.

pkg/operator/injection/injection.go
njtran: I don't personally see a way around it either

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -94,7 +98,26 @@ func (in *NodeClaim) ConvertFrom(ctx context.Context, from apis.Convertible) err
...
+// only need to set expireAfter for v1beta1 to v1
+func (in *NodeClaim) setExpireAfter(ctx context.Context, v1beta1nc *v1beta1.NodeClaim) error {
+	kubeClient := injection.GetClient(ctx)
+	nodePoolName, ok := v1beta1nc.Labels[NodePoolLabelKey]
+	if !ok {
+		// If we don't have a nodepool for this nodeclaim, there's nothing to look up
+		return nil
+	}
+	nodePool := &NodePool{}
+	if err := kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
```
engedaam, line 113: What do you think about adding an agnation that has the expireAfter value? That way we can make less call to the API server

pkg/controllers/disruption/drift_test.go
```diff
@@ -49,14 +49,14 @@ var _ = Describe("Drift", func() {
...
 					ConsolidateAfter: &v1.NillableDuration{Duration: nil},
-					ExpireAfter:      v1.NillableDuration{Duration: nil},
 					// Disrupt away!
 					Budgets: []v1.Budget{{
 						Nodes: "100%",
 					}},
 				},
 			},
 		})
+		nodePool.Spec.Template.Spec.ExpireAfter = v1.NillableDuration{Duration: nil}
```
engedaam, line 59: nit. Can we just add this as part of the `test.NodePool()`?

pkg/controllers/nodeclaim/disruption/expiration_test.go
```diff
@@ -42,11 +42,13 @@ var _ = Describe("Expiration", func() {
...
 				Labels: map[string]string{v1.NodePoolLabelKey: nodePool.Name},
 			},
+			Spec: v1.NodeClaimSpec{
+				ExpireAfter: v1.NillableDuration{Duration: lo.ToPtr(time.Second * 30)},
+			},
 		})
 	})
 	Context("Metrics", func() {
 		It("should fire a karpenter_nodeclaims_disrupted metric when expired", func() {
-			nodePool.Spec.Disruption.ExpireAfter.Duration = lo.ToPtr(time.Second * 30)
```
engedaam, line 49: nit. Can we now just remove the nodePool from this test file?

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -120,7 +120,7 @@ var _ = Describe("Static Drift Hash", func() {
 		nodePool.Spec.Limits = v1.Limits(corev1.ResourceList{"cpu": resource.MustParse("16")})
 		nodePool.Spec.Disruption.ConsolidationPolicy = v1.ConsolidationPolicyWhenEmpty
 		nodePool.Spec.Disruption.ConsolidateAfter = &v1.NillableDuration{Duration: lo.ToPtr(30 * time.Second)}
-		nodePool.Spec.Disruption.ExpireAfter.Duration = lo.ToPtr(30 * time.Second)
+		nodePool.Spec.Template.Spec.ExpireAfter.Duration = lo.ToPtr(30 * time.Second)
```
engedaam, line 123: Don't we expect the hash to change here? Isn't expireAfter a driftable field?

pkg/controllers/nodeclaim/disruption/suite_test.go
```diff
@@ -116,7 +116,7 @@ var _ = Describe("Disruption", func() {
 	})
 	It("should remove multiple disruption conditions simultaneously", func() {
```
engedaam, line 117: I think we should testing that we can trigger drift by updating the expireAfter field

pkg/apis/v1/nodeclaim.go
```diff
@@ -61,6 +61,16 @@ type NodeClaimSpec struct {
...
+	// ExpireAfter is the duration the controller will wait
+	// before terminating a node, measured from when the node is created. This
+	// is useful to implement features like eventually consistent node upgrade,
+	// memory leak protection, and disruption testing.
+	// +kubebuilder:default:="720h"
+	// +kubebuilder:validation:Pattern=`^(([0-9]+(s|m|h))+)|(Never)$`
+	// +kubebuilder:validation:Type="string"
+	// +kubebuilder:validation:Schemaless
+	// +optional
+	ExpireAfter NillableDuration `json:"expireAfter,omitempty"`
```
engedaam, line 73: Since this is a field that can be hashed, we should problem add a test for it

engedaam: Just validating, we would like this change to be included in v1, right? if it's not included we will need to make sure that we bump the hash version

pkg/utils/disruption/disruption.go
```diff
@@ -36,9 +36,9 @@ import (
 // disruption cost is highest, and it approaches zero as the node ages towards its expiration time.
 func LifetimeRemaining(clock clock.Clock, nodePool *v1.NodePool, nodeClaim *v1.NodeClaim) float64 {
 	remaining := 1.0
-	if nodePool.Spec.Disruption.ExpireAfter.Duration != nil {
+	if nodePool.Spec.Template.Spec.ExpireAfter.Duration != nil {
```
engedaam, line 39: Does it matter here where we are getting the expireAfter value? NodePool or NodeClaims?  Only asked for the case of a drifted NodeClaims where the two values can be different

pkg/apis/v1/nodepool_validation_cel_test.go
```diff
@@ -64,15 +64,15 @@ var _ = Describe("CEL/Validation", func() {
 	})
 	Context("Disruption", func() {
 		It("should fail on negative expireAfter", func() {
```
engedaam, line 66: nit. Can we move these test outside the disruption block? since they are meant to be in NodePool Template Spec

engedaam (commented):
Just few nits

hack/kwok/conversion.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
njtran: I think that's what I'm doing, no?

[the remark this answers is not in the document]
engedaam: I meant creating a hack folder in the kowk provider

pkg/apis/v1/nodeclaim.go
njtran: Good call. Will add it in

pkg/utils/disruption/disruption.go
njtran: That's a good point, this should check the nodeclaim.

pkg/apis/v1/nodeclaim_conversion.go
ellistarn: Reads are cached, no?

pkg/apis/v1/nodepool_validation_cel_test.go
njtran: I was thinking they're fine as is, since it's about disruption. The other sections are just labels, taints, and requirements. WDYT?

pkg/controllers/nodeclaim/disruption/expiration_test.go
njtran: good call. I kept it in for some of the metrics tests, but removed it elsewhere

hack/kwok/conversion.sh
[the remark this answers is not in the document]
njtran: I think that might be more confusing to have some hack files not at the top level. Having them in their own folder here I think is the more logical way to find these, but know not to use them if you're not using kwok

pkg/apis/v1/nodeclaim.go
njtran: Yeah it should be in v1

pkg/apis/v1/nodeclaim_conversion.go
engedaam: That'a fair point.

ellistarn (approved):
/lgtm

@engedaam had some good comments, and can probably provide a deeper review than me these days.

pkg/controllers/nodepool/hash/suite_test.go
```diff
@@ -103,6 +106,7 @@ var _ = Describe("Static Drift Hash", func() {
 
 		nodePool.Spec.Template.Labels = map[string]string{"keyLabeltest": "valueLabeltest"}
 		nodePool.Spec.Template.Annotations = map[string]string{"keyAnnotation2": "valueAnnotation2", "keyAnnotation": "valueAnnotation"}
+		ExpectApplied(ctx, env.Client, nodePool)
```
ellistarn, line 109: Can we do this at the top where we apply it on line 102?

PR #1427, opened 2024-07-15
test: implement nodeclaim.spec.terminationGracePeriod CEL tests

wmgroot:
**Description**
Implements CEL tests for the new `terminationGracePeriod` field on NodeClaim.

### DIFF
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -267,6 +267,7 @@ spec:
 
                     The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
                     If left undefined, the controller will wait indefinitely for pods to be drained.
+                  pattern: ^([0-9]+(s|m|h))+$
                   type: string
               required:
                 - nodeClassRef
@@ -731,6 +732,7 @@ spec:
 
                     The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
                     If left undefined, the controller will wait indefinitely for pods to be drained.
+                  pattern: ^([0-9]+(s|m|h))+$
                   type: string
               required:
                 - nodeClassRef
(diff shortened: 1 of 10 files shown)


pkg/apis/v1/nodeclaim.go
```diff
@@ -59,6 +59,9 @@ type NodeClaimSpec struct {
 	//
 	// The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
 	// If left undefined, the controller will wait indefinitely for pods to be drained.
+	// +kubebuilder:validation:Minimum:=0
```
njtran, line 62: you shouldn't need the minimum. We just do the validation by regex.

pkg/apis/v1beta1/nodeclaim.go
```diff
@@ -66,6 +66,8 @@ type NodeClaimSpec struct {
 	//
 	// The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
 	// If left undefined, the controller will wait indefinitely for pods to be drained.
+	// +kubebuilder:validation:Pattern=`^([0-9]+(\\.[0-9]+)?(ms|s|m|h|d|w|y))+$`
```
njtran, line 69: I discussed with @wmgroot, but we intentionally only keep it as s/m/h, due to inconsistencies in how days/weeks/years are calculated.  @cnmcavoy can you add that back in as it was? 

https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1/nodepool.go#L84

wmgroot: I'm testing out the branch with day+ removed.

wmgroot: Should be all good now.

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -267,6 +267,7 @@ spec:
 
                     The feature can also be used to allow maximum time limits for long-running jobs which can delay node termination with preStop hooks.
                     If left undefined, the controller will wait indefinitely for pods to be drained.
+                  pattern: ^([0-9]+(\\.[0-9]+)?(s|m|h))+$
```
njtran, line 270: why do we need the middle bit? 
```suggestion
                  pattern: ^(([0-9]+(s|m|h))+)$
```

njtran (commented):
/lgtm
/approve

Thanks for the commit!

PR #1447, opened 2024-07-18
feat: Adding StartDateTime, EndDateTime, and Timezones to NodePool Budgets

### LINKED ISSUE #1446: feat: StartDateTime and EndDateTime, Timezones for NodePool Disruption Budgets
### Description

As apart of planned events, there are instances in which budgets should be schedule-able within a human readable format that can be easily understood and reasoned about.

This feature should also follow the current standard of minimal ("most restrictive") budgeting when multiple budgets are defined with overlapping timeframes or crontab schedules.

This feature would enable operators to schedule budgets for many one-off events in the future when they are known about, without having to create and compare complex crontab definitions against one another.

ant-smith:
Fixes #1446

**Description**

This feature adds the capability for users to define a StartDateTime and EndDateTime for their nodepools, allowing single duration scheduling of allowed disruptions.

The functionality of the API stays the same if no StartDateTime and EndDateTime are defined.

The budget also now takes into account timezones for the defined cronSchedule and dateTime's. Without defining a timezone the functionality remains unchanged and will default to "UTC". Timezones follow the standard IANA "America/New_York" format.

Both StartDateTime and EndDateTime are defined in the RFC3339 standard formats minus the "(+|-)0700" timezone at the end of the format as timezones are supported as a separate parameter.

This feature overall will assist operations teams in the following ways:

- Overall more human readable, allowing an operator to easily reason if a budget is active or inactive at a glance
- Support for many differing, but short lived time periods of disruption budgets over a longer timespan. E.g. there may be 15-18, 3-6 hour windows, over the course of 6 weeks that limit the clusters disruption in different ways. Writing cron's for this would be more difficult that start-end datetimes.
- Support for empty start datetime, allows operators to say "From today until X, I want this to be active"
- Support for empty end datetime, allows for operators to say "From today moving forward, I want this to be active"
- Support for different timezones allow operators in separate timezones to manage separate sets of budgets, without converting to UTC first.
- Dates are communicated in PST, needing to convert them adds friction and leaves a chance for errors as well as harder to quickly verify.

Extensive tests were written within the test suite to test the existing functionality with timezones added, along with testing the functionality of the new DateTime feature.


tvonhacht-apple (approved):

njtran:
Hey @ant-smith, I notice there are a couple of use-case you're trying to solve, is it possible to pare them down to smaller distinct ideas? At a high level (stealing from your description), I see these as the core asks: 

1. human readable schedule formats
2. a one-off "active until" schedule
3. a one-off "active since" schedule
4. supporting timezones for a budget

Is it possible to tackle these one by one? The best way would be to have an RFC and cut issues to get a gist of how important each issue is to other users. 

Specifically speaking, would the following solutions help? 
- For 1: Post this information into the status. e.g. each budget could have a "nextHit" entry in the status
- For 3: an infinite duration - the current schedule semantic reasons about starts to the schedule, rather than number 2 which reasons about the ends.

For 4: Seems like a simple ease-of-use add that I think we should probably add in. Just need to verify how this works in practice.

flavono123:
> supporting timezones for a budget

may i do another PR only for this? @njtran @ant-smith

PR #1453, opened 2024-07-22
feat: implement consolidateAfter

### LINKED ISSUE #735: Consolidation ttl: `spec.disruption.consolidateAfter`

We have a cluster where there are a lot of cron jobs which run every 5 minutes...

This means we have 5 nodes for our base workloads and every 5 minutes we get additional nodes for 2-3 minutes which are scaled down or consolidated with existing nodes.

This leads to a constant flow of nodes joining and leaving the cluster. It looks like the docker image pull and node initialization creates more network traffic fees than the cost reduction of not having running the instances all the time.

It would be great if we could configure some time consolidation period maybe together with ttlSecondsAfterEmpty which would only cleanup or consolidate nodes if the capacity was idling for x amount of time.

Creating a special provisioner is quite time consuming because all app deployments have to be changed to leverage it...

We think about putting cronjobs into a special provisioner which would not use consolidation but the ttlSecondsAfterEmpty feature.

### Additional Context

_No response_

### Attachments

_No response_

### Community Note

njtran:
Fixes #735 

**Description**
Implements consolidateAfter as discussed in #735, and allows setting `consolidateAfter` when the `consolidationPolicy: WhenUnderutilized`. 

At a high level, Karpenter will track a timer per-node with a duration matching its NodePool's `consolidateAfter`. When the timer finishes, the node will be considered underutilized. The timer is restarted by pod events on the node:
1. When a pod is removed from the node
2. When a pod scheduled to the node succeeds/fails (goes terminal)
3. When a pod schedules to the node

make presubmit && make apply


pkg/controllers/nodeclaim/disruption/underutilization.go
```diff
@@ -0,0 +1,98 @@
...
+	cluster    *state.Cluster
+}
+
+//nolint:gocyclo
+func (e *Underutilization) Reconcile(ctx context.Context, nodePool *v1.NodePool, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
+	hasUnderutilizedCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeUnderutilized) != nil
+
+	// 1. If ConsolidationPolicyWhenUnderutilized is not configured or ConsolidateAfter isn't configured, remove the underutilization status condition
+	if nodePool.Spec.Disruption.ConsolidationPolicy != v1.ConsolidationPolicyWhenUnderutilized ||
+		nodePool.Spec.Disruption.ConsolidateAfter == nil ||
```
jonathan-innis, line 46: I think we should make it so that this struct can't be nil and only the duration can set to be nil (set to `Never`)

pkg/controllers/nodeclaim/disruption/underutilization.go
```diff
@@ -0,0 +1,98 @@
...
+			}
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+
+	// If node is not utilized, clear the condition, and requeue when we expect the node to be underutilized
+	underutilized, timeLeft, reason := e.cluster.IsNodeUnderutilized(n.Spec.ProviderID, lo.FromPtr(nodePool.Spec.Disruption.ConsolidateAfter.Duration))
+	if !underutilized {
+		_ = nodeClaim.StatusConditions().Clear(v1.ConditionTypeUnderutilized)
```
jonathan-innis, line 79: There seems to be a little bit of inconsistency between us calling Clear() inside of our if conditionals and calling Clear() outside. Ideally, we should be consistent with this everywhere. Since this is an idempotent call, seems like it makes sense to me that you would just always clear the condition and then only log when there is a change

pkg/controllers/nodeclaim/disruption/underutilization.go
```diff
@@ -0,0 +1,98 @@
...
+
+		if hasUnderutilizedCondition {
+			log.FromContext(ctx).V(1).WithValues("calculatedFrom", reason).Info("removing underutilization status condition")
```
jonathan-innis, line 82: It's not entirely clear to me what "calculatedFrom" is.

pkg/controllers/state/cluster.go
```diff
@@ -183,6 +187,40 @@ func (c *Cluster) IsNodeNominated(providerID string) bool {
 	return false
 }
 
+// IsNodeUnderutilized returns true if the last pod event on the node was at least consolidateAfter
+// in the past. Also returns a reason to inform the user on how this was calculated.
+func (c *Cluster) IsNodeUnderutilized(providerID string, consolidateAfter time.Duration) (bool, time.Duration, string) {
```
jonathan-innis, line 192: I wonder if rather than making the cluster state methods tightly scoped to this underutilization use-case if you just expose the last cluster event that happened on the state node. That would basically mean that you wouldn't have to pass down the consolidateAfter value into the function call and I think it should reduce the number of return arguments that you have for this function call

jonathan-innis: Also, the check would get much quicker

pkg/controllers/state/cluster.go
```diff
@@ -183,6 +187,40 @@ func (c *Cluster) IsNodeNominated(providerID string) bool {
...
+	lastPodEventTime, ok := c.lastPodEvents[providerID]
+	if !ok {
+		// If the node doesn't have a pod event, we assume it's because either:
+		// 1. Karpenter has restarted
+		// 2. No pods have ever scheduled to the node
+		//
+		// So treat the last pod time as the more recent of
+		// 1. When Karpenter restarted
+		// 2. When the Node was created
+		if c.clock.Since(c.initTime) < c.clock.Since(c.nodes[providerID].NodeClaim.CreationTimestamp.Time) {
```
jonathan-innis, line 206: Rather than capturing an init time, does it make more sense to just hydrate `time.Now()` when the cluster state node creation events are first being hydrated and then you can just know if the provider ID doesn't exist in the map that the node hasn't been seen yet by the cluster state informer and so you throw an error or requeue?

pkg/controllers/state/cluster.go
```diff
@@ -73,6 +75,8 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
 		daemonSetPods:             sync.Map{},
 		nodeNameToProviderID:      map[string]string{},
 		nodeClaimNameToProviderID: map[string]string{},
+		lastPodEvents:             map[string]time.Time{},
+		initTime:                  clk.Now(),
```
jonathan-innis, line 79: I wonder how much you really need this init time if you can just use time.Now() for the node creation if you haven't seen any pod events yet.

pkg/controllers/state/cluster.go
```diff
@@ -183,6 +187,40 @@ func (c *Cluster) IsNodeNominated(providerID string) bool {
...
+	reason := ""
+	lastPodEventTime, ok := c.lastPodEvents[providerID]
+	if !ok {
+		// If the node doesn't have a pod event, we assume it's because either:
+		// 1. Karpenter has restarted
+		// 2. No pods have ever scheduled to the node
+		//
+		// So treat the last pod time as the more recent of
+		// 1. When Karpenter restarted
+		// 2. When the Node was created
```
jonathan-innis, line 205: Why not just make it `time.Now()` if there is no data in the map?

pkg/controllers/state/cluster.go
```diff
@@ -183,6 +187,40 @@ func (c *Cluster) IsNodeNominated(providerID string) bool {
...
+		if c.clock.Since(c.initTime) < c.clock.Since(c.nodes[providerID].NodeClaim.CreationTimestamp.Time) {
+			lastPodEventTime = c.initTime
+			reason = "karpenter controller initialization time"
```
jonathan-innis, line 208: I'm not sure how needed the reason is here. I could see us doing without it to start

pkg/controllers/state/cluster.go
```diff
@@ -183,6 +187,40 @@ func (c *Cluster) IsNodeNominated(providerID string) bool {
...
+		} else {
+			lastPodEventTime = c.nodes[providerID].NodeClaim.CreationTimestamp.Time
+			reason = "node creation time"
+		}
+	}
+	timeSince := c.clock.Since(lastPodEventTime)
+	// If the lastPodEventTime is somehow in the future, treat the duration as 0
+	if timeSince < 0 {
```
jonathan-innis, line 216: `timeSince = lo.Max(timeSince, time.Duration(0))`?

pkg/controllers/state/cluster.go
```diff
@@ -183,6 +187,40 @@ func (c *Cluster) IsNodeNominated(providerID string) bool {
...
+		timeSince = 0 * time.Second
+	}
+	return timeSince >= consolidateAfter,
```
jonathan-innis, line 219: IMO: we could collapse a lot of this down into just getting the last time that there was movement on the node from any pod

pkg/controllers/state/cluster.go
```diff
@@ -497,7 +544,10 @@ func (c *Cluster) populateResourceRequests(ctx context.Context, n *StateNode) er
 		if err := n.updateForPod(ctx, c.kubeClient, pod); err != nil {
 			return err
 		}
-		c.cleanupOldBindings(pod)
+		// if the pod binding is new, cleanup old bindings
+		if oldNodeName, bindingKnown := c.bindings[client.ObjectKeyFromObject(pod)]; !bindingKnown || oldNodeName != pod.Spec.NodeName {
```
jonathan-innis, line 548: Is this just a refactor or is there a behavior change here? Seems like we just pulled the `if` conditional out?

pkg/controllers/state/cluster.go
```diff
@@ -510,16 +560,21 @@ func (c *Cluster) updateNodeUsageFromPod(ctx context.Context, pod *corev1.Pod) e
 	if pod.Spec.NodeName == "" {
 		return nil
 	}
-
-	n, ok := c.nodes[c.nodeNameToProviderID[pod.Spec.NodeName]]
+	providerID := c.nodeNameToProviderID[pod.Spec.NodeName]
```
jonathan-innis, line 563: just a refactor? Do we think this is needed?

pkg/controllers/state/cluster.go
```diff
@@ -532,27 +587,24 @@ func (c *Cluster) updateNodeUsageFromPodCompletion(podKey types.NamespacedName)
 	}
 
 	delete(c.bindings, podKey)
-	n, ok := c.nodes[c.nodeNameToProviderID[nodeName]]
+	providerID := c.nodeNameToProviderID[nodeName]
```
jonathan-innis, line 590: Same question here...just a refactor, do we think this is needed?

pkg/controllers/state/cluster.go
```diff
@@ -53,7 +53,9 @@ type Cluster struct {
 	nodeNameToProviderID      map[string]string               // node name -> provider id
```
jonathan-innis, line 53: Are there any tests that we should add to cluster state around this new function that you've introduced here?

pkg/controllers/nodeclaim/disruption/underutilization_test.go
```diff
@@ -0,0 +1,272 @@
...
+	var node *corev1.Node
+	BeforeEach(func() {
+		nodePool = test.NodePool()
+		nodePool.Spec.Disruption.ConsolidationPolicy = v1.ConsolidationPolicyWhenUnderutilized
+		nodePool.Spec.Disruption.ConsolidateAfter = &v1.NillableDuration{Duration: lo.ToPtr(1 * time.Minute)}
+		nodeClaim, node = test.NodeClaimAndNode(v1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					v1.NodePoolLabelKey:            nodePool.Name,
+					corev1.LabelInstanceTypeStable: "default-instance-type", // need the instance type for the cluster state update
```
jonathan-innis, line 50: I'm wondering if we should still be requiring this anymore. Maybe a tech debt thing that we should look at in a separate PR, but from what I remember, we shouldn't be caring about the instance type from the cluster state controller anymore since we should just be looking up the allocatable and the capacity from the NodeClaim directly

pkg/controllers/nodeclaim/disruption/underutilization_test.go
```diff
@@ -0,0 +1,272 @@
...
+				},
+			},
+		})
+		cluster.SetInitTime(fakeClock.Now().Add(-5 * time.Minute))
+	})
+	Context("Metrics", func() {
+		It("should fire a karpenter_nodeclaims_disrupted metric when underutilized", func() {
+			ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
+			fmt.Printf("this is the fake clock time: %s\n this is the apiserver time %s", fakeClock.Now(), nodeClaim.CreationTimestamp.Time)
```
jonathan-innis, line 59: drop?

pkg/controllers/nodeclaim/disruption/underutilization_test.go
```diff
@@ -0,0 +1,272 @@
...
+			// set the fake clock to the same time as the creation timestamp of the nodeclaim.
+			// this is necessary to avoid differences in the api server clock vs our injected clock
+			fakeClock.SetTime(nodeClaim.CreationTimestamp.Time)
+
+			fakeClock.Step(1 * time.Minute)
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*corev1.Node{node}, []*v1.NodeClaim{nodeClaim})
```
jonathan-innis, line 65: Seems odd to me that the initialization and state updates would happen after we step a minute. My expectation would be that when the node gets seen for the first time that this would be the first time that it gets "represented" inside of the cluster state. If there is a delay in this happening due to some networking issues or something that causes this value to be larger than the creation time, that should still be alright IMO.

pkg/controllers/nodeclaim/disruption/underutilization_test.go
```diff
@@ -0,0 +1,272 @@
...
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*corev1.Node{node}, []*v1.NodeClaim{nodeClaim})
+		fakeClock.Step(30 * time.Second)
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimDisruptionController, nodeClaim)
+
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+		Expect(nodeClaim.StatusConditions().Get(v1.ConditionTypeUnderutilized).IsTrue()).To(BeFalse())
+		fakeClock.Step(30 * time.Second)
+
+		timeNow := fakeClock.Now()
+		fmt.Printf("this is the time now %s\n", timeNow)
```
jonathan-innis, line 95: drop?

pkg/controllers/nodeclaim/disruption/underutilization_test.go
```diff
@@ -0,0 +1,272 @@
...
+		// Pod owned by a Deployment
+		pod := test.Pod(test.PodOptions{
+			NodeName: node.Name,
+		})
+		ExpectApplied(ctx, env.Client, pod)
+		ExpectReconcileSucceeded(ctx, podStateController, client.ObjectKeyFromObject(pod))
+
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimDisruptionController, nodeClaim)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+		Expect(nodeClaim.StatusConditions().Get(v1.ConditionTypeUnderutilized).IsTrue()).To(BeFalse())
```
jonathan-innis, line 146: I wonder if we should have a stronger bar here. Rather than just checking if True() is false, can we validate that the status condition doesn't exist? e.g. We get `nil` from the `Get()` operation?

pkg/controllers/nodeclaim/disruption/underutilization_test.go
```diff
@@ -0,0 +1,272 @@
...
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimDisruptionController, nodeClaim)
+
+		fakeClock.Step(30 * time.Second)
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*corev1.Node{node}, []*v1.NodeClaim{nodeClaim})
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimDisruptionController, nodeClaim)
+
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+		Expect(nodeClaim.StatusConditions().Get(v1.ConditionTypeUnderutilized).IsTrue()).To(BeTrue())
+	})
+	It("should clear the underutilized condition when a pod is added", func() {
```
jonathan-innis, line 127: Do we have a test that adds a pods and then ensures that the underutilized condition isn't removed until we have fully exceeded the consolidateAfter timeout?

pkg/apis/v1/nodeclaim_status.go
```diff
@@ -27,6 +27,7 @@ const (
 	ConditionTypeInitialized          = "Initialized"
 	ConditionTypeEmpty                = "Empty"
 	ConditionTypeDrifted              = "Drifted"
+	ConditionTypeUnderutilized        = "Underutilized"
```
jonathan-innis, line 30: Did we think about if representing the status condition as "Underutilized" is the right name here? Thinking more on this -- if a user is setting this value to 0 to get their old behavior, they are going to have all of their nodes marked as underutilized which I could see being confusing -- I wonder if we considered other names here

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -50,17 +50,20 @@ type Controller struct {
 	kubeClient    client.Client
 	cloudProvider cloudprovider.CloudProvider
 
-	drift     *Drift
-	emptiness *Emptiness
+	drift            *Drift
+	emptiness        *Emptiness
+	underutilization *Underutilization
```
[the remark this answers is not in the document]
njtran: done

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,134 @@
...
+
+	// If the node isn't owned by Karpenter, don't do anything
+	if _, ok := node.Labels[v1.NodePoolLabelKey]; !ok {
+		return reconcile.Result{}, nil
+	}
+
+	// If there's no associated node claim, it's not a karpenter owned node.
+	nodeClaims, err := nodeutils.GetNodeClaims(ctx, node, c.kubeClient)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("getting nodeclaims for node, %w", err))
```
jonathan-innis, line 81: There's ways to ignore errors within the `nodeutils` file for NodeClaims that I think are different than `client.IgnoreNotFound`. We should use those functions instead because I believe we return custom errors here

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,134 @@
...
+		return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("getting node, %w", err))
+	}
+
+	// If the node isn't owned by Karpenter, don't do anything
+	if _, ok := node.Labels[v1.NodePoolLabelKey]; !ok {
+		return reconcile.Result{}, nil
+	}
+
+	// If there's no associated node claim, it's not a karpenter owned node.
+	nodeClaims, err := nodeutils.GetNodeClaims(ctx, node, c.kubeClient)
```
jonathan-innis, line 79: Should we have the same logic that we have in the nodeclaim util package where we have a function called `NodeClaimForNode` that just assumes that there is a single NodeClaim/Node mapping. If that's not the case, then it will throw a DuplicateNodeError to indicate that there are duplicates and that this is an invalid state

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,134 @@
...
+	// If we've set the lastPodEvent before and it hasn't been before the timeout, don't do anything
+	if !nc.Status.LastPodEvent.Time.IsZero() && c.clock.Since(nc.Status.LastPodEvent.Time) < dedupeTimeout {
+		return reconcile.Result{}, nil
+	} else {
+		// otherwise, set the pod event time to now
+		nc.Status.LastPodEvent.Time = c.clock.Now()
+	}
+	if !equality.Semantic.DeepEqual(stored, nc) {
+		if err := c.kubeClient.Status().Patch(ctx, nc, client.MergeFrom(stored)); err != nil {
+			if errors.IsConflict(err) {
```
jonathan-innis, line 100: You don't need to handle conflicts for patches -- this should only be an issue with update operations

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,134 @@
...
+				terminal := (newPod.Spec.NodeName != "" && !podutils.IsTerminal(oldPod) && podutils.IsTerminal(newPod))
+				// if this is a newly terminating pod
+				terminating := (newPod.Spec.NodeName != "" && !podutils.IsTerminating(oldPod) && podutils.IsTerminating(newPod))
+				// return true if it was bound to a node, went terminal, or went terminating
+				return bound || terminal || terminating
+			},
+			// If the pod is deleted, but wasn't already treated as terminal or terminating
+			DeleteFunc: func(e event.TypedDeleteEvent[client.Object]) bool {
+				pod := (e.Object).(*corev1.Pod)
+				return pod.Spec.NodeName != "" && !podutils.IsTerminal(pod) && !podutils.IsTerminating(pod)
```
jonathan-innis, line 130: Is this going to work? The second you try to grab the pod with controller-runtime, it's not going to find it because there is no finalizer around. I'm also wondering how well this is going to work only because we may miss events this  way since if the pod is gone by the time we do the reconciliation, then we won't pick-up the pod and we will just exit without updating the nodeclaim at all

pkg/controllers/disruption/emptiness.go
```diff
@@ -18,74 +18,120 @@ package disruption
...
 func (e *Emptiness) Type() string {
-	return metrics.EmptinessReason
+	return metrics.ConsolidationReason
 }
 
 func (e *Emptiness) Class() string {
 	return GracefulDisruptionClass
 }
 
 func (e *Emptiness) ConsolidationType() string {
```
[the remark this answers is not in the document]
njtran: It's pretty much just because of the timeouts that we have. I can remove it in favor of hard-coded values to simplify the interface, but the metric + label would still exist

pkg/controllers/nodeclaim/podevents/controller.go
njtran: That's for nodeclaim -> node. We're looking for node first since we have the pod's nodename.

pkg/controllers/disruption/emptiness.go
```diff
@@ -18,74 +18,120 @@ package disruption
...
+		log.FromContext(ctx).V(1).Info(fmt.Sprintf("abandoning empty node consolidation attempt due to pod churn, command is no longer valid, %s", cmd))
+		return Command{}, scheduling.Results{}, nil
+	}
+
+	return cmd, scheduling.Results{}, nil
 }
 
 func (e *Emptiness) Type() string {
-	return metrics.EmptinessReason
+	return metrics.ConsolidationReason
```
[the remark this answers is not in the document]
njtran: I think this should actually just be the ConsolidationReason, since that's what limits this disruption method. If we call this emptiness, we should call the others underutilized. It's more clear IMO to name them all consolidation, and then make the consolidationTypes differentiate them.

pkg/controllers/nodeclaim/disruption/consolidation.go
```diff
@@ -0,0 +1,86 @@
...
+	timeToCheck := lo.Ternary(!nodeClaim.Status.LastPodEventTime.IsZero(), nodeClaim.Status.LastPodEventTime.Time, initialized.LastTransitionTime.Time)
+
+	// Consider a node consolidatable by looking at the lastPodEvent status field on the nodeclaim.
+	if c.clock.Since(timeToCheck) < lo.FromPtr(nodePool.Spec.Disruption.ConsolidateAfter.Duration) {
+		if hasConsolidatableCondition {
+			_ = nodeClaim.StatusConditions().Clear(v1.ConditionTypeConsolidatable)
+			log.FromContext(ctx).V(1).Info("removing consolidatable status condition")
+		}
+		consolidatableTime := nodeClaim.Status.LastPodEventTime.Add(lo.FromPtr(nodePool.Spec.Disruption.ConsolidateAfter.Duration))
+		return reconcile.Result{RequeueAfter: time.Until(consolidatableTime)}, nil
```
jonathan-innis, line 72: `time.Until` mixes the clock with the global `time.Time` IIRC. So, you probably don't want to use this function because you are going to be mixed `time.Now()` with the time value that you actually control through the clock interface

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *Controller) Reconcile(ctx context.Context, pod *corev1.Pod) (reconcile.Result, error) {
+	// If the pod doesn't have a node name, we don't know which node this pod refers to.
+	// or if this is a daemonset
+	if pod.Spec.NodeName == "" || podutils.IsOwnedByDaemonSet(pod) {
+		return reconcile.Result{}, nil
+	}
+
+	node := &corev1.Node{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: pod.Spec.NodeName}, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting node, %w", err)
```
jonathan-innis, line 69: Do we want to error if the node doesn't exist?

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,128 @@
...
+	// If the node isn't owned by Karpenter, don't do anything
+	if _, ok := node.Labels[v1.NodePoolLabelKey]; !ok {
+		return reconcile.Result{}, nil
+	}
+
+	// If there's no associated node claim, it's not a karpenter owned node.
+	nc, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+	if err != nil {
+		// if the nodeclaim doesn't exist, or has duplicates,
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
```
jonathan-innis, line 81: I think there were some Ignore functions that we had created for our NodeClaim helpers -- you could consider creating functions called `IgnoreDuplicateNodeClaimError` and `IgnoreNodeClaimNotFoundError` to make it easier for you to do this check and to match `client.IgnoreNotFound` functionality

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,128 @@
...
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("getting nodeclaims for node, %w", err)
+	}
+
+	stored := nc.DeepCopy()
+	// If we've set the lastPodEvent before and it hasn't been before the timeout, don't do anything
+	if !nc.Status.LastPodEventTime.Time.IsZero() && c.clock.Since(nc.Status.LastPodEventTime.Time) < dedupeTimeout {
+		return reconcile.Result{}, nil
+	} else {
```
jonathan-innis, line 91: Do you need an else statement if you are returning on L90?

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,128 @@
...
+				bound := oldPod.Spec.NodeName == "" && newPod.Spec.NodeName != ""
+				// if this is a newly terminal pod
+				terminal := (newPod.Spec.NodeName != "" && !podutils.IsTerminal(oldPod) && podutils.IsTerminal(newPod))
+				// if this is a newly terminating pod
+				terminating := (newPod.Spec.NodeName != "" && !podutils.IsTerminating(oldPod) && podutils.IsTerminating(newPod))
+				// return true if it was bound to a node, went terminal, or went terminating
+				return bound || terminal || terminating
+			},
+			// If the pod is deleted, but wasn't already treated as terminal or terminating
+			DeleteFunc: func(e event.TypedDeleteEvent[client.Object]) bool {
```
jonathan-innis, line 122: Didn't we decide to drop this section since we agreed its immaterial to this controller?

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+		ExpectApplied(ctx, env.Client, nodePool, node, pod)
+		fakeClock.SetTime(pod.CreationTimestamp.Time)
+		ExpectObjectReconciled(ctx, env.Client, podEventsController, pod)
+
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeZero())
+	})
+	It("should not set the nodeclaim lastPodEvent when the node does not exist", func() {
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim, pod)
+		fakeClock.SetTime(pod.CreationTimestamp.Time)
+		_ = ExpectObjectReconcileFailed(ctx, env.Client, podEventsController, pod)
```
jonathan-innis, line 125: We shouldn't fail reconciliation on this, right?

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeZero())
+	})
+	It("should not set the nodeclaim lastPodEvent when the nodeclaim does not exist", func() {
+		ExpectApplied(ctx, env.Client, nodePool, node, pod)
+		fakeClock.SetTime(pod.CreationTimestamp.Time)
+		_ = ExpectObjectReconcileFailed(ctx, env.Client, podEventsController, pod)
```
jonathan-innis, line 133: We shouldn't expect the reconciliation to fail on this, right?

jonathan-innis, line 133: Seems like expecting the reconciliation to fail should be used extremely sparingly since this means that we are returning errors and we generally want to avoid returning errors back to the user

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,169 @@
...
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeZero())
+	})
+	It("should only set the nodeclaim lastPodEvent when it hasn't been set before", func() {
+		ExpectApplied(ctx, env.Client, nodePool, node, nodeClaim, pod)
+		fakeClock.SetTime(pod.CreationTimestamp.Time)
```
jonathan-innis, line 138: Does it matter if this is set to the pod creation timestamp? It may be a little misleading IMO since this is really just ensuring that the lastPodEventTime should be set equal to clock.Now() (whatever that is)

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,169 @@
+/*
```
jonathan-innis, line 1: How do we validate the requeue functionality that we have here? Assuming that we need to add some E2Es to cover that. A lot of the core logic for how this is going to work is captured within the event filtering which means that we need some external actor to validate that we are processing the right events -- including the pod delete events that are going to cause the reconciliation here

pkg/controllers/disruption/emptiness.go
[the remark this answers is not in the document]
njtran: So are you thinking we should name the Multi and Single Node consolidations Underutilization? At a core, the method we're doing is consolidation, empty or underutilized would be a consolidationType, which is how I've modeled it. I personally think it makes more sense this way.

pkg/controllers/nodeclaim/podevents/suite_test.go
njtran: You're right this isn't needed

pkg/controllers/disruption/emptiness.go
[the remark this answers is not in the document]
njtran: I've definitely seen people reference it. We should minimally keep the metric, but keeping the label has value, as it's a core part of how Karpenter works, especially at v1. The only way this changes is if we merge them in the future, which I don't know how that will be implemented, if at all.

pkg/controllers/disruption/emptiness.go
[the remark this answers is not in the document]
jonathan-innis: > empty or underutilized would be a consolidationType

We should discuss this more offline since it sounds like we don't agree, but my thought is that if we are going to call the reason of the budgets: "Empty" and "Underutilized"; then, we should call the reason for the `nodeclaims_disrupted_total` by the same name. In terms of the method type, I could be convinced that it's consolidation, but I also don't know if you don't just model this with a reason label as well

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -0,0 +1,119 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	node := &corev1.Node{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: pod.Spec.NodeName}, node); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("getting node, %w", err))
+	}
+
+	// If the node isn't owned by Karpenter, don't do anything
+	if _, ok := node.Labels[v1.NodePoolLabelKey]; !ok {
```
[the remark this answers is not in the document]
njtran: Good call, don't know why I added in

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeZero())
+	})
+	It("should not set the nodeclaim lastPodEvent when the nodeclaim does not exist", func() {
+		ExpectApplied(ctx, env.Client, nodePool, node, pod)
+		ExpectObjectReconciled(ctx, env.Client, podEventsController, pod)
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeZero())
+	})
+	It("should only set the nodeclaim lastPodEvent when it hasn't been set before", func() {
+		ExpectApplied(ctx, env.Client, nodePool, node, nodeClaim, pod)
+		timeToCheck := fakeClock.Now().Truncate(time.Second)
```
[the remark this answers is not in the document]
njtran: There are added in ms and nanoseconds when doing fakeClock.Now()

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+	})
+	It("should set the nodeclaim lastPodEvent", func() {
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node, pod)
+		timeToCheck := fakeClock.Now().Truncate(time.Second)
+		ExpectObjectReconciled(ctx, env.Client, podEventsController, pod)
+
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeEquivalentTo(timeToCheck))
+	})
+	It("should not set the nodeclaim lastPodEvent for a node that Karpenter doesn't own", func() {
```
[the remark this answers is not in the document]
njtran: Good call

pkg/controllers/nodeclaim/podevents/suite_test.go
```diff
@@ -0,0 +1,165 @@
...
+		ExpectApplied(ctx, env.Client, nodePool, node, nodeClaim, pod)
+		timeToCheck := fakeClock.Now().Truncate(time.Second)
+		ExpectObjectReconciled(ctx, env.Client, podEventsController, pod)
+
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeEquivalentTo(timeToCheck))
+
+		fakeClock.Step(5 * time.Second)
+		ExpectObjectReconciled(ctx, env.Client, podEventsController, pod)
+		Expect(nodeClaim.Status.LastPodEventTime.Time).To(BeEquivalentTo(timeToCheck))
```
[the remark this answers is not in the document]
njtran: good catch

pkg/controllers/disruption/emptiness.go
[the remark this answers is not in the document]
jonathan-innis: Sounds good. Discussed offline. Seems like we are fine to keep this for now

jonathan-innis (approved):
/lgtm
/approve

PR #1433, opened 2024-07-16
docs: add consolidation policies rfc

cnmcavoy:
Fixes #N/A 

**Description**

RFC for https://github.com/kubernetes-sigs/karpenter/pull/1429 and https://github.com/kubernetes-sigs/karpenter/pull/1430

Let me know what topics or areas of interest need to be included or weren't fully covered. First time writing one of these.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1454, opened 2024-07-22
feat: nodeclaim.spec.minimumPriceImprovementPercent

### LINKED ISSUE #1440: Price Improvement Threshold
### Description

One piece of implementation for RFC #1433 (Consolidation Policies)

Make it possible to balance the cost of disrupting pods against the cost of using a more expensive instance type.

Currently, single-node consolidation will replace any instance with a cheaper version that can house the same pods, resulting in very frequent pod disruption. Disruption pods has its own cost, which for certain workloads can negate savings or result in a net loss of money.

This concept has already been documented as an improvement for spot consolidation, but I see no reason why it wouldn't be useful for generalized consolidation.
https://github.com/kubernetes-sigs/karpenter/blob/main/designs/spot-consolidation.md#2-price-improvement-factor

Very important. We are currently running a fork of Karpenter with single-node consolidation completely disabled due to the frequent node disruption we are seeing for small price improvements (individual pods are frequently being disrupted multiple times an hour).

wmgroot:
Fixes #1440 
Implementation component of #1433.

**Description**
This PR adds a configurable `MinimumPriceImprovementPercent` to the NodeClaim spec. This allows users control over how large of a cost improvement should be required for consolidation to disrupt pods operating on a node to replace it with a cheaper alternative.

TBD. This PR is currently a draft for implementation discussion.


pkg/apis/v1/nodeclaim.go
```diff
@@ -73,6 +73,12 @@ type NodeClaimSpec struct {
 	// +kubebuilder:validation:Schemaless
 	// +optional
 	ExpireAfter NillableDuration `json:"expireAfter,omitempty"`
+	// MinimumPriceImprovementPercent is the minimum price improvement necessary to disrupt this node, as an integer percentage.
+	// The default is 0%, which maintains the existing consolidation behavior prior to this feature.
+	// +kubebuilder:validation:Minimum:=0
+	// +kubebuilder:validation:Maximum:=100
+	// +optional
+	MinimumPriceImprovementPercent *int32 `json:"minimumPriceImprovementPercent,omitempty"`
```
ellistarn, line 81: I think this one makes sense at the NodePool level, rather than the NodeClaim level. This is because the decision to consolidate is at the pool level, unlike expiry/tgp, which make sense outside of the context of a pool (i.e. standalone nodeclaim).

wmgroot: I put it in NodeClaim because I think it could be helpful to have the flexibility to set different improvement thresholds for each NodeClaim in a shared NodePool. The existing candidate filtering logic already considers the price of each NodeClaim individually, so there's no added complexity there.

PR #1458, opened 2024-07-23
chore: drop metrics in favour of workqueue metrics

jigisha620:
Fixes #N/A 

**Description**
1. Drop the `karpenter_disruption_queue_depth` and `karpenter_nodes_eviction_queue_depth` metrics in favor of the workqueue_depth metrics. 


pkg/controllers/disruption/orchestration/metrics.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 func init() {
```
jonathan-innis, line 26: There is more than just a single place where we create the workqueue. Should this go in the disruption orchestration package or should this be in the top-level metrics package? I believe this workqueue is also going to apply to the eviction queue that we create

jonathan-innis (commented):
Nice work!

pkg/controllers/disruption/orchestration/metrics.go
jigisha620: Agreed. Moved it to the top level metrics package.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -115,13 +115,18 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
...
-		providerIDToCommand:   map[string]*Command{},
-		kubeClient:            kubeClient,
-		recorder:              recorder,
-		cluster:               cluster,
-		clock:                 clock,
-		provisioner:           provisioner,
+		RateLimitingInterface: workqueue.NewRateLimitingQueueWithConfig(
+			workqueue.NewItemExponentialFailureRateLimiter(queueBaseDelay, queueMaxDelay),
+			workqueue.RateLimitingQueueConfig{
+				Name:            "disruptionQueue",
```
jonathan-innis, line 121:
```suggestion
				Name:            "disruption.queue",
```

I'd prefer to have a similar style to what the actual controller name is; though, since the controller itself is going to have a workqueue, we need to make sure that the name that we choose here doesn't conflict with the controller workqueue. Maybe that means renaming the controller name to something like `disruption.queue.reconciler` or naming the workqueue to something like `disruption.workqueue`?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -68,7 +68,7 @@ var (
 		prometheus.SummaryOpts{
 			Namespace:  "karpenter",
 			Subsystem:  metrics.PodSubsystem,
-			Name:       "startup_time_seconds",
+			Name:       "startup_duration_seconds",
```
jonathan-innis, line 71: I'm fine to couple these in the same PR since the change is small but it might be worthwhile to decouple these into separate PRs since the changes are unrelated

pkg/controllers/node/termination/suite_test.go
```diff
@@ -809,8 +808,6 @@ var _ = Describe("Termination", func() {
 			node = ExpectNodeExists(ctx, env.Client, node.Name)
 			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
 			ExpectSingletonReconciled(ctx, queue) // Reconcile the queue so that we set the metric
-
-			ExpectMetricGaugeValue(terminator.EvictionQueueDepth, 5, map[string]string{})
```
jonathan-innis, line 813: Is there a way for us to check the workqueue metric to get this eviction queue depth?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -90,10 +92,15 @@ type Queue struct {
...
 func NewQueue(kubeClient client.Client, recorder events.Recorder) *Queue {
 	queue := &Queue{
-		RateLimitingInterface: workqueue.NewRateLimitingQueue(workqueue.NewItemExponentialFailureRateLimiter(evictionQueueBaseDelay, evictionQueueMaxDelay)),
-		set:                   sets.New[QueueKey](),
-		kubeClient:            kubeClient,
-		recorder:              recorder,
+		RateLimitingInterface: workqueue.NewRateLimitingQueueWithConfig(
+			workqueue.NewItemExponentialFailureRateLimiter(evictionQueueBaseDelay, evictionQueueMaxDelay),
+			workqueue.RateLimitingQueueConfig{
+				Name:            "evictionQueue",
```
jonathan-innis, line 98: Same comment here about naming. Can we make a convention for both the queue and the reconciler so that we get consistent naming

pkg/metrics/metricsProvider.go
```diff
@@ -0,0 +1,113 @@
...
+	AddsKey                    = "adds_total"
+	QueueLatencyKey            = "queue_duration_seconds"
+	WorkDurationKey            = "work_duration_seconds"
+	UnfinishedWorkKey          = "unfinished_work_seconds"
+	LongestRunningProcessorKey = "longest_running_processor_seconds"
+	RetriesKey                 = "retries_total"
+)
+
+var (
+	depth = prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
jonathan-innis, line 36: nit: Since these are all related to workqueue, depth may not be specific enough. consider something like `workqueueDepth` for the variable name

pkg/metrics/metricsProvider.go
```diff
@@ -0,0 +1,113 @@
...
+	DepthKey                   = "depth"
+	AddsKey                    = "adds_total"
+	QueueLatencyKey            = "queue_duration_seconds"
+	WorkDurationKey            = "work_duration_seconds"
+	UnfinishedWorkKey          = "unfinished_work_seconds"
+	LongestRunningProcessorKey = "longest_running_processor_seconds"
+	RetriesKey                 = "retries_total"
+)
+
+var (
```
jonathan-innis, line 35: nit: consider putting the init for the metrics in the same file as the metrics themselves

pkg/metrics/metricsProvider.go
```diff
@@ -0,0 +1,113 @@
...
+*/
+
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/util/workqueue"
+)
+
+const (
```
jonathan-innis, line 24: Should we add a TODO in-code for when we can drop this?

pkg/metrics/metricsProvider.go
```diff
@@ -0,0 +1,113 @@
+/*
```
jonathan-innis, line 1: `/metrics/metricsProvider` stutters. Consider just `/metrics/workqueue_provider` or similar. Also, it's generally against go style to camel-case file names, snake case is standard.

jonathan-innis (approved):
/lgtm
/approve

PR #401, opened 2023-07-11
feat: Block readiness until caches are synced

ellistarn:
Fixes #N/A 

**Description**

Improve readiness checks to include cache hydration. Currently, readiness is blocked on webhooks, but which can become ready before the manager, giving us a false readiness signal.

**How was this change tested?**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1460, opened 2024-07-24
chore: remove readiness check on cache sync

jmdeal:
Fixes #N/A 

**Description**
Temporarily removes the readiness check on cache sync ahead of v1. This check was introduced by #401, but has an unexpected consequence for the upgrade path to v1. Consider the following scenario:
- A cluster is upgraded from v0.37.0 to a version with the v1 APIs / conversion webhooks
- The newly deployed Karpenter pods must list all resources before going ready
- This list call is going to attempt to invoke the conversion webhook
- Karpenter's endpoint is not ready since the pod is not ready, and the webhook can't be called through the service

In this scenario, Karpenter will be stuck in an unready state and be unable to progress.

manual validation

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -219,9 +219,6 @@ func NewOperator() (context.Context, *Operator) {
 		return []string{o.(*v1.NodePool).Spec.Template.Spec.NodeClassRef.Name}
 	}), "failed to setup nodepool nodeclassref name indexer")
 
-	lo.Must0(mgr.AddReadyzCheck("manager", func(req *http.Request) error {
-		return lo.Ternary(mgr.GetCache().WaitForCacheSync(req.Context()), nil, fmt.Errorf("failed to sync caches"))
-	}))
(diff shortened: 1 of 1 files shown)


pkg/operator/operator.go
```diff
@@ -219,9 +219,6 @@ func NewOperator() (context.Context, *Operator) {
 		return []string{o.(*v1.NodePool).Spec.Template.Spec.NodeClassRef.Name}
 	}), "failed to setup nodepool nodeclassref name indexer")
 
-	lo.Must0(mgr.AddReadyzCheck("manager", func(req *http.Request) error {
```
jonathan-innis, line 222: It's not clear to me what the tangible impact of this change is -- this was obviously added for a reason when it was first introduced and I'm cautious of removing it when we know we are just going to add it back in as soon as we drop the webhooks.

jmdeal: Synced offline

njtran (commented):
/lgtm
/approve

PR #1465, opened 2024-07-24
test: remove kind due to flakes

njtran:
Fixes #N/A 

**Description**
Removes kind e2es from presubmit so that we don't have flakes blocking PRs. Keeps it on push so that we can still have it run whenever a PR is merged.

Testing here


jonathan-innis (approved):
/lgtm
/approve

PR #1468, opened 2024-07-24
chore: rename disruption taint

jmdeal:
Fixes #N/A 

**Description**
Replaces the disruption taint as per https://github.com/kubernetes-sigs/karpenter/pull/1222.

`karpenter.sh/disruption=disrupting:NoSchedule` -> `karpenter.sh/disrupted:NoSchedule`

When the taint is removed by the orchestration queue or disruption controller, we'll also remove the v1beta1 equivalent version of the taint. This ensures we don't leave nodes tainted if Karpenter is upgraded after they've been disrupted but before they've been terminated.

`make presubmit` and validation against the AWS provider (TODO)

### DIFF
--- a/pkg/apis/v1/taints.go
+++ b/pkg/apis/v1/taints.go
@@ -20,22 +20,21 @@ import (
 	v1 "k8s.io/api/core/v1"
 
 	"sigs.k8s.io/karpenter/pkg/apis"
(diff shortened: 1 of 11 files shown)


jonathan-innis (approved):
/lgtm
/approve

PR #1469, opened 2024-07-24
feat: Drop Support for Log Config, Add Log Paths

rschalo:
Fixes #N/A 

**Description**
Dropping our global logging configuration was a planned deprecation at v1beta1 and we will continue by fully dropping support for the ConfigMap-based configuration for our logging at v1.

make test


pkg/operator/logging/logging.go
```diff
@@ -51,6 +43,14 @@ func DefaultZapConfig(ctx context.Context, component string) zap.Config {
 		// Webhooks are deprecated, so support for changing their log level is also deprecated
 		logLevel = lo.Must(zap.ParseAtomicLevel(l))
 	}
+	outputPaths := []string{"stdout"}
```
jonathan-innis, line 46: Rather than the this being a default here, should we just rely on the options and their defaults?

jonathan-innis (approved):
/lgtm
/approve

PR #1472, opened 2024-07-25
feat: Remove karpenter.sh/managed-by var

rschalo:
Fixes #N/A 

**Description**
Dropping `karpenter.sh/managed-by` as it is an artifact of a migration path and not something needed in the future. Adding `eks:eks-cluster-name` to support pod identity sessions.

make test

### DIFF
--- a/pkg/apis/v1beta1/labels.go
+++ b/pkg/apis/v1beta1/labels.go
@@ -46,7 +46,6 @@ const (
 const (
 	DoNotDisruptAnnotationKey             = apis.Group + "/do-not-disrupt"
 	ProviderCompatibilityAnnotationKey    = apis.CompatibilityGroup + "/provider"
-	ManagedByAnnotationKey                = apis.Group + "/managed-by"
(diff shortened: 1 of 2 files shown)


pkg/apis/v1beta1/labels.go
```diff
@@ -46,7 +46,6 @@ const (
 const (
 	DoNotDisruptAnnotationKey             = apis.Group + "/do-not-disrupt"
 	ProviderCompatibilityAnnotationKey    = apis.CompatibilityGroup + "/provider"
-	ManagedByAnnotationKey                = apis.Group + "/managed-by"
```
engedaam, line 49: nit. 

Do we need to drop it from the v1beta1? I think it makes sense to leave it there, since that will be a breaking change

rschalo: What part does this break? The garbage collection is now being done via `NodePoolLabelKey` which i present on both v1beta1 and v1.

jonathan-innis (approved):
/lgtm
/approve

PR #1473, opened 2024-07-28
docs: Change example NodePool in kwok readme from v1beta to v1.

payall4u:
Fixes #N/A 

**Description**
When run the kwoks examples, I found that I had to configure the webhook to convert versions. So change example NodePool in kwok readme from v1beta to v1 to run the example directly.

Only docs.

### DIFF
--- a/kwok/README.md
+++ b/kwok/README.md
@@ -20,7 +20,7 @@ Once kwok is installed and Karpenter successfully applies to the cluster, you sh
 
 ```bash
 cat <<EOF | envsubst | kubectl apply -f -
-apiVersion: karpenter.sh/v1beta1
+apiVersion: karpenter.sh/v1
 kind: NodePool
 metadata:
   name: default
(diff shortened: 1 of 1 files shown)


kwok/README.md
```diff
@@ -20,7 +20,7 @@ Once kwok is installed and Karpenter successfully applies to the cluster, you sh
 
 ```bash
 cat <<EOF | envsubst | kubectl apply -f -
-apiVersion: karpenter.sh/v1beta1
+apiVersion: karpenter.sh/v1
```
engedaam, line 23: `expireAfter` will also need to be moved to template field

payall4u: My fault. Fixed, and I also checked other fields.

jonathan-innis: Thanks for the change. Looks good!

jonathan-innis (approved):
/lgtm
/approve

PR #1499, opened 2024-08-01
chore: Ensure that NodeClaims are held in Unknown until they reach a terminal state

jonathan-innis:
Fixes #N/A 

**Description**

This PR updates our handling of NodeClaims to leave them in an Unknown state for Launched, Registered, and Initialized status conditions until they have reached a terminal state. This aligns with the thinking around terminal status conditions that's detailed in: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties.

This now means that Karpenter will hold NodeClaims in an Unknown state (also Ready Unknown state) for a longer period of time before it either transitions them into a False or True state. Once it enters this False or True state, it will not leave its False or True state. 

`make presubmit`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ go 1.22.5
 require (
 	github.com/Pallinder/go-randomdata v1.2.0
 	github.com/avast/retry-go v3.0.0+incompatible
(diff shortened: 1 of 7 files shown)


njtran (commented):
/lgtm
/approve

PR #1516, opened 2024-08-06
chore: Add pattern check for group in nodeClassRef

jigisha620:
Fixes #N/A 

**Description**
If a user creates a nodePool with nodeClassRef as shown below, nodePool fails to watch the nodeClass and nodePool remains stuck in not ready condition if EC2NodeClass is created before nodePool.
```
nodeClassRef:
  group: karpenter.k8s.aws/v1
  kind: EC2NodeClass
  name: default
```
A check has been added for the field `group` such that it will not allow slash.

Added test for the change. Tested on local cluster.

### DIFF
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -86,6 +86,7 @@ spec:
                   properties:
                     group:
                       description: API version of the referent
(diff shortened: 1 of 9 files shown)


jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

PR #1520, opened 2024-08-07
cleanup: make the import order follow the best practice

jwcesign:
Fixes #N/A 

**Description**

make the import order follow the best practice.
We should use [gci](https://golangci-lint.run/usage/linters/#gci) to do lint checking

Nothing about logic changes. So we don't need to test.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -59,6 +59,12 @@ linters-settings:
       WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
       See the License for the specific language governing permissions and
       limitations under the License.
+  gci:
+    sections:
+      - standard # Standard section: captures all standard packages.
+      - default # Default section: contains all imports that could not be matched to another section type.
+      - localmodule # Local module section: contains all local packages. This section is not present unless explicitly enabled.
+    skip-generated: true
 issues:
   fix: true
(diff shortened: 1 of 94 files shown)


liangyuanpeng (commented):
cool! Maybe you want to update https://github.com/kubernetes-sigs/karpenter/blob/main/.golangci.yaml?

liangyuanpeng (commented):
LGTM

.golangci.yaml
```diff
@@ -59,6 +60,12 @@ linters-settings:
       WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
       See the License for the specific language governing permissions and
       limitations under the License.
+  gci:
+    sections:
+      - standard # Standard section: captures all standard packages.
+      - default # Default section: contains all imports that could not be matched to another section type.
```
jonathan-innis, line 66: Should we add "dot" here as a section since we have a number of dot imports that we should probably group together throughout the testing code?

jwcesign: cc @jonathan-innis @ellistarn

jwcesign: cc @jonathan-innis @ellistarn @njtran 
Please take a look, thanks

jwcesign: cc @jonathan-innis

PR #1524, opened 2024-08-07
ci: correct to create kubernetes with helm/kind-action.

liangyuanpeng:
Fixes #N/A 

**Description**

currently, create kubernetes cluster with helm/kind-action by default,  it's always using create the kubernetes v1.29.2, see https://github.com/kubernetes-sigs/karpenter/actions/runs/10275469972/job/28434158108


Makefile
```diff
@@ -43,11 +43,11 @@ apply-with-kind: verify build-with-kind ## Deploy the kwok controller from the c
 e2etests: ## Run the e2e suite against your local cluster
 	cd test && go test \
 		-count 1 \
-		-timeout 30m \
+		-timeout 120m \
```
liangyuanpeng, line 46: it's testing

.github/workflows/kind-e2e.yaml
```diff
@@ -1,7 +1,7 @@
 name: kind-e2e
 on:
   push:
-    branches: [main]
+  pull_request:
```
liangyuanpeng, line 4: for test

PR #1534, opened 2024-08-07
chore: fix nodeClassRef conversion

jigisha620:
Fixes #N/A 

**Description**
When converting nodeClassRef from v1beta1 to v1, group would be empty if apiVersion is of the type `karpenter.k8s.aws` i.e. it does not contain the version. This PR adds the fix for it.

Tested on local cluster

### DIFF
--- a/pkg/apis/v1/nodeclaim_conversion.go
+++ b/pkg/apis/v1/nodeclaim_conversion.go
@@ -148,7 +148,7 @@ func (in *NodeClaimSpec) convertFrom(ctx context.Context, v1beta1nc *v1beta1.Nod
 	in.NodeClassRef = &NodeClassReference{
 		Name:  v1beta1nc.NodeClassRef.Name,
 		Kind:  lo.Ternary(v1beta1nc.NodeClassRef.Kind == "", defaultNodeClassGVK.Kind, v1beta1nc.NodeClassRef.Kind),
-		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, nodeclassGroupVersion.Group),
+		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, lo.Ternary(nodeclassGroupVersion.Group == "", nodeclassGroupVersion.Version, nodeclassGroupVersion.Group)),
 	}
 
 	if v1beta1nc.Kubelet != nil {
--- a/pkg/apis/v1/nodepool_conversion.go
+++ b/pkg/apis/v1/nodepool_conversion.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/samber/lo"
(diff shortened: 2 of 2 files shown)


pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -139,16 +140,11 @@ func (in *NodeClaimSpec) convertFrom(ctx context.Context, v1beta1nc *v1beta1.Nod
...
-	if err != nil {
-		return "", err
-	}
+	nodeclasses := injection.GetNodeClasses(ctx)
 	in.NodeClassRef = &NodeClassReference{
 		Name:  v1beta1nc.NodeClassRef.Name,
-		Kind:  lo.Ternary(v1beta1nc.NodeClassRef.Kind == "", defaultNodeClassGVK.Kind, v1beta1nc.NodeClassRef.Kind),
-		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, nodeclassGroupVersion.Group),
+		Kind:  lo.Ternary(v1beta1nc.NodeClassRef.Kind == "", nodeclasses[0].Kind, v1beta1nc.NodeClassRef.Kind),
+		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", nodeclasses[0].Group, strings.Split(v1beta1nc.NodeClassRef.APIVersion, "/")[0]),
```
njtran, line 147: can't this just be the following? Looking at ParseGroupVersion, it looks it stores the correct information, but assumes that if there's no /, it treats the string as the version, not the group. 
```suggestion
		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, fmt.Sprintf("%s/%s", defaultNodeClassGVK.Group.Version, nodeclassGroupVersion.Version),
```

njtran: This way, we keep the upstream package that has more parsing logic for us, rather than implementing it ourselves

jigisha620: Yeah, makes sense. I believe this is what you meant - 
```
Group: lo.Ternary(v1beta1np.Spec.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, lo.Ternary(nodeclassGroupVersion.Group == "", nodeclassGroupVersion.Version, nodeclassGroupVersion.Group))
```

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -191,7 +183,7 @@ func (in *NodeClaimTemplate) convertFrom(ctx context.Context, v1beta1np *v1beta1
 	in.Spec.NodeClassRef = &NodeClassReference{
 		Name:  v1beta1np.Spec.NodeClassRef.Name,
 		Kind:  lo.Ternary(v1beta1np.Spec.NodeClassRef.Kind == "", defaultNodeClassGVK.Kind, v1beta1np.Spec.NodeClassRef.Kind),
-		Group: lo.Ternary(v1beta1np.Spec.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, nodeclassGroupVersion.Group),
+		Group: lo.Ternary(v1beta1np.Spec.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, lo.Ternary(nodeclassGroupVersion.Group == "", nodeclassGroupVersion.Version, nodeclassGroupVersion.Group)),
```
njtran, line 186: wouldn't this leave out the `/v1` if what it was converting from was `karpenter.k8s.aws`?

njtran (commented):
/lgtm
/approve

PR #1544, opened 2024-08-08
chore: Retain EC2NodeClass reference information during the conversion

jonathan-innis:
Fixes #N/A 

**Description**

Ensure that our EC2NodeClassReference information is retained during conversion from v1beta1 to v1 and back. This is to ensure that we don't drift when Karpenter starts reasoning in terms of the v1 version.

### DIFF
--- a/pkg/apis/v1/nodeclaim_conversion.go
+++ b/pkg/apis/v1/nodeclaim_conversion.go
@@ -20,11 +20,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 
 	"github.com/samber/lo"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/pkg/apis"
 
@@ -37,11 +37,17 @@ func (in *NodeClaim) ConvertTo(ctx context.Context, to apis.Convertible) error {
 	v1beta1NC := to.(*v1beta1.NodeClaim)
 	v1beta1NC.ObjectMeta = in.ObjectMeta
 
-	in.Status.convertTo((&v1beta1NC.Status))
-	return in.Spec.convertTo(ctx, &v1beta1NC.Spec, in.Annotations[KubeletCompatibilityAnnotationKey])
+	in.Status.convertTo(&v1beta1NC.Status)
+	if err := in.Spec.convertTo(&v1beta1NC.Spec, in.Annotations[KubeletCompatibilityAnnotationKey], in.Annotations[NodeClassReferenceAnnotationKey]); err != nil {
+		return err
+	}
+	// Remove the annotations from the v1beta1 NodeClaim on the convert back
+	delete(v1beta1NC.Annotations, KubeletCompatibilityAnnotationKey)
+	delete(v1beta1NC.Annotations, NodeClassReferenceAnnotationKey)
+	return nil
 }
 
-func (in *NodeClaimSpec) convertTo(ctx context.Context, v1beta1nc *v1beta1.NodeClaimSpec, kubeletAnnotation string) error {
+func (in *NodeClaimSpec) convertTo(v1beta1nc *v1beta1.NodeClaimSpec, kubeletAnnotation, nodeClassReferenceAnnotation string) error {
 	v1beta1nc.Taints = in.Taints
 	v1beta1nc.StartupTaints = in.StartupTaints
 	v1beta1nc.Resources = v1beta1.ResourceRequirements(in.Resources)
@@ -55,18 +61,16 @@ func (in *NodeClaimSpec) convertTo(ctx context.Context, v1beta1nc *v1beta1.NodeC
 			MinValues: v1Requirements.MinValues,
 		}
 	})
-
-	if in.NodeClassRef != nil {
-		nodeclass, found := lo.Find(injection.GetNodeClasses(ctx), func(nc schema.GroupVersionKind) bool {
-			return nc.Kind == in.NodeClassRef.Kind && nc.Group == in.NodeClassRef.Group
-		})
-		v1beta1nc.NodeClassRef = &v1beta1.NodeClassReference{
-			Kind:       in.NodeClassRef.Kind,
-			Name:       in.NodeClassRef.Name,
-			APIVersion: lo.Ternary(found, nodeclass.GroupVersion().String(), ""),
+	// Convert the NodeClassReference depending on whether the annotation exists
+	v1beta1nc.NodeClassRef = &v1beta1.NodeClassReference{}
+	if nodeClassReferenceAnnotation != "" {
+		if err := json.Unmarshal([]byte(nodeClassReferenceAnnotation), v1beta1nc.NodeClassRef); err != nil {
+			return fmt.Errorf("unmarshaling nodeClassRef annotation, %w", err)
 		}
+	} else {
+		v1beta1nc.NodeClassRef.Name = in.NodeClassRef.Name
+		v1beta1nc.NodeClassRef.Kind = in.NodeClassRef.Kind
 	}
-
 	if kubeletAnnotation != "" {
 		v1beta1kubelet := &v1beta1.KubeletConfiguration{}
 		err := json.Unmarshal([]byte(kubeletAnnotation), v1beta1kubelet)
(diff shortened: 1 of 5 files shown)


pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -37,11 +37,17 @@ func (in *NodeClaim) ConvertTo(ctx context.Context, to apis.Convertible) error {
...
 	v1beta1NC.ObjectMeta = in.ObjectMeta
 
-	in.Status.convertTo((&v1beta1NC.Status))
-	return in.Spec.convertTo(ctx, &v1beta1NC.Spec, in.Annotations[KubeletCompatibilityAnnotationKey])
+	in.Status.convertTo(&v1beta1NC.Status)
+	if err := in.Spec.convertTo(&v1beta1NC.Spec, in.Annotations[KubeletCompatibilityAnnotationKey], in.Annotations[NodeClassReferenceAnnotationKey]); err != nil {
+		return err
+	}
+	// Remove the annotations from the v1beta1 NodeClaim on the convert back
+	delete(v1beta1NC.Annotations, KubeletCompatibilityAnnotationKey)
```
njtran, line 45: Was this not correct before then?

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -55,18 +61,15 @@ func (in *NodeClaimSpec) convertTo(ctx context.Context, v1beta1nc *v1beta1.NodeC
...
-			Name:       in.NodeClassRef.Name,
-			APIVersion: lo.Ternary(found, nodeclass.GroupVersion().String(), ""),
+	// Convert the NodeClassReference depending on whether the annotation exists
+	v1beta1nc.NodeClassRef = &v1beta1.NodeClassReference{}
+	if nodeClassReferenceAnnotation != "" {
+		if err := json.Unmarshal([]byte(nodeClassReferenceAnnotation), v1beta1nc.NodeClassRef); err != nil {
+			return fmt.Errorf("unmarshaling nodeClassRef annotation, %w", err)
 		}
+	} else {
+		v1beta1nc.NodeClassRef.Name = in.NodeClassRef.Name
```
njtran, line 71: Thought we were also going to add kind?

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -141,14 +151,10 @@ func (in *NodeClaimSpec) convertFrom(ctx context.Context, v1beta1nc *v1beta1.Nod
...
 	defaultNodeClassGVK := injection.GetNodeClasses(ctx)[0]
-	nodeclassGroupVersion, err := schema.ParseGroupVersion(v1beta1nc.NodeClassRef.APIVersion)
-	if err != nil {
-		return "", err
-	}
 	in.NodeClassRef = &NodeClassReference{
 		Name:  v1beta1nc.NodeClassRef.Name,
 		Kind:  lo.Ternary(v1beta1nc.NodeClassRef.Kind == "", defaultNodeClassGVK.Kind, v1beta1nc.NodeClassRef.Kind),
-		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, lo.Ternary(nodeclassGroupVersion.Group == "", nodeclassGroupVersion.Version, nodeclassGroupVersion.Group)),
+		Group: lo.Ternary(v1beta1nc.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, strings.Split(v1beta1nc.NodeClassRef.APIVersion, "/")[0]),
```
njtran, line 157: I think I had the same question for Jigisha in her PR, why would we remove the upstream function for this and implement it ourselves? Seems like it's working as expected as is, no?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -40,17 +39,23 @@ func (in *NodePool) ConvertTo(ctx context.Context, to apis.Convertible) error {
 	// Convert v1 status
 	v1beta1NP.Status.Resources = in.Status.Resources
 	v1beta1NP.Status.Conditions = in.Status.Conditions
-	return in.Spec.convertTo(ctx, &v1beta1NP.Spec, in.Annotations[KubeletCompatibilityAnnotationKey])
+	if err := in.Spec.convertTo(&v1beta1NP.Spec, in.Annotations[KubeletCompatibilityAnnotationKey], in.Annotations[NodeClassReferenceAnnotationKey]); err != nil {
+		return err
+	}
+	// Remove the annotations from the v1beta1 NodeClaim on the convert back
+	delete(v1beta1NP.Annotations, KubeletCompatibilityAnnotationKey)
```
njtran, line 46: same question on if this was correct before or not

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -82,28 +87,15 @@ func (in *NodeClaimTemplate) convertTo(ctx context.Context, v1beta1np *v1beta1.N
...
-	})
-	matchingNodeClass, found := lo.Find(nodeClasses, func(nc schema.GroupVersionKind) bool {
-		return nc.Kind == in.Spec.NodeClassRef.Kind && nc.Group == in.Spec.NodeClassRef.Group
-	})
-	v1beta1np.Spec.NodeClassRef = &v1beta1.NodeClassReference{
-		Kind:       in.Spec.NodeClassRef.Kind,
-		Name:       in.Spec.NodeClassRef.Name,
-		APIVersion: lo.Ternary(found, matchingNodeClass.GroupVersion().String(), ""),
+	} else {
+		v1beta1np.Spec.NodeClassRef.Name = in.Spec.NodeClassRef.Name
```
njtran, line 97: same here on doing kind?

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -176,23 +175,17 @@ func (in *NodeClaimTemplate) convertFrom(ctx context.Context, v1beta1np *v1beta1
...
 	defaultNodeClassGVK := injection.GetNodeClasses(ctx)[0]
-	nodeclassGroupVersion, err := schema.ParseGroupVersion(v1beta1np.Spec.NodeClassRef.APIVersion)
-	if err != nil {
-		return "", err
-	}
 	in.Spec.NodeClassRef = &NodeClassReference{
 		Name:  v1beta1np.Spec.NodeClassRef.Name,
 		Kind:  lo.Ternary(v1beta1np.Spec.NodeClassRef.Kind == "", defaultNodeClassGVK.Kind, v1beta1np.Spec.NodeClassRef.Kind),
-		Group: lo.Ternary(v1beta1np.Spec.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, lo.Ternary(nodeclassGroupVersion.Group == "", nodeclassGroupVersion.Version, nodeclassGroupVersion.Group)),
+		Group: lo.Ternary(v1beta1np.Spec.NodeClassRef.APIVersion == "", defaultNodeClassGVK.Group, strings.Split(v1beta1np.Spec.NodeClassRef.APIVersion, "/")[0]),
```
njtran, line 181: same here

pkg/apis/v1/nodeclaim_conversion.go
jonathan-innis: Correct -- we had a brief discussion and realized that keeping the annotations around here is not the correct behavior.

engedaam: The compatibility annotation was there for observability, since this an annotation that is called out in the docs that customer will need drop when going to v1.1 or rolling back.

pkg/apis/v1/nodeclaim_conversion.go
jonathan-innis: You did and I think the upstream function is more confusing than this is. The string split is the most straightforward way to do this IMO. I had a lot of questions looking at the current implementation, wondering why we are setting it to version in some cases and not others

pkg/apis/v1/nodepool_conversion.go
jonathan-innis: Same response, discussed and agreed that this is correct

pkg/apis/v1/nodeclaim_conversion.go
njtran: Does our CEL validation cover the cases that the function covers? e.g. two `/` is invalid

jonathan-innis: It's a good callout. We could/should probably add validation for this

njtran (commented):
/lgtm
/approve

PR #1545, opened 2024-08-08
fix: ensure NillableDuration is round-trippable

jmdeal:
Fixes #N/A 

**Description**
Ensure our NillableDuration type is round-trippable for conversion webhooks.


pkg/apis/v1/duration.go
```diff
@@ -50,16 +56,14 @@ func (d *NillableDuration) UnmarshalJSON(b []byte) error {
...
 func (d NillableDuration) MarshalJSON() ([]byte, error) {
-	if d.Duration == nil {
-		return json.Marshal(Never)
-	}
-	return json.Marshal(d.Duration.String())
+	return json.Marshal(d.raw)
+	// if d.Duration == nil {
+	// 	return json.Marshal(Never)
+	// }
+	// return json.Marshal(d.Duration.String())
```
rschalo, line 63: Remove?

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -269,6 +270,24 @@ var _ = Describe("Convert V1 to V1beta1 NodePool API", func() {
 			Expect(v1beta1nodepool.Status.Resources[resource]).To(Equal(v1nodepool.Status.Resources[resource]))
 		}
 	})
+	Context("Round Trip", func() {
+		DescribeTable(
+			"NillableDuration",
+			func(value string, field func() *NillableDuration) {
+				str := fmt.Sprintf("%q", value)
```
njtran, line 277: curious why this needs to be quotes? does it do anything differently?
```suggestion
				str := fmt.Sprintf("%s", value)
```

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -269,6 +270,24 @@ var _ = Describe("Convert V1 to V1beta1 NodePool API", func() {
...
+				duration := NillableDuration{}
+				Expect(json.Unmarshal([]byte(str), &duration)).Should(Succeed())
+				*field() = duration
```
njtran, line 280: I don't quite understand what this is doing, is this setting the output of the return value of the func to duration?

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -537,4 +556,27 @@ var _ = Describe("Convert V1beta1 to V1 NodePool API", func() {
...
+				Expect(v1nodepool.ConvertFrom(ctx, v1beta1nodepool)).To(Succeed())
+				Expect(v1nodepool.ConvertTo(ctx, v1beta1nodepool)).To(Succeed())
+				result, err := json.Marshal(*field())
+				Expect(err).To(BeNil())
+				Expect(string(result)).To(Equal(str))
+			},
+			Entry("spec.template.spec.expireAfter", "720h", func() *v1beta1.NillableDuration { return &v1beta1nodepool.Spec.Disruption.ExpireAfter }),
+			Entry("spec.disruption.consolidateAfter", "15m", func() *v1beta1.NillableDuration {
+				if v1beta1nodepool.Spec.Disruption.ConsolidateAfter == nil {
```
njtran, line 575: What about for WhenEmpty?

pkg/apis/v1/nodepool_conversion_test.go
jmdeal: If you don't quote it, it will fail to parse, I believe because it's trying to parse it as a number rather than a string literal.

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -269,6 +270,24 @@ var _ = Describe("Convert V1 to V1beta1 NodePool API", func() {
 			Expect(v1beta1nodepool.Status.Resources[resource]).To(Equal(v1nodepool.Status.Resources[resource]))
 		}
 	})
+	Context("Round Trip", func() {
+		DescribeTable(
+			"NillableDuration",
+			func(value string, field func() *NillableDuration) {
```
jonathan-innis, line 276: Why do you need field to be a function that returns a NillableDuration rather than the pointer itself? I think this indirection might be creating some confusion here with reading through it. Also, considering the indirection here -- you might be better off just making these separate It() statements rather than a DescribeTable()

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -269,6 +270,24 @@ var _ = Describe("Convert V1 to V1beta1 NodePool API", func() {
...
+				str := fmt.Sprintf("%q", value)
+				duration := NillableDuration{}
+				Expect(json.Unmarshal([]byte(str), &duration)).Should(Succeed())
+				*field() = duration
+				Expect(v1nodepool.ConvertTo(ctx, v1beta1nodepool)).To(Succeed())
```
jonathan-innis, line 281: Just validating: This test failed prior to this change?

jonathan-innis: Also, any reason to not json marshal the entire object rather than just the specific field that we care about during the test? Seems like a stronger bar to hold if we do that

pkg/apis/v1/nodepool_conversion_test.go
jmdeal: Ya, I don't think the indirection is worth it here. I'm going to separate it out.

pkg/apis/v1/nodepool_conversion_test.go
jmdeal: Yep, this would have failed before. I was only marshaling this as json to ensure we populated the "raw" value correctly before doing the conversions, I've updated every instance of us creating a NillableDuration with a constructor that covers this instead.

pkg/apis/v1/nodepool_conversion_test.go
jmdeal: This didn't have to do with the consolidation type, it was just a hack since I was mutating the field through a pointer and needed to ensure it was non-nil. I've refactored the test to remove this.

pkg/apis/v1/nodepool_conversion_test.go
jmdeal: Removed this indirection, more confusing than it was worth. This was just a way to mutate an arbitrary field in the NodePool, specified by the Entry in the table.

pkg/apis/v1/duration.go
```diff
@@ -28,6 +32,16 @@ const Never = "Never"
...
 	*time.Duration
+
+	// Raw is used to ensure we remarshal the NillableDuration in the same format it was specified.
+	// This ensures tools like Flux and ArgoCD don't mistakenly detect drift due to our conversion webhooks.
+	Raw []byte `hash:"ignore"`
+}
+
+func MustParseNillableDuration(val string) NillableDuration {
+	nd := NillableDuration{}
+	lo.Must0(json.Unmarshal([]byte(fmt.Sprintf("%q", val)), &nd))
```
njtran, line 43:
```suggestion
        // Use %q instead of %s to ensure that we unmarshal the value as a string and not an int
	lo.Must0(json.Unmarshal([]byte(fmt.Sprintf("%q", val)), &nd))
```

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -269,6 +269,24 @@ var _ = Describe("Convert V1 to V1beta1 NodePool API", func() {
 			Expect(v1beta1nodepool.Status.Resources[resource]).To(Equal(v1nodepool.Status.Resources[resource]))
 		}
 	})
+	Context("Round Trip", func() {
```
njtran, line 272: Can we do some testing here with the `nil (Never)` value? This should be more important for ConsolidateAfter, since it's going from a pointer to a non-pointer.

njtran (commented):
two minor comments

pkg/apis/v1/nodepool_conversion_test.go
jmdeal: I don't think it makes a difference for ConsolidateAfter since the wrapped duration is a pointer regardless, but I can go ahead and add a test. I don't think it's something we explicitly need to test since it only has one valid representation ("Never") but it doesn't hurt to add.

pkg/apis/v1/nodepool_conversion_test.go
```diff
@@ -269,6 +269,40 @@ var _ = Describe("Convert V1 to V1beta1 NodePool API", func() {
...
+		})
+		It("spec.disruption.consolidateAfter", func() {
+			v1nodepool.Spec.Disruption.ConsolidateAfter = MustParseNillableDuration("10h")
+			Expect(v1nodepool.ConvertTo(ctx, v1beta1nodepool)).To(Succeed())
+			Expect(v1nodepool.ConvertFrom(ctx, v1beta1nodepool)).To(Succeed())
+			result, err := json.Marshal(v1nodepool.Spec.Disruption.ConsolidateAfter)
+			Expect(err).To(BeNil())
+			Expect(string(result)).To(Equal(`"10h"`))
+		})
+		It("spec.disruption.consolidateAfter", func() {
```
njtran, line 297:
```suggestion
		It("spec.disruption.consolidateAfter (Never)", func() {
```

njtran (commented):
one comment but LGTM

njtran (commented):
/lgtm
/approve

PR #1577, opened 2024-08-15
feat: add in cloudprovider disruption methods

njtran:
Fixes #N/A 

**Description**
Adds in cloudprovider disruption methods - Will be taken over by @engedaam 

TODO: add testing for cloudprovider methods. Either requires removing the CRD enum, or adding in the enum for each cloud provider installation, and ingesting a new CRD for the test environment

### DIFF
--- a/pkg/controllers/disruption/cloudprovider.go
+++ b/pkg/controllers/disruption/cloudprovider.go
@@ -1,0 +1,127 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package disruption
+
+import (
+	"context"
+	"errors"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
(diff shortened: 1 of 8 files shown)


pkg/controllers/disruption/cloudprovider.go
```diff
@@ -0,0 +1,127 @@
...
+
+		return Command{
+			candidates:   []*Candidate{candidate},
+			replacements: results.NewNodeClaims,
+		}, results, nil
+	}
+	return Command{}, scheduling.Results{}, nil
+}
+
+func (cp *CloudProvider) Reason() v1.DisruptionReason {
```
jonathan-innis, line 117: Why not just use the actual reason that was passed by the CloudProvider?

pkg/controllers/disruption/cloudprovider.go
```diff
@@ -0,0 +1,127 @@
...
+	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
+	"sigs.k8s.io/karpenter/pkg/controllers/provisioning"
+	"sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling"
+	"sigs.k8s.io/karpenter/pkg/controllers/state"
+	"sigs.k8s.io/karpenter/pkg/events"
+)
+
+// CloudProvider is a subreconciler that deletes candidates according to cloud provider specific reasons.
+// This should be methods that are
+type CloudProvider struct {
```
jonathan-innis, line 36: I'm not convinced that we need a separate method definition here. Can't we just combine this with the drift implementation so that drift is just another one of these "reasons" that we pull from the CloudProvider. I'm asking because this looks exactly the same as the drift implementation at this point.

pkg/controllers/disruption/controller.go
```diff
@@ -79,6 +79,8 @@ func NewController(clk clock.Clock, kubeClient client.Client, provisioner *provi
 		cloudProvider: cp,
 		lastRun:       map[string]time.Time{},
 		methods: []Method{
+			// Terminate any NodeClaims with StatusConditions specific to the CloudProvider implementation. For
+			NewCloudProvider(kubeClient, cluster, c.cloudProvider, recorder, provisioner),
```
jonathan-innis, line 83: What about appending `NewStatusConditionDisruption()` with new methods gen-ed based on a combination of the CloudProvider reason and the Drift reason?

jonathan-innis (commented):
I feel like this is a good time to do a code simplification on the way that we pull disruption reasons from CloudProviders with the way that we evaluate Drift

PR #1588, opened 2024-08-21
feat: add cloudprovider specific Eventual disruption methods 

engedaam:
Fixes #N/A 

**Description**
- Add eventual  disruption methods that will consist of drift and cloud provider supplied disruption 
- The objective is to allow a cloudproviders to update the status on a nodeclaim with a disruptionReasons and the disruption controller to orchestrate the termination and replacement of the nodeclaim 
- User will be able to block on the disruption type provided by the cloudprovider in their disruption controls on the nodepool.

- make presubmit


pkg/apis/v1/nodepool.go
```diff
@@ -129,6 +130,7 @@ const (
 )
 
 // DisruptionReason defines valid reasons for disruption budgets.
+// CloudProviders will need to append to the list of enums when implementing cloud provider disruption reasons
 // +kubebuilder:validation:Enum={Underutilized,Empty,Drifted}
```
engedaam, line 134: I'm not sure what the best option would be here for disruption reason validation. In my mind I see we have two options: 
1. Leave the OpenAPI validation and require cloudprovider to expand the disruption reason as part of the CRD validation
2. Remove the OpenAPI validation and use Runtime validation for the nodepool

I'm leaning towards option 1, since we have precedent of requiring cloud-providers to add additional validation for labels.

pkg/apis/v1/nodepool.go
```diff
@@ -326,7 +332,7 @@ func (in *NodePool) MustGetAllowedDisruptions(ctx context.Context, c clock.Clock
 func (in *NodePool) GetAllowedDisruptionsByReason(ctx context.Context, c clock.Clock, numNodes int) (map[DisruptionReason]int, error) {
 	var multiErr error
 	allowedDisruptions := map[DisruptionReason]int{}
-	for _, reason := range WellKnownDisruptionReasons {
+	for reason := range WellKnownDisruptionReasons {
```
jonathan-innis, line 335: I wonder if we even need a definition of WellKnownDisruptionReasons? Is it possible to say that if there are no constraints that are returned by this function that it's just the default? e.g. Unlimited?

pkg/controllers/disruption/eventual.go
```diff
@@ -32,32 +32,34 @@ import (
...
-	kubeClient  client.Client
-	cluster     *state.Cluster
-	provisioner *provisioning.Provisioner
-	recorder    events.Recorder
+type EventualDisruption struct {
+	kubeClient       client.Client
+	cluster          *state.Cluster
+	provisioner      *provisioning.Provisioner
+	recorder         events.Recorder
+	disruptionReason v1.DisruptionReason
```
jonathan-innis, line 40:
```suggestion
	reason v1.DisruptionReason
```

pkg/controllers/disruption/eventual.go
```diff
@@ -32,32 +32,34 @@ import (
...
+		recorder:         recorder,
+		disruptionReason: disruptionReason,
 	}
 }
 
 // ShouldDisrupt is a predicate used to filter candidates
-func (d *Drift) ShouldDisrupt(ctx context.Context, c *Candidate) bool {
-	return c.NodeClaim.StatusConditions().Get(v1.ConditionTypeDrifted).IsTrue()
+func (ed *EventualDisruption) ShouldDisrupt(ctx context.Context, c *Candidate) bool {
+	return c.NodeClaim.StatusConditions().Get(string(ed.disruptionReason)).IsTrue()
```
jonathan-innis, line 55:
```suggestion
	return c.NodeClaim.StatusConditions().Get(string(ed.DisruptionReason())).IsTrue()
```

pkg/apis/v1/nodepool.go
engedaam: I just made this as a passed in variable, so that we can define it as needed

pkg/controllers/disruption/controller.go
```diff
@@ -78,16 +86,21 @@ func NewController(clk clock.Clock, kubeClient client.Client, provisioner *provi
...
-			// Terminate any NodeClaims that have drifted from provisioning specifications, allowing the pods to reschedule.
-			NewDrift(kubeClient, cluster, provisioner, recorder),
-			// Delete any empty NodeClaims as there is zero cost in terms of disruption.
-			NewEmptiness(c),
-			// Attempt to identify multiple NodeClaims that we can consolidate simultaneously to reduce pod churn
-			NewMultiNodeConsolidation(c),
-			// And finally fall back our single NodeClaim consolidation to further reduce cluster cost.
-			NewSingleNodeConsolidation(c),
-		},
+		wellKnownDisruptionReason: append([]v1.DisruptionReason{
```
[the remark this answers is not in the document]
engedaam: Yeah I agree, this was the best way I could think to add well known disruption reason, without asking cloudprovider to inject it in to the list. This was in the hope to just ask cloudprovider to only add disruption reason as part of the cloudprovider interface method and no where else

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -69,6 +70,7 @@ func NewCloudProvider() *CloudProvider {
 		CreatedNodeClaims:        map[string]*v1.NodeClaim{},
 		InstanceTypesForNodePool: map[string][]*cloudprovider.InstanceType{},
 		ErrorsForNodePool:        map[string]error{},
+		disruptionReasons:        []v1.DisruptionReason{"CloudProviderDisruptionReason"},
```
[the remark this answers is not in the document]
engedaam: Seems reasonable to me

pkg/apis/v1/nodepool.go
```diff
@@ -314,34 +309,28 @@ func (nl *NodePoolList) OrderByWeight() {
...
-
 	for _, budget := range in.Spec.Disruption.Budgets {
 		val, err := budget.GetAllowedDisruptions(c, numNodes)
 		if err != nil {
 			multiErr = multierr.Append(multiErr, err)
 		}
-		// If reasons is nil, it applies to all well known disruption reasons
-		for _, reason := range lo.Ternary(budget.Reasons == nil, WellKnownDisruptionReasons, budget.Reasons) {
-			allowedDisruptions[reason] = lo.Min([]int{allowedDisruptions[reason], val})
+		if budget.Reasons == nil || lo.Contains(budget.Reasons, reason) {
```
jonathan-innis, line 329: It looks like we still fail closed here -- but we should validate through testing

pkg/controllers/disruption/controller.go
```diff
@@ -69,6 +69,13 @@ func NewController(clk clock.Clock, kubeClient client.Client, provisioner *provi
 	cp cloudprovider.CloudProvider, recorder events.Recorder, cluster *state.Cluster, queue *orchestration.Queue,
 ) *Controller {
 	c := MakeConsolidation(clk, cluster, kubeClient, provisioner, cp, recorder, queue)
+
+	// Generate eventually disruptable reason based on a combination of drift and cloudprovider disruption reason
+	eventualDisruptionMethods := []Method{}
+	for _, reason := range append([]v1.DisruptionReason{v1.DisruptionReasonDrifted}, cp.DisruptionReasons()...) {
```
jonathan-innis, line 75: nit: `lo.Map()`?

pkg/controllers/disruption/controller.go
```diff
@@ -107,7 +115,8 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	c.recordRun("disruption-loop")
 
 	// Log if there are any budgets that are misconfigured that weren't caught by validation.
-	c.logInvalidBudgets(ctx)
+	// Only validate the first reason, since CEL validation will catch invalid disruption reasons
+	c.logInvalidBudgets(ctx, c.methods[0].Reason())
```
jonathan-innis, line 119: This feels kind of hacky, can we catch this at the NodePool validation layer? I'm find with taking this iteratively, but I don't think that we should pass in a reason in the function signature

pkg/controllers/disruption/controller.go
```diff
@@ -159,7 +168,7 @@ func (c *Controller) disrupt(ctx context.Context, disruption Method) (bool, erro
 	if len(candidates) == 0 {
 		return false, nil
 	}
-	disruptionBudgetMapping, err := BuildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient, c.recorder)
+	disruptionBudgetMapping, err := BuildDisruptionBudgets(ctx, c.cluster, c.clock, c.kubeClient, c.recorder, disruption.Reason())
```
jonathan-innis, line 171:
```suggestion
	disruptionBudgetMapping, err := BuildDisruptionBudgetMapping(ctx, c.cluster, c.clock, c.kubeClient, c.recorder, disruption.Reason())
```

nit

pkg/controllers/disruption/controller.go
```diff
@@ -266,7 +275,7 @@ func (c *Controller) logAbnormalRuns(ctx context.Context) {
 }
 
 // logInvalidBudgets will log if there are any invalid schedules detected
-func (c *Controller) logInvalidBudgets(ctx context.Context) {
+func (c *Controller) logInvalidBudgets(ctx context.Context, reason v1.DisruptionReason) {
```
jonathan-innis, line 278: do we need reason to be passed here? I don't think this is dynamic

pkg/controllers/disruption/eventual.go
```diff
@@ -70,9 +72,9 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
 		}
 		// If there's disruptions allowed for the candidate's nodepool,
 		// add it to the list of candidates, and decrement the budget.
-		if disruptionBudgetMapping[candidate.nodePool.Name][d.Reason()] > 0 {
+		if disruptionBudgetMapping[candidate.nodePool.Name] > 0 {
 			empty = append(empty, candidate)
```
jonathan-innis, line 76: Calling this "empty" I don't think is correct anymore

pkg/controllers/disruption/eventual.go
```diff
@@ -86,7 +88,7 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
 		// If the disruption budget doesn't allow this candidate to be disrupted,
 		// continue to the next candidate. We don't need to decrement any budget
 		// counter since drift commands can only have one candidate.
-		if disruptionBudgetMapping[candidate.nodePool.Name][d.Reason()] == 0 {
+		if disruptionBudgetMapping[candidate.nodePool.Name] == 0 {
```
jonathan-innis, line 91: I know we mark if something is constrained by budgets in other implementations -- is there a reason that we don't do the same thing here?

pkg/controllers/disruption/eventual_test.go
```diff
@@ -952,6 +986,529 @@ var _ = Describe("Drift", func() {
...
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*corev1.Node{node}, []*v1.NodeClaim{nodeClaim})
+
+			ExpectSingletonReconciled(ctx, disruptionController)
+
+			// Expect to not create or delete more nodeclaims
+			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(1))
+			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
+			ExpectExists(ctx, env.Client, nodeClaim)
+		})
+		It("should drift nodes that have pods with the karpenter.sh/do-not-disrupt annotation when the NodePool's TerminationGracePeriod is not nil", func() {
```
jonathan-innis, line 1111: Looks like this is handling drift -- but I think this testing is for any eventual disruptions

test/suites/perf/disruption_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		env.TimeIntervalCollector.Start("KOWK Disruption")
+		nodeClaimList := &v1.NodeClaimList{}
+		Expect(env.Client.List(env, nodeClaimList, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
+		for i := range nodeClaimList.Items {
+			Expect(nodeClaimList.Items[i].StatusConditions().SetTrue("ExampleReason")).To(BeTrue())
+			Expect(env.Client.Status().Update(env, &nodeClaimList.Items[i])).To(Succeed())
+		}
+
+		Eventually(func(g Gomega) {
```
jonathan-innis, line 60: Do we need this Eventually statement? I think this just validates that the Update succeeds -- do we need this?

pkg/controllers/disruption/eventual_test.go
engedaam: Updated the naming

pkg/controllers/disruption/controller.go
engedaam: Yeah, we can do that. I'll drop it in the function signature and relay on the values in the controller

test/suites/perf/disruption_test.go
engedaam: This was mainly an extra layer of validation, which we can drop

pkg/controllers/disruption/eventual.go
engedaam: Why is that? we are checking if the candidate is empty, no?

pkg/controllers/disruption/eventual.go
engedaam: So if I understand it correctly `constrainedByBudgets` is used to track if karpenter will consider the cluster as consolidated. https://github.com/engedaam/karpenter-core/blob/1595ce081e448e40f97ea84fa649e84f5ae655ac/pkg/controllers/disruption/multinodeconsolidation.go#L92 For eventual disruption, I don't think the idea make sense.

pkg/controllers/disruption/controller.go
engedaam: How does `lo.Map()` improve things here?

jonathan-innis: It's a nit, you could use `lo.Map` here if you want

pkg/controllers/disruption/eventual.go
jonathan-innis: Yep, just misread this section.

pkg/controllers/disruption/eventual.go
jonathan-innis: Yep, makes sense. You're right that we don't need this.

pkg/apis/v1/nodepool.go
engedaam: We test the expected behavior of the budget here https://github.com/engedaam/karpenter-core/blob/1595ce081e448e40f97ea84fa649e84f5ae655ac/pkg/apis/v1/nodepool_budgets_test.go#L181

pkg/controllers/disruption/controller.go
engedaam: I'm gonna leave this as it, I'm fully groking how this would work with `lo.Map()`

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
Nice work!

PR #1592, opened 2024-08-22
chore: disable node lease in install-kwok script

### LINKED ISSUE #1591: install-kwok: consider disabling lease controller
I noticed when using the KWOK provider during scale testing in kind that one major contributor to the KWOK controller not able to keep up with large number of pods and nodes is that its renewing the node leases every 40s using the default `kustomize` [configuration](https://github.com/kubernetes-sigs/kwok/blob/main/kustomize/kwok/deployment.yaml#L21)

Should we consider setting `node-lease-duration-seconds=0` to disable the lease controller?

yibozhuang:
Fixes #1591  

**Description**
I think we can disable the lease controller in kwok to help with scaling and performance when simulating large number of pods and nodes. I think we don't need the consider the node lease and renewal aspect from cluster autoscaling perspective.

Tested the updated KWOK provider deployment in kind with ~3000 nodes and 35K pods and KWOK controller scales better and is able to keep up with the load.

### DIFF
--- a/hack/install-kwok.sh
+++ b/hack/install-kwok.sh
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
-# This script uses kustomize to generate the manifests of the latest 
-# version of kwok, and will then either apply/delete the resources 
-# depending on the UNINSTALL_KWOK environment variable. Set this to 
+# This script uses kustomize to generate the manifests of the latest
+# version of kwok, and will then either apply/delete the resources
+# depending on the UNINSTALL_KWOK environment variable. Set this to
 # true if you want to delete the generated objects.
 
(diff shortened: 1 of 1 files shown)


hack/install-kwok.sh
```diff
@@ -72,6 +75,9 @@ do
   - op: replace
     path: /spec/template/spec/containers/0/args/2
     value: --manage-nodes-with-label-selector=kwok-partition=${alphabet[i]}
+  - op: replace
+    path: /spec/template/spec/containers/0/args/9
+    value: --node-lease-duration-seconds=0
```
wzshiming, line 80: I think we should extend this time instead of turning it off, and we need to change the heartbeat check time of the kube-controller-manager (--node-monitor-period, --node-monitor-grace-period), otherwise the node status will move to unknown in about 2 min.

PR #1603, opened 2024-08-27
chore: added cluster utilization metric and a couple of node lifetime metrics 

Youssef-Beltagy:
**Description**

I'm adding a cluster utilization metric to track the utilization of resources and a couple of node metrics to track the longevity of the nodes.

I added unit tests for the new metrics.

### DIFF
--- a/pkg/controllers/metrics/node/controller.go
+++ b/pkg/controllers/metrics/node/controller.go
@@ -49,67 +49,91 @@ const (
 var (
 	allocatable = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Namespace: "karpenter",
-			Subsystem: "nodes",
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodeSubsystem,
 			Name:      "allocatable",
 			Help:      "Node allocatable are the resources allocatable by nodes.",
 		},
-		nodeLabelNames(),
+		nodeLabelNamesWithResourceType(),
 	)
 	totalPodRequests = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Namespace: "karpenter",
-			Subsystem: "nodes",
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodeSubsystem,
 			Name:      "total_pod_requests",
 			Help:      "Node total pod requests are the resources requested by pods bound to nodes, including the DaemonSet pods.",
 		},
-		nodeLabelNames(),
+		nodeLabelNamesWithResourceType(),
 	)
 	totalPodLimits = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Namespace: "karpenter",
-			Subsystem: "nodes",
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodeSubsystem,
 			Name:      "total_pod_limits",
 			Help:      "Node total pod limits are the resources specified by pod limits, including the DaemonSet pods.",
 		},
-		nodeLabelNames(),
+		nodeLabelNamesWithResourceType(),
 	)
 	totalDaemonRequests = prometheus.NewGaugeVec(
 		prometheus.GaugeOpts{
-			Namespace: "karpenter",
-			Subsystem: "nodes",
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodeSubsystem,
 			Name:      "total_daemon_requests",
 			Help:      "Node total daemon requests are the resource requested by DaemonSet pods bound to nodes.",
 		},
-		nodeLabelNames(),
+		nodeLabelNamesWithResourceType(),
 	)
 	totalDaemonLimits = prometheus.NewGaugeVec(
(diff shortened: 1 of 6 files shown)


pkg/utils/resources/resources.go
```diff
@@ -67,10 +67,7 @@ func Merge(resources ...v1.ResourceList) v1.ResourceList {
 // constructing a new one for each sum like Merge
 func MergeInto(dest v1.ResourceList, src v1.ResourceList) v1.ResourceList {
 	if dest == nil {
```
Youssef-Beltagy, line 69: I think it will be better if we panic here. Otherwise, the behavior of this MergeInto is inconsistent and confusing: sometimes it will allocate an object for you and you need to use the return value, and other times it will just use the destination used and directly modify it.

pkg/controllers/metrics/node/controller.go
```diff
@@ -157,6 +191,43 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
+		resources.MergeInto(allocatableAggregate, node.Allocatable())
+		resources.MergeInto(utilizedAggregate, node.PodRequests())
+	}
+
+	res := make([]*metrics.StoreMetric, 0, len(allocatableAggregate))
+
+	for resourceName, allocatableResource := range allocatableAggregate {
+
+		if allocatableResource.Value() == 0 {
+			// This zero check may be unnecessary. I'm erring towards caution.
```
Youssef-Beltagy, line 209: I'm not sure this 0 check is necessary because the resources I have in mind (cpu, memory, storage) should never be 0.

As for pod resources, I'm not sure if the default value is 0 (will be problematic) or just not existing (what I expect and shouldn't be problematic).

Finally, I'm not sure if there are other resources I don't know.

pkg/controllers/node/termination/metrics.go
```diff
@@ -38,4 +46,39 @@ var (
 		},
 		[]string{metrics.NodePoolLabel},
 	)
+	NodeLifetimeDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodeSubsystem,
+			Name:      "lifetime_duration_seconds",
+			Help:      "The lifetime duration of the nodes since creation.",
+			Buckets: []float64{
```
Youssef-Beltagy, line 55: I created these bucket thresholds to show if there churn at the beginning of the node's life (within an hour), an idea of long lived nodes (that live for days), and some distribution of where everything fits in between.

pkg/controllers/metrics/node/controller.go
jonathan-innis: I'd agree that if this value is 0 that we shouldn't publish it

pkg/controllers/metrics/node/suite_test.go
```diff
@@ -99,14 +104,30 @@ var _ = Describe("Node Metrics", func() {
...
 	})
-	It("should remove the node metric gauge when the node is deleted", func() {
-		resources := corev1.ResourceList{
-			corev1.ResourcePods:   resource.MustParse("100"),
-			corev1.ResourceCPU:    resource.MustParse("5000"),
-			corev1.ResourceMemory: resource.MustParse("32Gi"),
-		}
+	It("should update the node lifetime and cluster utilization metrics", func() {
+
+		nodeCreationTimestamp := v1.NewTime(time.Now())
```
jonathan-innis, line 109: I don't think you get to control this -- this should be auto-set by the apiserver when the resource is created

pkg/controllers/node/termination/metrics.go
```diff
@@ -24,9 +24,17 @@ import (
...
 
 func init() {
-	crmetrics.Registry.MustRegister(TerminationDurationSeconds)
+	crmetrics.Registry.MustRegister(
+		TerminationDurationSeconds,
+		NodeLifetimeDurationSeconds)
 }
 
+const (
+	minuteInSeconds = 60
```
jonathan-innis, line 33: Can you just represent these in the buckets as `time.Minute.Seconds()`, `time.Hour.Seconds()`, etc.?

pkg/utils/resources/resources.go
jonathan-innis: I tend to agree with you. Seems odd that we do this but I think that we probably shouldn't change this logic in the context of this PR

pkg/controllers/node/termination/metrics.go
Youssef-Beltagy: `time.Minute` and `time.Hour` are represented in milliseconds. So I didn't use them. But now I see there is a `Seconds` function for them. Ok, will make this change.

pkg/controllers/metrics/node/suite_test.go
Youssef-Beltagy: I wasn't sure if the testing Suite sets this or not.

jonathan-innis (approved):
/lgtm
/approve

PR #1608, opened 2024-08-29
fix: allow restricted labels in pod affinity/nodeSelector

### LINKED ISSUE #1596: kubernetes.io/hostname label not working in NodeAffinity for Aerospike Kubernetes Operator
Similar issue in Karpenter: https://github.com/aws/karpenter-provider-aws/issues/4671
Duplicate of https://github.com/aws/karpenter-provider-aws/issues/6844
Related issue in Aerospike: https://github.com/aerospike/aerospike-kubernetes-operator/issues/305

**Use-case:** There is a feature in Aerospike Kubernetes Operator (AKO) called [k8sNodeBlockList](https://aerospike.com/docs/cloud/kubernetes/operator/Cluster-configuration-settings#:~:text=in%20the%20cluster.-,k8sNodeBlockList,-No) (list of K8s node names)  where a user can define a list of K8s nodes that should be ignored from scheduling for Aerospike Cluster pods. This feature of AKO helps users in K8s cluster maintenance by migrating pods to other K8s nodes. It uses  `kubernetes.io/hostname` label along with `NotIn` operator in the NodeAffinity to move pods away from those nodes.

```
        affinity:
          nodeAffinity:
              nodeSelectorTerms:
              - matchExpressions:
                - key: kubernetes.io/hostname
                  operator: NotIn
                  values:
                  - gke-abhisek-test-default-pool-d04arw3-r5ts
```                  

I**ssue:** Karpenter has a sweeping check where it blocks the `kubernetes.io/hostname` in NodeAffinity. Ref code: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1/labels.go#L91
As a result if there are pending pods with `kubernetes.io/hostname` NodeAffinity, they remain in pending state as Karpenter doesn't scale K8s node. 

Questions:
1. Is there a plan to remove that sweeping check for `kubernetes.io/hostname` label?
2. Is it possible to only block `In` operator and allow `NotIn` operator for  `kubernetes.io/hostname` label?
3. Is there a work-around possible to bypass that check?

jwcesign:
Fixes  #1596

**Description**

In general, we should not filter out the pending pods that could trigger the node scaling out by label key. If the pods can't be scheduled, it will be ignored in the scheduling simulation process: https://github.com/kubernetes-sigs/karpenter/blob/75fcd2abcae13b7677a2c4d9ee2799662b2e9445/pkg/controllers/provisioning/scheduling/scheduler.go#L197

So, let's remove the limitations of the restricted label. Also, I checked the git blame, the related PR is: https://github.com/aws/karpenter-provider-aws/pull/2051

This PR, only tries to solve: `Implemented support for GT and LT requirement operators in pods`
It doesn't want to limit label key `kubernetes.io/hostname`, so let's cancel the limitations.

make e2etests

### DIFF
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -103,13 +103,22 @@ var (
 	}
 )
 
-// IsRestrictedLabel returns an error if the label is restricted.
-func IsRestrictedLabel(key string) error {
+// IsRestrictedRequirements returns an error if the requirement label is restricted.
+func IsRestrictedRequirements(key string) error {
 	if WellKnownLabels.Has(key) {
 		return nil
 	}
+
+	// In the node pool requirements configuration, labels like kops.k8s.io/x should not be allowed, as they do not make sense.
+	labelDomain := GetLabelDomain(key)
+	for exceptionLabelDomain := range LabelDomainExceptions {
+		if strings.HasSuffix(labelDomain, exceptionLabelDomain) {
+			return fmt.Errorf("requirement label key %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
+		}
+	}
+
 	if IsRestrictedNodeLabel(key) {
-		return fmt.Errorf("label %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
+		return fmt.Errorf("requirement label key %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
 	}
 	return nil
 }
--- a/pkg/apis/v1/nodeclaim_validation.go
+++ b/pkg/apis/v1/nodeclaim_validation.go
@@ -110,16 +110,23 @@ func (in *NodeClaimTemplateSpec) validateRequirements() (errs error) {
 	return errs
 }
 
+// ValidateRequirement checks if the requirements are valid, including restricted labels, and the format of the requirements.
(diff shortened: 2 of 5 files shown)


jwcesign:
Please take a look: 
cc @ellistarn @jonathan-innis @tzneal @bwagner5 @njtran

jwcesign:
Please take a look: 
cc @ellistarn @jonathan-innis @tzneal @bwagner5 @njtran

sftim:
In the PR description, consider explaining why these changes are the right way to fix it.

jwcesign:
> In the PR description, consider explaining why these changes are the right way to fix it.

Thanks, I added

pkg/apis/v1/labels.go
```diff
@@ -103,13 +103,22 @@ var (
...
 		return nil
 	}
+
+	// In the node pool requirements configuration, labels like kops.k8s.io/x should not be allowed, as they do not make sense.
+	labelDomain := GetLabelDomain(key)
+	for exceptionLabelDomain := range LabelDomainExceptions {
+		if strings.HasSuffix(labelDomain, exceptionLabelDomain) {
+			return fmt.Errorf("requirement label key %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
+		}
+	}
```
njtran, line 118: I'm not sure I understand this change. We're specifically allowing these sub-domains as an exception, understanding that some users use these labels. Are you saying all kops labels shouldn't be used?

jwcesign: For nodepool's requirements, I didn't see any reason to configure something like this.

jwcesign: It should be used as nodepool's labels, not requirements.

jwcesign: I check the logic, this function is used to check the nodepool's requirements.

jwcesign: cc @njtran

PR #1609, opened 2024-08-29
doc: make kwok docs better for local development

jwcesign:
Fixes #N/A 

**Description**

doc: make kwok docs better for local development 

no code change

### DIFF
--- a/kwok/README.md
+++ b/kwok/README.md
@@ -8,10 +8,16 @@ Before using the kwok provider, make sure that you don't have an installed versi
 - Have a cluster that you can install Karpenter on to.
   - For an example on how to make a cluster in AWS, refer to [karpenter.sh](https://karpenter.sh/docs/getting-started/getting-started-with-karpenter/)
 
+If you use a kind cluster, please set the the following environment variables:
+```bash
+export KWOK_REPO=kind.local
+export KIND_CLUSTER_NAME=<kind cluster name, for example, chart-testing>
+```
+
 ## Installing
 ```bash
 make install-kwok
-make apply
+make apply # Run this command again to redeploy if the code has changed
 ```
 
 ## Create a NodePool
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,7 +53,7 @@ e2etests: ## Run the e2e suite against your local cluster
 
 # Run make install-kwok to install the kwok controller in your cluster first
 # Webhooks are currently not supported in the kwok provider.
-apply: verify build ## Deploy the kwok controller from the current state of your git repository into your ~/.kube/config cluster
```
jwcesign, line 56: Separate verify and apply operations, if not, it costs much time to verify when I need to deploy again after the code changed.

jwcesign:
Please take a look: 
cc @ellistarn @jonathan-innis @tzneal @bwagner5 @njtran

Makefile
njtran: Why do you want to skip verify? This does some yq logic to add into CRDs before we apply them on the `make apply` target.

njtran (commented):
One comment

Makefile
jwcesign: Most of the time, I only modify some function logic. Verifying it every time is unnecessary, which makes my development process slow.

This is my original intention for this modification.

jwcesign: If someone wants to develop:
they should run:
```sh
make verify
make apply #Run this command again to redeploy if the code has changed
```

jwcesign:
Please take a look @ellistarn @jonathan-innis @tzneal @bwagner5 @njtran

njtran (commented):
I think the docs look good, I'm not convinced on removing the verify part of make apply.

jwcesign:
> I think the docs look good, I'm not convinced on removing the verify part of make apply.

I delete that modification part.

njtran (commented):
/lgtm
/approve

PR #1629, opened 2024-09-03
fix: handle webhook inability to connect error v37

rschalo:
Fixes #N/A 

**Description**

make presubmit

### DIFF
--- a/pkg/webhooks/webhooks.go
+++ b/pkg/webhooks/webhooks.go
@@ -217,12 +217,15 @@ func ValidateConversionEnabled(ctx context.Context, kubeclient client.Client) {
 	defer cancel()
 	var err error
 	v1np := &v1.NodePoolList{}
-	deadline, _ := listCtx.Deadline()
-	for time.Now().Before(deadline) {
-		if err = kubeclient.List(listCtx, v1np, &client.ListOptions{Limit: 1}); err == nil {
+	for {
+		err = kubeclient.List(listCtx, v1np, &client.ListOptions{Limit: 1})
+		if err == nil {
 			return
 		}
-		time.Sleep(10 * time.Second)
+		select {
+		case <-listCtx.Done():
+			panic("Conversion webhook enabled but unable to complete call: " + err.Error())
+		case <-time.After(10 * time.Second):
+		}
 	}
-	panic("Conversion webhook enabled but unable to complete call: " + err.Error())
 }
--- a/pkg/webhooks/webhooks.go
+++ b/pkg/webhooks/webhooks.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/samber/lo"
 	"go.uber.org/zap"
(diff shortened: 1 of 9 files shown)


pkg/webhooks/webhooks.go
```diff
@@ -208,3 +210,16 @@ func HealthProbe(ctx context.Context) healthz.Checker {
 		return nil
 	}
 }
+
+func ValidateConversionEnabled(ctx context.Context, kubeclient client.Client) {
+	// allow context to exist longer than cache sync timeout
+	listCtx, cancel := context.WithTimeout(ctx, 130*time.Second)
```
jonathan-innis, line 216: nice! Can we ref to where the cache sync timeout is defined? Is this configurable or will this always be 2m?

pkg/webhooks/webhooks.go
```diff
@@ -208,3 +210,16 @@ func HealthProbe(ctx context.Context) healthz.Checker {
...
+	defer cancel()
+	// sleep for cache hydration
+	// wait for cache to sync, controller-runtime defaults to 120 seconds
+	time.Sleep(120 * time.Second)
```
jonathan-innis, line 220: Why not continually poll and just exit out as soon as we succeed?

jmdeal: +1 on polling. Also, if we do keep the sleep, should it be greater than the controller-runtime cache invalidation period to prevent races, rather than equal.

pkg/webhooks/webhooks.go
rschalo: It's configurable here: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controller.go#L49

I don't see us changing this default anywhere in the repo.

pkg/webhooks/webhooks.go
```diff
@@ -208,3 +210,19 @@ func HealthProbe(ctx context.Context) healthz.Checker {
...
 	}
 }
+
+func ValidateConversionEnabled(ctx context.Context, kubeclient client.Client) {
+	// allow context to exist longer than cache sync timeout which has a default of 120 seconds
+	listCtx, cancel := context.WithTimeout(ctx, 130*time.Second)
+	defer cancel()
+	var err error
+	v1np := &v1.NodePoolList{}
+	deadline, _ := listCtx.Deadline()
```
jonathan-innis, line 220: It's a bit easier to do `listCtx.Done()` as a rule of thumb

pkg/webhooks/webhooks.go
```diff
@@ -208,3 +210,19 @@ func HealthProbe(ctx context.Context) healthz.Checker {
...
+	for time.Now().Before(deadline) {
+		if err = kubeclient.List(listCtx, v1np, &client.ListOptions{Limit: 1}); err == nil {
```
jonathan-innis, line 222: Nice!

jonathan-innis (approved):
/lgtm
/approve

PR #1632, opened 2024-09-03
fix: handle webhook inability to connect error v34

rschalo:
Fixes #N/A 

**Description**

make presubmit


jonathan-innis (approved):
/lgtm
/approve

PR #1633, opened 2024-09-03
fix: handle webhook inability to connect error v33

rschalo:
Fixes #N/A 

**Description**

make presubmit

### DIFF
--- a/pkg/webhooks/webhooks.go
+++ b/pkg/webhooks/webhooks.go
@@ -208,14 +208,20 @@ func HealthProbe(ctx context.Context) healthz.Checker {
 }
 
 func ValidateConversionEnabled(ctx context.Context, kubeclient client.Client) {
-	// allow context to exist longer than cache sync timeout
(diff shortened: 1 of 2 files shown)


pkg/webhooks/webhooks.go
```diff
@@ -204,3 +206,16 @@ func HealthProbe(ctx context.Context) healthz.Checker {
...
 	}
 }
+
+func ValidateConversionEnabled(ctx context.Context, kubeclient client.Client) {
+	// allow context to exist longer than cache sync timeout
+	listCtx, cancel := context.WithTimeout(ctx, 130*time.Second)
+	defer cancel()
+	// sleep for cache hydration
+	// wait for cache to sync, controller-runtime defaults to 120 seconds
+	time.Sleep(120 * time.Second)
```
njtran, line 216: Can we do something else here? a 120s sleep screams like we haven't thought about how to best wait for this to succeed.

jonathan-innis (approved):
/lgtm
/approve

PR #1650, opened 2024-09-10
chore: consolidationWarnings use goroutine

helen-frank:
**Description**
Asynchronously accelerating the execution of `consolidationWarnings`


njtran (commented):
same here. Why do we need this? What's the performance benefit?

helen-frank:
> same here. Why do we need this? What's the performance benefit?

The two for loops do not interfere with each other's logic, and I think that simple asynchronous processing can make it faster, more so when there are more pods.

PR #1657, opened 2024-09-12
[WIP] Migration controller

rschalo:
Fixes #N/A 

**Description**

### DIFF
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -51,6 +51,7 @@ const (
 	KubeletCompatibilityAnnotationKey          = apis.CompatibilityGroup + "/v1beta1-kubelet-conversion"
 	NodeClassReferenceAnnotationKey            = apis.CompatibilityGroup + "/v1beta1-nodeclass-reference"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
+	StorageVersion                             = apis.Group + "/storage-version"
 )
 
 // Karpenter specific finalizers
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -34,6 +34,7 @@ import (
 	metricsnode "sigs.k8s.io/karpenter/pkg/controllers/metrics/node"
 	metricsnodepool "sigs.k8s.io/karpenter/pkg/controllers/metrics/nodepool"
 	metricspod "sigs.k8s.io/karpenter/pkg/controllers/metrics/pod"
+	"sigs.k8s.io/karpenter/pkg/controllers/migration"
 	"sigs.k8s.io/karpenter/pkg/controllers/node/termination"
 	"sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator"
 	nodeclaimconsistency "sigs.k8s.io/karpenter/pkg/controllers/nodeclaim/consistency"
@@ -93,5 +94,7 @@ func NewControllers(
 		leasegarbagecollection.NewController(kubeClient),
 		status.NewController[*v1.NodeClaim](kubeClient, mgr.GetEventRecorderFor("karpenter")),
 		status.NewController[*v1.NodePool](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		migration.NewController[*v1.NodeClaim](kubeClient),
+		migration.NewController[*v1.NodePool](kubeClient),
 	}
 }
--- a/pkg/controllers/migration/controller.go
+++ b/pkg/controllers/migration/controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/awslabs/operatorpkg/object"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -68,7 +69,7 @@ func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (r
 	annotations[v1.StorageVersion] = "v1"
 
 	ctx = injection.WithControllerName(ctx, "migration")
-	log.FromContext(ctx).V(1).Info(fmt.Sprintf("Annotating v1 stored version on %s: %s",
+	log.FromContext(ctx).V(1).Info(fmt.Sprintf("annotating v1 stored version on %s: %s",
 		o.GetObjectKind().GroupVersionKind().Kind, o.GetName()))
 	o.SetAnnotations(annotations)
 
@@ -87,6 +88,6 @@ func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		For(o).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
-		Named(fmt.Sprintf("migration.%v", reflect.TypeOf(o))).
+		Named(fmt.Sprintf("migration.%s", strings.ToLower(reflect.TypeOf(object.New[T]()).Elem().Name()))).
 		Complete(c)
 }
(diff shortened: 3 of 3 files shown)


pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,89 @@
...
+	annotations[v1.StorageVersion] = "v1"
+
+	ctx = injection.WithControllerName(ctx, "migration")
+	log.FromContext(ctx).V(1).Info(fmt.Sprintf("Ensuring storage version migration by annotating %s: %s",
+		o.GetObjectKind().GroupVersionKind().Kind, o.GetName()))
+	o.SetAnnotations(annotations)
+
+	if !equality.Semantic.DeepEqual(stored, o) {
+		if err := c.kubeClient.Update(ctx, o); err != nil {
+			return reconcile.Result{}, fmt.Errorf("syncing node, %w", err)
```
rschalo, line 76: Wrong error here - will update after early feedback.

pkg/apis/v1/labels.go
```diff
@@ -51,6 +51,7 @@ const (
 	KubeletCompatibilityAnnotationKey          = apis.CompatibilityGroup + "/v1beta1-kubelet-conversion"
 	NodeClassReferenceAnnotationKey            = apis.CompatibilityGroup + "/v1beta1-nodeclass-reference"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
+	StorageVersion                             = apis.Group + "/storage-version"
```
rschalo, line 54: Probably better as `StoredVersion`

jonathan-innis, line 54:
```suggestion
	StorageMigrated                             = apis.Group + "/storage-migrated"
```

Should this just be a true/false to signal that we have migrated to the new storage version?

pkg/controllers/controllers.go
```diff
@@ -93,5 +94,7 @@ func NewControllers(
 		leasegarbagecollection.NewController(kubeClient),
 		status.NewController[*v1.NodeClaim](kubeClient, mgr.GetEventRecorderFor("karpenter")),
 		status.NewController[*v1.NodePool](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		migration.NewController[*v1.NodeClaim](kubeClient),
```
jonathan-innis, line 97: Nice use of generics 🙌🏼

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+)
+
+type Object interface {
```
jonathan-innis, line 40: Since this just wraps a regular `client.Object`, I don't think you need another interface

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	client.Object
+}
+
+type Controller[T Object] struct {
```
jonathan-innis, line 44:
```suggestion
type Controller[T client.Object] struct {
```

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+func NewController[T Object](client client.Client) *Controller[T] {
+	return &Controller[T]{
+		kubeClient: client,
+	}
+}
+
+func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	o := object.New[T]()
+
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, o); err != nil {
```
jonathan-innis, line 57: Can you use the `controller.As` to drop your need for doing the `Get` call yourself?

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("getting object, %w", err)
+	}
+	stored := o.DeepCopyObject()
+	annotations := o.GetAnnotations()
```
jonathan-innis, line 64: I wonder if we can just assume this is idempotent. e.g. Always set it and then do an equality check to see if it's changed to patch it

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	if _, ok := annotations[v1.StorageVersion]; ok {
+		return reconcile.Result{}, nil
+	}
+	annotations[v1.StorageVersion] = "v1"
+
+	ctx = injection.WithControllerName(ctx, "migration")
```
jonathan-innis, line 70: Can we put this at the top of the controller so that all of our controllers are consistent

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	log.FromContext(ctx).V(1).Info(fmt.Sprintf("annotating v1 stored version on %s: %s",
```
jonathan-innis, line 71: I think we want to log after the fact generally so we reduce log spam. The other thing: I'm not sure how valuable logging this is to users

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+		o.GetObjectKind().GroupVersionKind().Kind, o.GetName()))
```
jonathan-innis, line 72: A lot of this information is already in the log context so you don't have to add it yourself directly

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+	o.SetAnnotations(annotations)
+
+	if !equality.Semantic.DeepEqual(stored, o) {
+		if err := c.kubeClient.Patch(ctx, o, client.MergeFromWithOptions(client.Object(o), client.MergeFromWithOptimisticLock{})); err != nil {
```
jonathan-innis, line 76: I don't think you need the optimistic lock here since it's a map with a key that you own

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
+	o := object.New[T]()
+	return controllerruntime.NewControllerManagedBy(m).
+		For(o).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		Named(fmt.Sprintf("migration.%s", strings.ToLower(reflect.TypeOf(object.New[T]()).Elem().Name()))).
```
jonathan-innis, line 90: nice 🙌🏼

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, o); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("getting object, %w", err)
+	}
+	stored := o.DeepCopyObject()
+	annotations := o.GetAnnotations()
+	if _, ok := annotations[v1.StorageVersion]; ok {
```
njtran, line 65: nit: check this before doing deepcopy saves a tiny bit of time

jonathan-innis:
I think we're missing the controller here as well that does a list across all NodeClaims/NodePools and removes the `storedVersions` from the CRD once they are all migrated

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,92 @@
...
+		return reconcile.Result{}, nil
+	}
+	annotations[v1.StorageVersion] = "v1"
```
njtran, line 68: lo.assign

pkg/apis/v1/labels.go
njtran: storage-version-migrated?

PR #1666, opened 2024-09-16
fix: Fix race condition with using newer NodePool for hash annotation

jonathan-innis:
Fixes #N/A 

**Description**

Fix hash annotation being generated from a newer NodePool than used during scheduling. This new code guarantees that the NodePool that is used during scheduling is what will be used to generate the hash annotation.

This ensures that if the NodePool changes while we are scheduling, we won't use the new configuration to generate the hash, which could lead to use putting a hash on the NodeClaim that doesn't actually represent the configuration that the NodeClaim was generated from.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/suite_test.go
+++ b/pkg/controllers/provisioning/suite_test.go
@@ -29,6 +29,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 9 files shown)


pkg/controllers/provisioning/suite_test.go
```diff
@@ -246,6 +247,31 @@ var _ = Describe("Provisioning", func() {
 			ExpectScheduled(ctx, env.Client, pod)
 		}
 	})
+	It("should not use a different NodePool hash on the NodeClaim if the NodePool changes during scheduling", func() {
```
rschalo, line 250: I'm not following - where is the nodepool changed in the test?

jonathan-innis: lol good point. I'm an amazing test writer :D

PR #1667, opened 2024-09-16
fix: Add compat annotation key when using NodePool's kubelet config compat annotation

jonathan-innis:
Fixes https://github.com/aws/karpenter-provider-aws/issues/6987

**Description**

Ensure that we add the `compatability.karpenter.sh/v1beta1-kubelet-conversion` is propagated onto NodeClaims when NodePool's use the kubelet configuration conversion during scheduling. 

Prior to this change, it was possible for the scheduler to call the GetInstanceTypes() API with one kubelet configuration (parsed by the the `compatability.karpenter.sh/v1beta1-kubelet-conversion` annotation) which results in higher capacity/allocatable values to be passed to the scheduler. The NodeClaim wouldn't contain this annotation, though; so, when a CloudProvider goes to launch the NodeClaim or when the NodeClaim is considered later for capacity/allocatable, it might use a completely different set of values.

This change ensures that the `compatability.karpenter.sh/v1beta1-kubelet-conversion` so that the CloudProvider is able to reason about this Kubelet Config consistently across NodePools and NodeClaims.

`make presubmit`

### DIFF
--- a/pkg/apis/v1/nodepool_conversion.go
+++ b/pkg/apis/v1/nodepool_conversion.go
@@ -32,7 +32,7 @@ import (
 )
 
 // Convert v1 NodePool to v1beta1 NodePool
-func (in *NodePool) ConvertTo(ctx context.Context, to apis.Convertible) error {
+func (in *NodePool) ConvertTo(_ context.Context, to apis.Convertible) error {
 	v1beta1NP := to.(*v1beta1.NodePool)
 	v1beta1NP.ObjectMeta = in.ObjectMeta
 
@@ -123,7 +123,7 @@ func (in *NodePool) ConvertFrom(ctx context.Context, v1beta1np apis.Convertible)
 		return err
 	}
 	if kubeletAnnotation == "" {
-		in.Annotations = lo.OmitByKeys(in.Annotations, []string{KubeletCompatibilityAnnotationKey})
+		delete(in.Annotations, KubeletCompatibilityAnnotationKey)
 	} else {
 		in.Annotations = lo.Assign(in.Annotations, map[string]string{KubeletCompatibilityAnnotationKey: kubeletAnnotation})
 	}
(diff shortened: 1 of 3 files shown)


pkg/apis/v1/nodepool_conversion.go
```diff
@@ -123,9 +123,9 @@ func (in *NodePool) ConvertFrom(ctx context.Context, v1beta1np apis.Convertible)
 		return err
 	}
 	if kubeletAnnotation == "" {
-		in.Annotations = lo.OmitByKeys(in.Annotations, []string{KubeletCompatibilityAnnotationKey})
+		delete(in.Annotations, KubeletCompatibilityAnnotationKey)
 	} else {
-		in.Annotations = lo.Assign(in.Annotations, map[string]string{KubeletCompatibilityAnnotationKey: kubeletAnnotation})
+		in.Annotations = lo.Assign(in.Spec.Template.Annotations, map[string]string{KubeletCompatibilityAnnotationKey: kubeletAnnotation})
```
engedaam, line 128: If we are going to add the kubelet annotation to the nodeclaim template spec, we should update this here or else I think this will not work as expected:
- https://github.com/kubernetes-sigs/karpenter/pull/1667/files#diff-b8d4e10f084dfecb98330acb7dc0b082455afea1837c74bd59ab11d29b8518adR126
- https://github.com/kubernetes-sigs/karpenter/pull/1667/files#diff-b8d4e10f084dfecb98330acb7dc0b082455afea1837c74bd59ab11d29b8518adR42

jonathan-innis: I thought about this but figured that it would be more trouble than it's worth. Basically, it would mean that a user could have this annotation on either of two places -- and we would have to reason about both of them to not make a breaking change.

I don't think anything is wrong with the way that this is currently implemented. Can you describe how you are thinking the current configuration is going to cause problems?

engedaam (commented):
/lgtm

PR #1672, opened 2024-09-17
chore: Additional upstream metrics

jigisha620:
Fixes #N/A 

**Description**
Adding new metrics.
1. karpenter_nodes_drained_total
2. karpenter_ignored_pod_count
3. karpenter_pods_current_unbound_duration_seconds
4. karpenter_pods_bound_duration_seconds
5. karpenter_pods_unstarted_time_seconds
6. karpenter_cluster_state_unsynced_time_seconds
7. karpenter_nodes_eviction_requests_total

Added tests and tested on local cluster

### DIFF
--- a/pkg/controllers/node/termination/terminator/eviction.go
+++ b/pkg/controllers/node/termination/terminator/eviction.go
@@ -180,6 +180,11 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 				},
 			},
 		}); err != nil {
+		var apiStatus apierrors.APIStatus
+		if errors.As(err, &apiStatus) {
+			code := apiStatus.Status().Code
+			NodesEvictionRequestsTotal.With(map[string]string{CodeLabel: fmt.Sprint(code)}).Inc()
+		}
 		// status codes for the eviction API are defined here:
 		// https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/#how-api-initiated-eviction-works
 		if apierrors.IsNotFound(err) || apierrors.IsConflict(err) {
@@ -199,6 +204,7 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 		log.FromContext(ctx).Error(err, "failed evicting pod")
 		return false
 	}
+	NodesEvictionRequestsTotal.With(map[string]string{CodeLabel: "200"}).Inc()
 	q.recorder.Publish(terminatorevents.EvictPod(&corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: key.Name, Namespace: key.Namespace}}))
 	return true
 }
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -159,13 +159,14 @@ func (p *Provisioner) GetPendingPods(ctx context.Context) ([]*corev1.Pod, error)
 	if err != nil {
 		return nil, fmt.Errorf("listing pods, %w", err)
 	}
-	pods = lo.Reject(pods, func(po *corev1.Pod, _ int) bool {
+	rejectedPods, pods := lo.FilterReject(pods, func(po *corev1.Pod, _ int) bool {
 		if err := p.Validate(ctx, po); err != nil {
 			log.FromContext(ctx).WithValues("Pod", klog.KRef(po.Namespace, po.Name)).V(1).Info(fmt.Sprintf("ignoring pod, %s", err))
 			return true
 		}
 		return false
 	})
+	scheduler.IgnoredPodCount.Set(float64(len(rejectedPods)))
 	p.consolidationWarnings(ctx, pods)
 	return pods, nil
 }
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -59,8 +59,9 @@ type Cluster struct {
 	// changed about the cluster that might make consolidation possible. By recording
 	// the state, interested disruption methods can check to see if this has changed to
 	// optimize and not try to disrupt if nothing about the cluster has changed.
-	clusterState     time.Time
-	antiAffinityPods sync.Map // pod namespaced name -> *corev1.Pod of pods that have required anti affinities
+	clusterState      time.Time
+	unsyncedStartTime time.Time
+	antiAffinityPods  sync.Map // pod namespaced name -> *corev1.Pod of pods that have required anti affinities
 }
 
 func NewCluster(clk clock.Clock, client client.Client) *Cluster {
@@ -82,6 +83,18 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
 //
 //nolint:gocyclo
 func (c *Cluster) Synced(ctx context.Context) (synced bool) {
+	// Set the metric depending on the result of the Synced() call
+	defer func() {
+		if synced {
+			c.unsyncedStartTime = time.Time{}
+			ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(0)
+		} else {
+			if c.unsyncedStartTime.IsZero() {
+				c.unsyncedStartTime = c.clock.Now()
+			}
+			ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(c.clock.Since(c.unsyncedStartTime).Seconds())
+		}
+	}()
 	// Set the metric to whatever the result of the Synced() call is
 	defer func() {
 		ClusterStateSynced.Set(lo.Ternary[float64](synced, 1, 0))
--- a/pkg/controllers/state/suite_test.go
+++ b/pkg/controllers/state/suite_test.go
@@ -88,6 +88,7 @@ var _ = AfterSuite(func() {
 
 var _ = BeforeEach(func() {
 	fakeClock.SetTime(time.Now())
+	state.ClusterStateUnsyncedTimeSeconds.Reset()
 	cloudProvider.InstanceTypes = fake.InstanceTypesAssorted()
 	nodePool = test.NodePool(v1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: "default"}})
 	ExpectApplied(ctx, env.Client, nodePool)
@@ -1126,6 +1127,27 @@ var _ = Describe("Cluster State Sync", func() {
 		Expect(cluster.Synced(ctx)).To(BeTrue())
 		ExpectMetricGaugeValue(state.ClusterStateSynced, 1.0, nil)
 		ExpectMetricGaugeValue(state.ClusterStateNodesCount, 1000.0, nil)
+		metric, found := FindMetricWithLabelValues("karpenter_cluster_state_unsynced_time_seconds", map[string]string{})
+		Expect(found).To(BeTrue())
+		Expect(metric.GetGauge().GetValue()).To(BeEquivalentTo(0))
+	})
+	It("should emit cluster_state_unsynced_time_seconds metric when cluster state is unsynced", func() {
+		nodeClaim := test.NodeClaim(v1.NodeClaim{
+			Status: v1.NodeClaimStatus{
+				ProviderID: "",
+			},
+		})
+		nodeClaim.Status.ProviderID = ""
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		ExpectReconcileSucceeded(ctx, nodeClaimController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(cluster.Synced(ctx)).To(BeFalse())
+
+		fakeClock.Step(2 * time.Minute)
+		ExpectReconcileSucceeded(ctx, nodeClaimController, client.ObjectKeyFromObject(nodeClaim))
+		Expect(cluster.Synced(ctx)).To(BeFalse())
+		metric, found := FindMetricWithLabelValues("karpenter_cluster_state_unsynced_time_seconds", map[string]string{})
+		Expect(found).To(BeTrue())
+		Expect(metric.GetGauge().GetValue()).To(BeNumerically(">=", 120))
 	})
 	It("should consider the cluster state synced when nodes don't have provider id", func() {
 		// Deploy 1000 nodes and sync them all with the cluster
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -129,6 +129,9 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+	NodesDrainedTotal.With(prometheus.Labels{
+		metrics.NodePoolLabel: node.Labels[v1.NodePoolLabelKey],
+	}).Inc()
 	// In order for Pods associated with PersistentVolumes to smoothly migrate from the terminating Node, we wait
 	// for VolumeAttachments of drain-able Pods to be cleaned up before terminating Node and removing its finalizer.
 	// However, if TerminationGracePeriod is configured for Node, and we are past that period, we will skip waiting.
--- a/pkg/controllers/node/termination/metrics.go
+++ b/pkg/controllers/node/termination/metrics.go
@@ -28,7 +28,8 @@ import (
 func init() {
 	crmetrics.Registry.MustRegister(
 		TerminationDurationSeconds,
-		NodeLifetimeDurationSeconds)
+		NodeLifetimeDurationSeconds,
+		NodesDrainedTotal)
 }
 
 const dayDuration = time.Hour * 24
@@ -41,6 +42,15 @@ var (
 			Name:       "termination_duration_seconds",
 			Help:       "The time taken between a node's deletion request and the removal of its finalizer",
 			Objectives: metrics.SummaryObjectives(),
+		},
+		[]string{metrics.NodePoolLabel},
+	)
+	NodesDrainedTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodeSubsystem,
+			Name:      "drained_total",
+			Help:      "The total number of nodes drained by Karpenter",
 		},
 		[]string{metrics.NodePoolLabel},
 	)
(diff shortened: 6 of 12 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -346,6 +347,7 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 	results := s.Solve(ctx, pods).TruncateInstanceTypes(scheduler.MaxInstanceTypes)
 	if len(results.NewNodeClaims) > 0 {
 		log.FromContext(ctx).WithValues("Pods", pretty.Slice(lo.Map(pods, func(p *corev1.Pod, _ int) string { return klog.KRef(p.Namespace, p.Name).String() }), 5), "duration", time.Since(start)).Info("found provisionable pod(s)")
+		scheduler.PodsNominatedTotal.With(prometheus.Labels{scheduler.ControllerLabel: injection.GetControllerName(ctx)}).Add(float64(len(pods)))
```
jonathan-innis, line 350: So, is this going to double-count? What happens when the same pod is looped through the scheduler function multiple times?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -162,6 +162,7 @@ func (p *Provisioner) GetPendingPods(ctx context.Context) ([]*corev1.Pod, error)
 	pods = lo.Reject(pods, func(po *corev1.Pod, _ int) bool {
 		if err := p.Validate(ctx, po); err != nil {
 			log.FromContext(ctx).WithValues("Pod", klog.KRef(po.Namespace, po.Name)).V(1).Info(fmt.Sprintf("ignoring pod, %s", err))
+			metrics.IgnoredPodTotal.Inc()
```
jonathan-innis, line 165: This is going to be a counter, but I don't think we want it to be a counter. I think we want this to be a gauge that tracks the current number of ignored pods at any given point in time.

pkg/controllers/provisioning/provisioner.go
jigisha620: Resolving as per our discussion offline.

pkg/controllers/metrics/pod/controller.go
```diff
@@ -132,13 +141,19 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
-	c.recordPodStartupMetric(pod)
+	c.recordPodStartupMetric(pod, labels)
 	return reconcile.Result{}, nil
 }
 
-func (c *Controller) recordPodStartupMetric(pod *corev1.Pod) {
+func (c *Controller) recordPodStartupMetric(pod *corev1.Pod, labels prometheus.Labels) {
 	key := client.ObjectKeyFromObject(pod).String()
 	if pod.Status.Phase == phasePending {
+		cond, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
```
jonathan-innis, line 151: If the pod has the scheduled condition, will it also be pending? Don't you need to do something like track that the pod _was_ pending and then know that it has now been marked as scheduled

pkg/controllers/metrics/pod/controller.go
```diff
@@ -132,22 +164,57 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			Labels:   labels,
 		},
 	})
-	c.recordPodStartupMetric(pod)
+	c.recordPodStartupMetric(pod, labels)
 	return reconcile.Result{}, nil
 }
 
-func (c *Controller) recordPodStartupMetric(pod *corev1.Pod) {
+func (c *Controller) recordPodStartupMetric(pod *corev1.Pod, labels prometheus.Labels) {
```
jonathan-innis, line 171: nit:This function name is starting to get misleading -- can we consider renaming it/breaking out this single function into multiple so it's easier to track what things are happening here.

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -182,11 +184,15 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
-		if apierrors.IsNotFound(err) || apierrors.IsConflict(err) {
+		if apierrors.IsNotFound(err) {
 			// 404 - The pod no longer exists
 			// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L160
+			return true
+		}
+		if apierrors.IsConflict(err) {
 			// 409 - The pod exists, but it is not the same pod that we initiated the eviction on
 			// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L318
+			metrics.NodesEvictionRequestsTotal.With(map[string]string{metrics.ResponseLabel: "conflict"}).Inc()
```
jonathan-innis, line 195: Should we just consider surfacing the raw status codes and then you don't have to worry about creating custom labels yourself? `client_go_request_total` calls it `code`

pkg/controllers/state/cluster.go
```diff
@@ -82,6 +82,16 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
 //
 //nolint:gocyclo
 func (c *Cluster) Synced(ctx context.Context) (synced bool) {
+	start := time.Now()
+	ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(time.Since(start).Seconds())
```
jonathan-innis, line 86: Why do you need to set this metric twice?

pkg/controllers/state/cluster.go
```diff
@@ -82,6 +82,16 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
 //
 //nolint:gocyclo
 func (c *Cluster) Synced(ctx context.Context) (synced bool) {
+	start := time.Now()
```
jonathan-innis, line 85: This isn't going to be held onto through multiple calls to this function. You need to retain the information in the cluster state of the first time that we were unsynced and then diff from that

pkg/controllers/state/cluster.go
```diff
@@ -82,6 +82,16 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
...
+	ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(time.Since(start).Seconds())
+	// Set/Delete the metric depending on the result of the Synced() call
+	defer func() {
+		if synced {
+			ClusterStateUnsyncedTimeSeconds.Delete(map[string]string{})
```
jonathan-innis, line 90: I don't think we should delete this metric when we are synced -- we should just set the metric to 0

pkg/metrics/constants.go
```diff
@@ -33,6 +33,9 @@ const (
 	// Reasons for CREATE/DELETE shared metrics
 	ProvisionedReason = "provisioned"
 	ExpiredReason     = "expired"
+
+	// Response for eviction request
+	ResponseLabel = "response"
```
jonathan-innis, line 38:
```suggestion
	CodeLabel = "code"
```

Consider calling this `code` like `operatorpkg` does with its metric labels

pkg/metrics/metrics.go
```diff
@@ -89,9 +89,25 @@ var (
...
 		},
 	)
+	IgnoredPodCount = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Name:      "ignored_pod_count",
+			Help:      "Number of pods ignored during scheduling by Karpenter",
+		},
+	)
+	NodesEvictionRequestsTotal = prometheus.NewCounterVec(
```
jonathan-innis, line 99: This is only relevant to termination -- should this be in that package's metrics file

pkg/metrics/metrics.go
```diff
@@ -89,9 +89,25 @@ var (
 			NodePoolLabel,
 		},
 	)
+	IgnoredPodCount = prometheus.NewGauge(
```
jonathan-innis, line 92: This is only relevant to provisioning -- should this be in that package's metrics file

pkg/controllers/metrics/pod/controller.go
```diff
@@ -132,22 +164,57 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			podName:      pod.Name,
+			podNameSpace: pod.Namespace,
+		})
+		podBoundDurationSeconds.With(labels).Observe(condScheduled.LastTransitionTime.Sub(pod.CreationTimestamp.Time).Seconds())
+		c.unscheduledPods.Delete(key)
+	}
+	condReady, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
 		return c.Type == corev1.PodReady
 	})
 	if c.pendingPods.Has(key) && ok {
```
jonathan-innis, line 203: I'm a little worried about how complicated this function is getting and the number of conditional paths and data structures that we have to reason about. If this function gets broken up I think that it might help with being able to reason about everything

pkg/controllers/state/cluster.go
jigisha620: Agreed. Not needed.

pkg/controllers/metrics/pod/controller.go
jigisha620: Agreed

pkg/controllers/metrics/pod/controller.go
```diff
@@ -71,18 +72,47 @@ var (
 			Objectives: metrics.SummaryObjectives(),
 		},
 	)
+	podBoundDurationSeconds = prometheus.NewHistogramVec(
```
jonathan-innis, line 75: I have a question on metric cardinality that I'm trying to get an answer on before we close this out -- specifically: I'm trying to understand from the K8s side if we should be concerned about the total number of timeseries that this could produce or if we should just be concerned about the _active_ timeseries. 

If it's the latter, I don't think that we have a problem here. If it's the former, we'll have to do some more thinking about how we can add this metric -- or any metric that can have potentially undefined cardinality

pkg/controllers/metrics/pod/controller.go
```diff
@@ -71,18 +72,47 @@ var (
...
+		prometheus.HistogramOpts{
+			Namespace: "karpenter",
+			Subsystem: metrics.PodSubsystem,
+			Name:      "bound_duration_seconds",
+			Help:      "The time from pod creation until the pod is bound.",
+			Buckets:   metrics.DurationBuckets(),
+		},
+		labelNames(),
```
jonathan-innis, line 83: Since this is a histogram and historical, we shouldn't differentiate the buckets by this many labels -- we just want an aggregated view historically of all of the pod bound times -- I actually don't think we need any _labels_ here

pkg/controllers/metrics/pod/controller.go
```diff
@@ -71,18 +72,47 @@ var (
 			Objectives: metrics.SummaryObjectives(),
 		},
 	)
+	podBoundDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "karpenter",
+			Subsystem: metrics.PodSubsystem,
+			Name:      "bound_duration_seconds",
+			Help:      "The time from pod creation until the pod is bound.",
+			Buckets:   metrics.DurationBuckets(),
```
jonathan-innis, line 81: Our `pod_startup_duration_seconds` doesn't have any additional labels -- it's also not a histogram like it should be. At this point, we already marked it as stable so we can't change it from a summary metric. We can consider marking it as deprecated and replace it with a histogram metric called `karpenter_pods_start_duration_seconds`.

The other option: According to [the docs](https://prometheus.io/docs/concepts/metric_types/), the only difference between the summary metric and the histogram metric is the addition of the `_bucket` metric vs a metric on the basename that has the label (`quantile`).  That means that we could have them existing simultaneously as a histogram and summary metric -- we should look into if this is possible to avoid the breaking change/rename

pkg/controllers/metrics/pod/controller.go
```diff
@@ -133,21 +165,65 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+			podName:      pod.Name,
+			podNameSpace: pod.Namespace,
+		}).Set(time.Since(pod.CreationTimestamp.Time).Seconds())
 		c.pendingPods.Insert(key)
 		return
 	}
 	cond, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
 		return c.Type == corev1.PodReady
 	})
 	if c.pendingPods.Has(key) && ok {
```
jonathan-innis, line 185: Is it correct to not emit the metric when we can't find the ready status? Or should we treat the lack of ready status here as unknown?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -133,21 +165,65 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+}
+func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, labels prometheus.Labels) {
+	key := client.ObjectKeyFromObject(pod).String()
+	condScheduled, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
+		return c.Type == corev1.PodScheduled
+	})
+	if pod.Status.Phase == phasePending {
+		// The podsScheduled condition may be True or Unknown while a pod is in the pending state. Only when the pod is not bound,
+		// as shown by the PodScheduled condition not being set to true, do we wish to emit the pod_current_unbound_time_seconds metric.
+		if ok && condScheduled.Status != corev1.ConditionTrue {
```
jonathan-innis, line 210: Same comment here. Is it correct to treat the lack of ready status as something that we shouldn't emit a metric on?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -133,21 +165,65 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
+		}
+	}
+}
+func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, labels prometheus.Labels) {
+	key := client.ObjectKeyFromObject(pod).String()
+	condScheduled, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
+		return c.Type == corev1.PodScheduled
+	})
+	if pod.Status.Phase == phasePending {
+		// The podsScheduled condition may be True or Unknown while a pod is in the pending state. Only when the pod is not bound,
```
jonathan-innis, line 208: I think we just want this metric to increase from the pod's creation timestamp. I don't think that we want to emit the difference between it being scheduled and it being bound. Ideally, we're just looking to see how long it took for Karpenter to take a pod from created to bound

pkg/controllers/metrics/pod/controller.go
```diff
@@ -133,21 +165,65 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		},
 	})
 	c.recordPodStartupMetric(pod)
+	c.recordPodBoundMetric(pod, labels)
```
jonathan-innis, line 168: We aren't ruling out pods that Karpenter doesn't think can schedule. As it stands right now, you're going to have metrics that are going to increase forever because Karpenter will never be able to schedule them. Can we build this so that we filter out pods that Karpenter thinks it has no chance at getting to?

pkg/controllers/state/cluster.go
```diff
@@ -82,6 +83,17 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
 //
 //nolint:gocyclo
 func (c *Cluster) Synced(ctx context.Context) (synced bool) {
+	if c.unsyncedState.IsZero() {
```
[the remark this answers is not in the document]
jigisha620: But we only want to set this unsyncedState one time and then measure the time difference between unsyncedState and syncedState. Isn't that what this is? If I set it every time when this function returns false then the time difference is always going to be zero.
```
defer func() {
		if synced {
			ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(0)
		} else {
			c.unsyncedStartTime = c.clock.Now()
			ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(time.Since(c.unsyncedStartTime).Seconds())
		}
	}()
```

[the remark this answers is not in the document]
jonathan-innis: Seems like you could just move the check in the `defer` statement. I think what confused me was the fact that these things were placed in the code separately

pkg/controllers/state/suite_test.go
```diff
@@ -1241,6 +1245,9 @@ var _ = Describe("Cluster State Sync", func() {
 			ExpectReconcileSucceeded(ctx, nodeClaimController, client.ObjectKeyFromObject(nodeClaim))
 		}
 		Expect(cluster.Synced(ctx)).To(BeFalse())
+		metric, found := FindMetricWithLabelValues("karpenter_cluster_state_unsynced_time_seconds", map[string]string{})
```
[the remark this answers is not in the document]
jigisha620: I think this would work because we also reset this metric after each test. So even if it succeeded in previous iteration, the values would get cleaned up.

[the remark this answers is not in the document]
jigisha620: Updated it to use fake clock

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -180,6 +180,11 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 				},
 			},
 		}); err != nil {
+		var apiStatus apierrors.APIStatus
+		if errors.As(err, &apiStatus) {
+			code := apiStatus.Status().Code
+			NodesEvictionRequestsTotal.With(map[string]string{CodeLabel: fmt.Sprintf("%d", code)}).Inc()
```
jonathan-innis, line 186:
```suggestion
			NodesEvictionRequestsTotal.With(map[string]string{CodeLabel: fmt.Sprint(code)}).Inc()
```

nit

pkg/controllers/state/cluster.go
```diff
@@ -82,6 +83,17 @@ func NewCluster(clk clock.Clock, client client.Client) *Cluster {
 //
 //nolint:gocyclo
 func (c *Cluster) Synced(ctx context.Context) (synced bool) {
+	// Set the metric depending on the result of the Synced() call
+	defer func() {
+		if c.unsyncedStartTime.IsZero() {
+			c.unsyncedStartTime = c.clock.Now()
+		}
+		if synced {
+			ClusterStateUnsyncedTimeSeconds.With(map[string]string{}).Set(0)
```
jonathan-innis, line 92: This doesn't reset when the value gets synced. I think you need to set `c.unsyncedStartTime` to zero whenever you go synced

jigisha620: I misunderstood what we wanted to set to 0 earlier. Fixed this.

jonathan-innis (approved):
/lgtm
/approve

PR #1673, opened 2024-09-17
feat: Add VMReserved to InstanceTypeOverhead

jukie:
Relates:
- https://github.com/aws/karpenter-provider-aws/issues/5161
- https://github.com/aws/karpenter-provider-aws/pull/7004

**Description**
The underlying OS or VM hypervisor consumes additional resources that should be tracked as overhead. In the linked PR this overhead is being subtracted from the overall capacity but it may be preferable to instead add them here.


jukie:
Thinking more it probably makes sense to keep this within overall capacity measurements

PR #1683, opened 2024-09-17
feat: implement graceful error handling in operator package

danxwang22:
Fixes #N/A 

**Description**
This implements fail open logic within the operator package 


pkg/operator/operator.go
```diff
@@ -225,3 +198,66 @@ func (o *Operator) Start(ctx context.Context) {
 	}()
 	wg.Wait()
 }
+
+func setupIndexers(ctx context.Context, mgr manager.Manager) {
+	err := mgr.GetFieldIndexer().IndexField(ctx, &corev1.Pod{}, "spec.nodeName", func(o client.Object) []string {
+		return []string{o.(*corev1.Pod).Spec.NodeName}
+	})
+	if err != nil {
+		log.FromContext(ctx).Error(err, "failed to setup pod indexer")
```
jonathan-innis, line 207: The one question that I have from this: Is this going to cause weird failure states that we don't expect? Like, is it possible that we fail to setup the indexer due to some weird transient error and then we are in this weird half-state that we can't recover from?

We should probably look into the indexer code and see the scenarios where it can return an error. If it's just when the object isn't found, this is probably fine. 

It might also be good to use other projects as reference-points here to see what they do on top of what sig-apimachinery recommends here.

pkg/apis/crds/karpenter.sh_nodepools.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.2
```
jonathan-innis, line 6: I think these may be updated now so you may just need to rebase this PR based off of the current main

njtran (commented):
/lgtm
/approve

Issue #1674, opened 2024-09-17
Pod eviction will cause service interruption


andyblog:
### Description

**Observed Behavior**:

When all replicas of a Deployment are on the same node, for example, a deployment has 2 pods on this node, and the 2 pods are evicted when the node is terminated. From the time the 2 pods are evicted to the time the 2 pods are created and run successfully on the new node, the deployment has no pods to provide services. 
This also happens when a Deployment has only one replica.

**Expected Behavior**:

During Evicting, a judgment will be made here. If all replicas of the Deployment are on this node, or the Deployment has only one replica, restarting the Deployment is more elegant than evicting. This operation will first create a pod on the new node, wait for the new pod to run successfully, and then terminate the old pod, which will reduce service interruption time.

**Reproduction Steps** (Please include YAML):

**Versions**:
- Chart Version:
- Kubernetes Version (`kubectl version`):

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

njtran:
> This operation will first create a pod on the new node, wait for the new pod to run successfully, and then terminate the old pod, which will reduce service interruption time.

As I understand it, this is how it currently works. Can you show reproduction steps to show what you're talking about?

andyblog:
The current working method is:
1. Spot instance or ondemand instance is terminated for various reasons
2. Node starts to be deleted and finalizer method starts to be executed
3. Pods on the current node start to be evicted

I think when all replicas of the deployment are on this node, restarting is more elegant than evicting, because the service will not be interrupted during the restart

PR #1685, opened 2024-09-18
fix: when all replicas of a deployment are on one node, restart the deployment instead of evicting it

andyblog:
Fixes   [1674](https://github.com/kubernetes-sigs/karpenter/issues/1674)

**Description**
When all replicas of a Deployment are on the same node, for example, a deployment has 2 pods on this node, and the 2 pods are evicted when the node is terminated. From the time the 2 pods are evicted to the time the 2 pods are created and run successfully on the new node, the deployment has no pods to provide services.
This also happens when a Deployment has only one replica.

During Evicting, a judgment will be made here. If all replicas of the Deployment are on this node, or the Deployment has only one replica, restarting the Deployment is more elegant than evicting. This operation will first create a pod on the new node, wait for the new pod to run successfully, and then terminate the old pod, which will reduce service interruption time.

- make presubmit

### DIFF
--- a/pkg/controllers/node/termination/terminator/terminator.go
+++ b/pkg/controllers/node/termination/terminator/terminator.go
@@ -19,9 +19,11 @@ package terminator
 import (
 	"context"
 	"fmt"
+	"sync"
 	"time"
 
 	"github.com/samber/lo"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -29,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	terminatorevents "sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator/events"
 	"sigs.k8s.io/karpenter/pkg/events"
 	nodeutil "sigs.k8s.io/karpenter/pkg/utils/node"
@@ -36,18 +39,21 @@ import (
 )
 
 type Terminator struct {
-	clock         clock.Clock
-	kubeClient    client.Client
-	evictionQueue *Queue
-	recorder      events.Recorder
+	sync.RWMutex
+	clock                  clock.Clock
+	kubeClient             client.Client
+	nodeRestartDeployments map[string]map[string]struct{}
+	evictionQueue          *Queue
+	recorder               events.Recorder
 }
 
 func NewTerminator(clk clock.Clock, kubeClient client.Client, eq *Queue, recorder events.Recorder) *Terminator {
 	return &Terminator{
-		clock:         clk,
-		kubeClient:    kubeClient,
-		evictionQueue: eq,
-		recorder:      recorder,
+		clock:                  clk,
+		kubeClient:             kubeClient,
+		nodeRestartDeployments: make(map[string]map[string]struct{}),
+		evictionQueue:          eq,
+		recorder:               recorder,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jwcesign:
I like this idea

jwcesign:
One of the drawbacks is:
This generates a lot of replicasets, and if you want to roll back the specific version by replicasets, you may not be able to do it.

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -175,3 +195,146 @@ func (t *Terminator) podDeleteTimeWithGracePeriod(nodeGracePeriodExpirationTime
...
+}
+
+func (t *Terminator) RestartDeployments(ctx context.Context, deployments []*appsv1.Deployment, nodeName string) error {
+	var updateErrors []error
+
+	for _, deployment := range deployments {
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = make(map[string]string)
+		}
+		restartedNode, exists := deployment.Spec.Template.Annotations["kubectl.kubernetes.io/restartedNode"]
```
jwcesign, line 251: Why did we choose the name `kubectl.kubernetes.io/restartedNode` instead of something like x.karpenter.sh/xxx?

andyblog: It is OK to use "karpenter.sh/restartedNode". I refer to the restart method of k8s. It is more reasonable to change it to karpenter style here. Can you help review and see which parts are unreasonable and need to be modified? I will modify it together

jwcesign: Sure, I will take a look asap.

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -175,3 +207,145 @@ func (t *Terminator) podDeleteTimeWithGracePeriod(nodeGracePeriodExpirationTime
...
+
+func (t *Terminator) RestartDeployments(ctx context.Context, deployments []*appsv1.Deployment, nodeName string) error {
+	var updateErrors []error
+
+	for _, deployment := range deployments {
+		if deployment.Spec.Template.Annotations == nil {
+			deployment.Spec.Template.Annotations = make(map[string]string)
+		}
+		restartedNode, exists := deployment.Spec.Template.Annotations[v1.DeploymentRestartNodeAnnotationKey]
+		if exists && restartedNode == nodeName {
```
jwcesign, line 264: Why skip if `restartedNode == nodeName`? And other annotation value is better, like `true`?

jwcesign: Why not take the annotation from `deployment.Annotations`?

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -175,3 +207,145 @@ func (t *Terminator) podDeleteTimeWithGracePeriod(nodeGracePeriodExpirationTime
...
+		}
+	}
+
+	for _, pod := range pods {
+		deployment := deploymentCache[pod.Namespace+"/"+pod.Name]
+
+		if deployment != nil {
+			key := deployment.Namespace + "/" + deployment.Name
+			if nodeDeploymentReplicas[key] >= *deployment.Spec.Replicas {
+				// If a deployment has multiple pods on this node, there will be multiple deployments here, and deduplication is required.
```
jwcesign, line 313: If it's in the scaling-up process, this checking maybe wrong. how about check the ready replicas?

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -175,3 +207,145 @@ func (t *Terminator) podDeleteTimeWithGracePeriod(nodeGracePeriodExpirationTime
...
+					uniqueDeployments[key] = struct{}{}
+					restartDeployments = append(restartDeployments, deployment)
+				}
+				continue
+			} else {
+				// If a deployment has multiple copies, all of which are on this node,
+				// when the restart begins, the number of copies of the deployment on this node will gradually decrease.
+				// This situation needs to be judged separately.
+				t.RLock()
+				_, exists := t.nodeRestartDeployments[nodeName][key]
```
jwcesign, line 324: Managing nodeRestartDeployments involves quite a bit of code. Maybe it's time for a refactor?

jwcesign (commented):
Other lgtm

pkg/controllers/node/termination/terminator/terminator.go
andyblog: Updating deployment.Spec.Template.Annotations can restart deployment, but updating deployment.Annotations will not cause deployment to restart

Here deployment.Spec.Template.Annotations is used as a sign of whether deployment has been restarted. If annotation already has nodeName, it means that the last Reconcile has restarted deployment, so it will be skipped this time

After testing, it can be implemented. Can you help to see if it is suitable?

njtran:
I'm also wondering if this is better suited for a standardized upstream termination flow, rather than implementing this behavior on Karpenter. Effectively we'll be doing a lot more checks, and querying pods/deployments across the cluster for every pod we're terminating, but only doing something different in a smaller subset of cases, where all pods in the deployment live on the cluster. This is nice to let the deployment controller handle up rollouts, but i'm not sure if the juice is worth the squeeze here. 

If you feel confident about the change here and the drawbacks, can you write up an RFC detailing the trade-offs so I can better understand?

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -175,3 +207,145 @@ func (t *Terminator) podDeleteTimeWithGracePeriod(nodeGracePeriodExpirationTime
...
+
+		t.Lock()
+		if t.nodeRestartDeployments[nodeName] == nil {
+			t.nodeRestartDeployments[nodeName] = make(map[string]struct{})
+		}
+		t.nodeRestartDeployments[nodeName][deployment.Namespace+"/"+deployment.Name] = struct{}{}
+		t.Unlock()
+
+		deployment.Spec.Template.Annotations[v1.DeploymentRestartNodeAnnotationKey] = nodeName
+		if err := t.kubeClient.Update(ctx, deployment); err != nil {
```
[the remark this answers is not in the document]
andyblog: Indeed, restarting the deployment operation will make the karpenter have too much permissions.
Is the upstream support "surge" eviction feature available in a short period of time?

andyblog:
This situation should be more common in non-production environments, such as test environments and pre-release environments, which are usually functional verifications. Most services are single-copy. If the service interruption time can be minimized when the node is terminated, the R&D experience will be better.

Currently, there seems to be no better way to solve this problem except restarting the deployment. If consider making it a feature and turning it off by default?

Our company is currently encountering this problem, and other companies should also encounter the same problem.

I will try to write an RFC, and we will discuss the advantages and disadvantages, as well as the feasibility.

Cristobal-M:
I'm preparing a cluster for dev/staging and facing this same issue. We have deployments with `replicas: 1` and if the only pod gets evicted = downtime for service. I am not an expert and maybe what I am about to say is dumb.

If Karpenter should not alter deployments/replicasets, maybe an alternative is to use pod-disruption-budget and have Karpenter just create an annotation on the node for that case: "I want to disrupt this node but I can't"

That would allow an external controller to do:
- Taint the node
- Find deployments responsible for the node pods
- Restart such deployments

They would not be scheduled to the same node, if the new pods are Ready, the old ones get deleted, PDB is respected and Karpenter would be able to disrupt it the next check.

I guess that would have minimal changes for this project. It could be something implemented by the user from a manifest in karpenter documentation, referencing another github project.

PR #1689, opened 2024-09-19
feat: Implement Migration Controller

rschalo:
Fixes #N/A 

**Description**


pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+)
+
+type Controller[T client.Object] struct {
+	kubeClient client.Client
+}
+
+func NewController[T client.Object](client client.Client) *Controller[T] {
```
njtran, line 45: Let's add some comments about why we need this controller and at a high level what it does

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// update annotations on all CRs
+	for _, item := range list.Items {
+		if lo.HasKey(item.GetAnnotations(), v1.StoredVersionMigrated) {
+			continue
+		}
+		item.SetAnnotations(lo.Assign(item.GetAnnotations(), map[string]string{
+			v1.StoredVersionMigrated: "true",
+		}))
+		if err := c.kubeClient.Update(ctx, &item); err != nil {
```
njtran, line 70: Can't we do a strategic merge from?

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// To get here, there shouldn't be any resources missing the annotation so filter for and update CRD
+	crdList := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := c.kubeClient.List(ctx, crdList); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting crds, %w", err)
+	}
+	// should filter tighter than this
+	crds := lo.Filter(crdList.Items, func(crd apiextensionsv1.CustomResourceDefinition, _ int) bool {
+		return strings.Contains(crd.ObjectMeta.Name, strings.ToLower(object.GVK(o).Kind))
```
njtran, line 82: We can't be exact about this? Why can't we ensure that the names are exactly the same?

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+	o := object.New[T]()
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(object.GVK(o))
+	if err := c.kubeClient.List(ctx, list); err != nil {
+		return reconcile.Result{}, fmt.Errorf("listing objects, %w", err)
+	}
+
+	// update annotations on all CRs
+	for _, item := range list.Items {
+		if lo.HasKey(item.GetAnnotations(), v1.StoredVersionMigrated) {
```
njtran, line 64: don't we want to assert that storedversion migrated is true? Could really go either way on this.

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+			continue
+		}
+		item.SetAnnotations(lo.Assign(item.GetAnnotations(), map[string]string{
+			v1.StoredVersionMigrated: "true",
+		}))
+		if err := c.kubeClient.Update(ctx, &item); err != nil {
+			return reconcile.Result{}, fmt.Errorf("annotating: %w", err)
```
njtran, line 71: It's more convention to make it one continuous sentence with a bunch of comments. Worth putting in the gvk too though
```suggestion
			return reconcile.Result{}, fmt.Errorf("adding storage migration annotation %s, %w", insertGVKcode, err)
```

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+
+	// To get here, there shouldn't be any resources missing the annotation so filter for and update CRD
+	crdList := &apiextensionsv1.CustomResourceDefinitionList{}
+	if err := c.kubeClient.List(ctx, crdList); err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting crds, %w", err)
+	}
```
njtran, line 79: Don't we only want to get the specific CRDs that we own? Why do we need to get all of the CRDs?

njtran: you might be able to use reflection to get this

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+	// should filter tighter than this
+	crds := lo.Filter(crdList.Items, func(crd apiextensionsv1.CustomResourceDefinition, _ int) bool {
+		return strings.Contains(crd.ObjectMeta.Name, strings.ToLower(object.GVK(o).Kind))
+	})
+	for _, crd := range crds {
+		stored := crd.DeepCopy()
+		crd.Status.StoredVersions = []string{"v1"}
+		if err := c.kubeClient.Status().Patch(ctx, &crd, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			return reconcile.Result{}, fmt.Errorf("updating crd: %w", err)
```
njtran, line 88:
```suggestion
			return reconcile.Result{}, fmt.Errorf("updating crd, %w", err)
```

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
+	o := object.New[T]()
+	return controllerruntime.NewControllerManagedBy(m).
+		Named(fmt.Sprintf("migration.%s", strings.ToLower(reflect.TypeOf(o).Elem().Name()))).
+		For(o).
+		Watches(&apiextensionsv1.CustomResourceDefinition{}, CRDEventHandler(c.kubeClient)).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		Complete(singleton.AsReconciler(c))
```
njtran, line 102: nice!

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func CRDEventHandler(kubeClient client.Client) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(ctx context.Context, o client.Object) (requests []reconcile.Request) {
+		crdList := &apiextensionsv1.CustomResourceDefinitionList{}
+		if err := kubeClient.List(ctx, crdList); err != nil {
+			return requests
```
jmdeal, line 109: nit: can we just return `nil` here and drop the named return variable? We lean away from these in general, and it's not being used in the final return anyway.
```suggestion
			return nil
```

pkg/controllers/migration/suite_test.go
njtran, line 1: are there any tests we need to add to make sure that the crd has the annotation? Or is this in a later PR in the provider?

pkg/controllers/migration/controller.go
rschalo: My thought is presence/absence of the annotation is sufficient to indicate migration.

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller[T]) Register(_ context.Context, m manager.Manager) error {
+	o := object.New[T]()
+	return controllerruntime.NewControllerManagedBy(m).
+		Named(fmt.Sprintf("migration.%s", strings.ToLower(reflect.TypeOf(o).Elem().Name()))).
+		For(o).
```
jmdeal, line 99: Are we reconciling against the CRs or the CRDs? This `For` sets us to reconcile against the CRs, but the following `Watches` and the structure of the reconcile loop itself suggests we're reconciling against CRDs. The latter seems correct to me.

jmdeal: Seems like we could implement this in a single controller rather than 3 typed controllers. We could use an EventFilter to only look at CRDs which are part of one of our groups, and have a v1 storage version.

jmdeal (commented):
checkpointing

pkg/controllers/migration/controller.go
njtran: Sounds good to me

pkg/controllers/migration/controller.go
rschalo: Only issue is the NodeClass, right? Also, that introduces the need for a bunch of filtering for multiple CRs and then updating the CRD for each based on the state of CRs.

rschalo: IIUC this watches the CRDs so that it can list it from the shared informer and not as a hard list.

pkg/controllers/migration/controller.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+func NewController[T client.Object](client client.Client) *Controller[T] {
+	return &Controller[T]{
+		kubeClient: client,
+	}
+}
+
+func (c *Controller[T]) Reconcile(ctx context.Context) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "migration")
```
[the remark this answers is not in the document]
rschalo: Done.

pkg/controllers/migration/controller.go
rschalo: Done.

pkg/controllers/migration/controller.go
rschalo: We can patch but cannot use strategic merge from with CRDs: https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility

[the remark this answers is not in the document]
rschalo: Do we need the optimistic lock if we're altering a map with a key we own?

pkg/controllers/migration/resource/controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	o := object.New[T]()
+	ctx = injection.WithControllerName(ctx, fmt.Sprintf("migration.resource.%s", strings.ToLower(object.GVK(o).Kind)))
+
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, o); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("getting object %v, %w", req.Name, err)
```
njtran, line 64: nit
```suggestion
		return reconcile.Result{}, fmt.Errorf("getting %v, %w", req.Name, err)
```

pkg/test/environment.go
```diff
@@ -89,7 +89,7 @@ func NewEnvironment(options ...option.Function[EnvironmentOptions]) *Environment
 	ctx, cancel := context.WithCancel(context.Background())
 
 	os.Setenv(system.NamespaceEnvKey, "default")
-	version := version.MustParseSemantic(strings.Replace(env.WithDefaultString("K8S_VERSION", "1.30.x"), ".x", ".0", -1))
+	version := version.MustParseGeneric(strings.Replace(env.WithDefaultString("K8S_VERSION", "1.30.x"), ".x", ".0", -1))
```
njtran, line 92: intentional? What's the implication of this change

pkg/controllers/migration/crd/controller.go
```diff
@@ -0,0 +1,116 @@
...
+			Kind:    crd.Spec.Names.Kind,
+		},
+	)
+	if err := c.kubeClient.List(ctx, list); err != nil {
+		return reconcile.Result{}, fmt.Errorf("listing resources for %v, %w", crd.Spec.Names.Kind, err)
+	}
+	for _, item := range list.Items {
+		// requeue if a custom resource is found without annotation
+		if !lo.HasKey(item.GetAnnotations(), v1.StoredVersionMigrated) {
+			return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
```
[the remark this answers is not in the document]
rschalo: The intent is to let the other controller do it.

pkg/controllers/migration/crd/controller.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/*
+however many concurrent reconciles
+any change to CR re-reconciles, which then lists all CRs
+
+3 Get CR controllers to add annotation -> all resources annotated
+1 CRD controller filter on group (karpenter.sh) and cp.GetSupported
+
+*/
```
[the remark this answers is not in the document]
rschalo: Nope definitely not. Will remove.

pkg/test/environment.go
rschalo: This is something I've run into while doing local dev pretty often. I meant to remove for this change and will revisit.

pkg/controllers/migration/crd/controller.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+	// do nothing for CRDs that have dropped v1beta1
+	if !lo.Contains(crd.Status.StoredVersions, "v1beta1") {
+		return reconcile.Result{}, nil
+	}
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(
+		schema.GroupVersionKind{
+			Group:   crd.Spec.Group,
+			Version: "v1",
```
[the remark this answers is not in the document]
rschalo: So, the webhook is present as part of migration and then new objects should be stored as v1 after the annotation write to make it final.

pkg/apis/v1/nodepool_conversion.go
```diff
@@ -46,6 +46,7 @@ func (in *NodePool) ConvertTo(_ context.Context, to apis.Convertible) error {
 	// Remove the annotations from the v1beta1 NodeClaim on the convert back
 	delete(v1beta1NP.Annotations, KubeletCompatibilityAnnotationKey)
 	delete(v1beta1NP.Annotations, NodeClassReferenceAnnotationKey)
+	delete(v1beta1NP.Annotations, StoredVersionMigratedKey)
```
njtran, line 49: Can you add a comment on why we do this just so we don't forget in the future if we need

pkg/apis/v1/nodeclaim_conversion.go
```diff
@@ -45,6 +45,7 @@ func (in *NodeClaim) ConvertTo(ctx context.Context, to apis.Convertible) error {
 	// Remove the annotations from the v1beta1 NodeClaim on the convert back
 	delete(v1beta1NC.Annotations, KubeletCompatibilityAnnotationKey)
 	delete(v1beta1NC.Annotations, NodeClassReferenceAnnotationKey)
+	delete(v1beta1NC.Annotations, StoredVersionMigratedKey)
```
njtran, line 48: same for this

pkg/controllers/migration/crd/controller.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, item := range list.Items {
+		// requeue if a custom resource is found without annotation
+		if !lo.HasKey(item.GetAnnotations(), v1.StoredVersionMigratedKey) {
+			return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
+		}
+	}
+	// if all custom resources have been updated, patch the CRD
+	stored := crd.DeepCopy()
+	crd.Status.StoredVersions = []string{"v1"}
+	if err := c.kubeClient.Status().Patch(ctx, crd, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
```
njtran, line 88: Have we tried strategic merges? I know we can't do strategic merge on the CR itself, and I assume this extends to the CRD, but haven't tried.

pkg/apis/v1/nodeclaim_conversion.go
njtran, line 1: can we add tests for the conversions?

pkg/controllers/migration/resource/controller.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	o := object.New[T]()
+	ctx = injection.WithControllerName(ctx, fmt.Sprintf("migration.resource.%s", strings.ToLower(object.GVK(o).Kind)))
+
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, o); err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("getting %v, %w", req.Name, err)
```
njtran, line 64: This would just print the name of the resource, right? not the type. Probably want to print the type here

pkg/controllers/migration/resource/controller.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	stored := o.DeepCopyObject()
+	o.SetAnnotations(lo.Assign(o.GetAnnotations(), map[string]string{
+		v1.StoredVersionMigratedKey: "true",
+	}))
+	// update annotation on CR
+	if !equality.Semantic.DeepEqual(stored, o) {
+		if err := c.kubeClient.Patch(ctx, o, client.MergeFromWithOptions(stored.(client.Object), client.MergeFromWithOptimisticLock{})); err != nil {
+			return reconcile.Result{}, fmt.Errorf("adding stored version migration annotation %v, %w", req.Name, err)
```
njtran, line 73: same here on type vs name. Worth just typing the migration annotation as it appears in the annotations too

pkg/controllers/migration/crd/controller.go
rschalo: Done.

njtran (commented):
/lgtm
/approve

Issue #705, opened 2023-04-14
Unschedulable Pod Metric


sidewinder12s:
### Tell us about your request

We used this metric with the cluster_autoscaler `cluster_autoscaler_unschedulable_pods_count`. 

Would it be possible to have Karpenter expose a similar metric?

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

How to alert when Karpenter is unable to scale out

### Are you currently working around this issue?

Looking at Kubernetes Scheduler metric `scheduler_pending_pods`, however if you have multiple schedulers this gets more complicated. 

EKS also does not expose metrics from the Kubernetes Scheduler. 

### Additional Context

We might have been abusing this metric a bit as the OSS Autoscaler would also observe entire cluster state. 

We have 2 Autoscalers running (CAS and an Internal autoscaler), so we were somewhat overloading on this metric since CAS still took a whole cluster view even if it wasn't scaling most of our nodes. 

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Horiodino:
Hii, how about I work on it!

Bryce-Soghigian:
@njtran many users with large clusters probably will want to avoid scraping the high cardinality series of kube_pod*

all the kube_pod* are rather high cardinality metrics and aren't as scalable as the CAS metric imo.

jonathan-innis:
@Horiodino Assigned!

jonathan-innis:
Seems like this one might be difficult for the same reason that #686 is difficult. The second you get multiple NodePools into the mix, we now have to figure out which NodePool we think you _intended_ to schedule the pod to; otherwise, we might fire multiple metrics, one for each NodePool. There's basically a 1-1 reasoning between why the pod wouldn't have scheduled to a NodePool with Karpenter. Maybe this means what we are really looking for here is a pod metric with a nodepool label and pod label that tells us the reason why the pod didn't schedule against each NodePool.

I still haven't looked into how complex that might be to orchestrate or how much cardinality that would add to pod metrics that already have pretty high cardinality to begin with.

njtran:
[...]

PR #1698, opened 2024-09-19
Implemented UnschedulablePodsCount metric

edibble21:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/705

**Description**
Implemented UnschedulablePodsCount metric for Scheduler

implemented tests
make presubmit


pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3679,9 +3680,45 @@ var _ = Context("Scheduling", func() {
...
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go func() {
+				defer GinkgoRecover()
+				defer wg.Done()
+				Eventually(func(g Gomega) {
+					m, ok := FindMetricWithLabelValues("karpenter_scheduler_unschedulable_pods_count", map[string]string{"controller": "provisioner"})
+					g.Expect(ok).To(BeTrue())
+					g.Expect(lo.FromPtr(m.Gauge.Value)).To(BeNumerically(">=", 0))
+				}, time.Second).Should(Succeed())
```
njtran, line 3716: Looks like this is racey. Any reason we can't make this 10 seconds?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3679,9 +3680,45 @@ var _ = Context("Scheduling", func() {
...
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go func() {
+				defer GinkgoRecover()
+				defer wg.Done()
+				Eventually(func(g Gomega) {
+					m, ok := FindMetricWithLabelValues("karpenter_scheduler_unschedulable_pods_count", map[string]string{"controller": "provisioner"})
+					g.Expect(ok).To(BeTrue())
+					g.Expect(lo.FromPtr(m.Gauge.Value)).To(BeNumerically(">=", 0))
+				}, 10*time.Second).Should(Succeed())
```
rschalo, line 3716: Does this just test for the existence of the metric or do we expect that it should be greater than 0?

edibble21: If I am understanding correctly the metric should be 0 or greater because 0 would just mean that all pods were schedulable. Is that worth testing or should I just remove this check.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -191,10 +191,10 @@ func (r Results) TruncateInstanceTypes(maxInstanceTypes int) Results {
 		}
 	}
 	r.NewNodeClaims = validNewNodeClaims
-	return r
+
```
jmdeal, line 194: nit
```suggestion
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -204,7 +204,8 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 	// had 5xA pods and 5xB pods were they have a zonal topology spread, but A can only go in one zone and B in another.
 	// We need to schedule them alternating, A, B, A, B, .... and this solution also solves that as well.
 	errors := map[*corev1.Pod]error{}
-	QueueDepth.DeletePartialMatch(prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx)}) // Reset the metric for the controller, so we don't keep old ids around
+	UnschedulablePodsCount.DeletePartialMatch(prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx)}) // Reset the metric for the controller, so we don't keep old ids around
+	QueueDepth.DeletePartialMatch(prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx)})             // Reset the metric for the controller, so we don't keep old ids around
```
jmdeal, line 208: super-nit: let's consolidate the comment for the two metrics
```suggestion
    // Reset the metric for the controller, so we don't keep old scheduler-ids around
	UnschedulablePodsCount.DeletePartialMatch(prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx)})
	QueueDepth.DeletePartialMatch(prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx)})
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -240,11 +241,16 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 			delete(errors, k)
 		}
 	}
-	return Results{
+	var results = Results{
 		NewNodeClaims: s.newNodeClaims,
 		ExistingNodes: s.existingNodes,
 		PodErrors:     errors,
 	}
+	results.TruncateInstanceTypes(MaxInstanceTypes)
```
jmdeal, line 249: nit: prefer walrus assignment
```suggestion
    results := Results{
		NewNodeClaims: s.newNodeClaims,
		ExistingNodes: s.existingNodes,
		PodErrors:     errors,
	}
	results.TruncateInstanceTypes(MaxInstanceTypes)
```

pkg/controllers/provisioning/scheduling/suite_test.go
jmdeal: You're correct for this test, but it would be good to either adapt this test or add an additional test for pods we know are unschedulable. The most straightforward way to do this would be to create a deployment with some unfulfillable (anti-)affinity terms.

kwok/apis/crds/karpenter.kwok.sh_kwoknodeclasses.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.2
```
jmdeal, line 6: I believe these changes should go away if you rebase.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3673,9 +3675,39 @@ var _ = Context("Scheduling", func() {
...
 				}, time.Second).Should(Succeed())
 			}()
-			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods)
+			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods, scheduling.MaxInstanceTypes)
 			wg.Wait()
 		})
+		It("should surface the UnschedulablePodsCount metric while executing the scheduling loop", func() {
+			nodePool := test.NodePool()
+			nodePool.Spec.Template.Spec.Requirements = []v1.NodeSelectorRequirementWithMinValues{
+				{NodeSelectorRequirement: corev1.NodeSelectorRequirement{Key: corev1.LabelInstanceTypeStable, Operator: corev1.NodeSelectorOpIn, Values: []string{"default-instance-type"}}}}
```
jmdeal, line 3684: nit: if you're creating a new test NodePool, you can set these properties inline through the `test.NodePool` function.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3673,9 +3675,39 @@ var _ = Context("Scheduling", func() {
...
+			ExpectApplied(ctx, env.Client, nodePool)
+			//Creates 15 pods, 5 schedulable and 10 unschedulable and adds and UID
```
jmdeal, line 3686: Could you elaborate on why you need to add the UID here?

jmdeal (commented):
Great job! Couple of small comments, and I'm going to test locally, but overall looks good.

pkg/controllers/provisioning/scheduling/suite_test.go
edibble21: When the pod fails to schedule it is pushed back onto the queue and the length of the queue at the time it is pushed is mapped to the pod via its UID. If you do not set it in the test the length is mapped to every pod and not just the pod that failed to schedule because they all have an empty UID. That is what caused the issue we were discussing Friday.

jmdeal: Can you leave something to that effect as a comment? It's not clear looking at the test why it's necessary.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -239,11 +240,16 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
 		}
 	}
-	return Results{
+	results := Results{
 		NewNodeClaims: s.newNodeClaims,
 		ExistingNodes: s.existingNodes,
 		PodErrors:     errors,
 	}
+	results.TruncateInstanceTypes(MaxInstanceTypes)
+	UnschedulablePodsCount.With(
```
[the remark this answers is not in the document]
edibble21: The reason I did not pull it out is because in provisioner I did not have access to the scheduler ID. This was a point of discussion because it seemed I could either pull it out and not have SchedulingIDLabel or I would need to change the existing interface.

edibble21 (commented):
/karpenter snapshot

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -183,7 +183,7 @@ func benchmarkScheduler(b *testing.B, instanceCount, podCount int) {
 	podsScheduledInRound1 := 0
 	nodesInRound1 := 0
 	for i := 0; i < b.N; i++ {
-		results := scheduler.Solve(ctx, pods)
+		results := scheduler.Solve(ctx, pods).TruncateInstanceTypes(scheduling.MaxInstanceTypes)
```
jonathan-innis, line 186: I could go either way -- but doesn't seem necessary to change this

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3673,9 +3674,56 @@ var _ = Context("Scheduling", func() {
 					g.Expect(lo.FromPtr(m.Gauge.Value)).To(BeNumerically(">", 0))
 				}, time.Second).Should(Succeed())
 			}()
-			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods)
+			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods).TruncateInstanceTypes(scheduling.MaxInstanceTypes)
```
jonathan-innis, line 3677: Do you need to change this?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3673,9 +3674,56 @@ var _ = Context("Scheduling", func() {
...
+			for _, i := range pods {
+				ExpectApplied(ctx, env.Client, i)
+			}
+			var wg sync.WaitGroup
+			wg.Add(1)
+			go func() {
+				defer GinkgoRecover()
+				defer wg.Done()
+				Eventually(func(g Gomega) {
+					m, ok := FindMetricWithLabelValues("karpenter_scheduler_unschedulable_pods_count", map[string]string{"controller": "provisioner"})
```
jonathan-innis, line 3717: Can you just check if this metric is set once after running the scheduling loop rather than checking if it's set while it's running? I'm betting you don't need an Eventually statement here and can just still with an Expect

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3696,7 +3744,7 @@ var _ = Context("Scheduling", func() {
 			}) // Create 1000 pods which should take long enough to schedule that we should be able to read the queueDepth metric with a value
 			s, err := prov.NewScheduler(ctx, pods, nil)
 			Expect(err).To(BeNil())
-			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods)
+			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods).TruncateInstanceTypes(scheduling.MaxInstanceTypes)
```
jonathan-innis, line 3747: Did you need to change this one too?

jonathan-innis (approved):
/lgtm
/approve

sidewinder12s:
Thanks @edibble21!

PR #1699, opened 2024-09-20
fix: Ignore released volumes when deleting a node

willthames:
Fixes #1684

**Description**

Ignore released volumes when waiting for volume attachments
    
When deciding whether a node can be terminated, ignore persistent
volumes that have already been released (perhaps attached to a pod
that has since been drained)

**How was this change tested?**

Using `make test`. This change is split into two commits, the first is a test that fails before the change in the second commit and succeeds afterwards

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1700, opened 2024-09-20
Only wait for volume attachments for drainable nodes

### LINKED ISSUE #1684: Drift replacement stuck due to "Cannot disrupt NodeClaim"
### Description

**Observed Behavior**:

Two nodes were replaced during drift replacement, next one seems stuck with

```
  Normal   DisruptionBlocked      19s (x29 over 57m)  karpenter  Cannot disrupt Node: state node is marked for deletion
```

There is nothing in karpenter's logs to explain this. We did see similar behaviour during the karpenter 1.0.1 upgrade but put that down to API version mismatches but we don't seem to have any such mismatches this time.

```
k get nodeclaims -o custom-columns='APIVER:.apiVersion,NAME:.metadata.name,OWNER_API_VER:.metadata.ownerReferences[0].apiVersion,OWNERKIND:.metadata.ownerReferences[0].kind'
APIVER            NAME                               OWNER_API_VER     OWNERKIND
karpenter.sh/v1   bottlerocket-general-amd64-6k89s   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-general-amd64-cmdfv   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-general-amd64-hw9p6   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-general-amd64-jjmnz   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-general-amd64-ml9q4   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-general-amd64-z7r8c   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-smaller-amd64-gc2nt   karpenter.sh/v1   NodePool
karpenter.sh/v1   bottlerocket-smaller-amd64-sr2lr   karpenter.sh/v1   NodePool
```

**Expected Behavior**:

All nodes get replaced during drift replacement

**Versions**:
- Chart Version: 1.0.2
```
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.30.3-eks-2f46c53
```

willthames:
Fixes #1684

**Description**

This solves two problems identified so far:
* waiting for a volume attachment associated with a persistent volume claim that has been released
* waiting for a volume attachment associated with a persistent volume claim that has assigned to a new pod that can't start until the association is removed

Effectively this changes the logic from finding
all the PVCs associated with non-drainable pods
and ignoring those ones, to finding all the PVCs
associated with drainable pods and only blocking
on those ones.

As such some of the previous tests are now redundant because they check we wait for volume associations that aren't tied to a PVC at all.

`make test`

### DIFF
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -28,6 +28,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/clock"
@@ -193,6 +194,15 @@ func (c *Controller) ensureVolumesDetached(ctx context.Context, node *corev1.Nod
 	return len(filteredVolumeAttachments) == 0, nil
 }
 
+// volumeIsNotAttachedOrBound returns true if the persistent volume storage is neither Attached nor Bound
+func volumeIsNotAttachedOrBound(ctx context.Context, kubeClient client.Client, pvName string) (bool, error) {
+	pv := &corev1.PersistentVolume{}
+	if err := kubeClient.Get(ctx, types.NamespacedName{Name: pvName}, pv); err != nil {
+		return false, fmt.Errorf("getting persistent volume %q, %w", pvName, err)
+	}
+	return pv.Status.Phase == corev1.VolumePending || pv.Status.Phase == corev1.VolumeReleased || pv.Status.Phase == corev1.VolumeFailed, nil
+}
+
 // filterVolumeAttachments filters out storagev1.VolumeAttachments that should not block the termination
 // of the passed corev1.Node
 func filterVolumeAttachments(ctx context.Context, kubeClient client.Client, node *corev1.Node, volumeAttachments []*storagev1.VolumeAttachment, clk clock.Clock) ([]*storagev1.VolumeAttachment, error) {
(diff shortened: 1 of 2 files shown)


willthames:
Note that this is an alternative, much more impactful alternative to #1699. 

There may be cases that I have not considered.

#1699 is a safer PR but doesn't protect against a second problem I found where a pod with a PVC associated with a volume attachment got moved to a new node (statefulset-like behaviour) and the new pod couldn't start because its persistent volume was still attached to the old node.

AndrewSirenko:
Hey @willthames thanks for looking into this, I'm sorry that you're running into this drift replacement issue. I'll take a deeper look on the PR Tuesday, but I see one main issue with the following statement: 

> Effectively this changes the logic from finding
all the PVCs associated with non-drainable pods
and ignoring those ones, to finding all the PVCs
associated with drainable pods and only blocking
on those ones.

The issue is that Pod resources can get cleaned up in Kubernetes BEFORE the associated volumes are unpublished from the node (unmounted and/or detached). This is because the pod lifecycle is not affected by the lifecycle of the PVCs it is associated with.   

Therefore if you block node termination by filtering in attachments of drainable pods instead of filtering out those of non-drainable pod, you are not guaranteed to block node termination until all volumes are detached (because the pod resources might be deleted by K8s). You can run into the following race condition: 

1. Node starts being drained by Karpenter
2. Pod enters terminating state
3. Pod is drained and pod resource is deleted
4. K8s + CSI Driver sees that volume must be unmounted, begins unmount (but not detach) 
5. Karpenter does not see this yet-to-be-detached volume when looping through node's pods and therefore proceeds with node termination
6. CSI Driver begins volume detach, but node is already terminating (holding volume hostage) 
7. When stateful workload is rescheduled on another node, it now cannot attach volume and start until previous node fully terminated (Which this blocking of node termination is trying to prevent) 

Nonetheless, VolumeAttachment resources should not linger once a PV is released, so I'm curious about that case. Is there anything special about these volumes? This could be a bug in Attach/Detach controller instead of Kubernetes. 

You can see this [Preventing 6+ minute delays from StatefulSet Disruption Request for Comment](https://github.com/aws/karpenter-provider-aws/pull/6336/files) for more information on why this blocking until volumes detached happens. I'm working on a up-to-date/cleaned-up version of this document which should be up at the end of the month.

willthames:
Hi @AndrewSirenko, thanks for your detailed response, definitely appreciate the extra context!

The scenario you describe is the opposite problem to the one I have (nodes disappearing before the volume attachment, causing problems, rather than nodes not being terminated because of the volume attachment), so there's definitely a conflict somewhere.

I'm only having this problem with one file system type as far as I can tell (EFS) but also with one pod type (couchbase, which is managed as pods by the couchbase controller (rather than a higher level abstraction such as statefulsets)). But the couchbase pods are long gone by the time the clean up needs to happen.

This problem persists for days, the only way I can get karpenter to complete drift termination is through deleting affected volume attachments (which is not an approach that is at all satisfying).

I'll definitely check for any outstanding efs controller bugs. 

The alternative PR to this one doesn't block on unattached PVCs, but that one does not clean up if the volume is then mounted onto another pod

willthames:
This PR is no longer required (I'll add an explanation in #1684)

PR #1701, opened 2024-09-20
chore: Bump operatorpkg

jigisha620:
Fixes #N/A 

**Description**
Bump operatorpkg version

### DIFF
--- a/pkg/controllers/nodepool/readiness/suite_test.go
+++ b/pkg/controllers/nodepool/readiness/suite_test.go
@@ -92,6 +92,7 @@ var _ = Describe("Readiness", func() {
 		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
 		_ = ExpectObjectReconciled(ctx, env.Client, controller, nodePool)
 		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		nodePool.StatusConditions().SetTrue(v1.ConditionTypeValidationSucceeded)
 		Expect(nodePool.StatusConditions().IsTrue(status.ConditionReady)).To(BeTrue())
 	})
 	It("should have status condition on nodePool as not ready if nodeClass is not ready", func() {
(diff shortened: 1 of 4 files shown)


pkg/controllers/nodepool/readiness/suite_test.go
```diff
@@ -88,12 +88,6 @@ var _ = Describe("Readiness", func() {
 		nodePool = ExpectExists(ctx, env.Client, nodePool)
 		Expect(nodePool.StatusConditions().Get(status.ConditionReady).IsFalse()).To(BeTrue())
 	})
-	It("should have status condition on nodePool as ready if nodeClass is ready", func() {
```
jigisha620, line 91: Since operatorpkg has now changes for observed generation, nodePool status condition will not go ready unless the dependents are also set to true within the reconciliation lopp.

jonathan-innis: IMO: We shouldn't drop this test just because of this. Can we set the dependents for our other sub-conditions to true so that this test passes?

pkg/controllers/nodepool/validation/suite_test.go
```diff
@@ -69,7 +69,6 @@ var _ = Describe("Counter", func() {
 		ExpectApplied(ctx, env.Client, nodePool)
 		ExpectObjectReconciled(ctx, env.Client, nodePoolValidationController, nodePool)
 		nodePool = ExpectExists(ctx, env.Client, nodePool)
-		Expect(nodePool.StatusConditions().IsTrue(status.ConditionReady)).To(BeTrue())
```
jonathan-innis, line 72: Why do we drop this? Can we hold the same testing line and just make changes to ensure that we can keep this around?

jonathan-innis (approved):
/lgtm
/approve

PR #1704, opened 2024-09-21
chore: Allow setting leader election namespace when running out of cluster

jonathan-innis:
Fixes #N/A 

**Description**

Leader election won't currently work if you are running Karpenter outside of the cluster. This allows a user to set `--leader-election-namespace` to run Karpenter outside of the cluster but still run it with leader election (this enables hub and spoke models -- like the ones that CAPI is proposing)

`make presubmit`


engedaam (commented):
/lgtm

PR #1715, opened 2024-09-25
fix: only patch CRDs when storage version is v1

jmdeal:
Fixes #N/A 

**Description**
Ensures that we only patch the stored versions on a CRD when the storage version is v1beta1. This prevents an edge-case where a user upgrading from a pre-v1 version, who doesn't have any resources created for a CRD, and who upgraded their controller before their CRDs.

`make test`

### DIFF
--- a/pkg/controllers/migration/crd/controller.go
+++ b/pkg/controllers/migration/crd/controller.go
@@ -26,6 +26,7 @@ import (
 	"github.com/samber/lo"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/migration/crd/controller.go
```diff
@@ -62,10 +62,20 @@ func (c *Controller) Reconcile(ctx context.Context, crd *apiextensionsv1.CustomR
...
 	// If v1beta1 is not stored, no-op
 	if !lo.Contains(crd.Status.StoredVersions, "v1beta1") {
 		return reconcile.Result{}, nil
 	}
+	// We only want to reconcile against CRDs that have v1 as the storage version.
+	// Note: we don't need to check this in the resource migration controller since the conversion webhook drops the
+	// annotation when applying the v1 resource with the annotation.
+	if storageVersion, found := lo.Find(crd.Spec.Versions, func(v apiextensionsv1.CustomResourceDefinitionVersion) bool {
+		return v.Storage
+	}); !found || storageVersion.Name != "v1" {
```
njtran, line 75: Should we add some sort of log here to let the user know that they haven't upgraded their CRD and that we expect it to be?

jmdeal: Good call, I'll add one 👍

njtran (commented):
/lgtm
/approve

PR #1716, opened 2024-09-26
fix: Fix webhook owner reference causing ArgoCD leak

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that:
1. Knative doesn't add ownerReferences to webhook configurations that it creates by adding `DisableNamespaceOwnership` to the `webhook.Options`
2. We remove old OwnerReferences that were already added by Knative onto the webhook configuration resources by patching out the namespace ownership that was added by Knative

`make presubmit`

- Ran this on an ArgoCD cluster where we started on v0.37.3, upgraded to this version and then upgraded to v1.0.3. This removed the webhook configurations and I validated that we removed the ownerReferences on this version


pkg/operator/operator.go
```diff
@@ -223,11 +228,27 @@ func (o *Operator) WithControllers(ctx context.Context, controllers ...controlle
...
-func (o *Operator) WithWebhooks(ctx context.Context, ctors ...knativeinjection.ControllerConstructor) *Operator {
+func (o *Operator) WithWebhooks(ctx context.Context, ctors ...knativeinjection.NamedControllerConstructor) *Operator {
 	if !options.FromContext(ctx).DisableWebhook {
 		o.webhooks = append(o.webhooks, ctors...)
 		lo.Must0(o.Manager.AddReadyzCheck("webhooks", webhooks.HealthProbe(ctx)))
 		lo.Must0(o.Manager.AddHealthzCheck("webhooks", webhooks.HealthProbe(ctx)))
+
+		kCtx, _ := knativeinjection.EnableInjectionOrDie(ctx, o.GetConfig())
+		kCtx = injection.WithClient(kCtx, o.GetClient())
+		kCtx = injection.WithNodeClasses(kCtx, []schema.GroupVersionKind{})
```
njtran, line 239: Why do we add in the nodeclasses key in here with an empty kind? Shouldn't this be done in the cloud provider provider?

pkg/controllers/webhookownerreference/controller.go
```diff
@@ -0,0 +1,83 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+)
+
+// Controller for the resource
+type Controller[T client.Object] struct {
+	kubeClient client.Client
+	name       string
```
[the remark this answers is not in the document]
jonathan-innis: It's the name -- it has to do with the name of the configuration object -- the kind is determined from the `client.Object` type

pkg/controllers/webhookownerreference/controller.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if obj.GetName() != c.name {
+		return reconcile.Result{}, nil
+	}
+	stored := obj.DeepCopyObject().(client.Object)
+	obj.SetOwnerReferences(lo.Reject(obj.GetOwnerReferences(), func(o metav1.OwnerReference, _ int) bool {
+		return o.APIVersion == "v1" && o.Kind == "Namespace" && o.Name == system.Namespace()
+	}))
+	if !equality.Semantic.DeepEqual(obj, stored) {
+		if err := c.kubeClient.Update(ctx, obj); err != nil {
```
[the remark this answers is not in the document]
jonathan-innis: It's because we don't have patch permission in our current role and I didn't see a reason to change the role just for this

pkg/operator/operator.go
```diff
@@ -223,11 +228,27 @@ func (o *Operator) WithControllers(ctx context.Context, controllers ...controlle
...
+
+		// Add webhook OwnerReference controller (to clean-up owner references added by knative)
+		// for any Validating or Mutating WebhookConfigurations
+		for _, ctor := range ctors {
```
njtran, line 243: Are we getting to a point where we need to think about the number of controllers we have? we're adding in 3 controllers, all each for one resource in the cluster. 

Also as I think more, does it make sense that we have this implemented as 3 genericized controllers? We're only caring about three specific resources in the cluster from a previous installation, right? Can we just do this on startup?

pkg/operator/operator.go
njtran: I'm confused though, shouldn't we add in the cp.GetSupportedNodeClasses() then?

jonathan-innis: I mean -- we could, bit it's not needed. What's needed is the key to be there -- what is stored in the key doesn't matter here for this use-case

pkg/operator/operator.go
jonathan-innis: Not sure if I'm following -- this really doesn't do much from a goroutine perspective once it's done with its task. This is also only present on the pre-v1.0 versions so we aren't carrying it forward -- also, you get leader election for free and don't have to deal with retry behavior like you would if you had to create your own thread to do the task for you

njtran (commented):
/lgtm
/approve

PR #1721, opened 2024-09-29
fix: handle NodeClaim termination in the lifecycle controller

jmdeal:
Fixes #N/A 

**Description**
Collapses the NodeClaim termination and lifecycle controllers. This eliminates a race where the termination controller may finalize a NodeClaim while the lifecycle controller is launching an instance, resulting in an instance leak. The garbage collection controller ensures that this leaked instance is cleaned up, but this prevents the leak from occurring in the first place.

`make presubmit` and E2Es in karpenter-provider-aws (TODO)


pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -74,11 +86,35 @@ func NewController(clk clock.Clock, kubeClient client.Client, cloudProvider clou
 	}
 }
 
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
```
jonathan-innis, line 89: How much of this Register function changed from the current Register function? What was added from the termination controller that wasn't in the lifecycle controller already?

jmdeal: The only change was the removal of the event filter for NodeClaims. Previously, we were only watching `Create` events. Since we already had the watch on Nodes, we only need this change to capture the events for both controllers.

jonathan-innis: Hmmm, I wonder why the diff was so large then. Makes sense if we are just adding watches on `Update`

jmdeal: Oh, probably just because I moved it above the Reconcile function so it didn't split up reconcile and finalize.

jonathan-innis (approved):
/lgtm
/approve

PR #1727, opened 2024-10-01
fix: Fix handling for multiple Daemonset nodeAffinity selectorTerms

jonathan-innis:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1337

**Description**

This change ensures that we consider _all_ nodeAffinity terms when determining if the DS is going to schedule to the node.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/preferences.go
+++ b/pkg/controllers/provisioning/scheduling/preferences.go
@@ -36,7 +36,6 @@ type Preferences struct {
 }
 
 func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) bool {
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(pod.Namespace, pod.Name)))
 	relaxations := []func(*v1.Pod) *string{
 		p.removeRequiredNodeAffinityTerm,
 		p.removePreferredPodAffinityTerm,
@@ -50,7 +49,7 @@ func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) bool {
 
 	for _, relaxFunc := range relaxations {
 		if reason := relaxFunc(pod); reason != nil {
-			log.FromContext(ctx).V(1).Info(fmt.Sprintf("relaxing soft constraints for pod since it previously failed to schedule, %s", lo.FromPtr(reason)))
+			log.FromContext(ctx).WithValues("Pod", klog.KRef(pod.Namespace, pod.Name)).V(1).Info(fmt.Sprintf("relaxing soft constraints for pod since it previously failed to schedule, %s", lo.FromPtr(reason)))
 			return true
 		}
 	}
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -352,23 +352,33 @@ func (s *Scheduler) calculateExistingNodeClaims(stateNodes []*state.StateNode, d
 	})
 }
 
+// getDaemonOverhead determines the overhead for each NodeClaimTemplate required for daemons to schedule for any node provisioned by the NodeClaimTemplate
 func getDaemonOverhead(nodeClaimTemplates []*NodeClaimTemplate, daemonSetPods []*corev1.Pod) map[*NodeClaimTemplate]corev1.ResourceList {
-	overhead := map[*NodeClaimTemplate]corev1.ResourceList{}
+	return lo.SliceToMap(nodeClaimTemplates, func(nct *NodeClaimTemplate) (*NodeClaimTemplate, corev1.ResourceList) {
+		return nct, resources.RequestsForPods(lo.Filter(daemonSetPods, func(p *corev1.Pod, _ int) bool { return isDaemonPodCompatible(nct, p) })...)
+	})
+}
 
-	for _, nodeClaimTemplate := range nodeClaimTemplates {
-		var daemons []*corev1.Pod
-		for _, p := range daemonSetPods {
-			if err := scheduling.Taints(nodeClaimTemplate.Spec.Taints).Tolerates(p); err != nil {
-				continue
-			}
-			if err := nodeClaimTemplate.Requirements.Compatible(scheduling.NewPodRequirements(p), scheduling.AllowUndefinedWellKnownLabels); err != nil {
-				continue
-			}
-			daemons = append(daemons, p)
-		}
-		overhead[nodeClaimTemplate] = resources.RequestsForPods(daemons...)
+// isDaemonPodCompatible determines if the daemon pod is compatible with the NodeClaimTemplate for daemon scheduling
+func isDaemonPodCompatible(nodeClaimTemplate *NodeClaimTemplate, pod *corev1.Pod) bool {
+	preferences := &Preferences{}
+	// Add a toleration for PreferNoSchedule since a daemon pod shouldn't respect the preference
+	_ = preferences.toleratePreferNoScheduleTaints(pod)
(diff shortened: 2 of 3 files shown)


pkg/controllers/provisioning/scheduling/preferences.go
```diff
@@ -36,28 +36,27 @@ type Preferences struct {
...
-			log.FromContext(ctx).V(1).Info(fmt.Sprintf("relaxing soft constraints for pod since it previously failed to schedule, %s", lo.FromPtr(reason)))
+			log.FromContext(ctx).WithValues("Pod", klog.KRef(pod.Namespace, pod.Name)).V(1).Info(fmt.Sprintf("relaxing soft constraints for pod since it previously failed to schedule, %s", lo.FromPtr(reason)))
 			return true
 		}
 	}
 	return false
 }
 
-func (p *Preferences) removePreferredNodeAffinityTerm(pod *v1.Pod) *string {
+func (p *Preferences) RemovePreferredNodeAffinityTerm(pod *v1.Pod) *string {
```
jwcesign, line 59: Why export all these functions?

Raj-Popat:
Hey @jonathan-innis  any update on the following PR we are really waiting for this fix to land so we can use karpenter on our clusters currently this is a huge blocked for us

david-dumke:
Would love to see this get merged as well @jonathan-innis.

pkg/controllers/provisioning/scheduling/preferences.go
jonathan-innis: Good call -- made the update so that these aren't exported anymore. Forgot that everything here is happening in the same package

jonathan-innis:
Hey folks -- had dropped this in favor of other things but I'm back on it now and we should have something that gets merged in relatively soon I hope 🎉

pkg/controllers/provisioning/suite_test.go
```diff
@@ -923,22 +932,153 @@ var _ = Describe("Provisioning", func() {
 							Values:   []string{"node-name"},
 						},
 					},
-					ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4"), corev1.ResourceMemory: resource.MustParse("4Gi")}},
+					ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourceMemory: resource.MustParse("1Gi")}},
```
jmdeal, line 935: Why do the resource requirements not meet the DS here? Is this intentional?

pkg/controllers/provisioning/suite_test.go
```diff
@@ -857,26 +874,18 @@ var _ = Describe("Provisioning", func() {
...
+			// We have multiple instance types that we can launch, one with 2Gi and one with 4Gi
+			// If we launch with 2Gi, this means the Daemon pod was respected
+			// If we launch with 4Gi, this means the Daemon pod was respected
 			allocatable := instanceTypeMap[node.Labels[corev1.LabelInstanceTypeStable]].Capacity
 			Expect(*allocatable.Cpu()).To(Equal(resource.MustParse("4")))
 			Expect(*allocatable.Memory()).To(Equal(resource.MustParse("4Gi")))
 		})
-		It("should account for daemonset spec affinity", func() {
+		It("should remove daemonset node hostname affinity when considering daemonset schedulability", func() {
+			// We make two NodePools: one that can't schedule the DaemonSet due to no matching
```
jmdeal, line 886: It looks like this comment needs to be updated? There's only the one NodePool, and it is compatible with the DS.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -923,22 +932,153 @@ var _ = Describe("Provisioning", func() {
...
+				},
+			)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+			node := ExpectScheduled(ctx, env.Client, pod)
+
+			// We have multiple instance types that we can launch, one with 2Gi and one with 4Gi
+			// If we launch with 2Gi, this means the Daemon pod was respected
+			// If we launch with 4Gi, this means the Daemon pod was respected
+			allocatable := instanceTypeMap[node.Labels[corev1.LabelInstanceTypeStable]].Capacity
+			Expect(*allocatable.Cpu()).To(Equal(resource.MustParse("4")))
```
jmdeal, line 1015: I take it we get 4 here due to capacity overhead? Probably worth making the pod or DS slightly larger to make it more clear since currently they sum to 2. This applies to other tests as well.

pkg/controllers/provisioning/scheduling/preferences.go
```diff
@@ -36,7 +36,6 @@ type Preferences struct {
 }
 
 func (p *Preferences) Relax(ctx context.Context, pod *v1.Pod) bool {
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(pod.Namespace, pod.Name)))
```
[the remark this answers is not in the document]
jonathan-innis: Yeah, performance -- didn't really think that we needed to be re-creating the logger everytime here

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -352,23 +352,33 @@ func (s *Scheduler) calculateExistingNodeClaims(stateNodes []*state.StateNode, d
...
-			if err := scheduling.Taints(nodeClaimTemplate.Spec.Taints).Tolerates(p); err != nil {
-				continue
-			}
-			if err := nodeClaimTemplate.Requirements.Compatible(scheduling.NewPodRequirements(p), scheduling.AllowUndefinedWellKnownLabels); err != nil {
-				continue
-			}
-			daemons = append(daemons, p)
+// isDaemonPodCompatible determines if the daemon pod is compatible with the NodeClaimTemplate for daemon scheduling
+func isDaemonPodCompatible(nodeClaimTemplate *NodeClaimTemplate, pod *corev1.Pod) bool {
+	preferences := &Preferences{}
```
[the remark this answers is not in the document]
jonathan-innis: > If we're going to use these compatibility checks elsewhere, it might be good to develop a more general interface

Maybe -- I could definitely see an argument for creating an ORed set of requirements -- but it definitely makes the entire interface more complicated and it's few and far-between when we actually need it -- there are these cases like node affinity, though, when it definitely becomes useful to have it

pkg/controllers/provisioning/suite_test.go
jonathan-innis: > I take it we get 4 here due to capacity overhead

Agree 👍🏼

pkg/controllers/provisioning/suite_test.go
jonathan-innis: Yes, it's intentional -- comment at the top of the test states that we should be using the pod representation for resource requests but need to use the DaemonSet representation for affinity

jmdeal: Got it, makes sense. I am wondering though, is this the correct behavior? What scenarios are there where the resource requests on the pod will differ from the owning daemonset. I imagine it would be after an update to the daemonset, but shouldn't we respect the updated value rather than the value on current pods since that will be what's provisioned for the new Node?

jmdeal:
Had a clarifying question, but approved otherwise.
/lgtm
/hold

pkg/controllers/provisioning/suite_test.go
jonathan-innis: > resource requests on the pod will differ from the owning daemonset

It's in the comment, LimitRange is one example

PR #1728, opened 2024-10-01
fix: spurious logging from migration controller

rschalo:
Fixes #N/A 

**Description**
Suppresses conflict errors on custom resources by no longer using optimistic locking for an annotation we own and suppresses not found errors.

`make presubmit`

### DIFF
--- a/pkg/controllers/migration/crd/controller.go
+++ b/pkg/controllers/migration/crd/controller.go
@@ -31,6 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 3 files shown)


pkg/controllers/migration/crd/controller.go
```diff
@@ -99,6 +100,9 @@ func (c *Controller) Reconcile(ctx context.Context, crd *apiextensionsv1.CustomR
 	stored := crd.DeepCopy()
 	crd.Status.StoredVersions = []string{"v1"}
 	if err := c.kubeClient.Status().Patch(ctx, crd, client.StrategicMergeFrom(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+		if errors.IsConflict(err) {
```
jonathan-innis, line 103: `client.IgnoreNotFound` here as well? I can't imagine how that would happen but probably doesn't hurt

rschalo: Added to `:102`

pkg/controllers/migration/resource/controller.go
```diff
@@ -69,7 +69,7 @@ func (c *Controller[T]) Reconcile(ctx context.Context, req reconcile.Request) (r
 	}))
 	// update annotation on CR
 	if !equality.Semantic.DeepEqual(stored, o) {
-		if err := c.kubeClient.Patch(ctx, o, client.MergeFromWithOptions(stored.(client.Object), client.MergeFromWithOptimisticLock{})); err != nil {
+		if err := c.kubeClient.Patch(ctx, o, client.MergeFrom(stored.(client.Object))); client.IgnoreNotFound(err) != nil {
```
jmdeal, line 72: We still need to do an optimistic lock here to avoid reserializing parts of the NodeClaim spec, right?

jonathan-innis: Shouldn't, right? It's just an annotation update?

jmdeal: Sorry, I got my wires crossed on the issues we were running into. Since it's still a patch we're good.

njtran (commented):
/lgtm
/approve

PR #1734, opened 2024-10-03
fix: increase the eviction cost baseline value for do-not-disruptable pods

cnmcavoy:
Fixes #N/A 

**Description**
Karpenter manages the order for disruption based on the total disruption cost of the pods on a node. However, this does not take into account pods Karpenter can not evict due to the  `karpenter.sh/do-not-disrupt: true`. As a result, nodes that can't have all their pods evicted may be selected as candidates for disruption before nodes with equal or greater number of pods that are all evictable. When this occur, the unevictable node wastes the nodepool's disruption budget.

This change fixes that bug by increasing the base pod eviction cost for pods annotated with `karpenter.sh/do-not-disrupt: true` in Karpenter's simulation models. 

Unit tests + real workloads in our clusters.

### DIFF
--- a/pkg/controllers/disruption/suite_test.go
+++ b/pkg/controllers/disruption/suite_test.go
@@ -796,6 +796,16 @@ var _ = Describe("Pod Eviction Cost", func() {
 		})
 		Expect(cost).To(BeNumerically("<", standardPodCost))
 	})
+	It("should have a high disruptionCost for a pod with a do-not-disrupt annotation set", func() {
+		cost := disruptionutils.EvictionCost(ctx, &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Annotations: map[string]string{
+					v1.DoNotDisruptAnnotationKey: "true",
+				},
+			},
+		})
+		Expect(cost).To(BeNumerically(">", standardPodCost))
+	})
 })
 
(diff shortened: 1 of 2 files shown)


njtran:
Does this PR actually change anything behavior wise? 

The only disruption method that cares about disruption cost is Consolidation, but we don't consider pods with `do-not-disrupt` for consolidation, only for Drift with TGP enabled. https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/types.go#L102, so this code path gets exercised, but doesn't actually change anything for drift or consolidation.

cnmcavoy:
> The only disruption method that cares about disruption cost is Consolidation, but we don't consider pods with do-not-disrupt for consolidation, only for Drift with TGP enabled. https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/types.go#L102, so this code path gets exercised, but doesn't actually change anything for drift or consolidation.

Huh. I opened this PR because we have been testing out various tweaks to Karpenter and I thought this one had an empirical impact from some of our metrics. 

I added some more logging today and it looks like the it bears out the code you linked, the change here only affects the eventual disruption class, which doesn't sort candidates based on the disruption cost anyway. Which agrees with your point. So I'm not sure why we saw an improvement with this, possibly it was coincidental noise.

PR #1735, opened 2024-10-03
perf: don't include terminated nodes in budget

njtran:
Fixes #N/A 

**Description**
Nodes that have the `InstanceTerminating` status condition have had cloudprovider.Delete() succeed. In cases where this delete happens asynchronously in the cloud provider (transitions from Running -> Terminating in AWS), Karpenter will wait to remove the finalizer until the cloud provider machine no longer exists (transitions from Terminating -> Terminated in AWS). Instance termination can vary from seconds to minutes, and counting these against the budget (when there's no pods on the nodes) only impacts expected performance of the disruption controller.

make presubmit and testing locally

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -212,7 +212,7 @@ func BuildDisruptionBudgetMapping(ctx context.Context, cluster *state.Cluster, c
 		// Additionally, don't consider nodeclaims that have the terminating condition. A nodeclaim should have
 		// the Terminating condition only when the node is drained and cloudprovider.Delete() was successful
 		// on the underlying cloud provider machine.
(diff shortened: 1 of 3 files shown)


pkg/controllers/disruption/helpers.go
```diff
@@ -209,6 +209,13 @@ func BuildDisruptionBudgetMapping(ctx context.Context, cluster *state.Cluster, c
 			continue
 		}
 
+		// Additionally, don't consider nodeclaims that have the terminating condition. A nodeclaim should have
+		// the Terminating condition only when the node is drained and cloudprovider.Delete() was successful
+		// on the underlying cloud provider machine.
+		if nc := node.NodeClaim; nc != nil && nc.StatusConditions().Get(v1.ConditionTypeInstanceTerminating).IsTrue() {
```
jonathan-innis, line 215: Doesn't the managed check already validate that a NodeClaim exists?

jonathan-innis (approved):
/lgtm
/approve

PR #1740, opened 2024-10-06
chore: Concurrency acceleration updateNodeClaimHash

helen-frank:
Fixes #N/A 

**Description**
Concurrency acceleration updateNodeClaimHash


njtran (commented):
I'm not sure I understand why this needs to change. How does this speed up performance? Also, is this not something you can do with workqueue.ParallelizeUntil?

helen-frank:
> I'm not sure I understand why this needs to change. How does this speed up performance? Also, is this not something you can do with workqueue.ParallelizeUntil?

Asynchronously to use `kubeClient.Patch`

I went and read the code for workqueue.ParallelizeUntil, and it's really better than what I was doing, thanks for the guidance!

Issue #1733, opened 2024-10-02
 Parameterize Multinode + Single node consolidation timeout


Pokom:
### Description
> [!NOTE] 
> This is similar to #903 but distinct in that this happens to our largest clusters regardless of scale up/scale down activity. #1031 was opened but closed due to needing an RFC, and I would like to work on putting that together. 

**What problem are you trying to solve?**

Provide the ability to configure the values for [multinode](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/multinodeconsolidation.go#L35) and [single node consolidation](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/singlenodeconsolidation.go#L30) timeouts. 

In large clusters with somewhat complex nodepool setups and anti affinity rules, we're consistently running into timeouts for the consolidation process. The impact is that our clusters are over provisioned as nodes aren't being taken offline. 

We're also working on profiling karpenter and working to identify where the bottleneck is in the code, as a stopgap this would be a nice feature to have.  

It would be really for us to be able to configure these values at runtime and set them to values that would allow the consolidation process to finish. It may not be as fast as the default values, but having it finish slower is preferable to having the process timeout and never complete.

**How important is this feature to you?**

This is very important as Karpenter is causing a fairly large uptick in spend for large clusters because the consolidation can't process fast enough.  

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Pokom:
@njtran are you open to an outside contribution for the issue? I should have some time this week to get a PR ready

nantiferov:
I can add that issue with consolidation timeout is happening on smaller clusters too when Karpenter pod is limited by CPU.

We hit this during update to v1.0, when Karpenter running on Fargate hit CapacityProvisioned in terms of CPU. I guess same could happen with `resources.limits` are set and Karpenter pod will be throttled.

We fixed this issue by bumping CPU resources on Fargate and now consolidation is working relatively well.

[...]

PR #1754, opened 2024-10-15
feat(options): Add consolidation timeout options

Pokom:
Adds two options for the following timeouts
- multinodeconsolidation
- singlenodeconsolidation

These are exposed on the following ways:
- `--multi-node-consolidation-timeout` or `MULTI_NODE_CONSOLIDATION_TIMEOUT`
- `--single-node-consolidation-timeout` or `SINGLE_NODE_CONSOLIDATION_TIMEOUT`

The primary way of testing this was by building the image and running within dev and production clusters within Grafana Labs fleet.

---

- refs #1733

Fixes #N/A 

**Description**

### DIFF
--- a/pkg/controllers/disruption/multinodeconsolidation.go
+++ b/pkg/controllers/disruption/multinodeconsolidation.go
@@ -117,7 +117,7 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context, 
 	lastSavedCommand := Command{}
 	lastSavedResults := scheduling.Results{}
 	// Set a timeout
-	timeout := m.clock.Now().Add(options.FromContext(ctx).SinglenodeConsolidationTimeout)
+	timeout := m.clock.Now().Add(options.FromContext(ctx).MultiNodeConsolidationTimeout)
 	// binary search to find the maximum number of NodeClaims we can terminate
 	for min <= max {
 		if m.clock.Now().After(timeout) {
--- a/pkg/controllers/disruption/multinodeconsolidation.go
+++ b/pkg/controllers/disruption/multinodeconsolidation.go
@@ -20,7 +20,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"time"
 
 	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -29,10 +28,9 @@ import (
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 	scheduler "sigs.k8s.io/karpenter/pkg/scheduling"
 )
-
-const MultiNodeConsolidationTimeoutDuration = 1 * time.Minute
 
 type MultiNodeConsolidation struct {
 	consolidation
@@ -119,7 +117,7 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context, 
 	lastSavedCommand := Command{}
 	lastSavedResults := scheduling.Results{}
 	// Set a timeout
-	timeout := m.clock.Now().Add(MultiNodeConsolidationTimeoutDuration)
+	timeout := m.clock.Now().Add(options.FromContext(ctx).SinglenodeConsolidationTimeout)
 	// binary search to find the maximum number of NodeClaims we can terminate
 	for min <= max {
 		if m.clock.Now().After(timeout) {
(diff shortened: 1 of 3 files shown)


pkg/operator/options/options.go
```diff
@@ -96,6 +98,8 @@ func (o *Options) AddFlags(fs *FlagSet) {
...
 	fs.DurationVar(&o.BatchIdleDuration, "batch-idle-duration", env.WithDefaultDuration("BATCH_IDLE_DURATION", time.Second), "The maximum amount of time with no new pending pods that if exceeded ends the current batching window. If pods arrive faster than this time, the batching window will be extended up to the maxDuration. If they arrive slower, the pods will be batched separately.")
 	fs.StringVar(&o.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "SpotToSpotConsolidation=false"), "Optional features can be enabled / disabled using feature gates. Current options are: SpotToSpotConsolidation")
+	fs.DurationVar(&o.MultinodeConsolidationTimeout, "multi-node-consolidation-timeout", env.WithDefaultDuration("MULTI_NODE_CONSOLIDATION_TIMEOUT", 1*time.Minute), "The maximum amount of time that can be spent doing multinode consolidation before timing out. Defaults to 1 minute")
+	fs.DurationVar(&o.SinglenodeConsolidationTimeout, "single-node-consolidation-timeout", env.WithDefaultDuration("SINGLE_NODE_CONSOLIDATION_TIMEOUT", 3*time.Minute), "The maximum amount of time that can be spent doing single node consolidation before timing out. Defaults to 3 minute")
```
nikimanoledaki, line 102: Nit 😄 
```suggestion
	fs.DurationVar(&o.SinglenodeConsolidationTimeout, "single-node-consolidation-timeout", env.WithDefaultDuration("SINGLE_NODE_CONSOLIDATION_TIMEOUT", 3*time.Minute), "The maximum amount of time that can be spent doing single node consolidation before timing out. Defaults to 3 minutes")
```

pkg/operator/options/options.go
```diff
@@ -46,21 +46,23 @@ type FeatureGates struct {
...
+	LeaderElectionNamespace        string
+	MemoryLimit                    int64
+	LogLevel                       string
+	LogOutputPaths                 string
+	LogErrorOutputPaths            string
+	BatchMaxDuration               time.Duration
+	BatchIdleDuration              time.Duration
+	FeatureGates                   FeatureGates
+	MultinodeConsolidationTimeout  time.Duration
+	SinglenodeConsolidationTimeout time.Duration
```
nikimanoledaki, line 65: Could we rename these to `MultiNodeConsolidationTimeoutDuration` (capitalised `N` for `Node`) and `SingleNodeConsolidationTimeout`? This would be similar to the previous const and keep consistency with the `MultiNodeConsolidation` struct.

nikimanoledaki (changes requested):
Spotted a few changes that are needed.
🚀 let's go!

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -119,7 +117,7 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
 	lastSavedCommand := Command{}
 	lastSavedResults := scheduling.Results{}
 	// Set a timeout
-	timeout := m.clock.Now().Add(MultiNodeConsolidationTimeoutDuration)
+	timeout := m.clock.Now().Add(options.FromContext(ctx).SinglenodeConsolidationTimeout)
```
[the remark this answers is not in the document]
nikimanoledaki: Fixing this might fix the failing test - not sure 🤔 https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/state/suite_test.go#L1469, specifically [this](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/state/suite_test.go#L1469)

pkg/operator/options/options.go
Pokom: I'm going to remove those all together since it's redundant if you pass the `--help` flag:

```
  -multi-node-consolidation-timeout duration
        The maximum amount of time that can be spent doing multinode consolidation before timing out. Defaults to 1 minute (default 1m0s)
  -single-node-consolidation-timeout duration
        The maximum amount of time that can be spent doing single node consolidation before timing out. Defaults to 3 minute (default 3m0s)
```

njtran (commented):
The main problem I have with this is that it locks in API that looks at our implementation. If we change the implementation in the future, it makes it tougher to do so. I could see this being an "alpha" feature, but I also wonder if there's a way to make this apply to both types without specifically naming them?

Pokom:
> The main problem I have with this is that it locks in API that looks at our implementation. If we change the implementation in the future, it makes it tougher to do so. I could see this being an "alpha" feature, but I also wonder if there's a way to make this apply to both types without specifically naming them?

That's fair. I'm not tied to any implementation here, and having one variable + config is certainly easier to manage then multiple. I'll be at KubeCon next week, so it'll be a bit of time before I can pick this up again

Issue #1659, opened 2024-09-12
Node NotReady Disruption Controller


diranged:
### Description

**What problem are you trying to solve?**
Sometimes nodes just become `NotReady` for a variety of reasons (bad cloud provider instance, non-responsive kubelet, etc). When a Node has been in a `Ready` state and then transitions into `NotReady`, I think that Karpenter should have another Disruption Controller that monitors for these nodes and terminates them.

Third party controllers like the Spot.io Ocean Product, and the Cluster Autoscaler both handle nodes that become `NotReady` for you automatically. Karpenter should be able to do the same thing.

(_Note we have also raised this with our AWS TAM via a support ticket, and we were recommended to open a feature-request here_)

Related: https://github.com/kubernetes-sigs/karpenter/issues/1573

**How important is this feature to you?**

This is actually a blocker for us migrating off of our current tools - we launch enough nodes and we have enough failures throughout the day that we cannot fully migrate unless we have a completely automated self healing system where these nodes get cycled out once they become `NotReady`.

(_separate but related, is the ongoing discussion at https://github.com/bottlerocket-os/bottlerocket/issues/4075 about EKS nodes becoming unready due to heavy memory pressure_)

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jigisha620:
Hi @diranged,
The use case that you have presented should be taken care by NodeRepair which was implemented in this PR - https://github.com/kubernetes-sigs/karpenter/pull/1793. It was released with [v1.1.0](https://github.com/kubernetes-sigs/karpenter/releases/tag/v1.1.0).

saurav-agarwalla:
Resolving this now that Node Repair has been released.

/triage accepted
/remove-lifecycle rotten
/close

PR #1768, opened 2024-10-22
RFC: Node Auto Repair

engedaam:
Fixes #N/A 

**Description**
The purpose of this RFC is to Propose option around the best ways to support node auto repair for unhealthy Karpenter provisioned nodes. Related issue: https://github.com/kubernetes-sigs/karpenter/issues/1659

- N/A

### DIFF
--- a/designs/node-repair.md
+++ b/designs/node-repair.md
@@ -1,0 +1,129 @@
+# Node Auto Repair 
+
+## Problem Statement
+
+Nodes can experience failure modes that cause degradation to the underlying hardware, file systems, or container environments. Some of these failure modes are surfaced through the Node object such as network unavailability, disk pressure, or memory pressure, while others are not surfaced at all such as accelerator health. A Diagnostic Agent such as the [Node Problem Detector (NPD)](https://github.com/kubernetes/node-problem-detector) offers a way to surface these failures as additional status conditions on the node object.
+
+When a status condition is surfaced through the Node, it indicates that the Node is unhealthy. Karpenter does not currently react to those conditions.
+
+* Mega Issue: https://github.com/kubernetes-sigs/karpenter/issues/750
+    * Related (Unreachable): https://github.com/aws/karpenter-provider-aws/issues/2570
+    * Related (Remove by taints): https://github.com/aws/karpenter-provider-aws/issues/2544
+    * Related (Known resource are not registered) Fixed by v0.28.0: https://github.com/aws/karpenter-provider-aws/issues/3794
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The alpha implementation will not consider these features:
+
+  - Disruption Budgets
+  - Customer-Defined Conditions
+  - Customer-Defined Remediation Time
+
+The team does not have enough data to determine the right level of configuration that users will utilize. The opinionated mechanism would be responsible for defining unhealthy notes. The advantage of creating the mechanism would be to reduce the configuration burden for customers. **The feature will be gated under an alpha NodeRepair=true feature flag. This will  allow for additional feedback from customers. Additional feedback can support features that were originally considered out of scope for the Alpha stage.**
+
+## Recommendation: CloudProvider-Defined RepairPolicies
+
+```
+type RepairStatement struct {
+    // Type of unhealthy state that is found on the node
+    Type metav1.ConditionType 
+    // Status condition of when a node is unhealthy
+    Status metav1.ConditionStatus
+    // TolerationDuration is the duration the controller will wait
+    // before attempting to terminate nodes that are marked for repair.
+    TolerationDuration time.Duration
+}
+
+type CloudProvider interface {
+  ...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The RepairPolicy will contain a set of statements that the Karpenter controller will use to watch node conditions. On any given node, multiple node conditions may exist simultaneously. In those cases, we will chose the shortest `TolerationDuration` for a given condition. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types. The `TolerationDuration` will wait until a unhealthy state has passed the duration and is considered terminal:
+
+1. A diagnostic agent will add a status condition on a node 
+2. Karpenter will reconcile nodes and match unhealthy conditions with repair policy statements
+3. Node Health controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by the TolerationDuration
+
+### Example
+
+The example will look at the supporting Node Problem Detector for the AWS Karpenter Provider:
+```
+func (c *CloudProvider) RepairPolicy() []cloudprovider.RepairStatement {
+    return cloudprovider.RepairStatement{
+        {
+            Type: "Ready"
+            Status: corev1.ConditionFalse,
+            TrolorationDuration: "30m"
+        },
+        {
+            Type: "NetworkUnavailable"
+            Status: corev1.ConditionTrue,
+            TrolorationDuration: "10m"
+        },
+        ...
+    }
+}
+```
+
+In the example above, the AWS Karpenter Provider supports monitoring and terminating two node status conditions of the Kubelet Ready condition and the NPD  NetworkUnavailable condition. Below are the two cases of when we will act on nodes:
+
+```
+apiVersion: v1
+kind: Node
+metadata:
+  ...
+status:
+  condition:
+    - lastHeartbeatTime: "2024-11-01T16:29:49Z"
+      lastTransitionTime: "2024-11-01T15:02:48Z"
+      message: no connection
+      reason: Network is not available
+      status: "False"
+      type: NetworkUnavailable
+    ...
+    
+- The Node here will be eligible for node repair after at `2024-11-01T15:12:48Z`    
+---
+apiVersion: v1
+kind: Node
+metadata:
+  ...
+status:
+  condition:
+    - lastHeartbeatTime: "2024-11-01T16:29:49Z"
+      lastTransitionTime: "2024-11-01T15:02:48Z"
+      message: kubelet is posting ready status
+      reason: KubeletReady
+      status: "False"
+      type: NetworkUnavailable
+    - lastHeartbeatTime: "2024-11-01T16:29:49Z"
+      lastTransitionTime: "2024-11-01T15:02:48Z"
+      message: kubelet is posting ready status
+      reason: KubeletReady
+      status: "False"
+      type: Ready
+    ...
+    
+- The Node here will be eligible for node repair after at `2024-11-01T15:32:48Z`    
+```
+
+
+## Forceful termination
+
+For a first iteration approach, Karpenter will implement the force termination. Today, the graceful termination in Karpenter will attempt to wait for the pod to be fully drained on a node and all volume attachments to be deleted from a node. This raises the problem that during the graceful termination, the node can be stuck terminating when the pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to forcefully terminate nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future considerations 
+
+There are additional features we will consider including after the initial iteration. These include:
+
+* Disruption controls (budgets, terminationGracePeriod) for unhealthy nodes 
+* Node Reboot (instead of replacement)
+* Configuration surface for graceful vs forceful termination 
+* Additional consideration for the availability zone resiliency 
+


designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
+#  Karpenter Node Auto Repair 
+
+## Problem Statement
+
+Karpenter users are observing that occasionally nodes can face failure mode that prevent them to be healthy in a Kubernetes clusters. These nodes stay in unhealthy state for a period of time without any actions, and in some cases forever with manual intervention. Karpenter continues to simulate pods to these nodes, but the Kubernetes scheduler won't schedule to them. This will leave pods unscheduled and cause downtime for applications. Customers have the need for karpenter to repair unhealthy nodes. The karpenter team has seen several failure mode such as disk pressure, memory pressure, networking outage etc. that can cause nodes to get into an unhealthy state Here are issues related to the topic:
```
ellistarn, line 5: I'd be cautious of repairing on disk/memory pressure, as opposed to letting taint based eviction do their job https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/#taint-based-evictions. That said, I do strongly agree there are cases where the only fix is to toss the node.

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+
+```
+type UnhealthyConditions struct {
+    // Type of unhealthy state that is found on the node
+    Type string 
+    // Status condition of when a node is unhealthy
+    status string
+    // RepairAfter is the duration the controller will wait
+    // before attempting to terminate nodes that are underutilized.
+    RepairAfter time.Duration
```
ellistarn, line 27:
```suggestion
    Duration time.Duration
```

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+The scope of the solution suggested bellow will be an opinionated method that will not give customer configurability. The team does not have enough data to determine the right level of configuration surface that users will utilize. The implementation will not consider budgets, API surface for repairing unhealthy nodes, or option for nodes to be rebooted instead of replaced. The solution will also not give any Karpenter based approach to enable AZ resiliency. **The feature will be gated under an alpha feature flag to allow for additional feedback from customers, and make subsequent changes in the future.**
+
+### Option  1 (recommended) : Unhealthy condition set cloud provider interface  
+
+```
+type UnhealthyConditions struct {
+    // Type of unhealthy state that is found on the node
+    Type string 
+    // Status condition of when a node is unhealthy
+    status string
```
ellistarn, line 24: https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ConditionStatus
```suggestion
    Status metav1.ConditionStatus
```

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+    // RepairAfter is the duration the controller will wait
+    // before attempting to terminate nodes that are underutilized.
+    RepairAfter time.Duration
+}
+
+type CloudProvider interface {
+  ...
+    // UnhealthyConditions is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    UnhealthyConditions() []v1.UnhealthyConditions
```
ellistarn, line 34:
```suggestion
    RepairConditions() []v1.RepairCondition
```

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+  ...
+}
+```
+
+The `UnhealthyConditions` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on to a node
```
ellistarn, line 41: Worth noting that https://github.com/kubernetes/node-problem-detector can be used here.

designs/node-repair.md
ellistarn, line 34: Thoughts on 
```suggestion
    RepairPolicy() []v1.RepairPolicy
```

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+The `UnhealthyConditions` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and match unhealthy node condition to cloud provider defined unhealthy condition 
+3. Unhealthy condition are added to the NodeClaim
+4. Disruption controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by `repairAfter` of the unhealthy condition 
+
+### Option  2: cloud provider interface  for determining node health
```
ellistarn, line 46: Worth calling out the main downside of this one. Providers will need to implement much more undifferentiated logic. It's also not clear how we'd model the concept of "duration" and "action" in this API.

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+Cloud provider hydrate a `UnhealthyConditions` map that will hold the condition to track on nodes. The map will track the condition and `repairAfter` will dictate the length of time the node will be inside the cluster before action is taken against it. Node health validation will follow as such:
+
+1. A diagnostic agent will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and identify unhealthy nodes defined by `UnhealthyConditions` map
+3. Unhealthy condition is added to the NodeClaim
+4. Disruption controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by `UnhealthyConditions` map
+
+## Recommended Solution
```
ellistarn, line 86: +1 ship it.

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+    // to monitor on the node. Customer will need 
+    UnhealthyConditions() []v1.UnhealthyConditions
+  ...
+}
+```
+
+The `UnhealthyConditions` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and match unhealthy node condition to cloud provider defined unhealthy condition
```
mariuskimmina, line 42: Does this imply that the cloud provider needs to be aware of the node being unhealthy?
We observed cases where the node in our EKS cluster was in an unreachable state (kubelet died) but the from the AWS console point of view it was fine (health-checks passing).

Will this implementation consider the case where the node is unhealthy from kubernetes point of view but healthy from the cloud provider side?

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+
+* Mega Issue: https://github.com/kubernetes-sigs/karpenter/issues/750
+    * Related (Unreachable): https://github.com/aws/karpenter-provider-aws/issues/2570
+    * Related (Remove by taints): https://github.com/aws/karpenter-provider-aws/issues/2544
+    * Related (Known resource are not registered) Fixed by v0.28.0: https://github.com/aws/karpenter-provider-aws/issues/3794
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The scope of the solution suggested bellow will be an opinionated method that will not give customer configurability. The team does not have enough data to determine the right level of configuration surface that users will utilize. The implementation will not consider budgets, API surface for repairing unhealthy nodes, or option for nodes to be rebooted instead of replaced. The solution will also not give any Karpenter based approach to enable AZ resiliency. **The feature will be gated under an alpha feature flag to allow for additional feedback from customers, and make subsequent changes in the future.**
```
GnatorX, line 15:
```suggestion
The scope of the solution suggested below will be an opinionated method that will not give customer configurability. The team does not have enough data to determine the right level of configuration surface that users will utilize. The implementation will not consider budgets, API surface for repairing unhealthy nodes, or option for nodes to be rebooted instead of replaced. The solution will also not give any Karpenter based approach to enable AZ resiliency. **The feature will be gated under an alpha feature flag to allow for additional feedback from customers, and make subsequent changes in the future.**
```

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+### Option  1 (recommended) : Unhealthy condition set cloud provider interface  
+
+```
+type UnhealthyConditions struct {
+    // Type of unhealthy state that is found on the node
+    Type string 
+    // Status condition of when a node is unhealthy
+    status string
+    // RepairAfter is the duration the controller will wait
+    // before attempting to terminate nodes that are underutilized.
```
GnatorX, line 26:
```suggestion
    // before attempting to terminate nodes that are marked for repair.
```

designs/node-repair.md
GnatorX: We use this however I don't think I fully grok how to input the set of conditions for Karpenter.

designs/node-repair.md
```diff
@@ -0,0 +1,92 @@
...
+  ...
+}
+```
+
+The `UnhealthyConditions` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and match unhealthy node condition to cloud provider defined unhealthy condition 
+3. Unhealthy condition are added to the NodeClaim
+4. Disruption controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by `repairAfter` of the unhealthy condition
```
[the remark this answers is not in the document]
engedaam: We don't need another orchestration the main idea will be to add the status condition to get operatorPkg status condition metrics for free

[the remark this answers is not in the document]
GnatorX: But is there value in adding status conditions to the node claim vs leaving it in only the node?

designs/node-repair.md
engedaam: Yes, all the solution here focus on Kubernetes status condition. We plan on acting when a node is unhealthy from a Kubernetes point of view or an agent has marked as such. Cloud providers need to have a supported node diagnostic agent like [node-problem-detector](https://github.com/kubernetes/node-problem-detector). A cloud provider will specify the node condition that Karpenter will watch for and then act against.

designs/node-repair.md
engedaam: We are not going to do that for this first iteration, but in subsequent iteration. These condition will need to be supported by the cloud provider for now.

designs/node-repair.md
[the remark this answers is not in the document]
njtran: I worry about the propagation delay here, and am not totally convinced that the extra syncing work + apiserver load is worth the metrics we'd get. Maybe I'm underestimating the value we get though

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
+#  Karpenter Node Auto Repair
```
jonathan-innis, line 1:
```suggestion
# Karpenter Node Auto Repair 
```
nit

jonathan-innis, line 1:
```suggestion
# Node Auto Repair 
```

The fact that this is being proposed in the Karpenter project implies that this is a Karpenter design so I'd drop that bit

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+
+* Mega Issue: https://github.com/kubernetes-sigs/karpenter/issues/750
+    * Related (Unreachable): https://github.com/aws/karpenter-provider-aws/issues/2570
+    * Related (Remove by taints): https://github.com/aws/karpenter-provider-aws/issues/2544
+    * Related (Known resource are not registered) Fixed by v0.28.0: https://github.com/aws/karpenter-provider-aws/issues/3794
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The scope of the solution suggested below will be an opinionated method that will not give customer configurability. The team does not have enough data to determine the right level of configuration surface that users will utilize. The implementation will not consider budgets, API surface for repairing unhealthy nodes, or option for nodes to be rebooted instead of replaced. The solution will also not give any Karpenter based approach to enable AZ resiliency. **The feature will be gated under an alpha feature flag to allow for additional feedback from customers, and make subsequent changes in the future.**
```
jonathan-innis, line 15: I'd consider using more positive wording here. This makes it sound like we are making a decision that makes the customer experience worse since we aren't surfacing configuration. In reality, we are choosing not to surface cofiguration because we believe this will make the overall customer experience better

jonathan-innis, line 15: Perhaps we should talk about the fact that this feature will be introduced in ALPHA with the `NodeAutoRepair=true` feature flag? That might help someone understand that our purpose here is to receive feedback over the feature before considering it a stable feature in the Karpenter project

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
+#  Karpenter Node Auto Repair 
+
+## Problem Statement
+
+Karpenter users are observing that occasionally nodes can face failure mode that prevent them to be healthy in a Kubernetes clusters. These nodes stay in unhealthy state for a period of time without any actions, and in some cases forever with manual intervention. Karpenter continues to simulate pods to these nodes, but the Kubernetes scheduler won't schedule to them. This will leave pods unscheduled and cause downtime for applications. Customers have the need for karpenter to repair unhealthy nodes. The karpenter team has seen several failure mode such as disk pressure, memory pressure, networking outage etc. that can cause nodes to get into an unhealthy state Here are issues related to the topic:
```
jonathan-innis, line 5:
```suggestion
Nodes can experience failure modes that cause degredation to the underlying hardware, filesystem, or container environment. Some of these failure modes are surfaced through the Node object (consider referencing some of them) while others are not surfaced at all (consider referencing some of these). Node Problem Detector offers a way to surface these failures as additional status conditions on the node object.

In either case, even if a status condition is surfaced through the Node that indicates that the Node is unhealthy, Karpenter doesn't currently react to this unhealthiness in any way today. 
```

Just a suggestion on wording for this -- I'd consider framing this in how nodes experience failures today and which of these are currently surfaced -- along with what other components are needed to surface other types of failures. 

It might also be useful to show what this failure looks like on an actual node by showing status conditions for a node that is failing.

I'd also say that your comment on Karpenter continuing to schedule pods to these nodes may or may not be true. If the node gets tainted after initialization, we shouldn't be continuing to schedule there -- and the kube-scheduler won't either. I thought the larger issue was that we will just leave the node around even if it's unhealthy or even dead?

designs/node-repair.md
jonathan-innis, line 15: I'd also say that you could cull a lot of this wording out. In general, I think you are bringing up a lot of feature adds that we have talked about for this addition to Karpenter but folks may not think about if you don't bring them up.

It's a bit of answering a question before it's being asked and I think you could probably just drop a bunch of it.

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+    * Related (Unreachable): https://github.com/aws/karpenter-provider-aws/issues/2570
+    * Related (Remove by taints): https://github.com/aws/karpenter-provider-aws/issues/2544
+    * Related (Known resource are not registered) Fixed by v0.28.0: https://github.com/aws/karpenter-provider-aws/issues/3794
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The scope of the solution suggested below will be an opinionated method that will not give customer configurability. The team does not have enough data to determine the right level of configuration surface that users will utilize. The implementation will not consider budgets, API surface for repairing unhealthy nodes, or option for nodes to be rebooted instead of replaced. The solution will also not give any Karpenter based approach to enable AZ resiliency. **The feature will be gated under an alpha feature flag to allow for additional feedback from customers, and make subsequent changes in the future.**
+
+### Option  1 (recommended) : Unhealthy condition set cloud provider interface
```
jonathan-innis, line 17:
```suggestion
### Option 1 (recommended): Unhealthy condition set cloud provider interface  
```

nit: Watch the number of spaces you are using throughout your design docs

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The scope of the solution suggested below will be an opinionated method that will not give customer configurability. The team does not have enough data to determine the right level of configuration surface that users will utilize. The implementation will not consider budgets, API surface for repairing unhealthy nodes, or option for nodes to be rebooted instead of replaced. The solution will also not give any Karpenter based approach to enable AZ resiliency. **The feature will be gated under an alpha feature flag to allow for additional feedback from customers, and make subsequent changes in the future.**
+
+### Option  1 (recommended) : Unhealthy condition set cloud provider interface  
+
+```
+type RepairPolicy struct {
```
jonathan-innis, line 20: I feel like you can't argue that we should introduce all of this configuration without providing tangible examples of how these values would differ for Cloud Poviders for different types of status conditions. For example, if you were to implement this for Node Problem Detector with Kwok, what's your expectation of how this would look (I also get that Kwok may not be the best example because they aren't real nodes)

jonathan-innis: Maybe if we were to implement Node Problem Detector with AWS, how would this look?

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The `RepairPolicy` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent, such as [node problem detector](https://github.com/kubernetes/node-problem-detector), will add a status condition on to a node
```
jonathan-innis, line 41: nit: This is the first time that you are bringing up the term "diagnostic agent". If you are going to use this throughout the doc, I'd recommend bringing up the term earlier in a definitions section

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+type CloudProvider interface {
+  ...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The `RepairPolicy` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal:
```
jonathan-innis, line 39:
```suggestion
The `RepairPolicy` will be a set of conditions that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
```

nit

designs/node-repair.md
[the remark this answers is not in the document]
jonathan-innis: > be to add the status condition to get operatorPkg status condition metrics for free

You could do this by using the operatorpkg status controller on the Node directly. I don't think you have to force status conditions into the NodeClaim from the Node just to get more metrics

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+
+type CloudProvider interface {
+  ...
+    // IsHealthy returns whether a NodeClaim is considered operating as expected
+    IsHealthy(context.Context, *v1.NodeClaim) (UnhealthyReason, error)
+  ...
+}
+```
+
+`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. IsHealthy is a Turing complete operation and allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`.
```
jonathan-innis, line 59:
```suggestion
`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. IsHealthy allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`. 
```

nit: It feels a little pedantic to use "Turing complete" here. I'm not sure that it adds much to think of it this way so I'd generally drop this

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+type CloudProvider interface {
+  ...
+    // IsHealthy returns whether a NodeClaim is considered operating as expected
+    IsHealthy(context.Context, *v1.NodeClaim) (UnhealthyReason, error)
+  ...
+}
+```
+
+`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. IsHealthy is a Turing complete operation and allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`. 
+Main drawback of this option is providers will need to implement much more undifferentiated logic, without having tha ability to model duration and remediation on nodes. The reconciliation step for replacing an unhealthy node will look as such:
```
jonathan-innis, line 60: I'd like to point out that CloudProviders still have the ability to introduce delay -- it would just have to be hard-coded or defined in the NodeClass interface. You could just implement the delay in the IsHealthy() method

jonathan-innis, line 60:
```suggestion
Main drawback of this option is providers will need to implement much more undifferentiated logic, without having the ability to model duration and remediation on nodes. The reconciliation step for replacing an unhealthy node will look like: 
```

nit

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+    IsHealthy(context.Context, *v1.NodeClaim) (UnhealthyReason, error)
+  ...
+}
+```
+
+`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. IsHealthy is a Turing complete operation and allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`. 
+Main drawback of this option is providers will need to implement much more undifferentiated logic, without having tha ability to model duration and remediation on nodes. The reconciliation step for replacing an unhealthy node will look as such: 
+
+1. Karpenter will reconcile on the NodeClaim and make a cloud provide call with IsHealthy
+2. If the node is determined to unhealthy, a status condition will be added to the NodeClaim
```
jonathan-innis, line 63: What status condition? What's the type? Reason? If we are going to propose this as a real option I think we should have an example of what this will tangibly look like if we take it

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+The `RepairPolicy` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent, such as [node problem detector](https://github.com/kubernetes/node-problem-detector), will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and match unhealthy node condition to cloud provider defined unhealthy condition 
+3. Unhealthy condition are added to the NodeClaim
+4. Disruption controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by `repairAfter` of the unhealthy condition 
+
+### Option  2: cloud provider interface  for determining node health
```
jonathan-innis, line 46: This option confuses me a bit without having some kind of graceful termination. There's basically no reason to pipe something like this through the neutral codebase. This is very similar to the interruption controller in `aws/karpenter-provider-aws` where a cloud provider can just implement forceful disruption on their own without having to hook into any additional orchestration

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. IsHealthy is a Turing complete operation and allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`. 
+Main drawback of this option is providers will need to implement much more undifferentiated logic, without having tha ability to model duration and remediation on nodes. The reconciliation step for replacing an unhealthy node will look as such: 
+
+1. Karpenter will reconcile on the NodeClaim and make a cloud provide call with IsHealthy
+2. If the node is determined to unhealthy, a status condition will be added to the NodeClaim
+3. Disruption controller will force terminate the node claim immediately 
+
+### Option  3:  Cloud provider injected unhealthy condition
```
jonathan-innis, line 66: How is this tangibly different than Option 1? Isn't this just the same thing -- one is just a static value and the other is an interface type?

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Option  1 (recommended) : Unhealthy condition set cloud provider interface  
+
+```
+type RepairPolicy struct {
+    // Type of unhealthy state that is found on the node
+    Type metav1.ConditionType 
+    // Status condition of when a node is unhealthy
+    Status metav1.ConditionStatus
+    // RepairAfter is the duration the controller will wait
```
jonathan-innis, line 25: This comment doesn't match the name of the variable

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+    // before attempting to terminate nodes that are marked for repair.
+    TolerationDuration time.Duration
+}
+
+type CloudProvider interface {
+  ...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
```
jonathan-innis, line 34: Is there any way for a user to block repair from happening after enabling this feature? I'm assuming that `karpenter.sh/do-not-disrupt` won't work here but what about if I'm a user and I want to debug?

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+  ...
+}
+```
+
+The `RepairPolicy` will be a set of condition that the Karpenter controller will watch. The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent, such as [node problem detector](https://github.com/kubernetes/node-problem-detector), will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and match unhealthy node condition to cloud provider defined unhealthy condition 
+3. Unhealthy condition are added to the NodeClaim
```
jonathan-innis, line 43: Can we show an example of what you mean here? I can't tell if you are intending that we mirror conditions onto the NodeClaim from the Node or if we just stick an `Unhealthy` condition on the NodeClaim once one of the conditions on the Node has reached its toleration duration? I think an example would shed some light.

designs/node-repair.md
```diff
@@ -0,0 +1,93 @@
...
+4. Disruption controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by `repairAfter` of the unhealthy condition
```
jonathan-innis, line 44: Name doesn't match what you are calling it in the API above

designs/node-repair.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Forceful termination
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node or if set, termination grace period will force terminate the node. For the first iteration of problem, the recommendation will be to force terminate nodes now and follow-up to allow configurability for either the cloud providers or the Karpenter users once we have sufficient data to support the need for graceful termination for unhealthy nodes.
```
Bryce-Soghigian, line 95: Thoughts on having other autohealing actions that we can take? Repair Policy allows us to define conditions for unhealthy, but also it might make sense to have an RepairAction the customer can eventually configure. 

AKS has found great success to repairing node not ready by simply issuing a vm.Restart operation to azure, there are some benefits to not fully deleting the vm and instead doing softer operations for repair. 

This is out of scope for this design but wanted to mention it so its in peoples minds.

Bryce-Soghigian: https://learn.microsoft.com/en-us/azure/aks/node-auto-repair for reference/inspiration

Bryce-Soghigian: https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-repair

designs/node-repair.md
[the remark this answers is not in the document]
engedaam: Dropped this to relay on operatorpkg status controller

designs/node-repair.md
engedaam: Yeah, this would is considering how we want to the cloud providers to interact with the API

designs/node-repair.md
engedaam: We could add it a block just for that annotation, but it might be best for users to disable the feature to investigate. They would also have metrics and alarming to be able to give a better insight

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
+# Node Auto Repair 
+
+## Problem Statement
+
+Nodes can experience failure modes that cause degradation to the underlying hardware, filesystem, or container environment. Some of these failure modes are surfaced through the Node object (consider referencing some of them) while others are not surfaced at all (consider referencing some of these). A Diagnostic Agent such as [Node Problem Detector(NPD)](https://github.com/kubernetes/node-problem-detector) offers a way to surface these failures as additional status conditions on the node object.
```
jonathan-innis, line 5: "consider referencing some of them" is in here twice and I was under the impression that you were going to add examples rather than leaving the boilerplate in here

jonathan-innis, line 5:
```suggestion
Nodes can experience failure modes that cause degradation to the underlying hardware, filesystem, or container environment. Some of these failure modes are surfaced through the Node object (consider referencing some of them) while others are not surfaced at all (consider referencing some of these). A Diagnostic Agent such as [Node Problem Detector (NPD)](https://github.com/kubernetes/node-problem-detector) offers a way to surface these failures as additional status conditions on the node object.
```

nit

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+In either case, even if a status condition is surfaced through the Node that indicates that the Node is unhealthy, Karpenter doesn't currently react to this unhealthiness in any way today.
```
jonathan-innis, line 7:
```suggestion
Even if a status condition is surfaced through the Node that indicates that the Node is unhealthy, Karpenter doesn't currently react to this unhealthiness in any way today.
```

What are the two cases? I think you could drop this bit

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+    * Related (Unreachable): https://github.com/aws/karpenter-provider-aws/issues/2570
+    * Related (Remove by taints): https://github.com/aws/karpenter-provider-aws/issues/2544
+    * Related (Known resource are not registered) Fixed by v0.28.0: https://github.com/aws/karpenter-provider-aws/issues/3794
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The alpha implementation will not consider disruption budgets, or API surface for repairing unhealthy nodes. The team does not have enough data to determine the right level of configuration surface that users will utilize. The goal will be to provide an opinionated mechanism help customer not need to consider the definition of unhealthy nodes, ultimately reducing configuration burden for most customers. **The feature will be gated under an alpha NodeRepair=true feature flag to allow customer to enable the feature. This will  allow for additional feedback from customers, and make subsequent changes in the future that can support the features we have considered out of scope for the Alpha stage.**
```
jonathan-innis, line 17: nit: I'd consider removing the prose and just use bullet points to say, "Here's the things that we aren't considering for this iteration"

- Disruption Budgets
- Customer-Defined Conditions
- Customer-Defined Remediation Time

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Option  1(recommended): Unhealthy condition set cloud provider interface
```
jonathan-innis, line 19:
```suggestion
### Option 1 (recommended): Unhealthy condition set cloud provider interface  
```

nit: Watch your spacing

jonathan-innis, line 19:
```suggestion
### Option 1 (recommended): CloudProvider-Defined RepairPolicies  
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+    // TolerationDuration is the duration the controller will wait
+    // before attempting to terminate nodes that are marked for repair.
+    TolerationDuration time.Duration
+}
+
+type CloudProvider interface {
+  ...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
```
jonathan-innis, line 36:
```suggestion
    RepairPolicy() []v1.RepairStatement
```

If we are considering the whole thing to be a policy, then the underlying parts of the policy are statements

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+    * Related (Stuck on NotReady): https://github.com/aws/karpenter-provider-aws/issues/2439
+
+#### Out of scope
+
+The alpha implementation will not consider disruption budgets, or API surface for repairing unhealthy nodes. The team does not have enough data to determine the right level of configuration surface that users will utilize. The goal will be to provide an opinionated mechanism help customer not need to consider the definition of unhealthy nodes, ultimately reducing configuration burden for most customers. **The feature will be gated under an alpha NodeRepair=true feature flag to allow customer to enable the feature. This will  allow for additional feedback from customers, and make subsequent changes in the future that can support the features we have considered out of scope for the Alpha stage.**
+
+### Option  1(recommended): Unhealthy condition set cloud provider interface  
+
+```
+type RepairPolicy struct {
```
jonathan-innis, line 22:
```suggestion
type RepairStatement struct {
```

Statements are combined together to form a policy

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+type CloudProvider interface {
+  ...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The RepairPolicy will be a set of condition that the Karpenter controller will watch. On any given node, multiple node conditions may exist simultaneously, in those cases we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal:
```
jonathan-innis, line 41:
```suggestion
The RepairPolicy will be a set of conditions that the Karpenter controller will watch. On any given node, multiple node conditions may exist simultaneously, in those cases we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The RepairPolicy will be a set of condition that the Karpenter controller will watch. On any given node, multiple node conditions may exist simultaneously, in those cases we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on to a node
```
jonathan-innis, line 43:
```suggestion
1. A diagnostic agent will add a status condition on a Node 
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The RepairPolicy will be a set of condition that the Karpenter controller will watch. On any given node, multiple node conditions may exist simultaneously, in those cases we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and match unhealthy node condition to cloud provider defined unhealthy condition
```
jonathan-innis, line 44:
```suggestion
2. Karpenter will reconcile nodes and match unhealthy conditions with repair policy statements
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+type CloudProvider interface {
+  ...
+    // IsHealthy returns whether a NodeClaim is considered operating as expected
+    IsHealthy(context.Context, *v1.NodeClaim) (UnhealthyReason, error)
+  ...
+}
+```
+
+`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. `IsHealthy` allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`.
```
jonathan-innis, line 123: The start of this paragraph doesn't make sense -- I think you meant to say "The `IsHealthy` CloudProvider method allows..."

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+}
+```
+
+`IsHealthy` cloud provide methods allows each implementation of Karpenter to determine the healthy state of a NodeClaim and return a specified `UnhealthyReason` indicating that an action on the node is required. `IsHealthy` allows the cloud providers to determine when and on what condition to act on cleaning up the node. Cloud providers will also need to consider short lived failure prior to determining if a node is unhealthy. The `IsHealthy` follows the format defined by `IsDrifted`. 
+
+Main drawback of this option is providers will need to implement much more undifferentiated logic, without having the ability to model duration and remediation on nodes. The reconciliation step for replacing an unhealthy node will look like: 
+
+1. Karpenter will reconcile on the NodeClaim and make a cloud provide call with IsHealthy
+2. If the node is determined to unhealthy, a status condition will be added to the NodeClaim
+3. Disruption controller will force terminate the node claim immediately
```
jonathan-innis, line 129: I still don't follow why this is even an option -- if a cloud provider wanted to do this today they would just write their own controller -- there isn't any value pushing this in the neutral code so this doesn't feel like a real option we would ever consider

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+status:
+  conditions:
+    - lastTransitionTime: "2024-11-04T21:06:06Z"
+      status: "True"
+      reason: <cloud provider returned UnhealthyReason>
+      type: Unhealthy
+    ...
+```
+
+### Option 3: Cloud provider injected unhealthy condition
```
jonathan-innis, line 145: This also doesn't feel like a real option -- given that we are considering forceful termination here after a toleration duration, it feels like the real options are:
1. Each CloudProvider implements this behavior themselves and does the force termination on their own (we did this with the interruption controller)
2. The neutral code implements this behavior and the forceful termination is driven through the neutral code
3. There is some in-between option with `IsHealthy` and a graceful termination option where there is more value in pushing things into the neutral code.

There might also be more value talking about the upside/downside of why we are doing forceful termination and why we are proposing adding a toleration duration

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+1. A diagnostic agent will add a status condition on to a node 
+2. Karpenter will reconcile on nodes and identify unhealthy nodes defined by `UnhealthyConditions` map
+3. Unhealthy condition is added to the NodeClaim
+4. Disruption controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by `UnhealthyConditions` map
+
+## Recommended Solution
+
+The recommended approach is for cloud provider interface should expose a layer of configuration for Karpenter to monitor nodes and take action. I propose we implement option one. Cloud provider can define any set of status conditions that are tracked on the node. An interface approach allows for easy extensibility. Option 2 has the approach of being over generalized as we don’t have use-cases to show the need. All solutions outlined below will be under a `NodeRepair=true`  feature flag to indicate that these feature are not stable.
```
jonathan-innis, line 168: nit: You can probably drop this section -- you already said you are recommending Option 1 above, so this feels unnecessary IMO

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Forceful termination
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future consideration 
+
+The recommended solution includes a limited set of features that will improve the customer experience. There are additional feature we will like to include once more user date can help drive the community decision. These include:
+
+* Disruption controllers for unhealthy nodes
```
jonathan-innis, line 178:
```suggestion
* Disruption controls (budgets, terminationGracePeriod) for unhealthy nodes 
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+### Forceful termination
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future consideration 
+
+The recommended solution includes a limited set of features that will improve the customer experience. There are additional feature we will like to include once more user date can help drive the community decision. These include:
+
+* Disruption controllers for unhealthy nodes 
+* Node Reboot
```
jonathan-innis, line 179:
```suggestion
* Node Reboot (instead of replacement) 
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future consideration 
+
+The recommended solution includes a limited set of features that will improve the customer experience. There are additional feature we will like to include once more user date can help drive the community decision. These include:
+
+* Disruption controllers for unhealthy nodes 
+* Node Reboot 
+* Configuration surface for graceful vs force termination
```
jonathan-innis, line 180:
```suggestion
* Configuration surface for graceful vs forceful termination 
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future consideration 
+
+The recommended solution includes a limited set of features that will improve the customer experience. There are additional feature we will like to include once more user date can help drive the community decision. These include:
+
+* Disruption controllers for unhealthy nodes 
+* Node Reboot 
+* Configuration surface for graceful vs force termination 
+* Additional consideration for availability zone resiliency
```
jonathan-innis, line 181: What does this mean?

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Forceful termination
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future consideration
```
jonathan-innis, line 174:
```suggestion
## Future considerations
```

designs/node-repair.md
```diff
@@ -0,0 +1,182 @@
...
+
+### Forceful termination
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future consideration 
+
+The recommended solution includes a limited set of features that will improve the customer experience. There are additional feature we will like to include once more user date can help drive the community decision. These include:
```
jonathan-innis, line 176: You may not need this statement here -- you could probably shorten this and just say something similar to "Here's what we are considering after releasing this feature: ..."

designs/node-repair.md
engedaam: I missed this when going back and forth between the doc and this PR. Updated.

designs/node-repair.md
engedaam: Dropping this option after an offline discussion

designs/node-repair.md
engedaam: Dropping this option after an offline discussion

designs/node-repair.md
engedaam: This is meant to call out that we will add in safeguards for AZ resiliency as a follow-up

designs/node-repair.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Out of scope
+
+The alpha implementation will not consider these features:
+
+  - Disruption Budgets
+  - Customer-Defined Conditions
+  - Customer-Defined Remediation Time
+
+The team does not have enough data to determine the right level of configuration surface that users will utilize. The goal will be to provide an opinionated mechanism help customer not need to consider the definition of unhealthy nodes, ultimately reducing configuration burden for most customers. **The feature will be gated under an alpha NodeRepair=true feature flag to allow customer to enable the feature. This will  allow for additional feedback from customers, and make subsequent changes in the future that can support the features we have considered out of scope for the Alpha stage.**
```
jonathan-innis, line 23: This paragraph still has awkward grammar -- can you read through it again and then pull out things that either aren't necessary or don't have the correct grammar

designs/node-repair.md
jonathan-innis: I think we have a similar question with expiration -- should we allow users to add `karpenter.sh/do-not-disrupt` to block the deletion of a node when we don't support any budget for the disruption type -- maybe more generally -- should we support budgets for "forceful" disruption mechanism where customers don't want application devs to get in the way but they are fine getting in their own way

designs/node-repair.md
```diff
@@ -0,0 +1,129 @@
...
+type CloudProvider interface {
+  ...
+    // RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter 
+    // to monitor on the node. Customer will need 
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The RepairPolicy will be a set of conditions that the Karpenter controller will watch. On any given node, multiple node conditions may exist simultaneously, in those cases we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal:
```
jonathan-innis, line 47:
```suggestion
The RepairPolicy will contain a set of statements that Karpenter controller will use to watch node conditions. On any given node, multiple node conditions may exist simultaneously. In those cases, we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered terminal: 
```

designs/node-repair.md
```diff
@@ -0,0 +1,129 @@
...
+    RepairPolicy() []v1.RepairPolicy
+  ...
+}
+```
+
+The RepairPolicy will be a set of conditions that the Karpenter controller will watch. On any given node, multiple node conditions may exist simultaneously, in those cases we will chose the shortest `TolerationDuration` for a given condition.  The cloud provider can define compatibility with any node diagnostic agent, and track a list of node unhealthy condition types and a duration period to wait until a unhealthy state is considered a terminal: 
+
+1. A diagnostic agent will add a status condition on a node 
+2. Karpenter will reconcile nodes and match unhealthy conditions with repair policy statements
+3. Node Health controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by TolerationDuration of the unhealthy condition
```
jonathan-innis, line 51:
```suggestion
3. Node Health controller will forcefully terminate the the NodeClaim once the node has been in an unhealthy state for the duration specified by TolerationDuration 
```

nit: you can drop this

designs/node-repair.md
```diff
@@ -0,0 +1,129 @@
...
+```
+
+
+## Forceful termination
+
+For a first iteration approach Karpenter will implement force termination. Today, graceful termination in Karpenter will attempt to wait for pod to be fully drained on a node and all volume attachment to be deleted from a node. This raises the problem that during graceful termination node can be stuck terminating as pod eviction or volume detachment may be broken. In these cases, users will need to take manual action against the node. **For the Alpha implementation, the recommendation will be to non-graceful force termination nodes. Furthermore, unhealthy nodes will not respect the customer configured terminationGracePeriod.**
+
+## Future considerations 
+
+There are additional feature we will like to include once more user date can help drive the community decision. These include:
```
jonathan-innis, line 123:
```suggestion
There are additional features we will consider including after the initial iteration. These include:
```

designs/node-repair.md
engedaam: I think for the time being we should not added it, and see if that is something that customers need

jonathan-innis (approved):
/lgtm
/approve

PR #1773, opened 2024-10-24
chore: Update max delay for NodeClaim lifecycle controller

edibble21:
Fixes #N/A 

**Description**
Updated max delay in Nodeclaim lifecycle counter

make presubmit


pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -95,7 +95,7 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 		WithOptions(controller.Options{
 			RateLimiter: workqueue.NewTypedMaxOfRateLimiter[reconcile.Request](
 				// back off until last attempt occurs ~90 seconds before nodeclaim expiration
-				workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](time.Second, 300*time.Second),
+				workqueue.NewTypedItemExponentialFailureRateLimiter[reconcile.Request](time.Second, time.Minute),
```
jonathan-innis, line 98: @rschalo Can you take a look at this change? I recall that we made this change for a reason but I know you had the most context on why we originally made this change

jonathan-innis (commented):
I think this needs to be updated so that CI passes, but overall it looks good

PR #1774, opened 2024-10-24
[WIP] feat: Drop disruption reason and support dirft sub-reason for budgets

engedaam:
Fixes #N/A 

**Description**


jonathan-innis:
I posted my thoughts in the K8s slack but I had a few high-level concerns with this approach:
1. What does it look like when you intersect the top-level `Drifted` condition with sub-conditions? I'm assuming that the most restrictive one applies but just want to confirm
2. The current model only allows a single condition at a given time which means that the ownership for the condition as well as the relative ordering of conditions gets a little ambiguous. If I have a NodeClaim that has both Subnets drifted and AMIs drifted and I have a budget that is more restrictive on AMIs, I would expect the AMI budget to apply, but today if the subnet condition came first in the checks, we would ignore the more restrictive budget. How are we thinking about filling the gap on this issue?

PR #1786, opened 2024-10-29
feat(provisioning): Add node name to consolidation warning logs

omerap12:
Fixes https://github.com/aws/karpenter-provider-aws/issues/7176

**Description**
Add Node Name in Pod getting evicted.


omerap12:
PTAL @njtran

omerap12:
Sorry for pinging again.. just don't want that we will miss that out.
@engedaam 
@jmdeal

jonathan-innis:
@omerap12 I think I'm a little confused by this change -- doesn't `consolidationWarnings` only get called for pending pods, how would this help for these pods? If they are pending (and aren't scheduled), then by definition they won't be attached to a node.

omerap12:
> @omerap12 I think I'm a little confused by this change -- doesn't `consolidationWarnings` only get called for pending pods, how would this help for these pods? If they are pending (and aren't scheduled), then by definition they won't be attached to a node.

Yes, you are right. probably misunderstanding on my side. thanks for the review :)
/close

PR #1787, opened 2024-10-29
feat: add instance type drift for instance types that are no longer discoverable

njtran:
Fixes #N/A 

**Description**
For instance types that are deprecated, AWS cloud provider doesn't include those instance types. We want to make sure that we can still drift these instances since the absence of the instance type implies that it's not compatible with the NodePool requirements.

make presubmit

### DIFF
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -95,11 +95,11 @@ func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
 		return false
 	}
 	if _, ok := cn.Labels()[v1.CapacityTypeLabelKey]; !ok {
-		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", v1.CapacityTypeLabelKey))...)
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("Node does not have label %q", v1.CapacityTypeLabelKey))...)
 		return false
 	}
 	if _, ok := cn.Labels()[corev1.LabelTopologyZone]; !ok {
-		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", corev1.LabelTopologyZone))...)
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("Node does not have label %q", corev1.LabelTopologyZone))...)
 		return false
 	}
 	if cn.nodePool.Spec.Disruption.ConsolidateAfter.Duration == nil {
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -90,6 +90,16 @@ func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("Instance Type %q not found", cn.Labels()[corev1.LabelInstanceTypeStable]))...)
 		return false
 	}
+	// skip any candidates that don't have capacity type as we can't look up pricing
+	if _, ok := cn.Labels()[v1.CapacityTypeLabelKey]; !ok {
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", v1.CapacityTypeLabelKey))...)
+		return false
+	}
+	// skip any candidates that don't have zone info as we can't look up pricing
+	if _, ok := cn.Labels()[corev1.LabelTopologyZone]; !ok {
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", corev1.LabelTopologyZone))...)
+		return false
+	}
 	if cn.nodePool.Spec.Disruption.ConsolidateAfter.Duration == nil {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("NodePool %q has consolidation disabled", cn.nodePool.Name))...)
 		return false
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -85,17 +85,19 @@ func (c *consolidation) markConsolidated() {
 
 // ShouldDisrupt is a predicate used to filter candidates
 func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
-	// skip any candidates that we can't determine the instance type of
+	// We need the following to know what the price of the instance for price comparison. If one of these doesn't exist, we can't
+	// compute consolidation decisions for this candidate.
+	// 1. Instance Type
+	// 2. Capacity Type
+	// 3. Zone
 	if cn.instanceType == nil {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("Instance Type %q not found", cn.Labels()[corev1.LabelInstanceTypeStable]))...)
 		return false
 	}
-	// skip any candidates that don't have capacity type as we can't look up pricing
 	if _, ok := cn.Labels()[v1.CapacityTypeLabelKey]; !ok {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", v1.CapacityTypeLabelKey))...)
 		return false
 	}
-	// skip any candidates that don't have zone info as we can't look up pricing
 	if _, ok := cn.Labels()[corev1.LabelTopologyZone]; !ok {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", corev1.LabelTopologyZone))...)
 		return false
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -85,6 +85,11 @@ func (c *consolidation) markConsolidated() {
 
 // ShouldDisrupt is a predicate used to filter candidates
 func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
+	// skip any candidates that we can't determine the instance type of
+	if cn.instanceType == nil {
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("Instance Type %q not found", cn.Labels()[corev1.LabelInstanceTypeStable]))...)
+		return false
+	}
 	if cn.nodePool.Spec.Disruption.ConsolidateAfter.Duration == nil {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("NodePool %q has consolidation disabled", cn.nodePool.Name))...)
 		return false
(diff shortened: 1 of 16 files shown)


pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -81,6 +84,21 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1.NodePool, nodeClaim
 	}); reason != "" {
 		return reason, nil
 	}
+	// Mark the NodeClaim as drifted due to a stale instance type if
+	// 1. The NodeClaim doesn't have the instance type label
+	// 2. The NodeClaim has an instance type that doesn't exist in the cloudprovider instance types
+	// 3. There are no offerings that match the requirements
+	its, err := d.cloudProvider.GetInstanceTypes(ctx, nodePool)
```
engedaam, line 91: nit Any thought on moving this check to the separate function like the other drift checks?

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -81,6 +84,21 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1.NodePool, nodeClaim
 	}); reason != "" {
 		return reason, nil
 	}
+	// Mark the NodeClaim as drifted due to a stale instance type if
```
engedaam, line 87: Can we define what we mean by a state instance type? It would be nice to have the definition as a comment

pkg/controllers/state/statenode.go
```diff
@@ -195,9 +195,6 @@ func (in *StateNode) ValidateNodeDisruptable(ctx context.Context, kubeClient cli
 	}
 	// check whether the node has all the labels we need
 	for _, label := range []string{
-		v1.CapacityTypeLabelKey,
```
engedaam, line 198: Why remove these labels?

pkg/controllers/disruption/suite_test.go
```diff
@@ -1676,7 +1676,7 @@ var _ = Describe("Candidate Filtering", func() {
 		Expect(err.Error()).To(Equal(fmt.Sprintf("nodepool %q can't be resolved for state node", nodePool.Name)))
 		Expect(recorder.DetectedEvent(fmt.Sprintf("Cannot disrupt Node: NodePool %q not found", nodePool.Name))).To(BeTrue())
 	})
-	It("should not consider candidates that do not have the karpenter.sh/capacity-type label", func() {
+	It("should consider candidates that do not have the karpenter.sh/capacity-type label", func() {
```
engedaam, line 1679: I'm a little confused here. Aren't we making sure that we don't consider these nodes that are missing the labels listed in the changes of this file

njtran: Maybe I could factor it a bit better in the future (but this would require a bit of an overhaul on the candidate struct type), but effectively we just check the labels on the candidate only for Consolidation when it matters. We wouldn't do it for drift or other eventual methods.

pkg/controllers/nodeclaim/disruption/drift.go
njtran: Does the 3 definitions below help? It's mostly "does the instance type with the offering exist anymore in the instance type offerings cache", if not it's stale

pkg/controllers/nodeclaim/disruption/drift.go
njtran: Would probably be a bit cleaner, will do

pkg/controllers/state/statenode.go
njtran: We don't need to check for these to see if a node is explicitly disruptable. It only matters for Consolidation now and is done in the shouldDisrupt step

pkg/controllers/state/statenode.go
```diff
@@ -195,9 +195,6 @@ func (in *StateNode) ValidateNodeDisruptable(ctx context.Context, kubeClient cli
 	}
 	// check whether the node has all the labels we need
 	for _, label := range []string{
```
jonathan-innis, line 197: nit: You almost don't need to use a `for` loop anymore if you are just checking a single key

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -81,6 +84,21 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1.NodePool, nodeClaim
...
+	// 2. The NodeClaim has an instance type that doesn't exist in the cloudprovider instance types
+	// 3. There are no offerings that match the requirements
+	its, err := d.cloudProvider.GetInstanceTypes(ctx, nodePool)
+	if err != nil {
+		return "", err
+	}
+	it, ok := lo.Find(its, func(it *cloudprovider.InstanceType) bool {
+		return it.Name == nodeClaim.Labels[corev1.LabelInstanceTypeStable]
+	})
+	if !ok || len(it.Offerings.Compatible(scheduling.NewLabelRequirements(nodeClaim.Labels))) == 0 {
```
jonathan-innis, line 98: Does the offerings compatibility in the NodeClaim compatibility? I forget if these are scoped together or not?

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -25,14 +25,17 @@ import (
...
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
 const (
-	NodePoolDrifted     cloudprovider.DriftReason = "NodePoolDrifted"
-	RequirementsDrifted cloudprovider.DriftReason = "RequirementsDrifted"
+	NodePoolDrifted        cloudprovider.DriftReason = "NodePoolDrifted"
+	RequirementsDrifted    cloudprovider.DriftReason = "RequirementsDrifted"
+	StaleInstanceTypeDrift cloudprovider.DriftReason = "StaleInstanceTypeDrifted"
```
jonathan-innis, line 38:
```suggestion
	StaleInstanceTypeDrift cloudprovider.DriftReason = "InstanceTypeNotFound"/"InstanceTypeUnavailable"
```

There was no drift here so I wonder if the reason should contain "Drifted in it"

pkg/controllers/disruption/types.go
```diff
@@ -90,12 +90,8 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 		recorder.Publish(disruptionevents.Blocked(node.Node, node.NodeClaim, fmt.Sprintf("NodePool %q not found", nodePoolName))...)
 		return nil, fmt.Errorf("nodepool %q can't be resolved for state node", nodePoolName)
 	}
+	// We only care if instanceType in non-empty consolidation to do price-comparison.
 	instanceType := instanceTypeMap[node.Labels()[corev1.LabelInstanceTypeStable]]
```
jonathan-innis, line 94: I wonder if we should even populate this as part of building the candidates here or if this should all be encapsulated inside of the consolidation bits because now we have components of our candidate that can be nil (which feels a bit awkward to me) e.g. capacity type, zone, instance type and if you're the caller or we extended this later it's unclear that these are going to be nil or empty when I'm grabbing them

pkg/controllers/disruption/helpers.go
```diff
@@ -156,7 +156,7 @@ func GetCandidates(ctx context.Context, cluster *state.Cluster, kubeClient clien
 		return cn, e == nil
 	})
 	// Filter only the valid candidates that we should disrupt
-	return lo.Filter(candidates, func(c *Candidate, _ int) bool { return shouldDeprovision(ctx, c) }), nil
+	return lo.Filter(candidates, func(c *Candidate, _ int) bool { return shouldDisrupt(ctx, c) }), nil
```
jonathan-innis, line 159: nice!

pkg/controllers/disruption/consolidation.go
```diff
@@ -85,6 +85,11 @@ func (c *consolidation) markConsolidated() {
 
 // ShouldDisrupt is a predicate used to filter candidates
 func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
+	// skip any candidates that we can't determine the instance type of
```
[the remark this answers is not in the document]
jonathan-innis: I guess there's some amount of capacity type filtering but it might be more clear if we gave a bit more info on what we were using this information for in the first place

pkg/controllers/disruption/types.go
njtran: You have to pass in a lot of NodePool -> Instance Type -> Offerings map data if you pass it through into the consolidation calls. It's a bit more of an involved change, but I think this generally makes sense to me that you'd  pass in a nil if the instance type doesn't exist. I do agree it expands the possibility that we mess up and get a nil dereference. WDYT

jonathan-innis: I'm fine with a fast-follow -- it definitely doesn't completely encapsulate the behavior right now but I'm also fine not churning on it

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -81,13 +84,43 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1.NodePool, nodeClaim
...
+// InstanceTypeNotFoundDrift is computed by looking directly at the list of instance types and comparing the NodeClaim
+// to the instance types and its offerings.
+// 1. The NodeClaim doesn't have the instance type label
+// 2. The NodeClaim has an instance type that doesn't exist in the cloudprovider instance types
+// 3. There are no offerings that match the requirements
+func instanceTypeNotFound(its []*cloudprovider.InstanceType, nodeClaim *v1.NodeClaim) cloudprovider.DriftReason {
+	it, ok := lo.Find(its, func(it *cloudprovider.InstanceType) bool {
+		return it.Name == nodeClaim.Labels[corev1.LabelInstanceTypeStable]
+	})
+	if !ok || !it.Offerings.HasCompatible(scheduling.NewLabelRequirements(nodeClaim.Labels)) {
```
jonathan-innis, line 118: nit: Consider a comment that says "Check that the offering exists for the NodeClaim capacity type/zone" since it's not immediately clear from reading it at first-glance

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -25,14 +25,17 @@ import (
...
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
 const (
-	NodePoolDrifted     cloudprovider.DriftReason = "NodePoolDrifted"
-	RequirementsDrifted cloudprovider.DriftReason = "RequirementsDrifted"
+	NodePoolDrifted           cloudprovider.DriftReason = "NodePoolDrifted"
+	RequirementsDrifted       cloudprovider.DriftReason = "RequirementsDrifted"
+	InstanceTypeNotFoundDrift cloudprovider.DriftReason = "InstanceTypeNotFound"
```
jonathan-innis, line 38:
```suggestion
	InstanceTypeNotFound cloudprovider.DriftReason = "InstanceTypeNotFound"
```

nit: The other types don't append drift at the end -- if you wanted to be more inline with K8s upstream, you could consider doing `NodePoolDriftedDriftReason`, `RequirementsDriftedDriftReason`, etc.

pkg/controllers/disruption/consolidation.go
```diff
@@ -85,6 +85,23 @@ func (c *consolidation) markConsolidated() {
...
+	if cn.instanceType == nil {
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("Instance Type %q not found", cn.Labels()[corev1.LabelInstanceTypeStable]))...)
+		return false
+	}
+	if _, ok := cn.Labels()[v1.CapacityTypeLabelKey]; !ok {
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", v1.CapacityTypeLabelKey))...)
+		return false
+	}
+	if _, ok := cn.Labels()[corev1.LabelTopologyZone]; !ok {
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("node does not have label %q", corev1.LabelTopologyZone))...)
```
jonathan-innis, line 102: nit: We have a mix of capatalization and non-capatalization in the event messages that you have here -- I believe that we have a standard around this but we should check which one we want to use across all of these

pkg/controllers/nodeclaim/disruption/drift.go
njtran: It's in the paragraph comment above, but understand that'll probably get scanned through

jonathan-innis (approved):
/lgtm
/approve

PR #1788, opened 2024-10-30
feat: add log for scheduling progress for long running scheduling sim…

njtran:
…ulations

Fixes #N/A 

**Description**
This adds a log in case provisioning takes too long, and the user wants to know if scheduling is occurring

locally


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -205,10 +209,19 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
+	startTime := s.clock.Now()
+	lastLogTime := s.clock.Now()
+	batchSize := len(q.pods)
 	for {
 		QueueDepth.With(
 			prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx), schedulingIDLabel: string(s.id)},
 		).Set(float64(len(q.pods)))
+
+		if s.clock.Since(lastLogTime) > time.Minute {
+			log.FromContext(ctx).Info("printing scheduling progress", "podsScheduled", batchSize-len(q.pods), "podsLeft", len(q.pods), "timeSpent", s.clock.Since(startTime).Truncate(time.Second))
```
jonathan-innis, line 222:
```suggestion
			log.FromContext(ctx).Info("scheduling pods...", "pods-scheduled", batchSize-len(q.pods), "pods-left", len(q.pods), "duration", fmt.Sprintf("%ss", s.clock.Since(startTime).Truncate(time.Second)))
```

jonathan-innis (commented):
A few comments on the wording

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -205,10 +209,19 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
+	startTime := s.clock.Now()
+	lastLogTime := s.clock.Now()
+	batchSize := len(q.pods)
 	for {
 		QueueDepth.With(
 			prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx), schedulingIDLabel: string(s.id)},
 		).Set(float64(len(q.pods)))
+
+		if s.clock.Since(lastLogTime) > time.Minute {
+			log.FromContext(ctx).Info("printing scheduling progress", "pods-scheduled", batchSize-len(q.pods), "pods-left", len(q.pods), "time-spent", s.clock.Since(startTime).Truncate(time.Second))
```
jonathan-innis, line 222:
```suggestion
			log.FromContext(ctx).Info("printing scheduling progress", "pods-scheduled", batchSize-len(q.pods), "pods-left", len(q.pods), "duration", s.clock.Since(startTime).Truncate(time.Second))
```

I still think it's more ergonomic if we go with "scheduling pods..." to indicate that we are still making progress. Also `duration` feels more ergonomic to me compared to `time-spent` since it's consistent with wording elsewhere in the code.

Also, can we see an example of this happening? I'm also curious if we want to just show this during provisioning -- I'm worried that this is going to be extra noisy with provisioning and disruption running at the same time.

njtran: This fires once a minute, per provisioning controller. Don't think it'll be too noisy.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -205,10 +209,19 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
+	startTime := s.clock.Now()
+	lastLogTime := s.clock.Now()
+	batchSize := len(q.pods)
 	for {
 		QueueDepth.With(
 			prometheus.Labels{ControllerLabel: injection.GetControllerName(ctx), schedulingIDLabel: string(s.id)},
 		).Set(float64(len(q.pods)))
+
+		if s.clock.Since(lastLogTime) > time.Minute {
+			log.FromContext(ctx).Info("printing scheduling progress", "pods-scheduled", batchSize-len(q.pods), "pods-left", len(q.pods), "duration", s.clock.Since(startTime).Truncate(time.Second))
```
jmdeal, line 222: Should we use `WithValues` here? I also think we should include scheduling-id to help correlate with metrics.
```suggestion
			log.FromContext(ctx).WithValues("pods-scheduled", batchSize-len(q.pods), "pods-remaining", len(q.pods), "duration", s.clock.Since(startTime).Truncate(time.Second), "scheduling-id", string(s.id)).Info("scheduling progress")
```

jmdeal, line 222: Should this be a Debug log? If we were only logging this for the provisioner, I think it's rare enough that keeping it as Info is appropriate, but I'm wondering if it will be spammy due to parallel disruption.

njtran: 2 a minute at peak times I don't think is spammy

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: Per scheduler -- not provisioning controller and per-scheduling loop. With multiple scheduling loops, I think this could get extremely noisy

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: We discussed -- we don't add a real logger for any of our disruption loops so this isn't going to print anything out for anything but provisioning.

jonathan-innis (approved):
/lgtm
/approve

PR #1793, opened 2024-10-30
feat: Node Repair implementation

engedaam:
Fixes #N/A 

**Description**
- RFC: https://github.com/kubernetes-sigs/karpenter/pull/1768
- This PR is the implementation of the recommend solution defined in the node repair RFC 
- Defining a cloud provider interface `RepairPolicy` that will support node conditions that Karpenter will forcefully terminate nodes. The cloud provider policies will be unhealthy conditions a node can enter and the duration for Karpenter to react. 

- `make resubmit`

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -23,6 +23,7 @@ import (
 	"math"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/awslabs/operatorpkg/status"
 	"github.com/samber/lo"
@@ -40,6 +41,16 @@ var (
 )
 
 type DriftReason string
+
+type RepairPolicy struct {
+	// Type of unhealthy state that is found on the node
+	Type corev1.NodeConditionType
+	// Status condition of when a node is unhealthy
+	Status corev1.ConditionStatus
+	// TolerationDuration is the duration the controller will wait
+	// before attempting to terminate nodes that are underutilized.
+	TolerationDuration time.Duration
+}
 
 // CloudProvider interface is implemented by cloud providers to support provisioning.
 type CloudProvider interface {
@@ -63,6 +74,9 @@ type CloudProvider interface {
 	// IsDrifted returns whether a NodeClaim has drifted from the provisioning requirements
 	// it is tied to.
 	IsDrifted(context.Context, *v1.NodeClaim) (DriftReason, error)
+	// RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter
+	// to monitor on the node.
+	RepairPolicy() []RepairPolicy
 	// Name returns the CloudProvider implementation name.
 	Name() string
 	// GetSupportedNodeClasses returns CloudProvider NodeClass that implements status.Object
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -42,13 +42,13 @@ var (
 
 type DriftReason string
 
-type RepairPolicy struct {
+type RepairStatement struct {
 	// Type of unhealthy state that is found on the node
 	Type corev1.NodeConditionType
-	// Status condition of when a node is unhealthy
+	// Status condition when a node is unhealthy
 	Status corev1.ConditionStatus
 	// TolerationDuration is the duration the controller will wait
-	// before attempting to terminate nodes that are underutilized.
+	// before force terminating nodes that are unhealthy.
 	TolerationDuration time.Duration
 }
 
@@ -76,7 +76,7 @@ type CloudProvider interface {
 	IsDrifted(context.Context, *v1.NodeClaim) (DriftReason, error)
 	// RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter
 	// to monitor on the node.
-	RepairPolicy() []RepairPolicy
+	RepairPolicy() []RepairStatement
 	// Name returns the CloudProvider implementation name.
 	Name() string
 	// GetSupportedNodeClasses returns CloudProvider NodeClass that implements status.Object
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -42,11 +42,11 @@ var (
 
 type DriftReason string
 
-type RepairStatement struct {
-	// Type of unhealthy state that is found on the node
-	Type corev1.NodeConditionType
-	// Status condition when a node is unhealthy
-	Status corev1.ConditionStatus
+type RepairPolicy struct {
+	// ConditionType of unhealthy state that is found on the node
+	ConditionType corev1.NodeConditionType
+	// ConditionStatus condition when a node is unhealthy
+	ConditionStatus corev1.ConditionStatus
 	// TolerationDuration is the duration the controller will wait
 	// before force terminating nodes that are unhealthy.
 	TolerationDuration time.Duration
@@ -76,7 +76,7 @@ type CloudProvider interface {
 	IsDrifted(context.Context, *v1.NodeClaim) (DriftReason, error)
 	// RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter
 	// to monitor on the node.
-	RepairPolicy() []RepairStatement
+	RepairPolicies() []RepairPolicy
 	// Name returns the CloudProvider implementation name.
 	Name() string
 	// GetSupportedNodeClasses returns CloudProvider NodeClass that implements status.Object
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -98,8 +98,8 @@ func NewControllers(
 		health.NewController(kubeClient, cloudProvider, clock),
 	}
 
-	// The cloud provider must define status conation for the node repair controller to used for detecting unhealthy nodes
-	if len(cloudProvider.RepairPolicy()) != 0 && options.FromContext(ctx).FeatureGates.NodeRepair {
+	// The cloud provider must define status conditions for the node repair controller to use to detect unhealthy nodes
+	if len(cloudProvider.RepairPolicies()) != 0 && options.FromContext(ctx).FeatureGates.NodeRepair {
 		controllers = append(controllers, health.NewController(kubeClient, cloudProvider, clock))
 	}
 
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -32,6 +32,7 @@ import (
 	metricsnode "sigs.k8s.io/karpenter/pkg/controllers/metrics/node"
 	metricsnodepool "sigs.k8s.io/karpenter/pkg/controllers/metrics/nodepool"
 	metricspod "sigs.k8s.io/karpenter/pkg/controllers/metrics/pod"
+	"sigs.k8s.io/karpenter/pkg/controllers/node/health"
 	"sigs.k8s.io/karpenter/pkg/controllers/node/termination"
 	"sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator"
 	nodeclaimconsistency "sigs.k8s.io/karpenter/pkg/controllers/nodeclaim/consistency"
@@ -90,5 +91,6 @@ func NewControllers(
 		status.NewController[*v1.NodeClaim](kubeClient, mgr.GetEventRecorderFor("karpenter"), status.EmitDeprecatedMetrics),
 		status.NewController[*v1.NodePool](kubeClient, mgr.GetEventRecorderFor("karpenter"), status.EmitDeprecatedMetrics),
 		status.NewGenericObjectController[*corev1.Node](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		health.NewController(kubeClient, cloudProvider, clock),
 	}
 }
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -23,7 +23,6 @@ import (
 	"github.com/awslabs/operatorpkg/status"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	corev1 "k8s.io/api/core/v1"
@@ -99,11 +98,9 @@ func NewControllers(
 		health.NewController(kubeClient, cloudProvider, clock),
 	}
 
-	// The cloud proivder must define status condation for the node repair controller to used for dectecting unhealthy nodes
-	if len(cloudProvider.RepairPolicy()) != 0 && !options.FromContext(ctx).FeatureGates.NodeRepair {
+	// The cloud provider must define status conation for the node repair controller to used for detecting unhealthy nodes
+	if len(cloudProvider.RepairPolicy()) != 0 && options.FromContext(ctx).FeatureGates.NodeRepair {
 		controllers = append(controllers, health.NewController(kubeClient, cloudProvider, clock))
-	} else {
-		log.FromContext(ctx).V(1).Info("node repair has been disabled")
 	}
 
 	return controllers
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -17,10 +17,13 @@ limitations under the License.
 package controllers
 
 import (
+	"context"
+
 	"github.com/awslabs/operatorpkg/controller"
 	"github.com/awslabs/operatorpkg/status"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 
 	corev1 "k8s.io/api/core/v1"
@@ -49,9 +52,11 @@ import (
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/controllers/state/informer"
 	"sigs.k8s.io/karpenter/pkg/events"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 )
 
 func NewControllers(
+	ctx context.Context,
 	mgr manager.Manager,
 	clock clock.Clock,
 	kubeClient client.Client,
@@ -64,7 +69,7 @@ func NewControllers(
 	evictionQueue := terminator.NewQueue(kubeClient, recorder)
 	disruptionQueue := orchestration.NewQueue(kubeClient, recorder, cluster, clock, p)
 
-	return []controller.Controller{
+	controllers := []controller.Controller{
 		p, evictionQueue, disruptionQueue,
 		disruption.NewController(clock, kubeClient, p, cloudProvider, recorder, cluster, disruptionQueue),
 		provisioning.NewPodController(kubeClient, p),
@@ -93,4 +98,13 @@ func NewControllers(
 		status.NewGenericObjectController[*corev1.Node](kubeClient, mgr.GetEventRecorderFor("karpenter")),
 		health.NewController(kubeClient, cloudProvider, clock),
 	}
+
+	// The cloud proivder must define status condation for the node repair controller to used for dectecting unhealthy nodes
+	if len(cloudProvider.RepairPolicy()) != 0 && !options.FromContext(ctx).FeatureGates.NodeRepair {
+		controllers = append(controllers, health.NewController(kubeClient, cloudProvider, clock))
+	} else {
+		log.FromContext(ctx).V(1).Info("node repair has been disabled")
+	}
+
+	return controllers
 }
--- a/pkg/controllers/node/health/controller.go
+++ b/pkg/controllers/node/health/controller.go
@@ -21,13 +21,12 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
 	"k8s.io/utils/clock"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -63,67 +62,62 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 }
 
 func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
-	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("getting nodeclaim, %w", err)
-	}
-
 	ctx = injection.WithControllerName(ctx, "node.health")
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("nodeclaim", nodeClaim.Name))
-	nodeHealthCondition := corev1.NodeCondition{}
-	foundCloudProviderPolicy := cloudprovider.RepairStatement{}
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(node.Namespace, node.Name)))
 
 	// Validate that the node is owned by us and is not being deleted
-	if !node.GetDeletionTimestamp().IsZero() || !controllerutil.ContainsFinalizer(node, v1.TerminationFinalizer) {
-		return reconcile.Result{}, nil
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+	if err != nil {
+		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
 	}
 
-	for _, policy := range c.cloudProvider.RepairPolicy() {
-		nodeHealthCondition = nodeutils.GetCondition(node, policy.Type)
-		if nodeHealthCondition.Status == policy.Status {
-			// found unhealthy condition on the node
-			foundCloudProviderPolicy = policy
-			break
+	log.FromContext(ctx).V(1).Info("found nodeclaim")
+	// If find if a node is unhealthy
+	healthCondition, foundHealthCondition := lo.Find(c.cloudProvider.RepairPolicies(), func(policy cloudprovider.RepairPolicy) bool {
+		nodeCondition := nodeutils.GetCondition(node, policy.ConditionType)
+		for _, condition := range node.Status.Conditions {
+			if condition.Type == policy.ConditionType && nodeCondition.Status == policy.ConditionStatus {
+				terminationTime := condition.LastTransitionTime.Add(policy.TolerationDuration)
+				return !c.clock.Now().Before(terminationTime)
+			}
 		}
+		return false
+	})
+
+	log.FromContext(ctx).V(1).Info(fmt.Sprintf("found status: %v", foundHealthCondition))
+	// If the Node is unhealthy, but has not reached it's full toleration disruption
+	// requeue at the termination time of the unhealthy node
+	terminationTime := nodeutils.GetCondition(node, healthCondition.ConditionType).LastTransitionTime.Add(healthCondition.TolerationDuration)
+	log.FromContext(ctx).V(1).Info(fmt.Sprintf("time now: %s, terminate: %s", c.clock.Now().String(), terminationTime.String()))
+	if !foundHealthCondition {
+		log.FromContext(ctx).V(1).Info("termination time")
+		return reconcile.Result{RequeueAfter: terminationTime.Sub(c.clock.Now())}, nil
 	}
 
-	// From here there are three scenarios to handle:
-	// 1. If node is healthy, exit node repair loop
-	if foundCloudProviderPolicy.Type == "" {
-		return reconcile.Result{}, nil
-	}
-
-	// 2. If the Node is unhealthy, but has not reached it's full toleration disruption, exit the loop
-	disruptionTime := nodeHealthCondition.LastTransitionTime.Add(foundCloudProviderPolicy.TolerationDuration)
-	if c.clock.Now().Before(disruptionTime) {
-		return reconcile.Result{RequeueAfter: disruptionTime.Sub(c.clock.Now())}, nil
-	}
-
+	// For unhealthy past the tolerationDisruption window we can forcefully terminate the node
 	if err := c.annotateTerminationGracePeriod(ctx, nodeClaim); err != nil {
-		return reconcile.Result{}, fmt.Errorf("annotated termination grace period on nodeclaim, %w", err)
-	}
-
-	// 3. Otherwise, if the Node is unhealthy and past it's tolerationDisruption window we can forcefully terminate the node
-	if err := c.kubeClient.Delete(ctx, nodeClaim); err != nil {
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
-	// 4. The deletion timestamp has successfully been set for the Node, update relevant metrics.
+	if err := c.kubeClient.Delete(ctx, nodeClaim); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// The deletion timestamp has successfully been set for the Node, update relevant metrics.
 	log.FromContext(ctx).V(1).Info("deleting unhealthy node")
-	metrics.NodeClaimsDisruptedTotal.With(prometheus.Labels{
-		metrics.ReasonLabel:       metrics.UnhealthyReason,
-		metrics.NodePoolLabel:     nodeClaim.Labels[v1.NodePoolLabelKey],
-		metrics.CapacityTypeLabel: nodeClaim.Labels[v1.CapacityTypeLabelKey],
-	}).Inc()
+	metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
+		metrics.ReasonLabel:       string(healthCondition.ConditionType),
+		metrics.NodePoolLabel:     node.Labels[v1.NodePoolLabelKey],
+		metrics.CapacityTypeLabel: node.Labels[v1.CapacityTypeLabelKey],
+	})
 	return reconcile.Result{}, nil
 }
 
 func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeClaim *v1.NodeClaim) error {
 	stored := nodeClaim.DeepCopy()
-	terminationTime := c.clock.Now().Format(time.RFC3339)
-	nodeClaim.ObjectMeta.Annotations = lo.Assign(nodeClaim.ObjectMeta.Annotations, map[string]string{v1.NodeClaimTerminationTimestampAnnotationKey: terminationTime})
+	nodeClaim.ObjectMeta.Annotations = lo.Assign(nodeClaim.ObjectMeta.Annotations, map[string]string{v1.NodeClaimTerminationTimestampAnnotationKey: c.clock.Now().Format(time.RFC3339)})
 
 	if err := c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
-		return client.IgnoreNotFound(err)
+		return err
 	}
 
 	return nil
--- a/pkg/controllers/node/health/controller.go
+++ b/pkg/controllers/node/health/controller.go
@@ -18,7 +18,6 @@ package health
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/samber/lo"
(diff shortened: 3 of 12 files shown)


pkg/controllers/controllers.go
```diff
@@ -87,5 +93,16 @@ func NewControllers(
...
 		status.NewController[*v1.NodePool](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		// status.NewGenericObjectController[*corev1.Node](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		// status.NewGenericObjectController[*corev1.Pod](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+	}
+
+	// The cloud proivder must define status condation for the node repair controller to used for dectecting unhealthy nodes
+	if len(cloudProvider.RepairPolicy()) != 0 && !options.FromContext(ctx).FeatureGates.NodeRepair {
+		controllers = append(controllers, health.NewController(kubeClient, cloudProvider, clock))
+	} else {
+		log.FromContext(ctx).V(1).Info("node repair has been disabled")
```
jonathan-innis, line 104: I don't think you want to log that it's disabled -- you probably want to log the opposite since this is disabled by default.

Also, you generally want to log when you are doing actions, not when you aren't doing them

pkg/controllers/controllers.go
```diff
@@ -87,5 +93,16 @@ func NewControllers(
 		nodeclaimdisruption.NewController(clock, kubeClient, cloudProvider),
 		status.NewController[*v1.NodeClaim](kubeClient, mgr.GetEventRecorderFor("karpenter")),
 		status.NewController[*v1.NodePool](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		// status.NewGenericObjectController[*corev1.Node](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+		// status.NewGenericObjectController[*corev1.Pod](kubeClient, mgr.GetEventRecorderFor("karpenter")),
+	}
+
+	// The cloud proivder must define status condation for the node repair controller to used for dectecting unhealthy nodes
+	if len(cloudProvider.RepairPolicy()) != 0 && !options.FromContext(ctx).FeatureGates.NodeRepair {
```
jonathan-innis, line 101:
```suggestion
	if len(cloudProvider.RepairPolicy()) != 0 && options.FromContext(ctx).FeatureGates.NodeRepair {
```

We want to add the controller when we enable the feature flag

pkg/cloudprovider/types.go
```diff
@@ -41,6 +42,16 @@ var (
 
 type DriftReason string
 
+type RepairStatements struct {
```
jonathan-innis, line 45:
```suggestion
type RepairStatement struct {
```

pkg/cloudprovider/types.go
```diff
@@ -63,6 +74,9 @@ type CloudProvider interface {
 	// IsDrifted returns whether a NodeClaim has drifted from the provisioning requirements
 	// it is tied to.
 	IsDrifted(context.Context, *v1.NodeClaim) (DriftReason, error)
+	// RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter
+	// to monitor on the node.
+	RepairPolicy() []RepairStatements
```
jonathan-innis, line 79:
```suggestion
	RepairPolicy() []RepairStatement
```

pkg/cloudprovider/types.go
```diff
@@ -41,6 +42,16 @@ var (
 
 type DriftReason string
 
+type RepairStatements struct {
+	// Type of unhealthy state that is found on the node
+	Type corev1.NodeConditionType
+	// Status condition of when a node is unhealthy
+	Status corev1.ConditionStatus
+	// TolerationDuration is the duration the controller will wait
+	// before attempting to terminate nodes that are underutilized.
```
jonathan-innis, line 51:
```suggestion
	// before force terminating nodes that are unhealthy.
```

pkg/cloudprovider/types.go
```diff
@@ -41,6 +42,16 @@ var (
 
 type DriftReason string
 
+type RepairStatements struct {
+	// Type of unhealthy state that is found on the node
+	Type corev1.NodeConditionType
+	// Status condition of when a node is unhealthy
```
jonathan-innis, line 48:
```suggestion
	// Status condition when a node is unhealthy
```

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+// Controller for the resource
+type Controller struct {
+	clock         clock.Clock
+	kubeClient    client.Client
+	cloudProvider cloudprovider.CloudProvider
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, clock clock.Clock) *Controller {
+	return &Controller{
```
jonathan-innis, line 50: nit: the order of that you are doing the assignment here doesn't match the order you have declared the same parameters on L42 -- why switch them up? Would have expected args in the order: clock, kubeClient, cloudProvider -- since I think this matches the priority order in other controllers

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+		if nodeHealthCondition.Status == policy.Status {
+			// found unhealthy condition on the node
+			foundCloudProviderPolicy = policy
+			break
+		}
+	}
+
+	// From here there are three scenarios to handle:
+	// 1. If node is healthy, exit node repair loop
+	if foundCloudProviderPolicy.Type == "" {
```
jonathan-innis, line 84: I'm confused -- didn't you already validate that you found the type on the node -- so what is this check for?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+		For(&corev1.Node{}).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "node.health")
+	nodeHealthCondition := corev1.NodeCondition{}
+	foundCloudProviderPolicy := cloudprovider.RepairStatements{}
+
+	if !node.GetDeletionTimestamp().IsZero() {
```
jonathan-innis, line 69: How do we know this node is owned by us?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+	nodeClaims, err := nodeutils.GetNodeClaims(ctx, node, c.kubeClient)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("listing nodeclaims, %w", err)
+	}
+	if err := c.annotateTerminationGracePeriod(ctx, nodeClaims[0]); err != nil {
+		return reconcile.Result{}, fmt.Errorf("annotated termination grace period on nodeclaim, %w", err)
+	}
+
+	// 3. Otherwise, if the Node is unhealthy and past it's tolerationDisruption window we can forcefully terminate the node
+	if err := c.kubeClient.Delete(ctx, node); err != nil {
```
jonathan-innis, line 103: Why did we get the NodeClaim and then delete the node? I thought the intention of grabbing the NodeClaim was to get the NodeClaim associated with the node. I think we should adhere to the pattern that we are doing elsewhere which is to grab the NodeClaim, and then use the NodeClaim to do the deletion -- we do the same thing in the karpenter-provider-aws repo for the interruption controller

jonathan-innis, line 103: What about ignoring not found?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// 2. If the Node is unhealthy, but has not reached it's full toleration disruption, exit the loop
+	disruptionTime := nodeHealthCondition.LastTransitionTime.Add(foundCloudProviderPolicy.TolerationDuration)
+	if c.clock.Now().Before(disruptionTime) {
+		return reconcile.Result{RequeueAfter: disruptionTime.Sub(c.clock.Now())}, nil
+	}
+
+	nodeClaims, err := nodeutils.GetNodeClaims(ctx, node, c.kubeClient)
```
jonathan-innis, line 94: Should we ignore cases where we either don't find a nodeclaim or get duplicate nodeclaims?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+	if err := c.annotateTerminationGracePeriod(ctx, nodeClaims[0]); err != nil {
+		return reconcile.Result{}, fmt.Errorf("annotated termination grace period on nodeclaim, %w", err)
+	}
+
+	// 3. Otherwise, if the Node is unhealthy and past it's tolerationDisruption window we can forcefully terminate the node
+	if err := c.kubeClient.Delete(ctx, node); err != nil {
+		return reconcile.Result{}, err
+	}
+	// 4. The deletion timestamp has successfully been set for the Node, update relevant metrics.
+	log.FromContext(ctx).V(1).Info("deleting unhealthy node")
```
jonathan-innis, line 107: Is this consistent with other forms of disruption? Do we use info or debug?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("node.health").
+		For(&corev1.Node{}).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "node.health")
```
jonathan-innis, line 65: Should we attach the nodeclaim name to this node since we are looking up the information and we have it?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,126 @@
...
+		metrics.ReasonLabel:       metrics.UnhealthyReason,
+		metrics.NodePoolLabel:     nodeClaims[0].Labels[v1.NodePoolLabelKey],
+		metrics.CapacityTypeLabel: nodeClaims[0].Labels[v1.CapacityTypeLabelKey],
+	}).Inc()
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	stored := nodeClaim.DeepCopy()
+	terminationTime := c.clock.Now().Format(time.RFC3339)
```
jonathan-innis, line 118: I didn't see a discussion of this in the RFC -- let's talk about this more -- we're choosing to just blast past PDBs -- I'm worried about how dangerous this might be -- also, does the node termination controller respect a previously set terminationGracePeriod annotation set on the Node?

pkg/controllers/node/health/controller.go
engedaam: With other forms of disruption we log using info.

pkg/controllers/node/health/controller.go
engedaam: Ahh I found `NodeClaimForNode` that will validate the duplicate check. Switch to using that now. Might be better if use that else where

pkg/controllers/node/health/controller.go
engedaam: Found that saw didn't do that for other delete calls, and just re-queue, but I can add that

engedaam: We only get the nodeclaim to add the `NodeClaimTerminationTimestampAnnotationKey`, not for the delete. I wanted to keep the focus dealing with the node. I can also remove getting the nodeclaim  from flow. Do you have a strong preference that we delete the nodeclaim?

pkg/controllers/node/health/controller.go
engedaam: Good call, added a check for the karpenter finalizer

pkg/controllers/node/health/controller.go
engedaam: In the case, we find the node is healthy we want to make sure we exit the control loop

pkg/controllers/node/health/controller.go
engedaam: There was no reason. I can sync up the order

pkg/controllers/controllers.go
engedaam: I was trying to follow how we approached the drift feature flag. In there, we would log each time we removed a status condition, and I was thinking it made sense to log the feature setting once on startup. I removed it now

pkg/controllers/node/health/controller.go
engedaam: The node termination controller will respect a previously set `terminationGracePeriod`https://github.com/kubernetes-sigs/karpenter/blob/fea57a1de71d9d9767439a793ee2428d6cebc9e6/pkg/controllers/nodeclaim/lifecycle/controller.go#L243. I mentioned in the section on forceful termination on the need to not wait on pod termination https://github.com/kubernetes-sigs/karpenter/pull/1768/files#diff-997c287391c6ff3ac98a6fada129945da32e2c6de6d9c225352315b207527aeaR119

pkg/cloudprovider/types.go
```diff
@@ -63,6 +74,9 @@ type CloudProvider interface {
 	// IsDrifted returns whether a NodeClaim has drifted from the provisioning requirements
 	// it is tied to.
 	IsDrifted(context.Context, *v1.NodeClaim) (DriftReason, error)
+	// RepairPolicy is for CloudProviders to define a set Unhealthy condition for Karpenter
+	// to monitor on the node.
+	RepairPolicy() []RepairStatement
```
njtran, line 79: Any reason you can't do this? 
```suggestion
	RepairPolicies() []RepairPolicy
```

pkg/cloudprovider/types.go
```diff
@@ -41,6 +42,16 @@ var (
 
 type DriftReason string
 
+type RepairStatement struct {
+	// Type of unhealthy state that is found on the node
+	Type corev1.NodeConditionType
```
njtran, line 47: Might be nice to call it ConditionType so it hints to the mechanism it relies on?
```suggestion
	ConditionType corev1.NodeConditionType
```

pkg/cloudprovider/types.go
```diff
@@ -41,6 +42,16 @@ var (
...
+	// Status condition when a node is unhealthy
+	Status corev1.ConditionStatus
```
njtran, line 49: consider this too
```suggestion
	ConditionStatus corev1.ConditionStatus
```

pkg/cloudprovider/types.go
```diff
@@ -41,6 +42,16 @@ var (
...
+	// TolerationDuration is the duration the controller will wait
+	// before force terminating nodes that are unhealthy.
+	TolerationDuration time.Duration
```
njtran, line 52: I wonder if we should consider doing tolerationSeconds, since it's a bit easier to serialize and follows the pod tolerations semantic, but I wonder how long these tolerations are realistically going to be. Thoughts on one or the other? 

I also question if toleration is really the best word we can do here. It's nice because it implies that it works the same way as taints and tolerations. Could we name it `RepairAfter`? It would match with our other `__After` semantics.

pkg/controllers/controllers.go
```diff
@@ -88,4 +93,11 @@ func NewControllers(
 		status.NewController[*v1.NodeClaim](kubeClient, mgr.GetEventRecorderFor("karpenter")),
 		status.NewController[*v1.NodePool](kubeClient, mgr.GetEventRecorderFor("karpenter")),
 	}
+
+	// The cloud provider must define status conation for the node repair controller to used for detecting unhealthy nodes
```
njtran, line 97:
```suggestion
	// The cloud provider must define status conditions for the node repair controller to use to detect unhealthy nodes
```

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+	// Validate that the node is owned by us and is not being deleted
+	if !node.GetDeletionTimestamp().IsZero() || !controllerutil.ContainsFinalizer(node, v1.TerminationFinalizer) {
+		return reconcile.Result{}, nil
+	}
+
+	for _, policy := range c.cloudProvider.RepairPolicy() {
+		nodeHealthCondition = nodeutils.GetCondition(node, policy.Type)
+		if nodeHealthCondition.Status == policy.Status {
+			// found unhealthy condition on the node
+			foundCloudProviderPolicy = policy
```
njtran, line 84: couldn't we just check here if the repairAfter and lastTransition time means it should be deleted and exit here?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// 2. If the Node is unhealthy, but has not reached it's full toleration disruption, exit the loop
+	disruptionTime := nodeHealthCondition.LastTransitionTime.Add(foundCloudProviderPolicy.TolerationDuration)
+	if c.clock.Now().Before(disruptionTime) {
+		return reconcile.Result{RequeueAfter: disruptionTime.Sub(c.clock.Now())}, nil
+	}
+
+	if err := c.annotateTerminationGracePeriod(ctx, nodeClaim); err != nil {
```
njtran, line 101: Do we need to annotate it? Can't we just let the termination controller do that?

pkg/controllers/node/health/controller.go
njtran: Probably worth info level IMO

pkg/controllers/node/health/controller.go
njtran: I missed this too, but feels like it should just follow the same method as any other termination and not reinvent a special case.

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting nodeclaim, %w", err)
+	}
+
+	ctx = injection.WithControllerName(ctx, "node.health")
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("nodeclaim", nodeClaim.Name))
```
jonathan-innis, line 71: This doesn't match our log formatting that we have been using when injecting objects -- see: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/provisioning/scheduling/scheduler.go#L112

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("node.health").
+		For(&corev1.Node{}).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("getting nodeclaim, %w", err)
```
jonathan-innis, line 67: Should we really requeue if we don't get a nodeclaim for the node or if we get multiple nodeclaims? If this is how we are checking ownership, then we definitely shouldn't be erroring here for nodes that we don't own

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	ctx = injection.WithControllerName(ctx, "node.health")
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("nodeclaim", nodeClaim.Name))
+	nodeHealthCondition := corev1.NodeCondition{}
+	foundCloudProviderPolicy := cloudprovider.RepairStatement{}
+
+	// Validate that the node is owned by us and is not being deleted
+	if !node.GetDeletionTimestamp().IsZero() || !controllerutil.ContainsFinalizer(node, v1.TerminationFinalizer) {
```
jonathan-innis, line 76: nit: I think there's some weird ordering in this controller -- I think we should use the patterns in other controllers as a point of reference. In general, we have the following structure to controllers:

1. Inject the controller name at the top
2. Do any filtering that we can do before we start pulling things from the apiserver
3. Do any pulling from the apiserver
4. Filter again if there's anything to filter out at this point

In general: I'd shift the deletion timestamp check up and I'd use the fact that the NodeClaim exists as a check for ownership rather than relying on things like the finalizer

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	for _, policy := range c.cloudProvider.RepairPolicy() {
```
jonathan-innis, line 80: nit: `lo.Find()`?

pkg/controllers/node/health/controller.go
jonathan-innis: But you found a condition that matched your Node Repair policy so I'm lost why you need this additional check

pkg/controllers/node/health/controller.go
jonathan-innis, line 76: I'm curious if we actually _want_ to deal with deleting nodes -- if a node is deleting but it unhealthy for an extended period of time, do we jump-in, intervene and add a forceful termination to it? Do we know if Node agents that add conditions add these to deleting nodes?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,129 @@
...
+	return reconcile.Result{}, nil
+}
+
+func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	stored := nodeClaim.DeepCopy()
+	terminationTime := c.clock.Now().Format(time.RFC3339)
+	nodeClaim.ObjectMeta.Annotations = lo.Assign(nodeClaim.ObjectMeta.Annotations, map[string]string{v1.NodeClaimTerminationTimestampAnnotationKey: terminationTime})
+
+	if err := c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
+		return client.IgnoreNotFound(err)
```
jonathan-innis, line 125: If we failed to patch due to NotFound, do we really want to ignore the error at this layer? I think we should just bail out of the provisioning loop in general when we get a NodeClaim notFound since there's nothing we can really do at that point

pkg/controllers/node/health/suite_test.go
```diff
@@ -0,0 +1,295 @@
...
+			ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
+			// Determine to delete unhealthy nodes
+			ExpectObjectReconciled(ctx, env.Client, healthController, node)
+			// Let the termination controller terminate the node
+			ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			ExpectExists(ctx, env.Client, node)
+		})
+		It("should set termination grace period to now when not defined ", func() {
```
jonathan-innis, line 190:
```suggestion
		It("should set termination grace period to now when not defined", func() {
```

nit

jonathan-innis, line 190: Don't you want to set it to "now" always? Not just when it's not defined?

pkg/controllers/node/health/suite_test.go
```diff
@@ -0,0 +1,295 @@
...
+			// Determine to delete unhealthy nodes
+			ExpectObjectReconciled(ctx, env.Client, healthController, node)
+			// Let the termination controller terminate the node
+			ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+			Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1.NodeClaimTerminationTimestampAnnotationKey, fakeClock.Now().Format(time.RFC3339)))
+		})
+		It("should respect termination grace period if set on the nodepool", func() {
```
jonathan-innis, line 208: Should we really respect it if it's set on the NodePool? I thought we mentioned that these were special circumstances and so we would react faster? Also, I'm not sure this matches with what you have in the code with you overriding the annotation, right?

pkg/controllers/node/health/suite_test.go
```diff
@@ -0,0 +1,295 @@
...
+			ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+			Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1.NodeClaimTerminationTimestampAnnotationKey, fakeClock.Now().Format(time.RFC3339)))
+		})
+	})
+
+	Context("Forceful termination", func() {
+		It("should not respect node disruption budgets ", func() {
```
jonathan-innis, line 230:
```suggestion
		It("should ignore node disruption budgets", func() {
```

pkg/controllers/node/health/suite_test.go
```diff
@@ -0,0 +1,295 @@
...
+			ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
+			// Determine to delete unhealthy nodes
+			ExpectObjectReconciled(ctx, env.Client, healthController, node)
+			// Let the termination controller terminate the node
+			ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			ExpectObjectReconciled(ctx, env.Client, terminationController, node)
+			ExpectNotFound(ctx, env.Client, node)
+		})
+		It("should not respect do-not-disrupt on node", func() {
```
jonathan-innis, line 254:
```suggestion
		It("should ignore do-not-disrupt on a node", func() {
```

pkg/controllers/node/health/controller.go
engedaam: I readjusted the flow of the controller to close to what you call out here.

pkg/controllers/node/health/controller.go
engedaam: I was using the exists of the finalizer on the node as a way to determine if we own the node

pkg/controllers/node/health/controller.go
engedaam: I rewrote this whole section, let me know what you think

pkg/controllers/node/health/controller.go
engedaam: I rewrote this whole section, let me know what you think

pkg/cloudprovider/types.go
engedaam: > I wonder if we should consider doing tolerationSeconds, since it's a bit easier to serialize and follows the pod tolerations semantic, but I wonder how long these tolerations are realistically going to be. Thoughts on one or the other?

These values are expected by in ball park of 30 min. So it would feel awkward to name the field `tolerationSeconds`. This was something we went back and forth on in the RFC. I'm sure it really matter much since this is not a customer exposed field, since I'm assuming you don't mean `consolidateAfter`. If other feel a strong way I can change it.

pkg/controllers/node/health/controller.go
engedaam: So I removed the check on deleting nodes. I think it would be safe to expect pods that monitor nodes to system critical and be removed evicted last.

pkg/cloudprovider/types.go
njtran: yeah like consolidateAfter or expireAfter. Personally I think it follows Karpenter convention to do repairAfter

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,121 @@
...
+func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+	return controllerruntime.NewControllerManagedBy(m).
+		Named("node.health").
+		For(&corev1.Node{}).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "node.health")
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("node", klog.KRef(node.Namespace, node.Name)))
```
jonathan-innis, line 67:
```suggestion
	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(node.Namespace, node.Name)))
```

use capatalization as-is used in the other places in Karpenter

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,121 @@
...
+
+	// Validate that the node is owned by us and is not being deleted
+	if !controllerutil.ContainsFinalizer(node, v1.TerminationFinalizer) {
```
jonathan-innis, line 70: Is checking that we have a NodeClaim enough?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,121 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	// If find if a node is unhealthy
+	healthCondition, repairNode := lo.Find(c.cloudProvider.RepairPolicies(), func(policy cloudprovider.RepairPolicy) bool {
+		unhealthyNodeCondition := nodeutils.GetCondition(node, policy.ConditionType)
```
jonathan-innis, line 76:
```suggestion
		nodeCondition := nodeutils.GetCondition(node, policy.ConditionType)
```

nit

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,121 @@
...
+		if unhealthyNodeCondition.Status != policy.ConditionStatus {
+			return false
+		}
+		// If the Node is unhealthy, but has not reached it's full toleration disruption
+		terminationTime := unhealthyNodeCondition.LastTransitionTime.Add(policy.TolerationDuration)
+		return c.clock.Now().After(terminationTime)
+	})
+	if !repairNode {
+		return reconcile.Result{}, nil
```
jonathan-innis, line 85: This won't cause us to requeue at the termination time, we need to do this right?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, node *corev1.Node) error {
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+	if err != nil {
+		return fmt.Errorf("getting nodeclaim, %w", err)
+	}
+
+	stored := nodeClaim.DeepCopy()
+	terminationTime := c.clock.Now().Format(time.RFC3339)
```
jonathan-innis, line 113: inline this with the next line?

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,121 @@
...
+		terminationTime := unhealthyNodeCondition.LastTransitionTime.Add(policy.TolerationDuration)
+		return c.clock.Now().After(terminationTime)
+	})
+	if !repairNode {
+		return reconcile.Result{}, nil
+	}
+
+	// For unhealthy past the tolerationDisruption window we can forcefully terminate the node
+	if err := c.annotateTerminationGracePeriod(ctx, node); err != nil {
+		return reconcile.Result{}, fmt.Errorf("annotated termination grace period on nodeclaim, %w", err)
```
jonathan-innis, line 90: client.IgnoreNotFound?

engedaam: I'm little confused here. Did I miss understand this comment? https://github.com/kubernetes-sigs/karpenter/pull/1793#discussion_r1843348779

engedaam: I get what you mean the previous comment. Discussed ofline

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,125 @@
...
+		Named("node.health").
+		For(&corev1.Node{}).
+		Complete(reconcile.AsReconciler(m.GetClient(), c))
+}
+
+func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "node.health")
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(node.Namespace, node.Name)))
+
+	// Validate that the node is owned by us and is not being deleted
```
jonathan-innis, line 67:
```suggestion
	// Validate that the node is owned by us
```

nit: we aren't doing the deletion check anymore

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
+	}
+
+	requeueTime := time.Time{}
+	foundHealthCondition := false
+	cpPolicyFound := cloudprovider.RepairPolicy{}
+	// Find a node with a condition that matches one of the unhealthy conditions defined by the cloud provider
+	// If there are multiple unhealthy status condition we will requeue based on the condition closest to its terminationDuration
+	for _, policy := range c.cloudProvider.RepairPolicies() {
```
jonathan-innis, line 78: nit: This feels like it should be a separate function given it feels like it has "arguments" and it's pretty self-contained

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,125 @@
...
+	if !foundHealthCondition {
+		return reconcile.Result{RequeueAfter: requeueTime.Sub(c.clock.Now())}, nil
+	}
+
+	// For unhealthy past the tolerationDisruption window we can forcefully terminate the node
+	if err := c.annotateTerminationGracePeriod(ctx, nodeClaim); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	if err := c.kubeClient.Delete(ctx, nodeClaim); err != nil {
+		return reconcile.Result{}, err
```
jonathan-innis, line 103: This delete could have a NotFound error -- consider adding a `client.IgnoreNotFound`

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -265,7 +265,10 @@ func (c *Controller) annotateTerminationGracePeriodTerminationTime(ctx context.C
 	stored := nodeClaim.DeepCopy()
 	nodeClaim.ObjectMeta.Annotations = lo.Assign(nodeClaim.ObjectMeta.Annotations, map[string]string{v1.NodeClaimTerminationTimestampAnnotationKey: terminationTime})
 
-	if err := c.kubeClient.Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
+	// We use client.MergeFromWithOptimisticLock because patching a terminationGracePeriod annotation
+	// can cause races with the health controller, as that controller sets the current time as the terminationGracePeriod annotation
+	// Here, We want to resolve any conflict and not overwrite the terminationGracePeriod annotation
+	if err := c.kubeClient.Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
 		return client.IgnoreNotFound(err)
```
jonathan-innis, line 272: This doesn't handle the conflict gracefully

pkg/controllers/node/health/controller.go
```diff
@@ -0,0 +1,127 @@
...
+	ctx = injection.WithControllerName(ctx, "node.health")
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(node.Namespace, node.Name)))
+
+	// Validate that the node is owned by us
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+	if err != nil {
+		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
+	}
+
+	requeueTime, foundHealthCondition, cpPolicyFound := c.findUnhealthyNodes(node)
```
jonathan-innis, line 73: nit: I find this naming confusing since it's not talking about nodes
```suggestion
	requeueTime, foundHealthCondition, cpPolicyFound := c.findUnhealthy Condition(node)
```

jonathan-innis, line 73: Idea: Can this function return the condition that is nearest to hitting its toleration duration and then check the current time based off of that -- I think that would make this code a bit easier to reason about -- as it stands, foundHealthCondition doesn't actually tell me if a condition is found, it tells me if there is a condition that is past its toleration duration, which I don't think is intuitive to a reader

pkg/controllers/nodeclaim/lifecycle/controller.go
engedaam: Not sure what you mean here. I added a conflict check in the reconciliation result to not provide an error but requeue.

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

PR #1794, opened 2024-10-31
fix: dedupe expiration reconciliations

jmdeal:
Fixes #N/A 

**Description**
Currently, the expiration controller reconciles against all NodeClaim updates, and doesn't have a check to ensure that the NodeClaim hasn't already been terminated. This can result in duplicate expiration attempts against the same NodeClaim. The main problem with this (other than unnecessary deletes) is that we increment the expiration metric each time.

This change ensures we don't double count expirations through two mechanisms:
a. Check that the NodeClaim's deletion timestamp is zero before proceeding with expiration.
b. Only enqueue requests for NodeClaims when they are first added to the informer cache.

`make test`

### DIFF
--- a/pkg/controllers/nodeclaim/expiration/controller.go
+++ b/pkg/controllers/nodeclaim/expiration/controller.go
@@ -22,9 +22,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/utils/clock"
 	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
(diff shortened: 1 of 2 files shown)


pkg/controllers/nodeclaim/expiration/controller.go
```diff
@@ -72,9 +78,14 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
...
+	return controllerruntime.NewControllerManagedBy(m).
 		Named("nodeclaim.expiration").
-		For(&v1.NodeClaim{}).
+		For(&v1.NodeClaim{}, builder.WithPredicates(predicate.TypedFuncs[client.Object]{
+			CreateFunc: func(_ event.TypedCreateEvent[client.Object]) bool { return true },
+			DeleteFunc: func(_ event.TypedDeleteEvent[client.Object]) bool { return false },
+			UpdateFunc: func(_ event.TypedUpdateEvent[client.Object]) bool { return false },
+			GenericFunc: func(_ event.TypedGenericEvent[client.Object]) bool { return false },
+
+		})).
```
njtran, line 89: I think this is actually harder for us to test and reason about that I'd rather we just do this in the reconcile loop. The computation benefit we get from short circuiting three simple if statements I think is negligible.

pkg/controllers/nodeclaim/expiration/controller.go
```diff
@@ -72,9 +78,13 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
 }
 
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
-	builder := controllerruntime.NewControllerManagedBy(m)
-	return builder.
+	return controllerruntime.NewControllerManagedBy(m).
 		Named("nodeclaim.expiration").
-		For(&v1.NodeClaim{}).
+		For(&v1.NodeClaim{}, builder.WithPredicates(predicate.TypedFuncs[client.Object]{
```
jonathan-innis, line 83: My vote is to keep the same `time.Sleep` mechanism that we have in the NodeClaim lifecycle controller and just apply that to reading our own writes here

njtran (commented):
/lgtm
/approve

PR #1797, opened 2024-11-06
chore: make leader election ID configurable

saurav-agarwalla:
Fixes #N/A 

**Description**

chore: make leader election ID configurable

Unit tests

### DIFF
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -53,6 +53,7 @@ type Options struct {
 	KubeClientBurst         int
 	EnableProfiling         bool
 	DisableLeaderElection   bool
+	LeaderElectionID        string
 	LeaderElectionNamespace string
 	MemoryLimit             int64
 	LogLevel                string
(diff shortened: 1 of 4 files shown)


pkg/operator/options/options.go
```diff
@@ -88,6 +89,7 @@ func (o *Options) AddFlags(fs *FlagSet) {
 	fs.IntVar(&o.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
 	fs.BoolVarWithEnv(&o.EnableProfiling, "enable-profiling", "ENABLE_PROFILING", false, "Enable the profiling on the metric endpoint")
 	fs.BoolVarWithEnv(&o.DisableLeaderElection, "disable-leader-election", "DISABLE_LEADER_ELECTION", false, "Disable the leader election client before executing the main loop. Disable when running replicated components for high availability is not desired.")
+	fs.StringVar(&o.LeaderElectionID, "leader-election-id", env.WithDefaultString("LEADER_ELECTION_ID", "karpenter-leader-election"), "Leader election ID to create and monitor the lease if running outside the cluster")
```
jonathan-innis, line 92:
```suggestion
	fs.StringVar(&o.LeaderElectionID, "leader-election-name", env.WithDefaultString("LAEADER_ELECTION_NAME", "karpenter-leader-election"), "Leader election ID to create and monitor the lease if running outside the cluster")
```

consider calling it `LEADER_ELECTION_NAME`

saurav-agarwalla: I chose this to be consistent with the controller runtime terminology: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/operator/operator.go#L133

Moving away from it could be confusing in my opinion.

njtran: I had the same thought, but think it's better to be consistent with controller runtime 👍

saurav-agarwalla: Yeah we could deviate if we really wanted to but users are likely going to be more familiar with what's in the controller runtime. It does seem like an odd choice in the controller runtime for sure. 

If it helps, I also looked at other popular controllers and they have followed the same convention as the controller runtime so it would be a good idea for us to do the same such that we aren't deviating from the norm.

jonathan-innis: The counter-argument is that I looked at some of the upstream CLI arguments and it seems like they use `name`. I'd say that the fact that use controller-runtime is an implementation detail and would tend to opt for the more standard name/namespace api surface syntax

jonathan-innis (approved):
/lgtm
/approve

PR #1801, opened 2024-11-08
Abstract prometheus metrics into interfaces

jonathan-innis:
Fixes #N/A 

**Description**

This leverages https://github.com/awslabs/operatorpkg/pull/92 to create abstract metrics so that you can create an arbitrary pipeline for routing metrics beyond prometheus

For instance,
```
status.TerminationDuration = pmetrics.NewMultiObservation(
	status.TerminationDuration,
	pmetrics.NewEMFObservation(writer, "operatorpkg", "operator_termination_duration_seconds", []string{"group", "kind", "namespace"}),
)
```

`make presubmit`

### DIFF
--- a/pkg/cloudprovider/metrics/cloudprovider.go
+++ b/pkg/cloudprovider/metrics/cloudprovider.go
@@ -19,6 +19,7 @@ package metrics
 import (
 	"context"
 
+	opmetrics "github.com/awslabs/operatorpkg/metrics"
 	"github.com/prometheus/client_golang/prometheus"
 	crmetrics "sigs.k8s.io/controller-runtime/pkg/metrics"
 
@@ -44,7 +45,8 @@ const (
 // decorator implements CloudProvider
 var _ cloudprovider.CloudProvider = (*decorator)(nil)
 
-var methodDuration = prometheus.NewHistogramVec(
+var MethodDuration = opmetrics.NewPrometheusHistogram(
+	crmetrics.Registry,
 	prometheus.HistogramOpts{
 		Namespace: metrics.Namespace,
 		Subsystem: "cloudprovider",
@@ -59,7 +61,8 @@ var methodDuration = prometheus.NewHistogramVec(
 )
 
 var (
-	errorsTotal = prometheus.NewCounterVec(
+	ErrorsTotal = opmetrics.NewPrometheusCounter(
+		crmetrics.Registry,
 		prometheus.CounterOpts{
 			Namespace: metrics.Namespace,
 			Subsystem: "cloudprovider",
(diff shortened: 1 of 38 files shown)


kwok/apis/crds/karpenter.kwok.sh_kwoknodeclasses.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.5
+    controller-gen.kubebuilder.io/version: v0.16.3
```
njtran, line 6: why downgrade?

pkg/controllers/metrics/node/controller.go
njtran, line 1: it's trivial but the metrics are public where they don't need to be. Thoughts on just reverting to private?

pkg/operator/operator.go
```diff
@@ -82,10 +83,14 @@ var (
...
 func init() {
-	crmetrics.Registry.MustRegister(BuildInfo)
 	opmetrics.RegisterClientMetrics(crmetrics.Registry)
 
-	BuildInfo.WithLabelValues(Version, runtime.Version(), runtime.GOARCH, env.GetRevision()).Set(1)
+	BuildInfo.Set(1, map[string]string{
+		"version":   Version,
+		"goversion": runtime.Version(),
+		"goarch":    runtime.GOARCH,
+		"commit":    env.GetRevision(),
```
njtran, line 92: where's the .Set(1)?

pkg/controllers/metrics/node/controller.go
jonathan-innis: The goal of this is that they _can_ be overridden by anyone from the outside

pkg/operator/operator.go
jonathan-innis: It's on L88

pkg/cloudprovider/metrics/cloudprovider.go
```diff
@@ -59,7 +61,8 @@ var methodDuration = prometheus.NewHistogramVec(
 )
 
 var (
-	errorsTotal = prometheus.NewCounterVec(
+	ErrorsTotal = opmetrics.NewPrometheusCounter(
```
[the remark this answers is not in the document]
jonathan-innis: Same reason as the others -- allows folks to override and define their own custom metric pipelines

pkg/operator/operator.go
jonathan-innis: It's on L88

njtran: oopie

njtran (commented):
/lgtm

PR #1803, opened 2024-11-08
feat: add pod acknowledged time metric

njtran:
Fixes #N/A 

**Description**
This emits a metric when we consider a pod for scheduling. This will give an accurate representation of when Karpenter actually first considers the pod for scheduling. Currently this would emit for every pod for every scheduling loop, so if a pod fails to schedule, it would be considered again.

havent yet


pkg/controllers/state/cluster.go
```diff
@@ -53,6 +53,11 @@ type Cluster struct {
 	nodeClaimNameToProviderID map[string]string               // node claim name -> provider id
 	daemonSetPods             sync.Map                        // daemonSet -> existing pod
 
+	// podAckTimes keeps track of when Karpenter first tries to schedule the pod.
+	// This is orchestrated internally as opposed to writing to state in the cluster to reduce overall
+	// writes to the APIServer, but also since it only needs to be orchestrated cross-controller internally.
+	podAckTimes map[types.NamespacedName]time.Time // pod namespaced name -> when it was first considered for scheduling
```
jonathan-innis, line 59: How much does this increase our memory utilization when there is a large scale-up? Users perform 5000pod scale-ups -- when these size of scale-ups happen is storing this data expensive? I doubt it, but we should check

pkg/controllers/state/cluster.go
```diff
@@ -307,12 +313,40 @@ func (c *Cluster) UpdatePod(ctx context.Context, pod *corev1.Pod) error {
...
 
+// SetPodAckTime marks the pod as validating for scheduling. This is used to construct metrics
+// when the pod actually schedules to when Karpenter actually thought it could schedule.
+// This allows users to understand how long it took for Karpenter to actually start scheduling
+// for this pod.
+func (c *Cluster) SetPodAckTime(pods ...*corev1.Pod) {
+	for _, pod := range pods {
+		// Check the pod to see if we've already emitted a metric for it, and just continue
+		nn := types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}
+		c.mu.RLock()
```
jonathan-innis, line 324: This mutex is used for state nodes -- do we want to take another mutex for this data since this data is unrelated to the node data -- and I don't think we want the speed of grabbing state nodes to be affected by setting this observability metric

pkg/controllers/state/metrics.go
```diff
@@ -26,6 +26,8 @@ import (
 
 const (
 	stateSubsystem = "cluster_state"
+	PodName        = "name"
```
jonathan-innis, line 29: Do we need these constants?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -100,19 +101,40 @@ var (
...
 			Help:      "The time from pod creation until the pod is running.",
 		},
 		[]string{podName, podNamespace},
 	)
+	PodScheduledDurationSeconds = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.PodSubsystem,
+			Name:      "scheduled_seconds",
```
jonathan-innis, line 116:
```suggestion
			Name:      "unscheduled_time_seconds",
```

Are we tracking the amount of time that it took a pod to schedule as a gauge or are we tracking how long the pod is in a state where it hasn't scheduled? I think the latter is more valuable -- the first one you should be able to get from the histogram metric

pkg/controllers/provisioning/provisioner.go
```diff
@@ -345,6 +345,8 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 		}
 		return scheduler.Results{}, fmt.Errorf("creating scheduler, %w", err)
 	}
+	// Emit a metric for when we first acknowledged the pod for scheduling
+	p.cluster.SetPodAckTime(pendingPods...)
```
jonathan-innis, line 349: This emits when we first see the pod but doesn't signal whether the pod can actually schedule -- aren't the metrics going to be affected by whether we _can_ actually schedule the pod? Seems like we need a metric that marks down how long it took for us to go from pod seen to considered by the scheduler (either able to be scheduled or not) and another metric which tells us how long it took us to go from determined that we can schedule it to actually getting the pod bound.

pkg/controllers/metrics/pod/controller.go
njtran: Ah good call

pkg/controllers/provisioning/provisioner.go
njtran: That's a good point. So what we really need is to track the state transitions between four points and emit metrics from `1 -> 2`, `2->3`, and `2-> 4`.
1. Pod created
2. When did we first see the pod in our scheduling loop
3. When did we first think that we could fit the pod on a node (theoretical or not)
4. When did the pod actually bind

Then we track metrics of 
i. Ack (emitted when ACK'ing the pods on this line in code)
  a. Per-pod ack Gauge (1 -> now)
  b. Histogram ack metric (1 -> 2)
ii. Schedulable (should be from when it was ACK'd, not created, emitted when we actually successfully node.add(pod) in the scheduling loop)
  a. Per-pod schedulable Gauge (2 -> now) 
  b. Histogram schedulable (2 -> 3) 
iii. Scheduled (same as bound, but since it was ACK'd, not created) (emitted when the pod actually binds)
  a. Per-pod scheduled Gauge (2 -> now)
  b. Histogram scheduled (2 -> 4)

jonathan-innis: Yep, fully aligned with this 👍🏼

pkg/controllers/state/cluster.go
njtran: Did a scale up to 3000 and the memory fluctuated to 250 to 350m. Without the changes it's pretty much at the exact same level

pkg/controllers/metrics/pod/controller.go
```diff
@@ -100,19 +101,40 @@ var (
...
+	)
+	PodScheduledDurationSeconds = opmetrics.NewPrometheusHistogram(
+		crmetrics.Registry,
+		prometheus.HistogramOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.PodSubsystem,
+			Name:      "scheduled_duration_seconds",
+			Help:      "The time it takes for pods to be bound from when it's first considered for scheduling in memory",
+		},
+		[]string{},
```
engedaam, line 129: Any thought around adding the pod name and namespace in these metrics?

njtran: This is a histogram, so you want it to just be a function of all the pods, right?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -207,20 +213,20 @@ func (c *Controller) recordPodStartupMetric(pod *corev1.Pod) {
...
 		}
 	}
 }
-func (c *Controller) recordPodBoundMetric(pod *corev1.Pod) {
+func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.Time) {
 	key := client.ObjectKeyFromObject(pod).String()
 	condScheduled, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
 		return c.Type == corev1.PodScheduled
 	})
 	if pod.Status.Phase == phasePending {
```
jonathan-innis, line 226: With the schedulable time tracking, does that mean that we get to drop all of the tracking around seeing the pod/seeing that the pod was in a pending state inside of this controller?

pkg/controllers/metrics/pod/controller.go
```diff
@@ -68,42 +69,42 @@ var (
...
+			Namespace:  metrics.Namespace,
 			Subsystem:  metrics.PodSubsystem,
 			Name:       "startup_duration_seconds",
-			Help:       "The time from pod creation until the pod is running.",
+			Help:       "The time from when Karpenter first thought the pod could schedule until it started running.",
 			Objectives: metrics.SummaryObjectives(),
 		},
 		[]string{},
 	)
 	PodBoundDurationSeconds = opmetrics.NewPrometheusHistogram(
```
jonathan-innis, line 80: Should this be scheduled instead of bound? We haven't released this change yet so we still have a chance to change it -- I still personally prefer scheduled but I'm open to discussing

pkg/controllers/state/cluster.go
```diff
@@ -307,15 +314,54 @@ func (c *Cluster) UpdatePod(ctx context.Context, pod *corev1.Pod) error {
...
+}
+
+// AckPods marks the pod as acknowledged for scheduling from the provisioner. This is only done once per-pod.
+// If the pod hasn't been previously ACK'd, this emits a metric for the pod.
+func (c *Cluster) AckPods(pods ...*corev1.Pod) {
+	now := c.clock.Now()
+	for _, pod := range pods {
+		// store the value as now only if it doesn't exist.
+		if _, alreadySet := c.podAcks.LoadOrStore(types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, true); !alreadySet {
+			timeStart := lo.Ternary(c.initTime.Compare(pod.CreationTimestamp.Time) >= 0, c.initTime, pod.CreationTimestamp.Time)
```
jonathan-innis, line 344: Should we use compare or does it make more semantic sense to use Before/After

jonathan-innis, line 344: I'd question if we should even log a metric based off of the start time of our pods -- I think that's going to drastically skew our metrics on restart

pkg/controllers/state/metrics.go
```diff
@@ -60,4 +59,14 @@ var (
 		},
 		[]string{},
 	)
+	PodAcknowledgedDuration = opmetrics.NewPrometheusHistogram(
```
jonathan-innis, line 62: Should we have an unacknowledged duration as well to mirror this and track when we are hanging?

pkg/controllers/state/cluster.go
njtran: My thought was that the metrics could just be a function of the container, but I think actually it's probably better for it to be more consistent across restarts.

pkg/controllers/provisioning/scheduling/metrics.go
```diff
@@ -78,4 +78,14 @@ var (
 			ControllerLabel,
 		},
 	)
+	PodSchedulerDecisionSeconds = opmetrics.NewPrometheusHistogram(
```
jonathan-innis, line 81: I think we need a gauge metric that is paired with this to track the amount of time that we have a pod but the pod doesn't have a decision made against it

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -226,7 +226,10 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 		}
 
 		// Schedule to existing nodes or create a new node
-		if errors[pod] = s.add(ctx, pod); errors[pod] == nil {
+		errors[pod] = s.add(ctx, pod)
+		s.cluster.MarkPodSchedulingDecision(pod, errors[pod] == nil)
```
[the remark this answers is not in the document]
jonathan-innis: Should this be in the Nominate function call for the provisioning loop?

pkg/controllers/provisioning/scheduling/metrics.go
jonathan-innis: Also, looks like you also declared this metric in cluster state -- did you intend to declare the metric there as well?

pkg/controllers/state/metrics.go
```diff
@@ -60,4 +59,14 @@ var (
 		},
 		[]string{},
 	)
+	PodSchedulingDecisionSeconds = opmetrics.NewPrometheusHistogram(
```
jonathan-innis, line 62: This seems like a duplicate of the scheduler metric that you had above

pkg/controllers/provisioning/scheduling/metrics.go
njtran: > I think we need a gauge metric that is paired with this to track the amount of time that we have a pod but the pod doesn't have a decision made against it 

This would have to be a separate thread, yeah?

pkg/controllers/provisioning/scheduling/scheduler.go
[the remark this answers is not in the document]
njtran: Synced offline, going to call it just after the loop ends in the provisioner code

pkg/controllers/metrics/pod/controller.go
njtran: Keeping it as bound just to reduce the number of changes, since we've creeped up a bit in this pr already

pkg/controllers/state/cluster.go
```diff
@@ -307,15 +314,81 @@ func (c *Cluster) UpdatePod(ctx context.Context, pod *corev1.Pod) error {
...
+			if ackTime := c.PodAckTime(nn); !ackTime.IsZero() {
+				PodSchedulingDecisionSeconds.Observe(c.clock.Since(ackTime).Seconds(), nil)
+			}
+		}
+	}
+}
+
+// PodSchedulingDecision returns when Karpenter first decided if a pod could schedule a pod in scheduling simulations.
+// This returns 0, false if Karpenter never made a decision on the pod.
+func (c *Cluster) PodSchedulingDecision(podKey types.NamespacedName) time.Time {
```
jonathan-innis, line 358:
```suggestion
func (c *Cluster) PodSchedulingDecisionTime(podKey types.NamespacedName) time.Time {
```

nit: since this is returning a time -- I thought it was a boolean when I first read the function name

pkg/controllers/state/cluster.go
```diff
@@ -307,15 +314,81 @@ func (c *Cluster) UpdatePod(ctx context.Context, pod *corev1.Pod) error {
...
+	val, found := c.podsSchedulingAttempted.Load(podKey)
+	if !found {
```
jonathan-innis, line 360: This has the opposite logic of PodAckTime -- any reason?

pkg/controllers/state/cluster.go
```diff
@@ -307,15 +314,81 @@ func (c *Cluster) UpdatePod(ctx context.Context, pod *corev1.Pod) error {
...
+		return time.Time{}
+	}
+	return val.(time.Time)
+}
+
+// PodSchedulingSuccess returns when Karpenter first thought it could schedule a pod in its scheduling simulation.
+// This returns 0, false if the pod was never considered in scheduling as a pending pod.
+func (c *Cluster) PodSchedulingSuccess(podKey types.NamespacedName) time.Time {
```
jonathan-innis, line 368:
```suggestion
func (c *Cluster) PodSchedulingSuccessTime(podKey types.NamespacedName) time.Time {
```

nit: same confusion on naming

pkg/controllers/state/cluster.go
```diff
@@ -307,15 +314,81 @@ func (c *Cluster) UpdatePod(ctx context.Context, pod *corev1.Pod) error {
...
+	val, found := c.podsSchedulableTimes.Load(podKey)
+	if !found {
```
jonathan-innis, line 370: Same statement on the order/polarity of condition checking

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -56,6 +56,7 @@ var _ = Describe("Drift", func() {
 		})
 		// NodeClaims are required to be launched before they can be evaluated for drift
 		nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeLaunched)
+		Expect(nodeClaim.StatusConditions().Clear(v1.ConditionTypeDrifted)).To(Succeed())
```
jonathan-innis, line 59: Was this a flake? Seems unrelated to your change

njtran: Yeah

jonathan-innis (approved):
/lgtm
/approve

PR #1805, opened 2024-11-09
chore: Make nodeclaimlifecycle.Controller.Name() method private in lifecycle controller code

YanyChoi:
Fixes #N/A 

**Description**
The original public method `c.Name()` caused IntelliJ to spit out errors for not matching the `ObjectReconciler` interface.
This method won't be used outside of struct `nodeclaimlifecycle.Controller`, so I privatized this method to resolve the object infer errors.

image

ran `make presubmit`

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/controller.go
+++ b/pkg/controllers/nodeclaim/lifecycle/controller.go
@@ -85,7 +85,7 @@ func NewController(clk clock.Clock, kubeClient client.Client, cloudProvider clou
 
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
-		Named(c.Name()).
+		Named(c.name()).
 		For(&v1.NodeClaim{}).
 		Watches(
 			&corev1.Node{},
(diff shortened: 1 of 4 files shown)


pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -85,7 +85,7 @@ func NewController(clk clock.Clock, kubeClient client.Client, cloudProvider clou
 
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
-		Named(c.Name()).
+		Named(c.name()).
```
jonathan-innis, line 88: I'm not sure that I quite understand why this needs to be done -- if a struct has more methods that aren't defined on an interface, that doesn't mean that it doesn't fulfill the interface. This seems like a change that Intellij should take up vs. one that we should make in _our_ code

jonathan-innis (changes requested):

pkg/controllers/nodeclaim/lifecycle/controller.go
YanyChoi: I just found out that this type of error occurs inconsistently in rare cases, and yes, I understood that this doesn't really go against the go interface rules. I'll let Jetbrains know about this. Thank you!

PR #1811, opened 2024-11-14
chore: create new error type to wrap message and error

jigisha620:
Fixes #N/A 

**Description**
Create a new error type that can wrap error message and actual error.

NA

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -373,3 +373,16 @@ func IsNodeClassNotReadyError(err error) bool {
 	var nrError *NodeClassNotReadyError
 	return errors.As(err, &nrError)
 }
+
+// CreateError is an error type returned by CloudProviders when instance creation fails
+type CreateError struct {
+	error
+	ConditionMessage string
+}
+
+func NewCreateError(err error, message string) *CreateError {
+	return &CreateError{
+		error:            err,
+		ConditionMessage: message,
+	}
+}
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/types.go
```diff
@@ -373,3 +373,24 @@ func IsNodeClassNotReadyError(err error) bool {
 	var nrError *NodeClassNotReadyError
 	return errors.As(err, &nrError)
 }
+
+// InstanceCreateError is an error type returned by CloudProviders when instance creation fails
+type InstanceCreateError struct {
```
jonathan-innis, line 378: I wonder if we should just call it `Error` or `ConditionError` or `PrettyError`. Realistically, I don't think we have to be specific and call this a create error.

pkg/cloudprovider/types.go
```diff
@@ -373,3 +373,24 @@ func IsNodeClassNotReadyError(err error) bool {
...
+	error
+	message string
+}
+
+func NewInstanceCreateError(err error, message string) *InstanceCreateError {
```
jonathan-innis, line 383: We should handle parsing this out and setting the condition in the lifecycle controller in this same code change -- otherwise, there's not a ton of value in just adding the error

pkg/cloudprovider/types.go
jigisha620: Addressed

jonathan-innis (approved):
/lgtm
/approve

PR #1816, opened 2024-11-15
chore: fix unfinishedWorkSeconds metric

jigisha620:
Fixes #N/A 

**Description**
Fix unfinishedWorkSeconds metric that is being leaked because a  a unique metric is getting created each time with the id and controller.


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -242,6 +242,7 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 		}
 	}
 	UnfinishedWorkSeconds.Set(0, map[string]string{ControllerLabel: injection.GetControllerName(ctx), schedulingIDLabel: string(s.id)})
```
jonathan-innis, line 244: You don't need this Set call anymore if you are going to call Delete

PR #1818, opened 2024-11-16
feat: only operate on cloudprovider managed resources

jmdeal:
Fixes #N/A 

**Description**
Karpenter should only operate on resources which are managed by the cloudprovider consuming this library. The CloudProvider provides a list of supported NodeClasses, Karpenter will assume ownership of any resource with a ref on one of those NodeClasses. 

`make test`


pkg/controllers/node/termination/controller.go
```diff
@@ -87,7 +89,11 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 	if !controllerutil.ContainsFinalizer(node, v1.TerminationFinalizer) {
 		return reconcile.Result{}, nil
 	}
-	nodeClaims, err := nodeutils.GetNodeClaims(ctx, node, c.kubeClient)
+	if !nodeutils.IsManaged(node, c.cloudProvider) {
+		return reconcile.Result{}, nil
+	}
```
rschalo, line 94: Does this ever happen because of the event filter?

pkg/controllers/node/termination/controller.go
```diff
@@ -283,7 +289,7 @@ func (c *Controller) nodeTerminationTime(node *corev1.Node, nodeClaims ...*v1.No
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("node.termination").
-		For(&corev1.Node{}).
+		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsMangedPredicates(c.cloudProvider))).
```
rschalo, line 292:
```suggestion
		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicates(c.cloudProvider))).
```

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -67,6 +68,9 @@ func NewController(clk clock.Clock, kubeClient client.Client, cloudProvider clou
 func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
 	ctx = injection.WithControllerName(ctx, "nodeclaim.disruption")
 
+	if !nodeclaimutil.IsManaged(nodeClaim, c.cloudProvider) {
```
rschalo, line 71: Same question as nodes.

pkg/controllers/nodeclaim/expiration/controller.go
```diff
@@ -22,30 +22,38 @@ import (
...
-		kubeClient: kubeClient,
-		clock:      clk,
+		clock:         clk,
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
 	}
 }
 
 func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
+	if !nodeclaimutils.IsManaged(nodeClaim, c.cloudProvider) {
```
rschalo, line 54: Same question.

pkg/controllers/nodeclaim/lifecycle/launch_test.go
```diff
@@ -34,24 +35,43 @@ var _ = Describe("Launch", func() {
 	BeforeEach(func() {
 		nodePool = test.NodePool()
 	})
-	It("should launch an instance when a new NodeClaim is created", func() {
-		nodeClaim := test.NodeClaim(v1.NodeClaim{
-			ObjectMeta: metav1.ObjectMeta{
-				Labels: map[string]string{
-					v1.NodePoolLabelKey: nodePool.Name,
+	DescribeTable(
```
rschalo, line 38: Since it looks like these tables just change behavior for the not-managing cloudproviders with a single param, should this be a `Context` that sets up the behavior for the managing cloudprovider and then alters the nodeClaimOpts in the `It` block for managed?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -212,53 +213,48 @@ var ErrNodePoolsNotFound = errors.New("no nodepools found")
...
-	nodePoolList.OrderByWeight()
+	nodepoolutils.OrderByWeight(nodePools)
 
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	domains := map[string]sets.Set[string]{}
-	var notReadyNodePools []string
-	for _, nodePool := range nodePoolList.Items {
-		// Get instance type options
-		instanceTypeOptions, err := p.cloudProvider.GetInstanceTypes(ctx, lo.ToPtr(nodePool))
+	for _, np := range nodePools {
```
rschalo, line 238: It looks like a lot of renaming that we could skip to reduce diff.

pkg/utils/nodepool/nodepool.go
```diff
@@ -18,32 +18,130 @@ package nodepool
...
 )
 
+func IsManaged(nodePool *v1.NodePool, cp cloudprovider.CloudProvider) bool {
+	return lo.ContainsBy(cp.GetSupportedNodeClasses(), func(nodeClass status.Object) bool {
+		return object.GVK(nodeClass).GroupKind() == nodePool.Spec.Template.Spec.NodeClassRef.GroupKind()
+	})
+}
+
+// IsMangedPredicates is used to filter controller-runtime NodeClaim watches to NodeClaims managed by the given cloudprovider.
+func IsMangedPredicates(cp cloudprovider.CloudProvider) predicate.Funcs {
```
rschalo, line 44:
```suggestion
func IsManagedPredicates(cp cloudprovider.CloudProvider) predicate.Funcs {
```

pkg/controllers/node/termination/controller.go
jmdeal: It can, the event filter only applies when enqueuing the reconcile request, it's possible it no longer holds true by the time it's processed.

pkg/controllers/nodeclaim/lifecycle/launch_test.go
jmdeal: The idea was to eliminate false positives from changes in the launch conditions. I think the best example is initialization, where we need to meet a bunch of initial conditions to pass the test. If those conditions change, but we don't update the non-managed test to reflect that, it may continue to pass even though it shouldn't. By having a "base" test and ensuring that the only difference is if the resource is managed or not, we ensure we don't have skew between tests causing false positives.

jmdeal: I do find this approach a little hacky fwiw, but I think it's worth ensuring we don't regress blindly due to test skew.

jmdeal:
Holding while I do testing with the AWS provider

/hold

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -20,71 +20,142 @@ import (
...
+// IsManagedPredicates is used to filter controller-runtime NodeClaim watches to NodeClaims managed by the given cloudprovider.
+func IsManagedPredicates(cp cloudprovider.CloudProvider) predicate.Funcs {
+	return predicate.NewPredicateFuncs(func(o client.Object) bool {
+		return IsManaged(o.(*v1.NodeClaim), cp)
+	})
+}
+
+type ListOptions struct {
+	filterPredicates []func(*v1.NodeClaim) bool
+	listOptions      []client.ListOption
```
jonathan-innis, line 56: I wonder if we should even add the option for Managed/Not Managed here -- especially if you are planning to only list the managed NodePools/NodeClaims the majority of the time

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -20,71 +20,142 @@ import (
...
+}
+
+func withClientListOptions(listOpts ...client.ListOption) option.Function[ListOptions] {
+	return func(opts *ListOptions) {
+		opts.listOptions = append(opts.listOptions, listOpts...)
+	}
+}
+
+func WithProviderIDFilter(providerID string) option.Function[ListOptions] {
+	return withClientListOptions(client.MatchingFields{"status.providerID": providerID})
```
jonathan-innis, line 74: Since we talked about removing this list, I wonder if we even need this list option anymore

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -20,71 +20,142 @@ import (
...
+}
+
+func WithNodePoolFilter(nodePoolName string) option.Function[ListOptions] {
```
jonathan-innis, line 77:
```suggestion
func ForNodePool(nodePoolName string) option.Function[ListOptions] {
```

I wonder if this is more natural naming?

pkg/utils/nodeclaim/nodeclaim.go
```diff
@@ -20,71 +20,142 @@ import (
...
+	return withClientListOptions(client.MatchingLabels(map[string]string{v1.NodePoolLabelKey: nodePoolName}))
+}
+
+func WithNodeClassFilter(nodeClass status.Object) option.Function[ListOptions] {
```
jonathan-innis, line 81:
```suggestion
func ForNodeClass(nodeClass status.Object) option.Function[ListOptions] {
```

I wonder if this is more natural naming

pkg/utils/node/node.go
```diff
@@ -159,3 +165,21 @@ func GetCondition(n *corev1.Node, match corev1.NodeConditionType) corev1.NodeCon
...
+		_, ok := node.Labels[nodeclaimutils.NodeClassLabelKey(&v1.NodeClassReference{
+			Group: gvk.Group,
+			Kind:  gvk.Kind,
+		})]
+		return ok
+	})
+}
+
+// IsManagedPredicates is used to filter controller-runtime NodeClaim watches to NodeClaims managed by the given cloudprovider.
+func IsManagedPredicates(cp cloudprovider.CloudProvider) predicate.Funcs {
```
jonathan-innis, line 181:
```suggestion
func IsManagedPredicateFuncs(cp cloudprovider.CloudProvider) predicate.Funcs {
```

pkg/test/nodeclass.go
```diff
@@ -25,6 +25,16 @@ import (
 	"sigs.k8s.io/karpenter/pkg/test/v1alpha1"
 )
 
+var (
+	// defaultNodeClass is the default NodeClass type used when creating NodeClassRefs for NodePools and NodeClaims
+	defaultNodeClass status.Object = &v1alpha1.TestNodeClass{}
+)
+
+// SetDefaultNodeClassType configures the default NodeClass type used when generating NodeClassRefs for test NodePools and NodeClaims.
+func SetDefaultNodeClassType(nc status.Object) {
```
jonathan-innis, line 34: noice

pkg/test/nodeclaim.go
```diff
@@ -57,7 +65,9 @@ func NodeClaim(overrides ...v1.NodeClaim) *v1.NodeClaim {
 
 func NodeClaimAndNode(overrides ...v1.NodeClaim) (*v1.NodeClaim, *corev1.Node) {
 	nc := NodeClaim(overrides...)
-	return nc, NodeClaimLinkedNode(nc)
+	node := NodeClaimLinkedNode(nc)
+	nc.Status.NodeName = node.Name
```
jonathan-innis, line 69: This changes the logic, right? This now schedules the node when it wasn't scheduled before?

pkg/test/cachesyncingclient.go
```diff
@@ -39,8 +39,8 @@ type CacheSyncingClient struct {
 // If we timeout on polling, the assumption is that the cache updated to a newer version
 // and we missed the current WRITE operation that we just performed
 var pollingOptions = []retry.Option{
-	retry.Attempts(100), // This whole poll should take ~1s
-	retry.Delay(time.Millisecond * 10),
+	retry.Attempts(1000), // This whole poll should take ~1s
```
jonathan-innis, line 42: nice!

pkg/controllers/state/suite_test.go
```diff
@@ -880,16 +886,7 @@ var _ = Describe("Node Resource Level", func() {
...
-			ObjectMeta: metav1.ObjectMeta{
-				Labels: map[string]string{
-					v1.NodePoolLabelKey:            nodePool.Name,
-					corev1.LabelInstanceTypeStable: cloudProvider.InstanceTypes[0].Name,
-				},
-				Finalizers: []string{v1.TerminationFinalizer},
-			},
-			ProviderID: nodeClaim.Status.ProviderID,
-		})
+		node := test.NodeClaimLinkedNode(nodeClaim)
```
jonathan-innis, line 889: I wonder if we need a function to ensure that we mark that the NodeClaim has scheduled so that you don't modify the existing function logic

pkg/controllers/state/informer/nodeclaim.go
```diff
@@ -21,27 +21,32 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
```
jonathan-innis, line 34: nit: We append utils for aliasing in some places and not in other places -- curious if there is a reason

pkg/controllers/state/cluster.go
```diff
@@ -106,8 +110,8 @@ func (c *Cluster) Synced(ctx context.Context) (synced bool) {
 	defer func() {
 		ClusterStateSynced.Set(lo.Ternary[float64](synced, 1, 0), nil)
 	}()
-	nodeClaimList := &v1.NodeClaimList{}
-	if err := c.kubeClient.List(ctx, nodeClaimList); err != nil {
+	nodeClaims, err := nodeclaim.List(ctx, c.kubeClient, nodeclaim.WithManagedFilter(c.cloudProvider))
```
jonathan-innis, line 113: Hmmm -- this is interesting -- so we only care that our NodeClaims are synced -- it's an interesting consideration if we should consider NodeClaims that we don't directly manage or not

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -57,7 +58,10 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 		v1.NodePoolHashAnnotationKey:        nodePool.Hash(),
 		v1.NodePoolHashVersionAnnotationKey: v1.NodePoolHashVersion,
 	})
-	nct.Labels = lo.Assign(nct.Labels, map[string]string{v1.NodePoolLabelKey: nodePool.Name})
+	nct.Labels = lo.Assign(nct.Labels, map[string]string{
+		v1.NodePoolLabelKey: nodePool.Name,
+		nodeclaimutils.NodeClassLabelKey(nodePool.Spec.Template.Spec.NodeClassRef): nodePool.Spec.Template.Spec.NodeClassRef.Name,
```
jonathan-innis, line 63: Should this be a utils function or something on the apis package?

jonathan-innis, line 63: Would be kinda cool if we reconciled this on the NodePool too so that folks could list all NodePools that have a given NodeClass

pkg/controllers/provisioning/provisioner.go
jonathan-innis: +1, I'd agree -- not blocking but it does make it harder to review

pkg/controllers/nodepool/validation/suite_test.go
```diff
@@ -65,6 +68,17 @@ var _ = Describe("Counter", func() {
 		nodePool = test.NodePool()
 		nodePool.StatusConditions().SetUnknown(v1.ConditionTypeValidationSucceeded)
 	})
+	It("should ignore NodePools which aren't managed by this instance of Karpenter", func() {
+		nodePool.Spec.Template.Spec.NodeClassRef = &v1.NodeClassReference{
+			Group: "karpenter.k8s.aws",
```
jonathan-innis, line 73: AWS-specific code in the neutral code? Can we mock a group/kind?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -54,20 +54,29 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
 func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reconcile.Result, error) {
 	ctx = injection.WithControllerName(ctx, "nodepool.readiness")
 	stored := nodePool.DeepCopy()
-	supportedNC := c.cloudProvider.GetSupportedNodeClasses()
 
-	if len(supportedNC) == 0 {
-		logger.Fatal("no supported node classes found for the cloud provider")
+	nodeClass, ok := lo.Find(c.cloudProvider.GetSupportedNodeClasses(), func(nc status.Object) bool {
```
jonathan-innis, line 58: We dropped our error checking on GetSupportedNodeClasses(). How do we dry-run to make sure that CloudProviders have defined this?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -54,20 +54,29 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
+	})
+	if !ok {
+		// Ignore NodePools which aren't using a supported NodeClass.
+		// Note: should be unreachable due to predicate in controller reconciliation
+		return reconcile.Result{}, nil
 	}
-	nodeClass, err := c.getNodeClass(ctx, nodePool, supportedNC)
-	if err != nil {
+	err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass)
+	if err != nil && !errors.IsNotFound(err) {
```
jonathan-innis, line 67:
```suggestion
	if err != nil {
```

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -54,20 +54,29 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
+		return object.GVK(nc).GroupKind() == nodePool.Spec.Template.Spec.NodeClassRef.GroupKind()
+	})
+	if !ok {
+		// Ignore NodePools which aren't using a supported NodeClass.
+		// Note: should be unreachable due to predicate in controller reconciliation
+		return reconcile.Result{}, nil
 	}
-	nodeClass, err := c.getNodeClass(ctx, nodePool, supportedNC)
-	if err != nil {
+	err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass)
```
jonathan-innis, line 66: Inline L66 and L67

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -54,20 +54,29 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
+	if err != nil && !errors.IsNotFound(err) {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
 		return reconcile.Result{}, err
 	}
-	if nodeClass == nil {
+
+	switch {
```
jonathan-innis, line 74: Seems like there's a lot of duplicate checks here. Maybe I'm lost but can we double-check this code

pkg/controllers/nodeclaim/expiration/suite_test.go
```diff
@@ -121,23 +118,40 @@ var _ = Describe("Expiration", func() {
 			})
 		})
 	})
+	DescribeTable(
+		"Expiration",
+		func(isNodeClaimManaged bool) {
+			nodeClaim.Spec.ExpireAfter = v1.MustParseNillableDuration("30s")
+			if !isNodeClaimManaged {
+				nodeClaim.Spec.NodeClassRef = &v1.NodeClassReference{
+					Group: "karpenter.k8s.aws",
```
jonathan-innis, line 127: This is AWS-specific again

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -109,24 +113,23 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
...
-			nodeclaimutil.NodeClassEventHandler(c.kubeClient),
-		)
-	}
-	return builder.
+	b := controllerruntime.NewControllerManagedBy(m).
 		Named("nodeclaim.disruption").
-		For(&v1.NodeClaim{}).
+		For(&v1.NodeClaim{}, builder.WithPredicates(nodeclaimutil.IsManagedPredicates(c.cloudProvider))).
 		WithOptions(controller.Options{MaxConcurrentReconciles: 10}).
+		// Note: We don't use the ManagedFilter (NodeClaim) for NodePool updates because drift should be captured when
```
jonathan-innis, line 120: Do we even want to consider this case? Do we simplify the code if we just reason about managed resources

pkg/controllers/node/termination/suite_test.go
```diff
@@ -126,25 +142,20 @@ var _ = Describe("Termination", func() {
...
-				nodes = append(nodes, node)
+				*node = *ExpectNodeExists(ctx, env.Client, node.Name)
 			}
 
 			// Reconcile twice, once to set the NodeClaim to terminating, another to check the instance termination status (and delete the node).
 			for range 2 {
 				var wg sync.WaitGroup
 				// this is enough to trip the race detector
-				for i := 0; i < numNodes; i++ {
+				for i := range nodes {
```
jonathan-innis, line 158: nit: These are changes that make the diff larger -- fine with them, just be careful with it for such a large PR already

pkg/controllers/metrics/nodepool/controller.go
```diff
@@ -95,6 +100,9 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 		}
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
+	if !nodepoolutils.IsManaged(nodePool, c.cloudProvider) {
```
jonathan-innis, line 103: musing: It would be nice if we could generalize this concept somehow outside of the context of the Reconcile loop -- particularly because we are having to repeat ourselves in a bunch of places

pkg/utils/nodepool/nodepool.go
```diff
@@ -18,32 +18,130 @@ package nodepool
...
 )
 
+func IsManaged(nodePool *v1.NodePool, cp cloudprovider.CloudProvider) bool {
+	return lo.ContainsBy(cp.GetSupportedNodeClasses(), func(nodeClass status.Object) bool {
+		return object.GVK(nodeClass).GroupKind() == nodePool.Spec.Template.Spec.NodeClassRef.GroupKind()
+	})
+}
+
+// IsManagedPredicates is used to filter controller-runtime NodeClaim watches to NodeClaims managed by the given cloudprovider.
+func IsManagedPredicates(cp cloudprovider.CloudProvider) predicate.Funcs {
```
jonathan-innis, line 44:
```suggestion
func IsManagedPredicateFuncs(cp cloudprovider.CloudProvider) predicate.Funcs {
```

nit

pkg/controllers/nodeclaim/lifecycle/launch_test.go
```diff
@@ -34,24 +35,43 @@ var _ = Describe("Launch", func() {
...
-		Expect(err).ToNot(HaveOccurred())
-	})
+			Expect(cloudProvider.CreateCalls).To(HaveLen(lo.Ternary(isNodeClaimManaged, 1, 0)))
+			Expect(cloudProvider.CreatedNodeClaims).To(HaveLen(lo.Ternary(isNodeClaimManaged, 1, 0)))
+			if isNodeClaimManaged {
+				_, err := cloudProvider.Get(ctx, nodeClaim.Status.ProviderID)
+				Expect(err).ToNot(HaveOccurred())
+			}
+		},
+		Entry("should launch an instance when a new NodeClaim is created", true),
```
jonathan-innis, line 72: I go either way on whether it's better to just make these separate It blocks. Personally, I would opt for different It() blocks rather than trying to collapse it into Tables

pkg/utils/nodeclaim/suite_test.go
```diff
@@ -131,4 +136,90 @@ var _ = Describe("NodeClaimUtils", func() {
 			BlockOwnerDeletion: lo.ToPtr(true),
 		}))
 	})
+	Context("List", func() {
```
jonathan-innis, line 139: Do we just want to do away with Listing based on providerID since we said that it's an edge case at this point. Fine with a fast-follow though

pkg/controllers/nodeclaim/lifecycle/launch_test.go
jmdeal: The reason I went with `DescribeTables` is to ensure that we don't have test skew over time. If we add additional conditions to reach a state, it would be easy to miss adding those additional conditions to a negative test. By using the same test block, you ensure that the **only** reason the conditions wouldn't be set is that the node is unmanaged.

pkg/controllers/nodepool/readiness/controller.go
jmdeal: We do want to ignore not found errors here, if the NodeClass is not found we want to mark the NodePool as unready.

pkg/controllers/nodepool/readiness/controller.go
jmdeal: I think there was just the one, the NotFound check on line 68 is a no-op and should be dropped. Were there any others you noticed?

pkg/controllers/metrics/nodepool/controller.go
jmdeal: Discussed offline, will look into a wrapper around the typed reconciler - not sure if it makes sense yet, but could be helpful. Going to decouple though.

pkg/controllers/nodepool/readiness/controller.go
jmdeal: The error is used in the switch below, I think leaving this as is is appropriate.

pkg/controllers/nodepool/readiness/controller.go
jmdeal: Discussed offline, we could potentially do this in the operator but it's not really necessary. We'll punt for now.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
jmdeal: Maybe, though you can do that today with a field selector right? Is there additional value?

jmdeal: Good call, this should probably be in the APIs package

pkg/controllers/state/informer/nodeclaim.go
jmdeal: It should be appended everywhere, I'll have to do another pass.

pkg/test/nodeclaim.go
jmdeal: Not really, we don't rely on this anywhere for testing (in this package). In my mind, `NodeClaimAndNode` should return a pair which have already been linked together, reflecting that in the status is appropriate.

pkg/utils/nodeclaim/suite_test.go
jmdeal: It's still used to get a NodeClaim for a given Node, we're just going to error when there are multiple values going forwards.

pkg/controllers/state/suite_test.go
jmdeal: I'm not sure I follow. I changed this to ensure the Node has the NodeClass label on it for the management check, I don't think it makes sense to require the use of the lifecycle controller in any test that needs to apply that label.

pkg/test/nodeclaim.go
jmdeal: I can drop for now then, this was helping out some tests that assume this link exists but I can just add it there. Longer term I think we should include some test helpers that give us a Node and NodeClaim at each stage of the lifecycle (e.g. registered, initialized, terminating).

pkg/utils/node/node.go
```diff
@@ -159,3 +164,17 @@ func GetCondition(n *corev1.Node, match corev1.NodeConditionType) corev1.NodeCon
 	}
 	return corev1.NodeCondition{}
 }
+
+func IsManaged(node *corev1.Node, cp cloudprovider.CloudProvider) bool {
+	return lo.ContainsBy(cp.GetSupportedNodeClasses(), func(nodeClass status.Object) bool {
```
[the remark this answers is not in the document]
jmdeal: Good point, discussed this some more offline but I'm adding a hydration sub-reconciler to the lifecycle controller to add this metadata to existing nodes.

pkg/controllers/nodeclaim/lifecycle/hydration.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	stored := node.DeepCopy()
+	node.Labels = lo.Assign(node.Labels, map[string]string{
+		v1.NodeClassLabelKey(nodeClaim.Spec.NodeClassRef.GroupKind()): nodeClaim.Spec.NodeClassRef.Name,
+	})
+	if !equality.Semantic.DeepEqual(stored, node) {
+		// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
+		// can cause races due to the fact that it fully replaces the list on a change
+		// Here, we are updating the taint list
+		if err := h.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
```
jonathan-innis, line 63: `client.IgnoreNotFound(err)`

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

PR #1819, opened 2024-11-17
perf: Unregister the topology domain when failing NodeClaim creation

jonathan-innis:
Fixes #N/A 

**Description**

Unregister the topology domain when failing to create a NodeClaim so that we don't hold topology domains around for hostname. Prior to this change, we weren't removing a mock domain that we would create for a topology on NewNodeClaim(). As a result, we would continually expand the number of hostname domains, making hostnmae topology spread and hostname anti-affinity way more inefficient than it should have.

### Before

__Example: Debug Image showing ~53,000 domains after only a few seconds iterating in the scheduling loop__

Screenshot 2024-11-16 at 5 12 51 PM

### After

__Example: Debug Image showing only a single domain after only a few seconds iterating in the scheduling loop__

Screenshot 2024-11-16 at 5 17 01 PM

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/nodeclaim.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaim.go
@@ -39,6 +39,7 @@ type NodeClaim struct {
 	topology        *Topology
 	hostPortUsage   *scheduling.HostPortUsage
 	daemonResources v1.ResourceList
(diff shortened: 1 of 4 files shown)


jmdeal (commented):
/lgtm

PR #1820, opened 2024-11-17
perf: Improve TopologyGroup node domain iteration

jonathan-innis:
Fixes #N/A 

**Description**

This change updates our iteration over domains for TSC, affinity, and anti-affinity. Often, we have cases where we have available domains across the cluster, but those domains aren't available for the node we are currently on. This is specifically true with hostname requirements where there is only one option that the node can carry. 

Rather than looping through all domain options and adding those as possibilities when they were never really possible because the node doesn't support it, this change adds an optimization to only loop through the available node domains if it's shorter than all available domains. 

For hostname, this turns our O(node) operation into an O(1) operation to validate whether the pod can fit on the node. 

### Before

Screenshot 2024-11-16 at 11 05 06 PM

### After

Screenshot 2024-11-17 at 12 38 48 AM

### DIFF
--- a/pkg/controllers/provisioning/scheduling/topologygroup.go
+++ b/pkg/controllers/provisioning/scheduling/topologygroup.go
@@ -268,9 +268,8 @@ func (t *TopologyGroup) nextDomainAffinity(pod *v1.Pod, podDomains *scheduling.R
 		return options
 	}
 
-	// If pod is self-selecting and no pod has been scheduled yet OR there is another pod that has scheduled but it's
-	// incompatible with our podDomains, we can pick a domain at random to bootstrap scheduling
-	// This also requires that there not already be some other domain where this a pod in this topology has scheduled
+	// If pod is self-selecting and no pod has been scheduled yet OR the pods that have scheduled are
+	// incompatible with our podDomains, we can pick a domain at random to bootstrap scheduling.
 	if t.selects(pod) && (len(t.domains) == len(t.emptyDomains) || !t.anyCompatiblePodDomain(podDomains)) {
 		// First try to find a domain that is within the intersection of pod/node domains. In the case of an in-flight node
 		// this causes us to pick the domain that the existing in-flight node is already in if possible instead of picking
--- a/pkg/controllers/provisioning/scheduling/topologygroup.go
+++ b/pkg/controllers/provisioning/scheduling/topologygroup.go
@@ -99,7 +99,7 @@ func (t *TopologyGroup) Get(pod *v1.Pod, podDomains, nodeDomains *scheduling.Req
 	case TopologyTypePodAffinity:
 		return t.nextDomainAffinity(pod, podDomains, nodeDomains)
 	case TopologyTypePodAntiAffinity:
-		return t.nextDomainAntiAffinity(podDomains)
+		return t.nextDomainAntiAffinity(podDomains, nodeDomains)
 	default:
 		panic(fmt.Sprintf("Unrecognized topology group type: %s", t.Type))
 	}
@@ -128,7 +128,6 @@ func (t *TopologyGroup) Register(domains ...string) {
 	}
 }
 
-// Unregister removes the topology group from being aware of the domain
 func (t *TopologyGroup) Unregister(domains ...string) {
 	for _, domain := range domains {
 		delete(t.domains, domain)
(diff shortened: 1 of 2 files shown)


pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -224,17 +243,34 @@ func (t *TopologyGroup) domainMinCount(domains *scheduling.Requirement) int32 {
...
+	if options.Len() != 0 {
+		return options
+	}
 
-	// If pod is self selecting and no pod has been scheduled yet, we can pick a domain at random to bootstrap scheduling
-
-	if options.Len() == 0 && t.selects(pod) {
+	// If pod is self-selecting and no pod has been scheduled yet, we can pick a domain at random to bootstrap scheduling
+	// This also requires that there not already be some other domain where this a pod in this topology has scheduled
+	if t.selects(pod) && len(t.domains) == len(t.emptyDomains) {
```
jmdeal, line 273: Just making sure I understand this change: previously, we may have chosen to "bootstrap" a node for a pod with self-affinity if there were no compatible domains with the pod for the affinity term. However, there may have been a pod scheduled in the topology group, and it's domain was just incompatible with this pod's requirements. In this case, previously Karpenter would have created a new node in a compatible domain, but with this change Karpenter would fail to schedule the pod. Are we sure this change matches `kube-scheduler's` behavior? I thought a pod w/ affinty could schedule to a separate domain if the only domain in it's topology group with a matching pod isn't compatible.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -172,25 +171,45 @@ func (t *TopologyGroup) Hash() uint64 {
...
-			// 'existing matching num' + 'if self-match (1 or 0)' - 'global min matching num' <= 'maxSkew'
-			count := t.domains[domain]
-			if selfSelecting {
-				count++
+
+	// If we are explicitly selecting on specific node domains ("In" requirement) and the number of node domains
+	// is less than our total domains, this is going to be more efficient to iterate through
+	// This is particularly useful when considering the hostname topology key that can have a
+	// lot of t.domains but only a single nodeDomain
+	if nodeDomains.Operator() == v1.NodeSelectorOpIn && nodeDomains.Len() < len(t.domains) {
```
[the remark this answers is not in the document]
jonathan-innis: Ah, you're saying because we calculate available domains at the beginning based on the possibilities on the nodepool that this is moot -- think I agree

pkg/controllers/provisioning/scheduling/topologygroup.go
jmdeal: Discussed offline, there are two actual issues:

We need to iterate over all domains in the topology group, not just node domains. If there is a pod in the topology group in a domain which is compatible with this pod's requirements, we should consider that an option. This is currently missed by only iterating over node domains, since it may be in a domain incompatible with the node. 

Additionally, if there are no options available, we don't want to necessarily short-circuit just because `len(t.domains) == len(t.emptyDomains)`. If there is a pod in the topology group that is **not** compatible with the pod's requirements, it should not be considered.

jonathan-innis: Fixed this by validating if there is a compatible domain that has a pod -- this would prevent us from generating a value for this node and scheduling the pod to the current node.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -224,17 +243,35 @@ func (t *TopologyGroup) domainMinCount(domains *scheduling.Requirement) int32 {
...
+	if options.Len() != 0 {
+		return options
+	}
 
-	// If pod is self selecting and no pod has been scheduled yet, we can pick a domain at random to bootstrap scheduling
-
-	if options.Len() == 0 && t.selects(pod) {
+	// If pod is self-selecting and no pod has been scheduled yet OR there is another pod that has scheduled but it's
+	// incompatible with our podDomains, we can pick a domain at random to bootstrap scheduling
+	// This also requires that there not already be some other domain where this a pod in this topology has scheduled
```
jmdeal, line 273: nit: I think this is a little bit clearer, the last sentence is redundant I think and multiple pods could be scheduled.
```suggestion
	// If pod is self-selecting and no pod has been scheduled yet OR the pods that have scheduled are
	// incompatible with our podDomains, we can pick a domain at random to bootstrap scheduling.
```

jmdeal (commented):
One non-blocking nit, I'll approve with a hold.

PR #1822, opened 2024-11-17
perf: Don't reparse selector on each toplogy select check

jonathan-innis:
Fixes #N/A 

**Description**

Don't reparse the selector on the TopologyGroup each time we validate whether a pod matches a selector. The call to `metav1.LabelSelectorAsSelector` causes a regex parse of the selector which is expensive to compute during the scheduling loop

### Flame Graph

Screenshot 2024-11-17 at 1 17 53 PM

`make presubmit`


njtran (commented):
/lgtm
/approve

PR #1823, opened 2024-11-17
perf: Drop unneeded recalculation for resources and available

jonathan-innis:
Fixes #N/A 

**Description**

We were recalculating `n.Available()` every time we were running through a scheduling loop which didn't end up being a cheap check when we were calculating pods for existing nodes

### Flame Graph

Screenshot 2024-11-17 at 12 32 08 AM

`make presubmit`


njtran (commented):
/lgtm
/approve

PR #1825, opened 2024-11-17
perf: Cache the pod request calculation in memory

jonathan-innis:
Fixes #N/A 

**Description**

Cache pod request calculations in memory during scheduling -- this prevents us from calling requests.RequestsForPods a ton of times during the scheduling simulation which causes us to reconstruct the requests. We can calculate this once at the beginning of the scheduling simulation and then just reached for the pre-computed cached value anytime we are trying to use a pod's requests

`make presubmit`


engedaam (commented):
/lgtm

jmdeal:
/hold
There's one more occurrence in NodeClaim which isn't using the cached value on line 110.

PR #1827, opened 2024-11-18
perf: Cache taints for existing nodes

jonathan-innis:
Fixes #N/A 

**Description**

Cache taints when checking pod tolerations for existing nodes. We were reconstructing these taints each time we attempted to schedule a pod, but we can realistically calculate this once and then re-use this calculated value

`make presubmit`


engedaam (commented):
/lgtm

PR #1831, opened 2024-11-20
chore: Limit Node Repair based by Nodepool

engedaam:
Fixes #N/A 

**Description**
- Limit node repair on a nodepool, more then 20% nodes are considered unhealthy based on the cloud provider repair policy. Karpenter will validate for each nodepool

- `make presubmit`


pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +141,38 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+		unhealthyNodeCount := lo.CountBy(nodeList.Items, func(node corev1.Node) bool {
+			nodeCondition := nodeutils.GetCondition(lo.ToPtr(node), policy.ConditionType)
+			return nodeCondition.Status == policy.ConditionStatus
+		})
+
+		// This will round up to the nearest whole number. Therefore, a unhealthy node can
+		// sometimes exceed the 20% limit. This is the same as how Kubernetes
+		// handles MaxUnavailable with PDBs. Take the case with 20% unhealthy node threshold, but
+		// 3 nodes. Karpenter will opt to allow 1 node unhealthy nodepool, rather than
+		// blocking all node repair for this nodepool.
```
jmdeal, line 168: nit: would probably be good to move this up to the method comment since this is part of the definition for "what makes a nodepool healthy".

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +72,15 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
 	}
 
+	nodePoolHealthy, err := c.isNodePoolHealthy(ctx, nodeClaim)
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	if !nodePoolHealthy {
+		log.FromContext(ctx).WithValues("nodepool", nodeClaim.Labels[v1.NodePoolLabelKey]).V(1).Info("nodepool are unhealthy")
```
jmdeal, line 80:
```suggestion
		log.FromContext(ctx).WithValues("nodepool", nodeClaim.Labels[v1.NodePoolLabelKey]).V(1).Info("nodepool is unhealthy")
```

pkg/controllers/node/health/suite_test.go
```diff
@@ -274,6 +274,61 @@ var _ = Describe("Node Health", func() {
...
+
+			// Determine to delete unhealthy nodes
+			for i := range 4 {
+				res := ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				nodeClaim = ExpectExists(ctx, env.Client, nodeClaims[i])
+				Expect(nodeClaim.DeletionTimestamp).To(BeNil())
+				Expect(res.RequeueAfter).To(BeNumerically("~", time.Minute*5, time.Second))
+			}
+		})
+		It("should allow at one node to be unhealthy in nodepool", func() {
```
jmdeal, line 307:
```suggestion
		It("should consider a NodePool healthy when a single Node is unhealthy and the threshold is exceeded", func() {
```

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +72,15 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
+		return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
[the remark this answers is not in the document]
engedaam: Reducing this down to a minute

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +141,38 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
 
 	return nil
 }
+
+// Consider a nodepool unhealthy if less then 20% of the nodes owned by the nodepool have an unhealthy condition
+// defined by the cloud provider
+func (c *Controller) isNodePoolHealthy(ctx context.Context, nodeClaim *v1.NodeClaim) (bool, error) {
```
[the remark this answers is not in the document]
engedaam: I'm now using the nodepool name here. What do you think of that?

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +141,38 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if !found {
+		return false, fmt.Errorf("nodepool cannot be found from node")
+	}
+
+	nodeList := &corev1.NodeList{}
+	if err := c.kubeClient.List(ctx, nodeList, client.MatchingLabels(map[string]string{v1.NodePoolLabelKey: nodePoolName})); err != nil {
```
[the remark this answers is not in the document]
engedaam: The nodepool label is added from the nodeclaim, which is added on nodeclaim create. I think we can take a deps on that

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +72,20 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
 	}
 
+	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if !found {
+		return reconcile.Result{}, fmt.Errorf("nodepool cannot be found from node")
```
jonathan-innis, line 77: This is going to requeue -- this shouldn't requeue IMO since this isn't recoverable -- we should perhaps think about how this is going to work for standalone NodeClaims in the future since this is going to remove support for those -- maybe something with just not considering percentage when they are standalone, or doing the whole cluster?

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +146,31 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+// Up to 20% of Nodes may be unhealthy before the NodePool becomes unhealthy (or the nearest whole number, rounding up).
+// For example, given a NodePool with three nodes, one may be unhealthy without rendering the NodePool unhealthy, even though that's 33% of the total nodes.
+// This is analogous to how minAvailable and maxUnavailable work for PodDisruptionBudgets: https://kubernetes.io/docs/tasks/run-application/configure-pdb/#rounding-logic-when-specifying-percentages.
+func (c *Controller) isNodePoolHealthy(ctx context.Context, nodePoolName string) (bool, error) {
+	nodeList := &corev1.NodeList{}
+	if err := c.kubeClient.List(ctx, nodeList, client.MatchingLabels(map[string]string{v1.NodePoolLabelKey: nodePoolName})); err != nil {
+		return false, err
+	}
+
+	for _, policy := range c.cloudProvider.RepairPolicies() {
```
jonathan-innis, line 161: Do we need a helper for this function in the CloudProvider since we are calling it in a couple places now?

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +146,31 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+		unhealthyNodeCount := lo.CountBy(nodeList.Items, func(node corev1.Node) bool {
+			nodeCondition := nodeutils.GetCondition(lo.ToPtr(node), policy.ConditionType)
+			return nodeCondition.Status == policy.ConditionStatus
+		})
+
+		threshold, err := intstr.GetScaledValueFromIntOrPercent(lo.ToPtr(intstr.FromString("20%")), len(nodeList.Items), true)
```
jonathan-innis, line 167: Can we define this as a constant somewhere?

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +146,31 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+		if err != nil {
```
jonathan-innis, line 168: AFAIK, this can't ever error -- should this be a `lo.Must`

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +72,20 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
 	}
 
+	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if !found {
+		return reconcile.Result{}, fmt.Errorf("nodepool cannot be found from node")
+	}
+
+	nodePoolHealthy, err := c.isNodePoolHealthy(ctx, nodePoolName)
```
jonathan-innis, line 80: This is screaming to add a status condition on the NodePool for observability

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +72,20 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
+	if err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
```
jonathan-innis, line 82: Is NotFound possible on a list?

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +72,20 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
+	}
+	if !nodePoolHealthy {
+		log.FromContext(ctx).WithValues("nodepool", nodeClaim.Labels[v1.NodePoolLabelKey]).V(1).Info("nodepool is unhealthy")
```
jonathan-innis, line 85: Feels like this should be a status condition metric, not a log

pkg/controllers/node/health/suite_test.go
```diff
@@ -274,6 +274,61 @@ var _ = Describe("Node Health", func() {
...
+
+			// Determine to delete unhealthy nodes
+			for i := range 4 {
+				res := ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				nodeClaim = ExpectExists(ctx, env.Client, nodeClaims[i])
+				Expect(nodeClaim.DeletionTimestamp).To(BeNil())
+				Expect(res.RequeueAfter).To(BeNumerically("~", time.Minute*1, time.Second))
+			}
+		})
+		It("should consider a NodePool healthy when a single Node is unhealthy and the threshold is exceeded", func() {
```
jonathan-innis, line 307: I _think_ I understand what you mean here -- the description might be a little confusing -- is this meant to catch the round down case?

jonathan-innis (commented):
I think we need more observability around this if we are going to be blocking action that a user would expect would happen

pkg/controllers/node/health/suite_test.go
engedaam: Yeah, it is meant to catch that case the round down case. Updated the title

pkg/controllers/node/health/controller.go
engedaam: Can we do this as a follow-up?

pkg/apis/v1/nodepool_status.go
```diff
@@ -27,6 +27,8 @@ const (
 	ConditionTypeValidationSucceeded = "ValidationSucceeded"
 	// ConditionTypeNodeClassReady = "NodeClassReady" condition indicates that underlying nodeClass was resolved and is reporting as Ready
 	ConditionTypeNodeClassReady = "NodeClassReady"
+	// ConditionTypeUnhealthy = "Unhealthy" condition indicates when the nodepool has more 20% of nodes in an unhealthy state
+	ConditionTypeUnhealthy = "Unhealthy"
```
jonathan-innis, line 31: Can we mark this status condition as alpha and only add it when we have enabled the auto-repair feature?

jonathan-innis: Should this be unhealthy or should this be a condition `Healthy` that can be marked as `False`?

pkg/controllers/node/health/controller.go
jonathan-innis: Yeah, totally -- more of a musing than something we absolutely have to do now

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +78,26 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
 	}
 
+	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if found {
+		nodePoolHealthy, err := c.isNodePoolHealthy(ctx, nodePoolName)
+		if err != nil {
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+		if !nodePoolHealthy {
+			return reconcile.Result{RequeueAfter: time.Minute}, nil
```
jonathan-innis, line 88: I don't think you need a RequeueAfter -- just watch the nodepool and requeue when there is a change to its status

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +158,59 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+	if err := c.kubeClient.List(ctx, nodeList, client.MatchingLabels(map[string]string{v1.NodePoolLabelKey: nodePoolName})); err != nil {
+		return false, err
+	}
+	nodePool := &v1.NodePool{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
+		return false, err
+	}
+	stored := nodePool.DeepCopy()
+
+	healthy := c.validateHealthyCloudProviderHealthCondition(nodeList.Items)
```
jonathan-innis, line 178:
```suggestion
	healthy := c.isHealthyForNodes(nodeList.Items)
```

nit

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +158,59 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+	if !healthy {
+		nodePool.StatusConditions().SetTrueWithReason(v1.ConditionTypeUnhealthy, "Unhealthy", fmt.Sprintf("more then %s nodes are unhealthy", allowedUnhealthyPercent.String()))
+	} else {
+		_ = nodePool.StatusConditions().Clear(v1.ConditionTypeUnhealthy)
+	}
+
+	if !equality.Semantic.DeepEqual(stored, nodePool) {
+		if err := c.kubeClient.Patch(ctx, nodePool, client.MergeFrom(stored)); err != nil {
```
jonathan-innis, line 186: This needs to patch the Status, not patch the spec like you have it right now -- as it stands, I don't think this works

jonathan-innis, line 186: This should be optimistically locked to avoid patching out status conditions that are raced with another controller

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +158,59 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+
+func (c *Controller) validateHealthyCloudProviderHealthCondition(nodes []corev1.Node) bool {
+	for _, policy := range c.cloudProvider.RepairPolicies() {
+		unhealthyNodeCount := lo.CountBy(nodes, func(node corev1.Node) bool {
+			nodeCondition := nodeutils.GetCondition(lo.ToPtr(node), policy.ConditionType)
+			return nodeCondition.Status == policy.ConditionStatus
+		})
+
+		threshold := lo.Must(intstr.GetScaledValueFromIntOrPercent(lo.ToPtr(allowedUnhealthyPercent), len(nodes), true))
+		if unhealthyNodeCount > threshold {
```
jonathan-innis, line 211: Does this make sense to be per-policy? I had assumed it would be across all policies: Is the Node healthy and what's the count for that?

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +78,26 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
+		if !nodePoolHealthy {
+			return reconcile.Result{RequeueAfter: time.Minute}, nil
+		}
+	} else {
+		clusterHealthy, err := c.isClusterHealthy(ctx)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !clusterHealthy {
+			log.FromContext(ctx).V(1).Info(fmt.Sprintf("more then %s nodes are unhealthy", allowedUnhealthyPercent.String()))
```
jonathan-innis, line 96: Event maybe? This is going to be super noisy

pkg/apis/v1/nodepool_status.go
engedaam: I added a note in a comment above , and should only be marked on the nodepool if node repair is enabled

pkg/controllers/node/health/controller.go
engedaam: This was an issue with rebasing and should now be fixed

pkg/controllers/node/health/controller.go
engedaam: Do you mean the watch the nodes here? I removed to just use events we reconcile on

pkg/controllers/node/health/controller.go
engedaam: I changed it to be with any unhealthy conditions

pkg/controllers/node/health/controller.go
engedaam: Yeah, I can see that. Just updated this to be events

pkg/apis/v1/nodepool_status.go
```diff
@@ -27,6 +27,9 @@ const (
 	ConditionTypeValidationSucceeded = "ValidationSucceeded"
 	// ConditionTypeNodeClassReady = "NodeClassReady" condition indicates that underlying nodeClass was resolved and is reporting as Ready
 	ConditionTypeNodeClassReady = "NodeClassReady"
+	// ConditionTypeHealthy = "Healthy" condition indicates when the nodepool has more 80% of nodes in an healthy state
+	// This status condition is Alpha status condition
+	ConditionTypeHealthy = "Healthy"
```
jonathan-innis, line 32: I think we still need to mark that this is an "alpha" status condition that is associated with the NodeRepair feature

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +81,26 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
 	}
 
+	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if found {
```
jonathan-innis, line 85: nit: consider adding a comment here to indicate why we are doing both of these checks

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +81,26 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
+		if !nodePoolHealthy {
+			return reconcile.Result{}, nil
+		}
+	} else {
+		clusterHealthy, err := c.isClusterHealthy(ctx)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !clusterHealthy {
+			c.recorder.Publish(UnhealthyCluster(node, fmt.Sprintf("more then %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String())))
```
jonathan-innis, line 99: Should we also fire a metric for the NodeClaim? We've tended to do both throughout our disruption controller as well

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +161,57 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+	}
+	nodePool := &v1.NodePool{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
+		return false, err
+	}
+	stored := nodePool.DeepCopy()
+
+	healthy := c.isHealthyForNodes(nodeList.Items)
+	if !healthy {
+		nodePool.StatusConditions().SetTrueWithReason(v1.ConditionTypeHealthy, "Healthy", fmt.Sprintf("more then %s nodes are unhealthy", allowedUnhealthyPercent.String()))
```
jonathan-innis, line 183:
```suggestion
		nodePool.StatusConditions().SetTrueWithReason(v1.ConditionTypeHealthy, "ExceededUnhealthyNodeThreshold", fmt.Sprintf("more then %s nodes are unhealthy", allowedUnhealthyPercent.String()))
```

pkg/controllers/node/health/controller.go
```diff
@@ -130,3 +161,57 @@ func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeCla
...
+	} else {
+		_ = nodePool.StatusConditions().Clear(v1.ConditionTypeHealthy)
```
jonathan-innis, line 185: I don't think that we should clear it -- the whole point of this condition is that we mark it as True when the NodePool is healthy -- further, we need to make sure that we mark the NodePool as healthy when there are no nodes, which is something that we are going to have to probably codify into another controller -- I actually wonder if we just hang this logic on another controller entirely (nodepool status controller) and just validate whether a NodePool is or isn't healthy when we are making a decision on these nodes

pkg/controllers/node/health/events.go
```diff
@@ -0,0 +1,36 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/karpenter/pkg/events"
+)
+
+func UnhealthyCluster(node *corev1.Node, reason string) events.Event {
+	return events.Event{
+		InvolvedObject: node,
+		Type:           corev1.EventTypeWarning,
+		Reason:         "UnhealthyNodePool",
```
jonathan-innis, line 31: There's no NodePool attached to this node, so I'm not sure this event makes sense.

Also, should we fire events for Nodes/NodeClaims that are owned by a NodePool? I think, yes. Ideally, we apply the same eventing logic to all nodes across the system

pkg/apis/v1/nodepool_status.go
engedaam: Discussed offline, dropping the observability aspect from this PR and revising at a later time

pkg/controllers/node/health/controller.go
engedaam: Discussed offline, dropping the observability aspect from this PR and revising at a later time

pkg/controllers/node/health/controller.go
```diff
@@ -70,6 +79,30 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
+	// In the case of standalone nodeclaim, validate the nodes inside the cluster are healthy before proceeding
+	// to repair the nodes
+	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if found {
+		nodePoolHealthy, err := c.isNodePoolHealthy(ctx, nodePoolName)
+		if err != nil {
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+		if !nodePoolHealthy {
+			c.recorder.Publish(UnhealthyCluster(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the nodepool", allowedUnhealthyPercent.String()))...)
```
jonathan-innis, line 92:
```suggestion
			c.recorder.Publish(UnhealthyNodePool(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the nodepool", allowedUnhealthyPercent.String()))...)
```

Should we also trigger an event directly on the nodePool when this happpens too?

pkg/controllers/node/health/events.go
```diff
@@ -0,0 +1,76 @@
...
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/events"
+)
+
+func UnhealthyNodePool(node *corev1.Node, nodeClaim *v1.NodeClaim, nodePool *v1.NodePool, reason string) []events.Event {
+	return []events.Event{
+		{
+			InvolvedObject: node,
+			Type:           corev1.EventTypeWarning,
+			Reason:         "ExceededUnhealthyNodeThreshold",
```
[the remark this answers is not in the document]
engedaam: Yeah, that would keep with our current semantic for event  names

jonathan-innis (approved):
/lgtm
/approve

PR #1832, opened 2024-11-21
chore: Remove Eventual Disruption cloud provider interface

engedaam:
Fixes #N/A 

**Description**
- Dropping Eventual Disruption as the use case is not defined well. We will take up an RFC to implement this feature if the customer use case is found.

- `make presubmit`


pkg/apis/v1/nodepool_budgets_test.go
jmdeal, line 1: Looks like we need to still drop the custom reason from the provider on line 81.

charliedmcb (approved):
Overall, lgtm.

Excited to get the convo started around this area, and ensure we get the best working solution implemented :)

FYI: changes look good, but for `pkg/controllers/disruption/drift_test.go` I only did a quick pass over it.

jonathan-innis (approved):
/lgtm
/approve

PR #1835, opened 2024-11-21
chore: fix nodeclaim going unknown during instance termination

rschalo:
Fixes #N/A 

**Description**
This fixes an issue where during deletion, the generation of an object is incremented and then the status condition controller we use transitions the root condition of the object to `Unknown` due to a mismatch in `lastSeenGeneration` and `currentGeneration` for other status conditions.

Unit tests.


jonathan-innis (approved):
/lgtm
/approve

PR #1837, opened 2024-11-24
[WIP] refactor: split termination controller

jmdeal:
Fixes #N/A 

**Description**
This PR splits the node termination controller into three. Each controller has an associated finalizer, with the final controller (Instance Termination) being unable to reconciile until the drain and volume finalizers have been removed.

|Controller|Finalizer|
|-|-|
|Drain|`karpenter.sh/drain-protection`|
|Volume Detachment|`karpenter.sh/volume-protection`|
|Instance Termination|`karpenter.sh/termination`|

This change was motivated by the increased complexity of the termination controller once additional status conditions were added for drain and volume detachment monitoring. Alternatively, the subreconciler pattern (a la the NodeClaim lifecycle controller) could have been used. However, this approach has some additional observability benefits thanks to the per-controller metrics, and is subjectively easier to test and maintain.

`make test`


jmdeal:
/hold

I'm separating out the observability changes in this PR so I can get the essential feature change in and prioritize other work. I'll come back to this refactor, the current sticking point blocking this is the rollback story when adding additional finalizers.

jmdeal:
I'm going to close this for now, I still think it's a valuable refactor but I don't have the bandwidth to come back to it any time soon.

PR #1846, opened 2024-11-26
feat: Expose drifted nodeclaim condition status to printer column output

### LINKED ISSUE #1464: Enhance v1 NodeClaim print output to show drifted state
### Description

 As a cluster admin, when I am deploying changes to nodeclaims, or nodeclasses, it is hard to view how many nodeclaims have drifted. There are Prometheus metrics, but I am not aware of a way get the nodeclaim status without jq magic from the cli. Karpenter should show the nodeclaim drifted state in the wide output.

 This feature would make it easier to track the progress of nodepool changes and upgrades, when nodes are being replaced with drift.

cnmcavoy:
Fixes #1464

**Description**
feat: Expose drifted nodeclaim condition status to printer column output. Updates the nodeclaim reconcile to not clear the condition, but set the drifted condition state to false. Tests were similarly updated for this change in behavior.

`make presubmit`

### DIFF
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -51,6 +51,10 @@ spec:
           name: NodeClass
           priority: 1
           type: string
+        - jsonPath: .status.conditions[?(@.type=="Drifted")].status
+          name: Drifted
+          priority: 1
+          type: string
       name: v1
       schema:
         openAPIV3Schema:
--- a/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
+++ b/pkg/apis/crds/karpenter.sh_nodeclaims.yaml
@@ -51,6 +51,10 @@ spec:
           name: NodeClass
           priority: 1
           type: string
+        - jsonPath: .status.conditions[?(@.type=="Drifted")].status
+          name: Drifted
+          priority: 1
+          type: string
       name: v1
       schema:
         openAPIV3Schema:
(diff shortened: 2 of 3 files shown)


kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -32,6 +32,9 @@ spec:
         - jsonPath: .status.conditions[?(@.type=="Ready")].status
           name: Ready
           type: string
+        - jsonPath: .status.conditions[?(@.type=="Drifted")].status
```
jonathan-innis, line 35: How common do you think it will be for folks to look this up? I definitely can see it as useful -- what do you think about starting this in the `-o wide` output with `priority=1` and then elevating it if we hear from more folks that they'd like to see it in the non-wide output?

cnmcavoy: Sounds good to me.

kwok/charts/crds/karpenter.sh_nodeclaims.yaml
```diff
@@ -32,6 +32,10 @@ spec:
...
```
jonathan-innis, line 35: I think we need to move the wide output after all of the non-wide output so things don't shift around in a weird way

jonathan-innis: Also, given that this could be blank, does it make more sense for this to be at the end after all of the non-blank output

jonathan-innis (approved):
/lgtm
/approve

PR #1853, opened 2024-11-29
fix: hydration race for terminating nodes

jmdeal:
Fixes #N/A 

**Description**
Separates the hydration controller from the NodeClaim lifecycle controller since we need to ensure details are hydrated regardless of the lifecycle state of the NodeClaim. The motivating example for this change is Node's which were terminating during the upgrade. Since the hydration subreconciler only acted against non-terminating NodeClaims, the details would never get propagated to these Nodes leaving them orphaned after an upgrade.

`make test`


jonathan-innis (approved):
/lgtm
/approve

pkg/controllers/node/hydration/controller.go
```diff
@@ -0,0 +1,98 @@
...
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(n.Namespace, n.Name)))
+
+	nc, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, n)
+	if err != nil {
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("hydrating node, %w", err)
+	}
+	if !nodeclaimutils.IsManaged(nc, c.cloudProvider) {
```
engedaam, line 68: How is the label going to be on the if need to hydrate the node with the label?

engedaam (approved):
/lgtm

jonathan-innis (approved):
/lgtm
/approve

PR #1854, opened 2024-11-29
fix: spurious disruption budget eventing

jmdeal:
Fixes #N/A 

**Description**
Updates the disruption budget eventing to only report budgets as blocked if there are zero allowed disruptions, and a non-zero number of nodes. Otherwise, this event is always reported NodePools with zero nodes.


engedaam (commented):
/lgtm

jonathan-innis (approved):
/lgtm
/approve

PR #1862, opened 2024-12-03
fix: Lower casing disruption count metric on reason label 

engedaam:
Fixes #N/A 

**Description**
- Disruption metrics had a reason label that consisted of some upper case values and some lower case values. This PR will make them consistently lower case

- Manually validated 


pkg/controllers/node/health/controller.go
```diff
@@ -126,7 +127,7 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 	// The deletion timestamp has successfully been set for the Node, update relevant metrics.
 	log.FromContext(ctx).V(1).Info("deleting unhealthy node")
 	metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
-		metrics.ReasonLabel:       string(unhealthyNodeCondition.Type),
+		metrics.ReasonLabel:       strings.ToLower(string(unhealthyNodeCondition.Type)),
```
jonathan-innis, line 130: Should we use snake case when we see a capital in the middle of the reason? e.g NodeNotReady becomes node_not_ready

pkg/utils/pretty/pretty.go
```diff
@@ -85,3 +86,12 @@ func Taint(t v1.Taint) string {
 	}
 	return fmt.Sprintf("%s=%s:%s", t.Key, t.Value, t.Effect)
 }
+
+var matchFirstCap = regexp.MustCompile("(.)([A-Z][a-z]+)")
+var matchAllCap = regexp.MustCompile("([a-z0-9])([A-Z])")
+
+func ToSnakeCase(str string) string {
```
jonathan-innis, line 93: Let's consider adding a test to this so that we have confidence that this logic is working

jonathan-innis (approved):
/lgtm
/approve

PR #1863, opened 2024-12-03
Fix: Set Karpenter Consolidation Timeouts to zero  

edibble21:
Fixes #N/A 

**Description**
 Set Karpenter Consolidation Timeouts to zero so metrics are fired even when there are no timeouts
make presubmit


pkg/controllers/disruption/singlenodeconsolidation.go
```diff
@@ -51,7 +51,8 @@ func (s *SingleNodeConsolidation) ComputeCommand(ctx context.Context, disruption
 	// Set a timeout
 	timeout := s.clock.Now().Add(SingleNodeConsolidationTimeoutDuration)
 	constrainedByBudgets := false
-
+	//Set ConsolidationTimeoutsTotal to zero
+	ConsolidationTimeoutsTotal.Add(0, map[string]string{consolidationTypeLabel: s.ConsolidationType()})
```
jonathan-innis, line 55: Can we set this in the `init()` where we create the metrics rather than in the actual call-path? I think we can just do a `ConsolidationTimeousTotal.Set(0)` IIRC

edibble21: So ConsolidationTimeousTotal is a [counter metric](https://pkg.go.dev/github.com/awslabs/operatorpkg@v0.0.0-20241130193813-da0c28954330/metrics#CounterMetric) so it does not have set(). Also could you direct me to the init() you are referring to here? I cannot seem to find it.

jonathan-innis: `Add(0)` seems fine then

pkg/controllers/disruption/metrics.go
```diff
@@ -30,6 +30,11 @@ const (
 	consolidationTypeLabel       = "consolidation_type"
 )
 
+func init() {
+	ConsolidationTimeoutsTotal.Add(0, map[string]string{consolidationTypeLabel: "multi"})
```
jonathan-innis, line 34: nit: Do we have consts that we can use for these?

edibble21: We have these https://github.com/kubernetes-sigs/karpenter/blob/d3907afb22aecfdf9a6f62ec22cff6169ba9b60a/pkg/controllers/disruption/multinodeconsolidation.go#L226 but not consts

jonathan-innis (approved):
/lgtm
/approve

Issue #1348, opened 2024-06-24
Migrate Ginkgo Expectations to operatorpkg


jonathan-innis:
### Description

**What problem are you trying to solve?**

Right now, we have a bunch of duplicates for "expectations" for Ginkgo testing that are already present in operatorpkg. We should be able to just do a once-off migration over to it and drop all of these exepctations from our codebase.

See: https://github.com/awslabs/operatorpkg/blob/main/test/expectations/expectations.go

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

rlanhellas:
I'll work on this one.

PR #1876, opened 2024-12-11
feat: drain and volume detachment status conditions

jmdeal:
Fixes #N/A 

**Description**
Adds status conditions for node drain and volume detachment to improve observability for the individual termination stages. This is a scoped down version of #1837, which takes these changes along with splitting each termination stage into a separate controller. I will continue to work on that refactor, but I'm decoupling to work on higher priority work.

**Status Conditions:**
|Condition|Unknown|False|True|
|-|-|-|-|
|`Drained`|Karpenter hasn't attempted to drain the Node yet **or** the node is currently draining (marked with reason)|N/A|Karpenter has successfully drained the node and will proceed with the termination flow.|
|`VolumesDetached`|Karpenter hasn't checked for volume attachments yet or there are volume attachments which are currently blocking termination. This won't transition out of unknown until `Drained` transitions to true.|There are blocking volume attachments, but the nodeclaim's TerminationGracePeriod has elapsed. This state is **only** possible if TGP is configured.|All blocking volume attachment objects have been deleted.|

`make presubmit`

### DIFF
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -78,8 +78,6 @@ func NewController(clk clock.Clock, kubeClient client.Client, cloudProvider clou
 
 func (c *Controller) Reconcile(ctx context.Context, n *corev1.Node) (reconcile.Result, error) {
 	ctx = injection.WithControllerName(ctx, "node.termination")
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(n.Namespace, n.Name)))
-
 	if !n.GetDeletionTimestamp().IsZero() {
 		return c.finalize(ctx, n)
 	}
@@ -95,22 +93,18 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
-	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
-	if err != nil {
-		// This should not occur. The NodeClaim is required to track details about the termination stage and termination grace
-		// period and will not be finalized until after the Node has been terminated by Karpenter. If there are duplicates or
-		// the nodeclaim does not exist, this indicates a customer induced error (e.g. removing finalizers or manually
-		// creating nodeclaims with matching provider IDs).
-		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
-			return reconcile.Result{}, c.associatedNodeClaimError(err)
-		}
+	nodeClaim, err := c.NodeClaimForNode(ctx, node)
+	if nodeClaim == nil || err != nil {
 		return reconcile.Result{}, err
 	}
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("NodeClaim", klog.KRef(nodeClaim.Namespace, nodeClaim.Name)))
+
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("NodeClaim", klog.KObj(nodeClaim)))
 	if nodeClaim.DeletionTimestamp.IsZero() {
 		if err := c.kubeClient.Delete(ctx, nodeClaim); err != nil {
 			if errors.IsNotFound(err) {
-				return reconcile.Result{}, c.associatedNodeClaimError(err)
+				log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+				c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+				return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 			}
 			return reconcile.Result{}, fmt.Errorf("deleting nodeclaim, %w", err)
 		}
@@ -150,7 +144,35 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 			return *result, err
 		}
 	}
+	if err := c.removeFinalizer(ctx, node); err != nil {
+		return reconcile.Result{}, err
+	}
 	return reconcile.Result{}, nil
+}
+
+func (c *Controller) NodeClaimForNode(ctx context.Context, node *corev1.Node) (*v1.NodeClaim, error) {
+	nodeClaims, err := nodeutils.GetNodeClaims(ctx, c.kubeClient, node)
+	if err != nil {
+		return nil, err
+	}
+	// This should not occur. The NodeClaim is required to track details about the termination stage and termination grace
+	// period and will not be finalized until after the Node has been terminated by Karpenter. If there are duplicates or
+	// the nodeclaim does not exist, this indicates a customer induced error (e.g. removing finalizers or manually
+	// creating nodeclaims with matching provider IDs).
+	if len(nodeClaims) == 0 {
+		log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+		c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+		return nil, nil
+	}
+	if len(nodeClaims) > 1 {
+		log.FromContext(ctx).WithValues(
+			"node", klog.KObj(node),
+			"nodeclaims", lo.Map(nodeClaims, func(nc *v1.NodeClaim, _ int) klog.ObjectRef { return klog.KObj(nc) }),
+		).Error(fmt.Errorf("duplicate nodeclaims found"), "failed to terminate node")
+		c.recorder.Publish(terminatorevents.DuplicateNodeClaimsFound(node, nodeClaims...))
+		return nil, nil
+	}
+	return nodeClaims[0], nil
 }
 
 type terminationFunc func(context.Context, *v1.NodeClaim, *corev1.Node, *time.Time) (*reconcile.Result, error)
@@ -176,7 +198,9 @@ func (c *Controller) awaitDrain(
 					return &reconcile.Result{Requeue: true}, nil
 				}
 				if errors.IsNotFound(err) {
-					return &reconcile.Result{}, c.associatedNodeClaimError(err)
+					log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+					c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+					return &reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 				}
 				return &reconcile.Result{}, err
 			}
@@ -192,7 +216,9 @@ func (c *Controller) awaitDrain(
 				return &reconcile.Result{Requeue: true}, nil
 			}
 			if errors.IsNotFound(err) {
-				return &reconcile.Result{}, c.associatedNodeClaimError(err)
+				log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+				c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+				return &reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 			}
 			return &reconcile.Result{}, err
 		}
@@ -234,7 +260,9 @@ func (c *Controller) awaitVolumeDetachment(
 					return &reconcile.Result{Requeue: true}, nil
 				}
 				if errors.IsNotFound(err) {
-					return &reconcile.Result{}, c.associatedNodeClaimError(err)
+					log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+					c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+					return &reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 				}
 				return &reconcile.Result{}, err
 			}
@@ -242,7 +270,10 @@ func (c *Controller) awaitVolumeDetachment(
 			// operations on the NodeClaim.
 			return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 		}
-	} else if !c.hasTerminationGracePeriodElapsed(nodeTerminationTime) {
+		return nil, nil
+	}
+
+	if !c.hasTerminationGracePeriodElapsed(nodeTerminationTime) {
 		// There are volume attachments blocking instance termination remaining. We should set the status condition to
 		// unknown (if not already) and requeue. This case should never fall through, to continue to instance termination
 		// one of two conditions must be met: all blocking volume attachment objects must be deleted or the nodeclaim's TGP
@@ -255,31 +286,35 @@ func (c *Controller) awaitVolumeDetachment(
 					return &reconcile.Result{Requeue: true}, nil
 				}
 				if errors.IsNotFound(err) {
-					return &reconcile.Result{}, c.associatedNodeClaimError(err)
+					log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+					c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+					return &reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 				}
 				return &reconcile.Result{}, err
 			}
 		}
 		return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
-	} else {
-		// There are volume attachments blocking instance termination remaining, but the nodeclaim's TGP has expired. In this
-		// case we should set the status condition to false (requeing if it wasn't already) and then fall through to instance
-		// termination.
-		stored := nodeClaim.DeepCopy()
-		if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeVolumesDetached, "TerminationGracePeriodElapsed", "TerminationGracePeriodElapsed"); modified {
-			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
-				if errors.IsConflict(err) {
-					return &reconcile.Result{Requeue: true}, nil
-				}
-				if errors.IsNotFound(err) {
-					return &reconcile.Result{}, c.associatedNodeClaimError(err)
-				}
-				return &reconcile.Result{}, err
+	}
+
+	// There are volume attachments blocking instance termination remaining, but the nodeclaim's TGP has expired. In this
+	// case we should set the status condition to false (requeing if it wasn't already) and then fall through to instance
+	// termination.
+	stored := nodeClaim.DeepCopy()
+	if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeVolumesDetached, "TerminationGracePeriodElapsed", "TerminationGracePeriodElapsed"); modified {
+		if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if errors.IsConflict(err) {
+				return &reconcile.Result{Requeue: true}, nil
 			}
-			// We requeue after a patch operation since we want to ensure we read our own writes before any subsequent
-			// operations on the NodeClaim.
-			return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+			if errors.IsNotFound(err) {
+				log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+				c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+				return &reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+			}
+			return &reconcile.Result{}, err
 		}
+		// We requeue after a patch operation since we want to ensure we read our own writes before any subsequent
+		// operations on the NodeClaim.
+		return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
 	return nil, nil
 }
@@ -304,21 +339,14 @@ func (c *Controller) awaitInstanceTermination(
 	if !isInstanceTerminated {
 		return &reconcile.Result{RequeueAfter: 5 * time.Second}, nil
 	}
-	if err := c.removeFinalizer(ctx, node); err != nil {
-		return &reconcile.Result{}, err
-	}
 	return nil, nil
-}
-
-func (*Controller) associatedNodeClaimError(err error) error {
-	return reconcile.TerminalError(fmt.Errorf("failed to terminate node, expected a single associated nodeclaim, %w", err))
 }
 
 func (c *Controller) hasTerminationGracePeriodElapsed(nodeTerminationTime *time.Time) bool {
 	if nodeTerminationTime == nil {
 		return false
 	}
-	return !c.clock.Now().Before(*nodeTerminationTime)
+	return c.clock.Now().After(*nodeTerminationTime)
 }
 
 func (c *Controller) pendingVolumeAttachments(ctx context.Context, node *corev1.Node) ([]*storagev1.VolumeAttachment, error) {
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -139,24 +139,49 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		}
 		return reconcile.Result{}, serrors.Wrap(fmt.Errorf("tainting node, %w", err), "taint", pretty.Taint(v1.DisruptedNoScheduleTaint))
 	}
-	if err = c.terminator.Drain(ctx, node, nodeTerminationTime); err != nil {
+
+	for _, f := range []terminationFunc{
+		c.awaitDrain,
+		c.awaitVolumeDetachment,
+		c.awaitInstanceTermination,
+	} {
+		result, err := f(ctx, nodeClaim, node, nodeTerminationTime)
+		if result != nil || err != nil {
+			return *result, err
+		}
+	}
+	return reconcile.Result{}, nil
+}
+
+type terminationFunc func(context.Context, *v1.NodeClaim, *corev1.Node, *time.Time) (*reconcile.Result, error)
+
+// awaitDrain initiates the drain of the node and will continue to requeue until the node has been drained. If the
+// nodeClaim has a terminationGracePeriod set, pods will be deleted to ensure this function does not requeue past the
+// nodeTerminationTime.
+func (c *Controller) awaitDrain(
+	ctx context.Context,
+	nodeClaim *v1.NodeClaim,
+	node *corev1.Node,
+	nodeTerminationTime *time.Time,
+) (*reconcile.Result, error) {
+	if err := c.terminator.Drain(ctx, node, nodeTerminationTime); err != nil {
 		if !terminator.IsNodeDrainError(err) {
-			return reconcile.Result{}, fmt.Errorf("draining node, %w", err)
+			return &reconcile.Result{}, fmt.Errorf("draining node, %w", err)
 		}
 		c.recorder.Publish(terminatorevents.NodeFailedToDrain(node, err))
 		stored := nodeClaim.DeepCopy()
 		if modified := nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
 			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
 				if errors.IsConflict(err) {
-					return reconcile.Result{Requeue: true}, nil
+					return &reconcile.Result{Requeue: true}, nil
 				}
 				if errors.IsNotFound(err) {
-					return reconcile.Result{}, c.associatedNodeClaimError(err)
+					return &reconcile.Result{}, c.associatedNodeClaimError(err)
 				}
-				return reconcile.Result{}, err
+				return &reconcile.Result{}, err
 			}
 		}
-		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
 	if !nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained).IsTrue() {
 		stored := nodeClaim.DeepCopy()
@@ -164,27 +189,40 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		_ = nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeDrained)
 		if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
 			if errors.IsConflict(err) {
-				return reconcile.Result{Requeue: true}, nil
+				return &reconcile.Result{Requeue: true}, nil
 			}
 			if errors.IsNotFound(err) {
-				return reconcile.Result{}, c.associatedNodeClaimError(err)
+				return &reconcile.Result{}, c.associatedNodeClaimError(err)
 			}
-			return reconcile.Result{}, err
+			return &reconcile.Result{}, err
 		}
 		NodesDrainedTotal.Inc(map[string]string{
 			metrics.NodePoolLabel: node.Labels[v1.NodePoolLabelKey],
 		})
 		// We requeue after a patch operation since we want to ensure we read our own writes before any subsequent
 		// operations on the NodeClaim.
-		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+	return nil, nil
+}
 
+// awaitVolumeDetachment will continue to requeue until all volume attachments associated with the node have been
+// deleted. The deletion is performed by the upstream attach-detach controller, Karpenter just needs to await deletion.
+// This will be skipped once the nodeClaim's terminationGracePeriod has elapsed at nodeTerminationTime.
+//
+//nolint:gocyclo
+func (c *Controller) awaitVolumeDetachment(
+	ctx context.Context,
+	nodeClaim *v1.NodeClaim,
+	node *corev1.Node,
+	nodeTerminationTime *time.Time,
+) (*reconcile.Result, error) {
 	// In order for Pods associated with PersistentVolumes to smoothly migrate from the terminating Node, we wait
 	// for VolumeAttachments of drain-able Pods to be cleaned up before terminating Node and removing its finalizer.
 	// However, if TerminationGracePeriod is configured for Node, and we are past that period, we will skip waiting.
 	pendingVolumeAttachments, err := c.pendingVolumeAttachments(ctx, node)
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("ensuring no volume attachments, %w", err)
+		return &reconcile.Result{}, fmt.Errorf("ensuring no volume attachments, %w", err)
 	}
 	if len(pendingVolumeAttachments) == 0 {
 		// There are no remaining volume attachments blocking instance termination. If we've already updated the status
@@ -193,16 +231,16 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		if modified := nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeVolumesDetached); modified {
 			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
 				if errors.IsConflict(err) {
-					return reconcile.Result{Requeue: true}, nil
+					return &reconcile.Result{Requeue: true}, nil
 				}
 				if errors.IsNotFound(err) {
-					return reconcile.Result{}, c.associatedNodeClaimError(err)
+					return &reconcile.Result{}, c.associatedNodeClaimError(err)
 				}
-				return reconcile.Result{}, err
+				return &reconcile.Result{}, err
 			}
 			// We requeue after a patch operation since we want to ensure we read our own writes before any subsequent
 			// operations on the NodeClaim.
-			return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+			return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 		}
 	} else if !c.hasTerminationGracePeriodElapsed(nodeTerminationTime) {
 		// There are volume attachments blocking instance termination remaining. We should set the status condition to
@@ -214,15 +252,15 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		if modified := nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeVolumesDetached, "AwaitingVolumeDetachment", "AwaitingVolumeDetachment"); modified {
 			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
 				if errors.IsConflict(err) {
-					return reconcile.Result{Requeue: true}, nil
+					return &reconcile.Result{Requeue: true}, nil
 				}
 				if errors.IsNotFound(err) {
-					return reconcile.Result{}, c.associatedNodeClaimError(err)
+					return &reconcile.Result{}, c.associatedNodeClaimError(err)
 				}
-				return reconcile.Result{}, err
+				return &reconcile.Result{}, err
 			}
 		}
-		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	} else {
 		// There are volume attachments blocking instance termination remaining, but the nodeclaim's TGP has expired. In this
 		// case we should set the status condition to false (requeing if it wasn't already) and then fall through to instance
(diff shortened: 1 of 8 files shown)


pkg/controllers/node/termination/controller.go
```diff
@@ -92,20 +94,39 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
+			log.FromContext(ctx).Error(err, "failed to terminate node")
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, err
+	}
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("NodeClaim", klog.KRef(nodeClaim.Namespace, nodeClaim.Name)))
+	if nodeClaim.DeletionTimestamp.IsZero() {
+		if err := c.kubeClient.Delete(ctx, nodeClaim); err != nil {
+			return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("deleting nodeclaim, %w", err))
```
njtran, line 108: wouldn't we want to client.IgnoreNotFound the error handling block so that we continue to the rest of the controller

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		// on nodes that are no longer alive. We do a check on the Ready condition of the node since, even
-		// though the CloudProvider says the instance is not around, we know that the kubelet process is still running
-		// if the Node Ready condition is true
-		// Similar logic to: https://github.com/kubernetes/kubernetes/blob/3a75a8c8d9e6a1ebd98d8572132e675d4980f184/staging/src/k8s.io/cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go#L144
-		if nodeutils.GetCondition(node, corev1.NodeReady).Status != corev1.ConditionTrue {
-			if _, err = c.cloudProvider.Get(ctx, node.Spec.ProviderID); err != nil {
-				if cloudprovider.IsNodeClaimNotFoundError(err) {
-					return reconcile.Result{}, c.removeFinalizer(ctx, node)
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
```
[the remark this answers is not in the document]
njtran: Scoping this to the drain error handling block also means that we're not going to be adding this status condition if the node was empty in the first place. From a functionality perspective this is fine, but also makes it a bit confusing to trace the steps in history later. Thoughts on this?

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+				if errors.IsConflict(err) {
+					return reconcile.Result{Requeue: true}, nil
 				}
-				return reconcile.Result{}, fmt.Errorf("getting nodeclaim, %w", err)
+				return reconcile.Result{}, client.IgnoreNotFound(err)
```
njtran, line 147: Shouldn't we be doing ignore not found in this block and just continuing if it doesn't exist?

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		_ = nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeDrained)
+		if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+		NodesDrainedTotal.Inc(map[string]string{
+			metrics.NodePoolLabel: node.Labels[v1.NodePoolLabelKey],
+		})
```
njtran, line 163: Just an observation that we still emit this metric even if we didn't do any draining.

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		// We sleep here after a patch operation since we want to ensure that we are able to read our own write before
+		// getting the NodeClaim again. This prevents conflict errors on subsequent writes.
+		// USE CAUTION when determining whether to increase this timeout or remove this line
+		time.Sleep(time.Second)
+		nodeClaim, err = nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
```
njtran, line 168: Do you think we can just return here and requeue the controller?

pkg/controllers/node/termination/controller.go
```diff
@@ -145,48 +184,59 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 			return reconcile.Result{}, fmt.Errorf("ensuring no volume attachments, %w", err)
 		}
 		if !areVolumesDetached {
+			c.recorder.Publish(terminatorevents.NodeAwaitingVolumeDetachmentEvent(node))
+			stored := nodeClaim.DeepCopy()
+			if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeVolumesDetached, "AwaitingVolumeDetachment", "AwaitingVolumeDetachment"); modified {
+				if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
```
njtran, line 190: same here on ignoring not found here rather than L194

pkg/controllers/node/termination/controller.go
```diff
@@ -145,48 +184,59 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-			return reconcile.Result{}, fmt.Errorf("ensuring instance termination, %w", err)
+			return reconcile.Result{}, client.IgnoreNotFound(err)
 		}
-		if !isInstanceTerminated {
-			return reconcile.Result{RequeueAfter: 5 * time.Second}, nil
+		// We sleep here after a patch operation since we want to ensure that we are able to read our own write before
+		// getting the NodeClaim again. This prevents conflict errors on subsequent writes.
+		// USE CAUTION when determining whether to increase this timeout or remove this line
+		time.Sleep(time.Second)
+		nodeClaim, err = nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
```
njtran, line 213: same here on just returning

pkg/controllers/node/termination/controller.go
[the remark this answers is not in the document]
jmdeal: > Do we want both Reason and Message to be Draining? Any extra details we can add here?

It would be nice, but would result in a lot of additional writes to the resource. That's why I opted to leave additional information on the event where it can be appropriately deduped.

> Scoping this to the drain error handling block also means that we're not going to be adding this status condition if the node was empty in the first place.

Yeah, this is intentional. If there were no drainable pods on the Node in the first place, it wouldn't make sense to transition the status condition to false. We should transition from unknown -> true.

pkg/controllers/node/termination/controller.go
jmdeal: If the nodeClaim isn't found, we shouldn't be able to proceed with the rest of the loop anyway. This is just a short-circuit.

pkg/controllers/node/termination/controller.go
jmdeal: I think this is what we want. I'm considering "drained" the end state, not the process. It would also be confusing / concerning to me as an operator if total nodes drained was less than the total nodes terminated, since that would indicate to me Karpenter is terminating nodes.

We do need to check though that it drained successfully, and we haven't passed over the drain block due to TGP expiration. If that's what you were calling out, you're right and I'll address that.

pkg/controllers/node/termination/controller.go
jmdeal: The main reason I didn't do that was the additional testability burden. This would increase the number of reconciliations required for the termination controller. Requiring multiple reconciliations for instance termination can already be hard enough to reason about, I'd really rather not increase this further. 

Long-term I'm still tracking #1837 which will split these stages into individual controller or subreconcilers and address this.

pkg/controllers/node/termination/controller.go
jmdeal: Discussed offline, this isn't actually an issue with drain, but it is an issue with VolumeDrained. We shouldn't set that condition to true if we proceeded due to TPG expiration.

pkg/controllers/node/termination/controller.go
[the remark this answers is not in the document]
jmdeal: Discussed some additional follow-ups, we could set the reason based on the group of pods currently being evicted (e.g. `critical`, `system-critical`, etc.) This will result in up to 4 additional writes per Node. We're going to decouple for now, but I'll open an issue to track this as an additional feature once this has merged.

pkg/controllers/node/termination/controller.go
jmdeal: Discussed offline, we'll return and requeue if the NodeClaim isn't found. An appropriate error message will then be printed by the initial get call, and we'll return without requeue.

pkg/controllers/node/termination/controller.go
```diff
@@ -92,20 +94,39 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
-	nodeClaims, err := nodeutils.GetNodeClaims(ctx, c.kubeClient, node)
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("listing nodeclaims, %w", err)
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
+			log.FromContext(ctx).Error(err, "failed to terminate node")
+			return reconcile.Result{}, nil
```
engedaam, line 101: Nit. 

Why don't we want to do anything if we find more then one nodeclaims? Are we expecting the customer to act in those cases?

pkg/controllers/node/termination/controller.go
```diff
@@ -92,20 +94,39 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
-	nodeClaims, err := nodeutils.GetNodeClaims(ctx, c.kubeClient, node)
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("listing nodeclaims, %w", err)
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
+			log.FromContext(ctx).Error(err, "failed to terminate node")
```
engedaam, line 100: Shouldn't the error message be as such?
```suggestion
			log.FromContext(ctx).Error("failed to terminate node", err)
```

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+				return reconcile.Result{}, client.IgnoreNotFound(err)
 			}
 		}
-
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
-	NodesDrainedTotal.Inc(map[string]string{
-		metrics.NodePoolLabel: node.Labels[v1.NodePoolLabelKey],
-	})
+	if !nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained).IsTrue() {
```
engedaam, line 152: What do you think about creating a function the patch call for updating the status condition, since it seems to be the same for bother status condition

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		NodesDrainedTotal.Inc(map[string]string{
+			metrics.NodePoolLabel: node.Labels[v1.NodePoolLabelKey],
+		})
+		// We sleep here after a patch operation since we want to ensure that we are able to read our own write before
+		// getting the NodeClaim again. This prevents conflict errors on subsequent writes.
+		// USE CAUTION when determining whether to increase this timeout or remove this line
+		time.Sleep(time.Second)
+		nodeClaim, err = nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
+		if err != nil {
+			if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
```
engedaam, line 170: Same question here

pkg/controllers/node/termination/controller.go
```diff
@@ -117,25 +138,43 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+	if !nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained).IsTrue() {
+		stored := nodeClaim.DeepCopy()
+		_ = nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeDrained)
+		if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, client.IgnoreNotFound(err)
+		}
+		NodesDrainedTotal.Inc(map[string]string{
```
engedaam, line 161: Wondering if we can drop this metric and relay on the operatorpkg status condition? Does this `NodesDrainedTotal` give something more then operatorpkg?

pkg/controllers/node/termination/controller.go
jmdeal: No, it's not a sprintf. This function takes the error first, and the message next.

pkg/controllers/node/termination/controller.go
jmdeal: Yes, this means that an invariant was broken (e.g. a customer created a nodeclaim manually with the same provider ID). Making the termination controller tolerant to this significantly complicates things since the NodeClaim is the source of truth for where in the termination lifecycle we are. If there are multiple NodeClaims we have multiple sources of truth. Handling that isn't worth the extra complexity.

pkg/controllers/node/termination/controller.go
jmdeal: That's a good point, there's an argument we could drop it, but I think we should do that in a separate PR since it's an existing metric and that makes it easier to track a breaking change. 

There is still an argument to keep it - controllers aren't guaranteed to capture every edge and the status condition controller could potentially miss the transition if the NodeClaim's finalizer is removed too quickly after the drain completes. I don't think this would happen in practice since we wait for instance termination to complete before removing the finalizer, but the time that takes is going to be cloudprovider dependent. If we want to ensure correctness, the separate metric is better.

pkg/controllers/node/termination/controller.go
jmdeal: Personally I think it's over abstracting, and hides the control flow of the controller. I think the root issue is that this controller has too much responsibility, and is why I've been working on splitting it here: https://github.com/kubernetes-sigs/karpenter/pull/1837. Once I have more bandwidth, I plan on coming back to that refactor.

pkg/controllers/node/termination/controller.go
engedaam: > Yes, this means that an invariant was broken (e.g. a customer created a nodeclaim manually with the same provider ID)
Is this really an invariant we have taken? We have thus far created our nodeclaims to only link to one node, but this is not something we have definitively codified, no? To me this seems like a breaking change, since we have supported terminating multiple nodeclaims and have one providerID. To be fair, I don't think anyone should do this in practice, but was mainly thinking this could break some customers.

pkg/controllers/node/termination/controller.go
engedaam: Yeah, the refactor would definitely help, but The ask here was not add more blot this controller in the mean time. This is a small and not I'm okay with keeping it as is

pkg/controllers/node/termination/controller.go
engedaam: Yeah, would be better to discuss in the refactor PR

pkg/controllers/node/termination/controller.go
jmdeal: I've discussed this pretty extensively offline with @jonathan-innis and @njtran, this was an overly defensive design back from the machine migration. This has never worked properly (e.g. TGP just picks the first NodeClaim) and is an invariant enforced by the NodeClaim creation mechanism. If we are in this state we should fail loudly instead of silently.

pkg/controllers/node/termination/controller.go
jmdeal: I've tried it both ways, personally I find it easier to read as is. I'll plan on leaving it.

pkg/controllers/node/termination/controller.go
jmdeal: I don't think it should be in the refactor PR either. I wouldn't want to hide a breaking metric change in a PR labeled refactor. I think it can just be a standalone PR.

engedaam (approved):
/lgtm

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
-	nodeClaims, err := nodeutils.GetNodeClaims(ctx, c.kubeClient, node)
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("listing nodeclaims, %w", err)
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
```
jonathan-innis, line 99: nit: Should we throw a comment over this one that indicates that we don't expect this case to happen and (if it does) then we expect that something has gone wrong and we have broken some tenant of the system?

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+			if errors.IsNotFound(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, fmt.Errorf("deleting nodeclaim, %w", err)
+		}
 	}
 
-	if err = c.deleteAllNodeClaims(ctx, nodeClaims...); err != nil {
-		return reconcile.Result{}, fmt.Errorf("deleting nodeclaims, %w", err)
+	// If the underlying NodeClaim no longer exists, we want to delete to avoid trying to gracefully drain nodes that are
```
jonathan-innis, line 114: Underlying NodeClaim? I wonder if we should just call this a CloudProvider instance because this wording confused me a bit because there is already a check for the existence of the NodeClaim above

jonathan-innis: Note: I probably wrote this but I think my own writing is confusing here :D

jonathan-innis, line 114: Remind me again: I recall there was a bug and a reason that we moved this up -- something with us getting stuck on the terminationGracePeriod and continually trying to drain even if the instance was already terminated, right?

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
 
-	nodeTerminationTime, err := c.nodeTerminationTime(node, nodeClaims...)
+	nodeTerminationTime, err := c.nodeTerminationTime(node, nodeClaim)
 	if err != nil {
 		return reconcile.Result{}, err
 	}
-
 	if err = c.terminator.Taint(ctx, node, v1.DisruptedNoScheduleTaint); err != nil {
-		if errors.IsConflict(err) {
+		if errors.IsConflict(err) || errors.IsNotFound(err) {
```
jonathan-innis, line 132: If the node is no longer found, why why would we choose to requeue in that case?

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		// on nodes that are no longer alive. We do a check on the Ready condition of the node since, even
-		// though the CloudProvider says the instance is not around, we know that the kubelet process is still running
-		// if the Node Ready condition is true
-		// Similar logic to: https://github.com/kubernetes/kubernetes/blob/3a75a8c8d9e6a1ebd98d8572132e675d4980f184/staging/src/k8s.io/cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go#L144
-		if nodeutils.GetCondition(node, corev1.NodeReady).Status != corev1.ConditionTrue {
-			if _, err = c.cloudProvider.Get(ctx, node.Spec.ProviderID); err != nil {
-				if cloudprovider.IsNodeClaimNotFoundError(err) {
-					return reconcile.Result{}, c.removeFinalizer(ctx, node)
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
```
jonathan-innis, line 143: Should it be Drained (Unknown) here since we are in the process of draining but we haven't completed our drain logic -- at which point we would mark the status as Drained=true

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
-	nodeClaims, err := nodeutils.GetNodeClaims(ctx, c.kubeClient, node)
+	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
 	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("listing nodeclaims, %w", err)
+		if nodeutils.IsDuplicateNodeClaimError(err) || nodeutils.IsNodeClaimNotFoundError(err) {
+			return reconcile.Result{}, reconcile.TerminalError(fmt.Errorf("failed to terminate node, %w", err))
```
jonathan-innis, line 100: I wonder if we should have a more specific error message that indicates that the nodeClaim existing is a pre-condition for these nodes getting properly drained and terminated

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		// if the Node Ready condition is true
-		// Similar logic to: https://github.com/kubernetes/kubernetes/blob/3a75a8c8d9e6a1ebd98d8572132e675d4980f184/staging/src/k8s.io/cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go#L144
-		if nodeutils.GetCondition(node, corev1.NodeReady).Status != corev1.ConditionTrue {
-			if _, err = c.cloudProvider.Get(ctx, node.Spec.ProviderID); err != nil {
-				if cloudprovider.IsNodeClaimNotFoundError(err) {
-					return reconcile.Result{}, c.removeFinalizer(ctx, node)
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
+			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+				if errors.IsConflict(err) || errors.IsNotFound(err) {
```
jonathan-innis, line 145: Can we put a little comment over why we are retrying on NotFound here -- since it's not consistent with our typical error handling, I think we should add some detail here if we can -- I get why you are handling it this way though

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
 		}
-
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
-	NodesDrainedTotal.Inc(map[string]string{
-		metrics.NodePoolLabel: node.Labels[v1.NodePoolLabelKey],
-	})
+	if !nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained).IsTrue() {
+		stored := nodeClaim.DeepCopy()
+		_ = nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeDrained)
```
jonathan-innis, line 155: Should we use the modified check here as well? Would probably work just as well as the IsTrue check

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-			if errors.IsConflict(err) {
-				return reconcile.Result{Requeue: true}, nil
+			// We requeue after a patch operation since we want to ensure we read our own writes before any subsequent
+			// operations on the NodeClaim.
+			return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+	} else if !c.hasTerminationGracePeriodElapsed(nodeTerminationTime) {
+		c.recorder.Publish(terminatorevents.NodeAwaitingVolumeDetachmentEvent(node))
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeVolumesDetached, "AwaitingVolumeDetachment", "AwaitingVolumeDetachment"); modified {
```
jonathan-innis, line 193: Can this also be setting to Unknown since we are in the process of Detaching the volumes so it hasn't hit a terminal state yet

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+				if errors.IsConflict(err) || errors.IsNotFound(err) {
+					return reconcile.Result{Requeue: true}, nil
+				}
+				return reconcile.Result{}, err
+			}
+		}
+		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+	} else {
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeVolumesDetached, "TerminationGracePeriodElapsed", "TerminationGracePeriodElapsed"); modified {
```
jonathan-innis, line 204: I could see us setting it to False here since this indicates that we failed to attach the volumes and we had to terminate due to hitting our terminationGracePeriod on the node

pkg/controllers/node/termination/controller.go
```diff
@@ -92,99 +94,148 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		return reconcile.Result{}, fmt.Errorf("deleting nodeclaims, %w", err)
+		return reconcile.Result{}, fmt.Errorf("ensuring no volume attachments, %w", err)
 	}
-	for _, nodeClaim := range nodeClaims {
-		isInstanceTerminated, err := termination.EnsureTerminated(ctx, c.kubeClient, nodeClaim, c.cloudProvider)
-		if err != nil {
-			// 404 = the nodeClaim no longer exists
-			if errors.IsNotFound(err) {
-				continue
+	if volumesDetached {
```
jonathan-innis, line 177: Is there a way to cleanup some of this logic so that it's not so nested -- comments might also help too -- the fact that we fall-through when volumes are detached and are able to get to the bottom of the function (we don't requeue) is a bit confusing IMO

pkg/controllers/node/termination/terminator/events/events.go
```diff
@@ -56,6 +56,16 @@ func NodeFailedToDrain(node *corev1.Node, err error) events.Event {
 	}
 }
 
+func NodeAwaitingVolumeDetachmentEvent(node *corev1.Node) events.Event {
+	return events.Event{
+		InvolvedObject: node,
+		Type:           corev1.EventTypeNormal,
+		Reason:         "AwaitingVolumeDetachment",
+		Message:        "Awaiting deletion VolumeAttachments bound to node",
```
jonathan-innis, line 64: Is it useful to list out the volume detachments that we are waiting on (or maybe a pretty list of them) in this list that we have here

pkg/controllers/node/termination/controller.go
jmdeal: Yeah, I agree it's a bit confusing. I'm good with updating the wording, but I think we use it elsewhere as well.

jmdeal: Since we were only checking this in the drain logic, if we drained but were stuck awaiting volume attachments we never hit this check and could get stuck indefinitely. I don't think there was any interaction with `terminationGracePeriod`, if anything it would save users in that case.

pkg/controllers/node/termination/controller.go
jmdeal: I think this came out of a previous review, this deduplicated the Node not found logic by just relying on the check at the top, at the cost of an extra reconcile.

pkg/controllers/node/termination/controller.go
jmdeal: My thought is at this point we know it is not drained, since it's in the process of draining. Whereas before we do the check we don't know if there are any drainable pods on the node, so drained is unknown.

pkg/controllers/node/termination/controller.go
jmdeal: I didn't because we knew it always would be and this limited the scope of `stored`, I can go either way though. The alternative would be:
```
if nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeDrained) {
  // remaining logic
}

File: pkg/controllers/node/termination/terminator/events/events.go (hunk 23 above)
Comment (jmdeal) [reply pkg/controllers/node/termination/terminator/events/events.go:64]: I didn't for a similar reason that I didn't include the pods on the `Drained` status condition - we would either be hammering the API server as the list changes or the information would be out of date frequently. The former

File: pkg/controllers/node/termination/controller.go (hunk 20 above)
Comment (jmdeal) [reply pkg/controllers/node/termination/controller.go:193]: Similar to `Drained`, I think it's more clear to set it to `False` here since we know the volumes aren't detached. Unknown indicates to me that we don't know one way or the other.

File: pkg/controllers/node/termination/terminator/events/events.go (hunk 23 above)
Comment (jonathan-innis) [reply pkg/controllers/node/termination/terminator/events/events.go:64]: > hammering the API server as the list changes or the information would be out of date frequently

You can make it so that the events are only fired at a certain frequency and they are deduped without considering their message -- honestly, I think including the extra info could be helpful and, if there are ones that are actually stuck, would be really valuable information for a user to know

File: pkg/controllers/node/termination/controller.go (hunk 20 above)
Comment (jonathan-innis) [reply pkg/controllers/node/termination/controller.go:193]: I think this is counter to how we have been treating status conditions throughout the project -- we are indicating with status conditions that a process hasn't completed and we don't know whether it's going to succeed or fail (for instance, the Completed status condition for a job doesn't go into a False state while the job is running, it stays in Unknown because we don't know if the job is going to complete or not and then transitions to True/False based on whether it entered a terminal state or not

File: pkg/controllers/node/termination/controller.go (hunk 15 above)
Comment (jonathan-innis) [reply pkg/controllers/node/termination/controller.go:132]: Right -- it just seems odd because it's completely counter to every other controller where we perform this logic -- seems odd to not handle it because the error check is the same anyways and the check is free

File: pkg/controllers/node/termination/controller.go (hunk 16 above)
Comment (jonathan-innis) [reply pkg/controllers/node/termination/controller.go:143]: See: https://github.com/kubernetes-sigs/karpenter/pull/1876#discussion_r1927958481 but I personally disagree with this framing -- personally, I think we should have done InstanceTerminated and then gone from Unknown to True/False as well there -- transitioning from True -> False or False -> True for a terminal status condition in general is a little odd because it suggests that the process has finished when in fact it hasn't

File: pkg/controllers/node/termination/controller.go (hunk 19 above)
Comment (jonathan-innis) [reply pkg/controllers/node/termination/controller.go:155]: I don't really think it tangibly matters -- good either way, just thought it was odd that we didn't keep consistent style here

File: pkg/controllers/node/termination/controller.go (hunk 15 above)
Comment (jmdeal) [reply pkg/controllers/node/termination/controller.go]: Agreed offline to go with the original approach and duplicate the check and error, rather than rereconciling.

File: pkg/controllers/node/termination/controller.go (hunk 16 above)
Comment (jmdeal) [reply pkg/controllers/node/termination/controller.go:147]: Discussed more offline, agreed to stay consistent with the existing status conditions (e.g. `Launched`, `Initialized`, etc.) Honestly, I still find it strange to frame it as unknown given that we know the instance hasn't been drained, but I think the consistency is more valuable here. The same argument could have been made for the existing conditions.

File: pkg/controllers/node/termination/controller.go (hunk 20 above)
Comment (jmdeal) [reply pkg/controllers/node/termination/controller.go:206]: Discussed more offline, we'll go with `Unknown` for the non-terminal states.

File: pkg/controllers/node/termination/controller.go (hunk 22 above)
Comment (jmdeal) [reply pkg/controllers/node/termination/controller.go]: I've messed around with a few different ways of structuring this, but this ends up being the most readable IMO. I will add some comments to call out the fall-through behavior. FWIW, this does clean up in my larger refactor PR.

File: pkg/utils/node/node.go [hunk 24]
```diff
@@ -64,6 +64,7 @@ func IgnoreNodeClaimNotFoundError(err error) error {
 // DuplicateNodeClaimError is an error returned when multiple v1.NodeClaims are found matching the passed providerID
 type DuplicateNodeClaimError struct {
 	ProviderID string
+	NodeClaims []string
```
jonathan-innis, line 67:
```suggestion
	NodeClaimNames []string
```

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -60,6 +60,7 @@ func TestAPIs(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
+	fakeClock = clock.NewFakeClock(time.Now())
```
jonathan-innis, line 63: Why did you add the fakeClock here -- doesn't look like it is being used anywhere in this testing?

pkg/controllers/node/termination/controller.go
jonathan-innis: Maybe throw a comment over this to explain why it's coded this way? I found myself guessing about it again as I was reading through it -- had to re-read the context to understand that this is because of the patch

pkg/controllers/node/termination/controller.go
```diff
@@ -92,112 +94,193 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		return reconcile.Result{}, fmt.Errorf("deleting nodeclaims, %w", err)
+		return reconcile.Result{}, fmt.Errorf("ensuring no volume attachments, %w", err)
 	}
-	for _, nodeClaim := range nodeClaims {
-		isInstanceTerminated, err := termination.EnsureTerminated(ctx, c.kubeClient, nodeClaim, c.cloudProvider)
-		if err != nil {
-			// 404 = the nodeClaim no longer exists
-			if errors.IsNotFound(err) {
-				continue
+	if len(pendingVolumeAttachments) == 0 {
```
jonathan-innis, line 187: Given how large this volume attachment code is now, what do you think about moving this into a separate function?

pkg/controllers/node/termination/controller.go
```diff
@@ -92,112 +94,193 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
 		}
 	}
+
+	isInstanceTerminated, err := termination.EnsureTerminated(ctx, c.kubeClient, nodeClaim, c.cloudProvider)
+	if client.IgnoreNotFound(err) != nil {
+		// 409 - The nodeClaim exists, but its status has already been modified
+		if errors.IsConflict(err) {
+			return reconcile.Result{Requeue: true}, nil
+		}
+		return reconcile.Result{}, fmt.Errorf("ensuring instance termination, %w", err)
```
jonathan-innis, line 251: We don't seem to handle the NotFound error that we get back from this EnsureTerminated call in the same way as we do elsewhere here

pkg/controllers/node/termination/suite_test.go
```diff
@@ -908,3 +957,13 @@ func ExpectNodeWithNodeClaimDraining(c client.Client, nodeName string) *corev1.N
 	Expect(node.DeletionTimestamp).ToNot(BeNil())
 	return node
 }
+
+func ExpectNotRequeued(result reconcile.Result) {
```
jonathan-innis, line 961: These are two functions that I think could be generally useful and I would consider putting in operatorpkg

jonathan-innis (commented):
Close to me! Just a few small things! Nice work!

pkg/controllers/node/termination/terminator/suite_test.go
jmdeal: It's passed down to the terminator and was causing nil panics. We could probably scope it down and just pass it there, but I don't mind keeping it test global so it's obvious to use the fake clock passed to the terminator rather than a new one in any future tests that use it.

pkg/controllers/node/termination/controller.go
jmdeal: I've been resistant to that because I find the requeue logic to be significantly harder to follow as soon as you move it into a separate function - I had a similar issue when I tried to structure this as subreconcilers. I had been punting those architectural changes for my larger refactor PR. If you don't feel strongly about it I'd rather punt for that, since I think the longer reconcile function is a better than obfuscating the control flow, but let me know what you think.

pkg/controllers/node/termination/suite_test.go
jmdeal: Good call, I've opened a PR: https://github.com/awslabs/operatorpkg/pull/130

jmdeal: I'm only just getting back to this, but I think we can defer migration for https://github.com/kubernetes-sigs/karpenter/issues/1348 and continue to use the local version for now. This way we don't have a mix of operatorpkg and local expectations.

pkg/controllers/node/termination/controller.go
jmdeal: I could go either way on this - if we're patching the NodeClaim that means we successfully initiated instance termination. Do we want to orphan the Node at that point because we're unable to patch the NodeClaim? It is an invariant we should maintain, so I could see the argument for consistency.

pkg/controllers/node/termination/controller.go
jmdeal: I've not had the bandwidth to work on my larger refactor, so I'm going to go ahead and factor this out. Let me know what you think.

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +95,243 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		// 409 - The nodeClaim exists, but its status has already been modified
+		if errors.IsConflict(err) {
+			return &reconcile.Result{Requeue: true}, nil
+		}
+		return &reconcile.Result{}, fmt.Errorf("ensuring instance termination, %w", err)
+	}
+	if !isInstanceTerminated {
+		return &reconcile.Result{RequeueAfter: 5 * time.Second}, nil
+	}
 	if err := c.removeFinalizer(ctx, node); err != nil {
```
jonathan-innis, line 307: Can we remove the finalizer outside of this function? Feels weird that each thing is waiting for a pre-condition, but only this one actually removes the finalizer at the end

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +95,243 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+func (*Controller) associatedNodeClaimError(err error) error {
+	return reconcile.TerminalError(fmt.Errorf("failed to terminate node, expected a single associated nodeclaim, %w", err))
+}
+
+func (c *Controller) hasTerminationGracePeriodElapsed(nodeTerminationTime *time.Time) bool {
+	if nodeTerminationTime == nil {
+		return false
 	}
-	return nil
+	return !c.clock.Now().Before(*nodeTerminationTime)
```
jonathan-innis, line 321: Can we just do `.After`? Yes, it's technically "not correct" but it practically doesn't matter and makes it so much more readable

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +95,243 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		return reconcile.Result{}, serrors.Wrap(fmt.Errorf("tainting node, %w", err), "taint", pretty.Taint(v1.DisruptedNoScheduleTaint))
+	}
+
+	for _, f := range []terminationFunc{
+		c.awaitDrain,
+		c.awaitVolumeDetachment,
+		c.awaitInstanceTermination,
+	} {
+		result, err := f(ctx, nodeClaim, node, nodeTerminationTime)
+		if result != nil || err != nil {
```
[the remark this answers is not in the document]
jmdeal: Generally when I've used this pattern I haven't because it doesn't give you a way to encode that you should return without requeueing, and the presence of a `reconcile.Result{}` was a clear indicator that you wouldn't "fall-through" to the next subreconciler. I can go either way here though since we don't need to do that.

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +95,243 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		// on nodes that are no longer alive. We do a check on the Ready condition of the node since, even
-		// though the CloudProvider says the instance is not around, we know that the kubelet process is still running
-		// if the Node Ready condition is true
-		// Similar logic to: https://github.com/kubernetes/kubernetes/blob/3a75a8c8d9e6a1ebd98d8572132e675d4980f184/staging/src/k8s.io/cloud-provider/controllers/nodelifecycle/node_lifecycle_controller.go#L144
-		if nodeutils.GetCondition(node, corev1.NodeReady).Status != corev1.ConditionTrue {
-			if _, err = c.cloudProvider.Get(ctx, node.Spec.ProviderID); err != nil {
-				if cloudprovider.IsNodeClaimNotFoundError(err) {
-					return reconcile.Result{}, c.removeFinalizer(ctx, node)
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
```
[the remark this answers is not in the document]
jmdeal: The reason I chose not to was to reduce API load from updating the status condition. It would be one thing if it was a log, but I don't want to be updating the status for every terminating NodeClaim once a second.

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +95,243 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+			return &reconcile.Result{RequeueAfter: 1 * time.Second}, nil
+		}
+	} else if !c.hasTerminationGracePeriodElapsed(nodeTerminationTime) {
+		// There are volume attachments blocking instance termination remaining. We should set the status condition to
+		// unknown (if not already) and requeue. This case should never fall through, to continue to instance termination
+		// one of two conditions must be met: all blocking volume attachment objects must be deleted or the nodeclaim's TGP
+		// must have expired.
+		c.recorder.Publish(terminatorevents.NodeAwaitingVolumeDetachmentEvent(node, pendingVolumeAttachments...))
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeVolumesDetached, "AwaitingVolumeDetachment", "AwaitingVolumeDetachment"); modified {
```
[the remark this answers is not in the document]
jmdeal: Similar story to the draining message - I don't want to include a ton of detail here and opted to include detail on the event.

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +93,273 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		if modified := nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
+			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+				if errors.IsConflict(err) {
+					return &reconcile.Result{Requeue: true}, nil
 				}
-				return reconcile.Result{}, fmt.Errorf("getting nodeclaim, %w", err)
+				if errors.IsNotFound(err) {
+					log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
+					c.recorder.Publish(terminatorevents.NodeClaimNotFound(node))
+					return &reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
jonathan-innis, line 203: The more I think about the amount of error handling you are having to do now in this controller to handle NotFound and do this custom requeue logic for eventing, the more I think a separate controller might be the better place to do this health check to avoid having to handle this everywhere in termination -- also, it would be much more strongly encapsulated than it is right now

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +93,273 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
-	nodeClaims, err := nodeutils.GetNodeClaims(ctx, c.kubeClient, node)
-	if err != nil {
-		return reconcile.Result{}, fmt.Errorf("listing nodeclaims, %w", err)
+	nodeClaim, err := c.NodeClaimForNode(ctx, node)
```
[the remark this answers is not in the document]
jmdeal: The issue with the existing function was that it didn't return the duplicate NodeClaims, which I wanted to include in the error message for observability purposes. Honestly, it would probably make more sense to update that function to include the NodeClaims as part of a structured error. That being said, I think I agree with your other comment where it makes more sense to do this validation elsewhere and handle both of these as terminal errors here.

pkg/controllers/node/termination/controller.go
```diff
@@ -94,114 +93,273 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-					return reconcile.Result{}, c.removeFinalizer(ctx, node)
+		stored := nodeClaim.DeepCopy()
+		if modified := nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining"); modified {
+			if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+				if errors.IsConflict(err) {
+					return &reconcile.Result{Requeue: true}, nil
 				}
-				return reconcile.Result{}, fmt.Errorf("getting nodeclaim, %w", err)
+				if errors.IsNotFound(err) {
+					log.FromContext(ctx).WithValues("node", klog.KObj(node)).Error(fmt.Errorf("nodeclaim not found"), "failed to terminate node")
```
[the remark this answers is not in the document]
jmdeal: So it does, didn't realize that but I had assumed it would have been injected in the reconciler adapter but I guess it's injected further up the call stack: https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/builder/controller.go#L447-L454.

pkg/controllers/node/termination/controller.go
jmdeal: Discussed offline - we're going to take a middleground approach and not handle the event for now. We'll take a backlog task to consider a separate health controller.

jonathan-innis (approved):
/lgtm
/approve

PR #1885, opened 2024-12-17
RFC: Degraded NodePool Status Condition

rschalo:
**Description**
See Design File for full RFC but this change seeks to add the `Degraded` status condition to NodePools in order to better communicate to users when a NodePool may be misconfigured.

### DIFF
--- a/designs/degraded-nodepools.md
+++ b/designs/degraded-nodepools.md
@@ -1,4 +1,4 @@
-# RFC: Degraded NodePool Status Condition
+# RFC: Misconfigured NodePool Observability
 
 ## Overview
 
@@ -6,11 +6,11 @@ Karpenter can launch nodes with a NodePool that will never join a cluster when a
 
 One example is that when a network path does not exist due to a misconfigured VPC (network access control lists, subnets, route tables), Karpenter will not be able to provision compute with that NodeClass that joins the cluster until the error is fixed. Crucially, this will continue to charge users for compute that can never be used in a cluster.
 
-To improve visibility of these possible failure modes, this RFC proposes the addition of a `Degraded` status condition that indicates to cluster admins there may be a problem with a NodePool needs to be investigated and corrected.
+To improve visibility of these failure modes, this RFC proposes mechanisms that indicate to cluster users there may be a problem with a NodePool/NodeClass combination that needs to be investigated and corrected.
 
 ## Options
 
-### [Recommended] Option 1: Introduce a Generalized `Degraded` Status Condition
+### Option 1: Introduce a `Degraded` Status Condition on the NodePool
 
 ```
 // ConditionTypeDegraded = "Degraded" condition indicates that a misconfiguration exists that prevents the normal, successful use of a Karpenter resource
@@ -53,9 +53,10 @@ Finally, this status condition would not be a precondition for NodePool `Readine
 
 #### Considerations
 
-1. 👎 Three retries can still be a long time to wait on compute that would never succeed
-2. 👍 Observability improvements so that users can begin triaging misconfigurations
-3. 👍 `Degraded` is not a pre-condition for NodePool readiness
+1. 👎 Three retries can still be a long time to wait on compute that never provisions correctly
+2. 👎 Heuristics can be wrong and mask failures
+3. 👍 Observability improvements so that users can begin triaging misconfigurations
+4. 👍 `Degraded` is not a pre-condition for NodePool readiness
 
 ### Option 2: Expand `Validated` Status Condition and Use Reasons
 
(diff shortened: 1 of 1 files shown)


sftim (commented):
There's another option. Conditions can have three values, so we can have a condition `Stable` (or a different name).

If Karpenter self-detects flapping, or the logic for the proposed `Degraded` condition were to fire, `Stable` can become false. Crucially, if Karpenter is _not sure_, the `Stable` condition can become Unknown and we can provide a `reason` string for that.

For example, I'd set a brand-new NodePool to `Stable` Unknown, and eagerly transition (from `Stable` false to `Stable` unknown) as soon as just one NodeClaim fails to provision.

rschalo:
So in the case of utilizing `Unknown`, we'd still not have the new status condition be a pre-condition for readiness and then eagerly switch into `Unknown`... Yea, I think that makes sense as a third option and still leaves the possibility of it being able to affect functionality in the future if we want.

Flapping in and out of `Unknown` seems fine if we don't block provisioning for that NodePool. Do you see any disadvantages to changing the status condition from `Unknown` to `True` after a single success?

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+## Options
+
+### [Recommended] Option 1: Introduce a Generalized `Degraded` Status Condition
+
+```
+// ConditionTypeDegraded = "Degraded" condition indicates that a misconfiguration exists that prevents the normal, successful use of a Karpenter resource
+ConditionTypeDegraded = "Degraded"
+```
+
+This option would set `Degraded: true` on a NodePool whenever Karpenter suspects something is wrong with the launch path but isn't sure. In this case, if 3 or more NodeClaims fail to launch with a NodePool then the NodePool will be marked as degraded. The retries are included to account for transient errors. The number of failed launches is stored as a status on the NodePool and then reset to zero following an edit to the NodePool or a sufficient amount time has passed.
```
njtran, line 20: Have you thought of something we could do to make it scale to the size of the cluster? And which errors are included in "failing to launch"?

njtran: > The number of failed launches is stored as a status on the NodePool and then reset to zero following an edit to the NodePool or a sufficient amount time has passed.

I worry about false positives for "waiting to edit" the nodepool. I'm guessing this would be bubbled up from the nodeclass, but I think having the "amount of time" as a backstop makes sense. What amount of time are you thinking?

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+
+```
+// NodePoolStatus defines the observed state of NodePool
+type NodePoolStatus struct {
+	// Resources is the list of resources that have been provisioned.
+	// +optional
+	Resources v1.ResourceList `json:"resources,omitempty"`
+	// FailedLaunches tracks the number of times a nodepool failed before being marked degraded
+	// +optional
+	FailedLaunches int `json:"failedLaunches,omitempty"`
```
njtran, line 30: How can we indicate that this is "failed launches in a row"? Does this mechanically reset if we get a successful launch?
```suggestion
	FailedSuccessiveLaunches int `json:"failedLaunches,omitempty"`
```

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+	// Conditions contains signals for health and readiness
+	// +optional
+	Conditions []status.Condition `json:"conditions,omitempty"`
+}
+```
+
+Once a NodePool is `Degraded`, it recovers with `Degraded: false` after an update to the NodePool or when the NodeClaim registration expiration TTL (currently 15 minutes) passes since the `lastTransitionTime` for the status condition on the NodePool, whichever comes first. A `Degraded` NodePool is not passed over when provisioning and may continue to be chosen during scheduling. A successful provisioning could also remove the status condition but this may cause more apiserver and metric churn than is necessary.
```
njtran, line 37: > Once a NodePool is `Degraded`, it recovers with `Degraded: false` after an update to the NodePool or when the NodeClaim registration expiration TTL (currently 15 minutes) passes since the `lastTransitionTime` for the status condition on the NodePool, whichever comes first.

Doesn't this defeat the purpose? if it's 15m then I imagine if I'm having bootstrapping errors that I'll almost never hit this feature if I'm not scaling up 3 nodes at a time.

njtran: > A `Degraded` NodePool is not passed over when provisioning and may continue to be chosen during scheduling.

Maybe eventually configurable if we get it right. i could imagine users wanting to orchestrate a signal (or maybe leverage this one) to signal to Karpenter when they think something's wrong.

njtran: > A successful provisioning could also remove the status condition but this may cause more apiserver and metric churn than is necessary.

This seems like a natural time to un mark the status condition. This only causes churn if the issue is ephemeral or racey, which in the ephemeral case, I'd imagine the NodePool isn't degraded. I'd like to understand what some of the racey cases would be though.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+
+As additional misconfigurations are handled, they can be added to the `Degraded` status condition and the `Degraded` controller expanded to handle automated recovery efforts. This is probably most simpoly achieved by changing the Status Condition metrics to use comma-delimiting for `Reason`s with the most recent change present in the `Message`.
```
njtran, line 39: Is there prior art for doing this? I'm wondering if we're actually going to be able to bubble up these failures correctly anyway, or if this is work we're carving out that wouldn't even be usable.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+    message: "FizzBuzz component was misconfigured"
+    observedGeneration: 1
+    reason: FizzBuzzFailure,FooBarFailure
+    status: "True"
+    type: Degraded
+```
+
+This introduces challenges when determining when to evaluate contributors to the status condition but since the `Degraded` status condition only has a single contributor this decision can be punted. When the time comes to implement the multiple contributors to this status condition, this probably looks like a `Degraded` controller which acts as a "heartbeat" and evaluates each of the contributors.
+
+Finally, this status condition would not be a precondition for NodePool `Readiness` because the NodePool should still be considered for scheduling purposes.
```
njtran, line 52: But this would need to be rolled up somehow from the NodeClass too, right? So we'd need a new nodepool "Degraded" condition which relies on the nodeclass "Degraded"

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+
+#### Considerations
+
+1. 👎 Three retries can still be a long time to wait on compute that would never succeed
```
njtran, line 56: Maybe I misunderstood, but resetting the timer on the registration timeout means we never hit this, right?

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+2. 👍 Karpenter continues to try and launch with other potentially valid NodePools
```
njtran, line 57: I thought we weren't opting it out of scheduling?

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+3. 👍 Observability improvements so that users can begin triaging misconfigurations
+4. 👍 `Degraded` is not a pre-condition for NodePool readiness
+
+### Option 2: Expand `Validated` Status Condition and Use Reasons
+
+The implementation is roughly the same except that validation is a pre-condition for `Readiness`. This has impact in a larger portion of the code because `Validated` would no longer block provisioning or `Readiness`. However, it is still an option that Karpenter could expand the `Valdiated` status condition so that any time a misconfiguration is encountered, the NodePool is treated as having failed validation.
+
+#### Considerations
+
+1. 👎👎 Validation implies the original state of the NodePool was correct and is something Karpenter can vet with certainty. A NodePool could have been correctly validated but then degraded.
```
njtran, line 67: False positives. I think the severity of this con depends on what signals we're choosing to include.

njtran (commented):
I like this idea. Let me know if you have questions.

designs/degraded-nodepools.md
rschalo: Yup - missed in an update. Posting one now.

designs/degraded-nodepools.md
rschalo: I was thinking that the edit to a NodePool would be a signal that a user updated the NodePool and is expecting the issue to be resolved.

>What amount of time are you thinking?

To keep it simple, the NodeClaim registration TTL (15 minutes).

designs/degraded-nodepools.md
rschalo: >Doesn't this defeat the purpose? if it's 15m then I imagine if I'm having bootstrapping errors that I'll almost never hit this feature if I'm not scaling up 3 nodes at a time.

Yea, I was initially thinking it was the same as NodeClaim registration but it's probably 3x it to address this issue.

>Maybe eventually configurable if we get it right. i could imagine users wanting to orchestrate a signal (or maybe leverage this one) to signal to Karpenter when they think something's wrong.

Yea, I think for introducing this we don't want to make functional changes yet but it could be done in the future. One thing I want to sort out is how to appropriately add/filter metrics for NodeClaims being provisioned for NodePools that are degraded.

designs/degraded-nodepools.md
rschalo: This is probably actually `FailedRegistrations`

Bryce-Soghigian: So the number of times we hit registration TTL? 

The azure provider will use the native controller retries for successive launches of a nodeclaim. 

Attempt 1: Try with Cheapest sku  -> FAIL Quota error(need a better quota api), cache sku as lacking quota for 
Attempt 2: Try again but with the next cheapest sku -> Succeed. 


So sometimes with our provider we will have an expected launch failure, that we know will succeed on the retry. 

I like FailedRegistrations better, but also for some of these LRO operations, being able to report back from the cloudprovider side when a launch fails would be useful.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,175 @@
...
+
+```
+// NodePoolStatus defines the observed state of NodePool
+type NodePoolStatus struct {
+	// Resources is the list of resources that have been provisioned.
+	// +optional
+	Resources v1.ResourceList `json:"resources,omitempty"`
+	// LaunchedNode indicates if the current NodePool configuration has launched a node.
+	// +optional
+	Verified bool `json:"verified,omitempty"`
```
Bryce-Soghigian, line 119:
```suggestion
	// Verified indicates if the current NodePool configuration has launched a node.
	// +optional
	Verified bool `json:"verified,omitempty"`
```

Using language like launch is interesting, the node has launched but to a user is it clear the differences between Launched, Registered, Initialized etc?  Are we going to use the Launched condition, or will we check that its gone through the full nodeclaim lifecycle before verifying a nodepool?  

Maybe they aren't configuring labels for daemonsets on their nodes correctly, so a daemonset dropping configuration for CNI never initializes and pods can't schedule, the nodes startup taints for that CNI are never are removed, but from the controllers perspective we have moved past a launch so its considered verified. 

Should we only consider a nodepool verified if a nodeclaim has gone through all the create lifecycle operations? (Launch + Initialization + Registration)

designs/degraded-nodepools.md
sftim: `reason` should be a single machine readable value and we **don't**delimit it

If we wanted multiple reasons, the API would support a list here.

sftim: We can fire Events, though, and I think we can even store annotations onto those Events if we want to.

designs/degraded-nodepools.md
sftim: Maybe this should be `nodeClaimRecentLaunchSuccessRatio`. We really care about recent launch attempts and failures, not the entire lifecycle of the NodePool. But I also don't think that one successful launch should discard the info here.

Imagine if the NodePool spans multiple zones but is misconfigured, except for in a zone that the workload hardly uses. (eg: workload uses zones 1 and 2; cluster supports zones 1, 2 and 3). Launching in zone 3 might misleading make things look better than they are.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Option 2: Expand `Validated` Status Condition and Use Reasons
+
+The implementation is roughly the same except that validation is a pre-condition for `Readiness`. This has impact in a larger portion of the code because `Validated` would no longer block provisioning or `Readiness`. However, it is still an option that Karpenter could expand the `Valdiated` status condition so that any time a misconfiguration is encountered, the NodePool is treated as having failed validation.
+
+#### Considerations
+
+1. 👎👎 Validation implies the original state of the NodePool was correct and is something Karpenter can vet with certainty. A NodePool could have been correctly validated but then degraded.
+2. 👎👎 Changes the meaning of `Validated` in terms of `Readiness`
+3. 👎 Relies on statuses that were not part of the original validation of a NodePool.
```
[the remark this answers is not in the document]
sftim: I think a NodePool can start valid and become invalid (eg due to someone writing a new spec, or because they deleted the node class object, or potentially even by the cloud provider changing its offering).

- Example of a NodePool becoming invalid through a spec update: set a requirement for `karpenter.sh/capacity-type: ["nonexistent"]`
- Maybe if someone sets `expireAfter` to zero, that should also be invalid?

designs/degraded-nodepools.md
rschalo: Agree that the csv style reasons is a bad idea. 
>We can fire Events, though, and I think we can even store annotations onto those Events if we want to.

I think it's possible we fire an event for the NodeClaim if the backing compute fails to join the cluster. The only time these events would spam is if the NodePool/NodeClass is so misconfigured that every launch fails. Looks like we currently only event for initial NodeClaim launch failures and not for failures in Registration and Initialization.

designs/degraded-nodepools.md
rschalo: We should probably just track on the NodePool but we can do this in a way where which NodeClass was used is stored as a hash on the NodePool (see approach 4).

designs/degraded-nodepools.md
rschalo: >the node has launched but to a user is it clear the differences between Launched, Registered, Initialized etc?

>we have moved past a launch so its considered verified.

>Are we going to use the Launched condition, will we check that its gone through the full nodeclaim lifecycle before verifying a nodepool?

My mistake in wording - I think we consider a NodePool `Verified` after all NodeClaim lifecycle operations succeed for a given NodePool/NodeClass. 

As far as Karpenter could tell, everything _should_ have succeeded. Karpenter didn't see a cloudprovider create error but some other misconfiguration prevented a node from joining the cluster. If the NodePool/NodeClass has been verified then we can fire an event on the NodePool about the most recent failure with a probable cause. That could be the startup taints issue you mention or hitting registration TTLs because of no network path.

rschalo: Also, if a nodepool is updated then these verifications should be cleared.

rschalo:
Closing in favor of https://github.com/kubernetes-sigs/karpenter/pull/1910

PR #1888, opened 2024-12-18
chore: Avoid multiple logs messages for the node health controller

engedaam:
Fixes #N/A 

**Description**
- Avoid creating multiple log message on a repair action for an unhealthy node

- `make presubmit`

### DIFF
--- a/pkg/controllers/node/health/controller.go
+++ b/pkg/controllers/node/health/controller.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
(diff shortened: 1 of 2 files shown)


pkg/controllers/node/health/controller.go
```diff
@@ -155,11 +164,25 @@ func (c *Controller) findUnhealthyConditions(node *corev1.Node) (nc *corev1.Node
 }
 
 func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	if expirationTimeString, exists := nodeClaim.ObjectMeta.Annotations[v1.NodeClaimTerminationTimestampAnnotationKey]; exists {
+		expirationTime, err := time.Parse(time.RFC3339, expirationTimeString)
```
njtran, line 168: if we can't parse it, should we override it? Or should we fail close?

pkg/controllers/node/health/controller.go
```diff
@@ -155,11 +164,25 @@ func (c *Controller) findUnhealthyConditions(node *corev1.Node) (nc *corev1.Node
...
+		if err != nil {
+			return err
+		}
+		if c.clock.Now().Before(expirationTime) {
```
njtran, line 172: shouldn't this be the opposite? 

if the current time is before the expiration time, we actually do want to annotate it. If it isn't, then we keep it as is. 

Can you add a test for this?

pkg/controllers/node/health/controller.go
engedaam: Yeah, it seems reasonable to override it if we can't parse the time.

njtran: Seems like we fail close then if we can't parse it. Can you make that change? The extra call is definitely warranted here

pkg/controllers/node/health/controller.go
engedaam: Yeah, this was the wrong way around, and added a test to validate the behavior

pkg/controllers/node/health/controller.go
```diff
@@ -155,11 +164,24 @@ func (c *Controller) findUnhealthyConditions(node *corev1.Node) (nc *corev1.Node
...
 
 func (c *Controller) annotateTerminationGracePeriod(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	if expirationTimeString, exists := nodeClaim.ObjectMeta.Annotations[v1.NodeClaimTerminationTimestampAnnotationKey]; exists {
+		expirationTime, err := time.Parse(time.RFC3339, expirationTimeString)
+		if err == nil {
+			if expirationTime.Before(c.clock.Now()) {
+				return nil
+			}
+		}
+	}
```
njtran, line 174: nit can combine them
```suggestion
		if err == nil && expirationTime.Before(c.clock.Now()) {
			return nil
		}
	}
```

pkg/controllers/node/health/suite_test.go
```diff
@@ -186,6 +187,23 @@ var _ = Describe("Node Health", func() {
 			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 			Expect(nodeClaim.Annotations).To(HaveKeyWithValue(v1.NodeClaimTerminationTimestampAnnotationKey, fakeClock.Now().Format(time.RFC3339)))
 		})
+		It("should not update termination grace period if set past the current time", func() {
```
njtran, line 190: nit on the wording, seems like you're saying "past" indicating how far in the past it is
```suggestion
		It("should not update termination grace period if set before the current time", func() {
```

njtran (commented):
/lgtm
/approve

PR #1898, opened 2025-01-01
Use Expectations from awslabs/operatorpkg and remove duplicate expect…

jeevanpuchakay:
Fixes 1348 

**Description**
- Using expectations from `operatorpkg` so that we can avoid duplication at multiple places.
- Changing only one expectation `ExpectCleanedUp` from `pkg/controllers/disruption/suite_test.go` as part of this PR to align the overall change with reviewers before creating a bigger PR. 

**Issue tracker**
- https://github.com/kubernetes-sigs/karpenter/issues/1348

- Ran tests from VS code and it passed

### DIFF
--- a/pkg/apis/v1/suite_test.go
+++ b/pkg/apis/v1/suite_test.go
@@ -20,12 +20,12 @@ import (
 	"context"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


jonathan-innis (approved):
/lgtm
/approve

PR #1901, opened 2025-01-04
test: Use Expectations from awslabs/operatorpkg

jeevanpuchakay:
Fixes 1348 

**Description**
- Using expectations from `operatorpkg` so that we can avoid duplication at multiple places.
- Changing only those tests where a straightforward replace of imports is sufficient.
- Conversation in progress for other files https://github.com/kubernetes-sigs/karpenter/issues/1348#issuecomment-2571359883

**Issue tracker**
- https://github.com/kubernetes-sigs/karpenter/issues/1348

- Ran tests from VS code and it passed

### DIFF
--- a/pkg/utils/node/suite_test.go
+++ b/pkg/utils/node/suite_test.go
@@ -24,10 +24,10 @@ import (
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 
+	. "github.com/awslabs/operatorpkg/test/expectations"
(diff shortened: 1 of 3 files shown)


jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
Nice! This looks like it could get us closer to removing some of the duplicates of these functions that we have throughout the codebase! I think you just need to run a presubmit so that these CI tests succeed

PR #1907, opened 2025-01-09
[WIP]fix: Fix topology spread constraints with zonal volume

### LINKED ISSUE #1239: Zonal Volume Requirements Break Topology Spread Constraints
## Background

Over the past few months we've received a couple of issues regarding TopologySpreadConstraints with StatefulSets in the aws-provider repository:
- https://github.com/aws/karpenter-provider-aws/issues/5489
- https://github.com/aws/karpenter-provider-aws/issues/6141

In both cases the observed behavior was that Karpenter provisioned nodes as expected initially, but after scaling down and back up Karpenter either provisions nodes that violate* the TopologySpreadConstraint or Karpenter is unable to provision new nodes (allegedly due to TSC violation). 

\* I use violate loosely here, the TSC may be respected from kube-scheduler's perspective but Karpenter does not spread pods across all domains it is aware of.

## Reproduction and Validation
Reproducing this issue is relatively straightforward. The following NodePools and StatefulSet were used:

**NodePools:**
```yaml
kind: NodePool
metadata:
  name: on-demand-np
spec:
  disruption:
    consolidateAfter: 10m
    expireAfter: Never
  template:
    spec:
      nodeClassRef:
        name: default
      requirements:
          operator: In
          values: ['c', 'm', 'r']
          operator: In
          values: ['on-demand']
        # Virtual capacity-spread domain, 2/3 of nodes should launch on on-demand
        - key: capacity-spread
          operator: In
          values: ['1', '2']
---
kind: NodePool
metadata:
  name: spot-np
spec:
  disruption:
    consolidateAfter: 10m
    expireAfter: Never
  template:
    spec:
      nodeClassRef:
        name: default
      requirements:
        operator: In
        values: ['c', 'm', 'r']
        operator: In
        values: ['spot']
      # Virtual capacity-spread domain, 1/3 of nodes should launch on spot
      - key: capacity-spread
        operator: In
        values: ['3']
```

**StatefulSet:**
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: repro
spec:
  replicas: 3
  selector:
    matchLabels:
      app.kubernetes.io/name: repro
  template:
    metadata:
      labels:
        app.kubernetes.io/name: repro
    spec:
      affinity:
        podAntiAffinity:
              labelSelector:
                  matchLabels:
                    app.kubernetes.io/name: repro
      containers:
        - name: inflate
          resources:
            requests:
              cpu: 100m
      topologySpreadConstraints:
        - maxSkew: 1
          topologyKey: capacity-spread
          labelSelector:
            matchLabels:
              app.kubernetes.io/name: "repro"
        - maxSkew: 1
          labelSelector:
            matchLabels:
              app.kubernetes.io/name: "repro"
  volumeClaimTemplates:
  - apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      labels:
        app.kubernetes.io/name: repro
      name: data
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
      volumeMode: Filesystem
```

Each NodePool specifies a different capacity type, spot and on-demand. Additionally, each NodePool specifies a `capacity-type` label with three possible values (1, 2, or 3). The StatefulSet defines a TopologySpreadConstraint on this label. This is meant to ensure that 2/3 of the pods schedule to on-demand nodes and the remaining 1/3 schedule to spot nodes.

Follow these steps to reproduce:
- Apply the NodePools and StatefulSet to your cluster
- Wait for the StatefulSet to finish deploying (all volumes must exist)
- Scale the StatefulSet to 0
- Scale the StatefulSet back up to 3

On the initial scale-up, Karpenter correctly provisions nodes. Two on-demand nodes and one spot node, each in a different AZ (due to another TSC on the StatefulSet). When we scale-down and scale back up, Karpenter provisions three spot nodes, each in a different AZ. The pods are able to schedule, since kube-scheduler is only aware of the single domain (3), but this isn't what we expect Karpenter to do. 

While in that previous scenario the pods were able to schedule, we can create a scenario where Karpenter gets stuck and pods are left pending. Consider the following node distribution:

|Zone|Nodes (capacity-type value)|
|-|-|
|A|spot-3, od-2|
|B|od-1|
|C|spot-3|

We now scale up our StatefulSet to 3 replicas, and pods schedule to the spot node in zone A and the od node in zone B. The remaining pod remains pending while Karpenter continues to nominate it for the spot node in zone C. This would violate the TopologySpreadConstraint with a skew of 2 between domains `2` and `3`.

## Diagnosis

During scheduling, Karpenter not only needs to consider requirements on the pod itself, but also requirements on volumes for the pod. This typically comes in the form of an additional zonal requirement. Internally, Karpenter models this by adding a zonal node affinity term to the pod (in-memory only). Unfortunately, this causes issues for TopologySpreadConstraints. 

When computing skew across domains, only nodes that match a pod's NodeAffinity are counted by default. Since we inject a NodeAffinity term onto the pod after discovering it's volume requirements, this limits the nodes included when computing skew to the nodes in that volume's zone. This lines up with the behavior seen in reproduction. In the first case, where all nodes brought up in the second scale-up were spot, we only counted the single node in each zone for each pod so we believed there was a skew of 0 for each pods TSC. There's a similar story in the second example where we believed there was a skew of 1 for the pod in zone A, 0 in B, and 0 in C. 

The issue laid out in https://github.com/aws/karpenter-provider-aws/issues/6141 is a little bit different in terms of symptoms (unable to provision), but the same zonal injection behavior is to blame. In that issue there were 5 pods in the StatefulSet, each with zonal and hostname anti-affinity. When the StatefulSet was scaled to 3 pods and back to 5, Karpenter failed to provision additional nodes believing TSC would be violated. Since Karpenter is only counting nodes in the same zone as the volume when computing skew, and it is aware of all zones, it believes it must create a node in a different zone to fulfill the TSC requirement. Since this is not possible, we fail to provision.

**Versions**:
- Chart Version: 0.36.1
- Kubernetes Version (`kubectl version`): 1.29

leoryu:
Fixes #1239 

**Description**


leoryu:
@jmdeal @engedaam @tallaxes @jonathan-innis @njtran hi, can you help to review this PR?

codeeong:
Hi, any plans to release this soon? I am experiencing this issue as well 🙏

leoryu:
> Hi, any plans to release this soon? I am experiencing this issue as well 🙏

As of now, no one has reviewed the code. I have forked this repo in my project, but this is not what I expected.

leoryu:
@jmdeal Hi, I have resolved the confilcts of this PR, please help review it. We really want this isssue can be fixed.

ap-elmo:
Just ran into this issue aswell, would be good for a fix! :D

pkg/controllers/provisioning/provisioner.go
```diff
@@ -464,13 +464,13 @@ func validateKarpenterManagedLabelCanExist(p *corev1.Pod) error {
 	return nil
 }
 
-func (p *Provisioner) injectVolumeTopologyRequirements(ctx context.Context, pods []*corev1.Pod) []*corev1.Pod {
-	var schedulablePods []*corev1.Pod
+func (p *Provisioner) convertToPodVolumeRequirements(ctx context.Context, pods []*corev1.Pod) map[*corev1.Pod][]corev1.NodeSelectorRequirement {
+	var schedulablePods = make(map[*corev1.Pod][]corev1.NodeSelectorRequirement)
```
jmdeal, line 468: nit: this project uses this style for map initialization. Also, I think this name is more representative of what we're storing.
```suggestion
podVolumeRequirements := map[*corev1.Pod][]corev1.NodeSelectorRequirement{}
```

pkg/controllers/provisioning/provisioner.go
```diff
@@ -464,13 +464,13 @@ func validateKarpenterManagedLabelCanExist(p *corev1.Pod) error {
 	return nil
 }
 
-func (p *Provisioner) injectVolumeTopologyRequirements(ctx context.Context, pods []*corev1.Pod) []*corev1.Pod {
-	var schedulablePods []*corev1.Pod
+func (p *Provisioner) convertToPodVolumeRequirements(ctx context.Context, pods []*corev1.Pod) map[*corev1.Pod][]corev1.NodeSelectorRequirement {
```
jmdeal, line 467: We're not really converting anything here, right? We're just creating a mapping between pods and their volume requirements. I think something along these lines is more accurate.
```suggestion
func (p *Provisioner) volumeRequirementsForPods(ctx context.Context, pods []*corev1.Pod) map[*corev1.Pod][]corev1.NodeSelectorRequirement {
```

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -60,6 +60,9 @@ type Topology struct {
 	excludedPods sets.Set[string]
 	cluster      *state.Cluster
 	stateNodes   []*state.StateNode
+	// podVolumeRequirements links volume requirements to pods. This is used so we
+	// can track the volume requirements in simulate scheduler
+	podVolumeRequirements map[*corev1.Pod][]corev1.NodeSelectorRequirement
```
jmdeal, line 65: I would use the pod's UID as the key here rather than a pointer to the pod object. We still use the pod object as a key elsewhere in the project, but we've moved to pod UID here (see `excludedPods`) and I think it would be wise for us to move to using it elsewhere when possible since we don't need to worry about copies preventing us from indexing.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -79,17 +85,18 @@ func NewTopology(
 		topologyGroups:        map[uint64]*TopologyGroup{},
 		inverseTopologyGroups: map[uint64]*TopologyGroup{},
 		excludedPods:          sets.New[string](),
+		podVolumeRequirements: podsVolumeRequirements,
 	}
 
 	// these are the pods that we intend to schedule, so if they are currently in the cluster we shouldn't count them for
 	// topology purposes
-	for _, p := range pods {
+	for p := range podsVolumeRequirements {
```
jmdeal, line 93: Why are we iterating over the pods stored as keys in `podVolumeRequirements` rather than all pods? I believe this should be the same set in this implementation, but the intention is to exclude **all** pods we're attempting to schedule, not just those which have an associated volume requirement. Even if it's the same in practice today, this obfuscates intent.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -79,17 +85,18 @@ func NewTopology(
...
 		t.excludedPods.Insert(string(p.UID))
 	}
 
 	errs := t.updateInverseAffinities(ctx)
-	for i := range pods {
-		errs = multierr.Append(errs, t.Update(ctx, pods[i]))
+	for p := range podsVolumeRequirements {
```
jmdeal, line 98: Same comment here - we should still be using `pods` rather than the pods stored as keys in the `podVolumeRequirements`. Let me know if this is intentional and I'm missing the rationale, but as far as I can tell we should be updating the topology for all pods we're attempting to schedule.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -202,9 +206,47 @@ func (t *TopologyGroup) Hash() uint64 {
 // If there are multiple eligible domains, we return any random domain that satisfies the `maxSkew` configuration.
 // If there are no eligible domains, we return a `DoesNotExist` requirement, implying that we could not satisfy the topologySpread requirement.
 // nolint:gocyclo
-func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, nodeDomains *scheduling.Requirement) *scheduling.Requirement {
+func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, nodeDomains *scheduling.Requirement, hasVolumeRequirement bool) *scheduling.Requirement {
```
jmdeal, line 209: This is my high-level understanding of this change, correct me if I'm mistaken:
- When the pod we're attempting to find the next domain for has a volume induced requirement, we want to return the set of possible domains rather than a single domain. 
- We block domains if there are no existing nodes for the domain which are also compatible with the node filter. However, if there are no nodes in a domain (regardless of nodeFilter compatibility), we will add them to the set of compatible domains if the pod has a volume induced requirement.
- If the pod does not have volume induced requirements, we continue to use the single minDomain for the new requirement, the difference being we've no longer evaluated the blocked domains.

I have a few questions about this change:
- I don't understand the purpose of blocked domains, are you able to elaborate?
- Why do we need to return the full set of compatible domains? We're going to constrain the NodeClaim to the pod's volume requirement after anyway, why not just do that first since it's the only eligible domain?

I'll also note that the currently implementation will almost certainly not be acceptable from a performance standpoint - we're iterating over all of the state nodes in the cluster each time we attempt to constrain a pod to adhere to a topology spread constraint. This is one of the most important "hot-paths" in the scheduler, and we need to be extremely careful with what we add here. As far as I can tell this logic shouldn't be necessary if we inject the volume requirements into the nodeclaim requirements before performing the topology checks.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -137,6 +137,13 @@ func (n *NodeClaim) Add(ctx context.Context, pod *corev1.Pod, podData *PodData)
 	}
 	nodeClaimRequirements.Add(topologyRequirements.Values()...)
 
+	podVolumeRequirements := scheduling.NewNodeSelectorRequirements(volumeRequirements...)
+	// Check Pod Volume Requirements
+	if err = nodeClaimRequirements.Compatible(podVolumeRequirements, scheduling.AllowUndefinedWellKnownLabels); err != nil {
+		return err
```
jmdeal, line 143: We should wrap this error, this will be propagated out to the user if we fail to schedule the pod.

pkg/controllers/provisioning/scheduling/topologygroup.go
jmdeal: I will also add that, though this shouldn't cause issues in the common case where we're constraining the NodeClaim to a single zone to adhere to pod requirements, it is an invariant that we need to constrain requirements to a single value for TSC. We only record the domain to the topology group if there's a single domain, and failing to do so can result in Karpenter overprovisioning as it doesn't respect TSC.

jmdeal (commented):
Sorry this had slipped from notice, I didn't notice the GitHub notification and have a relatively deep backlog and just popped this off. I would encourage reaching out in the #karpenter-dev channel on the k8s slack if you're having issues getting traction on a PR, there's far less noise there than there is on GH.

pkg/controllers/provisioning/scheduling/topology.go
leoryu: @jmdeal Since we need to know whether the pod has `VolumeRequirements`, the `pods` is replaced by `podVolumeRequirements`. Please check line 238.

pkg/controllers/provisioning/scheduling/topologygroup.go
leoryu: @jmdeal 
1. The purpose of blocked domains.

I block the empty domains which all existing nodes with them don't match the pod in the caculations of skew and domainMincount, bescaue in real scheduling, these domains is not considered in `TopologySpread`.

2. Why do we need to return the full set of compatible domains?

At present, Karpenter inject volume nodeAffinity info to pod, so many nodes which is not compatible with the `added nodeAffinity` are ignored in karpenter but not in real scheduling.
 
To fix this, I decide to handle pod volume requirements independently, so in this function, we should return all compatible domains to let pod volume choose the suitable domain.

leoryu: @jmdeal As for performance, I have no better idea to fix this issue. But I think accuracy is more important than performance.

jmdeal:
Closing in favor of https://github.com/kubernetes-sigs/karpenter/pull/2780

Issue #1046, opened 2024-02-26
allow karpenter to create nodes with non-well-known kubernetes labels


elmiko:
### Description

**What problem are you trying to solve?**

As an openshift user, i require new nodes to take the label `node-role.kubernetes.io/worker`. Although this label uses the `kubernetes.io` prefix, it is not an [official well-known label](https://kubernetes.io/docs/reference/labels-annotations-taints/). Currently, karpenter does not allow NodePools to be created when a label with the `kubernetes.io` prefix is used on a non-well-known label for new node creation, or when checking predicates.

**How important is this feature to you?**

This is fairly important for openshift as the label is used frequently. It is possible to work around this, but for long term usage on openshift a permanent solution would be better.

This issue was [discussed at the Feb. 26 2024 community meeting](https://docs.google.com/document/d/18BT0AIMugpNpiSPJNlcAL2rv69yAE6Z06gUVj7v_clg/edit#bookmark=id.9809ti3z75rr).

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
Looks like there are a couple other issues that were asking for similar functionality here: https://github.com/aws/karpenter-provider-aws/issues/1941 and https://github.com/kubernetes-sigs/karpenter/issues/660

jonathan-innis:
What's the security implication of allowing users to set this custom label? From a little sleuthing, it doesn't seem to be any different than the `node-restriction.kubernetes.io` label that we already support today.

sftim:
As Karpenter is a Kubernetes project, I'm wary of even indirectly encouraging use of unregistered labels.

If `node-role.kubernetes.io/worker` should become an official label, let's get that registered first - it's not a particularly difficult process. However, registration **first**, then adoption.

“OpenShift uses it” does not count as grounds to add support into Kubernetes code.

--- 

OpenShift folks, if you want a label to mark nodes as OpenShift workers, you can make one, eg `openshift.example/worker: true`

elmiko:
> As Karpenter is a Kubernetes project, I'm wary of even indirectly encouraging use of unregistered labels.

[...]

PR #1908, opened 2025-01-09
fix: Decouple label propagation

enxebre:
Fixes [#N/A ](https://github.com/kubernetes-sigs/karpenter/issues/1046)

**Description**
nodeClaim Labels are the source for core karpenter to sync Node Labels in a centralized fashion.
Some bootstrap userdata provider implementations also consume this labels and pass them through kubelet self setting. 

That results in a coupling between a centralized and a kubelet self setting approach.
This coupling results in conflicting criteria for validations and degraded UX.
This PR removes the coupling.

As a consequence, Node Labels are not unncessarily restricted for the centralized sync anymore.
This better empowers administrators reducing the reliance on self setting kubelet and minimizing the risk of a Node steering privileged workloads to itself.

A subset of labels, excluding those disallowed by the node restriction admission, is now stored in json within the "karpenter.sh/kubelet-labels" NodeClaim annotation. This enables bootstrap userdata providers to continue utilizing the labels if needed.

### DIFF
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -58,19 +58,14 @@ const (
 )
 
 var (
-	// RestrictedLabelDomains are either prohibited by the kubelet or reserved by karpenter
+	// RestrictedLabelDomains are reserved by karpenter
 	RestrictedLabelDomains = sets.New(
-		"kubernetes.io",
-		"k8s.io",
 		apis.Group,
 	)
 
-	// LabelDomainExceptions are sub-domains of the RestrictedLabelDomains but allowed because
-	// they are not used in a context where they may be passed as argument to kubelet.
-	LabelDomainExceptions = sets.New(
-		"kops.k8s.io",
-		v1.LabelNamespaceSuffixNode,
-		v1.LabelNamespaceNodeRestriction,
+	K8sLabelDomains = sets.New(
+		"kubernetes.io",
+		"k8s.io",
 	)
 
 	// WellKnownLabels are labels that belong to the RestrictedLabelDomains but allowed.
@@ -109,32 +104,65 @@ func IsRestrictedLabel(key string) error {
 	if WellKnownLabels.Has(key) {
 		return nil
 	}
-	if IsRestrictedNodeLabel(key) {
+
+	labelDomain := GetLabelDomain(key)
+	for restrictedLabelDomain := range RestrictedLabelDomains {
+		if strings.HasSuffix(labelDomain, restrictedLabelDomain) {
+			return fmt.Errorf("label %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
+		}
+	}
+
+	if RestrictedLabels.Has(key) {
 		return fmt.Errorf("label %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
 	}
+
 	return nil
 }
 
-// IsRestrictedNodeLabel returns true if a node label should not be injected by Karpenter.
-// They are either known labels that will be injected by cloud providers,
-// or label domain managed by other software (e.g., kops.k8s.io managed by kOps).
-func IsRestrictedNodeLabel(key string) bool {
+// IsKubeletLabel returns true if the label key is one that kubelets are allowed to set on their own Node object.
+// This function is similar the one used by the node restriction admission https://github.com/kubernetes/kubernetes/blob/e319c541f144e9bee6160f1dd8671638a9029f4c/staging/src/k8s.io/kubelet/pkg/apis/well_known_labels.go#L67
+// but karpenter also restricts the known labels to be passed to kubelet. Only the kubeletLabelNamespaces are allowed.
+func IsKubeletLabel(key string) bool {
 	if WellKnownLabels.Has(key) {
+		return false
+	}
+
+	if !isKubernetesLabel(key) {
 		return true
 	}
-	labelDomain := GetLabelDomain(key)
-	for exceptionLabelDomain := range LabelDomainExceptions {
-		if strings.HasSuffix(labelDomain, exceptionLabelDomain) {
-			return false
-		}
-	}
-	for restrictedLabelDomain := range RestrictedLabelDomains {
-		if strings.HasSuffix(labelDomain, restrictedLabelDomain) {
+
+	namespace := getLabelNamespace(key)
+	for allowedNamespace := range kubeletLabelNamespaces {
+		if namespace == allowedNamespace || strings.HasSuffix(namespace, "."+allowedNamespace) {
 			return true
 		}
 	}
-	return RestrictedLabels.Has(key)
+
+	return false
 }
+
+func isKubernetesLabel(key string) bool {
+	namespace := getLabelNamespace(key)
+	if namespace == "kubernetes.io" || strings.HasSuffix(namespace, ".kubernetes.io") {
+		return true
+	}
+	if namespace == "k8s.io" || strings.HasSuffix(namespace, ".k8s.io") {
+		return true
+	}
+	return false
+}
+
+func getLabelNamespace(key string) string {
+	if parts := strings.SplitN(key, "/", 2); len(parts) == 2 {
+		return parts[0]
+	}
+	return ""
+}
+
+var kubeletLabelNamespaces = sets.NewString(
+	v1.LabelNamespaceSuffixKubelet,
+	v1.LabelNamespaceSuffixNode,
+)
 
 func GetLabelDomain(key string) string {
 	if parts := strings.SplitN(key, "/", 2); len(parts) == 2 {
--- a/pkg/controllers/nodeclaim/lifecycle/launch.go
+++ b/pkg/controllers/nodeclaim/lifecycle/launch.go
@@ -18,6 +18,7 @@ package lifecycle
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 
@@ -28,6 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"sigs.k8s.io/karpenter/pkg/apis"
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/events"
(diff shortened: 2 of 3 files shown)


enxebre:
If we are happy with the approach I'll sort out the tests.

enxebre:
cc @elmiko @javanthropus

elmiko:
thanks Alberto, i'll take a look tomorrow or monday

sftim (commented):
Do we want to restrict any of these, even as we loosen restrictions:
- `kubernetes.io/hostname`
- `kubernetes.io/arch`
- `kubernetes.io/os`
- `node.kubernetes.io/instance-type`
- `node.kubernetes.io/windows-build`

?

pkg/controllers/nodeclaim/lifecycle/launch.go
```diff
@@ -127,7 +129,14 @@ func PopulateNodeClaimDetails(nodeClaim, retrieved *v1.NodeClaim) *v1.NodeClaim
...
 		nodeClaim.Labels, // User-defined labels
 	)
-	nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, retrieved.Annotations)
+
+	kubeletLabels := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).KubeletLabels()
+	json, err := json.Marshal(kubeletLabels)
+	if err != nil {
+		panic(err)
+	}
+	kubeletLabelsAnnotation := map[string]string{apis.Group + "/kubelet-labels": string(json)}
```
sftim, line 138:
```suggestion
	kubeletLabelsAnnotation := map[string]string{apis.Group + "/node-labels": string(json)}
```
We shouldn't assume that the node runs a kubelet. Any conformant node implementation would be OK.

enxebre: makes sense, I feel node is still conflating things though. How about something like
```
// nodeRestrictedLabelsAnnotation contains a list of labels created from the requirements but filtering out those rejected by the node restriction admission
nodeRestrictedLabelsAnnotation apis:= .Group + /node-restricted-labels
```
?

enxebre:
>Do we want to restrict any of these, even as we loosen restrictions:
kubernetes.io/hostname
kubernetes.io/arch
kubernetes.io/os
node.kubernetes.io/instance-type
node.kubernetes.io/windows-build
?

That sounds reasonable. We can even make this a configurable flag with a this restriction as default if we want to. And with * allowing everything (note that in any case, now that we have decoupling KubeletLabels will filter out everything not admitted). We are most likely moving in a similar direction in capi with the flag. Thoughts?

elmiko (commented):
thanks @enxebre , this is definitely going the direction i wanted to see. i think @sftim's comments about the restricted labels is good too, i think having it configurable is nice perhaps with a sane default that covers the labels Tim mentioned?

enxebre:
Thanks for feedback. I addressed comments and I'll look into sorting out the tests next.

elmiko:
i think this is generally looking good, but i'm curious about @sftim 's comment https://github.com/kubernetes-sigs/karpenter/pull/1908#pullrequestreview-2542197656

i think it makes sense to consider preventing some of those labels, especially if there is overlap with what the ccm might manage.

enxebre:
>i think this is generally looking good, but i'm curious about @sftim 's comment https://github.com/kubernetes-sigs/karpenter/pull/1908#pullrequestreview-2542197656
i think it makes sense to consider preventing some of those labels, especially if there is overlap with what the ccm might manage.

This is now implemented in the PR now via https://github.com/kubernetes-sigs/karpenter/pull/1908/commits/a74f2fdfac879f24cb8276af873b7f8b29a90839#diff-ae13a9bbcddb80fed52372de837e250e0cec57957b6b58bc06c98130d0aed84fR123-R129

pkg/apis/v1/labels.go
```diff
@@ -104,38 +99,72 @@ var (
...
 		}
 	}
-	for restrictedLabelDomain := range RestrictedLabelDomains {
-		if strings.HasSuffix(labelDomain, restrictedLabelDomain) {
+
+	return false
+}
+
+func isKubernetesLabel(key string) bool {
+	for k8Domain := range K8sLabelDomains {
```
javanthropus, line 154: This is just a nit pick, but shouldn't `k8Domain` be called `k8sDomain`?  "K8s" is short for "kubernetes" while "k8" would be short for "kubernete".  Plus using "k8s" matches the name used elsewhere in this code.

pkg/apis/v1/labels.go
```diff
@@ -104,38 +99,72 @@ var (
...
 func IsRestrictedLabel(key string) error {
 	if WellKnownLabels.Has(key) {
 		return nil
 	}
-	if IsRestrictedNodeLabel(key) {
+
+	labelDomain := GetLabelDomain(key)
+	for restrictedLabelDomain := range RestrictedLabelDomains {
+		if labelDomain == restrictedLabelDomain || strings.HasSuffix(labelDomain, "."+restrictedLabelDomain) {
+			return fmt.Errorf("label %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
```
sftim, line 112: Not sure that this is the right advice. Maybe just tell people that Karpenter doesn't manage these labels on Nodes because that's left to other components.

pkg/apis/v1/labels.go
```diff
@@ -104,38 +99,72 @@ var (
...
+		}
+	}
+
+	if RestrictedLabels.Has(key) {
 		return fmt.Errorf("label %s is restricted; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
```
sftim, line 117: See https://github.com/kubernetes-sigs/karpenter/pull/1908/files#r1915062718

pkg/controllers/nodeclaim/lifecycle/launch.go
```diff
@@ -127,7 +129,14 @@ func PopulateNodeClaimDetails(nodeClaim, retrieved *v1.NodeClaim) *v1.NodeClaim
...
 		nodeClaim.Labels, // User-defined labels
 	)
-	nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, retrieved.Annotations)
+
+	kubeletLabels := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).KubeletLabels()
+	json, err := json.Marshal(kubeletLabels)
+	if err != nil {
+		panic(err)
+	}
+	kubeletLabelsAnnotation := map[string]string{apis.Group + "/node-restricted-labels": string(json)}
```
sftim, line 138: Maybe add a comment about why we set this.

pkg/scheduling/requirements.go
```diff
@@ -303,13 +303,29 @@ func (r Requirements) Intersects(requirements Requirements) (errs error) {
...
+			continue
+		}
+		if value := requirement.Any(); value != "" {
+			labels[key] = value
+		}
+	}
+	return labels
+}
+
+// KubeletLabels filters out labels from the requirements that will be rejected by the node restriction admission.
```
sftim, line 320:
```suggestion
// KubeletLabels identifies the labels that we expect the kubelet to manage
// Do not include labels here that the node restriction admission controller would reject.
```
?

elmiko:
> This is now implemented in the PR now via [a74f2fd#diff-ae13a9bbcddb80fed52372de837e250e0cec57957b6b58bc06c98130d0aed84fR123-R129](https://github.com/kubernetes-sigs/karpenter/commit/a74f2fdfac879f24cb8276af873b7f8b29a90839#diff-ae13a9bbcddb80fed52372de837e250e0cec57957b6b58bc06c98130d0aed84fR123-R129)

awesome, thanks!

pkg/apis/v1/labels.go
enxebre: This func is actually validating requirements input and not allowing those within .karpenter.sh (except for the well known CapacityTypeLabelKey). That's afaict a Karpenter design choice.
FWIW It should never return err for the nodepool/nodeclaim apis since this is validated with cel as well. It would return error for a pod with a conflicting node selector req.

So I think the current error message is actually accurate?
`IsValidToSyncCentrallyLabel` is what dictates what gets propagated to Nodes.

pkg/controllers/nodeclaim/lifecycle/launch.go
enxebre: Thanks! Added, please let me know if that's still not clear.

pkg/apis/v1/labels.go
elmiko: the error message makes sense to me given the context, but i do wonder if we should disambiguate from the restricted label error that is produced in the following conditional. looking at the error message on line 117 makes me think we should alter the other message.

pkg/apis/v1/labels.go
elmiko: +1, i think we need to adjust this message to be more accurate.

elmiko (commented):
left a couple comments

pkg/apis/v1/labels.go
enxebre: I updated the error in both places to clarify the conflicting label is not allowed because it might interfere with the internal provisioning logic. Let me know if that's still not clear.

pkg/apis/v1/labels.go
enxebre: let's continue here https://github.com/kubernetes-sigs/karpenter/pull/1908#discussion_r1916121412

elmiko (commented):
i like the new error text. i had thought we might want to differentiate those 2 error conditions, but thinking about it i'm not sure what would be a meaningful distinction.

enxebre:
@sftim @elmiko @javanthropus I think I addressed most comments. Please Let me know if there's more feedback

elmiko:
looking good to me

pkg/controllers/nodeclaim/lifecycle/launch.go
```diff
@@ -124,10 +126,19 @@ func PopulateNodeClaimDetails(nodeClaim, retrieved *v1.NodeClaim) *v1.NodeClaim
...
 		nodeClaim.Labels, // User-defined labels
 	)
-	nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, retrieved.Annotations)
+
+	// We store labels compliant with the node restriction admission as an annotation on the NodeClaim
+	// so that bootstrap provider implementation can use them for kubelet labels.
+	kubeletLabels := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).KubeletLabels()
+	json, err := json.Marshal(kubeletLabels)
+	if err != nil {
+		panic(err)
```
elmiko, line 138: i wonder about this panic. i realize there isn't a way for us to return the error, and it will be some wild error if the marshal fails, but is there anything we can do to help users triage or see this failure?

elmiko (commented):
i'm happy to add lgtm, just one quick question @enxebre

jonathan-innis:
@elmiko Just checking in on this one since we haven't seen a lot of movement on it lately -- do we need a review from an approver?

elmiko:
hey @jonathan-innis , yeah i think getting a review from an approver would be good.

PR #1910, opened 2025-01-10
chore: RFC for NodeRegistrationHealthy Status Condition on NodePool

jigisha620:
**Description**

Adding RFC for NodeRegistrationHealthy Status Condition on NodePool


designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
...
+
+**Note:** The issues that we want to address in the launch path with this status condition are the ones that happen during Nodeclaim launch and registration. Nodeclaim initialization never fails as of today, it can only transition from Unknown to True unless we plan on changing that in the future.   
+
+## 🔑 Introduce a Degraded Status Condition on the NodePool
+
+```
+// ConditionTypeDegraded = "Degraded" condition indicates that a misconfiguration exists that prevents the normal, successful use of a Karpenter resource
+ConditionTypeDegraded = "Degraded"
+```
+We want to introduce a `Degraded` status condition on the Nodepool. This condition can be set to either of Unknown, True or False. The state transition is not unidirectional meaning it can go from True to False and back to True or Unknown. `Degraded` status condition will not be a pre-condition for Nodepool readiness. A degraded Nodepool will continue to be considered for provisioning. We will only deprioritize it during provisioning among the nodepools with similar weight.
```
jmdeal, line 19: I think there are a few other points we should discuss in this section before discussing the different implementation options:
- What does each of the states indicate (true, false, and unknown)? I think it differs slightly in the two provided options. For example, in option 1 the condition should be unknown if we don't have sufficient data, whereas in option 2 it transitions to false. To make the doc easier to follow, we should establish the meaning of these states up front and use them consistently. If we think there's multiple options worth evaluating for those meanings, I think we should discuss that in a separate section from the implementation details.
- We should establish the properties that both solutions need to have:
    - Tolerates transient errors
    - Responds to corrections in external configuration (i.e. can remove the degraded status condition from a NodePool if an external fix allows Nodes to register)
    - Resets to unknown if there is an update to **either** the NodePool or linked NodeClass

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
+# RFC: Degraded NodePool Status Condition
+
+## Overview
+
+Karpenter can launch nodes with a NodePool that will never join a cluster when a NodeClass is misconfigured.
+
+One example is that when a network path does not exist due to a misconfigured VPC (network access control lists, subnets, route tables), Karpenter will not be able to provision compute with that NodeClass that joins the cluster until the error is fixed. Crucially, this will continue to charge users for compute that can never be used in a cluster.
+
+To improve visibility of these failure modes, this RFC proposes adding a `Degraded` status condition on the Nodepool that indicate to cluster users there may be a problem with a NodePool/NodeClass combination that needs to be investigated and corrected.
```
jmdeal, line 9: I think it would also be worth giving a counter example for the types of errors we're not trying to capture with the deprecated condition:
> The degraded status condition should not be used as a catch-all for any launch / registration failures. If Karpenter can guarantee that a launch / registration will not succeed based on the information provided in the NodePool / NodeClass configs, it shouldn't consider those resources ready. For example, if a NodePool is constrained to a zone that is not available via the provided subnets, Karpenter should treat that NodePool as not ready rather than degraded.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Overview
+
+Karpenter can launch nodes with a NodePool that will never join a cluster when a NodeClass is misconfigured.
+
+One example is that when a network path does not exist due to a misconfigured VPC (network access control lists, subnets, route tables), Karpenter will not be able to provision compute with that NodeClass that joins the cluster until the error is fixed. Crucially, this will continue to charge users for compute that can never be used in a cluster.
+
+To improve visibility of these failure modes, this RFC proposes adding a `Degraded` status condition on the Nodepool that indicate to cluster users there may be a problem with a NodePool/NodeClass combination that needs to be investigated and corrected. 
+
+**Note:** The issues that we want to address in the launch path with this status condition are the ones that happen during Nodeclaim launch and registration. Nodeclaim initialization never fails as of today, it can only transition from Unknown to True unless we plan on changing that in the future.
```
jmdeal, line 11: A bit of a nit on wording: I think calling this the "launch path" is confusing. That reads to me as anything leading up to the instance being launched, whereas we're really trying to capture registration failures (which may have been caused by unknown launch failures). I'd word this something like this:

> The presence of the degraded status condition will indicate failures with instance launch / registration that aren't well known to Karpenter. Since initialization does not have a timeout like launch and registration, there is no concept of an initialization failure today. However, this design could be extended to capture initialization failures in the future.
 
Also, we should probably merge this with the paragraph explaining the types of failures we're trying to capture rather than having it as a separate note.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
+# RFC: Degraded NodePool Status Condition
+
+## Overview
+
+Karpenter can launch nodes with a NodePool that will never join a cluster when a NodeClass is misconfigured.
```
jmdeal, line 5: I don't think I'd frame this as a NodeClass misconfiguration. While that's a possible failure mode, it doesn't capture everything we're looking to address. I'd frame it as "registration failures which can't be well understood or predicted by Karpenter", and then give the specific examples.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
...
+[+1] = 'Degraded: False'
+[-1, +1] = 'Degraded: False'
+[-1, -1] = 'Degraded: True'
+[-1, +1, -1] = 'Degraded: True'
+[-1, +1, +1, +1, +1, +1, +1, +1, +1, +1] = 'Degraded: False'
+```
+
+#### Considerations
+
+1. 👎 Heuristics can be wrong and mask failures
```
jmdeal, line 52: Could you elaborate on what type of failures are being masked? As for it being wrong, I'm wondering if we should only consider degraded unknown or true. Maybe we don't ever transition it to false?

jmdeal (commented):
Checkpointing

designs/degraded-nodepools.md
jonathan-innis, line 9: I think like Jason has called out online and offline, I would make our motivation front and center here. Why do we think that there is a need for something like this to exist? Does it make tracking down failures to NodePools easier? Does alarming get easier with this kind of a setup?

designs/degraded-nodepools.md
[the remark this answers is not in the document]
jonathan-innis: +1, I agree that we should talk about what it means for the Degraded status condition to be in a given state

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
+# RFC: Degraded NodePool Status Condition
+
+## Overview
+
+Karpenter can launch nodes with a NodePool that will never join a cluster when a NodeClass is misconfigured.
+
+One example is that when a network path does not exist due to a misconfigured VPC (network access control lists, subnets, route tables), Karpenter will not be able to provision compute with that NodeClass that joins the cluster until the error is fixed. Crucially, this will continue to charge users for compute that can never be used in a cluster.
```
jonathan-innis, line 7: I think listing out tangible, real faillure types that this will catch will help illuminate how we design this.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
...
+We want to introduce a `Degraded` status condition on the Nodepool. This condition can be set to either of Unknown, True or False. The state transition is not unidirectional meaning it can go from True to False and back to True or Unknown. `Degraded` status condition will not be a pre-condition for Nodepool readiness. A degraded Nodepool will continue to be considered for provisioning. We will only deprioritize it during provisioning among the nodepools with similar weight.
+
+### Option 1: In-memory Buffer to store history - Recommended
+
+This option will have an in-memory FIFO buffer, which will grow to a max size of 10 (this can be changed later). This buffer will store data about the success or failure in the launch path and is evaluated by a controller to determine the relative health of the NodePool. This will be an int buffer and a positive means `Degraded: False`, negative means `Degraded: True` and 0 means `Degraded: Unknown`. 
+
+Evaluation conditions -
+
+1. We start with an empty buffer with `Degraded: Unknown`.
+2. There have to be 2 minimum failures in the buffer for `Degraded` to transition to `True` or basically the threshold would be 80%.
```
jonathan-innis, line 28: One thing that I still feel like would be better here is if we considered flipping the polarity of this condition type at all -- Degraded: False meaning that it's healthy feels a bit weird to me but I get that we have to come up with some other word besides "Degraded" that isn't "Ready" and probably isn't "Healthy" to really reflect what this condition is evaluating

jonathan-innis, line 28: I feel like you might be able to just omit the bit about 80% -- you could just say, if there are two failures in the last 10 attempts, then we mark it into a `Degraded: True` state

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
...
+
+See below for example evaluations:
+
+```
+Successful Launch: 1
+Default: 0
+Unsuccessful Launch: -1
+
+[] = 'Degraded: Unknown'
+[-1] = 'Degraded: Unknown'
```
jonathan-innis, line 42: nit: It's _slightly_ confusing to call this "Degraded: Unknown". The only reason that I say that is becasue this doesn't necessarily mean that we transition the condition to Unknown when the condition is already set -- I know this is said above but I did find it a tad semantically odd as I was reading through this and trying to parse-out the design

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,119 @@
...
+[-1, +1, -1] = 'Degraded: True'
+[-1, +1, +1, +1, +1, +1, +1, +1, +1, +1] = 'Degraded: False'
+```
+
+#### Considerations
+
+1. 👎 Heuristics can be wrong and mask failures
+2. 👍 Keeps track of recent launch history.
+3. 👍 Can be easily expanded if we want to update the buffer size depending on the cluster size.
+4. 👍 Observability improvements so that users can begin triaging misconfigurations
```
jonathan-innis, line 55: This is the design goal in general, right? So I don't know if I'd call this a pro as a consideration for this particular implementation

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
...
+## 🔑 Introduce a Degraded Status Condition on the NodePool Status
+
+```
+// 'Degraded' condition indicates that a misconfiguration exists that prevents the normal, successful use of a Karpenter resource
+Status:
+  Conditions:
+    Last Transition Time:  2025-01-13T18:57:20Z
+    Message:               
+    Observed Generation:   1
+    Reason:                Degraded
```
saurav-agarwalla, line 22: One thing that I had discussed with Reed is making this a more structured object and putting a serialized string output of that here since I understand that this has to be a string. That way, we can expose more details including error codes mentioning the reason behind the degradation, expose resource IDs/dependents causing it as well as have more than one reason for the degradation. Making it a structured object will also allow us to parse it better for metrics.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
...
+
+This RFC proposes enhancing the visibility of these failure modes by introducing a `Degraded` status condition on the NodePool. We can then create new metric/metric-labels around this status condition which will improve the observability by alerting cluster administrators to potential issues within a NodePool that require investigation and resolution.
+
+The `Degraded` status would specifically highlight instance launch/registration failures that Karpenter cannot fully diagnose or predict. However, this status should not be a mechanism to catch all types of launch/registration failures. Karpenter should not mark resources as `Degraded` if it can definitively determine, based on the NodePool/NodeClass configurations or through dry-run, that launch or registration will fail. For instance, if a NodePool is restricted to a specific zone using the `topology.kubernetes.io/zone` label, but the specified zone is not accessible through the provided subnet configurations, this inconsistency shouldn't trigger a `Degraded` status.
```
enxebre, line 9: >For instance, if a NodePool is restricted to a specific zone using the `topology.kubernetes.io/zone` label, but the specified zone is not accessible through the provided subnet configurations, this inconsistency shouldn't trigger a `Degraded` status.

Can we enumerate different semantics for failures that we'd want to capture as different .Reasons that should trigger degraded == true, e.g. badSecurityGroup

designs/degraded-nodepools.md
jmdeal: Is `reason` the right field to surface that level of detail? I agree with the direction, but it seems like `message` would be more appropriate.

designs/degraded-nodepools.md
jonathan-innis: Major +1 to this -- I think what we need to explore here is how we are going to capture these failure modes -- if we are just relying on the registration timeout being hit, it's going to be tough to know what the reason was that the Node failed to join

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
+# RFC: Degraded NodePool Status Condition
+
+## Motivation
+
+Karpenter may initiate the creation of nodes based on a NodePool configuration, but these nodes might fail to join the cluster due to unforeseen registration issues that Karpenter cannot anticipate or prevent. An example illustrating this issue is when network connectivity is impeded by incorrect cluster security group configuration, such as missing outbound rule that allows outbound access to any IPv4 address. In such cases, Karpenter will continue its attempts to provision compute resources, but these resources will fail to join the cluster until the outbound rule for the security group is updated. The critical concern here is that users will incur charges for these compute resources despite their inability to be utilized within the cluster.
```
jonathan-innis, line 5: To be clear: I don't think that we're really solving for the problem of cost here -- we're still going to be launching instances and retrying

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
...
+
+The state transition is not unidirectional meaning it can go from True to False and back to True or Unknown. A NodePool marked as Degraded can still be used for provisioning workloads, as this status isn't a precondition for readiness. However, when multiple NodePools have the same weight, a degraded NodePool will receive lower priority during the provisioning process compared to non-degraded ones. 
+
+The approach that we go forward with should -
+1. Tolerate transient errors.
+2. Respond to corrections in external configuration (i.e. can remove the degraded status condition from a NodePool if an external fix allows Nodes to register).
+
+### Option 1: In-memory Buffer to store history - Recommended
+
+This option will have an in-memory FIFO buffer, which will grow to a max size of 10 (this can be changed later). This buffer will store data about the success or failure during launch/registration and is evaluated by a controller to determine the relative health of the NodePool. This will be an int buffer and a positive means `Degraded: False`, negative means `Degraded: True` and 0 means `Degraded: Unknown`.
```
jmdeal, line 39: If I understand correctly, the final sentence states that the buffer can have three values: `-1` (degraded true), `0` (unknown), `1` (degraded false). I don't think this matches the example, which only has two values, and the values map to launch success / failure, not the actual degraded state right? I think it might be more clear like this:
```suggestion
This option will have an in-memory FIFO buffer, which will grow to a max size of 10 (this can be changed later). This buffer will store data about the success or failure during launch/registration and is evaluated by a controller to determine the relative health of the NodePool. This would be implemented as a `[]bool`, where `true` indicates a launch success, and `false` represents a failure. The state of the degraded condition would be based on the number of `false` entries in the buffer.
```

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
...
+
+[] = 'Degraded: Unknown'
+[+1] = 'Degraded: False'
+[-1, +1] = 'Degraded: False'
+[-1, -1] = 'Degraded: True'
+[-1, +1, -1] = 'Degraded: True'
+[-1, +1, +1, +1, +1, +1, +1, +1, +1, +1] = 'Degraded: False'
+```
+
+#### Considerations
```
jonathan-innis, line 67: We discussed this but what happens if we have one success and that causes us to stop trying new NodeClaims -- when that happens, what's the way that we make sure that we eventually get out of a Degraded state

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
...
+1. If 3 or more NodeClaims fail to launch with a NodePool then the NodePool will be marked as degraded. The retries are included to account for transient errors.
+2. Once a NodePool is `Degraded`, it recovers with `Degraded: False` after an update to the NodePool. 
+3. A successful provisioning would also set `Degraded: False`.
+
+#### Considerations
+
+1. 👎 Three retries can still be a long time to wait on compute that never provisions correctly.
+2. 👎 Setting `Degraded: False` on an update to NodePool implies Karpenter can vet with certainty that NodePool is correctly configured which is misleading.
+
+### How Does this Affect Metrics and Improve Observability?
```
jonathan-innis, line 89: I'm still a bit fuzzy on the reason that we need this extra label

designs/degraded-nodepools.md
jigisha620: I haven't updated the RFC since our discussion about that. But @rschalo and I were thinking about expiring the entries in the buffer after some time - maybe 3x the registration ttl.

rschalo: I think one of the options we had discussed was expiring entries some amount of time after the last write so that there is some recency bias.

jigisha620: Expiring the entries like this should also see when was the last update made to the buffer if it is frequent enough (we can define the time), then we don't expire entries until the buffer is full.

designs/degraded-nodepools.md
jigisha620: Agreed. Will update.

designs/degraded-nodepools.md
jigisha620: Addressed. We will surface the recent cause of the launch/registration failure through status condition reason and message.

designs/degraded-nodepools.md
```diff
@@ -0,0 +1,107 @@
...
+    Reason:                Degraded
+    Status:                True
+    Type:                  Degraded
+```
+`Degraded` status condition is introduced in the NodePool status which can be set to - 
+1. Unknown - When the NodePool is first created, `Degraded` is set to Unknown. This means that we don't have enough data to tell if the NodePool is degraded or not. 
+2. True - NodePool has configuration issues that require customer investigation and resolution. Since Karpenter cannot automatically detect these specific launch or registration failures, we will document common failure scenarios and possible fixes in our troubleshooting guide to assist customers.
+3. False - There has been successful node registration using this NodePool.
+
+The state transition is not unidirectional meaning it can go from True to False and back to True or Unknown. A NodePool marked as Degraded can still be used for provisioning workloads, as this status isn't a precondition for readiness. However, when multiple NodePools have the same weight, a degraded NodePool will receive lower priority during the provisioning process compared to non-degraded ones.
```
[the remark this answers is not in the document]
jigisha620: Updated the RFC, we will no longer deprioritize the nodePool during scheduling.

designs/noderegistrationhealthy-status-condition.md
```diff
@@ -0,0 +1,95 @@
...
+
+1. 👍 This approach is particularly helpful for pod binding/ready metrics for pods that are scheduled against NodePools that have `NodeRegistrationHealthy: True`. Metrics collection will begin only after its first successful node registration. Once `NodeRegistrationHealthy: True`, and a node is launched with a bad AMI then we will still see pod metrics that can help identify node launch failures due to bad AMIs.
+
+### Option 2: In-memory Buffer to store history
+
+This option will have an in-memory FIFO buffer, which will grow to a max size of 10 (this can be changed later). This buffer will store data about the success or failure during launch/registration and is evaluated by a controller to determine the relative health of the NodePool. This would be implemented as a `[]bool`, where `true` indicates a launch success, and `false` represents a failure. The state of the `NodeRegistrationHealthy` condition would be based on the number of `false` entries in the buffer.
+
+![](./images/noderegistrationhealthy-nodepools1.png)
+
+Evaluation conditions -
```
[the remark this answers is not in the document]
jigisha620: Updated.

designs/noderegistrationhealthy-status-condition.md
```diff
@@ -0,0 +1,95 @@
...
+
+This RFC proposes enhancing the visibility of these failure modes by introducing a `NodeRegistrationHealthy` status condition on the NodePool. We can then create new metrics around this status condition which will improve observability by alerting cluster administrators to potential issues within a NodePool that require investigation and resolution.
+
+The `NodeRegistrationHealthy` status would specifically highlight instance launch/registration failures that Karpenter cannot fully diagnose or predict. However, this status should not be a mechanism to catch all types of launch/registration failures. Karpenter should not mark resources as `NodeRegistrationHealthy` if it can definitively determine, based on the NodePool/NodeClass configurations or through dry-run, that launch or registration will fail. For instance, if a NodePool is restricted to a specific zone using the `topology.kubernetes.io/zone` label, but the specified zone is not accessible through the provided subnet configurations, this inconsistency shouldn't trigger a `NodeRegistrationHealthy: False` status.
```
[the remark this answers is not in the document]
jigisha620: For this example, we actually ignore such nodePool during scheduling as we will not be able to resolve any instance types for this nodePool.

jonathan-innis (approved):
/lgtm
/approve

PR #1760, opened 2024-10-21
RFC: Capacity Reservations

jonathan-innis:
Fixes #N/A 

**Description**

This Capacity Reservation RFC proposes upstream changes for proposals initially made in the [`aws/karpenter-provider-aws` ODCR RFC](https://github.com/aws/karpenter-provider-aws/blob/main/designs/odcr.md). It describes changes to the scheduler, the CloudProvider interface, and consolidation that need to be made to ensure that capacity reservations work properly across all clouds.

It also proposes the introduction of the `CapacityReservations` feature flag in `ALPHA`. 

**How was this change tested?**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1911, opened 2025-01-13
feat: reserved capacity support

jmdeal:
Fixes #N/A 

**Description**
Implements the changes described in https://github.com/kubernetes-sigs/karpenter/pull/1760.

- `make test`
- integration tests in the AWS provider (WIP)


pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -103,26 +167,94 @@ func (n *NodeClaim) Add(pod *v1.Pod, podData *PodData) error {
 		return err
 	}
 
+	reservedOfferings, offeringsToRelease, err := n.reserveOfferings(remaining, nodeClaimRequirements)
```
jonathan-innis, line 170: Can you just diff the current `n.reservedOfferings` with the offerings that are returned from this function? Maybe there's a way to store this efficiently so you aren't having to do a find across all offerings to do this but that strikes me as simpler than having to build-up and handle partitioning

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -261,15 +293,17 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
+	UnfinishedWorkSeconds.Delete(map[string]string{ControllerLabel: injection.GetControllerName(ctx), schedulingIDLabel: string(s.uuid)})
 	for _, m := range s.newNodeClaims {
 		m.FinalizeScheduling()
 	}
 
 	return Results{
 		NewNodeClaims: s.newNodeClaims,
 		ExistingNodes: s.existingNodes,
-		PodErrors:     errors,
+		PodErrors: lo.OmitBy(errors, func(_ *corev1.Pod, err error) bool {
```
jonathan-innis, line 304: What do you think -- should we omit these higher up when we actually print out the errors? Realistically, it does make sense to me that this is included in these pod errors (from what gets returned in the results section) because this is the reason that the pod didn't schedule -- I think the real bit is that we don't want to print it back

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -309,25 +343,21 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
...
+			nodeClaimTemplate,
+			s.topology,
+			s.daemonOverhead[nodeClaimTemplate],
+			s.remainingResources[nodeClaimTemplate.NodePoolName],
+			pod,
+			s.cachedPodData[pod.UID],
+			s.reservedOfferingMode,
+		)
+		if err != nil {
+			if nodeClaim != nil {
```
jonathan-innis, line 357: This strikes me as a weird semantic and I think also leaks hostname values in our topology code just glancing at it -- because we return nil in some scenarios where we have already registered the topology

pkg/operator/options/options.go
```diff
@@ -98,7 +99,7 @@ func (o *Options) AddFlags(fs *FlagSet) {
...
 	fs.DurationVar(&o.BatchIdleDuration, "batch-idle-duration", env.WithDefaultDuration("BATCH_IDLE_DURATION", time.Second), "The maximum amount of time with no new pending pods that if exceeded ends the current batching window. If pods arrive faster than this time, the batching window will be extended up to the maxDuration. If they arrive slower, the pods will be batched separately.")
-	fs.StringVar(&o.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "NodeRepair=false,SpotToSpotConsolidation=false"), "Optional features can be enabled / disabled using feature gates. Current options are: SpotToSpotConsolidation")
+	fs.StringVar(&o.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "ReservedCapacity=false,NodeRepair=false,SpotToSpotConsolidation=false"), "Optional features can be enabled / disabled using feature gates. Current options are: ReservedCapacity, NodeRepair, and SpotToSpotConsolidation")
```
jonathan-innis, line 102: If we're updating the feature flag naming, let's update the design doc so that it reflects the same -- as it stands right now, that design doc says `CapacityReservations` is the feature flag

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -40,6 +41,10 @@ import (
 	"sigs.k8s.io/karpenter/pkg/utils/resources"
 )
 
+func init() {
+	v1.WellKnownLabels = v1.WellKnownLabels.Insert(v1alpha1.LabelReservationID)
```
jonathan-innis, line 45: If you have the reservation ID passed through to identify the offering, I'm thinking you may not need this label anymore to identify the offering

pkg/cloudprovider/types.go
```diff
@@ -239,27 +262,74 @@ func (i InstanceTypeOverhead) Total() corev1.ResourceList {
...
+}
+
 // Available filters the available offerings from the returned offerings
 func (ofs Offerings) Available() Offerings {
 	return lo.Filter(ofs, func(o Offering, _ int) bool {
 		return o.Available
 	})
 }
 
+func (ofs Offerings) PartitionCompatible(reqs scheduling.Requirements) (compatible Offerings, incompatible Offerings) {
```
jonathan-innis, line 322: Same kind of question here about creating two new offerings arrays, I think this is going to end up being pretty inefficient in the scheduler at the end of the day

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -162,10 +162,20 @@ func benchmarkScheduler(b *testing.B, instanceCount, podCount int) {
 		b.Fatalf("creating topology, %s", err)
 	}
 
-	scheduler := scheduling.NewScheduler(ctx, client, []*v1.NodePool{nodePool},
```
jonathan-innis, line 165: Can we consider either adding a benchmark test in this PR or in a follow-on PR that validates what the performance of scheduling is with the addition of the reservation code? We should: 1) Compare our current scheduling performance and 2) Ensure that we add a test case for using reserved capacity type so that we can ensure that we are tracking when the performance improves over time

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3778,6 +3778,98 @@ var _ = Context("Scheduling", func() {
...
+					Requirements: pscheduling.NewLabelRequirements(map[string]string{
+						v1.CapacityTypeLabelKey:     v1.CapacityTypeReserved,
+						corev1.LabelTopologyZone:    "test-zone-1",
+						v1alpha1.LabelReservationID: fmt.Sprintf("r-%s", it.Name),
+					}),
+					Price: fake.PriceFromResources(it.Capacity) / 10000.0,
+				})
+			}
+		})
+		FIt("shouldn't fallback to on-demand or spot when compatible reserved offerings are available", func() {
```
jonathan-innis, line 3820:
```suggestion
		It("shouldn't fallback to on-demand or spot when compatible reserved offerings are available", func() {
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3778,6 +3778,98 @@ var _ = Context("Scheduling", func() {
...
+				return bindings.Get(p) == nil
+			})
+
+			bindings = ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+			Expect(len(bindings)).To(Equal(1))
+			node = lo.Values(bindings)[0].Node
+			Expect(node.Labels).To(HaveKeyWithValue(v1.CapacityTypeLabelKey, Not(Equal(v1.CapacityTypeReserved))))
+			Expect(node.Labels).To(HaveKeyWithValue(corev1.LabelInstanceTypeStable, "small-instance-type"))
+		})
+		// It("should fallback to on-demand or spot when no compatible reserved offerings are available", func() {
```
jonathan-innis, line 3869: Is this next test meant to capture a case where there is a second pod that would cause us to filter out reserved capacity options so we don't schedule it but then we get a third pod that schedules to a completely different NodePool?

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -2050,29 +2050,29 @@ var _ = Describe("Consolidation", func() {
 				Name: "current-on-demand",
```
jonathan-innis, line 2050: Seems like we're missing consolidation and drift testing for reserved capacity

pkg/cloudprovider/types.go
```diff
@@ -226,6 +240,15 @@ func (its InstanceTypes) Truncate(requirements scheduling.Requirements, maxItems
 	return truncatedInstanceTypes, nil
 }
 
+func (its InstanceTypes) Difference(other InstanceTypes) InstanceTypes {
```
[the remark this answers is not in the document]
jmdeal: Looks like it was vestigial from a previous iteration, dropped it.

pkg/cloudprovider/types.go
```diff
@@ -239,27 +262,74 @@ func (i InstanceTypeOverhead) Total() corev1.ResourceList {
...
-	// Available is added so that Offerings can return all offerings that have ever existed for an instance type,
-	// so we can get historical pricing data for calculating savings in consolidation
-	Available bool
+	Available    bool
 }
 
 type Offerings []Offering
 
+// WithCapacityType filters the offerings by the provided capacity type.
+func (ofs Offerings) WithCapacityType(capacityType string) Offerings {
```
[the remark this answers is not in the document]
jmdeal: I was leaning into the use of set operations for handling reservations, but you're right that the extra readability probably isn't worth the additional memory allocations. I've refactored to drop the additional set operations I've introduced. Granted, in the common case this would have had little to no impact since we quickly short-circuit if there are no reserved offerings, and I generally expect the set of reserved offerings to be far smaller than the total offerings.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -35,36 +42,93 @@ import (
...
+	nodeClaimTemplate *NodeClaimTemplate,
+	topology *Topology,
+	daemonResources corev1.ResourceList,
+	remainingResources corev1.ResourceList,
+	pod *corev1.Pod,
+	podData *PodData,
+	reservedOfferingMode ReservedOfferingMode,
+) (*NodeClaim, error) {
+	// Ensure we don't consider instance types which would exceed the limits of the NodePool
+	instanceTypes := filterByRemainingResources(nodeClaimTemplate.InstanceTypeOptions, remainingResources)
```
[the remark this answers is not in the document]
jmdeal: I'll have to check again, it was definitely necessary in a previous iteration due to the way I had handled the relaxation logic but I think you're right that it's not anymore. I do think it's an improvement FWIW so if it's not necessary I'll defer to what you think for keeping it.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -103,26 +167,94 @@ func (n *NodeClaim) Add(pod *v1.Pod, podData *PodData) error {
...
+// not be released until we're ready to persist the changes to the nodeclaim.
+// nolint:gocyclo
+func (n *NodeClaim) reserveOfferings(
+	instanceTypes []*cloudprovider.InstanceType,
+	nodeClaimRequirements scheduling.Requirements,
+) (reservedOfferings map[string]cloudprovider.Offerings, offeringsToRelease []cloudprovider.Offering, err error) {
+	compatibleReservedInstanceTypes := sets.New[string]()
+	reservedOfferings = map[string]cloudprovider.Offerings{}
+	for _, it := range instanceTypes {
+		offerings := it.Offerings.Available().WithCapacityType(v1.CapacityTypeReserved)
```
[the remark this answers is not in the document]
jmdeal: Yeah, I've refactored to reduce this significantly. I don't think this would have had significant performance impact since the total number of reserved offerings are going to be far fewer than the total number of offerings. This call in particular should return `nil` in the common case. That being said, I think you're right that it's good practice to avoid these functions this deep into our scheduling code when possible.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -103,26 +167,94 @@ func (n *NodeClaim) Add(pod *v1.Pod, podData *PodData) error {
...
+		}
+	}
+	if n.reservedOfferingMode == ReservedOfferingModeStrict {
+		// If an instance type with a compatible reserved offering exists, but we failed to make any reservations, we should
+		// fail. We include this check due to the pessimistic nature of instance reservation - we have to reserve each potential
+		// offering for every nodeclaim, since we don't know what we'll launch with. Without this check we would fall back to
+		// on-demand / spot even when there's sufficient reserved capacity. This check means we may fail to schedule the pod
+		// during this scheduling simulation, but with the possibility of success on subsequent simulations.
+		// Note: while this can occur both on initial creation and on
+		if len(compatibleReservedInstanceTypes) != 0 && len(reservedOfferings) == 0 {
```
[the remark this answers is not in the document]
jmdeal: Yeah, so this is basically defending against the case where there is an offering available, but that offering has already been reserved by other NodeClaims. If we were concretely determining the offering we'd launch with at this point in time, we wouldn't need to handle this. However, since we pessimistically reserve all compatible offerings we want to fall back and try again on the next iteration.

Consider a case where we have two pods which much schedule on their own nodes (e.g. anti-affinity). We have two reserved offerings, each with a capacity of one, and both pods are compatible with both offerings. We create the first nodeclaim for the first pod, and we reserve one of each offering. When we go to create the second nodeclaim, we fail to reserve either offering (since all capacity was reserved by the first nodeclaim), but we don't want to fallback to on-demand. Instead, we fail to schedule this pod in this scheduling simulation and we'll retry in the next simulation.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -103,26 +167,94 @@ func (n *NodeClaim) Add(pod *v1.Pod, podData *PodData) error {
...
+			return nil, nil, NewReservedOfferingError(fmt.Errorf("one or more instance types with compatible reserved offerings are available, but could not be reserved"))
+		}
+		// If the nodeclaim previously had compatible reserved offerings, but the additional requirements filtered those out,
+		// we should fail to add the pod to this nodeclaim.
```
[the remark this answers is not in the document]
jmdeal: I'm not sure what you mean by optimistic in this case. This check ensures that if a NodeClaim was compatible with a reserved offering, we won't tighten its constraints such that it is no longer compatible.

pkg/cloudprovider/fake/cloudprovider.go
jmdeal: You still need it for compatibility checks (e.g. a pod with a node selector on a specific reservation)

pkg/controllers/provisioning/provisioner.go
```diff
@@ -261,7 +269,7 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*corev1.Pod, stat
 	if err != nil {
 		return nil, fmt.Errorf("getting daemon pods, %w", err)
 	}
-	return scheduler.NewScheduler(ctx, p.kubeClient, nodePools, p.cluster, stateNodes, topology, instanceTypes, daemonSetPods, p.recorder, p.clock), nil
+	return scheduler.NewScheduler(ctx, schedulerID, p.kubeClient, nodePools, p.cluster, stateNodes, topology, instanceTypes, daemonSetPods, p.recorder, p.clock, reservedOfferingMode), nil
```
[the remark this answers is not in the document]
jmdeal: Sure, are we sure fallback should be the default though? IMO it would be kind of surprising that the provisioner's behavior isn't the default, but it doesn't functionally matter either way.

pkg/cloudprovider/types.go
```diff
@@ -239,27 +262,74 @@ func (i InstanceTypeOverhead) Total() corev1.ResourceList {
 	return resources.Merge(i.KubeReserved, i.SystemReserved, i.EvictionThreshold)
 }
 
+// ReservationManager is used to track the availability of a reserved offering over the course of a scheduling
+// simulation. Reserved offerings may have a limited number of available instances associated with them,
+// This is exposed as an interface for cloudprovider's to implement to give flexibility when dealing with separate
+// offerings with associated availablility.
+type ReservationManager interface {
```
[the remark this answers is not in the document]
jmdeal: Noting this here since I've flaked since our discussion yesterday, but I do think we should go with the name `ReservationID`, at least until there's a non-reservation use case. Since there's no way to encode in the type system that "this field should only be set if capacity type == reserved", it seems reasonable to encode it in the name. I think we can change it to `OfferingID` down the road if / when there's a more general use case.

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal: Yeah, you're definitely right about the issue with `Destroy`. The refactor I did here shouldn't be necessary anymore with the current way reservations are tracked, so I'm going to revert this set of changes.

pkg/controllers/disruption/consolidation_test.go
jmdeal: Yeah, I'm planning on expanding the set of disruption and scheduling tests tonight. I definitely plan on including some.

pkg/controllers/disruption/helpers.go
```diff
@@ -100,6 +112,7 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 				// If the pod is on a deleting node, we assume one of two things has already happened:
 				// 1. The node was manually terminated, at which the provisioning controller has scheduled or is scheduling a node
 				//    for the pod.
+				// TODO: clarify this point, not clear to me
```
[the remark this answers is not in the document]
jmdeal: Don't remember, rereading it it makes sense to me today 🤷

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal: I don't have a strong opinion either way, you're right that at the moment the main goal is just preventing us from printing those errors. We can move this filter further up the call stack. I think I originally handled it here so we wouldn't need to duplicate the check in the provisioner and the disruption controller, but we should only see these errors in provisioning anyway due to the fallback policy.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -308,28 +371,79 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
...
 			}
 		}
-		nodeClaim := NewNodeClaim(nodeClaimTemplate, s.topology, s.daemonOverhead[nodeClaimTemplate], instanceTypes)
+
+		nodeClaim := NewNodeClaim(nodeClaimTemplate, s.topology, s.daemonOverhead[nodeClaimTemplate], instanceTypes, s.reservationManager, s.reservedOfferingMode)
 		if err := nodeClaim.Add(pod, s.cachedPodData[pod.UID]); err != nil {
-			nodeClaim.Destroy() // Ensure we cleanup any changes that we made while mocking out a NodeClaim
-			errs = multierr.Append(errs, fmt.Errorf("incompatible with nodepool %q, daemonset overhead=%s, %w",
+			nodeClaim.Destroy()
+			if IsReservedOfferingError(err) {
```
jonathan-innis, line 409: Rather than worrying about NodeClaims of equal weight and handling that within one scheduling loop, can we just bail-out here for this pod when we get an error like this and we are in non-fallback mode? Doesn't that mean that whichever NodeClaim was ordered first wasn't compatible and that we should try later for the pod in another scheduling loop? I feel like that might also simplify some of your logic here

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -308,28 +371,79 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
...
+					err,
+				))
+			}
+			continue
+		}
+
+		// TODO: If necessary, could mark the reserved offering status in the scheduling.NodeClaim rather than crawling the
+		// offerings. Granted, this won't be an issue unless there is a compatible reserved NodeClaimTemplate since we short
+		// circuit the conditional.
+		if s.reservedOfferingMode == ReservedOfferingModeStrict && compatibleReservedNCT != nil && !lo.ContainsBy(nodeClaim.InstanceTypeOptions, func(it *cloudprovider.InstanceType) bool {
```
jonathan-innis, line 430: Maybe I'm crazy but this seems pretty complicated for me to reason about -- does this problem just evaporate if we don't worry about equal-weight NodePools?

pkg/controllers/provisioning/scheduling/nodeclaim.go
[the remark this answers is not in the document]
jmdeal: Yeah, I think this warrants a more in-depth comment. I'll go ahead and make an update 👍

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -103,26 +136,111 @@ func (n *NodeClaim) Add(pod *v1.Pod, podData *PodData) error {
...
 // FinalizeScheduling is called once all scheduling has completed and allows the node to perform any cleanup
 // necessary before its requirements are used for instance launching
 func (n *NodeClaim) FinalizeScheduling() {
 	// We need nodes to have hostnames for topology purposes, but we don't want to pass that node name on to consumers
 	// of the node as it will be displayed in error messages
-	delete(n.Requirements, v1.LabelHostname)
+	delete(n.Requirements, corev1.LabelHostname)
+	// If there are any reserved offerings tracked, inject those requirements onto the NodeClaim. This ensures that if
+	// there are multiple reserved offerings for an instance type, we don't attempt to overlaunch into a single offering.
+	if len(n.reservedOfferings) != 0 {
```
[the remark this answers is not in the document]
jmdeal: I think it might be a little easier to follow just handling it in finalization, since you don't need to worry about both adding and removing offerings. I don't think there would be any technical reason we couldn't though.

pkg/controllers/provisioning/scheduling/reservationmanager.go
```diff
@@ -0,0 +1,85 @@
...
+	return &ReservationManager{
+		reservations: map[string]sets.Set[string]{},
+		capacity:     capacity,
+	}
+}
+
+func (rm *ReservationManager) Reserve(hostname string, offering *cloudprovider.Offering) bool {
+	reservations, ok := rm.reservations[hostname]
+	if ok && reservations.Has(offering.ReservationID()) {
+		return true
```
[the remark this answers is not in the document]
jmdeal: Yes

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -261,15 +308,30 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
+	}
+
 	return Results{
-		NewNodeClaims: s.newNodeClaims,
-		ExistingNodes: s.existingNodes,
-		PodErrors:     errors,
+		NewNodeClaims:          s.newNodeClaims,
+		ExistingNodes:          s.existingNodes,
+		PodErrors:              errors,
+		ReservedOfferingErrors: reservedOfferingErrors,
```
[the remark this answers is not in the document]
jmdeal: We could, the main reason I've been hesitant to though is that there's nowhere that these are actually treated as errors. Implementing the same error filtering anywhere we call schedule seems error prone. Granted, there aren't that many places, so I don't feel that strongly on this.

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -139,14 +144,26 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *v1.NodeClaim) (*v
...
 		}
 	}
-	// Find Offering
-	for _, o := range instanceType.Offerings.Available() {
-		if reqs.IsCompatible(o.Requirements, scheduling.AllowUndefinedWellKnownLabels) {
-			labels[corev1.LabelTopologyZone] = o.Requirements.Get(corev1.LabelTopologyZone).Any()
-			labels[v1.CapacityTypeLabelKey] = o.Requirements.Get(v1.CapacityTypeLabelKey).Any()
-			break
+	// Find offering, prioritizing reserved instances
+	offering := func() *cloudprovider.Offering {
```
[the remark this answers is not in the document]
jmdeal: Yeah, I think it's largely a personal preference of mine to limit the scope of intermediary variables without requiring frequent goto definition jumps for these small functions. I'm happy to inline this though, I don't feel strongly about it.

pkg/cloudprovider/types.go
```diff
@@ -239,30 +244,51 @@ func (i InstanceTypeOverhead) Total() corev1.ResourceList {
...
 // these properties are captured with labels in Requirements.
-// Requirements are required to contain the keys v1.CapacityTypeLabelKey and corev1.LabelTopologyZone
+// Requirements are required to contain the keys v1.CapacityTypeLabelKey and corev1.LabelTopologyZone.
 type Offering struct {
-	Requirements scheduling.Requirements
-	Price        float64
-	// Available is added so that Offerings can return all offerings that have ever existed for an instance type,
-	// so we can get historical pricing data for calculating savings in consolidation
-	Available bool
+	Requirements        scheduling.Requirements
```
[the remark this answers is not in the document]
jmdeal: Yeah, I now have the cloudprovider injecting the reservation ID label as a global since we need to inject it into the NodeClaim's requirements. I figured at that point we may as well drop the reservation ID as a separate field and just rely on the one injected into requirements.

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal: I could go either way on this one. On one hand, attempting to schedule to all equal-weight NodePools will minimize provisioning latency. On the other hand, it may make certain issues more difficult to debug since which NodePool is used will no longer be quite as consistent, though the alphabetical ordering for NodePools of equal weight is an implementation detail and could change anyway. Personally, I'd lean toward the reduced latency making this worth it, but given the reduced complexity and predictability arguments I could be convinced otherwise. What do you think?

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal: Yeah, pretty much. I still think this might be worth keeping in, given the latency advantages, but I can be convinced otherwise. If we do decide to keep it in though, I can work on some improved comments and variable names to hopefully make things more clear.

pkg/controllers/provisioning/provisioner.go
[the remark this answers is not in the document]
jmdeal: Discussed offline, going forward with fallback as the default with an option to be strict.

pkg/controllers/provisioning/scheduling/nodeclaim.go
jmdeal: I've changed my mind, I think the particular way I was building it up wasn't necessarily more efficient than taking the diff, since it included offerings we may have never reserved, and I do agree that taking the diff is a bit cleaner. I've updated it in the latest revision (which should be pushed in the next couple of hours).

pkg/operator/options/options.go
jmdeal: I'm actually wondering if we just include the gate at the cloudprovider layer. This logic stays entirely dormant until a cloudprovider provides reserved offerings via the `GetInstanceTypes` implementation. I don't see a reason to add any gating in the upstream code, so I think we can drop the feature flag from here as well.

pkg/controllers/provisioning/scheduling/nodeclaim.go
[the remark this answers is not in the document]
jmdeal: So there is actually a technical reason not to. Right now, if we failed to make a reservation for a given offering in one iteration, we may succeed on subsequent iterations. If we restricted the labels immediately, that would result in us never being able to make a reservation outside of the initial set of reservations we made. In addition to keeping logic simpler (imo), I think the increased NodeClaim flexibility is good justification to keep it as is.

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal: Discussed offline, this is something we can consider bringing back if we see issues with scheduling latency, but given the limited benefit and the increased complexity we're going to drop the changes for now.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -250,26 +297,45 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
...
-		// If unsuccessful, relax the pod and recompute topology
-		relaxed := s.preferences.Relax(ctx, pod)
-		q.Push(pod, relaxed)
-		if relaxed {
-			if err := s.topology.Update(ctx, pod); err != nil {
-				log.FromContext(ctx).Error(err, "failed updating topology")
+		// We should only relax the pod's requirements when the error is not a reserved offering error because the pod may be
+		// able to schedule later without relaxing constraints. This could occur in this scheduling run, if other NodeClaims
+		// release the required reservations when constrained, or in subsequent runs.
+		var relaxed bool
```
jonathan-innis, line 303: Discussed offline: Noted down that we need to provide an example in this comment OR reference out to the test that validates this scenario as an example

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -35,36 +38,66 @@ import (
 type NodeClaim struct {
 	NodeClaimTemplate
 
-	Pods            []*v1.Pod
-	topology        *Topology
-	hostPortUsage   *scheduling.HostPortUsage
-	daemonResources v1.ResourceList
-	hostname        string
+	Pods                 []*corev1.Pod
+	reservedOfferings    map[string]cloudprovider.Offerings
```
jonathan-innis, line 42: Discussed offline: Noting down that we should add a comment above this to note about why we are storing reservedOfferings in this way and calling FinalizeScheduling() with it rather than storing the data in labels as we schedule pods

pkg/controllers/provisioning/scheduling/nodeclaim.go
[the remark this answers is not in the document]
jonathan-innis: Discussed offline: We agreed to document why we took this decision -- this is also covered in a different review comment

pkg/operator/options/options.go
jonathan-innis: Discussed offline: We need to keep this around because we can't re-define a flag in the cloudprovider repo

pkg/cloudprovider/fake/cloudprovider.go
```diff
@@ -121,9 +128,16 @@ func (c *CloudProvider) Create(ctx context.Context, nodeClaim *v1.NodeClaim) (*v
 	reqs := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...)
 	np := &v1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}}
 	instanceTypes := lo.Filter(lo.Must(c.GetInstanceTypes(ctx, np)), func(i *cloudprovider.InstanceType, _ int) bool {
-		return reqs.IsCompatible(i.Requirements, scheduling.AllowUndefinedWellKnownLabels) &&
-			i.Offerings.Available().HasCompatible(reqs) &&
-			resources.Fits(nodeClaim.Spec.Resources.Requests, i.Allocatable())
+		if !reqs.IsCompatible(i.Requirements, scheduling.AllowUndefinedWellKnownLabels) {
```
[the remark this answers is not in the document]
jmdeal: Breaking it out like this made it significantly easier to figure out why an instance was being filtered out with delve when I was writing the scheduling tests. While it's definitely more verbose, I think that benefit makes it worth factoring out this way.

jmdeal:
Ok, this last push includes the consolidation and drift tests along with the updates for the remaining review feedback. I'm going to place a hold on this PR until we wrap out the AWS implementation, just in case any additional requirements come out of it, but this should be good to go.

/hold

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3778,6 +3778,478 @@ var _ = Context("Scheduling", func() {
...
+						{
+							Key:      v1.NodePoolLabelKey,
+							Operator: corev1.NodeSelectorOpIn,
+							Values:   []string{fmt.Sprintf("np-%d", i+1)},
+						},
+					},
+				})
+			})
+
+			// Even though the pods schedule to separate NodePools, those NodePools share a capacity reservation for the
```
jonathan-innis, line 3990: I think this comment is not correct given the code below

pkg/controllers/disruption/consolidation.go
```diff
@@ -308,24 +308,11 @@ func (c *consolidation) computeSpotToSpotConsolidation(ctx context.Context, cand
...
-			if c.capacityType != v1.CapacityTypeReserved {
-				return 0.0, fmt.Errorf("unable to determine offering for %s/%s/%s", c.instanceType.Name, c.capacityType, c.zone)
-			}
-			// If there are no compatible offerings, but the capacity type for the candidate is reserved, we can fall-back to
-			// the on-demand offering to derive pricing.
-			reqs[v1.CapacityTypeLabelKey] = scheduling.NewRequirement(v1.CapacityTypeLabelKey, corev1.NodeSelectorOpIn, v1.CapacityTypeOnDemand)
-			delete(reqs, cloudprovider.ReservationIDLabel)
-			reservedFallback = true
+		compatibleOfferings := c.instanceType.Offerings.Compatible(scheduling.NewLabelRequirements(c.StateNode.Labels()))
+		if len(compatibleOfferings) == 0 {
```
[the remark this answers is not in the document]
jmdeal: Yeah, it shouldn't make a tangible difference from a behavior perspective but I can see just returning a 0 given it is a more expected path.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -155,33 +166,34 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData) error {
...
+// to reserve compatible offerings when some were available.
+//
+//nolint:gocyclo
 func (n *NodeClaim) reserveOfferings(
+	ctx context.Context,
 	instanceTypes []*cloudprovider.InstanceType,
 	nodeClaimRequirements scheduling.Requirements,
-) (map[string]cloudprovider.Offerings, error) {
+) (cloudprovider.Offerings, error) {
+	if !opts.FromContext(ctx).FeatureGates.CapacityReservations {
```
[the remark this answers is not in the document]
jmdeal: Yeah, honestly that's the only place we need to gate it, but this opts users out of any potential performance penalties. While we're still working on building out our high-scale benchmark capabilities, I figured we can play it safe.

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -521,4 +525,78 @@ var _ = Describe("Drift", func() {
...
+					v1.CapacityTypeLabelKey:          v1.CapacityTypeReserved,
+					cloudprovider.ReservationIDLabel: reservedOffering.ReservationID(),
+				}))
+			}
+
+			ctx = options.ToContext(ctx, test.Options(test.OptionsFields{FeatureGates: test.FeatureGates{CapacityReservations: lo.ToPtr(true)}}))
+		})
+		// This is required to support cloudproviders dynamically updating the capacity type based on reservation expirations
+		It("should drift reserved nodeclaim if the capacity type label has been updated", func() {
+			cp.Drifted = ""
```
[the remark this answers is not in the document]
jmdeal: I'll have to double check then - without it I was running into issues with it being set, but it was intermittent. I assumed it was because it wasn't being reset.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -4379,4 +4380,142 @@ var _ = Describe("Consolidation", func() {
...
+					Case(v1.CapacityTypeSpot, spotNodeClaim).
+					Default(reservedNodeClaim)
+				node = lo.Switch[string, *corev1.Node](initialCapacityType).
+					Case(v1.CapacityTypeOnDemand, node).
+					Case(v1.CapacityTypeSpot, spotNode).
+					Default(reservedNode)
+
+				// If the capacity type is reserved, we will need a cheaper reserved instance to consolidat into
+				var leastExpensiveReservationID string
+				if initialCapacityType == v1.CapacityTypeReserved {
```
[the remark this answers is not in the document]
jmdeal: Discussed offline, I'll go ahead and separate it out into a different test

pkg/controllers/provisioning/provisioner.go
```diff
@@ -311,10 +316,30 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 		}
 		return scheduler.Results{}, fmt.Errorf("creating scheduler, %w", err)
 	}
+
 	results := s.Solve(ctx, pods).TruncateInstanceTypes(scheduler.MaxInstanceTypes)
-	scheduler.UnschedulablePodsCount.Set(float64(len(results.PodErrors)), map[string]string{scheduler.ControllerLabel: injection.GetControllerName(ctx)})
+	reservedOfferingErrors := results.ReservedOfferingErrors()
+	if len(reservedOfferingErrors) != 0 {
```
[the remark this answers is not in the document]
jmdeal: It's being used further down, to subtract from the `UnschedulablePodsCount` metric, that's why it's not inlined

pkg/controllers/provisioning/scheduling/suite_test.go
jmdeal: Good catch, you're right. Will update

pkg/controllers/nodeclaim/disruption/drift_test.go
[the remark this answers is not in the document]
jmdeal: It actually looks like it's the opposite - it's set to "drifted" in the reset function. I think there's a good argument that we should update the default there, but I think that can be decoupled from this PR.

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
I can see the light at the end of the tunnel. Thanks for your hard work on this @jmdeal!

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -516,4 +520,51 @@ var _ = Describe("Drift", func() {
...
+				o.SetLabels(lo.Assign(o.GetLabels(), map[string]string{
+					v1.CapacityTypeLabelKey:          v1.CapacityTypeReserved,
+					cloudprovider.ReservationIDLabel: reservedOffering.ReservationID(),
+				}))
+			}
+
+			ctx = options.ToContext(ctx, test.Options(test.OptionsFields{FeatureGates: test.FeatureGates{ReservedCapacity: lo.ToPtr(true)}}))
+		})
+		// This is required to support cloudproviders dynamically updating the capacity type based on reservation expirations
+		It("should drift reserved nodeclaim if the capacity type label has been updated", func() {
```
jonathan-innis, line 557: Is this not already covered by other drift testing? Kind of odd to me that we are specifically targeting this scenario when I would just expect it to be covered by generic drift requirement checking

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -115,9 +115,20 @@ func instanceTypeNotFound(its []*cloudprovider.InstanceType, nodeClaim *v1.NodeC
...
-	// Offerings should in most cases only have zone and capacity type. This likely shouldn't differ
-	// across cloud providers.
-	if !ok || !it.Offerings.HasCompatible(scheduling.NewLabelRequirements(nodeClaim.Labels)) {
+	if !ok {
+		return InstanceTypeNotFound
+	}
+	reqs := scheduling.NewLabelRequirements(nodeClaim.Labels)
+	// The reserved capacity type is special because a NodeClaim can be demoted from reserved to on-demand after creation.
+	// For this reason, when evaluating drift due to unavailable offerings, we should check both reserved and on-demand for
+	// reserved nodeclaims. This ensures we don't drift a nodeclaim whoes label hasn't been updated yet. If the NodePool
```
jonathan-innis, line 124:
```suggestion
	// reserved nodeclaims. This ensures we don't drift a nodeclaim whose label hasn't been updated yet. If the NodePool
```

nit

pkg/controllers/nodeclaim/disruption/drift_test.go
jmdeal: Good point, I don't think there's a reason we need to explicitly have this

pkg/controllers/nodeclaim/disruption/drift.go
jmdeal: Dropping with the test

jonathan-innis (approved):
/lgtm
/approve

PR #1914, opened 2025-01-14
fix: Modify conversion webhook check to check for v1 version existence

jonathan-innis:
Fixes #N/A 

**Description**

This change updates the conversion webhook check that we added to the latest pre-v1.0 patch versions so that we crash when we have an issue calling the v1 conversion endpoint converting from v1beta1, not when we get any error -- this allows the v1 CRD to be deployed after the application is deployed.

This also logs errors if we get any unknown errors when calling this API just in case there are errors that we are getting in this health check that are unexpected and we don't see.

This also ensures that this health check will continually fire throughout the lifetime of the process so that if the CRD is updated and deployed later after the application is deployed or if the NodePools aren't deployed until after the process is started, the check continues to be executed.

We also need to ensure that we don't fail immediately when we restart the container so that we give an unhealthy container a chance to become healthy again -- if a Karpenter container was crahsing with a blocked network path to the conversion webhook and that network path has now opened back up again -- we need to allow for that container to go back to ready -- we can make that happen by ensuring that we don't crash on the first failure but checking that we have consecutive failures.

`make presubmit`

### DIFF
--- a/pkg/webhooks/webhooks.go
+++ b/pkg/webhooks/webhooks.go
@@ -19,7 +19,7 @@ package webhooks
 import (
 	"context"
 	"crypto/tls"
-	"errors"
+	stderrors "errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -29,6 +29,8 @@ import (
 	"github.com/samber/lo"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/rest"
 	"knative.dev/pkg/configmap"
@@ -45,6 +47,7 @@ import (
 	"knative.dev/pkg/webhook/resourcesemantics/validation"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/apis/v1beta1"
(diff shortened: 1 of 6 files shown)


pkg/webhooks/webhooks.go
```diff
@@ -220,21 +223,23 @@ func HealthProbe(ctx context.Context) healthz.Checker {
...
+func ValidateConversionEnabled(ctx context.Context, directKubeClient client.Client) {
 	for {
-		err = kubeclient.List(listCtx, v1np, &client.ListOptions{Limit: 1})
-		if err == nil {
-			return
+		// We only need to try to get a single NodePool teh
+		err := directKubeClient.List(ctx, &v1.NodePoolList{}, &client.ListOptions{Limit: 1})
+		// We only want to bail out with a failure if we know the reason that we are not able to List on this API
+		// is due to an internal error on the conversion webhook
+		if errors.IsInternalError(err) && strings.Contains(err.Error(), "conversion webhook for karpenter.sh/v1beta1, Kind=NodePool failed") {
```
engedaam, line 232: Can we also not get an error as such:
`conversion webhook for karpenter.sh/v1, Kind=NodePool failed`

pkg/webhooks/webhooks.go
```diff
@@ -220,21 +223,23 @@ func HealthProbe(ctx context.Context) healthz.Checker {
...
+			panic(fmt.Sprintf("validating conversion webhook reachable, %s", err.Error()))
+		}
+		// If there's an error that occurs that we don't know about, and it's not about the API not existing, then log the error
+		if err != nil && !meta.IsNoMatchError(err) {
+			log.FromContext(ctx).Error(err, "failing validating conversion webhook reachable")
```
jmdeal, line 237: tiny nit:
```suggestion
			log.FromContext(ctx).Error(err, "failed validating conversion webhook reachable")
```

pkg/webhooks/webhooks.go
jonathan-innis: No, since this is going to just be pulling the v1 version and there isn't going to be any conversion from v1 to v1

jmdeal: I imagine we shouldn't see this if the stored version is v1 since we'll get the object directly without using the conversion webhook. I do wonder though, what happens when there's a mix of v1 and v1beta1 stored resources in the cluster? Since we're limiting the list to a single resource, could we get a false negative if the resource we get is v1?

PR #1916, opened 2025-01-15
chore: Add duplicate nodeclaim chack for Repair Controller

engedaam:
Fixes #N/A 

**Description**
- Adding nodeclaim check for the repair controller for duplicate nodeclaims. We should expect manual intervention for node has multiple nodeclaim mapping. We should not re-recocile on a node with multiple nodeclaim mapping 

- `make presubmit`

### DIFF
--- a/pkg/controllers/node/health/controller.go
+++ b/pkg/controllers/node/health/controller.go
@@ -77,32 +77,24 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef(node.Namespace, node.Name)))
 
 	// Validate that the node is owned by us
+	if !nodeutils.IsManaged(node, c.cloudProvider) {
+		return reconcile.Result{}, nil
+	}
 	nodeClaim, err := nodeutils.NodeClaimForNode(ctx, c.kubeClient, node)
 	if err != nil {
-		return reconcile.Result{}, nodeutils.IgnoreNodeClaimNotFoundError(err)
-	}
-
-	// If a nodeclaim does has a nodepool label, validate the nodeclaims inside the nodepool are healthy (i.e bellow the allowed threshold)
-	// In the case of standalone nodeclaim, validate the nodes inside the cluster are healthy before proceeding
-	// to repair the nodes
-	nodePoolName, found := nodeClaim.Labels[v1.NodePoolLabelKey]
-	if found {
-		nodePoolHealthy, err := c.isNodePoolHealthy(ctx, nodePoolName)
-		if err != nil {
-			return reconcile.Result{}, client.IgnoreNotFound(err)
-		}
-		if !nodePoolHealthy {
-			return reconcile.Result{}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
-		}
-	} else {
-		clusterHealthy, err := c.isClusterHealthy(ctx)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		if !clusterHealthy {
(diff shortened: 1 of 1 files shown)


pkg/controllers/node/health/controller.go
```diff
@@ -78,30 +78,16 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		if !clusterHealthy {
-			c.recorder.Publish(NodeRepairBlockedUnmanagedNodeClaim(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String()))...)
+		if nodeutils.IsDuplicateNodeClaimError(err) {
+			log.FromContext(ctx).Error(err, "failed to validate node health")
 			return reconcile.Result{}, nil
 		}
+		return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("repairing node, %w", err))
```
jmdeal, line 85: nit: should this match the printed error above? 
```suggestion
		return reconcile.Result{}, client.IgnoreNotFound(fmt.Errorf("validating node health, %w", err))
```

jmdeal (commented):
One nit

/lgtm
/hold

jmdeal (approved):
/lgtm

pkg/controllers/node/health/controller.go
```diff
@@ -79,30 +79,19 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
-		}
-	} else {
-		clusterHealthy, err := c.isClusterHealthy(ctx)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		if !clusterHealthy {
-			c.recorder.Publish(NodeRepairBlockedUnmanagedNodeClaim(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String()))...)
+		if nodeutils.IsDuplicateNodeClaimError(err) {
+			log.FromContext(ctx).Error(err, "failed to validate node health")
```
jonathan-innis, line 83: It's odd that we would just be logging the error and not returning the error. If you are looking to handle cases where you don't want to keep requeueing because the error isn't retryable, you can consider doing something like a TerminalError

jonathan-innis: Anyways, do you need to log an error for this kind of failure, I think this should propagate up as a registration error if this even happens at all -- I'd probably just consider ignoring this rather than logging for it

pkg/controllers/node/health/controller.go
```diff
@@ -79,30 +79,19 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
...
-			return reconcile.Result{}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
-		}
-	} else {
-		clusterHealthy, err := c.isClusterHealthy(ctx)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
-		if !clusterHealthy {
-			c.recorder.Publish(NodeRepairBlockedUnmanagedNodeClaim(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String()))...)
+		if nodeutils.IsDuplicateNodeClaimError(err) {
```
jonathan-innis, line 82: Why did we remove the check for validating if there is an owning NodeClaim?

engedaam: Discussed offline, switching back to use `nodeutils.IgnoreNodeClaimNotFoundError`

pkg/controllers/node/health/controller.go
engedaam: I can use `reconcile.TerminalError` here. The idea was just to fail load and not reconcile on the error. Customer should really be intervening and fixing the broken state

engedaam: Yeah, that seems reasonable to me. I can drop this check all together. Gonna open a new PR to avoid needing to alter to much in this commit

PR #1925, opened 2025-01-22
BREAKING: Allow custom condition reasons to be returned from Create

jonathan-innis:
Fixes #N/A 

**Description**

This change allows CloudProviders to not just affect the condition message when the Create returns an error but also to affect the condition reason. 

This change is a breaking change since any CloudProvider that uses the CreateError type now needs to specify a "reason" value with the CreateError or will get errors when trying to assign the reason to the Launched status condition.

### Motivation

We were already supporting adding a custom error message but we'd like to start allowing classification of "error codes" back from CloudProviders but specifying well-structured custom reasons. We may be able to build on this for metrics and when considering things like #1910 which may propagate similar information down into the degraded condition based on the failure code that NodeClaims are indicating (and whether they succeeded/failed to launch or not -- imagine differentiating a user-induced error vs. something that failed on the service-side)

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -388,12 +388,14 @@ func IsNodeClassNotReadyError(err error) bool {
 // CreateError is an error type returned by CloudProviders when instance creation fails
 type CreateError struct {
 	error
+	ConditionReason  string
 	ConditionMessage string
 }
 
-func NewCreateError(err error, message string) *CreateError {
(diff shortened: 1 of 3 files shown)


enxebre:
Is there a motivational use case / provider consumption driving this change we can link/articulate in the PR desc so is clear for folks with less context why the additional flexibility is desired?

jmdeal (commented):
/lgtm

Issue #1928, opened 2025-01-24
Ensure pods that can't drain aren't considered for provisioning rescheduling


jorgeperezc:
### Description

**Observed Behavior**:
The releases of v1 brings a gift of a new approach regarding the node lifecycle, with graceful and forceful methods for draining and expiration.

Regarding node expiration, according to the Nodepool schema specification, the NodeTerminationGracePeriod property acts as a feature flag to enable a maximum threshold for recycling. If defined as null, the expiration controller will wait indefinitely, respecting do-not-disrupt, PDBs and so on.(Note that I do not remember reading this in the documentation).

```
  terminationGracePeriod	<string>
    TerminationGracePeriod is the maximum duration the controller will wait
    before forcefully deleting the pods on a node, measured from when deletion
    is first initiated.
    
    [...]

    The feature can also be used to allow maximum time limits for long-running
    jobs which can delay node termination with preStop hooks.
    If left undefined, the controller will wait indefinitely for pods to be
    drained.
```

Having said that, two erratic behaviors can be observed:

- Forceful method enabled. The terminationGracePeriod property is defined as the maximum grace period threshold for draining the node's pods. When the expiration of NodeClaims begins (TTL specified in the expireAfter setting), they are marked with the annotation `karpenter.sh/nodeclaim-termination-timestamp`, indicating the maximum datetime for decommissioning, and the grace period countdown starts. The affected node workloads, regardless of PDBs and the do-not-disrupt annotation, are identified by the provisioner controller as reschedulable pods, causing the scheduler to determine whether to generate a new NodeClaim as a replacement based on the available capacity. We have use cases with extended grace periods and workloads with significant sizing but the scheduler does not consider the potential maximum grace period, provisioning replacements that might not be used until the application terminates. Additionally, there are pods nominated to be scheduled on the newly provisioned NodeClaims, blocking possible disruptions, lack of synchronization of the cluster state with the in-memory snapshot of Karpenter and extensive enqueuing of reconciliations by the provisioner, creating the perfect ingredients for a disaster. I believe it does not make sense to flip between provisioning and consolidation with resources that may not be used, leading to unnecessary costs. For example, jobs with a TTL of 5 hours that could use the entire grace period but from t0 already have an unused replacement. Aggressive consolidation budgets tend to worsen the situation, leading to more chaos.

[...]

PR #1929, opened 2025-01-24
Avoid overprovisioning capacity for workloads on expired nodes

jorgeperezc:
Fixes #N/A 

**Description**
Avoid overprovisioning capacity for workloads on expired nodes

Yes, issue opened: https://github.com/kubernetes-sigs/karpenter/issues/1928


pkg/controllers/provisioning/provisioner.go
```diff
@@ -319,11 +324,33 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 		return scheduler.Results{}, err
 	}
 
+	nodesMarkedForDeletion := nodes.Deleting()
+
+	nodesMarkedForDeletion = lo.Filter(nodesMarkedForDeletion, func(n *state.StateNode, _ int) bool {
+		expirationTimeString, exists := n.NodeClaim.ObjectMeta.Annotations[v1.NodeClaimTerminationTimestampAnnotationKey]
```
jonathan-innis, line 330: I don't think I'm convinced that we want to just ignore nodes marked for deletion in general when we are grabbing pods from these nodes -- it seems to me that we'd be better off checking the pods themselves and not including them in the scheduling calculation -- for instance, `karpenter.sh/do-not-disrupt` pods -- I think in general, we expect most pods to deprovision within a short amount of time, so we should definitely provision capacity for all of these so that we don't end-up trickling capacity onto the cluster

jorgeperezc: I do not agree with you @jonathan-innis . There are workloads of all kinds. For example, jobs have a broad execution time and need to finish before they die or apps require an extensive graceful shutdown. If both `expireAfter` and `TGP` are defined, you may end up paying for resources that you cannot use until the workload effectively terminates like your maximum grace period. And if `TGP` is undefined but `expireAfter` is defined, you will provision capacity for workloads that never die.

I really appreciate your review of this PR. It may not be the right place or code and could even affect other components but it reflects the expected behavior to prevent cloud providers from becoming even richer.

jonathan-innis: > jobs have a broad execution time and need to finish before they die or apps require an extensive graceful shutdown

I think in general we haven't seen that apps have extensively long graceful shutdowns -- also, if a node starts deprovisioning, it doesn't wait for the job pod to finish to evict it.

> you will provision capacity for workloads that never die

I think in general this isn't true except for in the case of stuck PDBs and `karpenter.sh/do-not-disrupt`. `karpenter.sh/do-not-disrupt` is used heavily so I agree with this one; however, I think that a fully blocking PDB is more of a misconfiguration than a pattern.

The other thing to keep in mind here is that the reason we originally added this feature way back when was to prevent trickle behavior as pods drain off of nodes. If we don't pre-create the capacity as pods drain off of nodes, what you end up seeing is that Karpenter will launch a bunch of small nodes one after the other as pods go pending and will create a bunch of sprawl on your cluster (and potentially a bunch of underutilized nodes).

jonathan-innis: Honestly, I think your solution isn't that far off from what I agree with -- ignoring `karpenter.sh/do-not-disrupt` pods because it may be a while before these pods get evicted, but eviction isn't an all-or-nothing thing and that the current code is handling this is going to create node trickle

PR #1938, opened 2025-01-28
chore: Use self-affinity to ensure all pods schedule

jonathan-innis:
Fixes #N/A 

**Description**

Use self-affinity to ensure that all pods schedule when executing the scheduling benchmark. This also expands the test case to run up to 20,000 pods since the tests are executing quicker now

### Scheduling Benchmark

```
=== RUN   TestSchedulingProfile
scheduled 40151 against 8031 nodes in total in 59.075917891s 679.6508870853593 pods/sec
400 instances 1 pods      1 nodes     183.429µs per scheduling      183.429µs per pod
400 instances 50 pods     10 nodes    42.179802ms per scheduling    843.596µs per pod
400 instances 100 pods    20 nodes    93.097613ms per scheduling    930.976µs per pod
400 instances 500 pods    100 nodes   415.767736ms per scheduling   831.535µs per pod
400 instances 1000 pods   200 nodes   890.825521ms per scheduling   890.825µs per pod
400 instances 1500 pods   300 nodes   1.300125958s per scheduling   866.75µs per pod
400 instances 2000 pods   400 nodes   1.625429875s per scheduling   812.714µs per pod
400 instances 5000 pods   1000 nodes  5.086005208s per scheduling   1.017201ms per pod
400 instances 10000 pods  2000 nodes  12.071574167s per scheduling  1.207157ms per pod
400 instances 20000 pods  4000 nodes  35.739747083s per scheduling  1.786987ms per pod
--- PASS: TestSchedulingProfile (66.78s)
PASS
```

`go test -bench=.`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
+++ b/pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
@@ -90,6 +90,12 @@ func BenchmarkScheduling2000(b *testing.B) {
 func BenchmarkScheduling5000(b *testing.B) {
 	benchmarkScheduler(b, 400, 5000)
 }
+func BenchmarkScheduling10000(b *testing.B) {
+	benchmarkScheduler(b, 400, 10000)
+}
+func BenchmarkScheduling20000(b *testing.B) {
+	benchmarkScheduler(b, 400, 20000)
+}
 
 // TestSchedulingProfile is used to gather profiling metrics, benchmarking is primarily done with standard
 // Go benchmark functions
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -90,6 +90,12 @@ func BenchmarkScheduling2000(b *testing.B) {
 func BenchmarkScheduling5000(b *testing.B) {
 	benchmarkScheduler(b, 400, 5000)
 }
+func BenchmarkScheduling10000(b *testing.B) {
+	benchmarkScheduler(b, 400, 10000)
+}
+func BenchmarkScheduling20000(b *testing.B) {
+	benchmarkScheduler(b, 400, 20000)
+}
```
rschalo, line 98: Nice!

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -259,15 +285,16 @@ func makePodAntiAffinityPods(count int, key string) []*corev1.Pod {
 func makePodAffinityPods(count int, key string) []*corev1.Pod {
 	var pods []*corev1.Pod
 	for i := 0; i < count; i++ {
+		labels := randomAffinityLabels() // self-affinity
```
rschalo, line 288: At thousands of pods scale, were we having collisions?

jonathan-innis: It's not that we were having collisions, it's that at smaller scale, you don't guarantee that there is a pod in every domain -- in order for the affinity to be satisfied, you have to have a pod in that domain (unless you are using self-affinity, then you are allowed to schedule the pod so long as there are already no pods across all domains)

PR #1943, opened 2025-01-29
chore: drift reason should be available in info logs

jigisha620:
Fixes #N/A 

**Description**
Customers usually need to enable debug logging if they want to understand why a nodeclaim is drifted from the logs. This change doesn't require customers to enable debug logging and they can still get this information when running Karpenter in INFO mode.

N/A


pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -70,7 +70,7 @@ func (d *Drift) Reconcile(ctx context.Context, nodePool *v1.NodePool, nodeClaim
 	// 3. Finally, if the NodeClaim is drifted, but doesn't have status condition, add it.
 	nodeClaim.StatusConditions().SetTrueWithReason(v1.ConditionTypeDrifted, string(driftedReason), string(driftedReason))
 	if !hasDriftedCondition {
-		log.FromContext(ctx).V(1).WithValues("reason", string(driftedReason)).Info("marking drifted")
+		log.FromContext(ctx).WithValues("reason", string(driftedReason)).Info("marking drifted")
```
jonathan-innis, line 73: I'm not 100% confident that we want to do this. How noisy do we think this is going to be when we set it to this log level?

PR #1944, opened 2025-01-31
chore: Update nodeclaim finalization flow to not depend on cloudProvider.Get()

jigisha620:
Fixes #N/A 

**Description**
With karpenter aws-provider we saw an issue where eventual consistency with EC2 can cause the underlying instance to be still around even when cloudProvider.Delete() call is made once against it. To tackle this in this PR, we remove `InstanceTerminating` status condition if, on cloudProvider.Get() we don't see deletion timestamp on the nodeClaim.

Added test for it.

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -57,7 +57,9 @@ type CloudProvider interface {
 	// Create launches a NodeClaim with the given resource requests and requirements and returns a hydrated
 	// NodeClaim back with resolved NodeClaim labels for the launched NodeClaim
 	Create(context.Context, *v1.NodeClaim) (*v1.NodeClaim, error)
-	// Delete removes a NodeClaim from the cloudprovider by its provider id
+	// Delete removes a NodeClaim from the cloudprovider by its provider id. Delete should return
+	// NodeClaimNotFoundError if the cloudProvider instance is already terminated and nil if deletion was triggered.
+	// Karpenter will keep retrying until Delete returns a NodeClaimNotFound error.
 	Delete(context.Context, *v1.NodeClaim) error
 	// Get retrieves a NodeClaim from the cloudprovider by its provider id
 	Get(context.Context, string) (*v1.NodeClaim, error)
--- a/pkg/utils/termination/termination.go
+++ b/pkg/utils/termination/termination.go
@@ -20,62 +20,34 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/api/equality"
+
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 )
 
-// EnsureTerminated is a helper function that takes a v1.NodeClaim and calls cloudProvider.Delete() if status condition
-// on nodeClaim is not terminating. If it is terminating then it will call cloudProvider.Get() to check if the instance
-// is terminated or not. It will return an error and a boolean that indicates if the instance is terminated or not. We simply return
-// conflict or a NotFound error if we encounter it while updating the status on nodeClaim.
+// EnsureTerminated is a helper function that takes a v1.NodeClaim and calls cloudProvider.Delete(). It will return an
+// error and a boolean that indicates if the instance is terminated or not. We simply return conflict or a NotFound
+// error if we encounter it while updating the status on nodeClaim.
 func EnsureTerminated(ctx context.Context, c client.Client, nodeClaim *v1.NodeClaim, cloudProvider cloudprovider.CloudProvider) (terminated bool, err error) {
-	// Check if the status condition on nodeClaim is Terminating
-	if !nodeClaim.StatusConditions().Get(v1.ConditionTypeInstanceTerminating).IsTrue() {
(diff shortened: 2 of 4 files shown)


pkg/utils/termination/termination.go
```diff
@@ -27,45 +27,35 @@ import (
...
-				// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
-				// can cause races due to the fact that it fully replaces the list on a change
-				// Here, we are updating the status condition list
-				if err = c.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
-					return false, err
-				}
-				// Instance is terminated
-				return true, nil
+	if err = cloudProvider.Delete(ctx, nodeClaim); err != nil {
+		if !nodeClaim.StatusConditions().Get(v1.ConditionTypeInstanceTerminating).IsTrue() {
```
jonathan-innis, line 34: Do we want to add this only when we get an error? That doesn't make a ton of sense to me -- I think what we may actually want is to set this when we either get a NotFound error or we get no error (in both cases meaning that the Delete was processed or not necessary)

pkg/utils/termination/termination.go
```diff
@@ -27,45 +27,35 @@ import (
...
+			stored := nodeClaim.DeepCopy()
+			updateStatusConditionsForDeleting(nodeClaim)
```
jonathan-innis, line 36: I think you may not need this separate function call anymore -- if we just idempotently set the status condition to true every time then we can just do a deep equality comparison and use that to determine if we should update/patch or not

pkg/utils/termination/termination.go
```diff
@@ -27,45 +27,35 @@ import (
...
+			// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
+			// can cause races due to the fact that it fully replaces the list on a change
+			// Here, we are updating the status condition list
+			if e := c.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); e != nil {
+				return false, e
 			}
-			return false, fmt.Errorf("terminating cloudprovider instance, %w", err)
 		}
-
+		if cloudprovider.IsNodeClaimNotFoundError(err) {
```
jonathan-innis, line 44: I think we should explore collapsing a bit of this -- there's basically duplicate code in two spots in this function

pkg/utils/termination/termination.go
```diff
@@ -20,62 +20,34 @@ import (
...
-				if err = c.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
-					return false, err
-				}
-				// Instance is terminated
-				return true, nil
+	stored := nodeClaim.DeepCopy()
+	err = cloudProvider.Delete(ctx, nodeClaim)
+	// Set InstanceTerminating to true when cloudProvider.Delete() returns -
+	// 1. nodeClaim not found error which indicates the instance was terminated
+	// 2. no error which indicates that Delete() has been processed but the instance has not terminated so requeue
```
elmiko, line 39: i think it would be nice to update the code comment on the `Delete` interface in https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/cloudprovider/types.go#L60 to ensure that provider developers understand the expectation. eg "delete should return nil if the instance is terminating and NotFoundError if the instance is deleted".

elmiko (commented):
code makes sense to me and generally looks good. i have one suggestion.

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
Like @elmiko suggested, we should update the CloudProvider interface so it's clear what the new behavior is

pkg/utils/termination/termination.go
jigisha620: Right, makes sense. I will update the comment.

pkg/cloudprovider/types.go
```diff
@@ -57,7 +57,9 @@ type CloudProvider interface {
 	// Create launches a NodeClaim with the given resource requests and requirements and returns a hydrated
 	// NodeClaim back with resolved NodeClaim labels for the launched NodeClaim
 	Create(context.Context, *v1.NodeClaim) (*v1.NodeClaim, error)
-	// Delete removes a NodeClaim from the cloudprovider by its provider id
+	// Delete removes a NodeClaim from the cloudprovider by its provider id. Delete should return
+	// NodeClaimNotFoundError if the cloudProvider instance is terminated and nil if deletion has been triggered
```
jonathan-innis, line 61:
```suggestion
	// NodeClaimNotFoundError if the cloudProvider instance is already terminated and nil if deletion was successful
```

pkg/cloudprovider/types.go
```diff
@@ -57,7 +57,9 @@ type CloudProvider interface {
...
+	// but the cloudProvider instance has not terminated
```
jonathan-innis, line 62: Rather than saying "but the cloudprovider instance has not terminated", I'd say something like, "Returning nil from this function simply means that the Deletion call was successful. It's up to the CloudProvider to define when to return NotFound based on each one's respective termination flows. Karpenter will continue polling for Deletion until it returns a NodeClaimNotFound error

elmiko: ++, that makes sense to me

jonathan-innis (approved):
/lgtm
/approve

PR #1954, opened 2025-02-03
chore: fix nil pointer dereference

jigisha620:
Fixes #N/A 

**Description**
Saw cases where cleaning up node from cluster state can cause nil pointer dereference when `c.nodes[id] = nil`. Added a check to validate this.

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -547,7 +547,7 @@ func (c *Cluster) newStateFromNode(ctx context.Context, node *corev1.Node, oldNo
 
 func (c *Cluster) cleanupNode(name string) {
 	if id := c.nodeNameToProviderID[name]; id != "" {
-		if c.nodes[id].NodeClaim == nil {
+		if c.nodes[id] != nil && c.nodes[id].NodeClaim == nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -547,7 +547,7 @@ func (c *Cluster) newStateFromNode(ctx context.Context, node *corev1.Node, oldNo
 
 func (c *Cluster) cleanupNode(name string) {
 	if id := c.nodeNameToProviderID[name]; id != "" {
-		if c.nodes[id].NodeClaim == nil {
+		if c.nodes[id] != nil && c.nodes[id].NodeClaim == nil {
```
jonathan-innis, line 550: I think we should understand the scenarios that this can occur under before we just ignore these cases -- it seems odd to me that we would get into this state but maybe this can happen when we have two nodes with the same providerID?

rschalo: Based on discussion in triage meeting, this could possibly be the node name instead?

jonathan-innis:
I think we should probably just track this with an issue -- since this isn't the solve that we wanted to take in -- we can ask someone from the community to take a look at this probably

PR #1957, opened 2025-02-04
test: Fix `ConsolidationState()` test failure in state testing

jonathan-innis:
Fixes #N/A 

**Description**

Fix test flakes coming from the `ConsolidationState()` and DataRace checks inside of the test suite. Resetting the cluster state without using a wait group to complete the goroutine was causing the data race during the Data Race test and not properly resetting all of the details inside of cluster state was causing the `ConsolidationState()` test to fail.

`make presubmit`


engedaam (commented):
/lgtm

PR #1963, opened 2025-02-05
test: Expand KOWKNodeClass to Support Generic NodeClass for E2E tests

engedaam:
Fixes #N/A 

##Description**

This PR extends the current KowkNodeClass implementation in the E2Es to support cloudprovider defined NodeClass, allowing for more flexible and reusable test management across different cloud providers integration testing.

### Changes
- Added abstraction layer for provider-specific implementations
- Updated node lifecycle handlers for generic support

- `make e2etests`

### DIFF
--- a/test/pkg/environment/common/default_nodepool.yaml
+++ b/test/pkg/environment/common/default_nodepool.yaml
@@ -1,0 +1,27 @@
+apiVersion: karpenter.sh/v1
+kind: NodePool
+metadata:
+ name: default
+spec:
+ disruption:
+   consolidationPolicy: WhenEmptyOrUnderutilized
+   consolidateAfter: Never
+   budgets:
+     - nodes: 100%
+ limits:
+   cpu: 1000
+   memory: 1000Gi
+ template:
+   spec:
+     expireAfter: Never
+     requirements:
+       - key: kubernetes.io/os
+         operator: In
+         values: ["linux"]
+       - key: karpenter.sh/capacity-type
+         operator: In
+         values: ["on-demand"]
+     nodeClassRef:
+      group: karpenter.kwok.sh
+      kind: KWOKNodeClass
+      name: default
--- a/test/pkg/environment/common/environment.go
+++ b/test/pkg/environment/common/environment.go
@@ -18,7 +18,10 @@ package common
 
 import (
 	"context"
+	_ "embed"
+	"flag"
 	"fmt"
+	"io"
 	"log"
 	"os"
 	"strconv"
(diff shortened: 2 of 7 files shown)


test/pkg/environment/common/environment.go
```diff
@@ -144,42 +164,45 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
...
+	decoder := serializeryaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
+	u := &unstructured.Unstructured{}
+	ul := &unstructured.UnstructuredList{}
+	_, gvk, _ := decoder.Decode(content, nil, u)
+	u.SetGroupVersionKind(lo.FromPtr(gvk))
+	ul.SetGroupVersionKind(lo.FromPtr(gvk))
+
+	return &GenericNodeClass{
+		Unstructured:     lo.FromPtr(u),
+		GroupVersionKind: lo.FromPtr(gvk),
```
jonathan-innis, line 206: Why do you need to create a separate struct for this? Is this just an `unstructured.Unstructured` or is there something that you are trying to do in the future to extend additional methods off of this object?

test/pkg/environment/common/environment.go
```diff
@@ -30,29 +33,44 @@ import (
...
-	GitRefContextKey = ContextKey("gitRef")
+const GitRefContextKey = ContextKey("gitRef")
+
+// I need to add the the default kwok nodeclass path
+// That way it's not defined in code but we use it when we initialize the nodeclass
+var (
+	//go:embed default_kowknodeclass.yaml
+	defaultNodeClass []byte
+	//go:embed default_nodepool.yaml
+	defaultNodePool []byte
```
jonathan-innis, line 64: What's the reason you need a DefaultNodePool at all? Don't you fully control it in the context of the test and you just need to update the NodeClass to point to the CloudProvider-specific NodeClass type?

test/pkg/environment/common/default_nodepool.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: karpenter.sh/v1
```
jonathan-innis, line 1: Do we absolutely need this default? Or can we control this until we get a request to have something different?

jonathan-innis: Is the concern with injecting cloudprovider-specific requirements into the requirements section?

test/pkg/environment/common/environment.go
```diff
@@ -144,42 +164,45 @@ func NewClient(ctx context.Context, config *rest.Config) client.Client {
...
 
-func (env *Environment) DefaultNodeClass() *v1alpha1.KWOKNodeClass {
-	return &v1alpha1.KWOKNodeClass{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: test.RandomName(),
-		},
+func (env *Environment) DefaultNodePool(nodeClass client.Object) *v1.NodePool {
+	nodePool := &v1.NodePool{}
+	if lo.FromPtr(nodePoolPath) == "" {
+		nodePool = object.Unmarshal[v1.NodePool](defaultNodePool)
```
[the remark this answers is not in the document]
engedaam: The idea here is we set the nodeClass ref by the default nodeclass that is provided by the users. Don't want to short-circuit here

test/pkg/environment/common/default_nodepool.yaml
engedaam: Yeah, the concern is exactly that. We need to be able to allow cloudprovider to inject requirements. We could give providers a global variable to define their requirements, but I figured that would be very restrictive without much upside IMO

test/pkg/environment/common/environment.go
engedaam: We control changing, but we don't control the capacity available  by the cloudprovider. Mainly concerned about updating the requirements

test/pkg/environment/common/environment.go
engedaam: Yeah, I was think we could expanded this struct to contain more configuration information, but I think we can wait on that when the requirement arises

test/pkg/environment/common/environment.go
jonathan-innis: Makes sense. Seems reasonable.

test/pkg/environment/common/environment.go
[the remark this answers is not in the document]
jonathan-innis: Ah, I think it was just reading the diff that confused me

jonathan-innis (approved):
/lgtm
/approve

PR #1964, opened 2025-02-05
test: Ensure that we don't step the idle duration since this can cause the batcher to exit

jonathan-innis:
Fixes #N/A 

**Description**

This ensures that we don't trigger a 5s move on the batcher during testing since this is exactly the same as the idle duration and was causing flakes in testing like in https://github.com/kubernetes-sigs/karpenter/actions/runs/13143439974/attempts/1?pr=1950

`make presubmit`


engedaam (commented):
/lgtm

PR #1966, opened 2025-02-05
perf: Reduce the memory overhead on cluster state `Synced()`

jonathan-innis:
Fixes #N/A 

**Description**

This change updates our cluster state `Synced()` logic so that we only wait for the Nodes and NodeClaims to match at the controller startup and then only check the providerID after that point.

This allows us to save on allocations that we were spending on DeepCopying NodeClaims and Nodes as well as saving CPU time on checking this

### Before PR 

Screenshot 2025-02-05 at 11 57 27 AM

Screenshot 2025-02-05 at 11 58 36 AM

alloc_space_before

(Note the portions of the heap that are used by the NodeClaim and Node deep copies)

### After PR

Screenshot 2025-02-05 at 11 57 20 AM

Screenshot 2025-02-05 at 11 59 44 AM

alloc_space_after

(Note how there are no longer any nodes in the heap that are dedicated toward the Node and NodeClaim deep copies)

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -378,7 +378,7 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 		metrics.NodePoolLabel:     nodeClaim.Labels[v1.NodePoolLabelKey],
 		metrics.CapacityTypeLabel: nodeClaim.Labels[v1.CapacityTypeLabelKey],
 	})
-	// Update the nodeclaim manually in state to avoid evenutal consistency delay races with our watcher.
+	// Update the nodeclaim manually in state to avoid eventual consistency delay races with our watcher.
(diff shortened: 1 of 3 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -378,7 +378,7 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 		metrics.NodePoolLabel:     nodeClaim.Labels[v1.NodePoolLabelKey],
 		metrics.CapacityTypeLabel: nodeClaim.Labels[v1.CapacityTypeLabelKey],
 	})
-	// Update the nodeclaim manually in state to avoid evenutal consistency delay races with our watcher.
+	// Update the nodeclaim manually in state to avoid eventual consistency delay races with our watcher.
```
rschalo, line 381: ty

rschalo (approved):
/lgtm

PR #1967, opened 2025-02-05
perf: Add event filter for NodeClaims that have resolved providerIDs

jonathan-innis:
Fixes #N/A 

**Description**

Add an event filter for the counter controller so that NodeClaims that do not have a providerID don't trigger reconciliation on the controller (since they won't have resolved resources yet and can't impact the resource count calculation anyways)

`make presubmit`


rschalo (approved):
/lgtm

PR #1968, opened 2025-02-05
perf: Don't DeepCopy DaemonSet Pods

jonathan-innis:
Fixes #N/A 

**Description**

When hydrating DaemonSet pod data to handle resource quotas and mutating webhooks, we can reduce the number of pods that we list by using the same labelSelector that the DaemonSet does to perform the list against the cache

`make presubmit`


rschalo (approved):
/lgtm

PR #1969, opened 2025-02-06
feat: add NodeRegistrationHealthy status condition to nodepool

jigisha620:
Fixes #N/A 

**Description**
This PR adds `NodeRegistrationHealthy` status condition to nodePool which indicates if a misconfiguration exists that is preventing successful node launch/registrations that requires manual investigation.

Added tests


pkg/controllers/nodeclaim/lifecycle/initialization.go
```diff
@@ -54,38 +57,73 @@ func (i *Initialization) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim)
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
 	node, err := nodeclaimutils.NodeForNodeClaim(ctx, i.kubeClient, nodeClaim)
 	if err != nil {
 		nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeInitialized, "NodeNotFound", "Node not registered with cluster")
 		return reconcile.Result{}, nil //nolint:nilerr
 	}
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KRef("", node.Name)))
+	if updateNodeClaimInitializedUnknown(node, nodeClaim) {
```
jigisha620, line 73: Moved this logic to a different function to reduce the complexity of this function.

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,101 @@
...
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
+	return &Controller{
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "nodepool.healthyregistration")
+	stored := nodePool.DeepCopy()
```
rschalo, line 58: Let's do this post validation.

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,101 @@
...
+
+	nodeClass, ok := lo.Find(c.cloudProvider.GetSupportedNodeClasses(), func(nc status.Object) bool {
```
rschalo, line 60: Is this a duplicate because of the event filtering done on `controller.go:95`?

pkg/controllers/nodeclaim/lifecycle/initialization.go
```diff
@@ -54,38 +57,73 @@ func (i *Initialization) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim)
 		return reconcile.Result{}, nil
 	}
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("provider-id", nodeClaim.Status.ProviderID))
+	requeue, err := i.updateNodePoolRegistrationHealth(ctx, nodeClaim)
```
rschalo, line 60: Want to capture all of the requeue logic here? Meaning return the errors and then check `IsConflict`

jigisha620: I was trying to avoid doing that here since that's going to contribute to increased complexity of this function and I already tried to remove some code in a different function to reduce it. Also we are only patching once in the function that's getting called so I don't know if there's an advantage to checking conflicts here rather than in the called function unless I am missing something.

pkg/controllers/nodepool/registrationhealth/controller.go
jigisha620: From what I understand `cloudProvider.GetSupportedNodeClasses()` can return a list of supported nodeClasses so basically if we have a cluster with kwok as well as aws-provider then we should only react to updates on nodeclass referenced in the nodePool and in that case, having this check makes sense. So, I don't think this is a duplicate.

pkg/controllers/nodeclaim/lifecycle/initialization.go
rschalo: The only branch where `requeue == true` in `updateNodePoolRegistrationHealth` is when there is a conflict, the others just end up falling into controller runtime's back off strategy when it returns an error. I don't think it overcomplicates or adds to the complexity of the function.

pkg/controllers/nodeclaim/lifecycle/initialization.go
```diff
@@ -54,38 +57,73 @@ func (i *Initialization) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim)
...
-	if !equality.Semantic.DeepEqual(stored, node) {
-		if err = i.kubeClient.Patch(ctx, node, client.MergeFrom(stored)); err != nil {
-			return reconcile.Result{}, err
+	return false
+}
+
+func (i *Initialization) updateNodePoolRegistrationHealth(ctx context.Context, nodeClaim *v1.NodeClaim) (bool, error) {
+	nodePool := &v1.NodePool{}
+	if err := i.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, nodePool); err != nil {
+		return false, err
```
rschalo, line 111: This should probably just return the error and be handled in the reconcile like we do in other reconcile loops.

pkg/controllers/nodeclaim/lifecycle/initialization.go
rschalo: Also, does this fit better as the last step in lifecycle registration?

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -51,6 +55,30 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 	if ttl := registrationTTL - l.clock.Since(registered.LastTransitionTime.Time); ttl > 0 {
 		return reconcile.Result{RequeueAfter: ttl}, nil
 	}
+	nodePool := &v1.NodePool{}
+	if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, nodePool); err != nil {
```
rschalo, line 59: Let's do all of this after calling delete on the nodeclaim - that way we don't hang a nodeclaim delete if we have to deal with a bunch of patch conflict requeues on the nodepool status condition.

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -51,6 +55,30 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
...
+		return reconcile.Result{}, err
+	}
+	if nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown() {
+		stored := nodePool.DeepCopy()
+		// If the nodeClaim failed to launch/register during the TTL set NodeRegistrationHealthy status condition on
+		// NodePool to False. If the launch failed get the launch failure reason and message from nodeClaim.
+		if nodeClaim.StatusConditions().IsTrue(v1.ConditionTypeLaunched) {
```
rschalo, line 66: What do you think about getting this once and storing it in the if statement? Then it can be used in the else to get the launch failure.
```suggestion
		if launchCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched); launchCondition.IsTrue() {
```

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,100 @@
...
+// NewController is a constructor
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
+	return &Controller{
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "nodepool.healthyregistration")
```
rschalo, line 57:
```suggestion
	ctx = injection.WithControllerName(ctx, "nodepool.registrationhealth")
```

pkg/controllers/nodeclaim/lifecycle/initialization.go
jigisha620: I see what you mean. I will update the code.
>does this fit better as the last step in lifecycle registration

I feel that this shouldn't be the last step because with this status condition we are just trying to capture launch/registration failures as of now. Also initialization doesn't fail as of today, it either succeeds or remains stuck in Unknown as we have seen from some GH issues. And when that happens we will not mark the NodePoolRegistrationHealthy SC until there is another nodeclaim that initializes successfully.

pkg/controllers/nodeclaim/lifecycle/initialization.go
jigisha620: Reverted since this is now moved to registration controller.

pkg/controllers/nodepool/registrationhealth/suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+	BeforeEach(func() {
+		nodePool = test.NodePool()
+		nodeClass = test.NodeClass(v1alpha1.TestNodeClass{
+			ObjectMeta: metav1.ObjectMeta{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name},
+		})
+		nodePool.Spec.Template.Spec.NodeClassRef.Group = object.GVK(nodeClass).Group
+		nodePool.Spec.Template.Spec.NodeClassRef.Kind = object.GVK(nodeClass).Kind
+		_ = nodePool.StatusConditions().Clear(v1.ConditionTypeNodeRegistrationHealthy)
+	})
+	It("should ignore setting NodeRegistrationHealthy status condition on NodePools which aren't managed by this instance of Karpenter", func() {
```
jxs1211, line 79: How about changing the reason to `the referenced nodeClass is not supported by cloudProvider`

jigisha620: Thanks for the comment. But the idea is to not capture these in the NodeRegistrationHealthy status condition. If the NodePool isn't referencing the supported nodeClass then the NodePool itself will not go ready.

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -141,6 +147,44 @@ var _ = Describe("Liveness", func() {
 		// If the node hasn't registered in the registration timeframe, then we deprovision the nodeClaim
 		fakeClock.Step(time.Minute * 20)
 		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		nodeRegistrationHealthySC := ExpectStatusConditionExists(nodePool, v1.ConditionTypeNodeRegistrationHealthy)
+		Expect(nodeRegistrationHealthySC.Status).To(Equal(metav1.ConditionFalse))
```
jonathan-innis, line 152: You could consider using `ExpectStatusConditions` that we use in a few places from https://github.com/awslabs/operatorpkg/blob/main/test/expectations/expectations.go#L131

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
+)
+
+// Controller for the resource
+type Controller struct {
```
jonathan-innis, line 43: Might be nice what the point of this controller is -- it's responsibility is to ensure that we reset the status condition for the registration health when there is a change that occurs in the spec of the NodePool

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "nodepool.registrationhealth")
+
+	nodeClass, ok := lo.Find(c.cloudProvider.GetSupportedNodeClasses(), func(nc status.Object) bool {
+		return object.GVK(nc).GroupKind() == nodePool.Spec.Template.Spec.NodeClassRef.GroupKind()
+	})
+	if !ok {
```
jonathan-innis, line 62: Is there a function that @jmdeal created for this? Something like IsManaged() that signals that we do or don't own this NodePool?

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	stored := nodePool.DeepCopy()
+	// If NodeClass/NodePool have been updated then NodeRegistrationHealthy = Unknown
+	if (nodePool.Status.NodeClassObservedGeneration != nodeClass.GetGeneration()) ||
+		(nodePool.Generation != nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).ObservedGeneration) {
+		nodePool.StatusConditions().SetUnknown(v1.ConditionTypeNodeRegistrationHealthy)
+		nodePool.Status.NodeClassObservedGeneration = nodeClass.GetGeneration()
```
jonathan-innis, line 74: I could go either way on this line -- yes, technically this will catch every case where the generation changes and always make sure that you have the correct generation updated in the status, but I'm also comfortable just setting it idempotently outside of this block since we can just always set it, and if it doesn't change, then we just don't patch

pkg/controllers/nodepool/registrationhealth/suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+
+		nodePool.Spec.Limits = map[corev1.ResourceName]resource.Quantity{corev1.ResourceCPU: resource.MustParse("14")}
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		_ = ExpectObjectReconciled(ctx, env.Client, controller, nodePool)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		Expect(nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown()).To(BeTrue())
+		Expect(nodePool.Status.NodeClassObservedGeneration).To(Equal(int64(1)))
+	})
+	It("should set NodeRegistrationHealthy status condition on nodePool as Unknown if the nodePool is updated", func() {
```
jonathan-innis, line 108: Can we have a check for making sure that we don't override with Unknown in the case that we have already set the value to True or False and everything already matches

pkg/controllers/nodepool/registrationhealth/controller.go
jigisha620: Added another function that also returns the status.Object

pkg/controllers/metrics/pod/controller.go
```diff
@@ -252,7 +252,7 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	})
 	c.recordPodSchedulingUndecidedMetric(pod)
 	// Get the time for when we Karpenter first thought the pod was schedulable. This should be zero if we didn't simulate for this pod.
-	schedulableTime := c.cluster.PodSchedulingSuccessTime(types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace})
+	schedulableTime := c.cluster.PodSchedulingSuccessTime(types.NamespacedName{Name: pod.Name, Namespace: pod.Namespace}, false)
```
jonathan-innis, line 255: Can we just create a separate function to access the `PodSchedulingNodeRegistrationHealthySuccessTime` or something like this -- I think a boolean here is a bit hard to reason about

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -302,6 +302,7 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
 	// Pick existing node that we are about to create
 	for _, nodeClaim := range s.newNodeClaims {
 		if err := nodeClaim.Add(pod, s.cachedPodData[pod.UID]); err == nil {
+			s.cluster.MarkPodToNodePoolSchedulingDecision(pod, nodeClaim.Labels[v1.NodePoolLabelKey])
```
jonathan-innis, line 305: How expensive is it to do one iteration through all the pods at the end? You could just iterate through the results and mark pod scheduling decisions with the NodePool attached in the same place I think rather than having to create an internal store to capture this

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -61,6 +72,34 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
...
 	})
-
 	return reconcile.Result{}, nil
 }
+
+// updateNodePoolRegistrationHealth sets the NodeRegistrationHealthy=False
+// on the NodePool if the nodeClaim fails to launch/register
+func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	nodePoolName, ok := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if ok && len(nodePoolName) != 0 {
```
jonathan-innis, line 82: nit: Just check `nodePoolName != ""` -- you don't even have to check `ok`, since if the label doesn't exist then it will just return an empty string -- and `""` is an invalid name anyways

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -61,6 +72,34 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
...
+		nodePool := &v1.NodePool{}
+		if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
```
jonathan-innis, line 84: Do you properly handle the NodePool NotFound error?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -82,9 +84,37 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+		return reconcile.Result{}, err
+	}
 	return reconcile.Result{}, nil
 }
 
+// updateNodePoolRegistrationHealth sets the NodeRegistrationHealthy=True
+// on the NodePool if the nodeClaim that registered is owned by a NodePool
+func (r *Registration) updateNodePoolRegistrationHealth(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	nodePoolName, ok := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if ok && len(nodePoolName) != 0 {
```
jonathan-innis, line 100: Same comment as in the liveness controller -- I would just check if the value is not equal to `""`

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -82,9 +84,37 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+		nodePool := &v1.NodePool{}
+		if err := r.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
```
jonathan-innis, line 102: Handle NotFound error here too

pkg/utils/nodepool/nodepool.go
```diff
@@ -39,6 +39,15 @@ func IsManaged(nodePool *v1.NodePool, cp cloudprovider.CloudProvider) bool {
 	})
 }
 
+func GetNodeClassStatusObject(nodePool *v1.NodePool, cp cloudprovider.CloudProvider) status.Object {
```
jonathan-innis, line 42: I would just call it `GetNodeClass()` -- yeah, the thing that you are returning is `status.Object`, but that's just because all NodeClasses have to implement this interface

pkg/controllers/state/cluster.go
```diff
@@ -58,9 +58,11 @@ type Cluster struct {
 	nodeClaimNameToProviderID map[string]string               // node claim name -> provider id
 	daemonSetPods             sync.Map                        // daemonSet -> existing pod
 
-	podAcks                 sync.Map // pod namespaced name -> time when Karpenter first saw the pod as pending
-	podsSchedulingAttempted sync.Map // pod namespaced name -> time when Karpenter tried to schedule a pod
-	podsSchedulableTimes    sync.Map // pod namespaced name -> time when it was first marked as able to fit to a node
+	podAcks                         sync.Map // pod namespaced name -> time when Karpenter first saw the pod as pending
+	podsSchedulingAttempted         sync.Map // pod namespaced name -> time when Karpenter tried to schedule a pod
+	podsSchedulableTimes            sync.Map // pod namespaced name -> time when it was first marked as able to fit to a node
+	podToNodePool                   sync.Map // pod namespaced name -> nodePool where it can be scheduled
```
jonathan-innis, line 64: I'm not convinced that you need this map

pkg/controllers/state/cluster.go
```diff
@@ -395,9 +420,15 @@ func (c *Cluster) PodSchedulingDecisionTime(podKey types.NamespacedName) time.Ti
 
 // PodSchedulingSuccessTime returns when Karpenter first thought it could schedule a pod in its scheduling simulation.
 // This returns 0, false if the pod was never considered in scheduling as a pending pod.
-func (c *Cluster) PodSchedulingSuccessTime(podKey types.NamespacedName) time.Time {
-	if val, found := c.podsSchedulableTimes.Load(podKey); found {
-		return val.(time.Time)
+func (c *Cluster) PodSchedulingSuccessTime(podKey types.NamespacedName, registrationHealthyCheck bool) time.Time {
```
jonathan-innis, line 423: I would drop this boolean because it strikes me as not intuitive and I would just create another function for accessing this data

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,94 @@
...
+	return &Controller{
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "nodepool.registrationhealth")
+
+	nodeClass := nodepoolutils.GetNodeClassStatusObject(nodePool, c.cloudProvider)
```
jonathan-innis, line 56: Is it too much to have this helper actually retrieve the NodeClass for us that the NodePool is referencing rather than just the schema?

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,94 @@
...
+		return reconcile.Result{}, nil
+	}
+	if err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass); err != nil {
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
+	stored := nodePool.DeepCopy()
+
+	// If NodeClass/NodePool have been updated then NodeRegistrationHealthy = Unknown
+	if (nodePool.Status.NodeClassObservedGeneration != nodeClass.GetGeneration()) ||
+		(nodePool.Generation != nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).ObservedGeneration) {
```
jonathan-innis, line 67: `Get` can return `nil` if the condition isn't found -- how are you going to handle that

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,94 @@
...
+// NewController will create a controller to reset NodePool's registration health when there is an update to NodePool/NodeClass spec
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
+	return &Controller{
+		kubeClient:    kubeClient,
+		cloudProvider: cloudProvider,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, "nodepool.registrationhealth")
```
jonathan-innis, line 54: Should we check whether the NodePool is managed in this Reconcile to match our Predicate or are you wanting to handle that in the GetNodeClass() call

pkg/controllers/nodepool/registrationhealth/suite_test.go
```diff
@@ -0,0 +1,131 @@
...
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		_ = ExpectObjectReconciled(ctx, env.Client, controller, nodePool)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		Expect(nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown()).To(BeTrue())
+		Expect(nodePool.Status.NodeClassObservedGeneration).To(Equal(int64(1)))
+	})
+	It("should not set NodeRegistrationHealthy status condition on nodePool as Unknown if it is already set to true", func() {
```
jonathan-innis, line 115: Not sure that I get this test -- why would we set the status condition to Unknown here -- all of the generation details match so I don't see our controller doing anything

pkg/controllers/nodepool/registrationhealth/suite_test.go
```diff
@@ -0,0 +1,131 @@
...
+		nodePool.StatusConditions().SetTrue(v1.ConditionTypeNodeRegistrationHealthy)
+		nodePool.Status.NodeClassObservedGeneration = int64(1)
+		ExpectApplied(ctx, env.Client, nodePool, nodeClass)
+		_ = ExpectObjectReconciled(ctx, env.Client, controller, nodePool)
+		nodePool = ExpectExists(ctx, env.Client, nodePool)
+		Expect(nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown()).To(BeFalse())
+	})
+	It("should not set NodeRegistrationHealthy status condition on nodePool as Unknown if it is already set to false", func() {
```
jonathan-innis, line 123: I think (at least) one of these tests should validate that we are properly updating the NodeClassObservedGeneration since that's the responsibility of this controller

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -78,6 +81,12 @@ var _ = Describe("Liveness", func() {
 			ExpectFinalizersRemoved(ctx, env.Client, nodeClaim)
```
jonathan-innis, line 81: We should validate this: Do we have a test that we succeed when the NodeClaim has no owning NodePool?

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -17,6 +17,10 @@ limitations under the License.
 package lifecycle_test
 
 import (
+	"time"
+
+	"github.com/awslabs/operatorpkg/status"
+	operatorpkg "github.com/awslabs/operatorpkg/test/expectations"
 	. "github.com/onsi/ginkgo/v2"
```
jonathan-innis, line 24: We should validate this: Do we have a test that we succeed when the NodeClaim has no owning NodePool?

pkg/controllers/nodepool/registrationhealth/controller.go
jigisha620: I wanted to handle that in the the GetNodeClass() call. That should do this for us.

pkg/controllers/nodepool/registrationhealth/controller.go
jigisha620: Will update to check if `nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy)` returns nil and if true, set it to Unknown.

pkg/controllers/nodepool/registrationhealth/suite_test.go
jigisha620: Updated

pkg/controllers/nodepool/registrationhealth/suite_test.go
jigisha620: Wasn't that what you meant [here](https://github.com/kubernetes-sigs/karpenter/pull/1969#discussion_r1955373767)?
```
Can we have a check for making sure that we don't override with Unknown in the case that we have already set the value to True or False and everything already matches
```

pkg/controllers/nodeclaim/lifecycle/registration_test.go
jigisha620: Added

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
jigisha620: Added

pkg/controllers/provisioning/scheduling/scheduler.go
jigisha620: Made an update

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -51,6 +56,12 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 	if ttl := registrationTTL - l.clock.Since(registered.LastTransitionTime.Time); ttl > 0 {
 		return reconcile.Result{RequeueAfter: ttl}, nil
 	}
+	if err := l.updateNodePoolRegistrationHealth(ctx, nodeClaim); err != nil {
+		if errors.IsConflict(err) {
```
jonathan-innis, line 60: nit: It's a little odd to send the conflict error back but not send the NotFound error back -- honestly, I would just handle both out here but it's your call

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -19,6 +19,9 @@ package lifecycle_test
 import (
 	"time"
 
+	"github.com/awslabs/operatorpkg/status"
+
+	operatorpkg "github.com/awslabs/operatorpkg/test/expectations"
```
jonathan-innis, line 24:
```suggestion
	. "github.com/awslabs/operatorpkg/test/expectations"
```

Can you dot-import this?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -83,9 +85,37 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
 	metrics.NodesCreatedTotal.Inc(map[string]string{
 		metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
 	})
+	if err := r.updateNodePoolRegistrationHealth(ctx, nodeClaim); err != nil {
+		if errors.IsConflict(err) {
+			return reconcile.Result{Requeue: true}, nil
```
jonathan-innis, line 90: Same comment as the handling elsewhere in liveness -- should we do NotFound out here

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -83,9 +85,37 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+// updateNodePoolRegistrationHealth sets the NodeRegistrationHealthy=True
+// on the NodePool if the nodeClaim that registered is owned by a NodePool
+func (r *Registration) updateNodePoolRegistrationHealth(ctx context.Context, nodeClaim *v1.NodeClaim) error {
+	nodePoolName := nodeClaim.Labels[v1.NodePoolLabelKey]
+	if nodePoolName != "" {
+		nodePool := &v1.NodePool{}
+		if err := r.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
+			return client.IgnoreNotFound(err)
+		}
+		storedNodePool := nodePool.DeepCopy()
```
jonathan-innis, line 106:
```suggestion
		stored := nodePool.DeepCopy()
```

nit

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -55,6 +59,7 @@ var _ = Describe("Registration", func() {
 				})
 			}
 			nodeClaim := test.NodeClaim(nodeClaimOpts...)
+			nodePool.StatusConditions().SetUnknown(v1.ConditionTypeNodeRegistrationHealthy)
```
jonathan-innis, line 62: Do you need to set this to Unknown? Can't the lack of its existence also serve as it being treated as Unknown?

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -55,15 +53,10 @@ func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reco
 	ctx = injection.WithControllerName(ctx, "nodepool.readiness")
 	stored := nodePool.DeepCopy()
 
-	nodeClass, ok := lo.Find(c.cloudProvider.GetSupportedNodeClasses(), func(nc status.Object) bool {
-		return object.GVK(nc).GroupKind() == nodePool.Spec.Template.Spec.NodeClassRef.GroupKind()
-	})
-	if !ok {
-		// Ignore NodePools which aren't using a supported NodeClass.
+	nodeClass, err := nodepoolutils.GetNodeClass(ctx, c.kubeClient, nodePool, c.cloudProvider)
+	if nodeClass == nil {
```
jonathan-innis, line 57: nit: Comment that we are ignoring NodeClasses that we don't support

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,95 @@
...
+
+	nodeClass, err := nodepoolutils.GetNodeClass(ctx, c.kubeClient, nodePool, c.cloudProvider)
+	if nodeClass == nil {
```
jonathan-innis, line 57: nit: consider adding the same comment here

pkg/controllers/nodepool/registrationhealth/controller.go
```diff
@@ -0,0 +1,95 @@
...
+		return reconcile.Result{}, nil
+	}
+	if err != nil {
```
jonathan-innis, line 60: nit: error check needs to come before the NodeClass nil check

pkg/controllers/nodepool/readiness/controller.go
```diff
@@ -55,15 +53,10 @@ func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reco
...
 		return reconcile.Result{}, nil
 	}
-
-	err := c.kubeClient.Get(ctx, client.ObjectKey{Name: nodePool.Spec.Template.Spec.NodeClassRef.Name}, nodeClass)
 	if client.IgnoreNotFound(err) != nil {
```
jonathan-innis, line 60: nit: error check needs to come before the NodeClass nil check

pkg/controllers/state/cluster.go
```diff
@@ -416,6 +448,7 @@ func (c *Cluster) ClearPodSchedulingMappings(podKey types.NamespacedName) {
 	c.podAcks.Delete(podKey)
 	c.podsSchedulableTimes.Delete(podKey)
 	c.podsSchedulingAttempted.Delete(podKey)
+	c.podHealthyNodePoolScheduledTime.Delete(podKey)
```
jonathan-innis, line 451: This looks to me that we are keeping this in our cluster state forever if the pod isn't deleted -- I think we should throw this element away once the pod goes ready -- are we handling that correctly in the metric controller?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -343,10 +343,25 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
...
 	results.Record(ctx, p.recorder, p.cluster)
 	return results, nil
 }
 
+// MarkPodSchedulingDecisionsNodeRegistrationHealthy iterates through the nodeClaims
+// and existing nodes and marks podHealthyNodePoolScheduledTime time for pods
+func (p *Provisioner) MarkPodSchedulingDecisionsNodeRegistrationHealthy(ctx context.Context, results scheduler.Results) {
+	now := time.Now()
+	for _, n := range results.NewNodeClaims {
+		p.cluster.UpdatePodHealthyNodePoolScheduledTime(ctx, n.Labels[v1.NodePoolLabelKey], now, n.Pods...)
```
jonathan-innis, line 356: nit: Do we need to pass the time through? If we are marking it, can we just assume that we are talking about `time.Now()`

pkg/controllers/state/cluster.go
```diff
@@ -384,6 +387,26 @@ func (c *Cluster) MarkPodSchedulingDecisions(podErrors map[*corev1.Pod]error, po
...
+// for pods scheduled against nodePool that have NodeRegistrationHealthy=true
+func (c *Cluster) UpdatePodHealthyNodePoolScheduledTime(ctx context.Context, nodePoolName string, schedulableTime time.Time, pods ...*corev1.Pod) {
+	nodePool := &v1.NodePool{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err == nil {
+		for _, p := range pods {
+			nn := client.ObjectKeyFromObject(p)
+			// If the pod is scheduled to a nodePool and if the nodePool has NodeRegistrationHealthy=true
+			// then mark the time when we thought it can schedule to now.
+			if nodePool.StatusConditions().IsTrue(v1.ConditionTypeNodeRegistrationHealthy) {
+				c.podHealthyNodePoolScheduledTime.Store(nn, schedulableTime)
```
jonathan-innis, line 400: This is going to continually overwrite values that are already stored, right? I think we need to check that the value isn't already there?

pkg/controllers/state/suite_test.go
```diff
@@ -99,6 +99,26 @@ var _ = AfterEach(func() {
...
+var _ = Describe("Pod Healthy NodePool", func() {
+	It("should not store pod schedulable time if the nodePool that pod is scheduled to does not have NodeRegistrationHealthy=true", func() {
+		pod := test.Pod()
+		ExpectApplied(ctx, env.Client, pod, nodePool)
+
+		cluster.UpdatePodHealthyNodePoolScheduledTime(ctx, nodePool.Name, time.Now(), []*corev1.Pod{pod}...)
+		setTime := cluster.PodSchedulingSuccessTimeRegistrationHealthyCheck(client.ObjectKeyFromObject(pod))
+		Expect(setTime.IsZero()).To(BeTrue())
+	})
+	It("should store pod schedulable time if the nodePool that pod is scheduled to has NodeRegistrationHealthy=true", func() {
```
jonathan-innis, line 111: I think we need another test that ensures we don't overwrite the time if it's already stored AND that we properly clean things up when the pod is deleted

pkg/controllers/provisioning/suite_test.go
```diff
@@ -245,6 +245,34 @@ var _ = Describe("Provisioning", func() {
...
+		ExpectApplied(ctx, env.Client, nodePool)
+		pod := test.UnschedulablePod()
+		ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+		nodes := &corev1.NodeList{}
+		Expect(env.Client.List(ctx, nodes)).To(Succeed())
+		Expect(len(nodes.Items)).To(Equal(1))
+		ExpectScheduled(ctx, env.Client, pod)
+		Expect(cluster.PodSchedulingSuccessTimeRegistrationHealthyCheck(client.ObjectKeyFromObject(pod)).IsZero()).To(BeFalse())
+	})
+	It("should not mark podHealthyNodePoolScheduledTime if it is scheduled against a nodePool with NodeRegistrationHealthy=False", func() {
```
jonathan-innis, line 262: What about Unknown?

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
jigisha620: We already have this - `. "sigs.k8s.io/karpenter/pkg/test/expectations"`

jigisha620: Discussed offline.

pkg/controllers/state/cluster.go
jigisha620: We flush this element here - https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/metrics/pod/controller.go#L329

pkg/controllers/state/cluster.go
```diff
@@ -384,6 +387,26 @@ func (c *Cluster) MarkPodSchedulingDecisions(podErrors map[*corev1.Pod]error, po
...
+// for pods scheduled against nodePool that have NodeRegistrationHealthy=true
+func (c *Cluster) UpdatePodHealthyNodePoolScheduledTime(ctx context.Context, nodePoolName string, pods ...*corev1.Pod) {
+	nodePool := &v1.NodePool{}
+	if err := c.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err == nil {
+		for _, p := range pods {
+			nn := client.ObjectKeyFromObject(p)
+			// If the pod is scheduled to a nodePool and if the nodePool has NodeRegistrationHealthy=true
+			// then mark the time when we thought it can schedule to now.
+			if nodePool.StatusConditions().IsTrue(v1.ConditionTypeNodeRegistrationHealthy) {
+				c.podHealthyNodePoolScheduledTime.LoadOrStore(nn, time.Now())
```
jonathan-innis, line 400:
```suggestion
				c.podHealthyNodePoolScheduledTime.LoadOrStore(nn, c.clock.Now())
```

nit: Use the clock interface instead of time -- it's the standard we are using and will make testing easier

pkg/controllers/state/suite_test.go
```diff
@@ -99,6 +99,53 @@ var _ = AfterEach(func() {
...
+		nodePool.StatusConditions().SetTrue(v1.ConditionTypeNodeRegistrationHealthy)
+		ExpectApplied(ctx, env.Client, pod, nodePool)
+
+		// This will store the pod schedulable time
+		cluster.UpdatePodHealthyNodePoolScheduledTime(ctx, nodePool.Name, []*corev1.Pod{pod}...)
+		setTime := cluster.PodSchedulingSuccessTimeRegistrationHealthyCheck(client.ObjectKeyFromObject(pod))
+		Expect(setTime.IsZero()).To(BeFalse())
+
+		// We try to update pod schedulable time, but it should not change as we have already stored it
+		cluster.UpdatePodHealthyNodePoolScheduledTime(ctx, nodePool.Name, []*corev1.Pod{pod}...)
```
jonathan-innis, line 131: If you update to use the clock interface, you'll have to make an update to this test here

jonathan-innis (approved):
/lgtm
/approve

PR #1973, opened 2025-02-07
perf: Remove using `Difference()` from set creation

jonathan-innis:
Fixes #N/A 

**Description**

Using the `Difference` function results in us creating a new set when we could have just iterated through the map and read the existing values

`make presubmit`


jigisha620 (approved):
/lgtm
/approve

engedaam (commented):
/lgtm

PR #1989, opened 2025-02-13
docs: RFC for configurable registration TTL (MaxNodeProvisionTime)

### LINKED ISSUE #357: Allow registration TTL to be user configurable

At this time registration TTL is hard code to 15 minutes. It would be greatly appreciated if we can make this configurable by the user.

At this time registration TTL is hard code, https://github.com/aws/karpenter-core/blob/main/pkg/controllers/machine/lifecycle/liveness.go#L38, however some of our nodes takes close to and some times over the 15 minute time window to run -> bootstrap -> join the cluster. This means that these nodes will be terminated after 15 minutes even though they would have been able to join the cluster. 

Since this isn't user configurable, we have the fork the code and build our own image for Karpenter to bypass this limitation.

### Additional Context

https://kubernetes.slack.com/archives/C02SFFZSA2K/p1685123613509889 Conversation from slack

### Attachments

_No response_

### Community Note

GnatorX:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/357

**Description**
RFC for configurable registerationTTL

NA


GnatorX:
@jmdeal Updated the doc lmk if we want more info

GnatorX:
Chatting during working hours.
- We think fully configurable is likely fine
- We wondered if moving this to Node overlay make sense. Currently, we are not sure however it doesn't "feel" right to wrap this with overlay atm.

daimaxiaxie:
Vote for Option 2.

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,35 @@
...
+spec:
+  registerationTTL:  30m
+```
+
+#### Evaluation conditions 
+
+1. When evaluating liveness, check if NodeClaim contains a registerationTTL if so utilize that TTL instead of the default
+
+### Option 2:
+#### Introduce a configuration registeration TTL to Node Pool then propagate to node claim
```
engedaam: I'm inclined to favor Option 2 for configuring the registrationTTL, but there are nuanced considerations to weigh. On one hand, setting registrationTTL on the NodeClass could streamline configuration, as it's the key property affecting node registration time and a single NodeClass can apply to multiple NodePools. This approach would reduce the overall configuration surface area. However, it's important to recognize that registrationTTL isn't inherently a node property, but rather a directive for Karpenter's behavior. This characteristic aligns it more closely with other Karpenter-specific settings like disruption controls, termination grace periods, and expireAfter configurations. These parameters all govern how Karpenter manages node lifecycles within the cluster. From an architectural standpoint, maintaining Karpenter behavior configurations within the NodePool specification could provide a cleaner separation of concerns. It would keep node-specific properties distinct from operational behaviors, potentially leading to a more intuitive and maintainable system design. Ultimately, the decision should balance configuration simplicity with logical organization of cluster management parameters. What are your thoughts on these trade-offs?

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,35 @@
...
+[Issue](https://github.com/kubernetes-sigs/karpenter/issues/357)
+
+Karpenter currently hardcodes [registeration TTL](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/liveness.go#L44) which means nodes must take < 15 minutes to come up and register with the Kubernetes control plane.
+
+This is problematic because some users have nodes that take longer that time to register. As mentioned in the issue, nodes that run GPU can often take > 15 minutes to start up. 
+
+
+## Configuration
+### Option 1 (Preferred):
+#### Introduce a configuration registeration TTL to NodeClass and propagate it to the NodeClaim
```
engedaam: With the introduction of this field, How are you thinking drift should be applied? Should it be hash based or dynamically resolved? Should this be left to the cloud providers?

designs/configurable-registeration-ttl.md
GnatorX: I see, i think i follow your reasoning and happy to switch it over to NodePool

designs/configurable-registeration-ttl.md
GnatorX: Chatted offline. We don't think its necessarily to Drift on this because this only happens at registration time and the node either would fail to register within the default TTL -> node gets terminated or it did join within the TTL and the variable won't be useful to drift against

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,35 @@
...
+
+#### Evaluation conditions 
+
+1. When evaluating liveness, check if NodeClaim contains a registerationTTL if so utilize that TTL instead of the default
+
+### Option 2 (Preferred):
+#### Introduce a configuration registeration TTL to Node Pool then propagate to node claim
+
+We would introduce the new argument in Node Pool and perform similar propagation of the argument to Node Claim. In addition, we won't consider this variable when evaluating drift as 
+this variable is only used duration registration. If a node needed a longer TTL then the node would have been killed by Karpenter and if it did register then the TTL didn't matter, so a change to this variable wouldn't require Drift to update.
```
engedaam: Discussed offline to add a few more details on the stance we are going to take updates for unregisterd nodeclaim that are out of sync with the nodepool.

GnatorX:
Note i swapped the options number (2 <-> 1)

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,40 @@
...
+### Option 1 (Preferred):
+#### Introduce a configuration registeration TTL to Node Pool then propagate to node claim
+
+We propose introducing a new argument in the Node Pool and propagating it to the NodeClaim during its creation. When the liveness controller evaluates the registrationTTL, it will use the value specified in the NodeClaim.
+
+Importantly, even if the registrationTTL is updated on the NodePool during NodeClaim registration, we will continue to use the NodeClaim's registrationTTL. This approach ensures clarity by demonstrating that the NodeClaim relies on variables defined in its own spec.
+
+Furthermore, this variable will not be considered when evaluating drift, as it is only utilized during the registration process. If a node requires a longer TTL, it would have already been terminated by Karpenter. In cases where the node successfully registers, the TTL becomes irrelevant. Consequently, changes to this variable do not necessitate updates to the drift evaluation process.
+
+Lastly, we will move registrationTTL to fully be in NodePool CRD and default it to 15m.
```
engedaam: Can you just add one quick example of what the field will look in the nodepool schema?

GnatorX: added an example

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
...
+      expireAfter: 288h0m0s
+      registrationTTL: 20m # New variable
+```
+
+#### Concerns
+
+N/A
+
+### Option 2:
+#### Introduce a configuration registeration TTL to NodeClass and propagate it to the NodeClaim
```
jonathan-innis:
```suggestion
#### Introduce a configuration registration TTL to NodeClass and propagate it to the NodeClaim
```

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
...
+  disruption: {}
+    consolidateAfter: 5m
+    consolidationPolicy: WhenEmptyOrUnderutilized
+  limits:
+    nodes: "800"
+  template:
+    metadata: {}
+    spec:
+      expireAfter: 288h0m0s
+      registrationTTL: 20m # New variable
```
jonathan-innis: nit: We don't have TTL really in our lingo for Karpenter -- is there a way to define this as `....After` -- maybe try ChatGPT and see what it can come up with 😆

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+#### Concerns
+
+N/A
+
+### Option 2:
+#### Introduce a configuration registeration TTL to NodeClass and propagate it to the NodeClaim
+
+We will keep a default for all nodes at 15 minute TTL however in the cases that the NodeClass would be used with GPUs or known to take a longer time to register, we will expose a new configuration registerationTTL. We will allow users to configure their own TTL without presets
```
jonathan-innis:
```suggestion
We will keep a default for all nodes at 15 minute TTL however in the cases that the NodeClass would be used with GPUs or known to take a longer time to register, we will expose a new configuration regisrationTTL. We will allow users to configure their own TTL without presets
```

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
...
+
+#### Evaluation conditions 
+
+1. When evaluating liveness, check if NodeClaim contains a registerationTTL if so utilize that TTL instead of the default.
```
jonathan-innis:
```suggestion
1. When evaluating liveness, check if NodeClaim contains a registrationTTL if so utilize that TTL instead of the default. 
```

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
...
+
+[Issue](https://github.com/kubernetes-sigs/karpenter/issues/357)
+
+Karpenter currently hardcodes [registeration TTL](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/liveness.go#L44) which means nodes must take < 15 minutes to come up and register with the Kubernetes control plane.
+
+This is problematic because some users have nodes that take longer that time to register. As mentioned in the issue, nodes that run GPU can often take > 15 minutes to start up. 
+
+## Configuration
+
+### Option 1 (Preferred):
```
jonathan-innis, line 20: Did we consider a NodeOverlay option here? Not sure that it's the right option, but it should be an option that we should think about

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
+# RFC: Configurable registeration TTL
+
+## Motivation
+
+[Issue](https://github.com/kubernetes-sigs/karpenter/issues/357)
+
+Karpenter currently hardcodes [registeration TTL](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/liveness.go#L44) which means nodes must take < 15 minutes to come up and register with the Kubernetes control plane.
```
jonathan-innis:
```suggestion
Karpenter currently hardcodes [registration TTL](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/liveness.go#L44) which means nodes must take < 15 minutes to come up and register with the Kubernetes control plane.
```

akestner:
Hey @GnatorX, what ends up taking the majority of the time for the instances that take >15m to boot?

GnatorX:
If you look https://github.com/kubernetes-sigs/karpenter/issues/357 there are a few cases there. For myself it is GPU bootstrapping due to having to install drivers

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,57 @@
...
+
+This is problematic because some users have nodes that take longer that time to register. As mentioned in the issue, nodes that run GPU can often take > 15 minutes to start up.
```
engedaam, line 9: Can we outline all the different use cases in this doc? This would be good for posterity

GnatorX: updated

jonathan-innis:
Have we considered smarter approaches here where the cloudprovider passes down some of this information based on the known startup time of instance types? It may be tougher to do with AMIs (especially if they are custom) but perhaps we just leave a bigger buffer for custom AMIs and then we can control shorter registrationTTLs from the CloudProvider side to prevent making this another value that gets user-configured

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,60 @@
+# RFC: Configurable registeration TTL
+
+## Motivation
+
+[Issue](https://github.com/kubernetes-sigs/karpenter/issues/357)
+
+Karpenter currently hardcodes [registeration TTL](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/liveness.go#L44) which means nodes must take < 15 minutes to come up and register with the Kubernetes control plane.
```
Bryce-Soghigian: Do we know what instances are taking that long? Do you have a list of instance types that frequently hit this TTL?  We ran into some issues before but improved our latency so we don't hit that anymore.

designs/configurable-registeration-ttl.md
GnatorX: Here are some:
- startupDuration
- registrationTimeout
- registrationDeadline
- registrationFailAfter
- registrationGracePeriod

designs/configurable-registeration-ttl.md
GnatorX: I don't have a specific set of instance types unfortunately

GnatorX:
> Have we considered smarter approaches here where the cloudprovider passes down some of this information based on the known startup time of instance types? It may be tougher to do with AMIs (especially if they are custom) but perhaps we just leave a bigger buffer for custom AMIs and then we can control shorter registrationTTLs from the CloudProvider side to prevent making this another value that gets user-configured

I think we kind of want both where maybe cloud provider can provide a default and user can override if necessary?

Bryce-Soghigian:
Cluster autoscaler calls it maxNodeProvisionTime
```
maxNodeProvisionTime      = flag.Duration("max-node-provision-time", 15*time.Minute, "The default maximum time CA waits for node to be provisioned - the value can be overridden per node group") 
```
Ive seen some customers set this to a lower number as well, to abandon P99 latency scale calls and re-roll so to speak for large scale ups. Not sure how effective that really is. 

For cluster autoscaler it's done with a global flag option, with per node-group override being added after some customer complaints about it only being configurable per cluster.

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,65 @@
...
+  creationTimestamp: "2025-02-12T17:59:44Z"
+  name: example-nodepool
+spec:
+  disruption: {}
+    consolidateAfter: 5m
+    consolidationPolicy: WhenEmptyOrUnderutilized
+  limits:
+    nodes: "800"
+  template:
+    metadata: {}
```
Bryce-Soghigian, line 38: nit: maybe a nodepool stub is enough to showcase the actual field of interest. We don't need to show the annotations, node limits etc.

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,65 @@
...
+- GPU driver installation during node bootstrap
+- Windows AMIs security tooling installation 
+
+## Configuration
+
+### Option 1 (Preferred):
+#### Introduce a configuration registeration TTL to Node Pool then propagate to node claim
+
+We propose introducing a new argument in the Node Pool and propagating it to the NodeClaim during its creation. When the liveness controller evaluates the registrationTTL, it will use the value specified in the NodeClaim.
+
```
Bryce-Soghigian: Did we consider a cloudprovider default where the cloudprovider sets instance types into a bucket(GPUs get longer window), or specific times for specific instance types?

designs/configurable-registeration-ttl.md
Bryce-Soghigian: thoughts on having parity with the cluster autosclaer flag name? MaxNodeProvisionTime?

GnatorX: That sounds good to me

designs/configurable-registeration-ttl.md
GnatorX: Yes, I think this make sense however it isn't scoped here because we think in general we can do better on exposing cloud provider defaults in Karpenter

designs/configurable-registeration-ttl.md
GnatorX: Added option 3

designs/configurable-registeration-ttl.md
GnatorX: cleaned up

Bryce-Soghigian (approved):
1. Flag options for configuration
Agree with the decision we came to that it should live in the nodepool spec.

That being said, we should talk maybe about tradeoffs between the different design options.

Cluster level flag: Cluster Autoscaler has this, and it wasn't enough, users ended up adding a nodepool level configurable flag.

Nodepool level flag: As different nodepools have different expectations and requirements around registration, nodepool level makes more sense and avoids forcing one-size-fits-all behavior.

Cluster Level Flag w Nodepool Override: Also an option. 

(I know you mentioned cloudprovider override out of scope, but i would mention that as the fourth option, with the fifth being node overlay) 


2. User stories (why this is needed)
Some notes from my experience

1. As different nodepools have different expectations and requirements around registration. I have seen customers set this value to 45m and to 5m, both extending provisioning time to give long P99 requests longer to finish without abandoning the call, and then separately, abandoning requests that are slower than we see on average and re-rolling. These are two things we could include as user stories for why this is needed, outside of standard GPU provisioning + bootstrapping times.

2. Also, different cloudproviders have different performance expectations. 15 minutes seems reasonable for most instance types + images for Azure + AWS, but maybe there are other cloudproviders that have different performance expectations.

3. Sometimes the node registration failures come from a source outside of Karpenter’s control. For example the kubelet is getting throttled by the apiserver, due to APIServer QPS. Deleting the underlying VM will not make the scale up goal faster in this case, it does the opposite. Some customers are willing to pay for a bigger apiserver, some customers don't have that option. Either way, it makes sense that we would want to prevent this VM GC if it’s not the VM provisioning that’s the bottleneck in registration, but rather the apiserver. Deleting and recreating adds additional pressure and churn to the apiserver that's already overloaded. This is yet another reason to let that registration be configurable by the user.

Bryce-Soghigian:
i dont have any sort of approval perms, but i agree that going through with a nodepool level override meets all the user stories i can think of for this type of feature.

GnatorX:
Thanks for the comment! Let me add some of those into the doc

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,59 @@
+# RFC: Configurable MaxNodeProvisionTime
```
ellistarn, line 1: Now that we have auto repair, do we even need a max registration time? Could we rely on repair semantics to remediate issues with registration, and otherwise allow registration to be indefinite (or much longer ~1h?)

GnatorX: I think "registration" is a different class of problem because during this time the node won't have "status" since it isn't registered to the K8s control plane. I am not sure if we can rely on repair semantics for handling registration issues.

designs/configurable-registeration-ttl.md
```diff
@@ -0,0 +1,83 @@
...
+## Motivation
+
+[Issue](https://github.com/kubernetes-sigs/karpenter/issues/357)
+
+Karpenter currently hardcodes [registration TTL](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/liveness.go#L44) which means nodes must take < 15 minutes to come up and register with the Kubernetes control plane.
+
+This is problematic because some users have nodes that take longer that time to register. As mentioned in the issue, nodes that run GPU can often take > 15 minutes to start up. 
+Use cases:
+- GPU driver installation during node bootstrap
+- Windows AMIs security tooling installation
```
ellistarn, line 12: Is it necessary to prevent the kubelet from registering in these cases? Maybe this is a better use case for startup taints?

GnatorX: Chatted with @DerekFrank and I think we agree with this (startup taint) Will update with a slightly different RFC for this.

GnatorX:
Actually, i want to change this PR to still be configurable TTL but limit it to 15 minutes and allow this to config the ttl down (lower than 15 minutes) and i will have a second [PR](https://github.com/kubernetes-sigs/karpenter/pull/2693) for readiness TTL using node repair. 

@DerekFrank @ellistarn if this make sense

ellistarn:
> Actually, i want to change this PR to still be configurable TTL but limit it to 15 minutes and allow this to config the ttl down (lower than 15 minutes) and i will have a second [PR](https://github.com/kubernetes-sigs/karpenter/pull/2693) for readiness TTL using node repair.

What happened to https://github.com/kubernetes-sigs/karpenter/pull/1989#discussion_r2402737305

GnatorX:
That is still something I want to do. The thought here is allowing users to reduce registration ttl which isn't the same use case as extending the ttl?

designs/configurable-registeration-ttl.md
ellistarn, line 12: This tooling should be baked into your operating system image. Installing this at runtime seems like a wildly expensive operation to repeat on every node launch.

PR #1990, opened 2025-02-14
ci: fix flaky test for stale instance type drift

### LINKED ISSUE #1986: Drift Test is Flaking
### Description

**Observed Behavior**:

`Drift [It] should detect stale instance type drift if the instance type offerings aren't compatible with the nodeclaim` test is flaking here and there throughout on different test runs.

See https://github.com/kubernetes-sigs/karpenter/actions/runs/13260477160/attempts/1?pr=852. There are other examples if you look through different [presubmit runs](https://github.com/kubernetes-sigs/karpenter/actions/workflows/presubmit.yaml)

**Expected Behavior**:

Test shouldn't flake

Run `make deflake` with FOCUS on `should detect stale instance type drift if the instance type offerings aren't compatible with the nodeclaim` and you should eventually be able to repro the flake

flavono123:
Fixes #1986 

**Description**

1. change Offering.Requirements of InstanceType over Requirements for setup test; ofs.HasCompatible compare Offering.Requirements with the argument(NodeClaim.Labels in this case) -> so i don't know how the test was passed sometimes
2. fix the key for changing value to Offering.Requirements concerns since a return of lo.Keys is not sorted, the previous key was random.
  a. just one changing is enough so commented another  

chore: remove trailing whitespaces in Makefile

check the test case is passed on make deflake FOCUS="should detect stale instance type drift if the instance type offerings aren't compatible with the nodeclaim"

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/drift_test.go
+++ b/pkg/controllers/nodeclaim/disruption/drift_test.go
@@ -115,7 +115,6 @@ var _ = Describe("Drift", func() {
 				for i := range it.Offerings {
 					it.Offerings[i].Requirements = scheduling.NewLabelRequirements(map[string]string{
 						corev1.LabelTopologyZone: test.RandomName(),
-						// v1.CapacityTypeLabelKey:  test.RandomName(),
 					})
 				}
 			}
--- a/pkg/controllers/nodeclaim/disruption/drift_test.go
+++ b/pkg/controllers/nodeclaim/disruption/drift_test.go
@@ -112,8 +112,12 @@ var _ = Describe("Drift", func() {
 	It("should detect stale instance type drift if the instance type offerings aren't compatible with the nodeclaim", func() {
 		cp.InstanceTypes = lo.Map(cp.InstanceTypes, func(it *cloudprovider.InstanceType, _ int) *cloudprovider.InstanceType {
 			if it.Name == nodeClaim.Labels[corev1.LabelInstanceTypeStable] {
-				newLabels := lo.Keys(nodeClaim.Labels)
-				it.Requirements = scheduling.NewLabelRequirements(map[string]string{newLabels[0]: test.RandomName()})
+				for i := range it.Offerings {
+					it.Offerings[i].Requirements = scheduling.NewLabelRequirements(map[string]string{
+						corev1.LabelTopologyZone: test.RandomName(),
+						// v1.CapacityTypeLabelKey:  test.RandomName(),
+					})
+				}
 			}
 			return it
 		})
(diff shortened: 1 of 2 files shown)


pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -112,8 +112,12 @@ var _ = Describe("Drift", func() {
 	It("should detect stale instance type drift if the instance type offerings aren't compatible with the nodeclaim", func() {
 		cp.InstanceTypes = lo.Map(cp.InstanceTypes, func(it *cloudprovider.InstanceType, _ int) *cloudprovider.InstanceType {
 			if it.Name == nodeClaim.Labels[corev1.LabelInstanceTypeStable] {
-				newLabels := lo.Keys(nodeClaim.Labels)
-				it.Requirements = scheduling.NewLabelRequirements(map[string]string{newLabels[0]: test.RandomName()})
+				for i := range it.Offerings {
```
jonathan-innis, line 115: Oh! This actually isn't good at all -- looks like the default state that we set the generic cloudprovider to is `"drifted"` after a `Reset()` which means that regardless of whether the test _should_ actually pass for the drifted condition that it's checking, it's still going to return drifted at all times in the cloudprovider reason -- probably best to check the reason for "Drifted" in this test as well (in the status condition) so that we avoid cases like this

jonathan-innis: We should absolutely update that cloudprovider.Reset() function though to not return "drifted" by default

flavono123: it sounds like a former work required. feel free to amend or just let me know after that done.

i will take another one from open issues.

pkg/controllers/nodeclaim/disruption/drift_test.go
```diff
@@ -112,8 +112,12 @@ var _ = Describe("Drift", func() {
...
+					it.Offerings[i].Requirements = scheduling.NewLabelRequirements(map[string]string{
+						corev1.LabelTopologyZone: test.RandomName(),
+						// v1.CapacityTypeLabelKey:  test.RandomName(),
```
jonathan-innis, line 118: Do you need this comment here if you aren't using this portion?

pkg/controllers/nodeclaim/disruption/drift_test.go
jonathan-innis: Just pushed the change -- one comment, then I think this looks like it's good to go

pkg/controllers/nodeclaim/disruption/drift_test.go
flavono123: removed -- the incompatible offering might be split into a topology zone and a capacity for each, but still enough for now

jonathan-innis (approved):
/lgtm
/approve

PR #1991, opened 2025-02-14
fix: remove registration taint enforcement

rschalo:
Fixes #N/A 

**Description**
When upgrading from pre-v1 to v1 versions of Karpenter using gitops, coordination of the resource and controller updates can result in an undesirable chicken and egg problem where:
- the controller is updated before Karpenter resources and the prevents new capacity from launching because the NodeClass has not added the unregistered taint in userData.
- the resources are updated before the controller and not only does Drift occur but then application pods will not schedule unless those deployments are updated with a toleration for the registration taint.

The fix is to remove the requirement that the unregistered taint be present, and instead make users aware of the race condition. By removing the requirement, the controller can be upgraded, pending pods can schedule to new capacity, and then the NodeClass userData can be updated to prevent the race condition.


rschalo (commented):
/snapshot

rschalo (commented):
/karpenter snapshot

pkg/controllers/nodeclaim/lifecycle/events.go
```diff
@@ -35,6 +35,16 @@ func InsufficientCapacityErrorEvent(nodeClaim *v1.NodeClaim, err error) events.E
 	}
 }
 
+func NodeClassUnregisteredTaintMissingEvent(nodeClaim *v1.NodeClaim, err error) events.Event {
+	return events.Event{
+		InvolvedObject: nodeClaim,
+		Type:           corev1.EventTypeWarning,
+		Reason:         events.NodeClassUnregisteredTaintMissing,
+		Message:        fmt.Sprintf("NodeClaim %s event: %s", nodeClaim.Name, truncateMessage(err.Error())),
```
jonathan-innis, line 43: I'm not sure how much we really need to prepend with "this is a NodeClaim event" -- that should be obvious based on the fact that we are firing an event that is attached to the NodeClaim

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -18,26 +18,29 @@ package lifecycle
...
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	coreerrors "k8s.io/apimachinery/pkg/api/errors"
```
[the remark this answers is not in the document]
rschalo: The addition of `errors.New` on :72

pkg/controllers/nodeclaim/lifecycle/initialization_test.go
```diff
@@ -133,30 +133,34 @@ var _ = Describe("Initialization", func() {
 		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
 		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 
-		node := test.Node(test.NodeOptions{
+		node1 := test.Node(test.NodeOptions{
```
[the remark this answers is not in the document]
jonathan-innis: Discussed offline: This was changed because we still need a way to set Registration to `False` and the only other way while still registering the Node is by creating two nodes with the same providerID

pkg/controllers/nodeclaim/lifecycle/initialization_test.go
```diff
@@ -133,30 +133,34 @@ var _ = Describe("Initialization", func() {
...
+		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+		ExpectMakeNodesReady(ctx, env.Client, node1, node2) // Remove the not-ready taint
 
 		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 		Expect(ExpectStatusConditionExists(nodeClaim, v1.ConditionTypeRegistered).Status).To(Equal(metav1.ConditionFalse))
 		Expect(ExpectStatusConditionExists(nodeClaim, v1.ConditionTypeInitialized).Status).To(Equal(metav1.ConditionUnknown))
 
-		node = ExpectExists(ctx, env.Client, node)
-		node.Status.Capacity = corev1.ResourceList{
+		node1 = ExpectExists(ctx, env.Client, node1)
```
jonathan-innis, line 152: If you really want to make sure that this test is valid, I think you need to do the same thing for `node2` -- basically build something that when removing the check that we have for Registration at the top of Initialization would successfully Initialize without that check

pkg/controllers/nodeclaim/lifecycle/termination_test.go
```diff
@@ -360,18 +360,22 @@ var _ = Describe("Termination", func() {
 		_, err := cloudProvider.Get(ctx, nodeClaim.Status.ProviderID)
 		Expect(err).ToNot(HaveOccurred())
 
-		node := test.NodeClaimLinkedNode(nodeClaim)
-		// Remove the unregistered taint to ensure the NodeClaim can't be marked as registered
-		node.Spec.Taints = nil
-		ExpectApplied(ctx, env.Client, node)
-		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
+		// Create multiple nodes that are linked to the same NodeClaim
+		node1 := test.NodeClaimLinkedNode(nodeClaim)
```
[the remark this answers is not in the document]
jonathan-innis: I don't really think that you need to create two nodes here -- you can just create a NodeClaim that has only set the Launched condition, generate a node with the same providerID and then issue the terminate

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -62,11 +64,11 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
 		return t.MatchTaint(&v1.UnregisteredNoExecuteTaint)
 	})
-	// check if sync succeeded but setting the registered status condition failed
-	// if sync succeeded, then the label will be present and the taint will be gone
+	// if the sync hasn't happened yet and the race protecting startup taint isn't present then log it as missing and proceed
+	// if the sync has happened then the startup taint has been removed if it was present
 	if _, ok := node.Labels[v1.NodeRegisteredLabelKey]; !ok && !hasStartupTaint {
-		nodeClaim.StatusConditions().SetFalse(v1.ConditionTypeRegistered, "UnregisteredTaintNotFound", fmt.Sprintf("Invariant violated, %s taint must be present on Karpenter-managed nodes", v1.UnregisteredTaintKey))
-		return reconcile.Result{}, fmt.Errorf("missing required startup taint, %s", v1.UnregisteredTaintKey)
+		log.FromContext(ctx).Error(nil, fmt.Sprintf("missing %s taint which prevents registration related race conditions on Karpenter-managed nodes", v1.UnregisteredTaintKey))
```
jonathan-innis, line 70: This prints a nil error in the log messaging -- which is sort of odd -- I do think what you've created here is an error -- I think you just need to find another sensible message to put next to that error

pkg/controllers/nodeclaim/lifecycle/termination_test.go
```diff
@@ -353,25 +353,14 @@ var _ = Describe("Termination", func() {
...
-		Expect(err).ToNot(HaveOccurred())
-
 		node := test.NodeClaimLinkedNode(nodeClaim)
-		// Remove the unregistered taint to ensure the NodeClaim can't be marked as registered
-		node.Spec.Taints = nil
-		ExpectApplied(ctx, env.Client, node)
-		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
-		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
-		Expect(nodeClaim.StatusConditions().Get(v1.ConditionTypeRegistered).IsFalse()).To(BeTrue())
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
```
jonathan-innis, line 357: IMO: You should launch the NodeClaim at least but just not register it -- so have one reconciliation for the launch but don't do the other

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -219,6 +219,7 @@ func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (rec
 			return reconcile.Result{}, fmt.Errorf("ensuring instance termination, %w", err)
 		}
 		if !isInstanceTerminated {
+			fmt.Println("no error terminating but we didn't terminate?")
```
jonathan-innis, line 222: I think you left this behind

jonathan-innis (approved):
/lgtm
/approve

PR #2010, opened 2025-02-19
test: increase Eventually timeout in e2e tests

### LINKED ISSUE #2007: Allow platform agnostic e2e tests to ignore existing NodeClass and allow configurable e2e test timeout
### Description

I am looking into trying to integrate the platform agnostic upstream e2e tests with our karpenter provider implementation and re-use the tests and thankfully, we noticed [this PR was merged recently](https://github.com/kubernetes-sigs/karpenter/pull/1963) that allowed cloudproviders to support generic nodepools and nodeclasses for these e2e tests.

But I run into an issue because, in our provider implementation, the clusters using karpenter on our k8s platform are not in control of the NodeClass CR and a controller outside of the cluster reconciles/clobbers most changes to the object including deletions. This unfortunately means our tests fail out of the box since the tests expect the cluster to be "clean" (no nodeclass/nodepool/nodeclaim exists) and because they also attempt to create a "default" NodeClass before every test. I'm wondering if it's viable for upstream to relax those steps in the karpenter core e2e suites, and if it makes sense for me to submit some patch for that. My argument for allowing this is, these core tests should probably not test different NodeClass implementations and configurations (the provider tests should) and therefore, don't need a strict restriction/check on enforcing a "default" NodeClass during test execution.

Additionally, I've noticed that running the tests with our cloud provider requires a longer `Eventually` timeout when waiting for pods to schedule and nodes to provision in order to pass ([as of writing this, the default is 5 minutes](https://github.com/kubernetes-sigs/karpenter/blob/37452822fcfb14b9db7121c7ee4a4440c275b8f5/test/pkg/environment/common/environment.go#L94)). Since provider implementations are potentially not using fake KWOK nodes which are must faster to spin up, I think it may be worth it to be able to configure the default timeout during test execution time as well.

It will help cloud providers that manage NodeClasses for the user be able to test the implementation's core karpenter functionality with better first-class support.  

maxcao13:
Fixes #2007

**Description**
For now, this PR bumps up the default `Eventually` timeout to 10 minutes. E2E tests may fail if the cloud provider cannot finish provisioning an instance within the previous low 5 minute default.

Running the e2e tests in a local kind cluster.

### DIFF
--- a/test/pkg/environment/common/environment.go
+++ b/test/pkg/environment/common/environment.go
@@ -91,7 +91,7 @@ func NewEnvironment(t *testing.T) *Environment {
 	// Get the output dir if it's set
 	outputDir, _ := os.LookupEnv("OUTPUT_DIR")
 
-	gomega.SetDefaultEventuallyTimeout(5 * time.Minute)
+	gomega.SetDefaultEventuallyTimeout(10 * time.Minute)
 	gomega.SetDefaultEventuallyPollingInterval(1 * time.Second)
 	return &Environment{
 		Context:               ctx,


maxcao13:
Waiting for https://github.com/kubernetes-sigs/karpenter/issues/2007#issuecomment-2667276062 comment to be resolved. 

Hoping it's possible that if the test detects there already exists a NodeClass in the cluster, it will proceed with the tests and doesn't clean it up at the end of the suite. If the test detects there doesn't exist a NodeClass, then it will create one at the beginning of the suite, and clean it up at the end of the suite.

jonathan-innis (approved):
/lgtm
/approve

PR #2011, opened 2025-02-19
perf: check validity before constructing a new set of requirements

tzneal:
**Description**
When checking if a Pod will work on an existing Node, we validate in several ways including
checking the Pods requirements vs the Node's requirements. If it fails at any step, 
we need to leave the Node's requirements unchanged. Previously we were:

1) Creating set of temporary requirements from the current node's requirements
2) Check if the Pod is compatible with them
3) Assigning those back to the Node if all compatibility checks passed

This changes the order so that we check for compatibility with the Pod first so that 
in the case of a Pod not being compatible, we can avoid the allocation that occurs when
creating the temporary set of requirements.

Unit testing & deployed


jonathan-innis (approved):
/lgtm
/approve

PR #2012, opened 2025-02-19
perf: sort Pods by node name to allow re-use of scheduling requirements

tzneal:
**Description**

We're iterating over Pods to count the topology domains, which requires looking up Nodes and computing the Nodes requirements.  The order that we do this doesn't matter, so we can just sort the Pods by the Node name to minimize the number of times we need to fetch the Node and recompute its requirements.


jonathan-innis (approved):
/lgtm
/approve

jonathan-innis:
Nice find!

PR #2013, opened 2025-02-19
perf: trade off less verbose errors for faster execution

tzneal:
Fixes #N/A 

**Description**

Returning the first error we find is both faster and allocates less. There is a tradeoff here in that we'll only report one of the reasons that a Pod won't schedule, but I think the tradeoff for faster speed & fewer allocations is worth it.

Unit testing & deployed


tzneal:
I think this is the most contentious change.  We're changing the experience a bit, but I'm not sure how useful the previous behavior of preventing a screen full of errors is vs just a single error.

jonathan-innis (approved):
/lgtm
/approve

PR #2018, opened 2025-02-21
fix(nodeclaim): dedup logs marking consolidatable

### LINKED ISSUE #2002: `marking consolidatable` log messages are duplicated due to read-after-write inconsistency
### Description

**Observed Behavior**:

```
{"level":"DEBUG","time":"2025-02-17T17:23:51.179Z","logger":"controller","caller":"disruption/controller.go:99","message":"marking consolidatable","commit":"a79abac","controller":"nodeclaim.disruption","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"default-pwzrz"},"namespace":"","name":"default-pwzrz","reconcileID":"05fd3c4d-cad7-4979-862f-de6cee15988e","Node":{"name":"ip-192-168-106-221.us-west-2.compute.internal"}}
{"level":"DEBUG","time":"2025-02-17T17:23:51.372Z","logger":"controller","caller":"disruption/controller.go:99","message":"marking consolidatable","commit":"a79abac","controller":"nodeclaim.disruption","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"default-pwzrz"},"namespace":"","name":"default-pwzrz","reconcileID":"957af1c8-0966-47ea-be7e-8ae2697ca4df","Node":{"name":"ip-192-168-106-221.us-west-2.compute.internal"}}
```

Karpenter will double-log that a status condition has transitioned for `Consolidatable` (the same probably applies to `Drifted`) because there is no read-after-write consistency so Karpenter needs enough time after it has mutated the object to actually see the cache update -- one way to resolve this is to do an event filter on status updates -- the other way to accomplish this is by adding an artificial sleep into the Reconcile to give the cache time to catch-up like [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/controller.go#L162)

**Expected Behavior**:

Karpenter should not produce the same log twice in a row.

flavono123:
Fixes #2002 

**Description**

after set a nodeclaim is consolidatable, pause a while for a node handler does not work duplicated jobs

no, but guess it works since there are other cases referenced


pkg/controllers/nodeclaim/disruption/consolidation.go
```diff
@@ -69,9 +70,13 @@ func (c *Consolidation) Reconcile(ctx context.Context, nodePool *v1.NodePool, no
...
 	}
 
-	// 6. Otherwise, add the consolidatable status condition
-	nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeConsolidatable)
+	// 3. Otherwise, add the consolidatable status condition if not already set
 	if !hasConsolidatableCondition {
+		nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeConsolidatable)
+		// We sleep here after a set operation since we want to ensure that we are able to read our own writes
+		// so that we avoid duplicating log lines due to quick re-queues from our node watcher
+		time.Sleep(100 * time.Millisecond)
```
jonathan-innis, line 78: So, this sleep here isn't actually going to do anything. The sleep needs to come after the Patch to give the cache time to update. If we just sleep here without doing a cache update, we aren't really accomplishing the deduping that we intended

flavono123: i did change following how lifecycle controller does, you mentioned in the issue, but not sure does not duplicating logs

pkg/controllers/nodeclaim/disruption/controller.go
```diff
@@ -110,6 +111,10 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
 			}
 			return reconcile.Result{}, client.IgnoreNotFound(err)
 		}
+		// We sleep here after a patch operation since we want to ensure that we are able to read our own writes
+		// so that we avoid duplicating metrics and log lines due to quick re-queues from our node watcher
+		// USE CAUTION when determining whether to increase this timeout or remove this line
+		time.Sleep(time.Second)
```
jonathan-innis, line 117: I get that we are doing this in the lifecycle controller but looking back at this change, this sleep actually makes me a tad nervous about performance. Given that we only have 10 async threads occurring at one time, this has the potential to drastically slow down the performance of this controller.

I wonder if a better option here is to create some caching mechanism for the controller where we track how recently a key has been seen and processed and then avoid re-processing that key within a certain amount of time since we know that we just processed it

flavono123: added *cache.Cache for the controller by referencing the consistency controller does https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/consistency/controller.go#L52

plz check the dedupe time window should be tuned

PR #2025, opened 2025-02-21
perf: respect multinode consolidation timeout in all cases

rschalo:
Fixes #N/A 

**Description**
Scheduling for 50 nodes in multinode consolidation can take a long time, especially in large clusters where a scheduling decision for a node can take 20 seconds or longer. This can cause multinode consolidation to block drift, emptiness, and single node consolidation for longer than intended.

Deployed with a 5 second timeout and saw multinode consolidation bail before exhausting the list of candidates.
```
disruption/multinodeconsolidation.go:83","message":"failed to find a multi-node consolidation after timeout, last considered batch had 8 candidates
```


pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -120,42 +120,48 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
 
-		cmd, results, err := m.computeConsolidation(ctx, candidatesToConsolidate...)
-		if err != nil {
-			return Command{}, scheduling.Results{}, err
-		}
+			// ensure that the action is sensical for replacements, see explanation on filterOutSameType for why this is
+			// required
+			replacementHasValidInstanceTypes := false
+			if cmd.Decision() == ReplaceDecision {
+				log.FromContext(ctx).V(1).Info(fmt.Sprintf("evaluating replace decision, candidates: %v, replacements: %v", len(lastSavedCommand.candidates), len(lastSavedCommand.replacements)))
```
tzneal, line 150: Do we need this log?

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -120,42 +120,48 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
 			return lastSavedCommand, lastSavedResults, nil
-		}
-		mid := (min + max) / 2
-		candidatesToConsolidate := candidates[0 : mid+1]
+		default:
+			mid := (min + max) / 2
+			candidatesToConsolidate := candidates[0 : mid+1]
+
+			// Pass the timeout context to ensure sub-operations can be canceled
+			cmd, results, err := m.computeConsolidation(timeoutCtx, candidatesToConsolidate...)
```
tzneal, line 141: Won't the context timeout produce an err here that we need to handle specially?

I was also expecting we might need to check for context cancellation in the scheduler more, but maybe its getting caught when we do something like list pods/nodes?

rschalo: So, when I looked we wrap this error in many cases. Additionally, there is a no-op logger for the scheduler that is created for consolidation actions. The trickiest part has been deciding where we want to make cutoffs, for example in `Solve` on `pkg/controllers/provisioning/scheduling/scheduler.go:214`.

The stance I've taken is that we'll allow the current item being evaluated to finish or error out with a wrapped error and since errors are aggregated, break out of the loop and report the next pod as unschedulable.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -120,42 +122,50 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
 	lastSavedResults := scheduling.Results{}
 	// Set a timeout
-	timeout := m.clock.Now().Add(MultiNodeConsolidationTimeoutDuration)
-	// binary search to find the maximum number of NodeClaims we can terminate
+	timeoutCtx, cancel := context.WithDeadline(ctx, m.clock.Now().Add(MultiNodeConsolidationTimeoutDuration))
+	defer cancel()
 	for min <= max {
-		if m.clock.Now().After(timeout) {
+		// Check for timeout using select
+		select {
```
tzneal, line 129: Can we just check for `if ctx.Err() != nil` here, and does that make this a bit cleaner? (no select effectively and a smaller diff)

pkg/controllers/disruption/helpers.go
```diff
@@ -106,6 +107,10 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 				if _, ok := deletingNodePodKeys[client.ObjectKeyFromObject(p)]; !ok {
 					results.PodErrors[p] = NewUninitializedNodeError(n)
 				}
+				// check for a pod that was not scheduled due to ctx timeout
```
tzneal, line 110: I'm probably missing something, but what is this for? Shouldn't we normally fail scheduling if there is a pod error?

rschalo: This is probably unnecessary. Was using it to error on line 142 after calling `SimulateScheduling` in `computeConsolidation` but we do check if pods failed to schedule on line 146. We'd then return empty results, no command nil error, and continue the loop. 

The loop then either breaks because min > max and returns the last saved items or the timeout bails out of consolidation and returns the last saved items plus an error if context deadline was exceeded.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -120,22 +122,27 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
 	lastSavedCommand := Command{}
 	lastSavedResults := scheduling.Results{}
 	// Set a timeout
-	timeout := m.clock.Now().Add(MultiNodeConsolidationTimeoutDuration)
-	// binary search to find the maximum number of NodeClaims we can terminate
+	timeoutCtx, cancel := context.WithDeadline(ctx, m.clock.Now().Add(MultiNodeConsolidationTimeoutDuration))
```
jonathan-innis, line 125:
```suggestion
	timeoutCtx, cancel := context.WithTimeout(ctx, MultiNodeConsolidationTimeoutDuration)
```

nit: A bit less verbose

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -120,22 +122,27 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
 			return lastSavedCommand, lastSavedResults, nil
 		}
 		mid := (min + max) / 2
 		candidatesToConsolidate := candidates[0 : mid+1]
 
-		cmd, results, err := m.computeConsolidation(ctx, candidatesToConsolidate...)
+		// Pass the timeout context to ensure sub-operations can be canceled
+		cmd, results, err := m.computeConsolidation(timeoutCtx, candidatesToConsolidate...)
+		// context deadline exceeded will return to the top of the loop and either return nothing or the last saved command
+		if errors.Is(err, context.DeadlineExceeded) {
```
jonathan-innis, line 143: nit: Checking this inside of the `err != nil` check is the typical pattern

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -244,6 +244,12 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 			break
 		}
 
+		// when context has been canceled or deadline exceeded, stop attempting to schedule pods and mark current pod as unschedulable
+		if ctx.Err() != nil {
+			errors[pod] = ctx.Err()
+			break
```
jonathan-innis, line 250: I'm curious what you think about this in the context of applying a scheduling timeout as well -- we can choose to either: 1) Mark all pods that have not been scheduled yet as failing due to context id -- this could take some time potentially depending on the number of pods that we have OR 2) Fail the scheduling loop with a context timeout error -- in that case, we'd have to somehow know that we didn't schedule all the pods -- given that we are using errors to communicate this right now, that might not work well

jonathan-innis: In either case, I don't think we can break here because I think we need to mark all pods that weren't scheduled as unschedulable to have a proper interface

pkg/controllers/provisioning/provisioner.go
```diff
@@ -236,6 +236,9 @@ func (p *Provisioner) NewScheduler(ctx context.Context, pods []*corev1.Pod, stat
 
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	for _, np := range nodePools {
+		if ctx.Err() != nil {
```
jonathan-innis, line 239: Does this section of code take so much time that we think we need to handle this error at this level? I get that there's a trade-off between the number of times that we write this and how quickly we can respond

pkg/controllers/disruption/suite_test.go
```diff
@@ -2022,6 +2022,62 @@ var _ = Describe("Metrics", func() {
...
+		// bind pods to nodes
+		ExpectManualBinding(ctx, env.Client, pods[0], nodes[0])
+		ExpectManualBinding(ctx, env.Client, pods[1], nodes[1])
+		ExpectManualBinding(ctx, env.Client, pods[2], nodes[2])
+		ExpectManualBinding(ctx, env.Client, pods[3], nodes[2])
+
+		// inform cluster state about nodes and nodeclaims
+		ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*corev1.Node{nodes[0], nodes[1], nodes[2]}, []*v1.NodeClaim{nodeClaims[0], nodeClaims[1], nodeClaims[2]})
+		// create deadline in the past
+		deadlineCtx, cancel := context.WithDeadline(ctx, fakeClock.Now().Add(-disruption.MultiNodeConsolidationTimeoutDuration))
```
jonathan-innis, line 2074: I'm not sure if you can create a timeout that's negative but you could consider doing that here

pkg/controllers/provisioning/provisioner.go
rschalo: We probably don't need to check here. It probably makes the most sense to just check the timeout between pods in scheduling.

rschalo: Except that this context is timed out and continues on to `cloudProvider.GetInstanceTypes`. Less that we're handling the error and more that we're silencing spurious logging.

jonathan-innis: That doesn't completely solve it right? I think we just have to handle it generally because we can still race pass this check and fire spurious errors

pkg/controllers/provisioning/provisioner.go
```diff
@@ -326,9 +330,17 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 			}), 5),
 		).Info("deferring scheduling decision for provisionable pod(s) to future simulation due to limited reserved offering capacity")
 	}
+	contextErrors := results.ContextDeadlineExceededErrors()
+	if len(contextErrors) != 0 {
+		log.FromContext(ctx).V(1).WithValues(
```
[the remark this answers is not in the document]
rschalo: Is that not the case with the `.V(1)`?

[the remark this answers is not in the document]
rschalo: This is actually not needed at all in this PR. It's something we'll have to figure out if we want to add a timeout to the provisioning path.

pkg/controllers/provisioning/provisioner.go
```diff
@@ -464,16 +466,20 @@ func validateKarpenterManagedLabelCanExist(p *corev1.Pod) error {
 	return nil
 }
 
-func (p *Provisioner) injectVolumeTopologyRequirements(ctx context.Context, pods []*corev1.Pod) []*corev1.Pod {
+func (p *Provisioner) injectVolumeTopologyRequirements(ctx context.Context, pods []*corev1.Pod) ([]*corev1.Pod, error) {
 	var schedulablePods []*corev1.Pod
 	for _, pod := range pods {
 		if err := p.volumeTopology.Inject(ctx, pod); err != nil {
+			// return because we can't consider any other pod schedulable
```
jonathan-innis, line 473: What's the reason that we need to handle this here?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -286,6 +286,10 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 	lastLogTime := s.clock.Now()
 	batchSize := len(q.pods)
 	for {
+		if ctx.Err() != nil {
+			log.FromContext(ctx).V(1).WithValues("duration", s.clock.Since(startTime).Truncate(time.Second), "scheduling-id", string(s.uuid)).Info("scheduling simulation timed out")
+			break
```
jonathan-innis, line 291: We can't break, right? This means that we won't throw an error back when from the scheduling loop, right?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3682,7 +3682,9 @@ var _ = Context("Scheduling", func() {
 					g.Expect(lo.FromPtr(m.Gauge.Value)).To(BeNumerically(">", 0))
 				}, time.Second).Should(Succeed())
 			}()
-			s.Solve(injection.WithControllerName(ctx, "provisioner"), pods)
+			_, err = s.Solve(injection.WithControllerName(ctx, "provisioner"), pods)
```
jonathan-innis, line 3685: Can we expect err to be Nil()

PR #2028, opened 2025-02-24
feat: scale disruption cost by the node utilization

cnmcavoy:
Fixes #N/A 

**Description**
Scales the disruption cost of nodes by their utilization of pod resources. A node with 1 pod and 99% utilization should have a higher disruption cost than a node with 1 pod and 10% utilization. 

We've been looking at improving utilization of resources in our clusters, and noticed that Karpenter tends to prefer consolidating underutilized nodes with fewer pods rather than nodes with the most wasted resources. It seems like this can happen because the `disruptionutils.ReschedulingCost(ctx, pods)` produces a value that is roughly equivalent to the pod count (scaled by pod priority class). So nodes with fewer pods and higher utilization will be candidates for consolidation, while underutilized nodes with many smaller pods are not.

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -4017,7 +4017,12 @@ var _ = Describe("Consolidation", func() {
 			rs := test.ReplicaSet()
 			ExpectApplied(ctx, env.Client, rs)
 
-			pods := test.Pods(3, test.PodOptions{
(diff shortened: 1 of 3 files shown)


jonathan-innis:
cc: @rschalo

jonathan-innis:
@rschalo Has been thinking on this problem -- might be worth having a discussion over Slack or in a GH issue to discuss how y'all are both thinking about this

PR #2032, opened 2025-02-26
feat(nodepool): support timezone for disruption budgets

### LINKED ISSUE #1446: feat: StartDateTime and EndDateTime, Timezones for NodePool Disruption Budgets

flavono123:
Fixes #1446, partially

**Description**

Add the optional `disruption.budgets[].timeZone` field to `NodePoolSpec`, following the upstream CronJob `spec.timeZone` precedent.

```yaml
disruption:
  budgets:
    - nodes: "0"
      schedule: "0 8 * * mon-fri"
      timeZone: "America/New_York"
      duration: 10h
```

- Interpret `schedule` in the configured IANA time zone, defaulting to UTC when unset.
- Reject `timeZone` without `schedule` through CEL validation.
- Validate IANA names before interpolating them into the cron expression. Invalid zones fail closed with zero allowed disruptions and surface through the NodePool `ValidationSucceeded` condition.
- Embed the IANA time zone database so behavior is deterministic on images without system tzdata.
- Follow the upstream CronJob behavior at daylight-saving transitions and cover both skipped and repeated hours with tests.

This revision incorporates and credits the additional validation, fail-closed behavior, tzdata, and DST test coverage from #3086, rebased onto the latest `main`.

- `go test -vet=off ./pkg/apis/v1 ./pkg/controllers/nodepool/validation`
- `go vet -p=1 ./pkg/apis/v1 ./pkg/controllers/nodepool/validation`
- `git diff --check origin/main...HEAD`


flavono123:
are we have to emit a new event for wrong timezone set?:

> If the time zone name becomes invalid during the lifetime of a CronJob or due to a change in host configuration, the controller will stop creating new new Jobs and will create a system event with the reason UnknownTimeZone

https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/cron-job-v1/#CronJobSpec

flavono123:
@njtran could you mind to review this?

jonathan-innis:
@flavono123 Would you mind coming to one of our [Working Group meetings](https://github.com/kubernetes-sigs/karpenter?tab=readme-ov-file#working-group-meetings) and we can discuss this one more? I think there are some tricky things with handling timezones that we have generally avoided but open to talk more about your use-case.

flavono123:
@jonathan-innis thanks for inviting me. i'd like to, but two things are hurdles for me:
- I'm afraid to understand the conversation in real time, honestly, I'm not good at English, both speaking and listening, so it would be good if there is support for translation.
- Time is dawning for me, another time zone problem.

so let me explain my use-case in "text", in advance:

```yaml
    - duration: 21h
      nodes: "0"
      schedule: 0 21 * * sun-thu
```

I use budgets like the one above to allow disruptions from 3am to 6am, Monday to Friday(actually, leveraging does not allow from 6am to 3am tomorrow to avoid working time). I think it is more confusing when the duration is over the date and with weekdays for my case.

In short, this is just a literal time zone issue to make it easier to figure out the time span of budgets.

elmiko:
putting this in my queue for the week, my intention is to review the various places where a time value is used and enumerate them in a doc. then we can review the doc to see if we should update all the time uses to include a user-specified timezone.

elmiko:
i poked around the karpenter code this week and i think we can confirm the the `nodepool.spec.disruption.budgets.schedule` is the only place where we currently accept a date-like format, or cron format.

if we wanted to support timezones for disruption budgets, i think we would want to add a `nodepool.spec.disruption.budgets.timeZone` field, similar to how [core kubernetes allows a timezone field in cronjob specs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#time-zones).

elmiko (commented):
this is looking good to me

pkg/apis/v1/nodepool_budgets_test.go
```diff
@@ -237,6 +235,29 @@ var _ = Describe("Budgets", func() {
 			Expect(err).To(Succeed())
 			Expect(active).To(BeFalse())
 		})
+		It("should reutn that a schedule is active with a timezone", func() {
```
nandorKollar: nit: reutn -> return :)

elmiko (commented):
/lgtm

flavono123:
fyi. the code changes are similar with the previous commit.
i just did `make verify` for `go generate`, which the previous CI blamed on. there was no checking step at that time since the PR was open quite a while ago i guess .

flavono123:
do we have to wait one more year? @jackfrancis @tallaxes

elmiko:
/remove-lifecycle rotten

@flavono123 i'm not sure who we need to get a review here, but i will raise this issue at a sig meeting.

AlexanderYastrebov:
There is also https://github.com/kubernetes-sigs/karpenter/pull/3086

flavono123:
> There is also #3086

it looks a better version than this, but closed

flavono123:
> > There is also https://github.com/kubernetes-sigs/karpenter/pull/3086

> it looks a better version than this, but closed

so i enriched with the original commit in here, let met know if it is not proper please.

Issue #1842, opened 2024-11-25
Endless nodes are created after `expireAfter` elapse on a node in some scenarios


otoupin-nsesi:
### Description

**Observed Behavior**:

After `expireAfter` elapse on a node, pods are starting to get evicted, and endless new nodes are created to try to schedule those pods. Also, pods that don't have PDBs are NOT evicted.

**Expected Behavior**:

After `expireAfter` elapse on a node, pods are starting to get evicted, and one node at most is created to schedule those pods. Also, pods that don't have PDBs are evicted. There may be an odd pod that has a PDB preventing the node from getting recycled, but if this is the case we can set `terminationGracePeriod`.

**Reproduction Steps**:

1. Have one CloudNativePG database in the cluster (or a similar workload => single replica & a PDB)
2. CloudNativePG will add a PDB to the primary.
3. Have a nodepool with relatively short expiry (`expireAfter`). In our case we have dev environments set at 24h, so we caught this early.
4. Once a node expires, a weird behaviour is triggered.
    1. As expected, in v1 expiries are now forceful, so Karpenter begins to evict the pods.
    2. As expected, a new node is spun out to take up the slack.
    3. But then the problems start,
        1. Since there is a PDB on a single replica (there is only one PG primary at the time), eviction is not happening. So far so good (this is also the old behaviour, in v0.37.x the node just can't expire until we restart the database manually (or kill the primary)).
        2. However, any other pods on this node are not evicted either, while the documentation, and the log messages appear to believe it should be the case.
        3. The new node from earlier is `nomitated` for those pods, but they never transfer to that node, as they are not evicted.
        4. Then at the next batch of pod scheduling, we get `found provisionable pod(s)` again, and a new nodeclaim is added (for the same pod as earlier)
        5. And again
        6. And again
        7. And again
    8. So we end up in a situation where we have a lot of unused nodes, containing only daemonset and new workloads.
9. At the point, I restart the database, the primary move, the PDB is removed, and everything can then slowly heal. However, there was no sign of "infinite nodeclaim creation" ever ending before.

We believe this is a bug, we couldn't find a workaround (aside from removing `expireAfter`), and reverted to v0.37.x series for now.

**A few clues**:
The state of the cluster 30m-45m after expiry. Node 53-23 is the one that expired. Any nodes younger than 30min are running mostly empty (aside from daemonsets).

[...]

PR #2033, opened 2025-02-26
fix: don't provision unnecessary capacity for pods which can't move to a new node

saurav-agarwalla:
Fixes  https://github.com/kubernetes-sigs/karpenter/issues/1842, https://github.com/kubernetes-sigs/karpenter/issues/1928, https://github.com/aws/karpenter-provider-aws/issues/7521

**Description**
This PR takes care of two cases where Karpenter provisions unnecessary capacity today:
1. Pods with `karpenter.sh/do-not-disrupt=true`: they aren't [evictable](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/utils/pod/scheduling.go#L56) so it makes sense for Karpenter to not consider them reschedulable as well since the only end state for these pods is either to get into a terminal state or be forcibly deleted (after a termination grace period). This prevents Karpenter from spinning up and reserving unnecessary capacity for these pods on new nodes.
2. Pods which can't be evicted due to fully blocking PDBs: Karpenter doesn't know how long it will take for these pods to be successfully evicted and in absence of TGP, it can take a really long time for these pods to be evicted (if at all). When this happens, Karpenter shouldn't provision unnecessary capacity for these pods.

This doesn't handle the case where there's a PDB that transiently blocks eviction since there's more work needed to handle that case as Karpenter can't predict if the eviction blocker is temporary or permanent. If Karpenter doesn't consider a pod with a transient eviction blocker (say due to a low disruption ratio due to a PDB), it can end up creating a lot of smaller nodeclaims instead of a large nodeclaim that would fit all these pods and it could end up costing more overall especially if customers don't have consolidation enabled.

Without this change:
* Karpenter will continue to bring up new nodeclaims when the original nodeclaim expires (but can't be terminated because these pods hold up the termination in the absence of a termination grace period)
* Karpenter will not be able to consolidate nodeclaims nominated for these pods because these pods are never going to move to that nodeclaim

Reproduced the scenario where a nodeclaim has pods with `karpenter.sh/do-not-disrupt=true` and saw that Karpenter was continuously spinning up new nodeclaims after the expiry of the original nodeclaim (even though it was stuck due to these pods). It wasn't able to consolidate new nodeclaims that were nominated for these pods either.

After the change, Karpenter doesn't spin up new nodeclaims for these pods.

Did the same thing for pods whose eviction was blocked due to PDB.

### DIFF
--- a/pkg/controllers/disruption/drift.go
+++ b/pkg/controllers/disruption/drift.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"sort"
 
+	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/karpenter/pkg/utils/pretty"
@@ -39,14 +40,16 @@ type Drift struct {
 	cluster     *state.Cluster
 	provisioner *provisioning.Provisioner
 	recorder    events.Recorder
+	clock       clock.Clock
 }
 
-func NewDrift(kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
+func NewDrift(clock clock.Clock, kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
 	return &Drift{
 		kubeClient:  kubeClient,
 		cluster:     cluster,
 		provisioner: provisioner,
 		recorder:    recorder,
+		clock:       clock,
 	}
 }
 
@@ -92,7 +95,7 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
 			continue
 		}
 		// Check if we need to create any NodeClaims.
-		results, err := SimulateScheduling(ctx, d.kubeClient, d.cluster, d.provisioner, candidate)
+		results, err := SimulateScheduling(ctx, d.clock, d.kubeClient, d.cluster, d.provisioner, candidate)
 		if err != nil {
 			// if a candidate is now deleting, just retry
 			if errors.Is(err, errCandidateDeleting) {
--- a/pkg/controllers/disruption/drift.go
+++ b/pkg/controllers/disruption/drift.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"sort"
 
-	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/karpenter/pkg/utils/pretty"
@@ -40,16 +39,14 @@ type Drift struct {
 	cluster     *state.Cluster
 	provisioner *provisioning.Provisioner
 	recorder    events.Recorder
-	clock       clock.Clock
 }
 
-func NewDrift(clock clock.Clock, kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
+func NewDrift(kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
 	return &Drift{
 		kubeClient:  kubeClient,
 		cluster:     cluster,
 		provisioner: provisioner,
 		recorder:    recorder,
-		clock:       clock,
 	}
 }
 
@@ -95,7 +92,7 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
 			continue
 		}
 		// Check if we need to create any NodeClaims.
-		results, err := SimulateScheduling(ctx, d.clock, d.kubeClient, d.cluster, d.provisioner, candidate)
+		results, err := SimulateScheduling(ctx, d.kubeClient, d.cluster, d.provisioner, candidate)
 		if err != nil {
 			// if a candidate is now deleting, just retry
 			if errors.Is(err, errCandidateDeleting) {
--- a/pkg/controllers/disruption/drift_test.go
+++ b/pkg/controllers/disruption/drift_test.go
@@ -575,7 +575,7 @@ var _ = Describe("Drift", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodeClaim)
 		})
-		It("should drift nodes that have pods with the blocking PDBs when the NodePool's TerminationGracePeriod is not nil", func() {
+		It("should ignore nodes that have pods with the blocking PDBs when the NodePool's TerminationGracePeriod is not nil", func() {
 			nodeClaim.Spec.TerminationGracePeriod = &metav1.Duration{Duration: time.Second * 300}
 			podLabels := map[string]string{"test": "value"}
 			pod := test.Pod(test.PodOptions{
@@ -595,8 +595,9 @@ var _ = Describe("Drift", func() {
 
 			ExpectSingletonReconciled(ctx, disruptionController)
 
-			// Expect to create a replacement but not delete the old nodeclaim
-			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(2)) // new nodeclaim is created for drift
+			// Pods with blocking PDBs can't be evicted and hence can't be rescheduled on a new node.
+			// Expect no new nodeclaims to be created.
+			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(1))
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodeClaim)
 		})
--- a/pkg/controllers/disruption/drift_test.go
+++ b/pkg/controllers/disruption/drift_test.go
@@ -553,8 +553,7 @@ var _ = Describe("Drift", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodeClaim)
 		})
-		It("should drift nodes that have pods with the karpenter.sh/do-not-disrupt annotation when the NodePool's TerminationGracePeriod is not nil", func() {
-			nodeClaim.Spec.TerminationGracePeriod = &metav1.Duration{Duration: time.Second * 300}
+		It("should ignore nodes that have pods with the karpenter.sh/do-not-disrupt annotation when the NodePool's TerminationGracePeriod is not nil", func() {
 			pod := test.Pod(test.PodOptions{
 				ObjectMeta: metav1.ObjectMeta{
 					Annotations: map[string]string{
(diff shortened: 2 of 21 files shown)


otoupin-nsesi:
Could we have the same “not reschedulable” behaviour for “problematic” PDBs? They cause the same issues and have the same behaviour as the `do-not-disrupt` pods, a.k.a. they are not truly reschedulable and will never move (without a TGP). 

For example, the following PDB:

```
NAMESPACE       NAME                 MIN-AVAILABLE     MAX-UNAVAILABLE     ALLOWED-DISRUPTIONS     CURRENT     DESIRED     EXPECTED 
cnpg-system     my-db-primary                    1                 n/a                       0           1           1            1
```

is “problematic” in a Karpenter context, causing the same behaviour you are trying to patch, so it would make sense if they are treated the same. The downside being it’s a lot more logic than just checking for do-not-disrupt annotations, and maybe it doesn’t belong here.

Similar problematic PDBs: singleton pod (replica 1, min-available 1) (likely done on purpose), and misconfigured PDBs (blocks in similar ways, but are a mistake. Could be ignored as you could argue it should fixed or detected by policies).

saurav-agarwalla:
That's precisely what we discussed today: https://github.com/kubernetes-sigs/karpenter/issues/1928#issuecomment-2688691245

I am exploring that option. But I want to keep these changes separate to make it easier to review and justify.

saurav-agarwalla:
Pushed changes to handle this for pods which don't get evicted due to PDB violation as well since they weren't huge.

pkg/utils/pod/scheduling.go
```diff
@@ -39,13 +39,15 @@ func IsActive(pod *corev1.Pod) bool {
 // - Is an active pod (isn't terminal or actively terminating) OR Is owned by a StatefulSet and Is Terminating
 // - Isn't owned by a DaemonSet
 // - Isn't a mirror pod (https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/)
+// - Does not have the "karpenter.sh/do-not-disrupt=true" annotation (https://karpenter.sh/docs/concepts/disruption/#pod-level-controls)
 func IsReschedulable(pod *corev1.Pod) bool {
 	// StatefulSet pods can be handled differently here because we know that StatefulSet pods MUST
 	// get deleted before new pods are re-created. This means that we can model terminating pods for StatefulSets
 	// differently for higher availability by considering terminating pods for scheduling
 	return (IsActive(pod) || (IsOwnedByStatefulSet(pod) && IsTerminating(pod))) &&
```
jonathan-innis, line 47: What do you think about adding the PDB handling in here so that we have all of the logic for pods that we think are reschedulable and should be considered in one place?

pkg/controllers/disruption/helpers.go
```diff
@@ -78,6 +78,18 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 	for _, n := range candidates {
 		pods = append(pods, n.reschedulablePods...)
 	}
+
+	// Don't provision capacity for pods which aren't getting evicted due to PDB violation.
+	// Since Karpenter doesn't know when these pods will be successfully evicted, spinning up capacity until
+	// these pods are evicted is wasteful.
+	pdbs, err := pdb.NewLimits(ctx, clock, kubeClient)
```
jonathan-innis, line 85: Just doing a raw check on whether the pod is evictable or not is going to lead to some weird behavior I think -- there are going to be cases where pods are going to be in the middle of evicting off a node and we had considered these same pods in a previous iteration (because the PDBs weren't blocking eviction at that time). I think we can really only consider PDBs where they are persistently blocking (0 unavailable, etc.). We could maybe think about considering a time-based check for persistently blocking PDBs in the future to scope into this (e.g. if a PDB has been fully blocking for more than an hour then it's "stuck")

saurav-agarwalla: > Just doing a raw check on whether the pod is evictable or not is going to lead to some weird behavior I think -- there are going to be cases where pods are going to be in the middle of evicting off a node and we had considered these same pods in a previous iteration (because the PDBs weren't blocking eviction at that time).

Do you mind elaborating this a little more? The way I understand this, if a pod is terminating due to eviction, we don't consider it `IsReschedulable` anyway. It is possible that whether a pod is reschedulable or not can somewhat quickly change due to this logic but I am not sure if that's necessarily a bad thing since eventually things will settle down. This isn't very different from when pods are blocked due to the `do-not-disrupt` annotation but the customer removes this annotation.

We also discussed the time-based thing in one of the community meetings. While it is doable, determining the actual value is tricky and no matter what we choose, some customers could always prefer a shorter time. But the trade-off is eventually with the downsides of the current approach so I am trying to understand that first and then we can see which approach we choose.

saurav-agarwalla: As we discussed offline, I have updated the PR to consider fully blocking PDBs only for now to avoid creating a lot of small nodes if we consider all PDBs.

pkg/utils/pod/scheduling.go
saurav-agarwalla: The reason I didn't do this is that this is used to compute reschedulable pods for candidate selection which is then used in a lot of places to determine things like empty nodes, etc. Felt like there could be unintentional side-effects since I don't fully understand those code paths yet hence trying to make this a targeted change. What do you think?

pkg/controllers/disruption/drift.go
```diff
@@ -39,14 +40,16 @@ type Drift struct {
...
 }
 
-func NewDrift(kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
+func NewDrift(clock clock.Clock, kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
 	return &Drift{
 		kubeClient:  kubeClient,
 		cluster:     cluster,
 		provisioner: provisioner,
 		recorder:    recorder,
+		clock:       clock,
```
DerekFrank, line 52: nit: The ordering is inconsistent

pkg/controllers/disruption/helpers.go
jonathan-innis: Rather than having to do the check below with PDBs in both places,  is it worthwhile to integrate the PDB check in the pod scheduling utils as well as integrate it with the `ReschedulablePods` call so that it's standardized?

jonathan-innis: Particularly, I think we need to make sure that we don't consider all of the candidate node pods for rescheduling if they aren't reschedulable

pkg/controllers/disruption/drift.go
saurav-agarwalla: Ordering of the struct members here? Not sure if there's a convention for that, is there?

pkg/controllers/disruption/helpers.go
saurav-agarwalla: Done - I added a concept of `IsCurrentlyReschedulable` which considers the `do-not-disrupt` annotation and fully blocking PDBs to determine whether or not a pod is reschedulable at a given point-in-time. This was needed to keep `IsReschedulable` untouched as that's used for more things like determining if the node is empty which shouldn't change as part of this PR.

pkg/utils/pdb/pdb.go
```diff
@@ -86,13 +87,45 @@ func (l Limits) CanEvictPods(pods []*v1.Pod) (client.ObjectKey, bool) {
 	return client.ObjectKey{}, true
 }
 
+// IsFullyBlocked returns true if the given pod is fully blocked by a PDB.
+func (l Limits) IsFullyBlocked(pod *v1.Pod) (client.ObjectKey, bool) {
+	// If the pod isn't eligible for being evicted, then a fully blocking PDB doesn't matter
+	// This is due to the fact that we won't call the eviction API on these pods when we are disrupting the node
+	if !podutil.IsEvictable(pod) {
```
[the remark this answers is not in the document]
saurav-agarwalla: Done.

pkg/controllers/disruption/helpers.go
```diff
@@ -65,20 +65,33 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
...
+	// Don't provision capacity for pods which will not get evicted due to fully blocking PDBs.
+	// Since Karpenter doesn't know when these pods will be successfully evicted, spinning up capacity until
+	// these pods are evicted is wasteful.
+	pdbs, err := pdb.NewLimits(ctx, kubeClient)
+	if err != nil {
+		return scheduling.Results{}, fmt.Errorf("tracking PodDisruptionBudgets, %w", err)
+	}
 	for _, n := range candidates {
-		pods = append(pods, n.reschedulablePods...)
+		currentlyReschedulablePods := lo.Filter(n.reschedulablePods, func(p *corev1.Pod, _ int) bool {
```
jonathan-innis, line 82: Is there a way to store this in NewCandidate so that we only store the pods that we have to model for scheduling and we somehow do the emptiness check through some other struct member? If emptiness is the only place we don't use `CurrentlReschedulable`, I question if there's a simpler way to store this

pkg/controllers/disruption/drift_test.go
```diff
@@ -553,8 +553,7 @@ var _ = Describe("Drift", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodeClaim)
 		})
-		It("should drift nodes that have pods with the karpenter.sh/do-not-disrupt annotation when the NodePool's TerminationGracePeriod is not nil", func() {
-			nodeClaim.Spec.TerminationGracePeriod = &metav1.Duration{Duration: time.Second * 300}
+		It("should ignore nodes that have pods with the karpenter.sh/do-not-disrupt annotation when the NodePool's TerminationGracePeriod is not nil", func() {
```
jonathan-innis, line 556: This changes the meaning of this test case -- did we mean to change this?

pkg/utils/pdb/pdb.go
```diff
@@ -55,44 +55,83 @@ func NewLimits(ctx context.Context, clk clock.Clock, kubeClient client.Client) (
...
-								ignorePod = true
-								continue
-							}
+	}
+	return client.ObjectKey{}, true
+}
+
+// isFullyBlocked returns true if the given pod is fully blocked by a PDB.
+func (l Limits) isFullyBlocked(pod *v1.Pod) (client.ObjectKey, bool) {
+	pdb, evictable := l.isEvictable(pod, func(pdb *pdbItem) bool {
```
jonathan-innis, line 71: Never a huge fan of callbacks -- is there a way to just make this an option on the function below that you can specify one or the other on?

pkg/controllers/disruption/drift_test.go
saurav-agarwalla: Yes, this is capturing the case that a drift doesn't spin up a new node when there are blocking pods on the disrupted node. But let me make some changes to convey this a little better.

pkg/controllers/disruption/helpers.go
saurav-agarwalla: Emptiness isn't the only place this is checked today. We also check this for things like determining if the disruption budgets are breached.

I could have a separate `currentlyReschedulablePods` struct member in `candidate` but seemed like duplicate information since it can be deducted from `reschedulablePods` although one could argue that the additional computation isn't worth it. Primarily based it off the fact that this is only fetched from `candidate` in a single place at the moment which is the above one.

pkg/utils/pdb/pdb.go
saurav-agarwalla: Updating to an enum.

jonathan-innis (approved):
/lgtm
/approve

PR #2044, opened 2025-03-02
feat: add karpenter_pods_drained_total metric to track pod draining by reason

### LINKED ISSUE #2021: Add `karpenter_pods_drained_total` during Node drain
### Description

It's useful to know how many pods are drained by a particular reason in the same way it's useful to know how many Nodes are drained by a particular reason (which we should also consider adding to the Prom metric). We basically need to add a Inc() on this counter metric each time we successfully drain a pod [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/terminator/eviction.go#L217).

Would provide more visibility into the number of pods that end up being drained by a particular reason

omerap12:
Fixes #2021

**Description**
Implements a new Prometheus metric to count pods drained during node termination, labeled by the reason for draining. This provides visibility into the number of pods affected by different termination scenarios.

### DIFF
--- a/pkg/controllers/node/termination/terminator/metrics.go
+++ b/pkg/controllers/node/termination/terminator/metrics.go
@@ -27,6 +27,8 @@ import (
 const (
 	// CodeLabel for eviction request
 	CodeLabel = "code"
+	// ReasonLabel for pod draining
+	ReasonLabel = "reason"
 )
 
 var NodesEvictionRequestsTotal = opmetrics.NewPrometheusCounter(
@@ -38,4 +40,15 @@ var NodesEvictionRequestsTotal = opmetrics.NewPrometheusCounter(
 		Help:      "The total number of eviction requests made by Karpenter",
 	},
 	[]string{CodeLabel},
+)
+
+var PodsDrainedTotal = opmetrics.NewPrometheusCounter(
+	crmetrics.Registry,
+	prometheus.CounterOpts{
+		Namespace: metrics.Namespace,
+		Subsystem: metrics.PodSubsystem,
+		Name:      "pods_drained_total",
+		Help:      "The total number of pods drained during node termination by Karpenter, labeled by reason",
+	},
+	[]string{ReasonLabel},
 )
--- a/pkg/controllers/node/termination/terminator/suite_test.go
+++ b/pkg/controllers/node/termination/terminator/suite_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/awslabs/operatorpkg/status"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/samber/lo"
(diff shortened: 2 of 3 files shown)


pkg/controllers/node/termination/terminator/metrics.go
```diff
@@ -39,3 +41,14 @@ var NodesEvictionRequestsTotal = opmetrics.NewPrometheusCounter(
 	},
 	[]string{CodeLabel},
 )
+
+var PodsDrainedTotal = opmetrics.NewPrometheusCounter(
+	crmetrics.Registry,
+	prometheus.CounterOpts{
+		Namespace: metrics.Namespace,
+		Subsystem: metrics.NodeSubsystem,
```
jonathan-innis, line 49: I think this needs to be in the `metrics.PodsSubsystem` -- as it stands right now this is called `karpenter_nodes_pods_drained_total` which doesn't sound right

jonathan-innis:
I also think we need a rebase with one of the latest changes that was made to this section of code

jonathan-innis:
It might also be nice to add a test to make sure that we are properly emitting the metric -- I think you should be able to find similar metric tests for node eviction requests total

pkg/controllers/node/termination/terminator/metrics.go
omerap12: Sure

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -164,6 +165,13 @@ var _ = Describe("Eviction/Queue", func() {
 				queue.Add(node, test.Pod())
 			}
 		})
+		It("should increment PodsDrainedTotal metric when a pod is evicted", func() {
```
engedaam, line 168: Can we add one more test around a reason that is set on the nodeclaim?

omerap12: Sure

omerap12: I tried adding another test but got this error:  
`listing nodeclaims, field label not supported: status.providerID.`

I believe this happens because our test setup uses a simpler Kubernetes API server that doesn't fully support CRDs with field selectors. In production, this would work fine, but our test environment lacks this feature.  

I added a test to ensure the metric updates, even if the reason is `UnknownReason`. INMHO the core functionality is still covered ( but I agree that testing with a specific NodeClaim reason would be idea )

omerap12: nvm, lol just needed to set the indexer :)

engedaam (commented):
/lgtm

omerap12:
Friendly ping @jonathan-innis , can we get this merged?

panicstevenson:
FYI that this metric is `karpenter_pods_pods_drained_total`. Not sure if the `pods` duplication is intentional given the subsystem.

PR #2048, opened 2025-03-04
ci: Automating scheduling benchmarking testing for PRs

DerekFrank:
Fixes #N/A 

**Description**

New CI to automatically run scheduling benchmark tests for both the existing code and proposed changes whenever a PR modifies a `.go` file. The new workflow will comment results into PR comments, and also store CPU and MEM profiles into a downloadable artifacts that reviewers and approvers can use to analyze any performance improvements or regressions. The current setup leverages the `matrix` functionality to make setting up new benchmark CI tests simple in the future. 

This is a first iteration of a CI suite for performance testing. There are a couple, sometimes contradictory natural extensions for the future:
* Automatically fail the test if performance regresses and remove performance results from PR comments
* Add Consolidation performance benchmark tests
* Parameterize more components of the `go test` call
* Refactor the approach to run a single bench test per GitHub action for cleaner replication

CI testing will always be a moving target as the needs of the community develop, this is intended to be a quick and dirty approach to gather feedback and allow the CI performance testing to adapt to the needs of developers.

Simulated PRs in my own fork to demonstrate the functionality.

PR that shows benchmarking in action:

* https://github.com/DerekFrank/karpenter-testing-fork/pull/13

PR that shows how doc and other non `.go` functionality changes won't get tested:
* https://github.com/DerekFrank/karpenter-testing-fork/pull/12

### DIFF
--- a/.github/workflows/scheduling-benchmarking.yaml
+++ b/.github/workflows/scheduling-benchmarking.yaml
@@ -1,0 +1,38 @@
+name: "Scheduling Benchmarking"
+
+on:
+  pull_request:
+    paths:
+      - "**.go"
+    branches:
+      - main
+
+jobs:
+  before:
+    name: Before PR
+    permissions:
+        pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        path: ["pkg/controllers/provisioning/scheduling"]
+    uses: ./.github/workflows/run-bench-test.yaml
+    secrets: inherit
+    with:
+      path: ${{ matrix.path }}
+      runName: before-${{ strategy.job-index }}
+      githubSha: ${{ github.event.pull_request.base.sha }}
+  after:
+    name: After PR
+    permissions:
+      pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        path: ["pkg/controllers/provisioning/scheduling"]
+    uses: ./.github/workflows/run-bench-test.yaml
+    secrets: inherit
+    with:
+      path: ${{ matrix.path }}
+      runName: after-${{ strategy.job-index }}  
+      githubSha: ${{ github.sha }}
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -61,3 +61,11 @@ updates:
       action-deps:
         patterns:
           - '*'
+  - package-ecosystem: github-actions
+    directory: .github/actions/run-bench-test
+    schedule:
+      interval: weekly
+    groups:
+      action-deps:
+        patterns:
+          - '*'
--- a/.github/workflows/run-bench-test.yaml
+++ b/.github/workflows/run-bench-test.yaml
@@ -1,0 +1,53 @@
+name: "Run Bench Test"
+
+on:
+  workflow_call:
+    inputs:
+      path:
+        description: Path to the performance test
+        required: true
+        type: string
+      runName:
+        description: Name of the run, for the purpose of file naming and github comments
+        required: true
+        type: string
+      githubSha:
+        description: Sha of the github commit to check out
+        required: true
+        type: string
+jobs:
+  run-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.githubSha }}
+      - name: Run Test
+        id: test-run
+        shell: bash
+        run: |
+          {
+            cd ${{ inputs.path }}
+            mkdir output
+            go test -tags=test_performance -run=1 -bench=. -count=1 -cpuprofile output/cpu.out -memprofile output/mem.out > output/results.txt
+            echo 'OUTPUT<<EOF'
+            cat output/results.txt
+            echo EOF
+          } >> "$GITHUB_OUTPUT"
+      - uses: actions/upload-artifact@v4
+        id: artifact-upload
+        with:
+          name: ${{ inputs.runName }}
+          path: ${{ inputs.path }}/output
+      - uses: actions/github-script@v7
+        env:
+          RESULTS: ${{ steps.test-run.outputs.OUTPUT }}
+          UPLOAD: ${{ steps.artifact-upload.outputs.artifact-url }}
+        with:
+          script: |
+              github.rest.issues.createComment({
+                issue_number: context.issue.number,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                body: `# :mag: ${{ inputs.runName }} :mag:\nResults of benchmarking testing in \`${{ inputs.path }}\`: \n\`\`\`${process.env.RESULTS}\n\`\`\`\n${process.env.UPLOAD}`
+              })


DerekFrank:
Absolutely! I wanted to keep the initial CI relatively lightweight so I avoided any specific analysis of the results. We've got lots of paths to take the general concept of CI performance testing, I tend to err on the side of incremental, fast changes that allow us to see how contributors actually use the CI in practice. If a multi-run and benchstat analysis of the results setup is preferred, I would be happy to add that to the PR or to tackle it in a follow up PR

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -0,0 +1,53 @@
...
+    permissions:
+        pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        path: ["pkg/controllers/provisioning/scheduling"]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
rschalo, line 28: This automatically fetches head in main?

DerekFrank: Ah yes, I think we need to grab the target of the PR actually if we are making a PR against a branch, maybe `github.event.pull_request.base.sha`? I'll do some testing in my fork

rschalo:
Sent a couple questions via k8s slack

rschalo (approved):
/lgtm

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -0,0 +1,54 @@
+name: "Scheduling Benchmarking"
+
+on:
+  pull_request:
```
jonathan-innis, line 4: Do we need to handle specific types here? I'm curious if we care which branch we open it again, which types that we use, and which paths that we modify -- feel like this may be a little too specific

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -0,0 +1,54 @@
...
+      - "**.go"
+    branches:
+      - main
+
+jobs:
+  before:
+    name: Before PR
+    runs-on: ubuntu-latest
+    permissions:
+        pull-requests: write
```
jonathan-innis, line 20: Can we validate whether pull request write only gives you permission on this specific PR? If it gives wider permission, we should be careful about this

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -0,0 +1,54 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.base.sha }}
+      - uses: ./.github/actions/run-bench-test
+        with:
+            path: ${{ matrix.path }}
+            runName: before-${{ strategy.job-index }}
+            githubToken: ${{ secrets.GITHUB_TOKEN }}
```
jonathan-innis, line 34: Can you not access the GH token secret from within the action directly

.github/actions/run-bench-test/action.yaml
```diff
@@ -0,0 +1,45 @@
+name: RunBenchTest
+description: 'Runs a given Benchmark Test'
+inputs:
```
jonathan-innis, line 3: I forget if you can mark some of these inputs as required, but if you can, can you mark down the ones that are needed for this action to run rather than defaulting to an empty string

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -0,0 +1,54 @@
...
+      pull-requests: write
+    strategy:
+      fail-fast: false
+      matrix:
+        path: ["pkg/controllers/provisioning/scheduling"]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
```
jonathan-innis, line 48: Doesn't it checkout the HEAD sha by default. For some reason I thought the fetch depth was also 0 by default too

.github/workflows/scheduling-benchmarking.yaml
DerekFrank: In general we could make this very broad, but I figured in general lets only run it when it makes sense to.

There are a lot of automatic triggers for [pull_request](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) that we don't need to run a benchmark against, such as `labeled` and `unlabeled`. Similarly, we don't need to test performance if we're making a doc only change, so limiting it to the `.go` files to me makes sense. The main branch was just so we don't benchmark every patch release but I'm all ears for benchmarking those if we'd like

.github/workflows/scheduling-benchmarking.yaml
DerekFrank: It gives wider permissions: 

https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-pull-requests

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions


We can skip the comment and just have them available as artifacts\in the job runs and then remove this permission if you'd prefer

.github/workflows/scheduling-benchmarking.yaml
DerekFrank: I believe I tried that, I can test it again to be sure. I think maybe we can use [secrets: inherit](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idsecretsinherit) to get around this

.github/workflows/scheduling-benchmarking.yaml
DerekFrank: Defaults can be found [here](https://github.com/actions/checkout?tab=readme-ov-file#usage). Looks like fetch-depth defaults to 1 (which is actually what we want now that I read it), and the ref says:

>  The branch, tag or SHA to checkout. When checking out the repository that
>  triggered a workflow, this defaults to the reference or SHA for that event.
>  Otherwise, uses the default branch.

The reference for [pull_request](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request) is: 

> Last merge commit on the GITHUB_REF branch

I wasn't sure if we wanted only the last merge commit so I was explicit about the reference to use.

.github/actions/run-bench-test/action.yaml
DerekFrank: I can mark them as required, I'll add that

.github/workflows/scheduling-benchmarking.yaml
jonathan-innis: Yeah, I guess we can play it by ear as we go

.github/workflows/scheduling-benchmarking.yaml
jonathan-innis: I'm not 100% sure that this is going to work just reading through the docs, but let's merge it in and find out

.github/workflows/scheduling-benchmarking.yaml
jonathan-innis: Oooo -- that must be a new feature -- yeah, we should try to use that

.github/workflows/scheduling-benchmarking.yaml
jonathan-innis: I think just keep the defaults for now

jonathan-innis (approved):
/lgtm
/approve

PR #2049, opened 2025-03-04
remove unschedulable_pods_count metric in disruption loop

### LINKED ISSUE #1993: `karpenter_scheduler_unschedulable_pods_count` has unnecessary `DeletePartialMatch` in the scheduler
### Description

`karpenter_scheduler_unschedulable_pods_count` is set at the top of every provisioning loop -- realistically, this value is continually set as provisioning loops are run and we shouldn't need this metric in the disruption loops (because we expect that there are going to be cases with consolidation where pods won't schedule) -- we should remove the `DeletePartialMatch` as well as the `controller` label on the metric since it's unneccesary

jxs1211:
Fixes #1993 https://github.com/kubernetes-sigs/karpenter/issues/1993
**Description**
remove the unschedulable_pods_count metric if the controller is not `provisioner`

make presubmit

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -275,7 +275,9 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 	// We need to schedule them alternating, A, B, A, B, .... and this solution also solves that as well.
 	errors := map[*corev1.Pod]error{}
 	// Reset the metric for the controller, so we don't keep old ids around
-	UnschedulablePodsCount.DeletePartialMatch(map[string]string{ControllerLabel: injection.GetControllerName(ctx)})
+	if injection.GetControllerName(ctx) == "provisioner" {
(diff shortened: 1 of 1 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -275,7 +275,9 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) Results {
 	// We need to schedule them alternating, A, B, A, B, .... and this solution also solves that as well.
 	errors := map[*corev1.Pod]error{}
 	// Reset the metric for the controller, so we don't keep old ids around
-	UnschedulablePodsCount.DeletePartialMatch(map[string]string{ControllerLabel: injection.GetControllerName(ctx)})
+	if injection.GetControllerName(ctx) == "provisioner" {
```
jonathan-innis, line 278: Can we just drop deleting this altogether? It's not clear to me why we just don't rely on the next round of scheduling to handle it -- the one thing that we do need to handle is setting the value to 0 when we don't find any pending pods/when we aren't getting a trigger to run scheduling.

jonathan-innis (commented):
Rather than deleting the metric, I'd like to just see us setting it to 0

PR #2050, opened 2025-03-04
chore: Refactor some detials of disruption reason condition

jonathan-innis:
Fixes #N/A 

**Description**

Set disruption reason as the `Reason` field in the `DisruptionReason` condition and don't return error from `evictionReason`

`make presubmit`


engedaam (commented):
/lgtm

PR #2052, opened 2025-03-04
fix: bump status condition observed generations to prevent nodeclaims…

### LINKED ISSUE #2034: Nodeclaim stuck in Unknown readiness status due to generation mismatch
### Description

**Observed Behavior**:
Nodeclaims are stuck in an `Unknown` readiness status when disruption is blocked.

**Expected Behavior**:
Nodeclaims should be ready even if the disruption is blocked.

The root cause of this is the [mismatch in observed generation of the status conditions and the generation of the nodeclaim](https://github.com/awslabs/operatorpkg/blob/main/status/condition_set.go#L269). This happens because once the nodeclaim is marked for deletion, the status conditions [aren't reconciled to the latest generation](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/launch.go#L47-L49) anymore since they [get short-circuited here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/controller.go#L119).

Spin up a nodeclaim with an aggressive `expireAfter` (via the nodepool) and then schedule pods with `karpenter.sh/do-not-disrupt: "true"` on them. After the nodeclaim expires, it moves to `Unknown` for status condition `Ready` due to `Normal  Ready              50m                 karpenter  Status condition transitioned, Type: Ready, Status: True -> Unknown, Reason: ReconcilingDependents, Message: Initialized=True, Registered=True, Launched=True`.

```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: default
spec:
  disruption:
    budgets:
    - nodes: 10%
    consolidateAfter: 1m
  limits:
    cpu: 1000
  template:
    spec:
      expireAfter: 5m
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: default
      requirements:
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
      - key: kubernetes.io/os
        operator: In
        values:
        - linux
        operator: In
        values:
        - on-demand
        operator: In
        values:
        - c
        - m
        - r
        operator: Gt
        values:
        - "2"
```

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: inflate
spec:
  replicas: 10
  selector:
    matchLabels:
      app: inflate
  template:
    metadata:
      labels:
        app: inflate
    spec:
      terminationGracePeriodSeconds: 0
      securityContext:
        runAsUser: 1000
        runAsGroup: 3000
        fsGroup: 2000
      containers:
      - name: inflate
        resources:
          requests:
            cpu: 1
        securityContext:
          allowPrivilegeEscalation: false
```

**Versions**:
- Chart Version:

rschalo:
… from going unknown during deletion

Fixes #2034 

**Description**
Performs a no-op status condition update during finalization to prevent NodeClaims from going into `Ready=Unknown` during deletion.

make presubmit

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/controller.go
+++ b/pkg/controllers/nodeclaim/lifecycle/controller.go
@@ -171,6 +171,21 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
 
 //nolint:gocyclo
 func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
+	// perform a no-op for whatever the status condition is currently set to
+	// so that we bump the observed generation to the latest and prevent the nodeclaim
+	// root status from entering an `Unknown` state
+	stored := nodeClaim.DeepCopy()
+	for _, condition := range nodeClaim.GetConditions() {
+		if condition.ObservedGeneration != nodeClaim.Generation {
+			nodeClaim.StatusConditions().Set(condition)
+		}
+	}
+	if !equality.Semantic.DeepEqual(stored, nodeClaim) {
+		if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
 	if !controllerutil.ContainsFinalizer(nodeClaim, v1.TerminationFinalizer) {
 		return reconcile.Result{}, nil
 	}
@@ -226,7 +241,7 @@ func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (rec
 			metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
 		})
 	}
-	stored := nodeClaim.DeepCopy() // The NodeClaim may have been modified in the EnsureTerminated function
+	stored = nodeClaim.DeepCopy() // The NodeClaim may have been modified in the EnsureTerminated function
 	controllerutil.RemoveFinalizer(nodeClaim, v1.TerminationFinalizer)
 	if !equality.Semantic.DeepEqual(stored, nodeClaim) {
 		// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
--- a/pkg/controllers/nodeclaim/lifecycle/suite_test.go
+++ b/pkg/controllers/nodeclaim/lifecycle/suite_test.go
@@ -87,7 +87,7 @@ var _ = AfterEach(func() {
 	cloudProvider.Reset()
 })
 
-var _ = Describe("Finalizer", func() {
+var _ = FDescribe("Finalizer", func() {
 	var nodePool *v1.NodePool
 
 	BeforeEach(func() {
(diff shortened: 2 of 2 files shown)


pkg/controllers/nodeclaim/lifecycle/suite_test.go
```diff
@@ -87,7 +87,7 @@ var _ = AfterEach(func() {
 	cloudProvider.Reset()
 })
 
-var _ = Describe("Finalizer", func() {
+var _ = FDescribe("Finalizer", func() {
```
jigisha620, line 90: Might want to drop the `F` here.

pkg/controllers/nodeclaim/lifecycle/suite_test.go
```diff
@@ -136,6 +136,39 @@ var _ = Describe("Finalizer", func() {
 			})
 			Expect(ok).To(BeFalse())
 		})
+		FIt("should update status conditions to the latest generation when finalizing", func() {
```
jigisha620, line 139: Same here. If you could drop the `F`.

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -171,6 +171,21 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
 
 //nolint:gocyclo
 func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
+	// perform a no-op for whatever the status condition is currently set to
+	// so that we bump the observed generation to the latest and prevent the nodeclaim
```
jonathan-innis, line 175: Can we add a note that the generation bumps here because of the deletionTimestamp being set?

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -171,6 +171,21 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
...
+	// root status from entering an `Unknown` state
+	stored := nodeClaim.DeepCopy()
+	for _, condition := range nodeClaim.GetConditions() {
+		if condition.ObservedGeneration != nodeClaim.Generation {
+			nodeClaim.StatusConditions().Set(condition)
```
jonathan-innis, line 180: Seems like you can just call set and not even do the check -- just blindly setting should be fine since if there is not change we just won't have to patch

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -171,6 +171,21 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
 
 //nolint:gocyclo
 func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
+	// perform a no-op for whatever the status condition is currently set to
+	// so that we bump the observed generation to the latest and prevent the nodeclaim
+	// root status from entering an `Unknown` state
+	stored := nodeClaim.DeepCopy()
+	for _, condition := range nodeClaim.GetConditions() {
```
jonathan-innis, line 178: I think we have to be careful here about the status conditions that we are updating -- we should just update the ones that are handled by the lifecycle controller -- no other ones

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -171,6 +171,21 @@ func (c *Controller) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (re
...
+		if condition.ObservedGeneration != nodeClaim.Generation {
+			nodeClaim.StatusConditions().Set(condition)
+		}
+	}
+	if !equality.Semantic.DeepEqual(stored, nodeClaim) {
+		if err := c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFrom(stored)); err != nil {
```
jonathan-innis, line 184: Can you make sure this is with the OptimisticLock? We're messing with the status condition so we need to make sure that we aren't overriding previous changes to status conditions

pkg/controllers/nodeclaim/lifecycle/controller.go
rschalo: We can fix with `nodeClaim.StatusConditions().List()`

pkg/controllers/nodeclaim/lifecycle/controller.go
rschalo: Thought about it but is it needed here given that this nodeclaim is being deleted? 

The other lifecycle status conditions won't race because we enter finalize before the other lifecycle methods reconcile and the other status conditions that would be added include the ones from Disruption, Drift, Consolidation, and Consistency checks.

Either way, I'm fine adding it.

jonathan-innis: I think it's safer to include it generally -- it's possible that we build something down the line that updates during the delete and we'll have this race because we forgot that we didn't optimistic lock here

jonathan-innis (approved):
/lgtm
/approve

Issue #2045, opened 2025-03-03
Option for KWOK additional `AllowUndefined` labels


alec-rabold:
### Description

**What problem are you trying to solve?**

When using KWOK as a simulation environment for other cloudproviders like AWS, I think there's some friction in syncing up the NodeSelector Labels that are allowed to be undefined for each; for example:

-  `karpenter.k8s.aws/instance-cpu` --> `karpenter.kwok.aws/instance-cpu`
-  `karpenter.k8s.aws/instance-memory` --> `karpenter.kwok.aws/instance-memory`
- `karpenter.k8s.aws/instance-local-nvme` --> custom requirement in Offering
etc.

Most of the labels I'm interested in have a KWOK counterpart (like those above) and it's not too difficult to translate those from AWS NodePools into KWOK NodePools, but for certain labels like `karpenter.k8s.aws/instance-generation` it's a bit harder since IIUC specifying that in the instance offering would mean that a Pod needs to explicitly set it.

I'm wondering if it makes sense to expose an [option in the KWOK provider](https://github.com/kubernetes-sigs/karpenter/blob/2ddac1074ec257661970b9e1a4b3308d418e3550/kwok/options/options.go#L41-L43) to allow additional requirements/labels to be undefined? 

(I'm assuming the `AllowUndefined` labels between KWOK<-->AWS is the issue but not 100% sure)

This is the `NodePool` and `Instance Type`/`Offering` I was testing with:

<details>
<summary>Instance Type</summary>
<br>

```JSON
[
  {
    "name": "m6a.12xlarge",
    "offerings": [
      {
        "Price": 2.0736,
        "Available": true,
        "Requirements": [
          {
            "key": "karpenter.sh/capacity-type",
            "operator": "In",
            "values": ["on-demand"]
          },
          {
            "key": "topology.kubernetes.io/zone",
            "operator": "In",
            "values": ["us-west-2a", "us-west-2b", "us-west-2c"]
          },
          {
            "key": "kubernetes.io/arch",
            "operator": "In",
            "values": ["amd64"]
          },
          { "key": "kubernetes.io/os", "operator": "In", "values": ["linux"] },
          {
            "key": "karpenter.k8s.aws/instance-family",
            "operator": "In",
            "values": ["m6"]
          },
          {
            "key": "karpenter.k8s.aws/instance-category",
            "operator": "In",
            "values": ["m"]
          },
          {
            "key": "karpenter.k8s.aws/instance-size",
[...]

PR #2053, opened 2025-03-05
feat: add requirement keys of KWOK instance offerings to well-known labels

alec-rabold:
**Description**

Adds the keys of all requirements for each instance-type offering to the list of Well Known Labels. 

This makes it easier to run simulations with non-KWOK NodePools (e.g. using KWOK to simulate how Karpenter would autoscale under certain circumstances in an AWS environment).

Issue: https://github.com/kubernetes-sigs/karpenter/issues/2045

Tested by running locally against a [KWOK all-in-one](https://kwok.sigs.k8s.io/docs/user/all-in-one-image) environment (pod). 

---

### DIFF
--- a/kwok/cloudprovider/cloudprovider.go
+++ b/kwok/cloudprovider/cloudprovider.go
@@ -222,8 +222,6 @@ func addInstanceLabels(labels map[string]string, instanceType *cloudprovider.Ins
 	// Randomly add each new node to one of the pre-created kwokPartitions.
 
 	ret[v1alpha1.KwokPartitionLabelKey] = lo.Sample(kwokPartitions)
-	ret[v1.CapacityTypeLabelKey] = offering.Requirements.Get(v1.CapacityTypeLabelKey).Any()
-	ret[corev1.LabelTopologyZone] = offering.Requirements.Get(corev1.LabelTopologyZone).Any()
 	ret[corev1.LabelHostname] = nodeClaim.Name
 
 	ret[v1alpha1.KwokLabelKey] = v1alpha1.KwokLabelValue
--- a/kwok/cloudprovider/cloudprovider.go
+++ b/kwok/cloudprovider/cloudprovider.go
@@ -211,6 +211,11 @@ func addInstanceLabels(labels map[string]string, instanceType *cloudprovider.Ins
 			ret[r.Key] = r.Values()[0]
 		}
 	}
+	for _, r := range offering.Requirements {
+		if r.Len() == 1 && r.Operator() == corev1.NodeSelectorOpIn {
+			ret[r.Key] = r.Values()[0]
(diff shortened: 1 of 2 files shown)


kwok/cloudprovider/helpers.go
```diff
@@ -169,6 +176,12 @@ func newInstanceType(options InstanceTypeOptions) *cloudprovider.InstanceType {
 		return req.Values
 	})))
 
+	for _, offering := range options.Offerings {
```
jonathan-innis, line 179: It's not just offering requirements, right? It's also the instance type requirements in general? I think we need both

kwok/cloudprovider/cloudprovider.go
```diff
@@ -211,6 +211,11 @@ func addInstanceLabels(labels map[string]string, instanceType *cloudprovider.Ins
 			ret[r.Key] = r.Values()[0]
 		}
 	}
+	for _, r := range offering.Requirements {
```
jonathan-innis, line 214: I think this bit means we can drop some of the static things that we are setting around zone and capacity type below

kwok/cloudprovider/helpers.go
```diff
@@ -139,6 +139,13 @@ func setDefaultOptions(opts InstanceTypeOptions) InstanceTypeOptions {
 		v1alpha1.InstanceCPULabelKey:    cpu,
 		v1alpha1.InstanceMemoryLabelKey: memory,
 	}
+	for _, offering := range opts.Offerings {
```
jonathan-innis, line 142: I'm a little confused by this update -- what is this trying to achieve?

alec-rabold: I originally thought this was needed to get the labels added onto the resulting `Node` but looks like that's not correct - removed in https://github.com/kubernetes-sigs/karpenter/pull/2053/commits/ee215df7c732a9ed6e404d33596291fdb252ea66!

kwok/cloudprovider/helpers.go
alec-rabold: Not sure I'm familiar enough to fully follow; just pushed a commit now to add requirements from both `[KWOK]offering.Requirements` and `[KWOK]offering.Offering.Requirements`: https://github.com/kubernetes-sigs/karpenter/blob/ee215df7c732a9ed6e404d33596291fdb252ea66/kwok/cloudprovider/helpers.go#L172-L179

Is this what you mean or something else?

jonathan-innis: Oh, I think what confused me is that our definition of "instance types" is called "offerings" in the options. What you did makes sense

jonathan-innis (approved):
/lgtm
/approve

PR #2054, opened 2025-03-05
feat(nodeclaim): use PodScheduled's status for lastPodEventTime

### LINKED ISSUE #1965: Pod events controllers should use PodScheduled's `lastTransitionTime` to mark when pod events occur
We use our in-memory view of the pod cache right now to mark that pods have been bound to nodes or that they have initiated deletion; however, this is just based on us seeing the events -- it would be nice if we could rely on pod status conditions for this information.

Also, with this change, we should remove our need to watch Create events when Karpenter first starts-up which currently removes the Consolidatable status condition across all NodeClaims across the board due to a reconciliation of the Pod being triggered on Create

flavono123:
Fixes #1965 

**Description**

use PodScheduled's lastTransitionTime for NodeClaim.LastPodEventTime
actually, only for the pod is bound to node so it still has same timestamp value for terminal and terminating
(should we follow deletionTimestamp or some risks are if then?)
the mitigations by dedupeTimestamp are removed
filter for registering more precise only for the PodScheduled has changed 

### DIFF
--- a/pkg/controllers/nodeclaim/podevents/controller.go
+++ b/pkg/controllers/nodeclaim/podevents/controller.go
@@ -32,18 +32,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
+	"github.com/samber/lo"
+
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
 	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
 	podutils "sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
-// dedupeTimeout is 10 seconds to reduce the number of writes to the APIServer, since pod scheduling and deletion events are very frequent.
-// The smaller this value is, the more writes Karpenter will make in a busy cluster. This timeout is intentionally smaller than the consolidation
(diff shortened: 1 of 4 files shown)


flavono123:
https://github.com/kubernetes-sigs/karpenter/actions/runs/14214081694/job/39826657535?pr=2054#step:5:642
the failure of ci seems to be flaky but no idea why

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -81,14 +78,28 @@ func (c *Controller) Reconcile(ctx context.Context, pod *corev1.Pod) (reconcile.
...
+	// If we don't have a valid time, skip
+	if eventTime.IsZero() {
 		return reconcile.Result{}, nil
 	}
 
 	// otherwise, set the pod event time to now
 	stored := nc.DeepCopy()
-	nc.Status.LastPodEventTime.Time = c.clock.Now()
+
+	nc.Status.LastPodEventTime.Time = eventTime
```
jonathan-innis, line 102: This needs to be max otherwise it's going to oscillate around as different pods get reconciled. Also, I think still applying a dedupe timeout makes sense -- basically, if the current stored time is within (say 10s) of the new time, then don't update it

jonathan-innis:
@flavono123 Just checking in on this one!

pkg/controllers/nodeclaim/podevents/controller.go
flavono123: ammended

PR #2055, opened 2025-03-06
fix: update CEL validation for expireAfter and consolidateAfter fields to block invalid values like 1hr

saurav-agarwalla:
Fixes #N/A 

**Description**
Current CEL validation allows values like `1hr` to be set in the spec which causes the resources to be stuck in a weird state.

Used this spec for nodepool:
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: malformed
spec:
  disruption:
    budgets:
    - nodes: 10%
    consolidateAfter: 1hr
  template:
    spec:
      expireAfter: 12h
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: default
      requirements:
        operator: In
        values:
        - on-demand
      terminationGracePeriod: 24h0m0s
```

**Before the change:** 
```
~ k apply -f malformed-nodepool.yaml
nodepool.karpenter.sh/malformed created
```

```
~ k get nodepool
NAME        NODECLASS   NODES   READY   AGE
default     default     0       True    8d
malformed   default                     89s
```

In the logs:
```
{"level":"INFO","time":"2025-03-06T18:14:19.413Z","logger":"controller","caller":"cache/reflector.go:370","message":"k8s.io/client-go@v0.32.2/tools/cache/reflector.go:251: failed to list *v1.NodePool: time: unknown unit \"hr\" in duration \"1hr\"","commit":"974a323-dirty"}
```

**After the change:** 
```
 ~ k apply -f malformed-nodepool.yaml
The NodePool "malformed" is invalid: spec.disruption.consolidateAfter: Invalid value: "1hr": spec.disruption.consolidateAfter in body should match '^(([0-9]+(s|m|h))+|Never)$'
```

I can't replicate this in the unit tests since the tests parse the duration so testing that kind of makes it pointless since what we really want to test is CEL server-side validation catching it like it did above. I've added other unit tests to ensure that there's no regression.

**Regex testing:**

Before: https://regex101.com/r/C8F0Wf/1
After: https://regex101.com/r/nHnhl2/1

### DIFF
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -83,7 +83,7 @@ spec:
                     before terminating a node, measured from when the node is created. This
                     is useful to implement features like eventually consistent node upgrade,
                     memory leak protection, and disruption testing.
-                  pattern: ^(([0-9]+(s|m|h))+)|(Never)$
+                  pattern: ^(([0-9]+(s|m|h))+|Never)$
                   type: string
                 nodeClassRef:
                   description: NodeClassRef is a reference to an object that defines provider specific configuration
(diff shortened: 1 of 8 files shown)


jonathan-innis (approved):
/lgtm
/approve

PR #2088, opened 2025-03-20
feat: Add Global Default Termination Grace Period Variable 

engedaam:
Fixes #N/A 

**Description**
- Adding a global termination grace period to set on all NodeClaims that are owned by NodePools that don't add contain a termination grace period.
- This would be a mechanisms for cloud provider to enforce a termination grace period on all nodeClaims provided by Karpenter
- This would allow CloudProvider to avoid manipulating the nodeClaim validation rules prior to setting termination grace period 

- `make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/suite_test.go
+++ b/pkg/controllers/provisioning/suite_test.go
@@ -98,6 +98,7 @@ var _ = BeforeEach(func() {
 	}
 	fakeClock.SetTime(time.Now())
 	state.PodSchedulingDecisionSeconds.Reset()
+	pscheduling.DefaultTerminationGracePeriod = nil
 })
 
 var _ = AfterSuite(func() {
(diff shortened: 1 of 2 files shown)


pkg/controllers/provisioning/provisioner.go
```diff
@@ -53,6 +54,8 @@ import (
 	"sigs.k8s.io/karpenter/pkg/utils/pretty"
 )
 
+var GlobalDefaultTerminationGracePeriod *metav1.Duration = nil
```
jmdeal, line 57: nit: want to leave a small comment here on the use case?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -382,6 +385,9 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 		return "", err
 	}
 	nodeClaim := n.ToNodeClaim()
+	if nodeClaim.Spec.TerminationGracePeriod == nil {
```
jmdeal, line 388: nit: would it make more sense to handle this in `ToNodeClaim` rather than in `Create`? That seems more natural to me since it's where we insert other implicit fields (e.g. the owner ref).

pkg/controllers/provisioning/suite_test.go
```diff
@@ -220,6 +220,43 @@ var _ = Describe("Provisioning", func() {
 		Expect(len(nodes.Items)).To(Equal(1))
 		ExpectScheduled(ctx, env.Client, pod)
 	})
+	It("should expect terminationGracePeriod to be nil when nodepool and global value is not set", func() {
+		nodePool := test.NodePool()
+		provisioning.GlobalDefaultTerminationGracePeriod = &metav1.Duration{Duration: 98 * time.Hour}
```
jmdeal, line 225: We should make sure we're resetting this to `nil` in the `BeforeEach` otherwise it leaks out of this test.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -220,6 +220,43 @@ var _ = Describe("Provisioning", func() {
 		Expect(len(nodes.Items)).To(Equal(1))
 		ExpectScheduled(ctx, env.Client, pod)
 	})
+	It("should expect terminationGracePeriod to be nil when nodepool and global value is not set", func() {
```
jmdeal, line 223: It doesn't look like this description is right, is it? We're setting the global to 98h in this test.

pkg/controllers/provisioning/suite_test.go
engedaam: The NodePools are getting created in each tests, we don't need to reset the value

pkg/controllers/provisioning/provisioner.go
engedaam: Sure, I can add that

pkg/controllers/provisioning/provisioner.go
engedaam: Yeah, that's fair I can shift it there

pkg/controllers/provisioning/suite_test.go
engedaam: Updated the description

pkg/controllers/provisioning/suite_test.go
```diff
@@ -220,6 +220,42 @@ var _ = Describe("Provisioning", func() {
...
+	It("should expect nodeclaim terminationGracePeriod to be the nil when the nodepool terminationGracePeriod and gloabl terminationGracePeriod is not set", func() {
+		nodePool := test.NodePool()
+		ExpectApplied(ctx, env.Client, nodePool)
+		pod := test.UnschedulablePod()
+		ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+		ExpectScheduled(ctx, env.Client, pod)
+		nodeClaims := &v1.NodeClaimList{}
+		Expect(env.Client.List(ctx, nodeClaims)).To(Succeed())
+		Expect(len(nodeClaims.Items)).To(Equal(1))
+		Expect(nodeClaims.Items[0].Spec.TerminationGracePeriod.Duration).To(BeNumerically("==", 78*time.Hour))
```
jmdeal, line 244: This should be nil right?

engedaam: Yeah, was running the test in parallel, updated now

pkg/controllers/provisioning/suite_test.go
engedaam: I added the `nil` to the `BeforeEach`

Issue #1772, opened 2024-10-23
Race condition with node startupTaints being restored


dpiddock:
### Description

**Observed Behavior**:
Karpenter restores the `startupTaints` to a node if it is removed too quickly at node startup. This results in a node being unusable. Node also never reaches a ready state, so Karpenter refuses to remove it: `Cannot disrupt Node: state node isn't initialized`

From AWS CloudWatch logs insights:
* created state:
  ```
  requestObject.spec.taints.0 {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
  requestObject.spec.taints.1 {"key":"efs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
  requestObject.spec.taints.2 {"key":"karpenter.sh/unregistered","effect":"NoExecute"}
  requestObject.spec.taints.3 {"key":"node.cloudprovider.kubernetes.io/uninitialized","value":"true","effect":"NoSchedule"}
  requestReceivedTimestamp 2024-10-22T13:21:21.946552Z
  ```
* node-controller updates some taints
* efs-csi removes its taint
  ```
  requestObject.0.op                test
  requestObject.0.path              /spec/taints
  requestObject.0.value.0.effect    NoExecute
  requestObject.0.value.0.key       ebs.csi.aws.com/agent-not-ready
  requestObject.0.value.1.effect    NoExecute
  requestObject.0.value.1.key       efs.csi.aws.com/agent-not-ready
  requestObject.0.value.2.effect    NoExecute
  requestObject.0.value.2.key       karpenter.sh/unregistered
  requestObject.0.value.3.effect    NoSchedule
  requestObject.0.value.3.key       node.kubernetes.io/not-ready
  requestObject.1.op                replace
  requestObject.1.path              /spec/taints
  requestObject.1.value.0.effect    NoExecute
  requestObject.1.value.0.key       ebs.csi.aws.com/agent-not-ready
  requestObject.1.value.1.effect    NoExecute
  requestObject.1.value.1.key       karpenter.sh/unregistered
  requestObject.1.value.2.effect    NoSchedule
  requestObject.1.value.2.key       node.kubernetes.io/not-ready
  requestReceivedTimestamp          2024-10-22T13:21:32.092879Z
  ...
  responseObject.spec.taints.0 {"key":"ebs.csi.aws.com/agent-not-ready","effect":"NoExecute"}
[...]

PR #2091, opened 2025-03-21
fix: Remove code that syncs taints from nodeclaims to nodes

JeremyBolster:
**Description**

This commit removes the code that syncs the taints from nodeclaims to nodes. This changes the behaviour of Karpenter as it is now expected that the providers will be responsible for setting taints on nodes. By giving up control of taints to the Karpenter provider, taints can be set via the kubelet `--register-with-taints` field. This ensures that taints exist as soon as the node starts. The Karpenter node registration taint is still managed/removed by Karpenter and labels and annotations will still be synchronized here.

This commit is required to work around daemonsets, such as the EBS and EFS drivers, that do not continuously reconcile. It is common practice for these daemonsets to tolerate all taints, and thus they will race with this section of code to remove the taint. If these daemonsets win then this code re-applies the taint that they removed. This leads to a node that will always be tainted.

Fixes: https://github.com/kubernetes-sigs/karpenter/issues/1772

I added `!` as this changes the behaviour of karpenter that could effect karpenter providers. Providers would need to ensure that their code registers taints on the node, rather than relying on this section of code to add taints.

I tested this in an EKS cluster with the karpenter aws provider.
I could prove the race condition by forcing it to occur with the following code.
```
func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, node *corev1.Node) error {
+	time.Sleep(30 * time.Second)
```
I could prove the race condition was no longer triggered after this change by adding in the same sleep statement.

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/initialization_test.go
+++ b/pkg/controllers/nodeclaim/lifecycle/initialization_test.go
@@ -366,6 +366,18 @@ var _ = Describe("Initialization", func() {
 		Expect(ExpectStatusConditionExists(nodeClaim, v1.ConditionTypeInitialized).Status).To(Equal(metav1.ConditionTrue))
 	})
 	It("should not consider the Node to be initialized when all startupTaints aren't removed", func() {
+		startupTaints := []corev1.Taint{
+			{
+				Key:    "custom-startup-taint",
+				Effect: corev1.TaintEffectNoSchedule,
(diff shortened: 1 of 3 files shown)


engedaam:
Can we close out this PR in favor of this PR? https://github.com/kubernetes-sigs/karpenter/pull/2125

PR #2106, opened 2025-03-31
fix: don't need to deepcopy nodeclaim when lastPodEvent has not been set before the timeout

rambohe-ch:
Fixes #N/A 

**Description**

we can deep copy nodeclaim when we need to change it. because deepcopy will not be executed if lastPodEvent has not been set before the timeout.


rschalo (approved):
Thanks for catching this!

/lgtm

jonathan-innis (approved):
/approve

PR #2110, opened 2025-04-01
perf: Actively store NodePool resources in cluster state

jonathan-innis:
Fixes #N/A 

**Description**

The counter controller is doing _a lot_ of repeated work by continually pulling nodes from cluster state to calculate the number of resources that a NodePool has deployed at any point in time. Rather than rebuilding the sum every time, we can use a diff-based approach where we just take the difference for resources as we make updates to the StateNodes. Likewise, when a StateNode is deleted, we just subtract the resources from the retained NodePool resources that is now stored in cluster state.

This makes our counter controller logic way more efficient and reduces CPU cycles that were given to it.

`make presubmit`

### DIFF
--- a/pkg/controllers/nodepool/counter/controller.go
+++ b/pkg/controllers/nodepool/counter/controller.go
@@ -18,9 +18,9 @@ package counter
 
 import (
 	"context"
-	"fmt"
 	"time"
 
+	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/resource"
@@ -47,14 +47,12 @@ type Controller struct {
 	cluster       *state.Cluster
 }
 
-var ResourceNode = corev1.ResourceName("nodes")
-
 var BaseResources = corev1.ResourceList{
 	corev1.ResourceCPU:              resource.MustParse("0"),
 	corev1.ResourceMemory:           resource.MustParse("0"),
 	corev1.ResourcePods:             resource.MustParse("0"),
 	corev1.ResourceEphemeralStorage: resource.MustParse("0"),
-	ResourceNode:                    resource.MustParse("0"),
+	resources.Node:                  resource.MustParse("0"),
 }
 
 // NewController is a constructor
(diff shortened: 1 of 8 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -609,22 +627,75 @@ func (c *Cluster) newStateFromNode(ctx context.Context, node *corev1.Node, oldNo
...
+		newResources[resources.Node] = resource.MustParse("1")
+	}
+	if _, ok := c.nodePoolResources[newNodePoolName]; !ok && newNodePoolName != "" {
+		c.nodePoolResources[newNodePoolName] = corev1.ResourceList{}
+	}
+	if oldNodePoolName != "" {
+		for resourceName, quantity := range oldResources {
+			current := c.nodePoolResources[oldNodePoolName][resourceName]
+			current.Sub(quantity)
+			c.nodePoolResources[oldNodePoolName][resourceName] = current
```
jmdeal, line 675: nit: this is an in-place update, any reason we can't inline it?
```suggestion
			c.nodePoolResources[oldNodePoolName][resourceName].Sub(quantity)
```

pkg/controllers/state/cluster.go
```diff
@@ -609,22 +627,75 @@ func (c *Cluster) newStateFromNode(ctx context.Context, node *corev1.Node, oldNo
...
+		for _, v := range c.nodePoolResources[name] {
+			if !v.IsZero() {
+				allZero = false
+				break
+			}
+		}
+		if allZero {
+			delete(c.nodePoolResources, name)
+		}
+	}
```
jmdeal, line 696: This is for garbage collection right, to ensure `nodePoolResources` doesn't grow unbounded? I think this seems reasonable, we should always scale to zero if a NodePool is deleted. I could also see us just having a function called by the NodePool informer controller upon deletion though, that's a little simpler and more obviously bulletproof. I don't feel strongly either way though.

pkg/controllers/nodepool/counter/controller.go
```diff
@@ -81,48 +79,19 @@ func (c *Controller) Reconcile(ctx context.Context, nodePool *v1.NodePool) (reco
...
 		Named("nodepool.counter").
-		For(&v1.NodePool{}, builder.WithPredicates(nodepoolutils.IsManagedPredicateFuncs(c.cloudProvider))).
-		Watches(&v1.NodeClaim{}, nodepoolutils.NodeClaimEventHandler(), builder.WithPredicates(predicate.NewPredicateFuncs(func(o client.Object) bool {
-			// Add a predicate here to filter out NodeClaims triggering reconciliation if they haven't resolved their
-			// providerID (and therefore, their resources)
-			nc := o.(*v1.NodeClaim)
-			return nc.Status.ProviderID != ""
-		}))).
-		Watches(&corev1.Node{}, nodepoolutils.NodeEventHandler()).
+		For(&v1.NodePool{}, builder.WithPredicates(nodepoolutils.IsManagedPredicateFuncs(c.cloudProvider), predicate.GenerationChangedPredicate{})).
```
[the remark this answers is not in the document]
jonathan-innis: I'm sort-of with you, sure -- I don't really care one way or the other

pkg/controllers/state/cluster.go
jonathan-innis: I thought about that but the NodePool informer could be called before the NodeClaim gets removed and I didn't want to have to handle the NodePool getting created and deleted -- if you think about it, coordinating the two actually becomes quite hard. e.g. How do you handle updates that happen after the NodePool has been deleted -- do you only take updates if the key exists, but then you have to hang key creation on the NodePool being created and then hydrate that off of the existing resources

pkg/controllers/state/cluster.go
jonathan-innis: `Cannot call a pointer method on 'c.nodePoolResources[oldNodePoolName][resourceName]`. This doesn't work because the element isn't a pointer so you do have to retrieve, update, and assign

pkg/controllers/provisioning/provisioner.go
```diff
@@ -378,7 +378,7 @@ func (p *Provisioner) Create(ctx context.Context, n *scheduler.NodeClaim, opts .
 	if err := p.kubeClient.Get(ctx, types.NamespacedName{Name: n.NodePoolName}, latest); err != nil {
 		return "", fmt.Errorf("getting current resource usage, %w", err)
 	}
-	if err := latest.Spec.Limits.ExceededBy(latest.Status.Resources); err != nil {
+	if err := latest.Spec.Limits.ExceededBy(p.cluster.NodePoolResourcesFor(n.NodePoolName)); err != nil {
```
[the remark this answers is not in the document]
jonathan-innis: > Shouldn't we be adding the resources for the NodeClaim to the resources stored in cluster state

This is a good point. I'll update this as a follow-on since it's a different concern, but I agree with you that this is a bug

pkg/controllers/state/cluster.go
jmdeal: Discussed some more offline and I agree, the coordination is more difficult than the current approach

PR #2119, opened 2025-04-03
perf: Relax preferences all the way per-pod

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that we relax a pod's preferences all the way before we consider the next pod. This change will have us add the original pod back to the queue and attempt it later if we don't succeed on the first attempt. Handling preferences this way is important because we want to tee-up a scheduling timeout change to allow us to react faster to pending pods during provisioning.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -268,7 +268,6 @@ func (r Results) TruncateInstanceTypes(maxInstanceTypes int) Results {
 	return r
 }
 
-//nolint:gocyclo
 func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, error) {
 	defer metrics.Measure(DurationSeconds, map[string]string{ControllerLabel: injection.GetControllerName(ctx)})()
 	// We loop trying to schedule unschedulable pods as long as we are making progress.  This solves a few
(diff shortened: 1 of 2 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -305,28 +305,36 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
...
 		// Schedule to existing nodes or create a new node
-		if podErrors[pod] = s.add(ctx, pod); podErrors[pod] == nil {
-			delete(podErrors, pod)
+		p := pod.DeepCopy()
+		for {
+			if podErrors[pod] = s.add(ctx, pod); podErrors[pod] == nil {
+				delete(podErrors, pod)
+				break
+			}
+			if IsReservedOfferingError(podErrors[pod]) {
```
jmdeal, line 315: Can we retain the previous comment for why we don't relax with reserved offerings?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -305,28 +305,36 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
...
+				break
+			}
+			if relaxed := s.preferences.Relax(ctx, pod); !relaxed {
+				break
+			}
+			if err := s.topology.Update(ctx, pod); err != nil && !errors.Is(err, context.DeadlineExceeded) {
```
jmdeal, line 321: It doesn't need to be coupled to this PR, but we don't need to rebuild the topology for every type of preference we relax, right (e.g. taints / node affinity)?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -305,28 +305,36 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
...
-				if err := s.topology.Update(ctx, pod); err != nil && !errors.Is(err, context.DeadlineExceeded) {
-					log.FromContext(ctx).Error(err, "failed updating topology")
-				}
-				// Update the cached podData since the pod was relaxed and it could have changed its requirement set
-				s.updateCachedPodData(pod)
+		// We relax the pod all the way the first time we see it
+		// If we don't schedule it, we store the original pod (with preferences)
+		// in the queue and give ourselves another chance to schedule it later
+		if _, ok = podErrors[pod]; ok {
+			*pod = *p // Ensure that the pod pointer remains the same when we push it back onto the queue
```
jmdeal, line 332: This is because we use the pointer as a key in a few maps, right? Worth adding that to the comment? Might also be worth using the pod's UID rather than the pointer in those cases to prevent this type of subtle bug from popping up in the future.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -305,28 +305,38 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
 		if !ok {
 			break
 		}
-
 		// Schedule to existing nodes or create a new node
-		if podErrors[pod] = s.add(ctx, pod); podErrors[pod] == nil {
-			delete(podErrors, pod)
+		p := pod.DeepCopy()
+		for {
```
jmdeal, line 310: Thoughts on breaking this out into a separate function?

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: Right -- we could be smarter for sure about it -- I'll take a note to add that as a perf improvement

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: Good call -- added the change

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -336,10 +329,38 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
 	return Results{
 		NewNodeClaims: s.newNodeClaims,
 		ExistingNodes: s.existingNodes,
-		PodErrors:     podErrors,
+		PodErrors: lo.MapEntries(podErrors, func(id types.UID, err error) (*corev1.Pod, error) {
```
jmdeal, line 331: I'm thinking we should update this to use UID as well - I don't think it needs to be scoped to this PR, but we could open an issue.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -336,10 +328,38 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
...
+			return podUIDToPod[id], err
+		}),
 	}, ctx.Err()
 }
 
+func (s *Scheduler) trySchedule(ctx context.Context, p *corev1.Pod, podErrors map[types.UID]error) {
```
jmdeal, line 337: nit: I don't think there's any reason we need to pass in the full set of pod errors right? Can we update the interface to just return an error, and delete the key from pod errors if we return nil up the stack?

pkg/controllers/provisioning/scheduling/scheduler.go
jonathan-innis: It's currently not possible based on how we handle pod errors in disruption -- agreed that for now it's fine -- and honestly, I don't think it hurts a ton to use the pod as the actual key

jmdeal: Probably not, we just have to be careful if we're every making copies.

PR #2121, opened 2025-04-04
perf: Timeout scheduling simulation after 1m to schedule pods/nodes

jonathan-innis:
Fixes #N/A 

**Description**

Timeout the scheduling simulation after 1m during provisioning to ensure we make faster progress when scheduling pods/nodes. 

To validate this change, I ran simulations of large scale-ups with and without the timeout. I found that, when running the provisioning loops with the timeouts, we were significantly faster (2x) compared to running without it.

### Before PR (32m)

Screenshot 2025-03-11 at 1 07 59 AM

### After PR (16m)

Screenshot 2025-03-23 at 9 50 39 PM

`make presubmit`
- Running scale testing with the Kwok cloudprovider

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -285,20 +285,10 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
 	q := NewQueue(pods, s.cachedPodData)
 
 	startTime := s.clock.Now()
-	lastLogTime := s.clock.Now()
-	batchSize := len(q.pods)
 	for {
-		if ctx.Err() != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -333,6 +327,9 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
 
 func (s *Scheduler) trySchedule(ctx context.Context, p *corev1.Pod) error {
 	for {
+		if ctx.Err() != nil {
```
jmdeal, line 330: We need to catch this and return at the call site now right, rather than emplacing it as a pod error?

jonathan-innis: Good catch :)

PR #2122, opened 2025-04-04
feat: Add `PreferencePolicy` as an environment variable option

jonathan-innis:
Fixes #N/A 

**Description**

This change adds an environment variable option called `PreferencePolicy` with the following definitions:

- `Respect`: Karpenter will attempt to respect all preferences as requirements. If Karpenter cannot simulate pods against the node given these requirements, it will begin removing preferences from the pod until it is able to schedule the pod. `Respect` can also block consolidation more frequently due to the way that the preference relaxation ordering is orchestrated.
- `Ignore`: Karpenter will ignore all preferences when it performs its scheduling simulations. In general, this is a more efficient way to run Karpenter because Karpenter will have to give less effort to satisfying all of the preferences on the application pods. In addition, it's currently not guaranteed that even if Karpenter does respect all of the preferences, then they will still be respected by the kube-scheduler. In particular, when considering things like `preferredDuringSchedulingIgnoredDuringExecution` anti-affinity preferences and `ScheduleAnyways` topologySpreadConstraints, it's very timing-dependent and non-deterministic on whether or not the kube-scheduler will see the pods in time to respect the preferences.

### CI Benchmark Testing Results

```
============== Generic Pods ==============
1 pods      1 nodes     162.759µs per scheduling      162.759µs per pod
50 pods     10 nodes    38.027579ms per scheduling    760.551µs per pod
100 pods    20 nodes    87.226431ms per scheduling    872.264µs per pod
500 pods    100 nodes   406.788638ms per scheduling   813.577µs per pod
1000 pods   200 nodes   589.950666ms per scheduling   589.95µs per pod
1500 pods   300 nodes   1.902119667s per scheduling   1.268079ms per pod
2000 pods   400 nodes   2.537366s per scheduling      1.268683ms per pod
5000 pods   1000 nodes  5.126445375s per scheduling   1.025289ms per pod
10000 pods  2000 nodes  20.038828125s per scheduling  2.003882ms per pod
20000 pods  4000 nodes  50.439955s per scheduling     2.521997ms per pod
============== Preference Pods ==============
4000 pods  4000 nodes  51.578981334s per scheduling  12.894745ms per pod
4000 pods  6 nodes     711.665062ms per scheduling   177.916µs per pod
scheduled 48151 against 12037 nodes in total in 2m15.715480943s 354.793717 pods/sec
```

### Scale Testing

I validated this change by creating a testing workload that had a preference that was unsatisfiable and therefore would have to be relaxed with `PreferencePolicy=Respect`. By setting `PreferencePolicy=Ignore`, we wouldn't capture these preferences in our cluster topology or consider it when adding pods to nodes.

#### Test Workload

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: inflate
  namespace: default
spec:
  replicas: 10000
  selector:
    matchLabels:
      app: inflate
  template:
    metadata:
      labels:
        app: inflate
    spec:
      affinity: 
        nodeAffinity:
            nodeSelectorTerms:
            - matchExpressions:
                operator: In
                values:
                - test-zone-a
                - test-zone-b
                - test-zone-c
                - test-zone-d
        podAntiAffinity:
              labelSelector:
                matchLabels:
                  app: inflate
          # Can't satisfy this term
          preferredDuringSchedulingIgnoredDuringExecution:
            - weight: 100
              podAffinityTerm:
                labelSelector:
                  matchLabels:
                    app: inflate
      topologySpreadConstraints:
      - maxSkew: 1
        labelSelector:
          matchLabels:
            app: inflate
      containers:
        - name: inflate
          image: public.ecr.aws/ubuntu/ubuntu:22.04_stable
          command: ["/bin/sh", "-ec", "while :; do echo '.'; sleep 5 ; done"]
          resources:
            requests:
              memory: 100Mi
              cpu: 1
      terminationGracePeriodSeconds: 0
```

#### `PreferencePolicy=Respect` (19m)

Screenshot 2025-03-25 at 12 00 12 AM

#### `PreferencePolicy=Ignore` (7m)

Screenshot 2025-03-26 at 12 25 06 AM

`make presubmit`

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -28,6 +28,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
+	"sigs.k8s.io/karpenter/pkg/operator/options"
+
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
@@ -92,7 +94,16 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 	}
 	pods = append(pods, deletingNodePods...)
 
-	scheduler, err := provisioner.NewScheduler(log.IntoContext(ctx, operatorlogging.NopLogger), pods, stateNodes)
+	var opts []scheduling.Options
+	if options.FromContext(ctx).PreferencePolicy == options.PreferencePolicyIgnore {
+		opts = append(opts, scheduling.IgnorePreferences)
+	}
+	scheduler, err := provisioner.NewScheduler(
+		log.IntoContext(ctx, operatorlogging.NopLogger),
+		pods,
+		stateNodes,
+		opts...,
+	)
 	if err != nil {
 		return scheduling.Results{}, fmt.Errorf("creating scheduler, %w", err)
 	}
--- a/pkg/controllers/provisioning/scheduling/topology.go
+++ b/pkg/controllers/provisioning/scheduling/topology.go
@@ -43,7 +43,8 @@ import (
 )
 
 type Topology struct {
-	kubeClient client.Client
+	kubeClient       client.Client
+	preferencePolicy PreferencePolicy
 	// Both the topologyGroups and inverseTopologies are maps of the hash from TopologyGroup.Hash() to the topology group
 	// itself. This is used to allow us to store one topology group that tracks the topology of many pods instead of
 	// having a 1<->1 mapping between topology groups and pods owned/selected by that group.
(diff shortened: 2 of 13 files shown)


pkg/controllers/disruption/helpers.go
```diff
@@ -79,7 +81,12 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 		pods = append(pods, n.reschedulablePods...)
 	}
 	pods = append(pods, deletingNodePods...)
-	scheduler, err := provisioner.NewScheduler(log.IntoContext(ctx, operatorlogging.NopLogger), pods, stateNodes)
+	scheduler, err := provisioner.NewScheduler(
+		log.IntoContext(ctx, operatorlogging.NopLogger),
+		pods,
+		stateNodes,
+		lo.Ternary(options.FromContext(ctx).PreferencePolicy == options.PreferencePolicyIgnore, scheduling.IgnorePreferences, nil),
```
rschalo, line 88: Is this more easily understood in scheduling code as a bool?

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -70,9 +71,11 @@ func NewTopology(
 	nodePools []*v1.NodePool,
 	instanceTypes map[string][]*cloudprovider.InstanceType,
 	pods []*corev1.Pod,
+	opts ...Options,
```
rschalo, line 74: Do we need all options passed?

jonathan-innis: I'd like to keep them just because it's a little more sensible than a boolean and it makes it have nice parity with the scheduler

pkg/controllers/disruption/helpers.go
jonathan-innis: Makes sense -- I'll make the update

jonathan-innis: I sort of like keeping the same parity with the options fields -- it also has nice parity with the current reserved capacity options for the scheduler.

rschalo (approved):
/lgtm

PR #2125, opened 2025-04-04
fix: Allow disabling syncing taints from nodeclaims to nodes (with node label)

JeremyBolster:
**Description**

This commit removes the code that syncs the taints from nodeclaims to nodes. This changes the behaviour of Karpenter as it is now expected that the providers will be responsible for setting taints on nodes. By giving up control of taints to the Karpenter provider, taints can be set via the kubelet `--register-with-taints` field. This ensures that taints exist as soon as the node starts. The Karpenter node registration taint is still managed/removed by Karpenter and labels and annotations will still be synchronized here.

This commit is required to work around daemonsets, such as the EBS and EFS drivers, that do not continuously reconcile. It is common practice for these daemonsets to tolerate all taints, and thus they will race with this section of code to remove the taint. If these daemonsets win then this code re-applies the taint that they removed. This leads to a node that will always be tainted.

Fixes: https://github.com/kubernetes-sigs/karpenter/issues/1772

I tested this in an EKS cluster with the karpenter aws provider.
I could prove the race condition by forcing it to occur with the following code.
```
func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, node *corev1.Node) error {
+	time.Sleep(30 * time.Second)
```
I could prove the race condition was no longer triggered after this change by adding in the same sleep statement.

This change adds a behaviour-preserving feature flag for those that still want the buggy behaviour enabled. Otherwise it is the same change as https://github.com/kubernetes-sigs/karpenter/pull/2091

### DIFF
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -51,6 +51,7 @@ const (
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey           = apis.Group + "/nodepool-hash-version"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
+	NodeClaimSyncTaintsAnnotationKey           = apis.Group + "/nodeclaim-sync-taints"
 )
 
 // Karpenter specific finalizers
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -38,10 +38,11 @@ const (
 
 // Karpenter specific domains and labels
 const (
-	NodePoolLabelKey        = apis.Group + "/nodepool"
-	NodeInitializedLabelKey = apis.Group + "/initialized"
-	NodeRegisteredLabelKey  = apis.Group + "/registered"
-	CapacityTypeLabelKey    = apis.Group + "/capacity-type"
+	NodePoolLabelKey            = apis.Group + "/nodepool"
+	NodeInitializedLabelKey     = apis.Group + "/initialized"
+	NodeRegisteredLabelKey      = apis.Group + "/registered"
+	NodeDoNotSyncTaintsLabelKey = apis.Group + "/nodeclaim-do-not-sync-taints"
+	CapacityTypeLabelKey        = apis.Group + "/capacity-type"
 )
 
 // Karpenter specific annotations
@@ -51,7 +52,6 @@ const (
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey           = apis.Group + "/nodepool-hash-version"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
-	NodeClaimSyncTaintsAnnotationKey           = apis.Group + "/nodeclaim-sync-taints"
 )
 
 // Karpenter specific finalizers
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -41,7 +41,7 @@ const (
 	NodePoolLabelKey            = apis.Group + "/nodepool"
 	NodeInitializedLabelKey     = apis.Group + "/initialized"
 	NodeRegisteredLabelKey      = apis.Group + "/registered"
-	NodeDoNotSyncTaintsLabelKey = apis.Group + "/nodeclaim-do-not-sync-taints"
+	NodeDoNotSyncTaintsLabelKey = apis.Group + "/do-not-sync-taints"
 	CapacityTypeLabelKey        = apis.Group + "/capacity-type"
 )
 
--- a/pkg/controllers/nodeclaim/lifecycle/initialization_test.go
+++ b/pkg/controllers/nodeclaim/lifecycle/initialization_test.go
@@ -366,18 +366,6 @@ var _ = Describe("Initialization", func() {
 		Expect(ExpectStatusConditionExists(nodeClaim, v1.ConditionTypeInitialized).Status).To(Equal(metav1.ConditionTrue))
 	})
 	It("should not consider the Node to be initialized when all startupTaints aren't removed", func() {
-		startupTaints := []corev1.Taint{
-			{
-				Key:    "custom-startup-taint",
-				Effect: corev1.TaintEffectNoSchedule,
-				Value:  "custom-startup-value",
-			},
-			{
-				Key:    "other-custom-startup-taint",
-				Effect: corev1.TaintEffectNoExecute,
-				Value:  "other-custom-startup-value",
-			},
-		}
 		nodeClaim := test.NodeClaim(v1.NodeClaim{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: map[string]string{
@@ -392,7 +380,18 @@ var _ = Describe("Initialization", func() {
 						corev1.ResourcePods:   resource.MustParse("5"),
 					},
 				},
-				StartupTaints: startupTaints,
+				StartupTaints: []corev1.Taint{
+					{
+						Key:    "custom-startup-taint",
+						Effect: corev1.TaintEffectNoSchedule,
+						Value:  "custom-startup-value",
+					},
+					{
+						Key:    "other-custom-startup-taint",
+						Effect: corev1.TaintEffectNoExecute,
+						Value:  "other-custom-startup-value",
+					},
+				},
 			},
 		})
 		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
@@ -411,13 +410,13 @@ var _ = Describe("Initialization", func() {
 				corev1.ResourceMemory: resource.MustParse("80Mi"),
 				corev1.ResourcePods:   resource.MustParse("110"),
 			},
-			Taints: append(startupTaints, v1.UnregisteredNoExecuteTaint),
+			Taints: []corev1.Taint{v1.UnregisteredNoExecuteTaint},
 		})
 		ExpectApplied(ctx, env.Client, node)
 		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
 		ExpectMakeNodesReady(ctx, env.Client, node) // Remove the not-ready taint
 
-		// Startup taints should be the only taints remaining
+		// Should add the startup taints to the node
 		ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
 		node = ExpectExists(ctx, env.Client, node)
 		Expect(node.Spec.Taints).To(ContainElements(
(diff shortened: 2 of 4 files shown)


pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -121,11 +120,17 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, no
...
+		// Sync all taints inside NodeClaim into the Node taints
+		node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.Taints)
+		node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.StartupTaints)
+	}
+
 	node.Annotations = lo.Assign(node.Annotations, nodeClaim.Annotations)
-	// Sync all taints inside NodeClaim into the Node taints
-	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.Taints)
-	node.Spec.Taints = scheduling.Taints(node.Spec.Taints).Merge(nodeClaim.Spec.StartupTaints)
+	// We do not sync the taints, unless the feature flag is enabled, from the nodeclaim as
```
engedaam, line 131: Can we update this message comment?

pkg/apis/v1/labels.go
```diff
@@ -51,6 +51,7 @@ const (
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey           = apis.Group + "/nodepool-hash-version"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
+	NodeClaimSyncTaintsAnnotationKey           = apis.Group + "/nodeclaim-sync-taints"
```
engedaam, line 54: This needs to be label, no? Kubelet configuration does not allow nodes to be registered with annotations

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -121,11 +120,17 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, no
 	controllerutil.AddFinalizer(node, v1.TerminationFinalizer)
 
 	node = nodeclaimutils.UpdateNodeOwnerReferences(nodeClaim, node)
-	node.Labels = lo.Assign(node.Labels, nodeClaim.Labels)
+
+	if _, ok := nodeClaim.Annotations[v1.NodeClaimSyncTaintsAnnotationKey]; ok {
```
engedaam, line 124: We need to reverse the polity, no? We only want to sync when the node does not have the label. This is meant to handle the case where the OS is Unknown to Karpenter and we are not able to add the values to the kubelet config

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -204,7 +204,7 @@ var _ = Describe("Registration", func() {
 			Expect(node.Annotations).To(HaveKeyWithValue(k, v))
 		}
 	})
-	It("should sync the taints to the Node when the Node comes online", func() {
+	It("should not sync the taints to the Node when the Node comes online, without SyncNodeClaimTaints", func() {
```
engedaam, line 207: I think we need too change the polarity this test, no?

engedaam:
Can we also update the title of this PR?

pkg/apis/v1/labels.go
JeremyBolster: This isn't going to be coming from the kubelet. It would come from the provider, I thought. This is an annotation on the NodeClaim, not the Node.

pkg/controllers/nodeclaim/lifecycle/registration.go
JeremyBolster: I thought that we wanted to only sync if the annotation exists. I guess it could be the reverse, it shouldn't make any difference though.

JeremyBolster:
> Can we also update the title of this PR?

Any suggestions? It seems descriptive to me

engedaam:
I just meant dropping the feature flag bit, since we are not doing that anymore

pkg/apis/v1/labels.go
engedaam: So in this case, the provider should be using kubelet configuration is the way that will ensure that taint is always added to the node on registration. We will use `--node-labels` flag for the kubelet configuration, which will tell Karpenter if the startup taint is already added or not https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

pkg/controllers/nodeclaim/lifecycle/registration.go
engedaam: So it will make a difference here. For the cases where the taint is not added, the cloud provider would not even be able to added startup taint and that's when we will need to reconcile the startup taints on the node

pkg/apis/v1/labels.go
```diff
@@ -38,10 +38,11 @@ const (
...
 // Karpenter specific domains and labels
 const (
-	NodePoolLabelKey        = apis.Group + "/nodepool"
-	NodeInitializedLabelKey = apis.Group + "/initialized"
-	NodeRegisteredLabelKey  = apis.Group + "/registered"
-	CapacityTypeLabelKey    = apis.Group + "/capacity-type"
+	NodePoolLabelKey            = apis.Group + "/nodepool"
+	NodeInitializedLabelKey     = apis.Group + "/initialized"
+	NodeRegisteredLabelKey      = apis.Group + "/registered"
+	NodeDoNotSyncTaintsLabelKey = apis.Group + "/nodeclaim-do-not-sync-taints"
```
engedaam, line 44: nit. 

```suggestion
	NodeDoNotSyncTaintsLabelKey = apis.Group + "/do-not-sync-taints"
```

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -121,11 +120,16 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, no
 	controllerutil.AddFinalizer(node, v1.TerminationFinalizer)
 
 	node = nodeclaimutils.UpdateNodeOwnerReferences(nodeClaim, node)
-	node.Labels = lo.Assign(node.Labels, nodeClaim.Labels)
+
+	// We do not sync the taints if the feature flag is enabled. We instead assume that the karpenter provider
+	// is managing taints. We still manage/remove the unregistered taint to signal the end of syncing.
+	if _, ok := node.Annotations[v1.NodeDoNotSyncTaintsLabelKey]; !ok {
```
engedaam, line 126: We need to make sure that value is label and not an annotation. The reason for this is that kubelet configuration does not allow us to add annotation, but will allow us to register nodes with labels.

pkg/controllers/nodeclaim/lifecycle/registration_test.go
```diff
@@ -204,7 +204,52 @@ var _ = Describe("Registration", func() {
 			Expect(node.Annotations).To(HaveKeyWithValue(k, v))
 		}
 	})
-	It("should sync the taints to the Node when the Node comes online", func() {
+	It("should not sync the taints to the Node when the Node comes online, with node label do not sync taints", func() {
+		nodeClaim := test.NodeClaim(v1.NodeClaim{
```
engedaam, line 208: Don't we need to the label added to the nodeclaim?

JeremyBolster: I don't think that is required by any piece of code. We are looking at the node, not the nodeclaim to make the decision to sync or not to sync.

engedaam: Ahh your right, I still see that we don't add the label on the node to validate they taint operation. Doesn't the label need to be added the node?

pkg/controllers/nodeclaim/lifecycle/initialization_test.go
```diff
@@ -366,6 +366,18 @@ var _ = Describe("Initialization", func() {
 		Expect(ExpectStatusConditionExists(nodeClaim, v1.ConditionTypeInitialized).Status).To(Equal(metav1.ConditionTrue))
 	})
 	It("should not consider the Node to be initialized when all startupTaints aren't removed", func() {
+		startupTaints := []corev1.Taint{
```
engedaam, line 369: Curious, why do we need to make these test changes?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -121,11 +120,16 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, no
 	controllerutil.AddFinalizer(node, v1.TerminationFinalizer)
 
 	node = nodeclaimutils.UpdateNodeOwnerReferences(nodeClaim, node)
-	node.Labels = lo.Assign(node.Labels, nodeClaim.Labels)
```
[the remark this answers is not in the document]
JeremyBolster: This is redundant code. See old line 133 or new line 137. This cleanup commit is separated out in the PR: https://github.com/kubernetes-sigs/karpenter/pull/2125/commits/e788e12b584b957f0cc04075102b6953ed6136e8

JeremyBolster:
Tests are failing locally. I am unsure of why.
```
Registration [It] should not sync the startupTaints to the Node when the Node comes online, with node label do not sync taints
 <redacted>/pkg/controllers/nodeclaim/lifecycle/registration_test.go:282

  [FAILED] Expected
      <[]v1.Taint | len:1, cap:1>: [
          {
              Key: "node.kubernetes.io/not-ready",
              Value: "",
              Effect: "NoSchedule",
              TimeAdded: nil,
          },
      ]
  to have length 2
  In [It] at: <redacted>/pkg/controllers/nodeclaim/lifecycle/registration_test.go:310
```

JeremyBolster:
@engedaam I keep getting random failures on pre-submit. Always with a different version. I don't think I have permissions to rerun a single check without re-running all, can you re-run just the failing check?

pkg/controllers/nodeclaim/lifecycle/initialization_test.go
JeremyBolster: I think was part of cleanup that I was doing earlier when this was not using labels as the feature flag. I can revert if you would like.

engedaam:
> @engedaam I keep getting random failures on pre-submit. Always with a different version. I don't think I have permissions to rerun a single check without re-running all, can you re-run just the failing check?

@JeremyBolster  I see all the test are now passing, do you still need to trigger the runner?

pkg/controllers/nodeclaim/lifecycle/initialization_test.go
engedaam: Yeah, let's drop this change if it's not needed

JeremyBolster: Sure

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -121,11 +120,16 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, no
...
+
+	// We do not sync the taints if the feature flag is enabled. We instead assume that the karpenter provider
```
engedaam, line 124: Can we update this comment just to reflect this version of the updates?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -121,11 +120,16 @@ func (r *Registration) syncNode(ctx context.Context, nodeClaim *v1.NodeClaim, no
 	controllerutil.AddFinalizer(node, v1.TerminationFinalizer)
 
 	node = nodeclaimutils.UpdateNodeOwnerReferences(nodeClaim, node)
-	node.Labels = lo.Assign(node.Labels, nodeClaim.Labels)
+
+	// We do not sync the taints if this label is present. We instead assume that the karpenter provider
+	// is managing taints. We still manage/remove the unregistered taint to signal the end of syncing.
+	if _, ok := node.Labels[v1.NodeDoNotSyncTaintsLabelKey]; !ok {
```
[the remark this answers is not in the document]
JeremyBolster: Sure, we can force it to be `"true"` as the value. I have updated the PR and added additional tests.

jonathan-innis (approved):
/lgtm
/approve

PR #2126, opened 2025-04-04
perf: Parallelize node filtering

jonathan-innis:
Fixes #N/A 

**Description**

This change adds parallelism to the node filtering logic in the scheduler. It adds the `NumConcurrentReconciles` option to the scheduler to increase the number of worker threads that can do node filtering at any given point in time. This transforms the code to first validate whether the pod is eligible to schedule to the node with the `CanAdd` call and then, once it finds a successful decision, it bails out of the filtering code and calls `Add` for the decision that it found.

Since there are multiple threads that are running at a given point in time, it's possible that an element that is further down in the array may succeed before something that is earlier in the array -- to keep the same ordering, I add a check that validates that if something succeeds and there is already a NodeClaim that has been assigned as something that has succeeded, that if the index of the current NodeClaim is lower than the NodeClaim that is stored, we replace it.

#### Before PR

```
=== RUN   TestSchedulingProfile
============== Generic Pods ==============
1 Pods      1 pods      1 nodes     108.284µs per scheduling      108.284µs per pod
50 Pods     50 pods     10 nodes    25.736993ms per scheduling    514.739µs per pod
100 Pods    100 pods    20 nodes    57.129186ms per scheduling    571.291µs per pod
500 Pods    500 pods    100 nodes   264.545427ms per scheduling   529.09µs per pod
1000 Pods   1000 pods   200 nodes   545.493791ms per scheduling   545.493µs per pod
1500 Pods   1500 pods   300 nodes   869.215333ms per scheduling   579.476µs per pod
2000 Pods   2000 pods   400 nodes   1.034069542s per scheduling   517.034µs per pod
5000 Pods   5000 pods   1000 nodes  3.207326958s per scheduling   641.465µs per pod
10000 Pods  10000 pods  2000 nodes  8.9329105s per scheduling     893.291µs per pod
20000 Pods  20000 pods  4000 nodes  25.585288208s per scheduling  1.279264ms per pod
============== Preference Pods ==============
PreferencePolicy=Respect  4000 pods  4000 nodes  35.857739167s per scheduling  8.964434ms per pod
PreferencePolicy=Ignore   4000 pods  3 nodes     418.773611ms per scheduling   104.693µs per pod

scheduled 48151 against 12034 nodes in total in 1m19.161600282s 608.262085 pods/sec
--- PASS: TestSchedulingProfile (90.31s)
```

#### After PR

```
=== RUN   TestSchedulingProfile
============== Generic Pods ==============
1 Pods      1 pods      1 nodes     121.96µs per scheduling       121.96µs per pod
50 Pods     50 pods     10 nodes    38.251325ms per scheduling    765.026µs per pod
100 Pods    100 pods    20 nodes    72.285394ms per scheduling    722.853µs per pod
500 Pods    500 pods    100 nodes   391.950055ms per scheduling   783.9µs per pod
1000 Pods   1000 pods   200 nodes   759.167396ms per scheduling   759.167µs per pod
1500 Pods   1500 pods   300 nodes   1.136690292s per scheduling   757.793µs per pod
2000 Pods   2000 pods   400 nodes   1.545552875s per scheduling   772.776µs per pod
5000 Pods   5000 pods   1000 nodes  3.823707166s per scheduling   764.741µs per pod
10000 Pods  10000 pods  2000 nodes  9.200160792s per scheduling   920.016µs per pod
20000 Pods  20000 pods  4000 nodes  20.614699417s per scheduling  1.030734ms per pod
============== Preference Pods ==============
PreferencePolicy=Respect  4000 pods  4000 nodes  17.314432333s per scheduling  4.328608ms per pod
PreferencePolicy=Ignore   4000 pods  11 nodes    480.080625ms per scheduling   120.02µs per pod

scheduled 48151 against 12042 nodes in total in 57.386706003s 839.061925 pods/sec
--- PASS: TestSchedulingProfile (66.11s)
```

#### Before PR

Screenshot 2025-03-11 at 1 07 59 AM (1)

#### After PR (8m)

Screenshot 2025-04-01 at 2 53 07 PM

`make presubmit`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ deflake: ## Run randomized, racing tests until the test fails to catch flakes
 		--randomize-all \
 		--until-it-fails \
 		-v \
-		./pkg/controllers/disruption
+		./pkg/...
 
 vulncheck: ## Verify code vulnerabilities
 	@govulncheck ./pkg/...
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ deflake: ## Run randomized, racing tests until the test fails to catch flakes
 		--randomize-all \
 		--until-it-fails \
 		-v \
-		./pkg/...
+		./pkg/controllers/disruption
 
 vulncheck: ## Verify code vulnerabilities
 	@govulncheck ./pkg/...
--- a/pkg/controllers/provisioning/scheduling/nodeclaim.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaim.go
@@ -109,32 +109,35 @@ func NewNodeClaim(
 	}
 }
 
-func (n *NodeClaim) Add(ctx context.Context, pod *corev1.Pod, podData *PodData) error {
+// CanAdd returns whether the pod can be added to the NodeClaim
+// based on the taints/tolerations, host port compatibility,
+// requirements, resources, reserved capacity reservations, and topology requirements
+func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodData) (updateRequirements scheduling.Requirements, updatedInstanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering, err error) {
 	// Check Taints
 	if err := scheduling.Taints(n.Spec.Taints).ToleratesPod(pod); err != nil {
-		return err
+		return nil, nil, nil, err
 	}
 
 	// exposed host ports on the node
 	hostPorts := scheduling.GetHostPorts(pod)
 	if err := n.hostPortUsage.Conflicts(pod, hostPorts); err != nil {
-		return fmt.Errorf("checking host port usage, %w", err)
+		return nil, nil, nil, fmt.Errorf("checking host port usage, %w", err)
 	}
 	nodeClaimRequirements := scheduling.NewRequirements(n.Requirements.Values()...)
 
 	// Check NodeClaim Affinity Requirements
 	if err := nodeClaimRequirements.Compatible(podData.Requirements, scheduling.AllowUndefinedWellKnownLabels); err != nil {
-		return fmt.Errorf("incompatible requirements, %w", err)
+		return nil, nil, nil, fmt.Errorf("incompatible requirements, %w", err)
 	}
 	nodeClaimRequirements.Add(podData.Requirements.Values()...)
 
 	// Check Topology Requirements
 	topologyRequirements, err := n.topology.AddRequirements(pod, n.NodeClaimTemplate.Spec.Taints, podData.StrictRequirements, nodeClaimRequirements, scheduling.AllowUndefinedWellKnownLabels)
 	if err != nil {
-		return err
+		return nil, nil, nil, err
 	}
 	if err = nodeClaimRequirements.Compatible(topologyRequirements, scheduling.AllowUndefinedWellKnownLabels); err != nil {
-		return err
+		return nil, nil, nil, err
 	}
 	nodeClaimRequirements.Add(topologyRequirements.Values()...)
 
@@ -145,28 +148,33 @@ func (n *NodeClaim) Add(ctx context.Context, pod *corev1.Pod, podData *PodData) 
 	if err != nil {
 		// We avoid wrapping this err because calling String() on InstanceTypeFilterError is an expensive operation
 		// due to calls to resources.Merge and stringifying the nodeClaimRequirements
-		return err
+		return nil, nil, nil, err
 	}
-
-	reservedOfferings, err := n.reserveOfferings(ctx, remaining, nodeClaimRequirements)
+	ofs, err := n.offeringsToReserve(ctx, remaining, nodeClaimRequirements)
 	if err != nil {
-		return err
+		return nil, nil, nil, err
 	}
+	return nodeClaimRequirements, remaining, ofs, nil
+}
 
+// Add updates the NodeClaim to schedule the pod to this NodeClaim, updating
+// the NodeClaim with new requirements, instance types, and offerings to reserve
+// based on the pod scheduling
+func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements scheduling.Requirements, instanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering) {
 	// Update node
 	n.Pods = append(n.Pods, pod)
-	n.InstanceTypeOptions = remaining
-	n.Spec.Resources.Requests = requests
+	n.InstanceTypeOptions = instanceTypes
+	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests)
 	n.Requirements = nodeClaimRequirements
 	n.topology.Record(pod, n.NodeClaim.Spec.Taints, nodeClaimRequirements, scheduling.AllowUndefinedWellKnownLabels)
-	n.hostPortUsage.Add(pod, hostPorts)
-	n.releaseReservedOfferings(n.reservedOfferings, reservedOfferings)
-	n.reservedOfferings = reservedOfferings
-	return nil
+	n.hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	n.reservationManager.Reserve(n.hostname, offeringsToReserve...)
+	n.releaseReservedOfferings(n.reservedOfferings, offeringsToReserve)
+	n.reservedOfferings = offeringsToReserve
 }
 
 // releaseReservedOfferings releases all offerings which are present in the current reserved offerings, but are not
-// present in the updated reserved offerings.
+// present in the updated reserved offerings
 func (n *NodeClaim) releaseReservedOfferings(current, updated cloudprovider.Offerings) {
 	updatedIDs := sets.New[string]()
 	for _, o := range updated {
@@ -184,7 +192,7 @@ func (n *NodeClaim) releaseReservedOfferings(current, updated cloudprovider.Offe
 // to reserve compatible offerings when some were available.
 //
 //nolint:gocyclo
-func (n *NodeClaim) reserveOfferings(
+func (n *NodeClaim) offeringsToReserve(
 	ctx context.Context,
 	instanceTypes []*cloudprovider.InstanceType,
 	nodeClaimRequirements scheduling.Requirements,
@@ -209,7 +217,7 @@ func (n *NodeClaim) reserveOfferings(
 			// Note that reservation is an idempotent operation - if we have previously successfully reserved an offering for
 			// this host, this operation is guaranteed to succeed. We may also succeed to make reservations for offerings which
 			// failed in previous iterations if other NodeClaims have released them since the last attempt.
-			if n.reservationManager.Reserve(n.hostname, o) {
+			if n.reservationManager.CanReserve(n.hostname, o) {
 				reservedOfferings = append(reservedOfferings, o)
 			}
 		}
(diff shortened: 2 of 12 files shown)


Makefile
```diff
@@ -3,10 +3,10 @@ export KWOK_REPO ?= ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.co
...
 
 HELM_OPTS ?= --set logLevel=debug \
-			--set controller.resources.requests.cpu=1 \
-			--set controller.resources.requests.memory=1Gi \
-			--set controller.resources.limits.cpu=1 \
-			--set controller.resources.limits.memory=1Gi
+			--set controller.resources.requests.cpu=4 \
+			--set controller.resources.requests.memory=40Gi \
+			--set controller.resources.limits.cpu=4 \
+			--set controller.resources.limits.memory=40Gi
```
rschalo, line 9: 40Gi? Quite the jump.

Makefile
```diff
@@ -38,7 +38,11 @@ apply-with-kind: verify build-with-kind ## Deploy the kwok controller from the c
 		--set controller.image.tag=$(IMG_TAG) \
 		--set serviceMonitor.enabled=true \
 		--set-string controller.env[0].name=ENABLE_PROFILING \
-		--set-string controller.env[0].value=true
+		--set-string controller.env[0].value=true \
+		--set-string controller.env[1].name=KUBE_CLIENT_BURST \
+		--set-string controller.env[1].value=5000 \
+		--set-string controller.env[1].name=KUBE_CLIENT_QPS \
+        --set-string controller.env[1].value=5000
```
rschalo, line 45: nit: formatting

Also, should we just disable client-side throttling for the Kwok controller?

Makefile
```diff
@@ -61,7 +65,11 @@ apply: verify build ## Deploy the kwok controller from the current state of your
 		--set controller.image.tag=$(IMG_TAG) \
 		--set controller.image.digest=$(IMG_DIGEST) \
 		--set-string controller.env[0].name=ENABLE_PROFILING \
-		--set-string controller.env[0].value=true
+		--set-string controller.env[0].value=true \
+		--set-string controller.env[1].name=KUBE_CLIENT_BURST \
+        --set-string controller.env[1].value=5000 \
+        --set-string controller.env[1].name=KUBE_CLIENT_QPS \
+        --set-string controller.env[1].value=5000
```
rschalo, line 72: nit: formatting and same question about client-side throttling

Makefile
```diff
@@ -83,7 +91,7 @@ deflake: ## Run randomized, racing tests until the test fails to catch flakes
 		--randomize-all \
 		--until-it-fails \
 		-v \
-		./pkg/...
+		./pkg/controllers/disruption
```
rschalo, line 94: Only deflaking disruption?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -394,75 +406,170 @@ func (s *Scheduler) updateCachedPodData(p *corev1.Pod) {
...
+				// A reserved offering error means that any subsequent successful after this NodeClaimTemplate isn't valid
+				if i >= idx {
+					return
+				}
+				newNodeClaim = nil
+				requirements = nil
+				instanceTypes = nil
+				offeringsToReserve = nil
+				idx = i
+				cancel() // Cancel the context to end the workqueue parallelization
```
jmdeal, line 547: I believe this is safe to cancel because the workqueue implementation guarantees that all calls for previous values of `i` (and potentially some future values) have already been invoked and won't be interrupted with the context. This isn't very clear to me without checking the workqueue implementation so if we stick with this implementation rather than implementing our own, I think it's worth a comment.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -394,75 +406,170 @@ func (s *Scheduler) updateCachedPodData(p *corev1.Pod) {
...
-				resources.String(s.daemonOverhead[nodeClaimTemplate]),
-				err,
-			))
-			continue
+			return
 		}
+		mu.Lock()
+		defer mu.Unlock()
 
+		// Ensure that we always take an earlier successful schedule to keep consistent ordering
```
jmdeal, line 554: Maybe worth noting that we care about consistent ordering because otherwise we could schedule to a lower priority nodepool as well as the debug benefits.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -210,6 +210,19 @@ func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, no
...
+	// We special-case kubernetes.io/hostname primarily for new NodeClaims since their domain won't be registered until we Add() them
+	if t.Key == corev1.LabelHostname && len(nodeDomains.Values()) == 1 {
+		hostName := nodeDomains.Values()[0]
+		count := t.domains[hostName] // t.domains[hostName] produces a 0 value for new NodeClaims
+		if selfSelecting {
+			count++
+		}
+		if count <= t.maxSkew {
+			return scheduling.NewRequirement(t.Key, corev1.NodeSelectorOpIn, hostName)
+		}
```
jmdeal, line 222: I think I get this now, but let's double check my understanding. This works because we assume we can always create additional hosts, so the maximum number of pods we should attempt to schedule to a node is the max skew since we can treat the global minimum as zero. If I've got that right I think it's worth a comment since it's not how Kubernetes reasons about maxSkew in general, it's just how Karpenter reasons about it since we can always create an additional domain.

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -245,9 +258,9 @@ func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, no
 	}
 	if minDomain == "" {
 		// avoids an error message about 'zone in [""]', preferring 'zone in []'
-		return scheduling.NewRequirement(podDomains.Key, corev1.NodeSelectorOpDoesNotExist)
```
jmdeal, line 248: Odd that we were using `podDomains` in the first place here, I thought there may have been a reason but it does look like `podDomains.Key` is always equal to `t.Key`. I do think using the topology group's key makes more sense.

pkg/test/options.go
```diff
@@ -38,6 +38,7 @@ type OptionsFields struct {
 	LeaderElectionName      *string
 	LeaderElectionNamespace *string
 	MemoryLimit             *int64
+	CPURequests             *int64
```
jmdeal, line 41: I'm wondering if we want to surface this as CPU requests or specifically surface this as the number of worker threads allocated to the scheduler. I see the argument to surface it this way for configuration complexity, but if this is a value we expect users may tune surfacing it directly makes more sense IMO.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -145,28 +147,34 @@ func (n *NodeClaim) Add(ctx context.Context, pod *corev1.Pod, podData *PodData)
...
-	return nil
+	n.hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	n.reservationManager.Reserve(n.hostname, offeringsToReserve...)
+	n.releaseReservedOfferings(n.reservedOfferings, offeringsToReserve)
+	n.reservedOfferings = offeringsToReserve
 }
 
 // releaseReservedOfferings releases all offerings which are present in the current reserved offerings, but are not
-// present in the updated reserved offerings.
+// present in the updated reserved offerings
```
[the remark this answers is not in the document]
jonathan-innis: No clue -- probably a copy error

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -233,11 +241,6 @@ func (n *NodeClaim) reserveOfferings(
 	return reservedOfferings, nil
 }
 
-func (n *NodeClaim) Destroy() {
```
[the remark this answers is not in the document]
jonathan-innis: Me neither 😄

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -210,6 +210,19 @@ func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, no
 	minDomain := ""
 	minCount := int32(math.MaxInt32)
 
+	// We special-case kubernetes.io/hostname primarily for new NodeClaims since their domain won't be registered until we Add() them
+	if t.Key == corev1.LabelHostname && len(nodeDomains.Values()) == 1 {
```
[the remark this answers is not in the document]
jonathan-innis: That's fair -- I'll make the update

pkg/test/options.go
jonathan-innis: I don't think I want users to tune this quite yet (I mean they can obviously tune this by messing with this value) but I'd rather view it as an "implementation detail" at this point

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -96,6 +100,12 @@ var IgnorePreferences = func(opts *options) {
 	opts.preferencePolicy = PreferencePolicyIgnore
 }
 
+var NumConcurrentReconciles = func(numConcurrentReconciles int) func(*options) {
+	return func(opts *options) {
+		opts.numConcurrentReconciles = numConcurrentReconciles
+	}
+}
```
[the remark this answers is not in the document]
jonathan-innis: This was consistent with the other options that were in the scheduling code

pkg/controllers/provisioning/scheduling/topologygroup.go
jonathan-innis: Yeah, you've got that right -- I can add a comment -- good call

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -109,32 +108,35 @@ func NewNodeClaim(
 	}
 }
 
-func (n *NodeClaim) Add(ctx context.Context, pod *corev1.Pod, podData *PodData) error {
+// CanAdd returns whether the pod can be added to the NodeClaim
+// based on the taints/tolerations, host port compatibility,
+// requirements, resources, reserved capacity reservations, and topology requirements
+func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodData) (updateRequirements scheduling.Requirements, updatedInstanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering, err error) {
```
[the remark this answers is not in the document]
jonathan-innis: I personally still think that it would make the code less readable. You're right, I do hate callbacks 😝

pkg/controllers/provisioning/scheduling/topologygroup.go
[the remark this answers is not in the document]
jmdeal: Is there a case where it could be zero values? We insert the hostname into the nodeclaim requirements when we create the NodeClaim from the NCT.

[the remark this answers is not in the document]
jonathan-innis: If there is an existing node that happens to not have the label I think

pkg/controllers/provisioning/scheduling/nodeclaim.go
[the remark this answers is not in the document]
jmdeal: It's not a hill I'm going to die on, it is a matter of preference. Generally I don't love more return values than result + error for functions because I think it decreases readability. Granted, this isn't as bad as go since you can include return value names in the signature.

pkg/controllers/provisioning/scheduling/topologygroup.go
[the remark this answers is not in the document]
jmdeal: Oh got it, so it never should happen but could if a user removes the label from the node manually. That makes sense. It's unfortunate that we can't take it as an invariant that the label is present, but given we can't this makes sense to me.

PR #2127, opened 2025-04-05
chore: Update well Known labels to be as part of  cloud providers

engedaam:
Fixes #N/A 

**Description**
- Inject well known labels as part of the controller initialization, instead of pulling from the global variable  

- `make presubmit`


pkg/utils/pretty/pretty.go
```diff
@@ -100,3 +102,16 @@ func ToSnakeCase(str string) string {
 func Sentence(str string) string {
 	return string(unicode.ToUpper(rune(str[0]))) + str[1:]
 }
+
+func GetWellKnownLabels() map[string]string {
```
jmdeal, line 106: I'm not sure this package is really the right choice for this function - it's specifically intended to format well known labels for prometheus metrics.

pkg/controllers/controllers.go
```diff
@@ -67,6 +67,7 @@ func NewControllers(
 	recorder events.Recorder,
 	cloudProvider cloudprovider.CloudProvider,
 	cluster *state.Cluster,
+	wellKnownLabels map[string]string,
```
jmdeal, line 70: Rather than injecting the well known labels **specifically** for the metrics controller here, I think it would make more sense to just inject them as part of the cloudprovider interface and **never** rely on them through global injection. That way, we aren't relying on two sources of truth for the same labels. Minimally, if we want to inject this here I'd just fame it as "additional node metric labels" rather than well-known labels.

jonathan-innis: +1 on this -- it would be nice if we could just get all the well-known labels from just calling GetInstanceTypes() and collecting all of the labels from the instance type requirements and offering requirements so we wouldn't have to create a separate method -- though, I don't know if that's really possible with things like capacity-reservation-id and it might lead to more misleading bugs in general -- but something to think about

pkg/utils/pretty/pretty.go
engedaam: Dropped the need to move this function into the utils package

pkg/controllers/controllers.go
engedaam: We won't need to do this anymore, since will be relaying on runtime metrics creation.

jonathan-innis (approved):
/lgtm
/approve

PR #2136, opened 2025-04-14
chore: Use structured errors in logger

jonathan-innis:
Fixes #N/A 

**Description**

Add structured errors from the `serrors` package in operatorpkg. This registers the logger with anything using the operator so the logger unwraps values and prints the controller-runtime error with any structured values that were injected through the error stack.

`make presubmit`

### DIFF
--- a/pkg/controllers/node/termination/terminator/eviction.go
+++ b/pkg/controllers/node/termination/terminator/eviction.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/awslabs/operatorpkg/serrors"
 	"github.com/awslabs/operatorpkg/singleton"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
@@ -202,7 +203,7 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 			q.recorder.Publish(terminatorevents.NodeFailedToDrain(&corev1.Node{ObjectMeta: metav1.ObjectMeta{
 				Name:      key.Name,
 				Namespace: key.Namespace,
-			}}, fmt.Errorf("evicting pod %s/%s violates a PDB", key.Namespace, key.Name)))
+			}}, serrors.Wrap(fmt.Errorf("evicting pod violates a PDB"), "Pod", klog.KRef(key.Namespace, key.Name))))
 			return false
 		}
 		log.FromContext(ctx).Error(err, "failed evicting pod")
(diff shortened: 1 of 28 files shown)


pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -202,7 +202,7 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 			q.recorder.Publish(terminatorevents.NodeFailedToDrain(&corev1.Node{ObjectMeta: metav1.ObjectMeta{
 				Name:      key.Name,
 				Namespace: key.Namespace,
-			}}, fmt.Errorf("evicting pod %s/%s violates a PDB", key.Namespace, key.Name)))
```
rschalo, line 205: Mean to drop the pod name here?

pkg/controllers/provisioning/provisioner.go
```diff
@@ -485,8 +486,7 @@ func (p *Provisioner) Validate(ctx context.Context, pod *corev1.Pod) error {
 func validateKarpenterManagedLabelCanExist(p *corev1.Pod) error {
 	for _, req := range scheduling.NewPodRequirements(p) {
 		if req.Key == v1.NodePoolLabelKey && req.Operator() == corev1.NodeSelectorOpDoesNotExist {
-			return fmt.Errorf("configured to not run on a Karpenter provisioned node via the %s %s requirement",
-				v1.NodePoolLabelKey, corev1.NodeSelectorOpDoesNotExist)
+			return serrors.Wrap(fmt.Errorf("configured to not run on a Karpenter provisioned node"), "requirement", fmt.Sprintf("%s %s", v1.NodePoolLabelKey, corev1.NodeSelectorOpDoesNotExist))
```
rschalo, line 489:
```suggestion
			return serrors.Wrap(fmt.Errorf("configured not to run on a Karpenter provisioned node"), "requirement", fmt.Sprintf("%s %s", v1.NodePoolLabelKey, corev1.NodeSelectorOpDoesNotExist))
```

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: Good catch!

pkg/controllers/provisioning/provisioner.go
jonathan-innis: This is tuh-mato, tomato to me -- I'd rather not change this just in case someone has a log query built on it

rschalo (approved):
/lgtm

PR #2147, opened 2025-04-16
fix: Fix pod scheduling undecided time metric

jigisha620:
Fixes #N/A 

**Description**
We emit pod scheduling undecided time metric when we have not made a pod scheduling decision gauge for it. If the pod gets bound to a node, we should delete this gauge.

### DIFF
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -261,8 +261,11 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 func (c *Controller) recordPodSchedulingUndecidedMetric(pod *corev1.Pod) {
 	nn := client.ObjectKeyFromObject(pod)
-	// If we've made a decision on this pod, delete the metric idempotently and return
-	if decisionTime := c.cluster.PodSchedulingDecisionTime(nn); !decisionTime.IsZero() {
+	_, scheduled := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
+		return c.Type == corev1.PodScheduled && c.Status == corev1.ConditionTrue
+	})
+	// If we've made a decision on this pod or the pod is already bound, delete the metric idempotently and return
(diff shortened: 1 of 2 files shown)


pkg/controllers/metrics/pod/controller.go
```diff
@@ -261,8 +261,11 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 
 func (c *Controller) recordPodSchedulingUndecidedMetric(pod *corev1.Pod) {
 	nn := client.ObjectKeyFromObject(pod)
-	// If we've made a decision on this pod, delete the metric idempotently and return
-	if decisionTime := c.cluster.PodSchedulingDecisionTime(nn); !decisionTime.IsZero() {
+	cond, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
+		return c.Type == corev1.PodScheduled
```
rschalo, line 265: nit: If we check that the status of PodScheduled is True here then the `ok` can be `scheduled` and we can use `|| scheduld` on line 268

jonathan-innis: +1, I'd just short-circuit here and delete the metric up top just to reduce the number of conditionals

pkg/controllers/metrics/pod/suite_test.go
```diff
@@ -313,6 +313,31 @@ var _ = Describe("Pod Metrics", func() {
 		})
 		Expect(found).To(BeFalse())
 	})
+	It("should delete provisioning undecided metrics based if the pod is already bound", func() {
```
jonathan-innis, line 316:
```suggestion
	It("should delete provisioning undecided metrics when the pod is already bound", func() {
```

rschalo (approved):
/lgtm

pkg/controllers/metrics/pod/controller.go
jigisha620: Done

jonathan-innis (approved):
/lgtm
/approve

PR #2154, opened 2025-04-21
chore: Update pod schedulable time

jigisha620:
Fixes #N/A 

**Description**
Store the mapping between the pod and the nodeClaim against which it should schedule.

N/A

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -169,7 +169,7 @@ func (p *Provisioner) GetPendingPods(ctx context.Context) ([]*corev1.Pod, error)
 	rejectedPods, pods := lo.FilterReject(pods, func(po *corev1.Pod, _ int) bool {
 		if err := p.Validate(ctx, po); err != nil {
 			// Mark in memory that this pod is unschedulable
-			p.cluster.MarkPodSchedulingDecisions(ctx, map[*corev1.Pod]error{po: fmt.Errorf("ignoring pod, %w", err)}, nil)
+			p.cluster.MarkPodSchedulingDecisions(ctx, map[*corev1.Pod]error{po: fmt.Errorf("ignoring pod, %w", err)}, nil, nil)
 			log.FromContext(ctx).WithValues("Pod", klog.KObj(po)).V(1).Info(fmt.Sprintf("ignoring pod, %s", err))
 			return true
 		}
@@ -331,7 +331,7 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 			log.FromContext(ctx).Info("no nodepools found")
 			p.cluster.MarkPodSchedulingDecisions(ctx, lo.SliceToMap(pods, func(p *corev1.Pod) (*corev1.Pod, error) {
 				return p, fmt.Errorf("no nodepools found")
-			}), nil)
+			}), nil, nil)
 			return scheduler.Results{}, nil
 		}
 		return scheduler.Results{}, fmt.Errorf("creating scheduler, %w", err)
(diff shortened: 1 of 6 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -419,6 +422,15 @@ func (c *Cluster) MarkPodSchedulingDecisions(ctx context.Context, podErrors map[
 			}
 		}
 	}
+	c.UpdatePodToNodeClaimMapping(ncPods)
+}
+
+func (c *Cluster) UpdatePodToNodeClaimMapping(ncPods map[string][]*corev1.Pod) {
+	for ncName, pods := range ncPods {
+		for _, p := range pods {
+			c.podToNodeClaim.LoadOrStore(client.ObjectKeyFromObject(p), ncName)
```
jonathan-innis, line 431: I don't think you want to LoadOrStore -- I think you want to just Store. If the mapping to the NodeClaim changes, I think we want to know it since this affects our metrics

pkg/controllers/state/cluster.go
```diff
@@ -419,6 +422,15 @@ func (c *Cluster) MarkPodSchedulingDecisions(ctx context.Context, podErrors map[
 			}
 		}
 	}
+	c.UpdatePodToNodeClaimMapping(ncPods)
+}
+
+func (c *Cluster) UpdatePodToNodeClaimMapping(ncPods map[string][]*corev1.Pod) {
```
jonathan-innis, line 428: nit: You could possibly consider inlining this in MarkPodSchedulingDecisions and just add a comment to it

pkg/controllers/state/cluster.go
```diff
@@ -391,6 +393,7 @@ func (c *Cluster) MarkPodSchedulingDecisions(ctx context.Context, podErrors map[
 			}
 		}
 		c.podHealthyNodePoolScheduledTime.Delete(nn)
+		c.podToNodeClaim.Delete(nn)
```
jonathan-innis, line 396: Do we not delete from this map when we are clearingDecisions? I think this map could grow unbounded

pkg/controllers/provisioning/provisioner.go
```diff
@@ -371,7 +371,15 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 		).Info("found provisionable pod(s)")
 	}
 	// Mark in memory when these pods were marked as schedulable or when we made a decision on the pods
-	p.cluster.MarkPodSchedulingDecisions(ctx, results.PodErrors, results.NodePoolToPodMapping())
+	p.cluster.MarkPodSchedulingDecisions(ctx, results.PodErrors, results.NodePoolToPodMapping(),
+		// Only passing existing nodes here and not new nodeClaims because
+		// these nodeClaims don't have a name until they are created
+		lo.SliceToMap(lo.Filter(results.ExistingNodes, func(n *scheduler.ExistingNode, _ int) bool {
```
jonathan-innis, line 377: `results.ExistingNodeToPodMapping()`

pkg/controllers/state/suite_test.go
```diff
@@ -159,28 +159,30 @@ var _ = Describe("Pod Ack", func() {
...
+		cluster.MarkPodSchedulingDecisions(ctx, nil, map[string][]*corev1.Pod{"n1": {pod}}, map[string][]*corev1.Pod{"nc1": {pod}})
 		setTime = cluster.PodSchedulingSuccessTime(nn)
 		Expect(setTime.IsZero()).To(BeFalse())
 
 		newTime := cluster.PodSchedulingSuccessTime(nn)
 		Expect(newTime.Compare(setTime)).To(Equal(0))
+		Expect(cluster.PodNodeClaimMapping(nn)).To(BeEquivalentTo("nc1"))
 	})
-	It("should delete pod schedulable time if we get error for the pod", func() {
+	It("should delete pod schedulable time and pod to nodeClaim mapping if we get error for the pod", func() {
```
jonathan-innis, line 170: You need another test here that deletes the pod when we are clearing the mappings -- frankly, can we double-check that we are clearing all of these mappings generally? Tracking all of this info is tricky business and we need to make sure that we are properly cleaning things up when they no longer need to be tracked

pkg/controllers/state/cluster.go
jigisha620: Agreed and updated.

pkg/controllers/state/cluster.go
jigisha620: Causes problem with the cyclomatic complexity of the function, hence did not implement it.

jonathan-innis (approved):
/lgtm
/approve

PR #2161, opened 2025-04-23
fix: Ensure DaemonSets with '/initialized' and '/registered' are considered during nodeclaim calculations

### LINKED ISSUE #2116: Karpenter underprovisions CPU requests for DS with certain node affinity
### Description

**Observed Behavior**:

Karpenter provisions a Node with too little allocatable CPU for the Pod and DS it needs to schedule.

**Expected Behavior**:

Karpenter accurately takes into account all DS that will end up being scheduled on the Node it will create.

Create a NodePool, DaemonSet, and Deployment by `kubectl apply -f` on the files in the comment below.
See that Karpenter logs a too-low CPU request value and provisions a Node that cannot hold the Deployment and DaemonSet together, and either the Pod of the Deployment or the Daemonset will be stuck Pending.
(If the Deployment is stuck Pending, Karpenter will provision another new too-small Node and repeat. If the DaemonSet is stuck Pending, you'll only have one Node.)

See the Karpenter logs in the comment below.

If you remove the 
```yaml
              - key: karpenter.sh/registered
                operator: In
                values:
                - "true"
```
from the DaemonSet affinity, Karpenter will accurately take into account the DaemonSet resources and provision a large enough Node.

This seems similar to (but the reverse of) https://github.com/kubernetes-sigs/karpenter/issues/715

**Versions**:
- Chart Version: v0.32.10 (yes, we're very behind)

1mwataru:
Fixes #2116 

**Description**

- When Daemonsets has nodeAffinity or nodeSelector related to `karpenter.sh/registered` and `karpenter.sh/initialized` labels, these are not considered during NodeClaim calculations, resulting in nodes being launched with smaller sizes than expected.
  - This leads to unexpected behaviors such as resource shortages on nodes, preventing Daemonset Pods running, or Deployment Pods remaining in Pending state causing new NodeClaims to be created.

- Modified the code to consider `karpenter.sh/registered` and `karpenter.sh/initialized` labels during Daemonset overhead calculations.

- Prerequisites
  - Prepared the following Daemonset and Deployment
    - Daemonset
      - Pod Identity
        - No setting
      - VPC CNI (aws-node)
        - Cpu request: 25m
      - Kube Proxy
        - Cpu request: 100m
      - My Daemonset (with required nodeAffinity that specifies nodeSelectorTerms with matchExpressions for the `karpenter.sh/registered` label)
        - Cpu request: 2 (2000m)
    - Deployment
      - My Deployment
        - Cpu request: 1 (1000m)

- Before the change
  - Confirmed from Karpenter logs that My Daemonset was not considered in the request value (`"cpu":"1150m"`)
   
    ```
    {"level":"INFO","time":"2025-04-23T13:11:26.806Z","logger":"controller","caller":"provisioning/provisioner.go:413","message":"created nodeclaim","commit":"3f9eb00-dirty","controller":"provisioner","namespace":"","name":"","reconcileID":"e786ef2c-95ea-4607-b317-d684dd519b89","NodePool":{"name":"my-nodepool"},"NodeClaim":{"name":"my-nodepool-jtzvc"},"requests":{"cpu":"1150m","pods":"4"},"instance-types":"c5.large, c5.xlarge, c5a.large, c5a.xlarge, c5d.large and 55 other(s)"}
    ```

  - As a result, an EC2 instance of type c5d.large (allocatable cpu: 1930m) was launched, but the Daemonset remained in Pending state

- After the change
  - Confirmed from Karpenter logs that My Daemonset was now considered in the request value (`"cpu":"3150m"`)

    ```
    {"level":"INFO","time":"2025-04-23T13:45:02.609Z","logger":"controller","caller":"provisioning/provisioner.go:413","message":"created nodeclaim","commit":"3f9eb00-dirty","controller":"provisioner","namespace":"","name":"","reconcileID":"061608ab-f9ad-41fa-ade5-fd19a2ccec58","NodePool":{"name":"my-nodepool"},"NodeClaim":{"name":"my-nodepool-wbh4m"},"requests":{"cpu":"3150m","pods":"5"},"instance-types":"c5.2xlarge, c5.xlarge, c5a.2xlarge, c5a.xlarge, c5d.xlarge and 55 other(s)"}
    ```

  - The Daemonset started successfully without issues
 
**Manifests used for verification**

- NodePool
  
  <details><summary>nodepool.yaml</summary>
  <p>

    ```yaml
    apiVersion: karpenter.sh/v1
    kind: NodePool
    metadata:
    name: my-nodepool
    spec:
    template:
        metadata:
        labels:
            cluster-name: cluster-31-3923-alb
        spec:
        nodeClassRef:
            group: karpenter.k8s.aws
            kind: EC2NodeClass
            name: default
        requirements:
            operator: In
            values:
            - m
            - c
            - r
            operator: Gt
            values:
            - "4"
        - key: kubernetes.io/arch
            operator: In
            values:
            - amd64
        - key: kubernetes.io/os
            operator: In
            values:
            - linux
    ```

  </p>
  </details> 

- Deployment and Deamonset

  </p>
  </details> 
  <details><summary>dep_deamon.yaml</summary>

    ```yaml
    apiVersion: apps/v1
    kind: Deployment
    metadata:
    name: my-deploy
    spec:
    replicas: 1
    selector:
        matchLabels:
        jeremy: my-deploy
    template:
        metadata:
        labels:
            jeremy: my-deploy
        spec:
        nodeSelector:
            karpenter.sh/nodepool: my-nodepool
        containers:
        - name: my-dep-container
            image: nginx:latest
            ports:
            - containerPort: 80
            resources:
            requests:
                cpu: "1"

    ---

    apiVersion: apps/v1
    kind: DaemonSet
    metadata:
    name: my-daemonset
    spec:
    selector:
        matchLabels:
        app.kubernetes.io/name: my-daemonset-app
    template:
        metadata:
        labels:
            app.kubernetes.io/name: my-daemonset-app
        spec:
        affinity:
            nodeAffinity:
                nodeSelectorTerms:
                - matchExpressions:
                - key: karpenter.sh/registered
                    operator: In
                    values:
                    - "true"
                - key: karpenter.sh/nodepool
                    operator: In
                    values:
                    - my-nodepool
        containers:
        - image: nginx:latest
            imagePullPolicy: IfNotPresent
            name: my-ds-container
            resources:
            requests:
                cpu: "2"
        tolerations:
        - operator: Exists
    ```

  <p>

### DIFF
--- a/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
@@ -135,13 +135,10 @@ func (i *NodeClaimTemplate) ToNodeClaim() *v1.NodeClaim {
 	// Karpenter can't reason about which label domains would belong to each instance type.
 	i.Labels = lo.Assign(i.Labels, i.resolveCustomLabelsFromRequirements())
 
-	// Filter out DaemonSet scheduling-only requirements for the actual NodeClaim
-	requirements := scheduling.NewRequirements()
-	for key, req := range i.Requirements {
-		if key != v1.NodeRegisteredLabelKey && key != v1.NodeInitializedLabelKey {
-			requirements.Add(req)
-		}
-	}
+	// Exclude node lifecycle requirements used only for scheduling simulation.
+	requirements := scheduling.NewRequirements(lo.Filter(i.Requirements.Values(), func(req *scheduling.Requirement, _ int) bool {
+		return req.Key != v1.NodeRegisteredLabelKey && req.Key != v1.NodeInitializedLabelKey
+	})...)
 
 	nc := &v1.NodeClaim{
 		ObjectMeta: metav1.ObjectMeta{
--- a/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
@@ -73,6 +73,12 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 	})
 	nct.Requirements.Add(scheduling.NewNodeSelectorRequirementsWithMinValues(nct.Spec.Requirements...).Values()...)
 	nct.Requirements.Add(scheduling.NewLabelRequirements(nct.Labels).Values()...)
+
+	// Add requirements for DaemonSet scheduling calculations
+	// These ensure DaemonSets with nodeAffinity for these labels are considered
+	nct.Requirements.Add(scheduling.NewRequirement(v1.NodeRegisteredLabelKey, corev1.NodeSelectorOpIn, "true"))
+	nct.Requirements.Add(scheduling.NewRequirement(v1.NodeInitializedLabelKey, corev1.NodeSelectorOpIn, "true"))
+
 	return nct
 }
 
@@ -129,6 +135,14 @@ func (i *NodeClaimTemplate) ToNodeClaim() *v1.NodeClaim {
 	// Karpenter can't reason about which label domains would belong to each instance type.
 	i.Labels = lo.Assign(i.Labels, i.resolveCustomLabelsFromRequirements())
 
+	// Filter out DaemonSet scheduling-only requirements for the actual NodeClaim
+	requirements := scheduling.NewRequirements()
+	for key, req := range i.Requirements {
+		if key != v1.NodeRegisteredLabelKey && key != v1.NodeInitializedLabelKey {
+			requirements.Add(req)
+		}
+	}
+
 	nc := &v1.NodeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", i.NodePoolName),
@@ -146,7 +160,7 @@ func (i *NodeClaimTemplate) ToNodeClaim() *v1.NodeClaim {
 		},
 		Spec: i.Spec,
 	}
-	nc.Spec.Requirements = i.Requirements.NodeSelectorRequirements()
+	nc.Spec.Requirements = requirements.NodeSelectorRequirements()
 	if nc.Spec.TerminationGracePeriod == nil {
 		nc.Spec.TerminationGracePeriod = DefaultTerminationGracePeriod
 	}
--- a/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
@@ -35,6 +36,14 @@ import (
 // This would be a mechanism to allow cloud providers to enforce a TerminationGracePeriod on all node
 // provisioned by Karpenter
 var DefaultTerminationGracePeriod *metav1.Duration = nil
+
+// schedulingSimulationKeys are requirement keys added to NodeClaimTemplate solely for
+// DaemonSet scheduling simulation. They must be excluded from NodeClaim labels and
+// spec.requirements to avoid affecting node lifecycle state (e.g. Initialized/Registered).
+var schedulingSimulationKeys = sets.New(
+	v1.NodeRegisteredLabelKey,
+	v1.NodeInitializedLabelKey,
+)
 
 // MaxInstanceTypes is a constant that restricts the number of instance types to be sent for launch. Note that this
 // is intentionally changed to var just to help in testing the code.
@@ -87,7 +96,7 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 func (i *NodeClaimTemplate) resolveCustomLabelsFromRequirements() map[string]string {
 	labels := map[string]string{}
 	for key, requirement := range i.Requirements {
-		if v1.WellKnownLabels.Has(key) || v1.RestrictedLabels.Has(key) {
+		if v1.WellKnownLabels.Has(key) || v1.RestrictedLabels.Has(key) || schedulingSimulationKeys.Has(key) {
 			continue
 		}
 		if value := requirement.Any(); value != "" {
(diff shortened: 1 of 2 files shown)


pkg/apis/v1/labels.go
```diff
@@ -89,6 +89,14 @@ var (
 		v1.LabelWindowsBuild,
 	)
 
+	// AllowedDaemonsetLabels is a set of label keys permitted for Daemonset resources.
+	// For Daemonsets, Karpenter also considers labels indicating registration and initialization status
+	// when performing scheduling.
+	AllowedDaemonsetLabels = sets.New(
```
jonathan-innis, line 95: This isn't quite right, right? Because we know that the value is going to be set to `true`, I don't think we can make the change within the requirements checking code. 

It's probably more fitting to do this in the same place we inject the `karpenter.sh/nodepool` label: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go#L68.

We do have to be careful that when we actually create the NodeClaim, we don't add these labels on there then, but other than that, I think that would fix the scheduling representation that we have

1mwataru: Thank you for your comment!
I have modified it to add within the nodeclaimtemplate.

To minimize changes to the original NodeClaimTemplate Requirements, I am not using `delete()` when removing items during NodeClaim creation.
Please let me know if using `delete()` would be better from a performance or memory perspective.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -70,6 +70,12 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 	})
 	nct.Requirements.Add(scheduling.NewNodeSelectorRequirementsWithMinValues(nct.Spec.Requirements...).Values()...)
 	nct.Requirements.Add(scheduling.NewLabelRequirements(nct.Labels).Values()...)
+
+	// Add requirements for DaemonSet scheduling calculations
```
jonathan-innis, line 74: This looks good! Can we add one test to make sure that this works properly on the two new labels? Maybe something with affinity?

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -80,6 +86,14 @@ func (i *NodeClaimTemplate) ToNodeClaim() *v1.NodeClaim {
 		return i.Name
 	})...))
 
+	// Filter out DaemonSet scheduling-only requirements for the actual NodeClaim
```
jonathan-innis, line 89: nit: These may not just be DaemonSet scheduling requirements -- I'm not sure why you would do this on things other than DaemonSets, but we should technically support this for any pod that selected against this label

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
1mwataru: I have implemented a test that verifies nodes are launched with properly calculated daemonOverhead and scheduled correctly when NodeAffinity is configured for each respective label.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
1mwataru: @jonathan-innis 
Thanks for your comment.

Apologies for the delay.
We exclude other workloads like Deployments due to the following scenarios:

- When node affinity for these labels are applied to Deployments, current [validation](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1/labels.go#L127) rejects them during NodeClaim creation, preventing NodeClaim creation entirely
- Simply removing this validation leads to unintended behavior:
  - Deployments create Pending Pods, triggering new NodeClaim creation
  - Pods remain unscheduled until Nodes become registered/initialized, causing redundant NodeClaim creation

Addressing this would require logic to prevent excessive NodeClaim creation when Pending Pods exist with registered/initialized labels. 
This represents a not small change beyond this Issue's scope, so we excluded it.
(Please let me know if we should create a separate Issue for investigation, or if any comments need modification.)

1mwataru:
HI,  @jonathan-innis 

Could you review the above my fix?

Thanks!

1mwataru:
/remove-lifecycle stale

Hi, @engedaam @jonathan-innis @tallaxes

I've confirmed that this issue is still not fixed in the latest release.

Would you mind reviewing it when you get a chance?

Thanks!

moko-poi (approved):

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
1mwataru: @jonathan-innis 

I've updated the comment to `Exclude node lifecycle requirements used only for scheduling simulation` to avoid implying this is DaemonSet-only behavior.

1mwataru:
@jonathan-innis @ellistarn @jmdeal @DerekFrank 

This PR has been waiting for a while. Could you take a look when you get a chance?

The main logic is fairly simple - it's mostly test code.

Would appreciate your review!

Thanks.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -81,7 +96,7 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 func (i *NodeClaimTemplate) resolveCustomLabelsFromRequirements() map[string]string {
 	labels := map[string]string{}
 	for key, requirement := range i.Requirements {
-		if v1.WellKnownLabels.Has(key) || v1.RestrictedLabels.Has(key) {
+		if v1.WellKnownLabels.Has(key) || v1.RestrictedLabels.Has(key) || schedulingSimulationKeys.Has(key) {
```
1mwataru, line 99: Due to this method and my previous changes, the `karpenter.sh/initialized=true` label was being assigned at NodeClaim creation time as well. 
This change excludes the registered/initialized labels as a special case to prevent that.

1mwataru:
Hi @jmdeal ,

Thank you for stepping in as the reviewer.

I've resolved the conflicts. Could you please take another look when you get a chance?

Thanks!

AndrewMitchell25 (commented):
Thanks for this fix — the approach is solid and the DaemonSet overhead calculation fix is well-validated.

One suggestion: could you add assertions verifying that the simulation-only requirements (`karpenter.sh/registered` and `karpenter.sh/initialized`) don't leak into the created NodeClaim? This would validate the `ToNodeClaim()` stripping logic alongside the overhead calculation.

Right after `node1 := ExpectScheduled(ctx, env.Client, initialPod)`, something like:

```go
// Verify simulation-only requirements don't leak into the created NodeClaim
Expect(cloudProvider.CreateCalls).To(HaveLen(1))
createdNodeClaim := cloudProvider.CreateCalls[0]
reqs := pscheduling.NewNodeSelectorRequirementsWithMinValues(createdNodeClaim.Spec.Requirements...)
Expect(reqs).ToNot(HaveKey(v1.NodeRegisteredLabelKey))
Expect(reqs).ToNot(HaveKey(v1.NodeInitializedLabelKey))
Expect(createdNodeClaim.Labels).ToNot(HaveKey(v1.NodeRegisteredLabelKey))
Expect(createdNodeClaim.Labels).ToNot(HaveKey(v1.NodeInitializedLabelKey))
```

I verified locally that these assertions pass on your branch. They use existing patterns in the file (`cloudProvider.CreateCalls[0]` and `pscheduling.NewNodeSelectorRequirementsWithMinValues`) and would guard against a regression in the requirement-stripping logic in `ToNodeClaim()` and `resolveCustomLabelsFromRequirements()`.

1mwataru:
@AndrewMitchell25 Thanks for the suggestion! 

I've incorporated your assertions to verify that simulation-only requirements don't leak into the created NodeClaim.

Regarding the CI failure — the `vulncheck` job is failing due to GO-2026-5026 (golang.org/x/net/idna), which is unrelated to this PR. 

I've opened #3041 3041to bump `golang.org/x/net` to v0.55.0 to resolve it.

Thank you

AndrewMitchell25 (approved):
LGTM

1mwataru:
Hi @jmdeal,

This PR has been reviewed and approved, and all review feedback is addressed. 
Could you take a look  when you get a chance?

Thank you!

PR #2180, opened 2025-05-01
feat: prioritize emptiness over other consolidation methods

rschalo:
# Prioritizing Emptiness - Reordering Graceful Consolidation copy

## Background:

Karpenter performs dataplane right-sizing for Kubernetes clusters. When performing consolidation, Karpenter considers if a node should be deleted, replaced, or if it isn't a candidate for disruption at all. There are multiple criteria used when determining a node's disrupt-ability and this PR proposes a re-ordering of them. Disruption via Expiration, Interruption, and Node Auto Repair are forceful consolidation methods, and are out of scope for this change.

## Goal:

Unblock emptiness consolidation as it is fast to validate and special in that it is the only graceful consolidation that results in only a deletion operation.

## Graceful Consolidation:

Karpenter evaluates four methods of graceful disruption.

The four methods happen sequentially and are:

* Drift
* Emptiness
* Multi-node consolidation
* Single-node consoldiation

If a valid disruption is found for any of these methods then it is sent to the termination orchestrator, exits the loop, and begins evaluating drift again.

## Consolidation Controls:

Disruption can be controlled by a Node Disruption Budget. These budgets are part of the NodePool spec and enables users to determine how many nodes should be disrupted for a given reason and during what time. For example, a user may want to block all consolidation except for an overnight maintenance window for 2 hours.

When multiple budgets are present on a NodePool, the most restrictive budget applies.

```
spec:
  disruption:
    budgets:
    - nodes: 50%
      reasons:
      - Drifted
      - Underutilized
    - nodes: 100%
      reasons:
      - Empty
```

While the semantic around the application of these budgets is not wrong - this PR argues that it’s unexpected that emptiness would be blocked by other disruption methods and that it instead should be the first graceful disruption performed.

## Problem:

Today, when Karpenter runs Drift, it first checks for any nodes that are both empty and drifted. 
These nodes are then sent to the termination orchestrator and Karpenter exits, and returns to the top of the graceful consolidation loop and starts evaluating Drift again. As a result, nodes that are `Drifted && Empty` are taking priority over `Drifted` nodes, being counted against the Drift node disruption budget, and are sending fewer than expected nodes to be terminated. 

Even more confusing for users, when Karpenter logs these disruptions, the disruption reason is Drifted as evidenced by this log line:

```
"controller":"disruption.queue","namespace":"","name":"","reason":"Drifted","decision":"delete","disrupted-node-count":1,"replacement-node-count":0,"disrupted-nodes":[{"Node":{"name":"ip-node"},"NodeClaim":{"name":"nodeclaim"}}],"replacement-nodes":[]}
```

While this isn’t strictly wrong, it’s perhaps unexpected.

To repro this, a sleep was added (otherwise the 1 node disruption budget causes Karpenter to skip over Drift entirely) before evaluating consolidation methods for the following cluster:

```
1. 1000 nodes one pod per node
2. Block all disruption
3. Drift NodePool
4. Scale down deployment by half
5. Set Drift Budget to: 1 Nodes and Empty Budget to: 500 nodes
6. Observe 1 Node terminated by Drift down to 500 nodes
```

### 1. Update Drift's Handling of Empty Nodes

Specifically handle Empty nodes within Drift as being Empty, applying relevant Node Disruption Budgets for Emptiness as well as correctly logging the disruption reason as Empty. Alternatively, empty nodes can be skipped over when evaluating Drift and then handled by the Emptiness consolidation method.

Pros:
* Keeps Drift the first considered graceful disruption, prioritizing patching and getting requirements in line over everything else

Cons:
* The issue covered in this PR still exists
* Complicates how we reason about disruption methods

### 2. Reorder Graceful Consolidation and Skip Emptiness in Other Methods (this PR)

Update Drift such that if it detects empty nodes, those nodes are skipped over like we do in other disruption methods. Then, these nodes are picked up in Emptiness and are deleted when evaluating Empty nodes. Additionally, Emptiness should be evaluated before Drift is evaluated so that zero-simulation consolidation options are performed first. In the above repro, 500 nodes would be consolidated before evaluating drift for the remaining nodes.

Pros:
* Higher priority for deleting empty nodes
* Requirements changing doesn't mean anything for a node destined for deletion
* Fully separates out Emptiness as its own method and budget

Cons:
* Drift not being performed first could introduce some unexpected behavior where a user is relying on patching due to AMI or requirements drift before any other changes are made to their dataplane.

### DIFF
--- a/pkg/controllers/disruption/drift_test.go
+++ b/pkg/controllers/disruption/drift_test.go
@@ -122,7 +122,7 @@ var _ = Describe("Drift", func() {
 			ExpectApplied(ctx, env.Client, rs)
 			Expect(env.Client.Get(ctx, client.ObjectKeyFromObject(rs), rs)).To(Succeed())
 		})
-		It("should allow all empty nodes to be disrupted", func() {
+		It("should not disrupt empty nodes during drift", func() {
 			nodeClaims, nodes = test.NodeClaimsAndNodes(numNodes, v1.NodeClaim{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
@@ -151,56 +151,18 @@ var _ = Describe("Drift", func() {
 			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, nodes, nodeClaims)
 			ExpectSingletonReconciled(ctx, disruptionController)
 
-			metric, found := FindMetricWithLabelValues("karpenter_nodepools_allowed_disruptions", map[string]string{
+			metric, found := FindMetricWithLabelValues("karpenter_voluntary_disruption_decisions_total", map[string]string{
 				"nodepool": nodePool.Name,
+				"reason":   string(v1.DisruptionReasonDrifted),
 			})
-			Expect(found).To(BeTrue())
-			Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 10))
+			Expect(found).To(BeFalse())
+			Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 0))
 
 			// Execute command, thus deleting all nodes
 			ExpectSingletonReconciled(ctx, queue)
-			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(0))
+			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(10))
 		})
-		It("should allow no empty nodes to be disrupted", func() {
-			nodeClaims, nodes = test.NodeClaimsAndNodes(numNodes, v1.NodeClaim{
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						v1.NodePoolLabelKey:            nodePool.Name,
-						corev1.LabelInstanceTypeStable: mostExpensiveInstance.Name,
-						v1.CapacityTypeLabelKey:        mostExpensiveOffering.Requirements.Get(v1.CapacityTypeLabelKey).Any(),
-						corev1.LabelTopologyZone:       mostExpensiveOffering.Requirements.Get(corev1.LabelTopologyZone).Any(),
-					},
-				},
-				Status: v1.NodeClaimStatus{
-					Allocatable: map[corev1.ResourceName]resource.Quantity{
-						corev1.ResourceCPU:  resource.MustParse("32"),
-						corev1.ResourcePods: resource.MustParse("100"),
-					},
-				},
-			})
-
-			nodePool.Spec.Disruption.Budgets = []v1.Budget{{Nodes: "0%"}}
-
-			ExpectApplied(ctx, env.Client, nodePool)
-			for i := 0; i < numNodes; i++ {
-				nodeClaims[i].StatusConditions().SetTrue(v1.ConditionTypeDrifted)
-				ExpectApplied(ctx, env.Client, nodeClaims[i], nodes[i])
-			}
-			// inform cluster state about nodes and nodeclaims
-			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, nodes, nodeClaims)
-			ExpectSingletonReconciled(ctx, disruptionController)
-
-			metric, found := FindMetricWithLabelValues("karpenter_nodepools_allowed_disruptions", map[string]string{
-				"nodepool": nodePool.Name,
-			})
-			Expect(found).To(BeTrue())
-			Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 0))
-
-			// Execute command, thus deleting no nodes
-			ExpectSingletonReconciled(ctx, queue)
-			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(numNodes))
-		})
-		It("should only allow 3 empty nodes to be disrupted", func() {
+		It("should only allow 3 nodes to be disrupted", func() {
 			nodeClaims, nodes = test.NodeClaimsAndNodes(numNodes, v1.NodeClaim{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
@@ -234,9 +196,9 @@ var _ = Describe("Drift", func() {
 			Expect(found).To(BeTrue())
 			Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 3))
 
-			// Execute command, thus deleting 3 nodes
+			// emptiness is not performed in drift, so expect all nodeclaims to exist
 			ExpectSingletonReconciled(ctx, queue)
-			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(7))
+			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(10))
 		})
 		It("should disrupt 3 nodes, taking into account commands in progress", func() {
 			nodeClaims, nodes = test.NodeClaimsAndNodes(numNodes, v1.NodeClaim{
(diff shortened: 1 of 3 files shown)


jonathan-innis (approved):
/lgtm
/approve

cnmcavoy:
lgtm

I like this change, and I don't think the "cons" is a big concern. Emptiness is scaling down nodes and not replacing them, so if you need to patch an AMI for a security issue, scaling down nodes is still the desired behavior because it reduces the attack surface (fewer vulnerable nodes) faster than drift can.

cnmcavoy (approved):

pkg/controllers/disruption/drift_test.go
```diff
@@ -616,63 +578,6 @@ var _ = Describe("Drift", func() {
...
-			ExpectSingletonReconciled(ctx, queue)
-			// Cascade any deletion of the nodeClaim to the node
-			ExpectNodeClaimsCascadeDeletion(ctx, env.Client, nodeClaim)
-
-			// We should delete the nodeClaim that has drifted
-			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(0))
-			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(0))
-			ExpectNotFound(ctx, env.Client, nodeClaim, node)
-		})
-		It("should disrupt all empty drifted nodes in parallel", func() {
```
rschalo, line 637: Emptiness is no longer performed in drift and drift replace/deletes happen one at a time.

pkg/controllers/disruption/drift_test.go
```diff
@@ -151,56 +151,18 @@ var _ = Describe("Drift", func() {
 			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, nodes, nodeClaims)
 			ExpectSingletonReconciled(ctx, disruptionController)
 
-			metric, found := FindMetricWithLabelValues("karpenter_nodepools_allowed_disruptions", map[string]string{
+			metric, found := FindMetricWithLabelValues("karpenter_voluntary_disruption_decisions_total", map[string]string{
```
jonathan-innis, line 154: This is a completely different metric -- did we mean to change this? (I haven't looked at the whole test in context so this might have been wrong all along)

pkg/controllers/disruption/drift_test.go
```diff
@@ -151,56 +151,18 @@ var _ = Describe("Drift", func() {
...
+				"reason":   string(v1.DisruptionReasonDrifted),
 			})
-			Expect(found).To(BeTrue())
-			Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 10))
-
-			// Execute command, thus deleting all nodes
-			ExpectSingletonReconciled(ctx, queue)
-			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(0))
-		})
-		It("should allow no empty nodes to be disrupted", func() {
```
jonathan-innis, line 164: We're removing this because the drift budget no longer applies, right?

pkg/controllers/disruption/drift_test.go
```diff
@@ -151,56 +151,18 @@ var _ = Describe("Drift", func() {
...
+			Expect(found).To(BeFalse())
 			Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 0))
 
-			// Execute command, thus deleting no nodes
+			// Execute command, deleting no nodes
 			ExpectSingletonReconciled(ctx, queue)
 			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(numNodes))
 		})
-		It("should only allow 3 empty nodes to be disrupted", func() {
+		It("should only allow 3 nodes to be disrupted", func() {
```
jonathan-innis, line 165: Should we change the name of this test -- or at least make sure that this test is still testing what it should now that we have made this change? With this update, I'm not sure if this is different from other tests in the suite

pkg/controllers/disruption/drift_test.go
```diff
@@ -434,9 +396,9 @@ var _ = Describe("Drift", func() {
 				Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 3))
 			}
 
-			// Execute the command in the queue, deleting all nodes
+			// emptiness is not performed in drift, so expect all nodeclaims to exist
```
jonathan-innis, line 399: This one too: Do we need to update the naming of our tests/take a look at all the tests in the suite in context to make sure we are still maintaining our testing coverage for the drift disruption?

pkg/controllers/disruption/suite_test.go
```diff
@@ -1833,18 +1833,20 @@ var _ = Describe("Metrics", func() {
 	})
 	It("should fire metrics for single node empty disruption", func() {
 		nodeClaim, node := nodeClaims[0], nodes[0]
```
jonathan-innis, line 1835: Do we not need to have this node marked as consolidatable for this metric to fire? If not, do we have a bug?

pkg/controllers/disruption/drift_test.go
rschalo: We should probably actually test for both here.

pkg/controllers/disruption/drift_test.go
rschalo: Correct - although, with your other comments, doing a quick run through of all these tests instead of just the functionality makes sense.

pkg/controllers/disruption/drift_test.go
```diff
@@ -352,91 +339,59 @@ var _ = Describe("Drift", func() {
...
-				})
-				Expect(found).To(BeTrue())
-				Expect(metric.GetGauge().GetValue()).To(BeNumerically("==", 2))
-			}
-
-			// Execute the command in the queue, only deleting 20 nodes
-			ExpectSingletonReconciled(ctx, queue)
-			Expect(len(ExpectNodeClaims(ctx, env.Client))).To(Equal(10))
-		})
-		It("should allow all nodes from each nodePool to be deleted", func() {
```
rschalo, line 380: This would now happen in emptiness because the test used to have empty && drifted nodes. There is a test that covers disruptions across multiple nodepools.

pkg/controllers/disruption/drift_test.go
```diff
@@ -601,79 +629,7 @@ var _ = Describe("Drift", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectExists(ctx, env.Client, nodeClaim)
 		})
-		It("should ignore nodes with the drifted status condition set to false", func() {
```
rschalo, line 604: Moved.

pkg/controllers/disruption/drift_test.go
```diff
@@ -601,79 +629,7 @@ var _ = Describe("Drift", func() {
...
-
-			fakeClock.Step(10 * time.Minute)
-
-			ExpectSingletonReconciled(ctx, disruptionController)
-
-			// Expect to not create or delete more nodeclaims
-			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(1))
-			ExpectExists(ctx, env.Client, nodeClaim)
-		})
-		It("can delete drifted nodes", func() {
```
rschalo, line 619: Covered by `should delete drifted nodes`

pkg/controllers/disruption/drift_test.go
```diff
@@ -941,24 +954,5 @@ var _ = Describe("Drift", func() {
 			ExpectExists(ctx, env.Client, nodeClaim)
 			ExpectExists(ctx, env.Client, node)
 		})
-		It("should delete nodes with the karpenter.sh/do-not-disrupt annotation set to false", func() {
```
rschalo, line 944: Moved to other annotation test in same file.

pkg/controllers/disruption/emptiness_test.go
```diff
@@ -411,6 +411,31 @@ var _ = Describe("Emptiness", func() {
...
+
+			ExpectSingletonReconciled(ctx, queue)
+			// Cascade any deletion of the nodeClaim to the node
+			ExpectNodeClaimsCascadeDeletion(ctx, env.Client, nodeClaim)
+
+			// we should delete the empty node
+			Expect(ExpectNodeClaims(ctx, env.Client)).To(HaveLen(0))
+			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(0))
+			ExpectNotFound(ctx, env.Client, nodeClaim, node)
+			ExpectMetricGaugeValue(disruption.EligibleNodes, 1, map[string]string{
```
rschalo, line 435: Validate via metrics that the disruption was due to emptiness.

pkg/controllers/disruption/emptiness_test.go
```diff
@@ -411,6 +411,31 @@ var _ = Describe("Emptiness", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(0))
 			ExpectNotFound(ctx, env.Client, nodeClaim, node)
 		})
+		It("can delete empty and drifted nodes", func() {
```
jonathan-innis, line 414: Nice! 🎉

jonathan-innis (approved):
/lgtm
/approve

PR #2182, opened 2025-05-01
fix: should consider pending pods with preferred nodeAffinity when making provisioning decisions

### LINKED ISSUE #2181: Karpenter does not scale the cluster when the pod was a preferred node affinity for a given hostname
### Description

**Observed Behavior**:

When deploying Karpenter with `preferencePolicy=Ignore`, I would expect that any `preferredDuringSchedulingIgnoredDuringExecution` added to pod would be considered as a soft constraint and Karpenter would still scale the cluster if such pod is marked as unschedulable by the kube-scheduler.

However, Karpenter is still ignoring some pods, depending of the `key` used in the `matchExpressions` even if `preferencePolicy=Ignore` is defined, essentially not scaling the cluster further. 

**Expected Behavior**:

If Karpenter is running with `preferencePolicy=Ignore`, I would expect that pending pods flagged as unschedulable by the kube-scheduler would trigger Karpenter provisioning capabilities.

Assuming a deployment like the following:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: pod-info-test-1
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/instance: pod-info-test-1
  revisionHistoryLimit: 10
  template:
    metadata:
      labels:
        app.kubernetes.io/instance: pod-info-test-1
    spec:
      containers:
        - name: pod-info-test-1
          image: "ghcr.io/stefanprodan/podinfo:6.8.0"
          imagePullPolicy: IfNotPresent
          resources:
            limits:
              cpu: 8
              memory: 16Gi
            requests:
              cpu: 8
              memory: 16Gi
      affinity:
        nodeAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - preference:
              matchExpressions:
              - key: kubernetes.io/hostname
                operator: In
                values:
                - ip-10-0-0-1.eu-west-1.compute.internal
            weight: 100
```
And assuming that there is no enough spare capacity in the cluster for a pod with 8 cpus to be scheduled, once the pod is created, it will stuck in Pending state.

**Versions**:
- Chart Version: v1.4.0
- Kubernetes Version (`kubectl version`): v1.30.11

rodrigorfk:
Fixes #2181 

**Description**
When deploying Karpenter with `preferencePolicy=Ignore`, I would expect that any `preferredDuringSchedulingIgnoredDuringExecution` added to pod would be considered as a soft constraint and Karpenter would still scale the cluster if such pod is marked as unschedulable by the kube-scheduler.

However, Karpenter is still ignoring some pods, depending of the `key` used in the `matchExpressions` even if `preferencePolicy=Ignore` is defined, essentially not scaling the cluster further. 

I am running a custom build of Karpenter internally, on a cluster with 300+ nodes and with higher CPU allocation / bin-packing (98%+), before the change, some pods containing the mentioned soft `preferredDuringSchedulingIgnoredDuringExecution` condition was stuck in pending state, however after the custom build was introduced, the issue stopped working and provisioning started working as expected. 


pkg/controllers/provisioning/provisioner.go
```diff
@@ -519,13 +519,16 @@ func validateNodeSelector(p *corev1.Pod) (errs error) {
...
-func validateAffinity(p *corev1.Pod) (errs error) {
+func validateAffinity(ctx context.Context, p *corev1.Pod) (errs error) {
+	policy := options.FromContext(ctx).PreferencePolicy
 	if p.Spec.Affinity == nil {
 		return nil
 	}
 	if p.Spec.Affinity.NodeAffinity != nil {
-		for _, term := range p.Spec.Affinity.NodeAffinity.PreferredDuringSchedulingIgnoredDuringExecution {
-			errs = multierr.Append(errs, validateNodeSelectorTerm(term.Preference))
+		if policy != options.PreferencePolicyIgnore {
```
jonathan-innis, line 528: I'm wondering if we should just ignore the hostname affinity preference in general -- or just ignore validating preferences? We'd have to double-check to ensure that we don't get strange behavior from doing this but since we relax/ignore preferences, it seems odd that we would block fully on any of these preferences

jonathan-innis: Sure, makes sense -- yeah, I'd say let's just ignore this in all cases then -- I'd agree that I don't think we should be overly defensive with blocking a valid configuration that Karpenter _could_ fulfill

jonathan-innis (approved):
/lgtm
/approve

jonathan-innis (approved):
/lgtm
/approve

PR #2188, opened 2025-05-02
test: Update Expiration test time

engedaam:
Fixes #N/A 

**Description**
- Validate that only one node is getting expired 

- `make presubmit`


test/suites/integration/expiration_test.go
```diff
@@ -54,7 +54,11 @@ var _ = Describe("Expiration", func() {
 		selector = labels.SelectorFromSet(dep.Spec.Selector.MatchLabels)
 	})
 	It("should expire the node after the expiration is reached", func() {
-		nodePool.Spec.Template.Spec.ExpireAfter = karpv1.MustParseNillableDuration("2m")
+		if env.IsDefaultNodeClassKWOK() {
+			nodePool.Spec.Template.Spec.ExpireAfter = karpv1.MustParseNillableDuration("30s")
+		} else {
+			nodePool.Spec.Template.Spec.ExpireAfter = karpv1.MustParseNillableDuration("5m")
```
jmdeal, line 60: 5 minutes seems a little long, right? I get why it can be shorter with KWOK, but any reason to increase it from 2 to 5 for non-KWOK NodeClasses?

test/suites/integration/expiration_test.go
```diff
@@ -71,7 +75,7 @@ var _ = Describe("Expiration", func() {
 			g.Expect(ok).To(BeTrue())
 		}).Should(Succeed())
 
-		env.EventuallyExpectCreatedNodeCount("==", 2)
+		env.EventuallyExpectCreatedNodeCount("==", 1)
```
jmdeal, line 78: Why should this be one? Shouldn't it still be two since we're creating the replacement node?

test/suites/integration/expiration_test.go
engedaam: It's not a given that nodes will register and become healthy in 5 mins for all cloud providers. This would be to give more buffer

test/suites/integration/expiration_test.go
engedaam: No, we reset the counter, so expect to just create one nodeclaim

test/suites/integration/expiration_test.go
jmdeal: I'd rather leave it at 2 minutes and increase as needed for other cloudproviders in the interest in keeping E2E tests fast.

test/suites/integration/expiration_test.go
jmdeal: Oh, missed the monitor reset. Looks good then 👍

test/suites/integration/expiration_test.go
engedaam: This from testing with the AWS Provider, and the 5 minites is needed to validate that the nodeclaim expired and new node is spun up.

PR #2193, opened 2025-05-05
test: Migrate the NodeClaim suite test from the AWS Karpenter Provider

engedaam:
Fixes #N/A 

**Description**
* High Level Description: Validate that action taken by the nodeclaim controllers. This validates standalone claims and the garbage collection mechanisms  
* Source: https://github.com/aws/karpenter-provider-aws/tree/main/test/suites/nodeclaim
* Test Cases: 
    * Standalone NodeClaim  
        * should create a standard NodeClaim based on resource requests
        * should create a NodeClaim propagating all NodeClaim spec details
        * should remove the cloudProvider NodeClaim when the cluster NodeClaim is deleted
        * should delete a NodeClaim from the node termination finalizer 
        * should delete a NodeClaim after the registration timeout when the node doesn’t register
        * should delete a NodeClaim if it references a NodeClass that doesn’t exist
        * should delete a NodeClaim if it references a NodeClass that isn’t Ready 
        * should succeed to garbage collect an Instance that was launched by a NodeClaim but has no cloudprovider mapping 

**How was this change tested?**
- `make presubmit`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2195, opened 2025-05-05
test: Migrate the Drift suite test from the AWS Karpenter Provider

engedaam:
Fixes #N/A 

**Description**
##### Drift

* High Level Description: Validate the expected drift functionality of each field and expected case to limit drift
* Source: https://github.com/aws/karpenter-provider-aws/tree/main/test/suites/drift
* Test Cases: 
    * Budgets 
        * should respect budgets for empty drift
        * should respect budgets for non-empty delete drift 
        * should respect budgets for non-empty replace drift 
        * should not allow drift if the budgets is fully blocking 
        * should not allow drift if te budget is fully blocking during a scheduled time 
    * NodePool
        * Static Drift 
            * Annotations 
            * Labels 
            * Taints 
            * Start-up Taints 
            * Node Requirement
    * Drift Failure 
        * should not disrupt a drifted node if the replacement node never registers 
        * should not disrupt a drifted node if the replacement node registers but never initialized 
        * should not drift any nodes if their pod disruption budgets are unhealthy 
       
- `make presubmit`

### DIFF
--- a/hack/kwok/stages/pod-unhealthy.yaml
+++ b/hack/kwok/stages/pod-unhealthy.yaml
@@ -1,0 +1,68 @@
+apiVersion: kwok.x-k8s.io/v1alpha1
+kind: Stage
+metadata:
+  annotations:
+  name: pod-unhealthy
+spec:
+  next:
+    statusSubresource: status
+    statusTemplate: |
+      {{ $now := Now }}
+
+      conditions:
+      - lastTransitionTime: {{ $now | Quote }}
+        status: "True"
+        type: Initialized
+      - lastTransitionTime: {{ $now | Quote }}
+        status: "False"
+        type: Ready
+      - lastTransitionTime: {{ $now | Quote }}
+        status: "True"
+        type: ContainersReady
+      {{ range .spec.readinessGates }}
+      - lastTransitionTime: {{ $now | Quote }}
+        status: "True"
+        type: {{ .conditionType | Quote }}
+      {{ end }}
+
+      containerStatuses:
+      {{ range .spec.containers }}
+      - image: {{ .image | Quote }}
+        name: {{ .name | Quote }}
+        ready: true
+        restartCount: 0
+        state:
+          running:
+            startedAt: {{ $now | Quote }}
+      {{ end }}
+
+      initContainerStatuses:
+      {{ range .spec.initContainers }}
+      - image: {{ .image | Quote }}
+        name: {{ .name | Quote }}
+        ready: true
+        restartCount: 0
+        state:
+          terminated:
+            exitCode: 0
+            finishedAt: {{ $now | Quote }}
+            reason: Completed
+            startedAt: {{ $now | Quote }}
+      {{ end }}
+
+      hostIP: {{ NodeIPWith .spec.nodeName | Quote }}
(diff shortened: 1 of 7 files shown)


hack/kwok/stages/pod-unhealthy.yaml
```diff
@@ -0,0 +1,68 @@
+apiVersion: kwok.x-k8s.io/v1alpha1
```
engedaam, line 1: This is a way to mock unhealthy pods after they have placed on a node.

test/suites/integration/drift_test.go
```diff
@@ -0,0 +1,622 @@
...
+		})
+	})
+	Context("Failure", func() {
+		It("should not disrupt a drifted node if the replacement node never registers", func() {
+			version, err := env.KubeClient.Discovery().ServerVersion()
+			Expect(err).To(BeNil())
+			if version.Minor < "28" {
+				Skip("This test is only valid for K8s >= 1.28")
+			}
+			admissionspolicy := &admissionregistrationv1.ValidatingAdmissionPolicy{
```
engedaam, line 430: Admissions policy are used to mock node registration failures.

hack/kwok/stages/pod-unhealthy.yaml
jonathan-innis, line 1: Why put this in the hack and not make this part of the Kwok package? e.g. `kwok/stages` or even make it part of the chart directly? I don't really see this as a hack? Also, is this something that we built or something that's available upstream from kwok?

test/pkg/environment/common/setup.go
```diff
@@ -65,6 +66,8 @@ var (
 		&schedulingv1.PriorityClass{},
 		&corev1.Node{},
 		&v1.NodeClaim{},
+		&admissionregistrationv1.ValidatingAdmissionPolicy{},
```
jonathan-innis, line 69: We have a validatingAdmissionPolicy now?

engedaam: Yeah, this is how we will mocking node registration failure

test/suites/integration/drift_test.go
engedaam: I was updating the status condition of the nodeclaim at first and that would reduce the time to less then a min: https://github.com/kubernetes-sigs/karpenter/pull/2193/files#diff-ec78a380674cc833d5255f7b850e5ae25337cd6099fed376cc7e177e449b19c8R216. However this felt more natural to me. We can go either way.

hack/kwok/stages/pod-unhealthy.yaml
engedaam: This is not a hack, and the way that KWOK recommends to mock pod and node health: https://kwok.sigs.k8s.io/docs/user/stages-configuration/. This seemed reasonable to me at the time, but I agree the chart is a better place to add this CR. This was inspired based from files available from upstream KWOK. However these is a CR we needed to build

test/suites/integration/drift_test.go
jonathan-innis: ? There is a registration delay on the KwokNodeClass -- we can't use that?

engedaam: I did consider using  the registration delay. However, as a tanant , I want to make sure these can be imported by any Karpenter cloud provider and run against any nodeclass. Using NodeClass specific logic would break that, like registration delay, and I have stayed away from solutions that can't work for all Cloud providers

jonathan-innis (approved):
/lgtm
/approve

PR #2197, opened 2025-05-05
test: Migrate the Integration suite test from the AWS Karpenter Provider

engedaam:
Fixes #N/A 

**Description**
##### Integration

* High Level Description: Multiple functionality are tested. This is the catch all suite for testing karpenter
* Source: https://github.com/aws/karpenter-provider-aws/tree/main/test/suites/integration
* Test Cases: 
    * DaemonSet
        * should account for LimitRange Default on daemonSet pods for resources
        * should account for LimitRange DefaultRequest on daemonSet pods for resources
    * Hash
        * should have NodePool hash
    * Repair Policy
        * should ignore disruption budgets
        * should ignore do-not-disrupt annotation on node
        * should ignore terminationGracePeriod on the nodepool
    * Utilization
        * should provision one pod per node
    * Validation 
        * NodePool
            * should error when a restricted label is used in labels (karpenter.sh/nodepool)
            * should error when a restricted label is used in labels (kubernetes.io/custom-label)
            * should allow a restricted label exception to be used in labels (node-restriction.kubernetes.io/custom-label)
            * should allow a restricted label exception to be used in labels ([*].node-restriction.kubernetes.io/custom-label)
            * should error when a requirement references a restricted label (karpenter.sh/nodepool)
            * should error when a requirement uses In but has no values
            * should error when a requirement uses an unknown operator
            * should error when Gt is used with multiple integer values
            * should error when Lt is used with multiple integer values
            * should error when consolidateAfter is negative
            * should succeed when ConsolidationPolicy=WhenEmptyOrUnderutilized is used with consolidateAfter
            * should error when minValues for a requirement key is negative
            * should error when minValues for a requirement key is zero
            * should error when minValues for a requirement key is more than 50
            * should error when minValues for a requirement key is greater than the values specified within In operator

- `make presubmit`


test/pkg/environment/common/expectations.go
```diff
@@ -91,6 +91,26 @@ func (env *Environment) ExpectUpdated(objects ...client.Object) {
 	}
 }
 
+// WARNING: This ignores the resource version check, which can result in
+// overwriting changes made by other controllers in the cluster.
+// This is useful in ensuring that we can clean up resources by patching
+// out finalizers.
+// Grab the object before making the updates to reduce the chance of this race.
+func (env *Environment) ExpectStatusUpdated(objects ...client.Object) {
```
jonathan-innis, line 99: nit: We should eventually think about which of these we can migrate over into operatorpkg to reduce some of our code duplication here

jonathan-innis (approved):
/lgtm
/approve

test/pkg/environment/common/expectations.go
engedaam: Absolutely, once the migration is over. We will be using the common environment package from here

PR #2203, opened 2025-05-06
chore: Reduce "waiting on cluster sync" spam

jonathan-innis:
Fixes #N/A 

**Description**

This change reduces the number of "waiting on cluster sync" logs that we receive, particularly when nothing is wrong. This ensures that we only log this log when the cluster isn't synced for an extended period of time

`make presubmit`


Bryce-Soghigian (approved):

rschalo (approved):
/lgtm

PR #2214, opened 2025-05-09
test: Add default disruption budgets for the drifted initialized failure tests

engedaam:
Fixes #N/A 

**Description**
- Update `should not disrupt a drifted node if the replacement node registers but never initialized` 
- The test is designed to work off the default disruption budgets. To make the E2E tests faster we have allowed full disruption for the E2E test global nodepool. 

- `make e2etests`

### DIFF
--- a/test/suites/integration/drift_test.go
+++ b/test/suites/integration/drift_test.go
@@ -484,6 +484,7 @@ var _ = Describe("Drift", Ordered, func() {
 					},
 				},
 			})
+			// Expect only a single node to get tainted due to default disruption budgets
 			nodePool.Spec.Disruption = v1.Disruption{}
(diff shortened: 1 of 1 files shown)


test/suites/integration/drift_test.go
```diff
@@ -531,6 +531,7 @@ var _ = Describe("Drift", Ordered, func() {
 					},
 				},
 			})
+			nodePool.Spec.Disruption = v1.Disruption{}
```
jmdeal, line 534: Can we add a comment for why we're overwriting this?

jmdeal (commented):
/lgtm
/approve

jonathan-innis (approved):
/approve

PR #2220, opened 2025-05-12
refactor: convert validation to an interface

rschalo:
Fixes #N/A 

**Description**
In the disruption tests, there is not currently a deterministic way to induce pod churn. This PR refactors validation to be an interface called "Validator" that has default behavior and can be implemented for different methods of consolidation. This also allows a custom test validator that can perform other actions when it calls `Validate`

Unit testing that creates pod churn for emptiness.


pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -38,13 +38,19 @@ const MultiNodeConsolidationType = "multi"
...
 func NewMultiNodeConsolidation(consolidation consolidation) *MultiNodeConsolidation {
-	return &MultiNodeConsolidation{consolidation: consolidation}
+	return &MultiNodeConsolidation{
+		consolidation: consolidation,
+		Validator:     NewConsolidationValidator(consolidation),
+	}
 }
 
 func (m *MultiNodeConsolidation) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[string]int, candidates ...*Candidate) (Command, scheduling.Results, error) {
+	// needed to reset the validator sync after every run
```
jonathan-innis, line 52: This is a weird semantic -- why are we holding state in the validator like this?

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -96,7 +102,7 @@ func (m *MultiNodeConsolidation) ComputeCommand(ctx context.Context, disruptionB
 		return cmd, scheduling.Results{}, nil
 	}
 
-	if err := NewValidation(m.clock, m.cluster, m.kubeClient, m.provisioner, m.cloudProvider, m.recorder, m.queue, m.Reason()).IsValid(ctx, cmd, consolidationTTL); err != nil {
+	if cmd, err = m.Validator.Validate(ctx, cmd, consolidationTTL); err != nil {
```
jonathan-innis, line 105:
```suggestion
	if cmd, err = m.Validate(ctx, cmd, consolidationTTL); err != nil {
```

nit

pkg/controllers/disruption/emptiness.go
```diff
@@ -100,8 +101,7 @@ func (e *Emptiness) ComputeCommand(ctx context.Context, disruptionBudgetMapping
 	case <-e.clock.After(consolidationTTL):
 	}
 
-	v := NewValidation(e.clock, e.cluster, e.kubeClient, e.provisioner, e.cloudProvider, e.recorder, e.queue, e.Reason())
-	validatedCandidates, err := v.ValidateCandidates(ctx, cmd.candidates...)
+	validCmd, err := e.Validator.Validate(ctx, cmd, consolidationTTL)
```
jonathan-innis, line 104:
```suggestion
	validCmd, err := e.Validate(ctx, cmd, consolidationTTL)
```

pkg/controllers/disruption/singlenodeconsolidation.go
```diff
@@ -39,25 +39,27 @@ const SingleNodeConsolidationType = "single"
...
 		consolidation:             consolidation,
 		PreviouslyUnseenNodePools: sets.New[string](),
+		Validator:                 NewConsolidationValidator(consolidation),
 	}
 }
 
 // ComputeCommand generates a disruption command given candidates
 // nolint:gocyclo
 func (s *SingleNodeConsolidation) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[string]int, candidates ...*Candidate) (Command, scheduling.Results, error) {
+	// needed to reset the validator sync after every run
```
jonathan-innis, line 56: Same comment here -- this is strange, we should see what we can do to make sure that we are holding state across runs

pkg/controllers/disruption/validation.go
```diff
@@ -66,49 +72,93 @@ type Validation struct {
...
+			cloudProvider: c.cloudProvider,
+			recorder:      c.recorder,
+			queue:         c.queue,
+			reason:        v1.DisruptionReasonEmpty,
+		},
+	}
+}
+
+func (e *EmptinessValidator) Validate(ctx context.Context, cmd Command, _ time.Duration) (Command, error) {
+	validatedCandidates, err := e.validateCandidates(ctx, e.ShouldDisrupt, cmd.candidates...)
```
jonathan-innis, line 95: Is ShouldDisrupt function not available through the emptiness method directly so that we don't repeat ourselves

pkg/controllers/disruption/validation.go
```diff
@@ -148,12 +198,23 @@ func (v *Validation) ValidateCandidates(ctx context.Context, candidates ...*Cand
 }
 
 // ShouldDisrupt is a predicate used to filter candidates
-func (v *Validation) ShouldDisrupt(_ context.Context, c *Candidate) bool {
+func (v *validation) ShouldDisrupt(_ context.Context, c *Candidate) bool {
```
jonathan-innis, line 201: Same comment here -- I'd try to stay DRY and just re-use the method's ShouldDisrupt

pkg/controllers/disruption/emptiness_test.go
```diff
@@ -39,6 +40,71 @@ import (
...
+
+type TestEmptinessValidatorOption func(*TestEmptinessValidator)
+
+func WithChurn() TestEmptinessValidatorOption {
+	return func(v *TestEmptinessValidator) {
+		v.churn = true
+	}
+}
+
+func NewTestEmptinessValidator(nodes []*corev1.Node, nodeClaims []*v1.NodeClaim, nodePool *v1.NodePool, e *disruption.EmptinessValidator, opts ...TestEmptinessValidatorOption) disruption.Validator {
```
jonathan-innis, line 59: Why does the emptiness validator have nodes, nodeclaims, and nodepools? Seems odd.

Separately -- why would this test validator contain the actual emptiness validator -- why not just return true all the time and then you can control when you send back false?

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -2302,6 +2302,38 @@ var _ = Describe("Consolidation", func() {
 			// and delete the old one
```
jonathan-innis, line 2302: We should think about writing our own ConsolidationValidator that pulls out the consolidateAfter so that we don't have to do the weird clock.Step() that we have everywhere throughout our testing

pkg/controllers/disruption/emptiness.go
```diff
@@ -32,11 +31,13 @@ import (
 // Emptiness is a subreconciler that deletes empty candidates.
 type Emptiness struct {
 	consolidation
+	Validator
```
jmdeal, line 34: nit: can we encapsulate the validator in the consolidation struct and inject the correct validation through the make consolidation call?

pkg/controllers/disruption/consolidation_test.go
rschalo: Discussed offline, taking as a fast follow so as not to block release.

pkg/controllers/disruption/emptiness.go
rschalo: We would have to make Consolidation exportable in order to access the Validator so that we can override it in testing. Because nothing else about consolidation is changing, I'd prefer to leave it as is.

pkg/controllers/disruption/emptiness_test.go
jmdeal: Generally I agree with @jonathan-innis - if we're separating these components I think our tests should be limited to testing the integration with the interface rather than trying to mock the actual behavior. I.E. have a dumb mock interface for the validator which you use to test the integration with the controller, and have standalone tests for the validator itself. Generally I think this makes tests easier to reason about, and it makes it easier to ensure you have full coverage. This is a general issue I think with our approach to testing up until now though - we have very few unit tests and mostly only have integration tests. Both are important, and I think we should rely more heavily on the former.

That being said, this is more in-line with what we have been doing up until now and I don't want to block here. I do think we want to think more about how we architect our code to be testable with mocks in the future though.

pkg/controllers/disruption/validation.go
```diff
@@ -50,60 +49,104 @@ func IsValidationError(err error) bool {
...
+	}
+	return cmd, nil
+}
+
+func (c *ConsolidationValidator) isValid(ctx context.Context, cmd Command, validationPeriod time.Duration) error {
 	var err error
-	v.once.Do(func() {
-		v.start = v.clock.Now()
-	})
+	c.start = c.clock.Now()
```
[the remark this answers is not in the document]
rschalo: None that I can tell, good catch.

pkg/controllers/disruption/validation.go
```diff
@@ -50,60 +49,102 @@ func IsValidationError(err error) bool {
...
+func (c *ConsolidationValidator) isValid(ctx context.Context, cmd Command, validationPeriod time.Duration) error {
+	var err error
+	validationStart := c.clock.Now()
+	waitDuration := validationPeriod - c.clock.Since(validationStart)
 	if waitDuration > 0 {
 		select {
 		case <-ctx.Done():
 			return errors.New("context canceled")
-		case <-v.clock.After(waitDuration):
+		case <-c.clock.After(waitDuration):
```
jmdeal, line 135: nit: we can just pass `validationPeriod` directly to `c.clock.After()` right?

PR #2224, opened 2025-05-13
perf: Speed-up resource checking for existing nodes

jonathan-innis:
Fixes #N/A 

**Description**

This change speeds-up our existing node check by doing a quick check on whether some set of resources fits rather than re-creating a new resource set every time

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/existingnode.go
+++ b/pkg/controllers/provisioning/scheduling/existingnode.go
@@ -31,32 +31,32 @@ type ExistingNode struct {
 	cachedAvailable v1.ResourceList // Cache so we don't have to re-subtract resources on the StateNode every time
 	cachedTaints    []v1.Taint      // Cache so we don't hae to re-construct the taints each time we attempt to schedule a pod
 
-	Pods         []*v1.Pod
-	topology     *Topology
-	requests     v1.ResourceList
-	requirements scheduling.Requirements
+	Pods               []*v1.Pod
+	topology           *Topology
+	remainingResources v1.ResourceList
+	requirements       scheduling.Requirements
 }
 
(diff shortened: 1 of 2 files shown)


engedaam (commented):
/lgtm

pkg/controllers/provisioning/scheduling/existingnode.go
```diff
@@ -31,32 +31,32 @@ type ExistingNode struct {
...
 	}
 	node := &ExistingNode{
-		StateNode:       n,
-		cachedAvailable: n.Available(),
-		cachedTaints:    taints,
-		topology:        topology,
-		requests:        remainingDaemonResources,
-		requirements:    scheduling.NewLabelRequirements(n.Labels()),
+		StateNode:          n,
+		cachedAvailable:    n.Available(),
```
AlexanderYastrebov, line 55: `cachedAvailable` is not used anymore but is still calculated.

pkg/utils/resources/resources.go
```diff
@@ -95,6 +95,19 @@ func Subtract(lhs, rhs v1.ResourceList) v1.ResourceList {
 	return result
 }
 
+// SubtractFrom subtracts the src v1.ResourceList from the dest v1.ResourceList in-place
+func SubtractFrom(dest v1.ResourceList, src v1.ResourceList) {
+	if dest == nil {
+		sz := len(src)
+		dest = make(v1.ResourceList, sz)
+	}
```
AlexanderYastrebov, line 103: This seems to allocate input similar to MergeInto but never returns the allocated value.

IMO it would be better to not allocate at all here and in MergeInto to avoid cases like
```go
var foo  v1.ResourceList
...
resources.MergeInto(foo, bar) // forgot to reassign return value
```

PR #2226, opened 2025-05-13
perf: Poll for DaemonSet resources rather than watching

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that we only re-poll for new DaemonSet pods every 1m so that we don't spend a ton of CPU time re-queuing this controller every time there is _any_ update to the DaemonSet

`make presubmit`


rschalo:
When we update between intervals, what's the impact to scheduling? Slightly incorrect provisioning that is then addressed by the next provisioning loop?

jonathan-innis:
Correct, this was generally put in place to account for mutating webhooks and resource quotas -- we're mainly just requeueing too aggressively here and the poll really only comes into play if someone changes something either on the resource quota or on the mutating webhook that they are using

PR #2234, opened 2025-05-16
test: Update E2E testing suite to be named Regression

engedaam:
Fixes #N/A 

**Description**
- Rename the existing E2E testing suite to "Regression Suite" to:
  1. Prevent naming conflicts with integration test suite
  2. Better reflect its purpose of catching regressions across the full system

- `make e2etests`


test/suites/integration/suite_test.go
jmdeal, line 1: The root folder should be renamed as well if this is now the regression suite, right?

PR #2239, opened 2025-05-19
feat: add metrics for disruption candidate validation

rschalo:
Fixes #N/A 

**Description**
In order to better understand Karpenter's behavior during graceful disruption, this PR adds metrics for whenever Karpenter found candidates for disruption but then bailed because the candidates failed validation. This information can be incredibly useful when determining things like candidate selection and further fine-tuning disruption cost. These metrics would also have surfaced issues where Emptiness consolidation is repeatedly invalidated because of pod churn to a single candidate.

Expanded unit testing.

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -217,7 +217,7 @@ var _ = Describe("Consolidation", func() {
 				metrics.ReasonLabel: "underutilized",
 			})
 		})
-		DescribeTable("should correctly report invalidated commands when", func(invalidatedReason string, validatorOpt TestEmptinessValidatorOption) {
+		DescribeTable("should correctly report invalidated commands for emptiness disruption when", func(invalidatedReason string, validatorOpt TestEmptinessValidatorOption) {
 			ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
 			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, nodeStateController, nodeClaimStateController, []*corev1.Node{node}, []*v1.NodeClaim{nodeClaim})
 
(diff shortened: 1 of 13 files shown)


rschalo:
/hold

Needs to be updated for recent emptiness batch change

pkg/controllers/disruption/metrics.go
```diff
@@ -75,7 +75,17 @@ var (
 			Name:      "consolidation_timeouts_total",
 			Help:      "Number of times the Consolidation algorithm has reached a timeout. Labeled by consolidation type.",
 		},
-		[]string{consolidationTypeLabel},
+		[]string{ConsolidationTypeLabel},
+	)
+	InvalidatedConsolidationTotal = opmetrics.NewPrometheusCounter(
```
jonathan-innis, line 80: Is it `failed_validations` or `invalidations`? I'm not sure that I have a strong opinion either way -- I do think that we need to make sure that the name of the actual go variable closely matches the naming in prometheus

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -214,6 +217,87 @@ var _ = Describe("Consolidation", func() {
 				metrics.ReasonLabel: "underutilized",
 			})
 		})
+		DescribeTable("should correctly report invalidated commands for emptiness disruption when", func(invalidatedReason string, validatorOpt TestEmptinessValidatorOption) {
```
jonathan-innis, line 220:
```suggestion
		DescribeTable("should correctly report invalidated commands for emptiness disruption", func(invalidatedReason string, validatorOpt TestEmptinessValidatorOption) {
```

nit: I'd include the "When" at the start of the entries

jonathan-innis, line 220: When you say "correctly report invalidated commands", does that mean that we return back modified commands?

rschalo: No just the metrics for them. I can make clarifications to the suite.

pkg/controllers/disruption/metrics.go
rschalo: I think it's probably better put as `failed_validations`

pkg/controllers/disruption/validation.go
```diff
@@ -151,29 +160,40 @@ func (c *ConsolidationValidator) isValid(ctx context.Context, cmd Command, valid
...
 
 func (e *EmptinessValidator) validateCandidates(ctx context.Context, candidates ...*Candidate) ([]*Candidate, error) {
+	// This GetCandidates call filters out nodes that were nominated
 	validatedCandidates, err := GetCandidates(ctx, e.cluster, e.kubeClient, e.recorder, e.clock, e.cloudProvider, e.filter, GracefulDisruptionClass, e.queue)
 	if err != nil {
 		return nil, fmt.Errorf("constructing validation candidates, %w", err)
 	}
 	validatedCandidates = mapCandidates(candidates, validatedCandidates)
 	if len(validatedCandidates) == 0 {
+		InvalidatedConsolidationTotal.Inc(map[string]string{ConsolidationTypeLabel: e.validationType, metrics.ReasonLabel: CandidatesFiltered})
```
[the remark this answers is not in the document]
rschalo: Thoughts on `karpenter_disruption_failed_validations_total`?

pkg/controllers/disruption/validation.go
```diff
@@ -32,6 +32,12 @@ import (
 	"sigs.k8s.io/karpenter/pkg/controllers/provisioning"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/events"
+	"sigs.k8s.io/karpenter/pkg/metrics"
+)
+
+const (
+	BlockingBudget     = "blocking_budget"
+	CandidatesFiltered = "candidates_filtered"
```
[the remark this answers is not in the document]
rschalo: I thought about this a while and this metric is a catch all case where the candidate failed validation. The number of reasons a candidate could fail validation is actually quite high if you look at the number of filters applied in the `GetCandidates` function when it calls `ValidateNodeDisruptable`: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/state/statenode.go#L200-L225.

We could make each of those a reason why a candidate failed validation but I think a catch all that says "this was a candidate but was then filtered out" makes sense.

I think it's probably better stated as `candidate_no_longer_eligible` or similar, thoughts?

pkg/controllers/disruption/metrics.go
```diff
@@ -75,7 +75,17 @@ var (
...
+		crmetrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: voluntaryDisruptionSubsystem,
+			Name:      "failed_validations_total",
+			Help:      "Number of times a consolidation decision was invalidated. Labeled by consolidation type and reason.",
+		},
+		[]string{ConsolidationTypeLabel, metrics.ReasonLabel},
```
jonathan-innis, line 88: We discussed offline and decided to drop the reason label for now since it was tough to reason about what each reason would mean out of context

pkg/controllers/disruption/validation.go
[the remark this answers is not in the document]
jonathan-innis: Discussed offline, we're taking away the reasons for now and going to add it if it becomes necessary later.

pkg/controllers/disruption/validation.go
```diff
@@ -58,33 +64,35 @@ type Validator interface {
...
+	filter         CandidateFilter
+	validationType string
 }
 
 type EmptinessValidator struct {
 	validation
 }
 
-func NewEmptinessValidator(c consolidation, filter CandidateFilter) *EmptinessValidator {
+func NewEmptinessValidator(c consolidation, filter CandidateFilter, validationType string) *EmptinessValidator {
```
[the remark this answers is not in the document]
jonathan-innis: Discussed offline: We're going to remove this in favor of creating it directly in the constructor

pkg/controllers/disruption/singlenodeconsolidation.go
```diff
@@ -43,12 +43,11 @@ type SingleNodeConsolidation struct {
 }
 
 func NewSingleNodeConsolidation(consolidation consolidation) *SingleNodeConsolidation {
-	s := &SingleNodeConsolidation{
+	return &SingleNodeConsolidation{
 		consolidation:             consolidation,
 		PreviouslyUnseenNodePools: sets.New[string](),
+		Validator:                 NewSingleConsolidationValidator(consolidation),
```
rschalo, line 49: Ended up making a single and multi node validator because the consolidation type label on a metric represents `single` and `multi` and not `underutilized`.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -42,9 +42,10 @@ type MultiNodeConsolidation struct {
 }
 
 func NewMultiNodeConsolidation(consolidation consolidation) *MultiNodeConsolidation {
```
jonathan-innis, line 44:
```suggestion
func NewMultiNodeConsolidation(c consolidation) *MultiNodeConsolidation {
```

nit: Naming difference across constructors

pkg/controllers/disruption/validation.go
```diff
@@ -165,15 +193,20 @@ func (e *EmptinessValidator) validateCandidates(ctx context.Context, candidates
...
+			FailedValidationsTotal.Inc(map[string]string{ConsolidationTypeLabel: e.validationType, metrics.ReasonLabel: CandidatesIneligible})
 			return false
 		}
 		disruptionBudgetMapping[cn.NodePool.Name]--
 		return true
 	}); len(valid) > 0 {
 		return valid, nil
 	}
-	return nil, NewValidationError(fmt.Errorf("a candidate failed validation because it was nominated for a pod or would violate disruption budgets"))
+	return nil, NewValidationError(fmt.Errorf("%d candidates failed validation because it they were nominated for a pod or would violate disruption budgets", len(candidates)))
```
jonathan-innis, line 209:
```suggestion
	return nil, NewValidationError(fmt.Errorf("%d candidates failed validation because they were nominated for a pod or would violate disruption budgets", len(candidates)))
```

pkg/controllers/disruption/metrics.go
```diff
@@ -75,7 +76,17 @@ var (
...
+	)
+	FailedValidationsTotal = opmetrics.NewPrometheusCounter(
+		crmetrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: voluntaryDisruptionSubsystem,
+			Name:      "failed_validations_total",
+			Help:      "Number of candidates that were selected for disruption but failed validation. Labeled by consolidation type.",
+		},
+		[]string{ConsolidationTypeLabel, metrics.ReasonLabel},
```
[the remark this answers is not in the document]
rschalo: We're still emitting `candidates_ineligible` as the reason - did we want to drop it entirely and just track failed validations by consolidation type?

pkg/controllers/disruption/validation.go
jonathan-innis: Does it make sense to attach each candidate to a reason why it failed so that way we can count unique failures and return that in the error message

pkg/controllers/disruption/metrics.go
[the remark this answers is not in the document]
jonathan-innis: I thought that's what we agreed on -- I personally still don't find the reason being helpful -- we can always add more prometheus labels later but taking them away is more difficult

jonathan-innis (approved):
/lgtm
/approve

PR #2252, opened 2025-05-23
WIP, DO NOT MERGE. feat: Refactor the eviction queue to be multithreaded

DerekFrank:
Fixes #N/A 

**Description**

For performance, this change makes the termination controller multithreaded on evicting pods. 

`make presubmit`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,10 @@ help: ## Display help
 presubmit: verify test licenses vulncheck ## Run all steps required for code to be checked in
 
 install-kwok: ## Install kwok provider
-	UNINSTALL_KWOK=false ./hack/install-kwok.sh
+	./hack/install-kwok.sh
 
 uninstall-kwok: ## Uninstall kwok provider
-	UNINSTALL_KWOK=true ./hack/install-kwok.sh
+	UNINSTALL=true ./hack/install-kwok.sh
 
 build-with-kind: # build with kind assumes the image will be uploaded directly onto the kind control plane, without an image repository
 	$(eval CONTROLLER_IMG=$(shell $(WITH_GOFLAGS) KO_DOCKER_REPO="$(KWOK_REPO)" ko build sigs.k8s.io/karpenter/kwok))
--- a/hack/install-kwok.sh
+++ b/hack/install-kwok.sh
@@ -24,15 +24,16 @@ mkdir ${BASE}
 # make the alphabet, so that we can set be flexible to the number of allowed partitions (inferred max at 26)
 alphabet=( {a..z} )
 # set the number of partitions to 1. Currently only one partition is supported. 
-num_partitions=1
+: "${PARTITION_COUNT:=1}"
+: "${UNINSTALL:=false}"
 
 # allow it to schedule to critical addons, but not schedule onto kwok nodes.
-cat <<EOF > "${BASE}/tolerate-all.yaml"
+mkdir "${BASE}/deployment"
+cat <<EOF > "${BASE}/deployment/tolerate-all.yaml"
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: kwok-controller
-  namespace: kube-system
 spec:
   template:
     spec:
@@ -49,22 +50,48 @@ spec:
                 operator: DoesNotExist
 EOF
 
+curl -s -o "${BASE}/deployment/deployment.yaml" "https://raw.githubusercontent.com/kubernetes-sigs/kwok/refs/tags/${KWOK_RELEASE}/kustomize/kwok/deployment.yaml"
 # TODO: Simplify the kustomize to only use one copy of the RBAC that all
 # controllers can use.  
-cat <<EOF > "${BASE}/kustomization.yaml"
+cat <<EOF > "${BASE}/deployment/kustomization.yaml"
   apiVersion: kustomize.config.k8s.io/v1beta1
   kind: Kustomization
+  namespace: kube-system
   images:
   - name: registry.k8s.io/kwok/kwok
     newTag: "${KWOK_RELEASE}"
   resources:
-  - "https://github.com/${KWOK_REPO}/kustomize/kwok?ref=${KWOK_RELEASE}"
+  - deployment.yaml
   patches:
   - path: tolerate-all.yaml
+  labels:
+  - includeSelectors: true
+    pairs:
+      app: kwok-controller
+EOF
+
+
+curl -s -o "${BASE}/service.yaml" "https://raw.githubusercontent.com/kubernetes-sigs/kwok/refs/tags/${KWOK_RELEASE}/kustomize/kwok/service.yaml"
+curl -s -o "${BASE}/kwok.yaml" "https://raw.githubusercontent.com/kubernetes-sigs/kwok/refs/tags/${KWOK_RELEASE}/kustomize/kwok/kwok.yaml"
+cat <<EOF > "${BASE}/kustomization.yaml"
+  apiVersion: kustomize.config.k8s.io/v1beta1
+  kind: Kustomization
+  namespace: kube-system
+  resources:
+  - service.yaml
+  - "https://github.com/${KWOK_REPO}/kustomize/crd?ref=${KWOK_RELEASE}"
+  - "https://github.com/${KWOK_REPO}/kustomize/rbac?ref=${KWOK_RELEASE}"
+  configMapGenerator:
+  - name: kwok
+    namespace: kube-system
+    options:
+      disableNameSuffixHash: true
+    files:
+    - kwok.yaml
 EOF
 
 # Create num_partitions
-for ((i=0; i<num_partitions; i++))
+for ((i=0; i<PARTITION_COUNT; i++))
 do
   SUB_LET_DIR=$HOME_DIR/${alphabet[i]}
   mkdir ${SUB_LET_DIR}
@@ -82,7 +109,7 @@ cat <<EOF > "${SUB_LET_DIR}/kustomization.yaml"
   - name: registry.k8s.io/kwok/kwok
     newTag: "${KWOK_RELEASE}"
   resources:
-  - ./../base
+  - ./../base/deployment
   nameSuffix: -${alphabet[i]}
   patches:
     - path: ${SUB_LET_DIR}/patch.yaml
@@ -97,21 +124,23 @@ done
 
 cat <<EOF > "${HOME_DIR}/kustomization.yaml"
 resources:
+ - ./base
 EOF
 
 # Create num_partitions
-for ((i=0; i<num_partitions; i++))
+for ((i=0; i<PARTITION_COUNT; i++))
 do
   echo " - ./${alphabet[i]}" >> "${HOME_DIR}/kustomization.yaml"
 done
 
+echo "${HOME_DIR}/kwok.yaml"
 kubectl kustomize "${HOME_DIR}" > "${HOME_DIR}/kwok.yaml"
 
 # v0.4.0 added in stage CRDs which are necessary for pod/node initialization
 crdURL="https://github.com/${KWOK_REPO}/releases/download/${KWOK_RELEASE}/stage-fast.yaml"
 
-# Set UNINSTALL_KWOK=true if you want to uninstall.
-if [[ ${UNINSTALL_KWOK} = "true" ]]
+# Set UNINSTALL=true if you want to uninstall.
+if [[ ${UNINSTALL} = "true" ]]
 then
   kubectl delete -f ${crdURL}
   kubectl delete -f ${HOME_DIR}/kwok.yaml


pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -69,122 +66,97 @@ func IsNodeDrainError(err error) bool {
...
-	}
-}
-
 type Queue struct {
-	workqueue.TypedRateLimitingInterface[QueueKey]
+	sync.Mutex
 
-	mu  sync.Mutex
-	set sets.Set[QueueKey]
+	source chan event.TypedGenericEvent[*corev1.Pod]
```
jonathan-innis, line 72: Should we make the channel buffered to avoid having to wait on the reconciler pulling items off of the queue before we proceed?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -69,122 +66,97 @@ func IsNodeDrainError(err error) bool {
...
+		&corev1.Pod{ObjectMeta: metav1.ObjectMeta{Namespace: pod.Namespace, Name: pod.Name}},
 		&policyv1.Eviction{
 			DeleteOptions: &metav1.DeleteOptions{
 				Preconditions: &metav1.Preconditions{
-					UID: lo.ToPtr(key.UID),
+					UID: lo.ToPtr(pod.UID),
 				},
 			},
 		}); err != nil {
+		fmt.Printf("I Hit an error! %s\n", err.Error())
```
jonathan-innis, line 159: nit: Make sure to drop this before we merge it in

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -69,122 +66,97 @@ func IsNodeDrainError(err error) bool {
...
 
-// Evict returns true if successful eviction call, and false if there was an eviction-related error
-func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(key.Namespace, key.Name)))
+// Evict returns nil if successful eviction call, and an error if there was an eviction-related error
+func (q *Queue) Evict(ctx context.Context, pod *corev1.Pod) error {
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(pod.Namespace, pod.Name)))
 	if err := q.kubeClient.SubResource("eviction").Create(ctx,
-		&corev1.Pod{ObjectMeta: metav1.ObjectMeta{Namespace: key.Namespace, Name: key.Name}},
+		&corev1.Pod{ObjectMeta: metav1.ObjectMeta{Namespace: pod.Namespace, Name: pod.Name}},
```
jonathan-innis, line 151: Rather than creating a new pod, can't you just use the pod as-is?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -197,27 +169,33 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
 			// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L160
 			// 409 - The pod exists, but it is not the same pod that we initiated the eviction on
 			// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L318
-			return true
+			fmt.Printf("Error is an acceptable one\n")
```
jonathan-innis, line 172: nit: make sure we drop this before we merge it

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -197,27 +169,33 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
+			return nil
 		}
 		if apierrors.IsTooManyRequests(err) { // 429 - PDB violation
 			q.recorder.Publish(terminatorevents.NodeFailedToDrain(&corev1.Node{ObjectMeta: metav1.ObjectMeta{
-				Name:      key.Name,
-				Namespace: key.Namespace,
-			}}, serrors.Wrap(fmt.Errorf("evicting pod violates a PDB"), "Pod", klog.KRef(key.Namespace, key.Name))))
-			return false
+				Name:      pod.Name,
```
jonathan-innis, line 177: This event looks wrong -- it's referencing a node but that node is based on the pod info?

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -127,53 +129,33 @@ var _ = Describe("Eviction/Queue", func() {
...
 				Labels:         testLabels,
 				MaxUnavailable: &intstr.IntOrString{IntVal: 0},
 			})
-			ExpectApplied(ctx, env.Client, pdb, pdb2, pod)
-			Expect(queue.Evict(ctx, terminator.NewQueueKey(pod, node.Spec.ProviderID))).To(BeFalse())
+			ExpectApplied(ctx, env.Client, pdb, pdb2, pod, node)
+			Expect(queue.Evict(ctx, pod)).ToNot(BeNil())
 			ExpectMetricCounterValue(terminator.NodesEvictionRequestsTotal, 1, map[string]string{terminator.CodeLabel: "500"})
 		})
 		It("should ensure that calling Evict() is valid while making Add() calls", func() {
```
jonathan-innis, line 150: Why did we delete this test?

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -127,53 +129,33 @@ var _ = Describe("Eviction/Queue", func() {
 				Labels:         testLabels,
 				MaxUnavailable: &intstr.IntOrString{IntVal: 1},
 			})
-			ExpectApplied(ctx, env.Client, pod)
-			Expect(queue.Evict(ctx, terminator.NewQueueKey(pod, node.Spec.ProviderID))).To(BeTrue())
+			ExpectApplied(ctx, env.Client, pod, node)
+			Expect(queue.Evict(ctx, pod)).To(BeNil())
```
jonathan-innis, line 133:
```suggestion
			Expect(queue.Evict(ctx, pod)).To(Succeed())
```
nit: This reads a little better -- you could change all the places you have `BeNil()` to this

pkg/test/expectations/expectations.go
```diff
@@ -109,13 +109,34 @@ func ExpectNotFound(ctx context.Context, c client.Client, objects ...client.Obje
 	GinkgoHelper()
 	for _, object := range objects {
 		Eventually(func() bool {
-			return errors.IsNotFound(c.Get(ctx, types.NamespacedName{Name: object.GetName(), Namespace: object.GetNamespace()}, object))
+			err := c.Get(ctx, types.NamespacedName{Name: object.GetName(), Namespace: object.GetNamespace()}, object)
+			fmt.Printf("Deleting pod: %s\n", object)
```
jonathan-innis, line 113: nit: Drop the print lines before merge

pkg/test/expectations/expectations.go
```diff
@@ -109,13 +109,34 @@ func ExpectNotFound(ctx context.Context, c client.Client, objects ...client.Obje
...
+				fmt.Printf("Error deleting pod: %s\n", err.Error())
+			}
+			return errors.IsNotFound(err)
 		}, ReconcilerPropagationTime, RequestInterval).Should(BeTrue(), func() string {
 			return fmt.Sprintf("expected %s/%s to be deleted, but it still exists", lo.Must(apiutil.GVKForObject(object, scheme.Scheme)), client.ObjectKeyFromObject(object))
 		})
 	}
 }
 
+func ExpectPodNotFound(ctx context.Context, c client.Client, pods ...*corev1.Pod) {
```
jonathan-innis, line 124: Why does ExpectNotFound work here?

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -222,11 +206,12 @@ var _ = Describe("Eviction/Queue", func() {
 		})
 		It("should delete a pod with less than terminationGracePeriodSeconds remaining before nodeTerminationTime", func() {
 			pod.Spec.TerminationGracePeriodSeconds = lo.ToPtr[int64](120)
-			ExpectApplied(ctx, env.Client, pod)
+			ExpectApplied(ctx, env.Client, pod, node)
 
 			nodeTerminationTime := time.Now().Add(time.Minute * 1)
 			Expect(terminatorInstance.DeleteExpiringPods(ctx, []*corev1.Pod{pod}, &nodeTerminationTime)).To(Succeed())
-			ExpectNotFound(ctx, env.Client, pod)
+			fmt.Println("deleted the pod!")
```
jonathan-innis, line 213: nit

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -194,26 +177,27 @@ var _ = Describe("Eviction/Queue", func() {
...
 
-			ExpectApplied(ctx, env.Client, nodeClaim, pod)
-			Expect(queue.Evict(ctx, terminator.NewQueueKey(pod, node.Spec.ProviderID))).To(BeTrue())
+			ExpectApplied(ctx, env.Client, nodeClaim, node, pod)
+			Expect(queue.Evict(ctx, pod)).To(BeNil())
 
 			ExpectMetricCounterValue(terminator.PodsDrainedTotal, 1, map[string]string{terminator.ReasonLabel: "SpotInterruption"})
 			ExpectMetricCounterValue(terminator.NodesEvictionRequestsTotal, 1, map[string]string{terminator.CodeLabel: "200"})
 			Expect(recorder.Calls(events.Evicted)).To(Equal(1))
+			fmt.Println("AHHHH")
```
jonathan-innis, line 186: nit

pkg/controllers/node/termination/terminator/suite_test.go
DerekFrank: We didn't delete the test. The test is setup this specific way due to the singleton nature of the old eviction queue. Given that it is no longer a singleton reconcile, we can simply spam .Add calls because they will test the queue's ability to have both Evict and Add called simultaneously

pkg/controllers/node/termination/terminator/eviction.go
DerekFrank: Now that you mention it, probably!

pkg/controllers/node/termination/terminator/eviction.go
DerekFrank: That was the way the event was before this change. Given that we are draining a node by evicting pods from it, we publish the an event on the node when we can't evict a pod

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -69,119 +66,91 @@ func IsNodeDrainError(err error) bool {
...
 
 	// Evict the pod
-	if q.Evict(ctx, item) {
-		q.TypedRateLimitingInterface.Forget(item)
-		q.mu.Lock()
-		q.set.Delete(item)
-		q.mu.Unlock()
-		return reconcile.Result{RequeueAfter: singleton.RequeueImmediately}, nil
+	err := q.Evict(ctx, pod)
+	if err != nil {
```
jonathan-innis: nit: Typical style is to collapse L134 and L135 so that it reads `if err := q.Evict(ctx, pod); err != nil`

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -69,119 +66,91 @@ func IsNodeDrainError(err error) bool {
...
+		Named(q.Name()).
+		WatchesRawSource(source.Channel(q.source, handler.TypedFuncs[*corev1.Pod, reconcile.Request]{
+			GenericFunc: func(_ context.Context, e event.TypedGenericEvent[*corev1.Pod], queue workqueue.TypedRateLimitingInterface[reconcile.Request]) {
+				queue.Add(reconcile.Request{
+					NamespacedName: client.ObjectKeyFromObject(e.Object),
+				})
+			},
+		})).
+		WithOptions(controller.Options{
+			RateLimiter:             reasonable.RateLimiter(),
```
jonathan-innis: How long do we want our backoff to be when we requeue pods? Do we want a really long wait time or do we want to reduce the max delay?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -69,119 +66,91 @@ func IsNodeDrainError(err error) bool {
...
-		return reconcile.Result{RequeueAfter: singleton.RequeueImmediately}, nil
+	err := q.Evict(ctx, pod)
+	if err != nil {
+		return reconcile.Result{}, err
 	}
 
-	// Requeue pod if eviction failed
-	q.TypedRateLimitingInterface.AddRateLimited(item)
-	return reconcile.Result{RequeueAfter: singleton.RequeueImmediately}, nil
+	q.Lock()
```
jonathan-innis: nit: Consider changing the Unlock() to a `defer` statement to match style elsewhere

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: Right -- I think the event before was wrong just reading through this -- it expects a node with a name and we are giving it the pod name and namespace for some reason

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -197,27 +166,32 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
-				Name:      key.Name,
-				Namespace: key.Namespace,
-			}}, serrors.Wrap(fmt.Errorf("evicting pod violates a PDB"), "Pod", klog.KRef(key.Namespace, key.Name))))
-			return false
+				Name:      pod.Name,
+				Namespace: pod.Namespace,
+			}}, serrors.Wrap(fmt.Errorf("evicting pod violates a PDB"), "Pod", klog.KRef(pod.Namespace, pod.Name))))
+			return err
 		}
 		log.FromContext(ctx).Error(err, "failed evicting pod")
```
jonathan-innis: We don't need this error anymore if we return the error out of the reconcile since that's automatically printed by controller-runtime

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -197,27 +166,32 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
-		return false
+		return err
 	}
 	NodesEvictionRequestsTotal.Inc(map[string]string{CodeLabel: "200"})
-	reason := evictionReason(ctx, key, q.kubeClient)
-	q.recorder.Publish(terminatorevents.EvictPod(&corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: key.Name, Namespace: key.Namespace}}, reason))
+	reason := evictionReason(ctx, pod, q.kubeClient)
+	q.recorder.Publish(terminatorevents.EvictPod(&corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: pod.Name, Namespace: pod.Namespace}}, reason))
```
jonathan-innis: Can we just use the pod directly in the publish rather than reconstructing it

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -223,11 +205,13 @@ var _ = Describe("Eviction/Queue", func() {
 		})
 		It("should delete a pod with less than terminationGracePeriodSeconds remaining before nodeTerminationTime", func() {
 			pod.Spec.TerminationGracePeriodSeconds = lo.ToPtr[int64](120)
-			ExpectApplied(ctx, env.Client, pod)
+			ExpectApplied(ctx, env.Client, pod, node)
 
 			nodeTerminationTime := time.Now().Add(time.Minute * 1)
 			Expect(terminatorInstance.DeleteExpiringPods(ctx, []*corev1.Pod{pod}, &nodeTerminationTime)).To(Succeed())
-			ExpectNotFound(ctx, env.Client, pod)
+			ExpectObjectReconciled(ctx, env.Client, queue, pod)
```
jonathan-innis: I don't think that we are supposed to be evicting the pod here -- I believe this is intended to just be a pod delete

pkg/controllers/node/termination/terminator/suite_test.go
jonathan-innis: How does that work? We're still not calling Reconcile in this test so I'm not clear how this exercises these things in parallel

pkg/controllers/node/termination/suite_test.go
```diff
@@ -944,29 +956,6 @@ var _ = Describe("Termination", func() {
 			Expect(ok).To(BeTrue())
 			Expect(lo.FromPtr(m.GetHistogram().SampleCount)).To(BeNumerically("==", 1))
 		})
-		It("should update the eviction queueDepth metric when reconciling pods", func() {
```
jonathan-innis, line 947: Nice! This means we can just use the controller-runtime metric for this! How does this affect the existing metric?

pkg/controllers/node/termination/suite_test.go
```diff
@@ -738,7 +746,8 @@ var _ = Describe("Termination", func() {
 			ExpectApplied(ctx, env.Client, node, pod)
 
 			// Trigger eviction queue with the pod key still in it
-			ExpectSingletonReconciled(ctx, queue)
+			Expect(queue.Has(pod)).To(BeTrue())
+			queue.Add(pod)
```
jonathan-innis: We're adding a pod to the queue when it's already there?

pkg/controllers/node/termination/suite_test.go
```diff
@@ -452,16 +458,16 @@ var _ = Describe("Termination", func() {
...
 				}
 				// Start draining the pod group, but don't complete it yet
 				EventuallyExpectTerminating(ctx, env.Client, lo.Map(podGroup, func(p *corev1.Pod, _ int) client.Object { return p })...)
 
 				// Look at the next pod group and ensure that none of the pods have started terminating on it
 				if i != len(podGroups)-1 {
-					for range podGroups[i+1] {
-						ExpectSingletonReconciled(ctx, queue)
+					for _, pod := range podGroups[i+1] {
+						Expect(queue.Has(pod)).To(BeFalse())
```
jonathan-innis, line 468: Is this reconciling the pod?

pkg/controllers/node/termination/suite_test.go
```diff
@@ -208,8 +208,11 @@ var _ = Describe("Termination", func() {
 			ExpectApplied(ctx, env.Client, node, nodeClaim, pod, pdb)
 			ExpectManualBinding(ctx, env.Client, pod, node)
 			Expect(env.Client.Delete(ctx, node)).To(Succeed())
+
 			// We only reconcile once since this label should be applied before draining the node
-			ExpectRequeued(ExpectObjectReconciled(ctx, env.Client, terminationController, node))
+			ExpectRequeued(ExpectObjectReconciled(ctx, env.Client, terminationController, node)) // DrainInitiation
+			_ = ExpectObjectReconcileFailed(ctx, env.Client, queue, pod)
```
jonathan-innis: now we reconcile twice?

DerekFrank: These are two different reconcilers being called.  Its probably better as its own test, but I wanted to verify here that the pod won't get evicted when there is a fully blocking pdb

pkg/controllers/node/termination/suite_test.go
DerekFrank: Nope, just checking that the pods haven't been added to the eviction queue. Because we moved the verification that the pod isn't a duplicate into `Add` and out of `Reconcile` we need to use the functions slightly differently

pkg/controllers/node/termination/terminator/eviction.go
cnmcavoy: Yeah this event seems to be bugged in the current release of Karpenter. Here's an example of this event from our datadog event collector:

image

We have no nodes named "arc/arc-4", that's a namespace and statefulset pod name.

pkg/controllers/node/termination/suite_test.go
DerekFrank: The existing metric will disappear afaik! That's part of the reason I put `feat` on the initial PR, it introduces a breaking change to metrics

pkg/controllers/node/termination/terminator/eviction.go
DerekFrank: I'm not exactly sure. I was going to start with the defaults and then we can tweak it as we get usecase feedback. If you have an opinion I'm happy to use that as the starting point

pkg/controllers/node/termination/terminator/suite_test.go
DerekFrank: Maybe there's a better way. I'm still getting used to the test env for this. Is there a better way to set this up? What we really need is to call Reconcile in parallel, maybe something like?


```
pods := /// generate 1000 pods
cancelCtx, cancel := context.WithCancel(ctx)
wg := sync.WaitGroup{}
DeferCleanup(func() {
	cancel()
	wg.Wait() // Ensure that we wait for reconcile loop to finish so that we don't get a RACE
})
// Keep calling Reconcile() for the entirety of this test

for _, pod := range pods {
        wg.Add(1)
        go func() {
                defer wg.Done()
                ExpectObjectReconciled(ctx, env.Client, queue, pod)
        }()
}

```

pkg/controllers/node/termination/suite_test.go
DerekFrank: That's intentional, we want to verify that when we attempt to add the pod to the queue that it doesn't then get evicted because its already returning true for `Has`

pkg/controllers/node/termination/suite_test.go
jonathan-innis: I guess I'm curious why since that's not the intention of the test -- if we wanted to test this, I feel like we could make a separate test to specifically validate this -- I'd personally prefer that we keep the previous tests as-is unless we see something that's blatantly wrong and then we can check for specific things that aren't currently covered in separate testing

pkg/controllers/node/termination/suite_test.go
jonathan-innis: In that case, it seems like the check on L472 is basically not doing anything anymore -- since it takes us reconciling the pod to even get into a state of pods being terminated, I think, like you are saying, we can just check if the pod exists in the queue and (if it doesn't), then we pass the testcase

pkg/controllers/node/termination/suite_test.go
jonathan-innis: I think this is actually making me realize that this change breaks this logic in this code -- it's possible for the following scenario to happen now and us to evict a new pod that shouldn't have been evicted:
1. We add a pod to the channel
2. This pod's namespace name is added to the controller workqueue
3. The pod has previously been evicted and a new pod has launched but we didn't see it -- this means that when we pull the pod from the cache, we actually get the new pod and then we actually evict the new pod

A core tenant of whatever we do here has to be that when we add data to the queue, we also add the UUID of the pod that we pull -- if we don't do that, we have the potential to evict the wrong pod.

pkg/controllers/node/termination/suite_test.go
jonathan-innis: Yeah, I don't think that a breaking change necessitates it being a feature -- figure that we can mark the breaking changes separately

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: Given that we aren't hitting any external APIs, I'd be more apt to do max backoff as like 10s -- 1m feels like a really long time to wait for pods that have blocking PDBs -- what was the previous rate limiting logic? We had some configuration before with the old queue implementation

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -223,11 +205,11 @@ var _ = Describe("Eviction/Queue", func() {
 		})
 		It("should delete a pod with less than terminationGracePeriodSeconds remaining before nodeTerminationTime", func() {
 			pod.Spec.TerminationGracePeriodSeconds = lo.ToPtr[int64](120)
-			ExpectApplied(ctx, env.Client, pod)
+			ExpectApplied(ctx, env.Client, pod, node)
 
 			nodeTerminationTime := time.Now().Add(time.Minute * 1)
 			Expect(terminatorInstance.DeleteExpiringPods(ctx, []*corev1.Pod{pod}, &nodeTerminationTime)).To(Succeed())
-			ExpectNotFound(ctx, env.Client, pod)
+			ExpectDeleted(ctx, env.Client, pod)
```
jonathan-innis: IIRC, this function actually does the deletion of the pod vs. just checking that the pod is NotFound -- can we double-check if this is calling delete or just checking if something is gone?

pkg/controllers/node/termination/terminator/suite_test.go
jonathan-innis: Yeah, I think we really need to get to a place where pods successfully evict and then we are adding pods to the queue through the Evict call at the same time -- we basically just need to test that the queue is thread-safe -- adding and deleting at the same time doesn't cause any issues

pkg/controllers/node/termination/terminator/eviction.go
DerekFrank: I can update it to the old limits

pkg/controllers/node/termination/terminator/suite_test.go
DerekFrank: I think the new version should exercise what we're trying to get at. Let me know if you see an issue with it

pkg/controllers/node/termination/suite_test.go
```diff
@@ -798,8 +805,9 @@ var _ = Describe("Termination", func() {
 			// The pod should be deleted 60 seconds before the node's TGP expires
 			fakeClock.Step(175 * time.Second)
 			ExpectRequeued(ExpectObjectReconciled(ctx, env.Client, terminationController, node))
-			pod = ExpectExists(ctx, env.Client, pod)
-			Expect(pod.DeletionTimestamp.IsZero()).To(BeFalse())
+			ExpectObjectReconciled(ctx, env.Client, queue, pod)
+			ExpectDeleted(ctx, env.Client, pod)
```
jonathan-innis: Like I mentioned below -- this actually does a deletion, which I don't think is what we were doing before -- is this what we want? Since we aren't really checking anything here after reconciling

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -71,117 +75,107 @@ func IsNodeDrainError(err error) bool {
...
-	if shutdown {
-		return reconcile.Result{}, fmt.Errorf("EvictionQueue is broken and has shutdown")
-	}
-
-	defer q.TypedRateLimitingInterface.Done(item)
+func (q *Queue) Reconcile(ctx context.Context, pod *corev1.Pod) (reconcile.Result, error) {
+	ctx = injection.WithControllerName(ctx, q.Name())
 
+	if !q.Has(pod) {
+		//This is a different pod than in the queue, we should exit without evicting
```
jonathan-innis: nit: It may be worth adding some more verbose commenting on why we need to handle this case, given that this race condition is a little hard to think about and the impact of not handling this is a little bit on the edge

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -71,117 +75,107 @@ func IsNodeDrainError(err error) bool {
...
+		return reconcile.Result{}, nil
+	}
 	// Evict the pod
-	if q.Evict(ctx, item) {
-		q.TypedRateLimitingInterface.Forget(item)
-		q.mu.Lock()
-		q.set.Delete(item)
-		q.mu.Unlock()
-		return reconcile.Result{RequeueAfter: singleton.RequeueImmediately}, nil
+	if err := q.Evict(ctx, pod); err != nil {
```
jonathan-innis: If we handle errors like this, we are going to be printing a lot of errors as we encounter different pods that have PDBs that are blocking their eviction -- I think we need to avoid returning an error here and handle the Requeuing in a different way with a rate limiter OR with RequeueAfter

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -71,117 +75,107 @@ func IsNodeDrainError(err error) bool {
...
+	q.set.Delete(NewQueueKey(pod))
+	return reconcile.Result{}, nil
 }
 
-// Evict returns true if successful eviction call, and false if there was an eviction-related error
-func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
-	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(key.Namespace, key.Name)))
+// Evict returns nil if successful eviction call, and an error if there was an eviction-related error
+func (q *Queue) Evict(ctx context.Context, pod *corev1.Pod) error {
+	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Pod", klog.KRef(pod.Namespace, pod.Name)))
```
jonathan-innis: This should automatically be injected by the Reconcile() method so there's actually no need to add it to the context here

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -197,27 +191,33 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
 		}
 		if apierrors.IsTooManyRequests(err) { // 429 - PDB violation
-			q.recorder.Publish(terminatorevents.NodeFailedToDrain(&corev1.Node{ObjectMeta: metav1.ObjectMeta{
-				Name:      key.Name,
-				Namespace: key.Namespace,
-			}}, serrors.Wrap(fmt.Errorf("evicting pod violates a PDB"), "Pod", klog.KRef(key.Namespace, key.Name))))
-			return false
+			node, err2 := podutils.NodeForPod(ctx, q.kubeClient, pod)
+			if err2 != nil {
+				log.FromContext(ctx).V(1).Error(err2, "pod has no node")
```
jonathan-innis: This will double-log the error -- controller-runtime automatically logs the error if you return an error from the Reconcile by passing it up the stack

pkg/controllers/node/termination/terminator/suite_test.go
jonathan-innis: You resolved this but I didn't get it you did or didn't want to change it

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -223,6 +215,8 @@ var _ = Describe("Eviction/Queue", func() {
 		})
 		It("should delete a pod with less than terminationGracePeriodSeconds remaining before nodeTerminationTime", func() {
 			pod.Spec.TerminationGracePeriodSeconds = lo.ToPtr[int64](120)
+			// overwrite the node name or the delete does not succeed
+			pod.Spec.NodeName = ""
```
jonathan-innis: Why does the delete not succeed if we don't override the nodeName with empty?

pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -128,53 +129,43 @@ var _ = Describe("Eviction/Queue", func() {
...
-				defer wg.Done()
+			// Ensure that we add enough pods to the queue while we are pulling items off of the queue (enough to trigger a DATA RACE)
+			pods := test.Pods(1000)
 
-				for {
-					ExpectSingletonReconciled(ctx, queue)
-					if cancelCtx.Err() != nil {
-						return
+			for _, pod := range pods {
+				go func() {
```
jonathan-innis, line 160: These goroutines are never canceled, so when they start up, they are going to continue to operate and run for the lifetime of the test. I think you need to either have them attempt to add the pod to the queue a single time and then exit or you build in cancelation with `context.WithCancel()` and cancel the new context that you build at the end of the function, also causing the goroutines to bail

pkg/controllers/node/termination/terminator/suite_test.go
DerekFrank: I'm not exactly sure. I think it has to do with the mock client we use for testing. I've verified that client.Delete() is getting called for the pod, but the pod doesn't actually get deleted. FWIW, before this PR if someone had set the nodename in the test it would have failed in the same way. This PR sets the nodename for the pod at the start of each test.

pkg/controllers/node/termination/terminator/eviction.go
DerekFrank: I'm returning err not err2 out of the function, so both get logged instead of swallowing err in favor of err2

pkg/controllers/node/termination/terminator/suite_test.go
DerekFrank: Whoops I missed that one!

pkg/controllers/node/termination/terminator/suite_test.go
DerekFrank: I'll have the node\pod bound in tests where they need to be, and not have the overwrite here

pkg/controllers/node/termination/terminator/eviction.go
jonathan-innis: I see -- that's a good point, maybe an if/else makes this more clear -- that way you would only need a single return statement?

DerekFrank: Do we ever want to return err2? I don't think so as its not the thing blocking eviction

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -192,32 +176,42 @@ func (q *Queue) Evict(ctx context.Context, key QueueKey) bool {
...
-			// 404 - The pod no longer exists
-			// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L160
-			// 409 - The pod exists, but it is not the same pod that we initiated the eviction on
-			// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L318
-			return true
+		// 404 - The pod no longer exists
+		// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L160
+		// 409 - The pod exists, but it is not the same pod that we initiated the eviction on
+		// https://github.com/kubernetes/kubernetes/blob/ad19beaa83363de89a7772f4d5af393b85ce5e61/pkg/registry/core/pod/storage/eviction.go#L318
+		if !(apierrors.IsNotFound(err) || apierrors.IsConflict(err)) {
```
jonathan-innis: I think we should just exit immediately if we have a pod that no longer exists or is a different pod than the one that we intended to evict

jonathan-innis: As it stands right now, you are adding metrics and logging that you are evicting pods when you are not actually doing any kind of eviction

jonathan-innis (approved):
/lgtm
/approve

PR #2253, opened 2025-05-24
feat: Add karpenter_pods_disrupted_total metric

### LINKED ISSUE #2020: Add `karpenter_pods_disrupted_total` when NodeClaims are disrupted
### Description

It's useful to know how many pods are disrupted by a particular reason in the same way it's useful to know how many NodeClaims are disrupted by a particular reason -- this update would also require an update to the CloudProviders. We basically need to look-up the count of reschedulable pods that we are disrupting right next to the call that we make to increment the `karpenter_nodeclaims_disrupted_total` metric 

Brings more visibility to the impact of the different disruption reasons across the system

omerap12:
Fixes #2020

**Description**
Added a new metric `karpenter_pods_disrupted_total` to track the number of pods disrupted during node termination events.

Added unit tests for the new metric


omerap12:
Friendly ping @jonathan-innis @engedaam

pkg/controllers/nodeclaim/garbagecollection/controller.go
```diff
@@ -99,6 +99,11 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 			errs[i] = client.IgnoreNotFound(err)
 			return
 		}
+		pods := &corev1.PodList{}
+		var podCount int
+		if err := c.kubeClient.List(ctx, pods, client.MatchingFields{"spec.nodeName": node.Name}); err == nil {
```
jonathan-innis, line 104: Should we include DS pods in this calculation or should we just stick to reschedulable pods (running pods)

pkg/controllers/node/health/controller.go
```diff
@@ -131,6 +131,10 @@ func (c *Controller) deleteNodeClaim(ctx context.Context, nodeClaim *v1.NodeClai
 	if err := c.kubeClient.Delete(ctx, nodeClaim); err != nil {
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
+	pods := &corev1.PodList{}
+	if err := c.kubeClient.List(ctx, pods, client.MatchingFields{"spec.nodeName": node.Name}); err != nil {
```
jonathan-innis, line 135: Since pods are quite big, I think we should either look at creating a helper that returns a count of reschedulable pods that are currently bound to a node and employs the `client.UnsafeDisableDeepCopy` method OR we should use the PartialMetadataList -- I'm personally more in favor of the first one

pkg/controllers/disruption/queue.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/awslabs/operatorpkg/singleton"
```
jonathan-innis, line 28: Is there some way that we could avoid duplicating the pod list logic and rather handle the pod disruption metric inside of the termination controller? One thought that I had when I was thinking about the implementation of this is that we could make sure that we add the `Disrupted` status condition on the NodeClaim always (expiration and health included) and then when we go to look-up the disruption reason for pods, we can just look-up the status condition reason to determine which type of resaon we should fire for that metric

pkg/controllers/nodeclaim/garbagecollection/controller.go
omerap12: I think we should only include rescheduled pods.DS pods are automatically handled and are meant to run on every node (or specific ones using nodeSelector or affinity). So, if we include DaemonSet pods in the disruption metrics, it might give a wrong idea about real workload disruptions.

pkg/controllers/disruption/queue.go
omerap12: Yeah I agree this is a better idea.

pkg/controllers/node/health/controller.go
omerap12: Got it, will adjust

pkg/controllers/disruption/queue.go
omerap12: On second thought - and please correct me if I'm wrong here, there could be a timing problem with this. If another controller, like garbage collection, deletes the `NodeClaim` before the termination controller handles it, we might lose the reason for the disruption that was saved in the NodeClaim's status.
So we have a situation that Karpenter terminate a node but the `NodeClaim` isn't available.

omerap12: Friendly ping @jonathan-innis  :)

Issue #2245, opened 2025-05-22
Consider pod disruptable with allow-disruption pod annotation


marc-ostrow:
### Description

**What problem are you trying to solve?**

Would like the ability to prevent pod eviction with a blocking PDB, yet not block Karpenter from consolidating the associated node.

**How important is this feature to you?**

We deploy thousands of single-pod deployments and would rather have them rollout restarted, than have their pods evicted and replaced.  To implement this, we deploy a blocking PDB for such pods, and utilize a rollout-node-controller that monitors Karpenter nodes for deletion. When the APIServer sets the `DeletionTimestamp` on the Karpenter node, the rollout-node-controller will wait a bit for pod evictions to occur, before rollout restarting the deployments for any remaining pods associated with a single-pod deployment. Once the node is emptied, the deprovisioning process will continue.

Unfortunately, blocking PDBs currently generate `DisruptionBlocked` events on the node, and Karpenter will not consider such a node for consolidation.  Albeit, we can manually delete several existing Karpenter nodes at once, and the new replicas rolled out will be considered as a workload for Karpenter to address; but, depending on timing, Karpenter could see more than one workload to scale out for; and, as such, would otherwise be consolidated, if not for the blocking PDBs.

This feature request for an additional pod annotation, `karpenter.sh/allow-disruption` when set to `"true"`, would allow such worker-group nodes to be consolidated, and our pods gracefully terminated as part of a rollout restart of their deployment.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

marc-ostrow:
I believe the following use case to be highly unlikely, but should both annotations be asserted on the pod

- `karpenter.sh/do-not-disrupt="true"`
- `karpenter.sh/allow-disruption="true"`

I believe the `karpenter.sh/allow-disruption="true"` annotation should be _ignored_, as the `karpenter.sh/do-not-disrupt` annotation is directly associated with the pod, whereas an existing blocking PDB is indirectly associated with the pod via a selector.

marc-ostrow:
Looking at the latest code base, the check is isolated to one function `karpenter/pkg/utils/pdb/pdb.go:85`

[...]

PR #2256, opened 2025-05-27
feat: consider pod disruptable with allow-disruption pod annotation (#2245)

Sk8rMarc:
Implements #2245 

**Description**

For those who wish to have their pods from single-pod deployments gracefully terminated as part of rollout restarts of their deployments, rather than evicted, can utilize a custom or open source rollout node controller, and deploy blocking PDBs for such pods, and still have the benefits of Karpenter Consolidation, if this feature is implemented.

This Pull Request implements the necessary changes so that pods having blocking PDBs can indicate that they should be treated as pods without blocking PDBs with respect to Karpenter Consolidation.  Such pods would indicate this by being annotated with the annotation `karpenter.sh/allow-disruption` being set to a value of `"true"`.

- Unit tests were added to pkg/utils/pdb/suite_test.go and pass
- Built an aws karpenter provider image based on these changes and tested in an EKS 1.32 cluster where karpenter v1.4.0 was already installed and running.  The controller deployment was updated to use this built image which was pushed to AWS ECR for in-cluster testing.

### DIFF
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -48,6 +48,7 @@ const (
 
 // Karpenter specific annotations
 const (
+	AllowDisruptionAnnotationKey               = apis.Group + "/allow-disruption"
 	DoNotDisruptAnnotationKey                  = apis.Group + "/do-not-disrupt"
 	ProviderCompatibilityAnnotationKey         = apis.CompatibilityGroup + "/provider"
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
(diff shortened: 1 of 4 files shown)


Sk8rMarc:
After consulting with @ellistarn on #karpenter-dev, I'm marking this PR as Draft as I attempt a cleaner and simpler avenue utilizing a validating webhook on the pods/eviction subresource with a failure policy set to `Ignore`.  The webhook will look to see if the associated Pod has a specific annotation and if so, will rollout-restart the associated deployment and invalidate the eviction request; otherwise validate all other requests.  From the following [2022 issue](https://github.com/kubernetes/kubernetes/issues/110169), I will not use an `objectSelector`, and probably not a namespaceSelector

If this works, there would be no changes to karpenter needed, no rollout-node-controller needed, and no blocking PDBs!

Sk8rMarc:
Went with the validating admission webhook solution, this PR is not needed.

PR #2264, opened 2025-05-29
perf: Perform quick checks in node health first

jonathan-innis:
Fixes #N/A 

**Description**

Avoid making calls to list out the cluster or nodepool health when the node itself isn't unhealthy. This changes the order of our checks so that we perform the quick checks first.

`make presubmit`


rschalo (approved):
/lgtm

PR #2265, opened 2025-05-29
chore: Consider node.kubernetes.io/not-ready:NoExecute as ephemeral

jigisha620:
Fixes issue in aws-provider - https://github.com/aws/karpenter-provider-aws/issues/8066

**Description**
When using Kyverno to apply tolerations to the pod, Karpenter was creating two nodeClaims because the node comes up with this taint that gets removed later - 
```
- key: "node.kubernetes.io/not-ready"
  operator: "Exists"
  effect: "NoExecute"
```
This PR will consider this taint as ephemeral if the node isn't initialized which will prevent karpenter from creating two nodeclaims for the same pod.

Added tests. Tested this on my dev cluster.

### DIFF
--- a/pkg/controllers/provisioning/scheduling/suite_test.go
+++ b/pkg/controllers/provisioning/scheduling/suite_test.go
@@ -2037,6 +2037,10 @@ var _ = Context("Scheduling", func() {
 				Expect(node1.Name).To(Equal(node2.Name))
 			})
 			It("should assume pod will schedule to a node with ephemeral taint node.kubernetes.io/not-ready:NoExecute when the node is uninitialized", func() {
+				// We try to provision a node for an initial unschedulable pod that will create nodeClaim and node bindings.
+				// The node will not be initialized but the initial pod will still be schedulable against the node
+				// because the node has node.kubernetes.io/not-ready:NoSchedule taint which is ephemeral.
+				nodePool.Spec.Template.Spec.Taints = []corev1.Taint{{Key: corev1.TaintNodeNotReady, Effect: corev1.TaintEffectNoExecute}}
 				ExpectApplied(ctx, env.Client, nodePool)
 				initialPod := test.UnschedulablePod()
 				bindings := ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, initialPod)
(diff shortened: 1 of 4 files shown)


pkg/controllers/state/suite_test.go
```diff
@@ -1832,9 +1834,10 @@ var _ = Describe("Taints", func() {
 			ExpectReconcileSucceeded(ctx, nodeController, client.ObjectKeyFromObject(node))
 
 			stateNode := ExpectStateNodeExists(cluster, node)
-			Expect(stateNode.Taints()).To(HaveLen(3))
```
jonathan-innis, line 1835: Can we add a test to ensure the schedulability of pods that don't tolerate this taint while the node is uninitialized

jigisha620: Added

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -2036,6 +2036,27 @@ var _ = Context("Scheduling", func() {
...
+				// delete the pod so that the node is empty
+				ExpectDeleted(ctx, env.Client, initialPod)
+				node1 := bindings.Get(initialPod).Node
+				node1.Spec.Taints = []corev1.Taint{{Key: corev1.TaintNodeNotReady, Effect: corev1.TaintEffectNoExecute}}
+				node1.Labels = lo.Assign(node1.Labels, map[string]string{v1.NodeInitializedLabelKey: "true"})
+
+				ExpectApplied(ctx, env.Client, node1)
+				ExpectReconcileSucceeded(ctx, nodeStateController, client.ObjectKeyFromObject(node1))
+
+				opts := test.PodOptions{Tolerations: []corev1.Toleration{{Key: corev1.TaintNodeUnreachable, Effect: corev1.TaintEffectNoSchedule}}}
```
rschalo: Why do we need this toleration for another ephemeral taint?

jigisha620: We don't really need it. I was just trying to replicate the behavior in the test where the pod has some toleration for some other taint that we don't consider ephemeral. I don't think it affects what we are testing here though.

pkg/test/expectations/expectations.go
```diff
@@ -293,6 +293,23 @@ func ExpectProvisioned(ctx context.Context, c client.Client, cluster *state.Clus
 	return bindings
 }
 
+func ExpectProvisionedFakeClient(ctx context.Context, c client.Client, cluster *state.Cluster, cloudProvider cloudprovider.CloudProvider, provisioner *provisioning.Provisioner, pods ...*corev1.Pod) Bindings {
```
jigisha620: Wrote this function because using fakeClient we need to manually bind the pod to the node.

jonathan-innis: Is there a way that you could just extend the existing function? e.g. you do an interface check for the fake client and then manually update the bindings if that's the concrete type that you have?

jigisha620: done

PR #2266, opened 2025-05-29
fix: fail open for missing nodeclaims in termination

jmdeal:
Fixes #N/A 

**Description**
Updates the node termination controller to fail open rather than closed when a NodeClaim can not be found for the node. This is an invariant that shouldn't be broken, but discovered that it's currently possible if the node joins the cluster in time for the same reconciliation of the lifecycle controller to perform launch and registration **and** it fails to persist the provider-id to the NodeClaim's status. This likely will never happen in a real cloudprovider, but does occur in KWOK simulation testing. 

`make presubmit`

### DIFF
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -91,21 +91,19 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
 	}
 
+	// We're not guaranteed to find a NodeClaim for the node (e.g. if we failed to persist the provider ID to the nodeclaim
+	// at launch). If there are duplicate NodeClaims, we will treat it as though there is no NodeClaim since there is no
+	// longer a single source of truth.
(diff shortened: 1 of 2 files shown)


pkg/controllers/node/termination/controller.go
```diff
@@ -91,21 +91,19 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
 		return reconcile.Result{}, nil
```
jonathan-innis, line 91: Should we add some testing for the no NodeClaim and duplicate NodeClaim cases?

jonathan-innis (approved):

pkg/controllers/node/termination/controller.go
jmdeal: Good call, done

jonathan-innis (approved):
/lgtm
/approve

pkg/controllers/node/termination/controller.go
jonathan-innis: Completed.

PR #2272, opened 2025-05-30
perf: Be more resilient to deletion failures in disruption controller

jonathan-innis:
Fixes #N/A 

**Description**

Be more resilient to single node failures during disruption deletion operations. Currently, if we have a single node failure while we are moving through `executeCommand`, it will stop the _whole_ command from executing. This is particularly problematic when you have a large emptiness command that you aren't able to make any progress on. This changes the code so that we can continue to make progress on deletion commands even if all of the nodes don't succeed.

It also adds retries for the apiserver actions to attempt to be more resilient to transient failures.

`make presubmit`

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -31,6 +31,7 @@ import (
 	"go.uber.org/multierr"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/uuid"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/utils/clock"
 	controllerruntime "sigs.k8s.io/controller-runtime"
@@ -128,6 +129,7 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	outdatedNodes := lo.Filter(c.cluster.Nodes(), func(s *state.StateNode, _ int) bool {
 		return !c.queue.HasAny(s.ProviderID()) && !s.Deleted()
 	})
+	c.cluster.UnmarkForDeletion(lo.Map(outdatedNodes, func(s *state.StateNode, _ int) string { return s.ProviderID() })...)
 	if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, false, outdatedNodes...); err != nil {
 		if errors.IsConflict(err) {
 			return reconcile.Result{Requeue: true}, nil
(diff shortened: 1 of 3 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -260,31 +261,36 @@ func (c *Controller) createReplacementNodeClaims(ctx context.Context, m Method,
...
 	workqueue.ParallelizeUntil(ctx, len(candidates), len(candidates), func(i int) {
+		if err := state.RequireNoScheduleTaint(ctx, c.kubeClient, true, candidates[i].StateNode); err != nil {
+			errs[i] = serrors.Wrap(fmt.Errorf("tainting nodes, %w", err), "taint", pretty.Taint(v1.DisruptedNoScheduleTaint))
+			return
+		}
 		// refresh nodeclaim before updating status
 		nodeClaim := &v1.NodeClaim{}
-
-		if err := c.kubeClient.Get(ctx, client.ObjectKeyFromObject(candidates[i].NodeClaim), nodeClaim); err != nil {
+		if err := retry.OnError(retry.DefaultBackoff, func(err error) bool { return client.IgnoreNotFound(err) != nil }, func() error {
```
jmdeal, line 273: Is there a specific set of errors we want to retry for here? The only error I think I'd expect us to see which wouldn't be transient is an RBAC permission issue. I don't think we need to consider that case since Karpenter needs RBAC permissions to read / write to nodes to operate but I'm wondering if there are any other cases we wouldn't want to retry on.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -283,26 +284,24 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
-			multiErr = multierr.Append(multiErr, client.IgnoreNotFound(err))
-		} else {
-			metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
-				metrics.ReasonLabel:       pretty.ToSnakeCase(string(cmd.reason)),
-				metrics.NodePoolLabel:     cmd.candidates[i].NodeClaim.Labels[v1.NodePoolLabelKey],
-				metrics.CapacityTypeLabel: cmd.candidates[i].NodeClaim.Labels[v1.CapacityTypeLabelKey],
-			})
+	errs := make([]error, len(cmd.candidates))
+	workqueue.ParallelizeUntil(ctx, len(cmd.candidates), len(cmd.candidates), func(i int) {
+		q.recorder.Publish(disruptionevents.Terminating(cmd.candidates[i].Node, cmd.candidates[i].NodeClaim, cmd.Reason())...)
```
jmdeal, line 289: Not specific to this PR, but shouldn't we publish this event after we successfully set the deletion timestamp on the node?

pkg/controllers/disruption/controller.go
jmdeal: > Yeah, I think if you start talking about RBAC permissions, things just start failing generally -- I'd rather not handle it here

Yeah, I agree. I haven't been able to think of anything else other than RBAC permissions, so this seems good to me.

pkg/controllers/disruption/controller.go
```diff
@@ -128,6 +129,7 @@ func (c *Controller) Reconcile(ctx context.Context) (reconcile.Result, error) {
 	outdatedNodes := lo.Filter(c.cluster.Nodes(), func(s *state.StateNode, _ int) bool {
 		return !c.queue.HasAny(s.ProviderID()) && !s.Deleted()
 	})
+	c.cluster.UnmarkForDeletion(lo.Map(outdatedNodes, func(s *state.StateNode, _ int) string { return s.ProviderID() })...)
```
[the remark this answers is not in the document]
jmdeal: I think my unease with this largely comes from this being an atypical controller - since it's a singleton controller we do have that guarantee, but if it were a more conventional controller-runtime controller we would not. I do agree that the simplicity of handling cleanup in a single place is nice, I'm just not sure that outweighs the overhead of reasoning about cleanup always occuring in a subsequent reconcile. I might be overthinking this though - I didn't find it that hard to reason about personally.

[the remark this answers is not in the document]
jonathan-innis: Yeah, I personally think it's preferrable and less bug-prone than trying to handle cleanup in every place where you might fail. FWIW, you might do the same kind of cleanup/error handling in a typical reconciler too since you would error and then cleanup in the same way on a subsequent reconcile.

This is also pretty standard behavior because it ensures that you are resilient to restarts.

PR #2283, opened 2025-06-03
chore: Ensure we can stand up multiple partitions with kwok

jonathan-innis:
Fixes #N/A 

**Description**

Fix our `./install-kwok.sh` script so that it works with multiple partitions again

`make install-kwok`


rschalo:
Out of curiosity - how do these partitions split up the heartbeats?

PR #2284, opened 2025-06-03
chore: Update NodeClaim E2E test to only replace one status condition

engedaam:
Fixes #N/A 

**Description**
- Ensure validation policies are applied before NodeClass reconciliation to prevent race conditions
- Implement selective status condition updates instead of overwriting all NodeClass conditions

- `make upstream-e2etestss`

### DIFF
--- a/kwok/cloudprovider/cloudprovider.go
+++ b/kwok/cloudprovider/cloudprovider.go
@@ -68,6 +68,9 @@ func (c CloudProvider) Create(ctx context.Context, nodeClaim *v1.NodeClaim) (*v1
 		}
 		return nil, fmt.Errorf("resolving node class from nodeclaim, %w", err)
 	}
+	if status := nodeClass.StatusConditions().Get(status.ConditionReady); status.IsFalse() {
+		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(status.Message))
+	}
 	// Kick-off a goroutine to allow us to asynchronously register nodes
 	// We're fine to leak this because failed registration can also happen in real providers
 	go func() {
@@ -78,10 +81,6 @@ func (c CloudProvider) Create(ctx context.Context, nodeClaim *v1.NodeClaim) (*v1
 			log.FromContext(ctx).Error(err, "failed creating node from nodeclaim")
 		}
 	}()
-	nodeClassReady := nodeClass.StatusConditions().Get(status.ConditionReady)
-	if nodeClassReady.IsFalse() {
-		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(nodeClassReady.Message))
-	}
 	// convert the node back into a node claim to get the chosen resolved requirement values.
 	return c.toNodeClaim(node)
 }
--- a/test/pkg/environment/common/expectations.go
+++ b/test/pkg/environment/common/expectations.go
@@ -26,7 +26,6 @@ import (
 	"time"
 
 	"github.com/awslabs/operatorpkg/object"
-	"github.com/awslabs/operatorpkg/status"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/samber/lo"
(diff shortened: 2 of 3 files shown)


kwok/cloudprovider/cloudprovider.go
```diff
@@ -68,6 +68,10 @@ func (c CloudProvider) Create(ctx context.Context, nodeClaim *v1.NodeClaim) (*v1
 		}
 		return nil, fmt.Errorf("resolving node class from nodeclaim, %w", err)
 	}
+	nodeClassReady := nodeClass.StatusConditions().Get(status.ConditionReady)
```
jonathan-innis: Oops -- nice catch :D

test/suites/regression/nodeclaim_test.go
```diff
@@ -248,19 +247,20 @@ var _ = Describe("NodeClaim", func() {
...
-			nodeClass = env.ExpectNodeClassCondition(nodeClass, []status.Condition{
-				{
-					Type:               "Ready",
-					Status:             metav1.ConditionFalse,
-					LastTransitionTime: metav1.Now(),
-					Reason:             "NotReady",
-					Message:            "NodeClass is not ready",
-				},
+			// Give 5 seconds to allow Karpenter to update the NodeClass
+			time.Sleep(5 * time.Second)
```
jonathan-innis: Why do we need to sleep and why can't we create an Eventually to wait until we get into a certain condition? I'm always weary of sleeps like this since they too can be flaky.

test/pkg/environment/common/expectations.go
```diff
@@ -977,7 +981,7 @@ func (env *Environment) ExpectBlockNodeClassStatus(obj *unstructured.Unstructure
 			},
 			Validations: []admissionregistrationv1.Validation{
 				{
-					Expression: "false",
+					Expression: "object.status.conditions.filter(c, c.type == 'Ready').all(c, c.reason == 'TestingNotReady')",
```
jonathan-innis, line 988: Can you describe what this does through a comment? It's not immediately clear to me what the intention of this is

test/pkg/environment/common/expectations.go
```diff
@@ -998,6 +1002,10 @@ func (env *Environment) ExpectBlockNodeClassStatus(obj *unstructured.Unstructure
 	}
 	// Create both the policy and binding in the cluster
 	env.ExpectCreated(admissionspolicy, admissionspolicybinding)
+
+	// Wait for the admission policy to become active
+	// Note: There can be a delay between resource creation and policy enforcement
+	time.Sleep(5 * time.Second)
```
jonathan-innis: Why can't we just wait for a condition on the policy rather than using a time-based wait.

test/suites/regression/nodeclaim_test.go
engedaam: Sure, I updated to add say if we get a condition struct then, the Karpenter has reconciled against the nodeclass.

test/pkg/environment/common/expectations.go
engedaam: There is no condition on the policy that will tell us this information unfortunately. I added an evenutally statement that will dry run and validate the the pokily is actively working

test/pkg/environment/common/expectations.go
engedaam: Added a comment for the what expression is meant to do here

test/suites/regression/nodeclaim_test.go
```diff
@@ -248,19 +248,25 @@ var _ = Describe("NodeClaim", func() {
...
 			env.ExpectCreated(nodeClass)
-			nodeClass = env.ExpectNodeClassCondition(nodeClass, []status.Condition{
-				{
-					Type:               "Ready",
-					Status:             metav1.ConditionFalse,
-					LastTransitionTime: metav1.Now(),
-					Reason:             "NotReady",
-					Message:            "NodeClass is not ready",
-				},
+			By("Validate the NodeClass status condition have been reconciled")
```
rschalo:
```suggestion
			By("Validating the NodeClass status condition has been reconciled")
```

kwok/cloudprovider/cloudprovider.go
```diff
@@ -68,6 +68,10 @@ func (c CloudProvider) Create(ctx context.Context, nodeClaim *v1.NodeClaim) (*v1
...
+	if nodeClassReady.IsFalse() {
+		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(nodeClassReady.Message))
+	}
```
jmdeal: nit
```suggestion
	if status := nodeClass.StatusConditions().Get(status.ConditionReady); status.IsFalse() {
		return nil, cloudprovider.NewNodeClassNotReadyError(stderrors.New(status.Message))
	}
```

PR #2286, opened 2025-06-04
chore: Update the Node Repair Controller for requeue time 

engedaam:
Fixes #N/A 

**Description**
- Implemented 5-minute requeue interval when >20% of node pools or clusters are unhealthy
- Added filtering to process node updates only when status conditions change
- Added unit tests to verify node disruption protection when cluster health is degraded (>20% unhealthy)

- Unit tests verify nodes are not disrupted when cluster health threshold is exceeded
- Validated requeue behavior with unhealthy node pools

### DIFF
--- a/pkg/controllers/node/health/controller.go
+++ b/pkg/controllers/node/health/controller.go
@@ -32,8 +32,10 @@ import (
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
@@ -68,7 +70,12 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("node.health").
-		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider))).
+		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider), predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return !equality.Semantic.DeepEqual(e.ObjectOld.(*corev1.Node).Status.Conditions, e.ObjectNew.(*corev1.Node).Status.Conditions)
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool { return true },
+		})).
 		Complete(reconcile.AsReconciler(m.GetClient(), c))
 }
 
@@ -104,7 +111,7 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 			return reconcile.Result{}, client.IgnoreNotFound(err)
 		}
 		if !nodePoolHealthy {
-			return reconcile.Result{}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
+			return reconcile.Result{RequeueAfter: 5 * time.Minute}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
 		}
 	} else {
 		clusterHealthy, err := c.isClusterHealthy(ctx)
@@ -113,7 +120,7 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 		}
 		if !clusterHealthy {
 			c.recorder.Publish(NodeRepairBlockedUnmanagedNodeClaim(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String()))...)
-			return reconcile.Result{}, nil
+			return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
 		}
 	}
 	// For unhealthy past the tolerationDisruption window we can forcefully terminate the node
--- a/pkg/controllers/node/health/suite_test.go
+++ b/pkg/controllers/node/health/suite_test.go
@@ -290,7 +290,7 @@ var _ = Describe("Node Health", func() {
 			nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 			Expect(nodeClaim.DeletionTimestamp).ToNot(BeNil())
 		})
-		It("should ignore unhealthy nodes if more then 20% of the nodes are unhealthy", func() {
+		It("should ignore unhealthy nodes if more then 20% of the nodes are unhealthy in a nodepool", func() {
 			ExpectApplied(ctx, env.Client, nodePool)
 			nodeClaims := []*v1.NodeClaim{}
 			nodes := []*corev1.Node{}
@@ -312,9 +312,42 @@ var _ = Describe("Node Health", func() {
 
 			fakeClock.Step(60 * time.Minute)
 
-			// Determine to delete unhealthy nodes
+			// Determine if we should delete unhealthy nodes
 			for i := range 4 {
-				ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				result := ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				if i < 3 {
+					Expect(result.RequeueAfter).To(BeNumerically("~", time.Minute*5, time.Second))
+				}
+				nodeClaim = ExpectExists(ctx, env.Client, nodeClaims[i])
+				Expect(nodeClaim.DeletionTimestamp).To(BeNil())
+			}
+		})
+		It("should ignore unhealthy nodes if more then 20% of the nodes are unhealthy in a cluster", func() {
+			ExpectApplied(ctx, env.Client, nodePool)
+			nodeClaims := []*v1.NodeClaim{}
+			nodes := []*corev1.Node{}
+			for i := range 10 {
+				nodeClaim, node = test.NodeClaimAndNode(v1.NodeClaim{ObjectMeta: metav1.ObjectMeta{Finalizers: []string{v1.TerminationFinalizer}}})
+				if i < 3 {
+					node.Status.Conditions = append(node.Status.Conditions, corev1.NodeCondition{
+						Type:               "BadNode",
+						Status:             corev1.ConditionFalse,
+						LastTransitionTime: metav1.Time{Time: fakeClock.Now()},
+					})
+				}
+				nodeClaims = append(nodeClaims, nodeClaim)
+				nodes = append(nodes, node)
+				ExpectApplied(ctx, env.Client, nodePool, nodeClaim, node)
+			}
+
+			fakeClock.Step(60 * time.Minute)
+
+			// Determine if we should delete unhealthy nodes
+			for i := range 4 {
+				result := ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				if i < 3 {
+					Expect(result.RequeueAfter).To(BeNumerically("~", time.Minute*5, time.Second))
+				}
 				nodeClaim = ExpectExists(ctx, env.Client, nodeClaims[i])
 				Expect(nodeClaim.DeletionTimestamp).To(BeNil())
 			}


pkg/controllers/node/health/controller.go
```diff
@@ -68,7 +70,12 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("node.health").
-		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider))).
+		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider), predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				return !equality.Semantic.DeepEqual(e.ObjectOld.(*corev1.Node).Status.Conditions, e.ObjectNew.(*corev1.Node).Status.Conditions)
```
jonathan-innis: Just sanity check this real quick for how this is handled with arrays, mainly just want to make sure it's looking at each of the object contents and not just looking at the items that are in the array

pkg/controllers/node/health/controller.go
```diff
@@ -68,7 +70,12 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool { return true },
```
jonathan-innis: Not definining a func is `true` by default. If you want to make it `false`, then you should make that explicit

pkg/controllers/node/health/controller.go
```diff
@@ -104,7 +111,7 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 			return reconcile.Result{}, client.IgnoreNotFound(err)
 		}
 		if !nodePoolHealthy {
-			return reconcile.Result{}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
+			return reconcile.Result{RequeueAfter: 5 * time.Minute}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
```
jonathan-innis: This alone won't work, right? We are going to get a heartbeat event that comes in every 30s or so and that's going to requeue us more aggressively even with the event filter that you have

jonathan-innis: We either need a cache rate limiter that ensures that we literally don't check as aggressively as we are today OR we need to make sure that our event filter is a bit better so that it checks something like: The status state has actually changed

pkg/controllers/node/health/controller.go
```diff
@@ -104,7 +111,7 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 			return reconcile.Result{}, client.IgnoreNotFound(err)
 		}
 		if !nodePoolHealthy {
-			return reconcile.Result{}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
```
jonathan-innis, line 116: Separately from just generally making sure that we don't reconcile as much as we are doing today, I _do_ think that we should generally rate limit the cluster check that we are doing for health -- we shouldn't list as much as we are when it comes to just checking whether the cluster is healthy

pkg/controllers/node/health/controller.go
engedaam: Dropped this function

pkg/controllers/node/health/controller.go
engedaam: Yeah, I opted to make the events filter better and explicitly better by checking each status condition. This check now drops the heartbeat check which will help and avoid needing to do the cache rate limiter.

pkg/controllers/node/health/controller.go
engedaam: Dropped this check and opted for explicit check on each conditions.

pkg/controllers/node/health/controller.go
engedaam: Can you elaborate more here? I'm not sure I follow the concern here. I have made a few updates to the filtering logic and we have caching for the listing objects in place. Are you still thinking this is a concern?

pkg/controllers/node/health/controller.go
```diff
@@ -68,7 +72,35 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
 func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("node.health").
-		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider))).
+		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider), predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				if len(e.ObjectOld.(*corev1.Node).Status.Conditions) != len(e.ObjectNew.(*corev1.Node).Status.Conditions) {
```
jonathan-innis: nit: Can you just typecast things into variables like

```
oldNode := e.ObjectOld.(*corev1.Node)
newNode := e.ObjectNew.(*corev1.Node)
```

I think it will make things easier to read overall

pkg/controllers/node/health/controller.go
```diff
@@ -68,7 +72,35 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
+					// 1. Condition type no longer exists in new node
+					// 2. Transition time has changed
+					// 3. Status has changed
+					// 4. Reason has changed
+					// 5. Message has changed
+					if newCond.Type == "" ||
+						oldCond.LastTransitionTime != newCond.LastTransitionTime ||
+						oldCond.Status != newCond.Status ||
+						oldCond.Reason != newCond.Reason ||
+						oldCond.Message != newCond.Message {
```
jonathan-innis: Why do you care if the message changes?

pkg/controllers/node/health/controller.go
```diff
@@ -68,7 +72,35 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
+					// Return true if any of these conditions are met:
+					// 1. Condition type no longer exists in new node
+					// 2. Transition time has changed
+					// 3. Status has changed
+					// 4. Reason has changed
+					// 5. Message has changed
+					if newCond.Type == "" ||
+						oldCond.LastTransitionTime != newCond.LastTransitionTime ||
+						oldCond.Status != newCond.Status ||
+						oldCond.Reason != newCond.Reason ||
```
jonathan-innis: Why do you care if the reason changes?

pkg/controllers/node/health/controller.go
jonathan-innis: Given the filters that you added, I think this mitigates the bulk of the concern -- since we aren't deep-copying anymore when we are listing across the cluster, that should pretty much eliminate the CPU and memory usage consumed here -- we can take another flame graph after the change to see if there's more that needs to be done

pkg/controllers/node/health/controller.go
```diff
@@ -113,7 +145,7 @@ func (c *Controller) Reconcile(ctx context.Context, node *corev1.Node) (reconcil
 		}
 		if !clusterHealthy {
 			c.recorder.Publish(NodeRepairBlockedUnmanagedNodeClaim(node, nodeClaim, fmt.Sprintf("more then %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String()))...)
-			return reconcile.Result{}, nil
+			return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
```
jonathan-innis, line 152: Just checking: This is more of a fix than a perf benefit, right? We probably should have been requeueing before but we weren't and that meant that the cluster could have gone healthy after being unhealthy and we wouldn't have requeued a node again until after we got a heartbeat?

pkg/controllers/node/health/suite_test.go
```diff
@@ -312,9 +312,42 @@ var _ = Describe("Node Health", func() {
 
 			fakeClock.Step(60 * time.Minute)
 
-			// Determine to delete unhealthy nodes
+			// Determine if we should delete unhealthy nodes
 			for i := range 4 {
-				ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				result := ExpectObjectReconciled(ctx, env.Client, healthController, nodes[i])
+				if i < 3 {
```
jonathan-innis: Why do we have this check and also why are we iterating through the first 4 nodes rather than just the 3 that we marked unhealthy? Are we actually checking anything with the nodes that don't have an unhealthy status condition?

pkg/controllers/node/health/suite_test.go
```diff
@@ -312,9 +312,42 @@ var _ = Describe("Node Health", func() {
...
+				}
+				nodeClaim = ExpectExists(ctx, env.Client, nodeClaims[i])
+				Expect(nodeClaim.DeletionTimestamp).To(BeNil())
+			}
+		})
+		It("should ignore unhealthy nodes if more then 20% of the nodes are unhealthy in a cluster", func() {
+			ExpectApplied(ctx, env.Client, nodePool)
+			nodeClaims := []*v1.NodeClaim{}
+			nodes := []*corev1.Node{}
+			for i := range 10 {
```
jonathan-innis, line 334: Same question here -- I feel like the way that this is written is a bit odd -- is there a way to simplify some of this?

pkg/controllers/node/health/controller.go
engedaam: Yeah, it's safer to requeue here. It's possible that a node lease renewal would have triggered a reconciliation. However, during some failure scenarios node heartbeats are not guaranteed to occur.

pkg/controllers/node/health/controller.go
engedaam: Yeah, since we are not looking at the message or reason for the cloud provider policy that will not impact the repair action decision. The idea here is that it might be wroth while to reconcile on a change to any part of the condition. For now I will drop it

pkg/controllers/node/health/controller.go
engedaam: Yeah, since we are not looking at the message or reason for the cloud provider policy that will not impact the repair action decision. The idea here is that it might be wroth while to reconcile on a change to any part of the condition. For now I will drop it

pkg/controllers/node/health/suite_test.go
engedaam: Moved away to going to through all the nodes and nodeclaims. Now focusing only the changes we are expecting to see here

pkg/controllers/node/health/controller.go
```diff
@@ -73,7 +75,29 @@ func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
...
+				}
+
+				for _, oldCond := range oldNode.Status.Conditions {
+					newCond := nodeutils.GetCondition(newNode, oldCond.Type)
+					// Return true if any of these conditions are met:
+					// 1. Condition type no longer exists in new node
+					// 2. Transition time has changed
+					// 3. Status has changed
+					if newCond.Type == "" ||
+						oldCond.LastTransitionTime != newCond.LastTransitionTime ||
```
jonathan-innis, line 93: Just spot-checking: The heartbeat doesn't update the lastTransitionTime, right? That's just lastHearbeatTime?

pkg/controllers/node/health/controller.go
jonathan-innis: Technically, you aren't supposed to fire an error alongside a requeue result (controller-runtime will start firing warning logs at you when you do this) -- so, we should restructure this now so that we check an error first and then return the requeue result later if we don't have an error

pkg/controllers/node/health/suite_test.go
```diff
@@ -288,34 +288,74 @@ var _ = Describe("Node Health", func() {
...
+			nodeTwo := nodes[1]
+			nodeClaimTwo := nodeClaims[1]
+			result = ExpectObjectReconciled(ctx, env.Client, healthController, nodeTwo)
+			Expect(result.RequeueAfter).To(BeNumerically("~", time.Minute*5, time.Second))
+			nodeClaim = ExpectExists(ctx, env.Client, nodeClaimTwo)
+			Expect(nodeClaim.DeletionTimestamp).To(BeNil())
+		})
+		It("should ignore unhealthy nodes if more then 20% of the nodes are unhealthy in a cluster", func() {
+			nodeClaims, nodes := test.NodeClaimsAndNodes(10, v1.NodeClaim{ObjectMeta: metav1.ObjectMeta{Finalizers: []string{v1.TerminationFinalizer}}})
 			for i := range 4 {
```
jonathan-innis: nit: Just curious why you would choose 4 here and wouldn't want to just match what you did above with only 3 exceeding the 20% limit

pkg/controllers/node/health/controller.go
engedaam: Yeah, heartbeat will not update `lastTransitionTime` if there is not a change in the status condition, and only update `lastHearbeatTime`

pkg/controllers/node/health/suite_test.go
engedaam: Yeah, this would work as well. This just to validate that any over 20% would block it. This would have resulted in 40% of nodes being unhealthy. For the sake of uniformity, I've updated to have 30% of nodes being unhealthy

PR #2293, opened 2025-06-09
WIP: DO NOT MERGE perf: multithreaded orchestration queue

DerekFrank:
Fixes #N/A 

**Description**

This allows the orchestration queue to be multi threaded

Working on testing still :)


pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -132,10 +140,10 @@ func IsUnrecoverableError(err error) bool {
 }
 
 type Queue struct {
-	workqueue.TypedRateLimitingInterface[*Command]
+	sync.RWMutex
```
jmdeal: nit: we're not great about doing this everywhere, but generally we want to group the mutex with the resource it's protecting (in this case `nodeClaimToCommand`). I'd revert the ordering such that it's next to it.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -177,41 +184,32 @@ func NewCommand(replacements []string, candidates []*state.StateNode, id types.U
...
 	}
+	return cmd
 }
 
 func (q *Queue) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("disruption.queue").
-		WatchesRawSource(singleton.Source()).
-		Complete(singleton.AsReconciler(q))
+		WatchesRawSource(source.Channel(q.source, &handler.TypedEnqueueRequestForObject[*v1.NodeClaim]{})).
```
jmdeal: What's populating this source? I would assume we want to do that via the Add method, but it doesn't look like we are atm.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -177,41 +184,32 @@ func NewCommand(replacements []string, candidates []*state.StateNode, id types.U
...
+		WithOptions(controller.Options{
+			RateLimiter:             reasonable.RateLimiter(),
+			MaxConcurrentReconciles: 100,
+		}).
+		Complete(reconcile.AsReconciler(m.GetClient(), q))
 }
 
-func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
+func (q *Queue) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
```
jmdeal: Rather than reconciling on the NodeClaim, would it be possible to reconcile on the command itself? It might require an adapter on top of the base reconciler type (a la what's done by controller-runtime's [ObjectReconciler](https://github.com/kubernetes-sigs/controller-runtime/blob/6ad5c1dd4418489606d19dfb87bf38905b440561/pkg/reconcile/reconcile.go#L140-L142)), but IMO the extra scaffolding is worth the readability improvement.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -177,41 +184,32 @@ func NewCommand(replacements []string, candidates []*state.StateNode, id types.U
...
 	if err := q.waitOrTerminate(ctx, cmd); err != nil {
 		// If recoverable, re-queue and try again.
 		if !IsUnrecoverableError(err) {
 			// store the error that is causing us to fail, so we can bubble it up later if this times out.
 			cmd.lastError = err
-			// mark this item as done processing. This is necessary so that the RLI is able to add the item back in.
-			q.TypedRateLimitingInterface.Done(cmd)
-			q.TypedRateLimitingInterface.AddRateLimited(cmd)
-			return reconcile.Result{RequeueAfter: singleton.RequeueImmediately}, nil
+			return reconcile.Result{Requeue: true}, nil
```
jmdeal: Note that controller-runtime is not going to do exponential backoff for `Requeue: true` anymore. We need to solve this for the whole project, but we should keep that in mind if we end up merging the project wide change before this one.

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -307,32 +305,26 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
-	defer q.mu.RUnlock()
+func (q *Queue) HasAny(candidates ...*state.StateNode) bool {
+	q.RLock()
+	defer q.RUnlock()
 
 	// If the mapping has at least one of the candidates' providerIDs, return true.
-	_, ok := lo.Find(ids, func(id string) bool {
-		_, ok := q.providerIDToCommand[id]
+	_, ok := lo.Find(candidates, func(n *state.StateNode) bool {
+		_, ok := q.nodeClaimToCommand[client.ObjectKeyFromObject(n.NodeClaim)]
```
jmdeal: We need to iterate through all of the candidates in the command, right? This is probably pretty expensive, we'd probably still want to maintain the providerID map

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: Oh that's clever, I can certainly add that. Maybe in a new file under `pkg/utils/reconcile.go`?

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: Which part are you concerned about being expensive? The additional iterations (more candidates than providerIds), or the way in which we are validating the existence in the map?

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: Whoops!

pkg/controllers/disruption/orchestration/queue.go
jmdeal: I could see us moving it over to operatorpkg, but that seems appropriate in the interim.

pkg/controllers/disruption/orchestration/queue.go
jmdeal: IIUC the map doesn't have every candidate as a key - it just has a candidate which is chosen as the representative for the command. To check that any given candidate is in a command we could first check if it's a key in the map, but then we'd need to iterate through every nodeclaim in each command

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: Looks like we'll need to add a lot of scaffolding to make that happen. I think we can punt on that for now

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: Ahh I see what you mean. We can use providerId, I don't think that'll be an issue

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -307,53 +311,60 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
-	q.mu.Lock()
-	for _, candidate := range cmd.candidates {
-		q.providerIDToCommand[candidate.ProviderID()] = cmd
+	for _, c := range cmd.candidates {
+		q.nodeClaimsInQueueSet.Insert(c.NodeClaim.Status.ProviderID)
 	}
-	q.mu.Unlock()
-	q.TypedRateLimitingInterface.Add(cmd)
+	q.providerIDToCommand[cmd.KeyNodeClaim.Status.ProviderID] = cmd
+	q.source <- event.TypedGenericEvent[*v1.NodeClaim]{Object: cmd.KeyNodeClaim}
```
DerekFrank, line 326: This might make a deadlock condition? If the channel is full, this maintains the lock until the channel has an item taken off of it. If every single concurrent reconciler is trying to call Remove() on another command they won't be able to finish. Does controller runtime eventually time out a reconcile?

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -150,24 +162,24 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
...
-			}),
-		providerIDToCommand: map[string]*Command{},
-		kubeClient:          kubeClient,
-		recorder:            recorder,
-		cluster:             cluster,
-		clock:               clock,
-		provisioner:         provisioner,
+		source:               make(chan event.TypedGenericEvent[*v1.NodeClaim], 10000),
+		providerIDToCommand:  map[string]*Command{},
+		nodeClaimsInQueueSet: make(sets.Set[string]),
```
jonathan-innis: nit: style, more standard syntax
```suggestion
		nodeClaimsInQueueSet: sets.New[string](),
```

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -56,6 +61,10 @@ const (
 )
 
 type Command struct {
+	// We need a k8s object to represent each command
+	// This approach assumes that no nodeclaim will be in the disruption queue twice.
+	KeyNodeClaim *v1.NodeClaim
```
jonathan-innis: Do we need to store this in the command, if it's anyways just the first candidate?

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -181,24 +193,19 @@ func NewCommand(replacements []string, candidates []*state.StateNode, id types.U
 func (q *Queue) Register(_ context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("disruption.queue").
-		WatchesRawSource(singleton.Source()).
-		Complete(singleton.AsReconciler(q))
+		WatchesRawSource(source.Channel(q.source, &handler.TypedEnqueueRequestForObject[*v1.NodeClaim]{})).
+		WithOptions(controller.Options{
+			RateLimiter:             reasonable.RateLimiter(),
```
jonathan-innis: What's the max backoff that can happen here? Should we make this more similar to the eviction queue because I think that we should think about using a consistent poll and not have such a long maxBackoff just because we aren't hitting external APIs and are only hitting the apiserver

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -181,24 +193,19 @@ func NewCommand(replacements []string, candidates []*state.StateNode, id types.U
...
-		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
-	}
-
-	// Get command from queue. This waits until queue is non-empty.
-	cmd, shutdown := q.TypedRateLimitingInterface.Get()
-	if shutdown {
-		panic("unexpected failure, disruption queue has shut down")
+	cmd, exists := q.providerIDToCommand[nodeClaim.Status.ProviderID]
+	if !exists {
+		return reconcile.Result{}, fmt.Errorf("no command found for nodeclaim %s", nodeClaim.Name)
```
jonathan-innis: This shouldn't happen but this is going to get into a continuous loop if we return an error here -- seems fine to just throw out the command if this happens and say that this is a no-op if the command isn't properly stored

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -207,10 +214,7 @@ func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
 		if !IsUnrecoverableError(err) {
 			// store the error that is causing us to fail, so we can bubble it up later if this times out.
 			cmd.lastError = err
-			// mark this item as done processing. This is necessary so that the RLI is able to add the item back in.
-			q.TypedRateLimitingInterface.Done(cmd)
-			q.TypedRateLimitingInterface.AddRateLimited(cmd)
-			return reconcile.Result{RequeueAfter: singleton.RequeueImmediately}, nil
+			return reconcile.Result{Requeue: true}, nil
```
jonathan-innis: This is requeueing with backoff -- I almost think that we should try to requeue with a consistent poll because we are almost always going to hit the maxBackoff given that it's going to take around a minute for the nodes to come up

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -305,53 +309,57 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
 	}
 
+	q.Lock()
+	defer q.Unlock()
 	cmd.timeAdded = q.clock.Now()
-	q.mu.Lock()
-	for _, candidate := range cmd.candidates {
-		q.providerIDToCommand[candidate.ProviderID()] = cmd
+	for _, c := range cmd.candidates {
+		q.nodeClaimsInQueueSet.Insert(c.NodeClaim.Status.ProviderID)
```
jonathan-innis: I don't have a strong opinion, but it seems like you _could_ just keep the same paradigm as before -- I don't quite see the reason to change it up since the command is a pointer anyways

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -305,53 +309,57 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
+	q.providerIDToCommand[cmd.KeyNodeClaim.Status.ProviderID] = cmd
+	q.source <- event.TypedGenericEvent[*v1.NodeClaim]{Object: cmd.KeyNodeClaim}
 	return nil
 }
 
 // HasAny checks to see if the candidate is part of an currently executing command.
-func (q *Queue) HasAny(ids ...string) bool {
-	q.mu.RLock()
-	defer q.mu.RUnlock()
+func (q *Queue) HasAny(candidates ...*state.StateNode) bool {
```
jonathan-innis: I'm not sure if there's a reason to change this since I think we're only changing the representation of the item in the queue

pkg/controllers/disruption/orchestration/queue.go
```diff
@@ -305,53 +309,57 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
-		_, ok := q.providerIDToCommand[id]
-		return ok
+	_, ok := lo.Find(candidates, func(n *state.StateNode) bool {
+		return q.nodeClaimsInQueueSet.Has(n.ProviderID())
 	})
 	return ok
 }
 
+// Used for testing, gets the commands in the current queue
+func (q *Queue) GetCommands() []*Command {
```
jonathan-innis: To me: This is somewhat weird to me -- we don't have similar testing that tests what controller-runtime has in its queue and it feels like if we just keep our functional testing, we reduce the amount of changes that we need to make here and keep our testing more functional

DerekFrank: This is intended to keep the tests functional instead of relying on specific invariants in the disruption controller. Frequently tests will follow this pattern:

 1. Make a bunch of nodeclaims
 2. run the disruption controller
 3. run the orchestration queue
 4. see if the right nodeclaims got removed

Since ExpectObjectReconciled() needs to be called with the object that is in the controller runtime queue, we need to know which nodeclaim got put first. Since the test didn't create the command (the disruption controller did), the test does not know which nodeclaim was first in the list when the command got made. We can't call HasAny() because that will return true for any of the nodeclaims in the command. Instead, to rely on the function of the queue, we can instead just grab and run however many commands we think the disruption controller should have made and not have to do any weird nodeclaim lookup

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: Its a very minor detail, but I chose to change this as a minor refactor. I think it improves the abstraction of the orchestration queue but it doesn't matter much either way

pkg/controllers/disruption/orchestration/queue.go
DerekFrank: We don't. I thought this made for a more readable file as we are being explicit about the paradigm we are creating, but I can drop it if you'd prefer

jonathan-innis: Seems like dropping it would be easier -- we can just make a very explicit comment somewhere about it -- especially since we dropping having a command specifically for the orchestration queue now, it seems like an unnecessary detail

pkg/controllers/disruption/queue.go
```diff
@@ -113,34 +112,30 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
...
-		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
-	}
-
-	// Get command from queue. This waits until queue is non-empty.
-	cmd, shutdown := q.TypedRateLimitingInterface.Get()
-	if shutdown {
-		panic("unexpected failure, disruption queue has shut down")
+	cmd, exists := q.providerIDToCommand[nodeClaim.Status.ProviderID]
+	if !exists {
+		log.FromContext(ctx).Error(fmt.Errorf("no command found for nodeclaim %s", nodeClaim.Name), "")
```
jonathan-innis: NodeClaim is automatically injected into the error since it's the item we are using so you could probably just go with "no command found"

pkg/controllers/disruption/queue.go
```diff
@@ -299,7 +297,7 @@ func (q *Queue) StartCommand(ctx context.Context, cmd *Command) error {
 	markedCandidates, markDisruptedErr := q.markDisrupted(ctx, cmd)
 	// If we get a failure marking some nodes as disrupted, if we are launching replacements, we shouldn't continue
 	// with disrupting the candidates. If it's just a delete operation, we can proceed
-	if markDisruptedErr != nil && len(cmd.Replacements) > 0 {
+	if markDisruptedErr != nil && (len(cmd.Replacements) > 0 || len(markedCandidates) == 0) {
```
jonathan-innis, line 300: I added this because I realized with your change to use a NodeClaim as a key, this would be necessary

pkg/controllers/disruption/orchestration/queue.go
jonathan-innis: Discussed offline: This is fine to me so we don't take a dependency on the thing that we use as the "key" in the orchestration queue

pkg/controllers/disruption/queue.go
```diff
@@ -348,35 +349,40 @@ func (q *Queue) StartCommand(ctx context.Context, cmd *Command) error {
...
+		}
+	}
+	return false
+}
+
+func (q *Queue) GetCommands() []*Command {
+	q.RLock()
+	defer q.RUnlock()
+
+	return lo.UniqValues(q.providerIDToCommand)
```
jonathan-innis, line 371: Since this gives back pointers to commands, can't this technically have parallel access? Maybe not a big blocker since we are only using this for testing -- we should perhaps put a comment above this that implies that this is just testing?

pkg/controllers/disruption/queue.go
```diff
@@ -348,35 +349,40 @@ func (q *Queue) StartCommand(ctx context.Context, cmd *Command) error {
...
+	defer q.RUnlock()
 
 	// If the mapping has at least one of the candidates' providerIDs, return true.
-	_, ok := lo.Find(ids, func(id string) bool {
-		_, ok := q.providerIDToCommand[id]
-		return ok
-	})
-	return ok
+	for _, i := range ids {
+		if _, exists := q.providerIDToCommand[i]; exists {
```
jonathan-innis:
```suggestion
	for _, id := range ids {
		if _, exists := q.providerIDToCommand[id]; exists {
```

nit: only because `i` typically means the index so I personally find this a tiny bit confusing

jonathan-innis (approved):
/lgtm
/approve

PR #2296, opened 2025-06-10
feat: Add the Node Overlay CRD

engedaam:
Fixes #N/A 

**Description**
This PR introduces the NodeOverlay CRD, a new Karpenter API that enables fine-tuning of the scheduling simulation for advanced use cases.

### Features

- Implemented NodeOverlay CRD based on the [Node Overlay RFC](https://github.com/kubernetes-sigs/karpenter/pull/2166)
  - requirements: Constrain when the NodeOverlay is applied
  - priceAdjustment: Modify the simulated price of matching nodes
  - capacity: Add extended resources to matching nodes
- Added Common Expression Language (CEL) rules for CRD validation
- Included unit tests to validate CEL rules

### Example Usage

```yaml
apiVersion: karpenter.sh/v1alpha1
kind: NodeOverlay
metadata:
  name: adjustment
spec:
  requirements:
      operator: In
      values: ["spot"]
  priceAdjustment: "0.109"
  capacity:
     smarter-devices/fuse: 1
```

- `make presubmit`


pkg/apis/v1alpha1/nodelay_status.go
```diff
@@ -0,0 +1,50 @@
+/*
```
panicstevenson: Should this be `nodelay_status.go` or `nodeoverlay_status.go`?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+	// Requirements constrain when this NodeOverlay is applied during scheduling simulation.
+	// These requirements can match:
+	// - Well-known labels (e.g., node.kubernetes.io/instance-type, karpenter.sh/nodepool)
+	// - Custom labels from NodePool's spec.template.labels
+	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
+	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt' or 'Lt' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
+	// +kubebuilder:validation:MaxItems:=100
+	// +optional
+	Requirements []v1.NodeSelectorRequirement `json:"requirements,omitempty"`
+	// PriceAdjustment specifies a price changes for matching instance types. Accepts either:
```
jmdeal:
```suggestion
	// PriceAdjustment specifies the price change for matching instance types. Accepts either:
```

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+	// - A fixed price modifier (e.g., -0.5, 1.2)
```
jmdeal, line 41: Did we figure out how this is going to work with different currencies?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+	// +optional
+	PriceAdjustment *string `json:"priceAdjustment,omitempty"`
+	// Price specifies amount for an instance types that match the specified labels. Users can override prices using a signed float representing the price override
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +optional
+	Price *string `json:"price,omitempty"`
+	// Capacity that will add extended resources only, and not replace any existing resources on the nodes.
+	// These extended resources will be appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
+	// +kubebuilder:validation:XValidation:message="invalid resource restricted",rule="self.all(x, !(x in ['cpu', 'memory', 'pods']))"
```
jmdeal: It looks like ephemeral storage is missing. Also, do we expect cloudproviders to extend this via custom validation for their well-known resources (e.g. `nvidia.com/gpu`)?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+	// +optional
+	Capacity v1.ResourceList `json:"capacity,omitempty"`
+	// Weight is the priority given to the nodeoverlay while overriding node attributes. A higher
+	// numerical weight indicates that this nodeoverlay will be ordered
+	// ahead of other nodeoverlay with lower weights. A nodeoverlay with no weight
+	// will be treated as if it is a nodeoverlay with a weight of 0. Two nodeoverlays that have the same weight,
+	// we will merge them in alphabetical order.
```
jmdeal: nit: wording (consistent tense / capitalization)

> Weight defines the priority of this NodeOverlay when overriding node attributes. NodeOverlays with higher numerical weights take precedence over those with lower weights. If no weight is specified, the NodeOverlay is treated as having a weight of 0. When multiple NodeOverlays have identical weights, they are merged in alphabetical order.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+	// +kubebuilder:validation:Pattern=`^[+-](?:\d+(?:\.\d*)?|\.\d+)(?:%)?$`
+	// +optional
+	PriceAdjustment *string `json:"priceAdjustment,omitempty"`
+	// Price specifies amount for an instance types that match the specified labels. Users can override prices using a signed float representing the price override
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +optional
+	Price *string `json:"price,omitempty"`
+	// Capacity that will add extended resources only, and not replace any existing resources on the nodes.
+	// These extended resources will be appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
```
jmdeal: nit: wording (consistent tense / capitalization, and reworking the first sentence to flow a little better)

> Capacity adds extended resources to nodes without replacing any existing resources. These extended resources are appended to the node's existing resource list. Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+	// +kubebuilder:validation:XValidation:message="invalid resource restricted",rule="self.all(x, !(x in ['cpu', 'memory', 'pods']))"
+	// +optional
+	Capacity v1.ResourceList `json:"capacity,omitempty"`
+	// Weight is the priority given to the nodeoverlay while overriding node attributes. A higher
+	// numerical weight indicates that this nodeoverlay will be ordered
+	// ahead of other nodeoverlay with lower weights. A nodeoverlay with no weight
+	// will be treated as if it is a nodeoverlay with a weight of 0. Two nodeoverlays that have the same weight,
+	// we will merge them in alphabetical order.
+	// +kubebuilder:validation:Minimum:=1
```
jmdeal, line 60: Shouldn't the minimum be 0?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+type NodeOverlaySpec struct {
+	// Requirements constrain when this NodeOverlay is applied during scheduling simulation.
+	// These requirements can match:
+	// - Well-known labels (e.g., node.kubernetes.io/instance-type, karpenter.sh/nodepool)
+	// - Custom labels from NodePool's spec.template.labels
+	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
+	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt' or 'Lt' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
+	// +kubebuilder:validation:MaxItems:=100
+	// +optional
+	Requirements []v1.NodeSelectorRequirement `json:"requirements,omitempty"`
```
jmdeal, line 39: Should requirements be required? Or does omitting them implicitly mean we should select against all instance types? I'd lean towards requiring it so you need to at least be explicit (e.g. you could do something like this to  select everything)
```yaml
requirements:
  operator: Exists
```

pkg/apis/v1alpha1/nodeoverlay_validation.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// RuntimeValidate will be used to validate any part of the CRD that can not be validated at CRD creation
+func (in *NodeOverlay) RuntimeValidate(ctx context.Context) error {
+	return multierr.Combine(in.Spec.validateRequirements(ctx), in.Spec.validateCapacity())
+}
+
+// This function is used by the NodeClaim validation webhook to verify the nodepool requirements.
+// When this function is called, the nodepool's requirements do not include the requirements from labels.
+// NodeClaim requirements only support well known labels.
```
jmdeal: Looks like this comment has some copy-paste artifacts

pkg/apis/v1alpha1/nodeoverlay_validation.go
```diff
@@ -0,0 +1,132 @@
...
+func (in *NodeOverlaySpec) validateCapacity() (errs error) {
+	for n := range in.Capacity {
+		if v1.WellKnownResources.Has(n) {
+			errs = multierr.Append(errs, fmt.Errorf("invalid capacity: %s in resource, restricted", n))
+		}
+	}
+	return errs
+}
+
+func ValidateRequirement(ctx context.Context, requirement corev1.NodeSelectorRequirement) error { //nolint:gocyclo
```
jmdeal:
```suggestion
//nolint:gocyclo
func ValidateRequirement(ctx context.Context, requirement corev1.NodeSelectorRequirement) error { 
```

pkg/apis/v1alpha1/nodeoverlay_validation.go
jmdeal, line 1: Remind me - the reason we need to do this runtime validation is just because of CEL cost, right?

pkg/apis/v1alpha1/nodeoverlay_validation.go
```diff
@@ -0,0 +1,132 @@
...
+
+	// If there are only invalid values, set an error to transition the nodepool's readiness to false
+	if len(values) == 0 {
+		return fmt.Errorf("no valid values found in %v for %s, expected one of: %v, got: %v",
+			requirement.Values, requirement.Key, knownValues, invalidValues)
+	}
+
+	// If there are valid and invalid values, log the invalid values and proceed with valid values
+	if len(invalidValues) > 0 {
+		log.FromContext(ctx).Error(fmt.Errorf("invalid values found for key"), "please correct found invalid values, proceeding with valid values", "key", requirement.Key, "valid-values", values, "invalid-values", invalidValues)
```
jmdeal: nit: we can use a structured error here, same comment for the rest of the errors we create with associated values in the file.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,132 @@
...
+// OrderByWeight orders the NodeOverlays in the provided slice by their priority weight in-place. This priority evaluates
+// the following things in precedence order:
+//  1. NodeOverlays that have a larger weight are ordered first
+//  2. If two NodeOverlays have the same weight, then the NodePool with the name later in the alphabet will come first
+func (nol *NodeOverlayList) OrderByWeight() {
+	sort.Slice(nol.Items, func(a, b int) bool {
+		weightA := lo.FromPtr(nol.Items[a].Spec.Weight)
+		weightB := lo.FromPtr(nol.Items[b].Spec.Weight)
+		if weightA == weightB {
+			// Order NodePools by name for a consistent ordering when sorting equal weight
```
jmdeal:
```suggestion
			// Order NodeOverlays by name for a consistent ordering when sorting equal weight
```

jmdeal (commented):
checkpointing

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: Yeah, We will be using the public price that is sent back to Karpenter from the provider. Since Karpenter does not make any assumption for the currencies, the user will be able to define it based on that.

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: I was going back an forth on including ephemeral storage. I opted to have the cloud provider add that. I was thinking the cloud provider might not want to include ephemeral storage, but I can go either way if we feel that this is a common enough resource

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: I kept this at one to make it consistent with the NodePool API. We are effectively reserving zero wright as no weight

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: Yeah, this makes sense. To be inline with the NodePool requirements, we can make it required

pkg/apis/v1alpha1/nodeoverlay_validation.go
engedaam: Yeah, it was due to the cost

pkg/apis/v1alpha1/nodeoverlay.go
jmdeal: Based on [this](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) doc, it appears ephemeral storage is also considered a well-known resource. I believe, as a general rule, any resource without a domain (e.g. `nvidia.com/gpu`) would be a well-known resource. However, that includes hugepages so that concept doesn't map to Karpenter's concept of well known resources. 

I'm still not convinced we want to prevent customers from overriding well-known resources since there are usecases and I can't think of a drawback, but if we do want to I think it makes sense to include include ephemeral storage as a well-known resource. It's supported by both the AWS and Azure provider and is in common use. We should also make sure that we're consistent [here](https://github.com/kubernetes-sigs/karpenter/pull/2296/files#diff-ae13a9bbcddb80fed52372de837e250e0cec57957b6b58bc06c98130d0aed84fR96-R101) (ephemeral storage is currently listed)

pkg/apis/v1alpha1/doc.go
jmdeal, line 1: Any idea what these defaults libraries were originally for? I'm wondering if this was to implement an interface that could be used downstream? If we're going to drop them, I'd drop them in a separate PR.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	Price *string `json:"price,omitempty"`
+	// Capacity adds extended resources only, and not replace any existing resources.
+	// These extended resources are appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
+	// +kubebuilder:validation:XValidation:message="invalid resource restricted",rule="self.all(x, !(x in ['cpu', 'memory', 'ephemeral-storage', 'pods']))"
+	// +optional
+	Capacity v1.ResourceList `json:"capacity,omitempty"`
+	// Weight defines the priority of this NodeOverlay when overriding node attributes.
+	// NodeOverlays with higher numerical weights take precedence over those with lower weights.
+	//  If no weight is specified, the NodeOverlay is treated as having a weight of 0.
```
jmdeal:
```suggestion
	// If no weight is specified, the NodeOverlay is treated as having a weight of 0.
```

pkg/apis/v1alpha1/nodeoverlay.go
jmdeal: I think we need to have a rationale for which resources are or are not supported. I think right now it's the set intersection of well-known resources to Karpenter and well-known resources to Kubernetes. Here's an example of how I understand we intend this to work:
|Resource|Well-known to Karpenter|Well-known to k8s|Supported on NodeOverlay|
|-|-|-|-|
|`cpu`|x|x||
|`memory`|x|x||
|`ephemeral-storage`|x|x|?|
|`hugepages`||x|x|
|`nvidia.com/gpu`|x||x|

If we don't support ephemeral storage I think there's at least a consistent rule, but I also think that it's difficult to reason about since end-users don't necessarily know or care which resources are well known to Karpenter and k8s.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	PriceAdjustment *string `json:"priceAdjustment,omitempty"`
+	// Price specifies amount for an instance types that match the specified labels. Users can override prices using a signed float representing the price override
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +optional
+	Price *string `json:"price,omitempty"`
+	// Capacity adds extended resources only, and not replace any existing resources.
+	// These extended resources are appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
+	// +kubebuilder:validation:XValidation:message="invalid resource restricted",rule="self.all(x, !(x in ['cpu', 'memory', 'ephemeral-storage', 'pods']))"
+	// +optional
```
jmdeal: I'm going to hold off on additional comments here while we're still determining the story for which resources are or are not supported, but it looks like the first sentence from my original suggestion was dropped and doesn't flow well at the moment.

pkg/apis/v1alpha1/nodeoverlay_validation.go
jmdeal: I was referring to `serror`, not just a `fmt.Errorf`

pkg/apis/v1alpha1/nodeoverlay_validation.go
```diff
@@ -0,0 +1,132 @@
...
+	for n := range in.Capacity {
+		if v1.WellKnownResources.Has(n) {
+			errs = multierr.Append(errs, fmt.Errorf("invalid capacity: %s in resource, restricted", n))
+		}
+	}
+	return errs
+}
+
+//nolint:gocyclo
+func ValidateRequirement(ctx context.Context, requirement corev1.NodeSelectorRequirement) error {
```
jmdeal: Are we able to consolidate implementation and testing for requirements validation for nodeoverlays and nodeclaims? It looks like this is largely the same implementation.

pkg/apis/v1alpha1/nodeoverlay_validation.go
```diff
@@ -0,0 +1,132 @@
...
+func (in *NodeOverlaySpec) validateRequirements(ctx context.Context) (errs error) {
+	for _, requirement := range in.Requirements {
+		if err := ValidateRequirement(ctx, requirement); err != nil {
+			errs = multierr.Append(errs, fmt.Errorf("invalid value: %w in requirements, restricted", err))
+		}
+	}
+	return errs
+}
+
+func (in *NodeOverlaySpec) validateCapacity() (errs error) {
```
jmdeal, line 50: Is there any specific reason we're choosing to do this validation at both runtime and admission time?

pkg/apis/v1alpha1/nodeoverlay_validation_cel_test.go
```diff
@@ -0,0 +1,327 @@
...
+				Requirements: []corev1.NodeSelectorRequirement{
+					{
+						Key:      v1.CapacityTypeLabelKey,
+						Operator: corev1.NodeSelectorOpExists,
+					},
+				},
+			},
+		}
+	})
+	Context("Requirements", func() {
```
jmdeal: Are there large differences between the nodepool, nodeclaim, and nodeoverlay CEL validations for requirements? Like with the runtime validation, it would be good if we could collapse this to a single set of tests to prevent accidental skew in the future.

pkg/apis/v1alpha1/doc.go
rschalo: +1

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/samber/lo"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type NodeOverlaySpec struct {
+	// Requirements constrain when this NodeOverlay is applied during scheduling simulation.
```
rschalo:
```suggestion
	// Requirements constrain when this NodeOverlay is applied during scheduling simulations.
```

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	// These requirements can match:
+	// - Well-known labels (e.g., node.kubernetes.io/instance-type, karpenter.sh/nodepool)
+	// - Custom labels from NodePool's spec.template.labels
+	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
+	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt' or 'Lt' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
```
rschalo, line 36: Are these the same validations from other requirements?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	// Price specifies amount for an instance types that match the specified labels. Users can override prices using a signed float representing the price override
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +optional
+	Price *string `json:"price,omitempty"`
+	// Capacity adds extended resources only, and not replace any existing resources.
+	// These extended resources are appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
+	// +kubebuilder:validation:XValidation:message="invalid resource restricted",rule="self.all(x, !(x in ['cpu', 'memory', 'ephemeral-storage', 'pods']))"
+	// +optional
+	Capacity v1.ResourceList `json:"capacity,omitempty"`
```
rschalo, line 55: I'm not sure capacity appropriately covers this? Is this `CapacityOverrides`?

pkg/apis/v1alpha1/nodeoverlay.go
rschalo: Also - I'm not sure about splitting the validation of this into runtime validation and into CEL validation. What does it look like to perform runtime validation on `NodeOverlay`s and then post an update to their status?

pkg/apis/v1alpha1/nodeoverlay_validation.go
rschalo: Agree - let's go with runtime if CEL cost is too expensive.

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: > Are these the same validations from other requirements?

Yes, these are the same validation as other requirements except for minValues  

> Also - I'm not sure about splitting the validation of this into runtime validation and into CEL validation. What does it look like to perform runtime validation on NodeOverlays and then post an update to their status?

Ideally, we would not have runtime validation, but rather keep everything as CEL validation. This would give users quicker feedback loop, they know any error at apply time. However, there are some validation around the requirements keys that are to expensive for CEL validation. I would favor having this split validation model, so that we are able to provide users with quicker feedback at apply time and then relay on runtime validation for any other rules that need to be enforced.

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: > I'm not sure capacity appropriately covers this?

The resource here will map directly to capacity notion on nodes directly and not necessarily allocatable. To my I still think this is the best naming. Can you share you feel that this might not be the best name for it?


>  Is this CapacityOverrides?

These are not really overrides, as users are adding resources that are not on the nodes. I hesitate to insinuate any notion of overrides for capacity just because we have explicit validation against that

pkg/apis/v1alpha1/nodeoverlay_validation.go
engedaam: We are able to do CEL validation for capacity. However, for Karpenter well known resources I wanted runtime validation and CEL validation as a backstop for cloud providers including the resource  in the binary but not as a CEL rule.

pkg/apis/v1alpha1/doc.go
engedaam: These were interfaces to satisfy the knative  interface for conversion webhook. These are no longer need and should not be used downstream

engedaam: Dropped these for this PR

pkg/apis/v1alpha1/nodeoverlay_validation.go
engedaam: I'm not sure I follow, but I dropped all of these section anyway

pkg/apis/v1alpha1/nodeoverlay_validation.go
engedaam: The main reason why had split validation was due to the fact that NodeOverlay version of the requirements did not have MinValues. I do agree  that we can consolidation the validation, however I do want to keep the testing in place since this will check against the CEL validation rules that in both APIs that need to defined separately

pkg/apis/v1alpha1/nodeoverlay_validation_cel_test.go
engedaam: The test are close related. However, the CEL rules all need to be tested separately since they are defined in their own CRDs. Included Runtime validation to help catch if their are any divergence between runtime and CEL validation. There is no path for us to drop these tests.

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: Karpenter defines a specific set of well-known resources: CPU, Memory, ephemeral-storage, and pods. These are the core resources that all cloud providers must identify and manage. While cloud providers can discover and support additional resources (like GPUs), it's their responsibility to add validation for these resources, similar to how well-known labels are handled. Currently, the Node Overlay capacity field only permits the addition of resources that are not well known to Karpenter.

After further offline discussions, we've decided to maintain the original intent of the capacity field, while clearly documenting that only non-KarpenterWellKnownResources may be specified. This restriction ensures consistent resource management across all cloud providers while allowing flexibility for custom resource types.

pkg/apis/v1/labels.go
```diff
@@ -91,6 +91,15 @@ var (
 		v1.LabelWindowsBuild,
 	)
 
+	// WellKnownResources are resources that are known expected from the instance types
+	// provided by the cloud providers.
+	KarpenterWellKnownResources = sets.New[v1.ResourceName](
```
jonathan-innis:
```suggestion
	WellKnownResources = sets.New[v1.ResourceName](
```

nit: To have parity with the other naming

pkg/apis/v1/labels.go
```diff
@@ -91,6 +91,15 @@ var (
 		v1.LabelWindowsBuild,
 	)
 
+	// WellKnownResources are resources that are known expected from the instance types
+	// provided by the cloud providers.
```
jonathan-innis:
```suggestion
	// WellKnownResources are resources that are expected from the instance types
	// provided by cloud providers.
```

pkg/apis/v1alpha1/labels.go
```diff
@@ -0,0 +1,24 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import "sigs.k8s.io/karpenter/pkg/apis"
+
+const (
+	PriceOverlayAppliedAnnotationKey    = apis.Group + "/price-overlay-applied"
```
jonathan-innis, line 22: What's the intended use of these annotations? Seems like we are just going to applying these to instance types pretty heavily so do we expect to really need to mark NodeClaims with annotations in this way?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	// +optional
+	PriceAdjustment *string `json:"priceAdjustment,omitempty"`
+	// Price specifies amount for an instance types that match the specified labels. Users can override prices using a signed float representing the price override
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +optional
+	Price *string `json:"price,omitempty"`
+	// Capacity adds extended resources only, and not replace any existing resources.
+	// These extended resources are appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
+	// +kubebuilder:validation:XValidation:message="invalid resource restricted",rule="self.all(x, !(x in ['cpu', 'memory', 'ephemeral-storage', 'pods']))"
```
jonathan-innis, line 53: How is this going to work if we don't support overriding memory but we are trying to support huge-pages on this first iteration? If hugepages requires that you pre-allocate the pages, this is going to take away from the reported memory that we see in the allocatable block, so without allowing the override of memory, I'm not sure how you can tangibly support this new resource

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	// +optional
+	Capacity v1.ResourceList `json:"capacity,omitempty"`
+	// Weight defines the priority of this NodeOverlay when overriding node attributes.
+	// NodeOverlays with higher numerical weights take precedence over those with lower weights.
+	// If no weight is specified, the NodeOverlay is treated as having a weight of 0.
+	// When multiple NodeOverlays have identical weights, they are merged in alphabetical order.
+	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Maximum:=100
```
jonathan-innis: I think we want to be more flexible than blocking at 100 -- perhaps we go to something like 1000 but if we start seeing a user create a lot of different NodeOverlays with a lot of different weights, it might be nice to have the flexibility to define a higher value -- 1000 or 10,000 sounds better to me. Honestly, we should think about bumping the NodePool weight value but we haven't really had a strong ask for that yet.

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	// +optional
+	Weight *int32 `json:"weight,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:storageversion
+// +kubebuilder:resource:path=nodeoverlays,scope=Cluster,categories=karpenter
```
jonathan-innis: Should something like `overlays` be a shortname?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type NodeOverlaySpec struct {
+	// Requirements constrain when this NodeOverlay is applied during scheduling simulations.
+	// These requirements can match:
+	// - Well-known labels (e.g., node.kubernetes.io/instance-type, karpenter.sh/nodepool)
+	// - Custom labels from NodePool's spec.template.labels
+	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
```
jonathan-innis, line 35: I'm realizing that we don't have a similar constraint for `NotIn` -- should we?

pkg/apis/v1alpha1/nodeoverlay_validation.go
jonathan-innis: We should talk to apimachinery in some WG meetings to see if we can slowly get rid of this runtime validation -- for some of our calculations, I think we can reason about them being cheaper to calculate and I think it might be worthwhile trying to see if there is a path here

pkg/apis/v1alpha1/nodeoverlay_validation_cel_test.go
```diff
@@ -0,0 +1,327 @@
...
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	. "sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
+)
+
+var _ = Describe("CEL/Validation", func() {
+	var nodeOverlay *NodeOverlay
+
+	BeforeEach(func() {
+		if env.Version.Minor() < 25 {
```
jonathan-innis: We're off of pre-1.25 versions and, in general, I'd consider these out of support now so I think we can remove this block from our CEL testing

pkg/apis/v1alpha1/nodeoverlay_validation_cel_test.go
jonathan-innis: I don't think the implication is that we would drop these tests, but perhaps there's a common way to build these tests or a way to group the tests together so it's more likely that we make the changes all at once rather than forgetting to update them like @jmdeal is suggesting -- not something we need to necessarily cover in this PR but perhaps something to track and follow-up on?

pkg/apis/v1alpha1/nodeoverlay_default_test.go
```diff
@@ -0,0 +1,60 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
+)
+
+var _ = Describe("CEL/Default", func() {
```
jonathan-innis: Do we even need this test? We don't have any defaults to set, right? So what are we validating here?

pkg/apis/v1alpha1/suite_test.go
```diff
@@ -0,0 +1,165 @@
+/*
```
jonathan-innis, line 1: We don't have any runtime validation testing?

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+		return lo.Must(strconv.ParseFloat(lo.FromPtr(in.Spec.Price), 64))
+	}
+
+	// Check if adjustment is a percentage
+	isPercentage := strings.HasSuffix(lo.FromPtr(in.Spec.PriceAdjustment), "%")
+	adjustment := lo.FromPtr(in.Spec.PriceAdjustment)
+
+	// Remove the percentage sign if present
+	if isPercentage {
+		adjustment = strings.TrimSuffix(lo.FromPtr(in.Spec.PriceAdjustment), "%")
```
jonathan-innis, line 121: Do we need to apply some validation that we don't allow values below `-100%` since that won't make any sense

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// Parse the adjustment value
+	// Due to the CEL validation we can assume that
+	// there will always be a valid float provided into the spec
+	adjValue := lo.Must(strconv.ParseFloat(adjustment, 64))
+	adjustedPrice := lo.Ternary(isPercentage, instanceTypePrice*(1+(adjValue/100)), instanceTypePrice+adjValue)
```
jonathan-innis: nit: It might be nice to just handle all of the percentage-based logic inside of the conditional on L119

pkg/apis/v1alpha1/labels.go
engedaam: This is the mechanism that I think will be useful and cheep way for customer to have visibility that Node Overlay is applied to their resources. This is also a way that will be used to inform the cloud provider as a way for them to make decision if the launch strategy needs to change if an overlay is applied.

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: I was thinking that we could do this increase later, since it's not a breaking change to increase the limit, however   this is an alpha it okay for us to reduce this at a later time anyways

pkg/apis/v1alpha1/nodeoverlay_default_test.go
engedaam: The idea here is that we want to make sure that we don't set default values. Today it is indented for us not to have default values

pkg/apis/v1alpha1/suite_test.go
engedaam: No, we have runtime validtion testing. It's grouped together with the CEL validation tests: pkg/apis/v1alpha1/nodeoverlay_validation_cel_test.go. It make sense to make it here since we can vaildate the runs are consistent for both runtime and CEL

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: Yeah, I think it make sense to do add it in for NotIn for this API. However, we can't make this change on the NodePool and NodeClaim since it would be breaking

pkg/apis/v1alpha1/nodeoverlay.go
engedaam: Discussed offline, since HugePages is a well know resource to k8s and spacial in that allocatable memory will be consumed. When customers configure node overlay for HugePage support, Karpenter will simulate scheduling by account for the memory usage consumed by HugePages.

pkg/apis/v1alpha1/nodeoverlay_validation_cel_test.go
```diff
@@ -0,0 +1,274 @@
+/*
```
jonathan-innis: nit: So, should we call this "CEL test" or is this just "validation_test"?

pkg/apis/v1alpha1/nodeoverlay_default_test.go
jonathan-innis: IMO: This feels unneccesary -- just drop it

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,135 @@
...
+	// - A fixed price modifier (e.g., -0.5, 1.2)
+	// - A percentage modifier (e.g., +10% for increase, -15% for decrees)
+	// +kubebuilder:validation:Pattern=`^(([+-]{1}(\d*\.?\d+))|(\+{1}\d*\.?\d+%)|(^(-\d{1,2}(\.\d+)?%)$)|(-100%))$`
+	// +optional
+	PriceAdjustment *string `json:"priceAdjustment,omitempty"`
+	// Price specifies amount for an instance types that match the specified labels. Users can override prices using a signed float representing the price override
+	// +kubebuilder:validation:Pattern=`^\d+(\.\d+)?$`
+	// +optional
+	Price *string `json:"price,omitempty"`
+	// Capacity adds extended resources only, and not replace any existing resources.
```
jonathan-innis:
```suggestion
	// Capacity adds extended resources only, and does not replace any existing resources.
```

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -0,0 +1,135 @@
...
+	// These extended resources are appended to the node's existing resource list.
+	// Note: This field does not modify or override standard resources like CPU, memory, ephemeral-storage, or pods.
```
jonathan-innis:
```suggestion
	// Note: This field does not modify or override standard resources like cpu, memory, ephemeral-storage, or pods.
```

nit

PR #2297, opened 2025-06-11
feat: Implement Node Ready Timeout Recovery feature

matanbaruch:
This commit introduces a new feature that automatically recovers nodes that fail to reach a ready state within a configurable timeout period. The feature includes a new controller, `NodeReadyTimeoutController`, which monitors node readiness and triggers recovery actions when necessary.

This feature complements existing node repair mechanisms by addressing a different failure mode, ensuring efficient resource utilization in the cluster.

**Description**
- Added a feature gate `NodeReadyTimeoutRecovery` to enable this functionality.
- Configurable timeout duration via the `NODE_READY_TIMEOUT` environment variable, defaulting to 10 minutes.
- Updated Karpenter's settings and deployment configurations to support the new feature.
- Introduced metrics and events for tracking recovery actions.


designs/node-ready-timeout-recovery.md
```diff
@@ -0,0 +1,171 @@
...
+
+## Problem Statement
+
+Nodes can sometimes fail to reach a ready state due to various issues such as:
+- Network connectivity problems during bootstrap
+- Issues with cloud provider instance initialization
+- Problems with container runtime or kubelet startup
+- Resource constraints during node initialization
+
+Currently, Karpenter has a node repair feature that handles nodes with specific unhealthy conditions defined by cloud providers. However, there's no mechanism to automatically recover nodes that simply fail to become ready within a reasonable timeframe. These nodes can remain in a NotReady state indefinitely, consuming cloud resources without contributing to cluster capacity.
```
engedaam, line 11: I'd like to understand how the current node repair implementation doesn't address this issue. The existing mechanism seems to provide coverage through two key components:

1. The cloud provider (e.g., AWS) actively monitors the Node Ready status condition
Karpenter's node repair feature automatically repairs nodes that remain in NotReady or Unknown states for 30 minutes, including nodes that have never reached a Ready state
2. I acknowledge that the 30-minute repair threshold is currently hardcoded. While we previously considered making this duration configurable during the initial implementation, we decided to defer that feature for a later release.

Could you clarify what specific aspects of this problem aren't addressed by the current node repair functionality? Are there particular scenarios or edge cases that the existing implementation doesn't handle effectively, given that it also covers nodes that never achieve a Ready state?

designs/node-ready-timeout-recovery.md
```diff
@@ -0,0 +1,171 @@
...
+2. Track the time since node creation
+3. Check if nodes are in Ready state
+4. Trigger node replacement for nodes that exceed the timeout while remaining NotReady
+
+### Recovery Logic
+
+1. **Target Nodes**: Only nodes with `karpenter.sh/nodepool` label (Karpenter-managed)
+2. **Grace Period**: Wait for the configured timeout period from node creation time
+3. **Ready Check**: Verify the node's Ready condition status
+4. **Recovery Action**: Delete the NodeClaim to trigger replacement via standard Karpenter mechanisms
```
engedaam, line 65: Are there case we needed to think about around pod eviction? Do think it would be worth while to have a some way   for graceful delation of pods that start running on these nodes? I'm mainly thinking about pods that tolerate the node ready taints

designs/node-ready-timeout-recovery.md
```diff
@@ -0,0 +1,171 @@
...
+
+### Integration with Existing Features
+
+- **Node Repair**: This feature complements existing node repair by handling a different failure mode
+- **Disruption Controls**: Will respect existing disruption budgets and controls
+- **Health Checks**: Works alongside cloud provider health policies
```
engedaam, line 71: How so? Is the cloud provider going to be giving us back health information for each node outside of the status condition?

designs/node-ready-timeout-recovery.md
```diff
@@ -0,0 +1,171 @@
...
+    NodeReadyTimeoutRecovery bool // New feature gate
+}
+```
+
+### Configuration
+
+The timeout duration will be configurable via environment variable with a reasonable default:
+
+- **Environment Variable**: `NODE_READY_TIMEOUT`
+- **Default Value**: `10m` (10 minutes)
```
engedaam, line 48: Why do you think 10 minutes is a reasonable timeout? We have seen some cases where nodes take long then 15 minutes to get registered and go ready with Karpenter. This is mainly for nodes that have extended resources and require drivers and bootstrap scripts the need to be run on the node on startup

designs/node-ready-timeout-recovery.md
```diff
@@ -0,0 +1,171 @@
...
+    SpotToSpotConsolidation bool
+    NodeReadyTimeoutRecovery bool // New feature gate
+}
+```
+
+### Configuration
+
+The timeout duration will be configurable via environment variable with a reasonable default:
+
+- **Environment Variable**: `NODE_READY_TIMEOUT`
```
engedaam, line 47: Can you give some detail on why you think this would better to have global configuration vs on the nodepools?

pkg/controllers/node/readytimeout/controller.go
```diff
@@ -0,0 +1,211 @@
...
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/events"
+	"sigs.k8s.io/karpenter/pkg/metrics"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
+	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
+)
+
+var allowedUnhealthyPercent = intstr.FromString("20%")
```
engedaam, line 48: Why are you think we should limit to only disrupting only 20% of the nodes?

pkg/controllers/node/readytimeout/controller.go
```diff
@@ -0,0 +1,211 @@
...
+// Controller for the resource
+type Controller struct {
+	clock         clock.Clock
+	recorder      events.Recorder
+	kubeClient    client.Client
+	cloudProvider cloudprovider.CloudProvider
+}
+
+// NewController constructs a controller instance
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, clock clock.Clock, recorder events.Recorder) *Controller {
```
engedaam, line 59: This controller seems like it's really close related to the node repair controller. Any thoughts on combining the controllers and making conditional based on the feature gates enabled?

pkg/controllers/node/readytimeout/controller_test.go
```diff
@@ -0,0 +1,238 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/events"
+	"sigs.k8s.io/karpenter/pkg/test"
+)
+
+// Test the core timeout logic without requiring full k8s infrastructure
+func TestTimeoutLogic(t *testing.T) {
```
engedaam, line 34: As a stylistic and consistency sake, this repo uses ginkgo and gomega as the testing framework. Would mind updating these to use that? You can use this as a prior art: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/health/suite_test.go

pkg/controllers/node/readytimeout/events.go
```diff
@@ -0,0 +1,57 @@
...
+	return events.Event{
+		InvolvedObject: node,
+		Type:           corev1.EventTypeWarning,
+		Reason:         "NodeReadyTimeoutRecovery",
+		Message:        fmt.Sprintf("Node failed to become ready within %v (node age: %v), triggering recovery", nodeAge, nodeAge),
+		DedupeValues:   []string{node.Name},
+	}
+}
+
+func NodeReadyTimeoutRecoveryBlocked(node *corev1.Node, nodeClaim *v1.NodeClaim, nodePool *v1.NodePool, message string) events.Event {
```
engedaam, line 39: What are cases where recovery is blocked?

pkg/controllers/node/readytimeout/controller.go
```diff
@@ -0,0 +1,211 @@
...
+		if !nodePoolHealthy {
+			return reconcile.Result{}, c.publishNodePoolHealthEvent(ctx, node, nodeClaim, nodePoolName)
+		}
+	} else {
+		clusterHealthy, err := c.isClusterHealthy(ctx)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if !clusterHealthy {
+			c.recorder.Publish(NodeReadyTimeoutRecoveryBlockedUnmanagedNodeClaim(node, nodeClaim, fmt.Sprintf("more than %s nodes are unhealthy in the cluster", allowedUnhealthyPercent.String())))
```
engedaam, line 124: Not sure I understand this event here.  Would ever get to think point for nodes that Karpenter does not own?

pkg/controllers/node/readytimeout/events.go
```diff
@@ -0,0 +1,57 @@
...
+	return events.Event{
+		InvolvedObject: nodePool,
+		Type:           corev1.EventTypeWarning,
+		Reason:         "NodeReadyTimeoutRecoveryBlocked",
+		Message:        message,
+		DedupeValues:   []string{nodePool.Name},
+	}
+}
+
+func NodeReadyTimeoutRecoveryBlockedUnmanagedNodeClaim(node *corev1.Node, nodeClaim *v1.NodeClaim, message string) events.Event {
```
engedaam, line 49: I'm not sure it makes sense to fire events for nodes that Karpenter does not own. This might be very noise in the cluster

pkg/controllers/node/readytimeout/metrics.go
```diff
@@ -0,0 +1,39 @@
...
+	crmetrics.Registry,
+	prometheus.CounterOpts{
+		Namespace: metrics.Namespace,
+		Subsystem: metrics.NodeClaimSubsystem,
+		Name:      "ready_timeout_recovered_total",
+		Help:      "Number of nodeclaims recovered due to ready timeout by Karpenter. Labeled by the owning nodepool and capacity type.",
+	},
+	[]string{
+		metrics.NodePoolLabel,
+		metrics.CapacityTypeLabel,
```
engedaam, line 37: Why do you think the capacity type label would relevant to users in this context?

designs/node-ready-timeout-recovery.md
matanbaruch: I wasn't familiar with node auto repair. And Yes the 30-min repair threshold is needed to be configurable per node pool.

matanbaruch:
Thanks for a detailed PR review. 
I will close this PR and open another one to adjust per node pool node repair
https://karpenter.sh/docs/concepts/disruption/#node-auto-repair

Issue #2078, opened 2025-03-17
Support `whenMinValuesUnsatisfiable` to handle flexibility as a preference


jonathan-innis:
### Description

**What problem are you trying to solve?**

I've been talking with some folks about the use of `minValues` when it comes to selecting instance types and increasing the flexibility when Karpenter launches with spot. The core problem here is, without more context, Karpenter can overpack your pods onto the largest instance type size, reducing the flexibility of your spot instance launch, and causing you to continually be launched in a really bad instance type pool.

`minValues` is a great solve for this because it forces the scheduler to enforce a minimum "flexibility" on a requirement; however, the current configuration enforces this as a hard requirement. This means that anyone who's looking to implement this flexibility is forced to create a fallback NodePool to ensure that they don't get put into a situation where they are unable to schedule pods when capacity is constrained for the instance types in their NodePool.

One proposed solution to this problem is to create a `whenMinValuesUnsatisfiable` that has the values `DoNotSchedule`, `ScheduleAnyways`, and `UseUnavailable`. The values would have the following defintiions

* `DoNotSchedule`: This is a hard requirement, it requires that the minValues in the requirement is satisfied by available offerings
* `ScheduleAnyways`: This is a soft requirement, it attempts to satisfy the minValues requirements with available offerings. If it can’t satisfy it, then it will attempt to support as many values as possible for the requirement, but it will still allow the NodeClaim to launch without satisfying the requirement
* `UseUnavailable`: This is a soft requirement, it attempts to satisfy the minValues requirements with available offerings. If it can’t satisfy it, it attempts to satisfy the minValues with unavailable offerings. If it can’t satisfy the minValues with unavailable offerings, the NodeClaim will fail to schedule and we will attempt other NodePools

**How important is this feature to you?**

This would reduce the configuration burden for folks who are using the `minValues` feature to increase the flexibility of their NodePools.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

PR #2299, opened 2025-06-11
feat: support auto relaxing min values

saurav-agarwalla:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/2078  

**Description**

Support auto-relaxing min values when using min values would lead to a scheduling error.

* Unit testing 
* This repo doesn't have any testing around scheduling and min values. To ensure that there's no regression, I ran E2E tests locally with `MinValuesPolicy` set to `BestEffort` and without `MinValuesPolicy` being set . 
* I am going to follow up with the team to see where we're with porting over the scheduling tests and accordingly add tests around min values.

### DIFF
--- a/pkg/controllers/provisioning/scheduling/nodeclaim.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaim.go
@@ -146,8 +146,8 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 	remaining, unsatisfiableKeys, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, podData.Requests, n.daemonResources, requests, relaxMinValues)
 	if relaxMinValues {
 		// Update min values on the requirements if they are relaxed
-		for _, e := range lo.Entries(unsatisfiableKeys) {
-			nodeClaimRequirements.Get(e.Key).MinValues = lo.ToPtr(e.Value)
+		for key, minValues := range unsatisfiableKeys {
+			nodeClaimRequirements.Get(key).MinValues = lo.ToPtr(minValues)
 		}
 	}
 	if err != nil {
--- a/pkg/operator/options/suite_test.go
+++ b/pkg/operator/options/suite_test.go
@@ -58,6 +58,8 @@ var _ = Describe("Options", func() {
 		"LOG_ERROR_OUTPUT_PATHS",
 		"BATCH_MAX_DURATION",
 		"BATCH_IDLE_DURATION",
+		"PREFERENCE_POLICY",
+		"MIN_VALUES_POLICY",
 		"FEATURE_GATES",
 	}
 
@@ -324,6 +326,8 @@ func expectOptionsMatch(optsA, optsB *options.Options) {
 	Expect(optsA.LogErrorOutputPaths).To(Equal(optsB.LogErrorOutputPaths))
 	Expect(optsA.BatchMaxDuration).To(Equal(optsB.BatchMaxDuration))
 	Expect(optsA.BatchIdleDuration).To(Equal(optsB.BatchIdleDuration))
+	Expect(optsA.PreferencePolicy).To(Equal(optsB.PreferencePolicy))
+	Expect(optsA.MinValuesPolicy).To(Equal(optsB.MinValuesPolicy))
 	Expect(optsA.FeatureGates.ReservedCapacity).To(Equal(optsB.FeatureGates.ReservedCapacity))
 	Expect(optsA.FeatureGates.NodeRepair).To(Equal(optsB.FeatureGates.NodeRepair))
 	Expect(optsA.FeatureGates.SpotToSpotConsolidation).To(Equal(optsB.FeatureGates.SpotToSpotConsolidation))


pkg/cloudprovider/types.go
```diff
@@ -224,11 +225,12 @@ func (its InstanceTypes) SatisfiesMinValues(requirements scheduling.Requirements
...
 // Truncate truncates the InstanceTypes based on the passed-in requirements
 // It returns an error if it isn't possible to truncate the instance types on maxItems without violating minValues
-func (its InstanceTypes) Truncate(requirements scheduling.Requirements, maxItems int) (InstanceTypes, error) {
+func (its InstanceTypes) Truncate(ctx context.Context, requirements scheduling.Requirements, maxItems int) (InstanceTypes, error) {
 	truncatedInstanceTypes := lo.Slice(its.OrderByPrice(requirements), 0, maxItems)
 	// Only check for a validity of NodeClaim if its requirement has minValues in it.
 	if requirements.HasMinValues() {
-		if _, err := truncatedInstanceTypes.SatisfiesMinValues(requirements); err != nil {
+		// If minValues is NOT met for any of the requirement across InstanceTypes, then still allow it if fallback is enabled.
+		if _, err := truncatedInstanceTypes.SatisfiesMinValues(requirements); err != nil && !options.FromContext(ctx).FeatureGates.AutoRelaxMinValues {
```
jmdeal: If we've enabled `AutoRelaxMinValues`, is there any reason to check if we've satisfied min values or do we want to short circuit here? It's a small change, but could have meaningful performance impact since this is in a hot-path.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -111,49 +111,49 @@ func NewNodeClaim(
 // CanAdd returns whether the pod can be added to the NodeClaim
 // based on the taints/tolerations, host port compatibility,
 // requirements, resources, reserved capacity reservations, and topology requirements
-func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodData) (updatedRequirements scheduling.Requirements, updatedInstanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering, err error) {
+func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodData, autoRelaxMinValues bool) (updatedRequirements scheduling.Requirements, updatedInstanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering, minValuesAutoRelaxed bool, err error) {
```
jmdeal: nit: any reason we're passing this in instead of extracting the feature gate value from the context?

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -364,9 +364,10 @@ func (e InstanceTypeFilterError) Error() string {
 }
 
 //nolint:gocyclo
-func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceType, requirements scheduling.Requirements, podRequests, daemonRequests, totalRequests corev1.ResourceList) (cloudprovider.InstanceTypes, error) {
+func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceType, requirements scheduling.Requirements, podRequests, daemonRequests, totalRequests corev1.ResourceList, autoRelaxMinValues bool) (cloudprovider.InstanceTypes, bool, error) {
```
jmdeal: nit: any reason to pass the flag directly here rather than passing the context and parsing out out of that?

pkg/test/options.go
```diff
@@ -52,6 +52,7 @@ type FeatureGates struct {
 	NodeRepair              *bool
 	ReservedCapacity        *bool
 	SpotToSpotConsolidation *bool
+	AutoRelaxMinValues      *bool
```
jonathan-innis: A feature flag like this implies to me that we are going to go through a promotion cycle for this feature, but I'm not 100% sure that we are going to go through a promotion cycle to promote this to be the default behavior at all times. To me, this is just a setting that gives you a change of behavior so it would better if it was its own environment variable or even changing the name to something like `MinValuesPolicy` or `WhenMinValuesUnsatsfiable` might give you more flexibility in the future to extend the enum

pkg/metrics/metrics.go
```diff
@@ -44,6 +44,18 @@ var (
 			CapacityTypeLabel,
 		},
 	)
+	NodeClaimsCreatedWithMinValuesAutoRelaxedTotal = opmetrics.NewPrometheusCounter(
```
jonathan-innis: Is there a way to make this metric more generic? Like something about preference relaxation? This is super super specific to this change and I don't want to explode the number of metrics that we have as we start adding more config and more features

pkg/apis/v1/nodeclaim.go
```diff
@@ -37,7 +37,6 @@ type NodeClaimSpec struct {
 	// Requirements are layered with GetLabels and applied to every node.
 	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
 	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt' or 'Lt' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
-	// +kubebuilder:validation:XValidation:message="requirements with 'minValues' must have at least that many values specified in the 'values' field",rule="self.all(x, (x.operator == 'In' && has(x.minValues)) ? x.values.size() >= x.minValues : true)"
```
jonathan-innis: I'm not sure that we want to drop this requirement. It feels like if we are relaxing the minValues, we should also change the minValues requirement that we have associated with that NodeClaim -- that way we still maintain the requirement being passed onto the NodeClaim AND we don't have to remove the validation that we added here in CEL

pkg/apis/v1/labels.go
```diff
@@ -53,6 +53,7 @@ const (
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey           = apis.Group + "/nodepool-hash-version"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
+	NodeClaimMinValuesAutoRelaxedAnnotationKey = apis.Group + "/nodeclaim-min-values-auto-relaxed"
```
jonathan-innis: I'm curious why we need this annotation? You can't query on annotations so I'm curious how we expect this being added to the NodeClaim to be helpful

pkg/controllers/provisioning/suite_test.go
```diff
@@ -2583,6 +2584,447 @@ var _ = Describe("Provisioning", func() {
 			})
 		})
 	})
+
+	Context("AutoRelaxMinValues", func() {
```
jonathan-innis: Do we have consolidation testing in place for this as well? I'd want to validate scenarios like:
- When I have a NodeClaim that is fulfilling the NodePool's minValues without relaxation, I shouldn't consolidate to a cheaper node if consolidating to a cheaper node implied that I would relax the minValues

jonathan-innis: Note: I haven't looked deeply over these tests -- going to iron-out the core logic and then look more scrupulously over the testing

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -44,6 +44,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/events"
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	karpopts "sigs.k8s.io/karpenter/pkg/operator/options"
```
jonathan-innis, line 47: Did we need to alias this? What other package up here is named `options`?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -581,6 +589,11 @@ func (s *Scheduler) addToNewNodeClaim(ctx context.Context, pod *corev1.Pod) erro
 		if i >= idx {
 			return false
 		}
+
+		if minValuesAutoRelaxed {
```
jonathan-innis: Do we think we need to expose this externally? If we do, does it make sense to put this as a property of the requirement vs. exposing it as a return value?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -529,29 +530,36 @@ func (s *Scheduler) addToNewNodeClaim(ctx context.Context, pod *corev1.Pod) erro
...
-					len(s.nodeClaimTemplates[i].InstanceTypeOptions)-len(its),
-					len(s.nodeClaimTemplates[i].InstanceTypeOptions),
+					len(s.nodeClaimTemplates[nodeClaimTemplateIndex].InstanceTypeOptions)-len(its),
+					len(s.nodeClaimTemplates[nodeClaimTemplateIndex].InstanceTypeOptions),
 				))
 			}
 		}
-		nodeClaim := NewNodeClaim(s.nodeClaimTemplates[i], s.topology, s.daemonOverhead[s.nodeClaimTemplates[i]], s.daemonHostPortUsage[s.nodeClaimTemplates[i]], its, s.reservationManager, s.reservedOfferingMode)
-		r, its, ofs, err := nodeClaim.CanAdd(ctx, pod, s.cachedPodData[pod.UID])
+		nodeClaim := NewNodeClaim(s.nodeClaimTemplates[nodeClaimTemplateIndex], s.topology, s.daemonOverhead[s.nodeClaimTemplates[nodeClaimTemplateIndex]], s.daemonHostPortUsage[s.nodeClaimTemplates[nodeClaimTemplateIndex]], its, s.reservationManager, s.reservedOfferingMode)
```
jonathan-innis: I think we had mentioned that you should relax within a NodePool first before you try the next one -- I actually think that this makes the implementation easier -- it means that you don't have to iterate through the NodePools twice and can just try and get as high of a minValues as possible within the instance type filtering

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -419,14 +420,18 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
 		// We don't care about the minimum number of instance types that meet our requirements here, we only care if they meet our requirements.
 		_, err.minValuesIncompatibleErr = remaining.SatisfiesMinValues(requirements)
 		if err.minValuesIncompatibleErr != nil {
```
jonathan-innis, line 428: We should probably avoid setting the err here if we are supporting relaxation since we technically don't have an error if we had the flexibility to relax it

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -419,14 +420,18 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
 		// We don't care about the minimum number of instance types that meet our requirements here, we only care if they meet our requirements.
 		_, err.minValuesIncompatibleErr = remaining.SatisfiesMinValues(requirements)
```
jonathan-innis: We should perhaps consider returning a modified requirement set with modified minValues based on the result of the relaxation and the remaining instance types attached to the NodeClaim

pkg/controllers/provisioning/scheduling/nodeclaim.go
jonathan-innis: What's the behavior in this filter function if you have another pod that comes in that doesn't cause us to relax the minValues further and can fit on the NodeClaim? Will we allow it to schedule? In my mind we should so long as it doesn't cause us to continue to relax our NodeClaim even further

pkg/test/options.go
jmdeal: +1. Question though - do we want this to be surfaced as a global option, or should it be a per-requirement setting embedded in the NodePool / NodeClaim?

pkg/operator/options/options.go
```diff
@@ -112,7 +113,7 @@ func (o *Options) AddFlags(fs *FlagSet) {
...
 	fs.StringVar(&o.preferencePolicyRaw, "preference-policy", env.WithDefaultString("PREFERENCE_POLICY", string(PreferencePolicyRespect)), "How the Karpenter scheduler should treat preferences. Preferences include preferredDuringSchedulingIgnoreDuringExecution node and pod affinities/anti-affinities and ScheduleAnyways topologySpreadConstraints. Can be one of 'Ignore' and 'Respect'")
-	fs.StringVar(&o.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "NodeRepair=false,ReservedCapacity=false,SpotToSpotConsolidation=false"), "Optional features can be enabled / disabled using feature gates. Current options are: NodeRepair, ReservedCapacity, and SpotToSpotConsolidation")
+	fs.StringVar(&o.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "NodeRepair=false,ReservedCapacity=false,SpotToSpotConsolidation=false,AutoRelaxMinValues=false"), "Optional features can be enabled / disabled using feature gates. Current options are: NodeRepair, ReservedCapacity, SpotToSpotConsolidation, and AutoRelaxMinValues")
```
jmdeal: A general nit on naming: do we think `AutoRelax` is redundant? I think it can just be relax, in the context of Karpenter everything is automatic. There is no scenario where we attempt to create a NodeClaim, and prompt a user for preferences on its creation.

pkg/test/options.go
saurav-agarwalla: @jonathan-innis Good point, I'll update the feature flag.

@jmdeal deliberately want to start off with a feature flag. As we get more feedback around this feature, we can decide to graduate this to the NodePool API.

pkg/operator/options/options.go
saurav-agarwalla: I am renaming this feature flag to a policy which should resolve this concern.

pkg/metrics/metrics.go
saurav-agarwalla: Yup, this makes even more sense with the switch to a policy type instead of the feature flag.

pkg/controllers/provisioning/scheduling/scheduler.go
saurav-agarwalla: I think we want to show visibility to the customer in case there were nodeclaims being launched with relaxed min values. Would definitely prefer for this to be in the requirement but trying to keep API changes to a minimum for now. I think once we see more adoption for this feature, we can expose both the policy flag and this field in the APIs.

My worry is once we expand the API, it is relatively hard to deprecate a field if we see that this isn't a widely popular feature.

What do you think?

pkg/controllers/provisioning/scheduling/scheduler.go
saurav-agarwalla: There's an `options` structure defined in this package which conflicts with the import.

```
type options struct {
	reservedOfferingMode    ReservedOfferingMode
	preferencePolicy        PreferencePolicy
	numConcurrentReconciles int
}
```

pkg/apis/v1/labels.go
saurav-agarwalla: Mentioned this somewhere else but mostly so that it is easy for customers to know when/why a nodeclaim isn't being launched with the min values constraint. Open to other ways of doing this if you have any suggestions.

pkg/controllers/provisioning/scheduling/nodeclaim.go
saurav-agarwalla: We want to only relax min values in certain cases even when relaxation is enabled (e.g. only relax during new launches and not for in-flight nodeclaims).

pkg/controllers/provisioning/scheduling/scheduler.go
saurav-agarwalla: Sounds good, updated.

pkg/controllers/provisioning/scheduling/nodeclaim.go
saurav-agarwalla: I thought about this initially but decided against it because it seemed to me that it'd complicate the outcome without a lot of value add (I could be wrong though). The way we evaluate min values today is we return on the first requirement for which min values is violated even when there are multiple requirements whose min values would be violated for a particular scheduling decision. With relaxation, how do we decide which one to relax in case there is more than 1? It seemed a little odd to relax multiple requirements. Which is why the way I am thinking of this is relaxation isn't restricted to a particular requirements but the overall scheduling decision.

What do you think?

pkg/controllers/provisioning/scheduling/nodeclaim.go
saurav-agarwalla: `autoRelaxMinValues` is what controls that behavior and hence I chose to only enable it when trying to launch new nodeclaims as opposed to in-flight nodeclaims (even when relaxation is enabled).

saurav-agarwalla: See response below.

pkg/controllers/provisioning/suite_test.go
saurav-agarwalla: Working on those. Should have them up soon.

pkg/apis/v1/nodeclaim.go
saurav-agarwalla: See my response to relaxing the requirements to another comment. We can keep this validation if we do that but I tried multiple ways to see if this can be validated on the basis of the annotation that I added but K8s doesn't seem to allow that.

Depending on what we decide for relaxing the requirements, I can update this.

pkg/cloudprovider/types.go
saurav-agarwalla: The idea is that even when relaxation is enabled, we still want to first see if we can schedule with the min values restriction in place.

pkg/controllers/provisioning/suite_test.go
saurav-agarwalla: Added a test.

pkg/operator/options/options.go
```diff
@@ -112,7 +121,8 @@ func (o *Options) AddFlags(fs *FlagSet) {
...
 	fs.StringVar(&o.preferencePolicyRaw, "preference-policy", env.WithDefaultString("PREFERENCE_POLICY", string(PreferencePolicyRespect)), "How the Karpenter scheduler should treat preferences. Preferences include preferredDuringSchedulingIgnoreDuringExecution node and pod affinities/anti-affinities and ScheduleAnyways topologySpreadConstraints. Can be one of 'Ignore' and 'Respect'")
-	fs.StringVar(&o.FeatureGates.inputStr, "feature-gates", env.WithDefaultString("FEATURE_GATES", "NodeRepair=false,ReservedCapacity=false,SpotToSpotConsolidation=false"), "Optional features can be enabled / disabled using feature gates. Current options are: NodeRepair, ReservedCapacity, and SpotToSpotConsolidation")
+	fs.StringVar(&o.relaxationPolicyRaw, "relaxation-policy", env.WithDefaultString("RELAXATION_POLICY", string(RelaxationPolicyDefault)), "Relaxation policy for scheduling. Options include 'Default' for existing behavior or 'RelaxWhenMinValuesUnsatisfiable' where Karpenter relaxes min values when it isn't satisfied.")
```
jonathan-innis: nit: I feel like this should just be called the `MIN_VALUES_POLICY` or something like that -- relaxation policy doesn't feel specific enough to suggest that it's specifically applying to the minValues portion of our API. To me, this is something like `MIN_VALUES_POLICY=Strict|BestEffort`

pkg/metrics/metrics.go
jonathan-innis: I almost don't think that you need the metric at all -- given this setting is so specific, what's the value of surfacing this metric? If you have the annotation, someone can use something like kube-state to capture the number of nodes with relaxed preferences at any given point in time

pkg/apis/v1/nodeclaim.go
jonathan-innis: After discussing offline, we decided that we will allow the requirement for minValues to be relaxed so that we can have the NodeClaim have minValues requirements that match the number of values that will be passed through from the scheduler.

This will ensure that NodeClaims continue to be a standalone concept and make sense in isolation (outside the context of the scheduler) -- if we were to go the other way, we would have to write special code on the launch path that would create an exception for values not satisfying minValues if we know that the minValuesPolicy is set in the environment variables

pkg/apis/v1/labels.go
jonathan-innis: I could really go either way on this annotation -- it doesn't seem strictly necessary since we will be updating the requirements directly but I can see some use-cases for it with respect to observability -- in general, I think it's fine to keep it.

pkg/cloudprovider/types.go
```diff
@@ -224,11 +225,12 @@ func (its InstanceTypes) SatisfiesMinValues(requirements scheduling.Requirements
...
 // Truncate truncates the InstanceTypes based on the passed-in requirements
 // It returns an error if it isn't possible to truncate the instance types on maxItems without violating minValues
-func (its InstanceTypes) Truncate(requirements scheduling.Requirements, maxItems int) (InstanceTypes, error) {
+func (its InstanceTypes) Truncate(ctx context.Context, requirements scheduling.Requirements, maxItems int) (InstanceTypes, error) {
 	truncatedInstanceTypes := lo.Slice(its.OrderByPrice(requirements), 0, maxItems)
 	// Only check for a validity of NodeClaim if its requirement has minValues in it.
 	if requirements.HasMinValues() {
-		if _, err := truncatedInstanceTypes.SatisfiesMinValues(requirements); err != nil {
+		// If minValues is NOT met for any of the requirement across InstanceTypes, then still allow it if relaxation is enabled.
+		if _, err := truncatedInstanceTypes.SatisfiesMinValues(requirements); err != nil && options.FromContext(ctx).RelaxationPolicy != options.RelaxationPolicyRelaxMinValuesWhenUnsatisfiable {
```
jonathan-innis: nit: I'd probably go the other way with also checking that this value is set to something like "Strict" since that's more intentionally checking what you want here (e.g. if we require that minValues is set and we don't satisfy it)

It's also not a huge deal here though because we only have two options to select from

pkg/metrics/metrics.go
```diff
@@ -44,6 +44,19 @@ var (
 			CapacityTypeLabel,
 		},
 	)
+	NodeClaimsCreatedWithRelaxedPreferencesTotal = opmetrics.NewPrometheusCounter(
```
jonathan-innis: I still don't know if I see a point for this metric to exist -- worst-case it seems like you could just add a label to the existing NodeClaimsCreated metric, but I'm not sure I even think that is too valuable here to have if you have an annotation on the NodeClaim

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -549,7 +550,7 @@ func (s *Scheduler) addToNewNodeClaim(ctx context.Context, pod *corev1.Pod) erro
 			}
 		}
 		nodeClaim := NewNodeClaim(s.nodeClaimTemplates[i], s.topology, s.daemonOverhead[s.nodeClaimTemplates[i]], s.daemonHostPortUsage[s.nodeClaimTemplates[i]], its, s.reservationManager, s.reservedOfferingMode)
-		r, its, ofs, err := nodeClaim.CanAdd(ctx, pod, s.cachedPodData[pod.UID])
+		r, its, ofs, minValuesRelaxed, err := nodeClaim.CanAdd(ctx, pod, s.cachedPodData[pod.UID], karpopts.FromContext(ctx).RelaxationPolicy == karpopts.RelaxationPolicyRelaxMinValuesWhenUnsatisfiable)
```
jonathan-innis: Can you just pull the relaxationPolicy from the scheduler struct directly?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -132,7 +133,7 @@ func NewScheduler(
 	// Pre-filter instance types eligible for NodePools to reduce work done during scheduling loops for pods
 	templates := lo.FilterMap(nodePools, func(np *v1.NodePool, _ int) (*NodeClaimTemplate, bool) {
 		nct := NewNodeClaimTemplate(np)
-		nct.InstanceTypeOptions, _ = filterInstanceTypesByRequirements(instanceTypes[np.Name], nct.Requirements, corev1.ResourceList{}, corev1.ResourceList{}, corev1.ResourceList{})
+		nct.InstanceTypeOptions, _, _ = filterInstanceTypesByRequirements(instanceTypes[np.Name], nct.Requirements, corev1.ResourceList{}, corev1.ResourceList{}, corev1.ResourceList{}, karpopts.FromContext(ctx).RelaxationPolicy == karpopts.RelaxationPolicyRelaxMinValuesWhenUnsatisfiable)
```
jonathan-innis: nit: Same comment here -- can you assign this as a member value on the scheduler and then just pull it from that member value? In general, I'm trying to keep option passthroughs as explicit as possible and trying to prevent us from pulling details out from the context arbitrarily in the scheduler code so the dependencies are clear here and the scheduler can be used as a library

pkg/controllers/provisioning/scheduling/nodeclaim.go
jonathan-innis: Discussed offline: We decided to update the requirements that we are putting in the NodeClaim so that the minValues stays consistent with the count of values that we are passing through to the NodeClaim

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -111,49 +111,49 @@ func NewNodeClaim(
 // CanAdd returns whether the pod can be added to the NodeClaim
 // based on the taints/tolerations, host port compatibility,
 // requirements, resources, reserved capacity reservations, and topology requirements
-func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodData) (updatedRequirements scheduling.Requirements, updatedInstanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering, err error) {
+func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodData, relaxMinValues bool) (updatedRequirements scheduling.Requirements, updatedInstanceTypes []*cloudprovider.InstanceType, offeringsToReserve []*cloudprovider.Offering, minValuesRelaxed bool, err error) {
 	// Check Taints
 	if err := scheduling.Taints(n.Spec.Taints).ToleratesPod(pod); err != nil {
-		return nil, nil, nil, err
+		return nil, nil, nil, minValuesRelaxed, err
```
jonathan-innis:
```suggestion
		return nil, nil, nil, false, err
```

nit: I could see you returning explicit `false` here but I think it's probably fine either way

pkg/cloudprovider/types.go
saurav-agarwalla: Intentionally checking against the relaxed option (`BestEffort`) since I want to make sure this continues to work if clients don't explicitly set this option (for backwards compatibility). Shouldn't ideally happen since we're defaulting to `Strict` but then we have provider implementations which use some of these methods directly.

pkg/controllers/provisioning/scheduling/nodeclaim.go
saurav-agarwalla: Addressed. Just to clarify: I am changing the min values to match how many were/can be satisfied from the scheduler. I added a unit test which should clarify the behavior but TL;DR is if I have nodepool requirements to say `zone-1, zone-2, zone-3` and minValues set to 3 for this key, if the instance type constraint can only support `zone-1`, minValues will be updated to 1 but the values in the requirements will still be `zone-1, zone-2, zone-3` to reflect what the user originally wanted.

pkg/controllers/provisioning/scheduling/scheduler.go
saurav-agarwalla: Updated in some places. It can't be set as a property of the underlying struct in some cases e.g. `NodeClaim` because we want to selectively relax minValues (i.e. only when launching new nodeclaims) so for those cases, I'll need to pass it in the function signature.

pkg/metrics/metrics.go
saurav-agarwalla: Updated this to be a label. Primary reason for this is observability - customers would likely want to know and potentially alarm on this metric/label when nodeclaims are launched with relaxed min values. While annotations are good for discoverability (and for providers to act differently if needed for nodeclaims which launch with relaxed min values), it doesn't help customers alarm in real-time AFAIK.

pkg/controllers/provisioning/scheduling/scheduler.go
saurav-agarwalla: Keeping the annotation for easier discoverability and for provider implementations to know when a nodeclaim launches with relaxed min values as we discussed offline.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -2583,6 +2584,599 @@ var _ = Describe("Provisioning", func() {
...
+					node := ExpectScheduled(ctx, env.Client, pod)
+					nodeClaim := cloudProvider.CreateCalls[0]
+					Expect(node.Labels[corev1.LabelInstanceTypeStable]).To(Equal("instance-type-1"))
+					Expect(node.Labels[corev1.LabelTopologyZone]).To(Or(Equal("test-zone-1"), Equal("test-zone-2")))
+					Expect(node.Annotations[v1.NodeClaimMinValuesRelaxedAnnotationKey]).To(Equal("true"))
+
+					ExpectMetricCounterValue(metrics.NodeClaimsCreatedTotal, 1, map[string]string{
+						metrics.ReasonLabel:           metrics.ProvisionedReason,
+						metrics.NodePoolLabel:         nodeClaim.Labels[v1.NodePoolLabelKey],
+						metrics.CapacityTypeLabel:     "",
```
saurav-agarwalla, line 3056: Trying to figure out why this is happening but doesn't look related to my change. The nodeclaim object has this label but for some reason when the metric is published, this label is missing.

pkg/apis/v1/labels.go
```diff
@@ -53,6 +53,7 @@ const (
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey           = apis.Group + "/nodepool-hash-version"
 	NodeClaimTerminationTimestampAnnotationKey = apis.Group + "/nodeclaim-termination-timestamp"
+	NodeClaimMinValuesRelaxedAnnotationKey     = apis.Group + "/nodeclaim-min-values-relaxed"
```
jmdeal, line 56: Have we considered exposing this information via a status field on the NodeClaim rather than an annotation? I don't think this needs to block - I agree with some of your earlier comments that annotations are good for non-stable features - but I could see value in something like this:
```yaml
status:
  # minValuesRelaxed is a list of requirements Karpenter needed to relax minValues for
  minValuesRelaxed:
    - node.kubernetes.io/instance-type
    - karpenter.k8s.aws/instance-family
```
The one caveat with this is I'm not aware if there would be a good way to create monitors using kube-state metrics for this. I believe it's possible, but I haven't tried it myself.

pkg/cloudprovider/types.go
jmdeal: We're not actually doing anything with this result though if the `MIN_VALUES_POLICY` is `BestEffort` though, right? I don't see the value in calling `SatisfiesMinValues` at all if we're never going to forward the error. This isn't that big of a deal since the behavior is correct regardless, just a bit of an optimization.

pkg/controllers/provisioning/scheduling/instance_selection_test.go
```diff
@@ -47,6 +48,8 @@ var _ = Describe("Instance Type Selection", func() {
 	}
 
 	BeforeEach(func() {
+		fmt.Println("Min Values Policy")
+		fmt.Println(options.FromContext(ctx).MinValuesPolicy)
```
jmdeal: I assume these were leftover from debugging?

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -143,7 +143,13 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 	// Check instance type combinations
 	requests := resources.Merge(n.Spec.Resources.Requests, podData.Requests)
 
-	remaining, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, podData.Requests, n.daemonResources, requests)
+	remaining, unsatisfiableKeys, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, podData.Requests, n.daemonResources, requests, relaxMinValues)
+	if relaxMinValues {
+		// Update min values on the requirements if they are relaxed
+		lo.ForEach(lo.Entries(unsatisfiableKeys), func(e lo.Entry[string, int], _ int) {
```
jmdeal: nit: generally we prefer vanilla for loops over `lo.ForEach` - it doesn't provide much additional value and is trickier to debug due to the callback.

pkg/operator/options/options.go
jmdeal, line 1: It looks like testing is missing for `PREFERENCES_POLICY` as well, but we should add unit tests for the new field.

pkg/controllers/provisioning/suite_test.go
jmdeal: Just spitballing, but does this happen when the NodeClaim is compatible with more than 1 capacity type? In this case, the NodePool doesn't specify a capacity type requirement so we shouldn't be able to determine the capacity type at NodeClaim creation. It's selected by the cloudprovider at launch time.

pkg/controllers/provisioning/suite_test.go
```diff
@@ -2583,6 +2584,599 @@ var _ = Describe("Provisioning", func() {
 			})
 		})
 	})
+
+	Context("MinValuesPolicy", func() {
```
jmdeal, line 2588: Why do we need tests for both zone and instance type? They shouldn't be fundamentally different, right? I feel like we could collapse this to only testing one and have the same coverage.

pkg/apis/v1/labels.go
saurav-agarwalla: For this iteration, I am trying not to touch the API. As we get more feedback about this feature, we can graduate it to the API both in terms of allowing customers to define this on a per-nodepool basis as well as exposing information like this on the status condition.

pkg/controllers/provisioning/scheduling/instance_selection_test.go
saurav-agarwalla: Good catch, will remove this.

pkg/controllers/provisioning/suite_test.go
saurav-agarwalla: Kind of. The reason I have both of these is because the instance-type labels are treated a little differently today since we filter out instance types based on cloud-provider's availability. Just wanted to ensure that I am testing at least a couple of different keys.

pkg/controllers/provisioning/suite_test.go
saurav-agarwalla: Don't think this has ever worked since the label is set by the Cloud Provider: https://github.com/aws/karpenter-provider-aws/blob/c0a5033d90336e32cd03871e0ce944f38582b138/pkg/cloudprovider/cloudprovider.go#L422.

I'll create a separate issue to ensure we clean up this label.

saurav-agarwalla: Created https://github.com/kubernetes-sigs/karpenter/issues/2362.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -143,7 +143,13 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 	// Check instance type combinations
 	requests := resources.Merge(n.Spec.Resources.Requests, podData.Requests)
 
-	remaining, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, podData.Requests, n.daemonResources, requests)
+	remaining, unsatisfiableKeys, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, podData.Requests, n.daemonResources, requests, relaxMinValues)
+	if relaxMinValues {
+		// Update min values on the requirements if they are relaxed
+		for _, e := range lo.Entries(unsatisfiableKeys) {
+			nodeClaimRequirements.Get(e.Key).MinValues = lo.ToPtr(e.Value)
+		}
```
jmdeal: nit: any reason to not use a normal range over map for loop?
```suggestion
		for key, minValues := range unsatisfiableKeys {
			nodeClaimRequirements.Get(key).MinValues = lo.ToPtr(minValues)
		}
```

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -417,16 +424,20 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
+				err.minValuesIncompatibleErr = nil
+			}
 		}
 	}
 	if len(remaining) == 0 {
-		return nil, err
+		return nil, unsatisfiableKeys, err
 	}
-	return remaining, nil
+	return remaining, unsatisfiableKeys, nil
```
jmdeal, line 440: In both of these cases `unsatisfiableKeys` is nil, right? If it's non nil, minValuesIncompatibleErr is always set.

pkg/operator/options/suite_test.go
```diff
@@ -110,6 +110,8 @@ var _ = Describe("Options", func() {
 				LogErrorOutputPaths:     lo.ToPtr("stderr"),
 				BatchMaxDuration:        lo.ToPtr(10 * time.Second),
 				BatchIdleDuration:       lo.ToPtr(time.Second),
+				PreferencePolicy:        lo.ToPtr(options.PreferencePolicyRespect),
```
jmdeal, line 115: It looks like we still need to add the check for these fields in `expectOptionsMatch`

jmdeal (commented):
Only a few more small comments

pkg/controllers/provisioning/scheduling/nodeclaim.go
saurav-agarwalla: Not `nil`, an empty map. Regardless of whether we consider min values as incompatible, I still want to return the keys that didn't satisfy min values. It is needed for both paths - if min values are relaxed, we need this to be able to relax the requirements. If min values aren't relaxed, we still need this for improved logging.

jmdeal: Nevermind, I misread the flow. I thought we returned in the error check above, this looks good.

PR #2300, opened 2025-06-12
fix: Fix bug with MarkForDeletion before creating replacements

jonathan-innis:
Fixes #N/A 

**Description**

Fix a bug where we were marking our pods for deletion before we were creating replacements. This could cause us to launch additional capacity in the provisioner even though the disruption controller is about to create more capacity

`make presubmit`


pkg/controllers/disruption/suite_test.go
```diff
@@ -2212,3 +2276,28 @@ func NewTestingQueue(kubeClient client.Client, recorder events.Recorder, cluster
 	q.TypedRateLimitingInterface = test.NewTypedRateLimitingInterface[*orchestration.Command](workqueue.TypedQueueConfig[*orchestration.Command]{Name: "disruption.workqueue"})
 	return q
 }
+
+type hangCreateClient struct {
```
DerekFrank, line 2280: This strikes me as a useful concept for lots of kinds of testing. Similar to the ExpectRequeued() we should perhaps generalize this and move it into a test/ package

pkg/controllers/disruption/suite_test.go
```diff
@@ -460,6 +456,74 @@ var _ = Describe("Simulate Scheduling", func() {
...
+	It("should ensure that we do not duplicate capacity for disrupted nodes with provisioning", func() {
+		// We create a client that hangs Create() so that when we try to create replacements
+		// we give ourselves time to check that we wouldn't provision additional capacity before the replacements are made
+		hangCreateClient := newHangCreateClient(env.Client)
+		defer hangCreateClient.Stop()
+
+		p := provisioning.NewProvisioner(hangCreateClient, recorder, cloudProvider, cluster, fakeClock)
+		dc := disruption.NewController(fakeClock, env.Client, p, cloudProvider, recorder, cluster, queue)
+
+		nodeClaim, node := test.NodeClaimAndNode(v1.NodeClaim{
```
DerekFrank, line 468: Do we need all of these overrides?

pkg/controllers/disruption/suite_test.go
DerekFrank: Outside the scope of this PR ofc :)

pkg/controllers/disruption/suite_test.go
jonathan-innis: yeah, we need all of these things defined for the node to be a valid candidate

pkg/controllers/disruption/suite_test.go
jonathan-innis: Sure, since I can't think of how it would be generally useful, I'm just going to leave it here for now

PR #2305, opened 2025-06-12
chore: Refactor Orchestration Queue and Handle Mark/Unmark Deletion in Queue

jonathan-innis:
Fixes #N/A 

**Description**

This change refactors the orchestration queue and disruption controller to remove some duplicated code and collapse more logic into the `disruption.Command`. It does this by:

1. Remove the `orchestration.Command` and instead using the `disruption.Command` construct in the orchestration queue
2. Removing the orchestration package so that we can access all of the same types across the controller and the queue
3. Collapse all of the logic in `executeCommand` into `StartCommand` in the orchestration queue so that we perform all of the command execution logic in the context of the queue
4. Remove the `UnmarkForDeletion` call at the top of the disruption Reconcile and handle Unmarking through error handling
5. Creating symmetric `UnmarkForDeletion` and `MarkForDeletion` inside of the orchestration queue

`make presubmit`


pkg/controllers/disruption/types.go
```diff
@@ -117,9 +118,25 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
...
+type Replacement struct {
+	*scheduling.NodeClaim
+
+	Name string
+	// Use a bool track if a node has already been initialized so we can fire metrics for intialization once.
+	// This intentionally does not capture nodes that go initialized then go NotReady after as other pods can
+	// schedule to this node as well.
+	Initialized bool
+}
+
```
rschalo, line 130: Why do we need a new type if these are available on the NodeClaim?

pkg/controllers/disruption/helpers.go
rschalo, line 1: If we definitely need `Replacement` then wdyt about a helper function for that `lo.Map()`?

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -218,7 +215,8 @@ func filterOutSameType(newNodeClaim *scheduling.NodeClaim, consolidate []*Candid
 		}
 	}
 	// swallow the error since we don't allow min values to impact reschedulability in multi node claim
-	newNodeClaim, err := newNodeClaim.RemoveInstanceTypeOptionsByPriceAndMinValues(newNodeClaim.Requirements, maxPrice)
+	var err error
+	newNodeClaim.NodeClaim, err = newNodeClaim.NodeClaim.RemoveInstanceTypeOptionsByPriceAndMinValues(newNodeClaim.Requirements, maxPrice)
```
rschalo: nit: `replacement`

pkg/controllers/disruption/queue.go
```diff
@@ -282,44 +216,135 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
 
-// Add adds commands to the Queue
-// Each command added to the queue should already be validated and ready for execution.
-func (q *Queue) Add(cmd *Command) error {
-	providerIDs := lo.Map(cmd.candidates, func(s *state.StateNode, _ int) string {
-		return s.ProviderID()
+// markDisrupted taints the node and adds the Disrupted condition to the NodeClaim for a candidate that is about to be disrupted
+func (q *Queue) markDisrupted(ctx context.Context, cmd *Command) ([]*Candidate, error) {
+	errs := make([]error, len(cmd.Candidates))
+	workqueue.ParallelizeUntil(ctx, len(cmd.Candidates), len(cmd.Candidates), func(i int) {
```
rschalo, line 240: Is this resource safe to run in parallel for a large number of candidates?

pkg/controllers/disruption/queue.go
```diff
@@ -282,44 +216,135 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) error {
...
+		return serrors.Wrap(fmt.Errorf("marking disrupted, %w", markDisruptedErr), "command-id", cmd.ID)
+	}
+
+	// Update the command to only consider the successfully MarkDisrupted candidates
+	cmd.Candidates = markedCandidates
+
+	if err := q.createReplacementNodeClaims(ctx, cmd); err != nil {
+		// If we failed to launch the replacement, don't disrupt.  If this is some permanent failure,
+		// we don't want to disrupt workloads with no way to provision new nodes for them.
+		return serrors.Wrap(fmt.Errorf("launching replacement nodeclaim, %w", err), "command-id", cmd.ID)
```
rschalo, line 312: When this fails, the replacement nodeclaims that were created will be garbage collected because there is no node for the nodeclaim?

pkg/controllers/disruption/queue.go
```diff
@@ -216,35 +146,39 @@ func (q *Queue) Reconcile(ctx context.Context) (reconcile.Result, error) {
...
+			ConsolidationTypeLabel: cmd.ConsolidationType(),
 		})
-		multiErr := multierr.Combine(err, cmd.lastError, state.RequireNoScheduleTaint(ctx, q.kubeClient, false, cmd.candidates...))
-		multiErr = multierr.Combine(multiErr, state.ClearNodeClaimsCondition(ctx, q.kubeClient, v1.ConditionTypeDisruptionReason, cmd.candidates...))
+		stateNodes := lo.Map(cmd.Candidates, func(c *Candidate, _ int) *state.StateNode { return c.StateNode })
+		multiErr := multierr.Combine(err, state.RequireNoScheduleTaint(ctx, q.kubeClient, false, stateNodes...))
+		multiErr = multierr.Combine(multiErr, state.ClearNodeClaimsCondition(ctx, q.kubeClient, v1.ConditionTypeDisruptionReason, stateNodes...))
 		// Log the error
 		log.FromContext(ctx).Error(multiErr, "failed terminating nodes while executing a disruption command")
+		q.CompleteCommand(cmd, false)
```
rschalo: If the result is on each `cmd` we can call this once on `:168`

pkg/controllers/disruption/queue.go
```diff
@@ -336,15 +361,17 @@ func (q *Queue) HasAny(ids ...string) bool {
 	return ok
 }
 
-// Remove fully clears the queue of all references of a hash/command
-func (q *Queue) Remove(cmd *Command) {
+// CompleteCommand fully clears the queue of all references of a hash/command
+func (q *Queue) CompleteCommand(cmd *Command, succeeded bool) {
```
rschalo: Do want a bool here or should there be some status on the `Command` that conveys whether or not that specific `cmd` succeeded.

pkg/controllers/disruption/queue.go
jonathan-innis: How do you mean? What does resource-safe mean?

pkg/controllers/disruption/queue.go
jonathan-innis: Not sure that I follow -- if a portion of the replacements fails to come up, then there will be a few more NodeClaims present than there should be -- this is part of the failure mode that we expect to clean-up through emptiness. There really isn't a better way to handle this than that since this is a declarative system and any imperative action that we would attempt to take could fail at any poitn

pkg/controllers/disruption/queue.go
jonathan-innis: I felt like that was the same either way -- I'm not a huge fan of boolean flags like this, but I really couldn't come up with a better way to code this than that -- the alternative was to just create a different function but I didn't like that either

pkg/controllers/disruption/types.go
jonathan-innis: scheduling.NodeClaim doesn't contain the field Name (because it assumes that this NodeClaim isn't launched yet) and also doesn't track initialized

pkg/controllers/disruption/helpers.go
jonathan-innis: I'm not sure which function that you are referring to here

pkg/controllers/disruption/queue.go
jonathan-innis: Discussed offline: Added a `Succeeded` boolean to the command to mark success and going to UnmarkForDeletion based on that boolean on Complete

pkg/controllers/disruption/queue.go
jonathan-innis: Discussed this offline: Given that we don't think there are going to be that large of a number of candidates and, even if there are, we are going to get client-side throttled before it becomes an issue, we should be fine to have the parallelism be this high

pkg/controllers/disruption/helpers.go
jonathan-innis: Discussed offline: This is the replacement mapping function

pkg/controllers/disruption/queue_test.go
rschalo, line 1: A test wasn't present for `queue.Remove` and isn't for `queue.CompleteCommand`. Because this isn't a regression in coverage, it isn't blocking but we should add a test to cover it.

PR #2307, opened 2025-06-13
chore: Bump lifecycle cache expiration to one hour

jonathan-innis:
Fixes #N/A 

**Description**

Bump lifecycle cache expiration from 1m to 1h to ensure that we keep track of launched NodeClaims to avoid double-launches, particularly when Watches are slow and take longer than 1m to update our internal caches

`make presubmit`


engedaam (commented):
/lgtm

PR #2308, opened 2025-06-13
chore: Use cluster state to check replacement NodeClaim existence

jonathan-innis:
Fixes #N/A 

**Description**

Rather than using a time-based mechanism to ensure that we bail out of `waitOrTerminate` when we get an ICE error, this change updates it to use cluster state as a consistent in-memory store. Cluster state is updated immediately when the NodeClaims are created. The NodeClaim name is only removed from the cluster state if the NodeClaim informer sees the object deleted.

`make presubmit`


engedaam (commented):
/lgtm

Issue #749, opened 2022-07-05
Mega Issue: Manual node provisioning


ellistarn:
**Tell us about your request**
What do you want us to build?

I'm seeing a number of feature requests to launch nodes separately from pending pods. This issue is intended to broadly track this discussion. 

Use Cases:
- Create a System pool to run components like karpenter, loadbalancer, coredns, etc
- Provision baseline capacity that never scales down
- Manually preprovision a set of nodes before a large event

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

chamarakera:
When migrating from Cluster Auto Scaler to Karpenter, we would like Karpenter to provision a node beforehand, before we perform a drain on the old node. It takes time for Karpenter to provision a new node based on the unscheduled workloads, and due to this, the pods are kept in a Pending state for too long.

Bryce-Soghigian:
@chamarakera would just requesting a node via nodeclaim be enough in this case? No managed solution is requried if you just want to create a node from my testing just applying nodeclaims with a reference to valid nodepool is often enough. 


Note this example uses a instance type size from azure. 

k apply -f nodeclaim.yaml

```yaml
apiVersion: karpenter.sh/v1beta1
kind: NodeClaim
metadata:
  name: temporary-capacity
  labels:
    karpenter.sh/nodepool: general-purpose
  annotations:
    karpenter.sh/do-not-disrupt: "true"
spec:
  nodeClassRef:
    name: default
  requirements:
  - key: kubernetes.io/arch
    operator: In
    values:
    - amd64
  - key: kubernetes.io/os
    operator: In
    values:
    - linux
  - key: karpenter.sh/capacity-type
    operator: In
    values:
    - on-demand
  - key: node.kubernetes.io/instance-type
    operator: In
    values:
    - Standard_D32_v3
  resources:
    requests:
      cpu: 2310m
      memory: 725280Ki
      pods: "7"
status: {}
 ```

njtran:
@chamarakera how bad is the pod latency you're describing here? Do you have a long bootstrap/startup time on your instance? What would be an acceptable amount of pod latency to not need prewarmed instances ?

[...]

PR #2309, opened 2025-06-13
docs: static capacity RFC

sumukha-radhakrishna:
Fixes 

* Static Capacity support for Karpenter - [#749]

## Description 

This RFC plans to address static provisioning feature request by extending existing Nodepool to support replicas.

### DIFF
--- a/designs/static-capacity.md
+++ b/designs/static-capacity.md
@@ -1,0 +1,237 @@
+# Karpenter - Static Capacity
+
+## Background
+Karpenter currently operates as a dynamic cluster autoscaler, automatically adjusting node counts based on pending pod demand. However, several important use cases require maintaining a fixed set of nodes such as:
+
+1. Performance-critical applications where just-in-time provisioning latency is unacceptable
+2. Workloads that require predictable, always-available capacity
+3. Operational models that rely on a fixed number of nodes for budgeting, isolation, or infrastructure boundary
+
+Currently, users attempt to achieve static capacity through workarounds:
+- Configuring provisioner requirements to run placeholder pods to maintain minimum capacity
+- Using separate node management tools alongside Karpenter
+
+## Proposal
+Extend the existing NodePool resource to support static provisioning capabilities by adding new fields:
+
+```yaml
+spec:
+  # Makes NodePool static when specified
+  replicas: 5  # Number of nodes to maintain
+limits:
+  # Prevents Karpenter from creating new replicas once the limit is exceeded
+  nodes: 10 # Constrain the size of static Nodepool
+status:
+  nodes: 10 # shows the current size of Nodepool
+```
+The existence of replicas field will determine if NodePool is Static
+
+Key aspects:
+1. Static NodePools maintain fixed node count regardless of pod demand
+2. Will not be considered as consolidation candidate, so consolidation settings are not allowed other than defaults (at least in v1alpha1)
+3. Optional limits.nodes can be set to constrain the maximum size of static Nodepool when bursting (e.g. during drift or expiration)
+4. Inherit existing Karpenter features
+5. Use existing NodeClaim infrastructure with owner references
+  
+We will support scaling static node pools using a command such as:
+
+```sh
+kubectl scale nodepool <name> --replicas=10
+```
+
+To support this we will add Nodes field of type int64 under status. `.status.Nodes` will contain the current Node count for a NodePool.
+
+### Modeling & Validation
+
+When `replicas` is specified, the `NodePool` enters a static provisioning mode where certain disruption-related fields, weight and limits become irrelevant or misleading. Specifically:
+
+- `consolidationPolicy` and `consolidateAfter` (which control Karpenter’s consolidation logic) must not be used when the node count is fixed via `replicas`.
+- limits other than `limits.nodes` must not be set
+- weight must not be set
+
+We will have strict validation against setting limits (other than limits.nodes) and weight but if consolidation parameters are set then it will simply be ignored.
+
+Once a NodePool has replicas set, it cannot be removed unless the NodePool itself is deleted. This means a NodePool cannot switch between static and dynamic modes:
+- If a NodePool is static (replicas set), replicas cannot be unset or modified to nil.
+- If a NodePool is dynamic (replicas not set), replicas cannot be added.
+- This validation prevents unintentional large-scale terminations or sudden behavior changes in consolidation logic.
+
+
+#### Validation rules
+
+There will not be a validation against consolidation settings (consolidationPolicy and consolidateAfter), if specified it will simply be ignored
+
+```yaml
+If has(self.replicas) != has(oldSelf.replicas)
+  => Validation Error
+```
+```yaml
+if replicas != nil:
+  - limits other than limits.nodes must not be set
+  - weight must not be set
+  => ValidationError if not
+```
+
+### Static NodePool limits
+
+We will not be supporting limits other than limits.node, while dynamic provisioning does filter NodeClaim templates based on limits (e.g., limits.cpu, limits.memory), it operates under the assumption that pods are batched and packed efficiently into a single NodeClaim. This is fundamentally different from static provisioning, where we attempt to create a fixed number of NodeClaims (replicas). Limits would start to dictate instance shape, not just aggregate capacity. This creates a conflict between honoring replicas vs. satisfying hard limits.
+
+However, we are introducing a new field, limits.nodes, to explicitly cap the maximum number of nodes which can be provisioned for a given NodePool at a point in time.
+
+### Disruption
+
+Karpenter already has Disruption for nodes we can use the same mechanism. 
+Since Static Nodepool and NodeClaim is a variant of existing Nodepool/NodeClaim it will inherit Karpenter dynamic provider integration. Drift Detection can also be inherited to trigger replacement of drifted nodes. During this Karpenter will respect the Disruption Budget. If limits.nodes are specified then Karpenter would wait until Nodes are terminated and count is well within node limits before creating replacements.
+
+Replacement will be one-for-one i.e create replacements before terminating the old one to honor replicas if Nodepool node count do not exceed the node limits. 
+When a user scales down a static NodePool, Karpenter will delete NodeClaims and gracefully terminate nodes.
+Key behaviors:
+- User-driven actions (scaling replicas) will not respect disruption budgets. The user explicitly intends to remove capacity, and we honor that request by deleting corresponding NodeClaims. This will be classified as an [automated forceful disruption method](https://karpenter.sh/docs/concepts/disruption/#automated-forceful-methods), a la expiration.
+- Supported Karpenter-driven actions (e.g., drift) respect disruption budgets and scheduling safety.
+
+Deletion of nodepool would disrupt all of its NodeClaims. Garbage collection logic do not differ between static and dynamic NodePools.
+
+**Important Note**: The scale-down of replicas in a static NodePool will follow standard forceful disruption semantics. This means that when the desired replica count is reduced, Karpenter will drain nodes and while respecting PodDisruptionBudgets (PDBs), terminationGracePeriod, and other workload disruption safeguards. We will not bypass these constraints, as PDBs represent an explicit agreement between the cluster operator and the controller to maintain application availability during changes. This aligns with Kubernetes conventions, where scaling actions are reconciled towards the desired state without ignoring safety guarantees, and ensures predictable behavior for workloads. While this may mean that scale-down is delayed when eviction blockers are present, it provides consistency with the broader Kubernetes ecosystem.
+
+
+### Consolidation
+
+Static NodePools are not eligible for consolidation. They act like any other static capacity source (e.g other nodes in the cluster). Their lifecycle is managed directly by the user, not by Karpenter’s dynamic optimization logic.
+However, the nodes provisioned by a static pool:
+- Participate in scheduling decisions for both launched capacity and yet to launch inflight capacity (i.e., pods can land on them)
+- Are monitored for drift, enabling graceful re-creation when necessary
+
+This means they remain first-class citizens in the cluster but are excluded from cost-based disruption decisions.
+
+## Requirements
+
+The requirements field in a static NodePool behaves identically to dynamic pools—it defines the constraints for all NodeClaims launched under that NodePool.
+In static pools, we must choose multiple concrete node configurations up front—i.e., for replicas: 10, we select 10 NodeClaims matching the requirement set.
+If the requirements allow multiple combinations:
+- Karpenter selects the optimal combination based on cost, availability, and zone balancing
+- This selection is done once at provisioning time (unlike dynamic pools, where evaluation occurs per provisioning event)
+
+In static provisioning, the NodeClaim requirements are directly derived from the NodeClaimTemplate on the NodePool. These are evaluated once per NodeClaim at creation, meaning the selection is based solely on what the template allows.
+As a result, even though all NodeClaims come from the same static NodePool, they may still result in different instance types (shapes/flavors), depending on availability, since that decision happens during cloud-provider Create() call.
+
+#### Zonal Distribution
+Static NodePools do not currently support topology-aware spreading in the same way that dynamic NodePools rely on pod-driven scheduling. Initially, we explored automatically spreading static NodeClaims across zones when the topology.kubernetes.io/zone requirement includes multiple values, for example:
+
+```yaml
+- key: topology.kubernetes.io/zone
+  operator: In
+  values: ["zone-2a", "zone-2b", "zone-2c"]
+```
+
+However, this introduces implicit behavior that conflicts with existing Karpenter semantics. In current behavior, absence of a requirement is treated as unconstrained, and specifying zones in requirements is interpreted as a hard constraint. Introducing automatic spreading implicitly redefines these semantics, which could have long-term consequences for other requirement types (e.g., capacity type).
+Given these concerns, we’re punting zonal spreading to a future enhancement. For now, we will continue to treat the requirements field as the single source of truth for topology decisions. Users who want to spread nodes across zones can do so explicitly by:
+- Specifying multiple zones in the topology.kubernetes.io/zone requirement, or
+- Creating multiple static NodePools, each pinned to a specific AZ.
+
+We plan to explore spread option to offer users a more flexible and declarative way to express zone-spreading behavior, similar to podTopologySpreadConstraints.
+
+
+## Example 
+I want a Static NodePool that:
+- Always keeps exactly 12 running nodes under normal operation.
+- Can temporarily scale up to a maximum of 20 nodes, but never exceed that.
+- Runs only in a single availability zone (zone-2a).
+- Uses a fixed instance type (m5.2xlarge) for predictable capacity.
+- Has nodes that expire after 720 hours.
+- Follows disruption budgets allowing at most 10% of nodes to be disrupted at a time,
+- and blocks all deprovisioning during weekday business hours.
+
+```yaml
+apiVersion: karpenter.sh/v1beta1
+kind: NodePool
+metadata:
+  name: static-prod-nodepool
+spec:
+  replicas: 12 # Maintains 12 nodes
+  template:
+    metadata:
+      labels:
+        nodepool-type: static
+    spec:
+      nodeClassRef:
+        group: karpenter.k8s.aws
+        kind: EC2NodeClass
+        name: myEc2NodeClass
+      expireAfter: 720h
+      taints:
+        - key: example.com/special-taint
+          effect: NoSchedule
+      requirements:
+        - key: topology.kubernetes.io/zone
+          operator: In
+          values: ["zone-2a"]
+        - key: karpenter.k8s.aws/instance-type
+          operator: In
+          values: ["m5.2xlarge"]
+  disruption:
+    budgets:
+      - nodes: 10%
+      # On Weekdays during business hours, don't do any deprovisioning.
+      - schedule: "0 9 * * mon-fri"
+        duration: 8h
+        nodes: "0"
+  limits:
+    # Constrain the size of static Nodepool 
+    nodes: 20 # Karpenter will not scale up nodes or create replacements if limits are breaching
+```
+
+## Implementation Details
+
+### Controller Architecture
+
+We will be creating Static Capacity by adding 2 new controllers under a feature flag. This will be added to v1alpha1 release to get feedback from users.
+
+#### Scale-Up Controller
+- **Purpose**: Scale up NodeClaims to meet desired replicas
+- **Behavior**: Monitors the difference between desired replicas and current healthy NodeClaims, provisioning new nodes as needed respecting limits.nodes value and topology requirements
+- **Instance Selection**: Scale-up controller will be creating NodeClaims and Instance selection will happen during CloudProvider call. During which we Will find the cheapest instance possible based on requirements
+- **Consolidation**: Static NodePools will not be consolidated to maintain predictable capacity
+
+#### Deprovisioning Controller  
+- **Purpose**: Scale down replicas when desired count is reduced
+- **Node Selection**: Will prioritize empty nodes for termination. Incase of non-empty nodes we will pick at random. In future iterations, we will add intelligent algorithms for selecting which nodes to scale down
+- **Termination Behavior**: Will gracefully terminate nodes to meet desired size respecting PodDisruptionBudgets or terminationGracePeriod
+
+### Scheduling Integration
+
+We will consider static nodeclaims that are already launched and to be launched during scheduling simulation. This approach will:
+- Lessen the churn and possibility of over-provisioning
+- Be accounted for during disruption/consolidation operations
+- Be considered during new dynamic capacity provisioning decisions
+
+This ensures that the scheduler is aware of both existing and planned static capacity when making provisioning decisions.
+
+### Testing Strategy
+
+For testing, we will add comprehensive integration tests to ensure the feature works correctly across different scenarios:
+- Scale-up and Scale-down operations 
+- Scale-up operation 
+- Ensure NodeClaims created respect topology requirements
+- Static NodeClaims get drifted
+
+### Observability
+
+Controller-runtime metrics already provide baseline visibility into reconcile performance and errors. Static NodePools also expose spec.replicas (desired count) and status.nodes (current count), which we will surface as metrics.
+We’re not adding extra status fields now to keep the API surface minimal, but can introduce fields like pending/terminating replicas in future if needed.
+
+
+## Alternative Proposal: New Static NodePool API
+
+
+The alternative approach would be to create a separate StaticNodePool API focused solely on static provisioning. This would include:
+- Dedicated API for static provisioning
+- Clear separation from existing Nodepool 
+- Validation rules specific to static provisioning
+
+However, this approach was rejected because:
+- Many core functionalities would be shared between static and dynamic capacity management
+- Creates unnecessary cognitive overhead for users
+- Requires duplicate documentation
+
+The better approach is to extend the existing NodePool API since the differences represent different modes of the same fundamental abstraction rather than entirely separate concepts.
+


designs/static-capacity.md
```diff
@@ -0,0 +1,141 @@
+# Karpenter - Static Capacity
+
+## Background
+Karpenter currently operates as a dynamic cluster autoscaler, automatically adjusting node counts based on pending pod demand. However, several important use cases require maintaining a fixed set of nodes:
```
engedaam: Can we call out the specific use-cases that need manual provisioning?

designs/static-capacity.md
```diff
@@ -0,0 +1,141 @@
...
+spec:
+  # Makes NodePool static when specified
+  replicas: 5  # Number of nodes to maintain
+```
+
+Key aspects:
+1. Static NodePools maintain fixed node count regardless of pod demand
+2. Will not be considered as consolidation candidate (at least in v1alpha1)
+3. Nodes are spread across AZs by default
+4. Inherit existing Karpenter features like drift detection and interruption handling
```
engedaam: Do you mean spot interruption handling? You should drop this part, since this is really a cloud provider feature

designs/static-capacity.md
```diff
@@ -0,0 +1,141 @@
...
+  
+
+### Modeling & Validation
+
+When `replicas` is specified, the `NodePool` enters a static provisioning mode where certain disruption-related fields become irrelevant or misleading. Specifically:
+
+- `limits` and `weight` (used to influence consolidation and node selection) are not meaningful for static capacity.
+- `consolidationPolicy` and `consolidateAfter` (which control Karpenter’s consolidation logic) must not be used when the node count is fixed via `replicas`.
+
+However, **Kubernetes' schema-level validation (OpenAPI-based) lacks the ability to express conditional logic between fields**. This introduces limitations:
```
engedaam: Can we not do CEL validation here?

designs/static-capacity.md
```diff
@@ -0,0 +1,141 @@
...
+
+- Non-pointer types defaulting at parse time
+- Default values being indistinguishable from explicitly set values in admission webhooks
+
+### Disruption
+
+Karpenter already has Disruption and Interruption Management for nodes we can use the same mechanism. 
+Since Static Nodepool and NodeClaim is a variant of existing Nodepool/NodeClaim it will inherit Karpenter dynamic provider integration. Drift Detection can also be inherited to trigger replacement of drifted nodes. During this Karpenter will respect the Disruption Budget.
```
engedaam: What do you mean by dynamic provider integration?

designs/static-capacity.md
```diff
@@ -0,0 +1,141 @@
...
+Extend the existing NodePool resource to support static provisioning capabilities by adding new fields:
+
+```yaml
+spec:
+  # Makes NodePool static when specified
+  replicas: 5  # Number of nodes to maintain
+```
+
+Key aspects:
+1. Static NodePools maintain fixed node count regardless of pod demand
```
engedaam, line 30: Can we define what you mean by static and dynamic NodePool to make it clear for new readers on the doc?

designs/static-capacity.md
sumukha-radhakrishna: Added.

designs/static-capacity.md
sumukha-radhakrishna: Its there in the yaml, will add explicitly

designs/static-capacity.md
sumukha-radhakrishna: will remove this

designs/static-capacity.md
sumukha-radhakrishna: Supposed to be dynamic provisioning integration, fixed

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+  - consolidationPolicy must be nil or default "WhenEmptyOrUnderutilized" or "Never"
+  - consolidateAfter must be nil or default 0s
+  - limits must be nil
+  => ValidationError if not
+```
+
+### Disruption
+
+Karpenter already has Disruption for nodes we can use the same mechanism. 
+Since Static Nodepool and NodeClaim is a variant of existing Nodepool/NodeClaim it will inherit Karpenter dynamic provider integration. Drift Detection can also be inherited to trigger replacement of drifted nodes. During this Karpenter will respect the Disruption Budget.
```
engedaam: What do you mean by Static NodeClaim? Are you thinking there will be any difference from a dynamic NodeClaim?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+3. Nodes are spread based on the topology requirements
+4. Inherit existing Karpenter features
+5. Use existing NodeClaim infrastructure with owner references
+  
+
+### Modeling & Validation
+
+When `replicas` is specified, the `NodePool` enters a static provisioning mode where certain disruption-related fields become irrelevant or misleading. Specifically:
+
+- `limits` (used to influence consolidation and node selection) are not meaningful for static capacity.
```
engedaam: Don't we want to enable setting on node count? Mainly in the case of involuntary disruption and there are hard scaling requirements that the cluster cannot exceed?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
+# Karpenter - Static Capacity
+
+## Background
+Karpenter currently operates as a dynamic cluster autoscaler, automatically adjusting node counts based on pending pod demand. However, several important use cases require maintaining a fixed set of nodes such as:
+
+1. Performance-critical applications where just-in-time provisioning latency is unacceptable
+2. Workloads that need guaranteed capacity regardless of pod demand
```
jmdeal: What does this mean? Is this referring to applications which are running on worker nodes, but aren't containerized?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
+# Karpenter - Static Capacity
+
+## Background
+Karpenter currently operates as a dynamic cluster autoscaler, automatically adjusting node counts based on pending pod demand. However, several important use cases require maintaining a fixed set of nodes such as:
+
+1. Performance-critical applications where just-in-time provisioning latency is unacceptable
+2. Workloads that need guaranteed capacity regardless of pod demand  
+3. Scenarios where fixed node counts are preferred
```
jmdeal: What would this scenario look like? I can't think of a scenario where the fixed count is important. Generally the only use-case I've heard for pre-provisioned capacity is the pod startup latency use-case, but there the number of nodes isn't really relevant.

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+
+Currently, users attempt to achieve static capacity through workarounds:
+- Running placeholder pods to maintain minimum capacity
+- Using separate node management tools alongside Karpenter (MNG/Fargate/Self-managed)
+- Configuring complex provisioner requirements to approximate static behavior
```
jmdeal: What would these requirements look like? I think this really only makes sense in combination with point 1 too - without placeholder pods, Karpenter will not provision capacity.

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+## Background
+Karpenter currently operates as a dynamic cluster autoscaler, automatically adjusting node counts based on pending pod demand. However, several important use cases require maintaining a fixed set of nodes such as:
+
+1. Performance-critical applications where just-in-time provisioning latency is unacceptable
+2. Workloads that need guaranteed capacity regardless of pod demand  
+3. Scenarios where fixed node counts are preferred
+
+Currently, users attempt to achieve static capacity through workarounds:
+- Running placeholder pods to maintain minimum capacity
+- Using separate node management tools alongside Karpenter (MNG/Fargate/Self-managed)
```
jmdeal: I'd exclude fargate - unless there's a provisioning mode I'm unaware of for it, that's only just-in-time capacity.

designs/static-capacity.md
jmdeal: +1, I think this would be necessary to support a "max burst" semantic during drift as well, though maybe that can be handled sufficiently through budgets.

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+  => ValidationError if not
+```
+
+### Disruption
+
+Karpenter already has Disruption for nodes we can use the same mechanism. 
+Since Static Nodepool and NodeClaim is a variant of existing Nodepool/NodeClaim it will inherit Karpenter dynamic provider integration. Drift Detection can also be inherited to trigger replacement of drifted nodes. During this Karpenter will respect the Disruption Budget.
+
+Replacement will be one-for-one i.e create replacements before terminating the old one to honor replicas.
+When a user scales down a static NodePool, Karpenter will forcefully drain nodes and terminate the corresponding NodeClaims/instances.
```
jmdeal: What does forcibly mean in this context? Will Karpenter respect PDBs, do-not-disrupt annotations, and volume attachments or will it bypass the drain process and begin terminating the instance?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+Key behavior distinctions:
+- User-driven actions (scaling replicas) bypass disruption budgets. The user explicitly intends to remove capacity, and we honor that request immediately.
```
jmdeal: I wouldn't consider this bypassing disruption budgets. By definition, disruption budgets only apply to voluntary disruption mechanisms (e.g. drift, consolidation, etc). We don't consider forceful mechanisms to bypass disruption budgets, those budgets just never applied to them in the first place. I'd reword this to avoid confusion.

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+- Karpenter-driven actions (e.g., consolidation, drift) respect disruption budgets and scheduling safety. These are only applicable when replicas is not set.
```
jmdeal: I thought drift still functioned the same for static nodepools?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+  
+
+### Modeling & Validation
+
+When `replicas` is specified, the `NodePool` enters a static provisioning mode where certain disruption-related fields become irrelevant or misleading. Specifically:
+
+- `limits` (used to influence consolidation and node selection) are not meaningful for static capacity.
+- `consolidationPolicy` and `consolidateAfter` (which control Karpenter’s consolidation logic) must not be used when the node count is fixed via `replicas`.
+
+#### Validation rules
```
jmdeal, line 60: Do we want to consider a validation rule to ensure a NodePool must stay static or dynamic? I.E. if replicas was previously set it can't be removed, and vice versa? If we don't add this validation, what would be the behavior if you added or removed replicas to an existing NodePool? Would there be any footguns that we need to inform users of?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+
+This means they remain first-class citizens in the cluster but are excluded from cost-based disruption decisions.
+
+## Requirements
+
+The requirements field in a static NodePool behaves identically to dynamic pools—it defines the constraints for all NodeClaims launched under that NodePool.
+In static pools, we must choose multiple concrete node configurations up front—i.e., for replicas: 10, we select 10 NodeClaims matching the requirement set.
+If the requirements allow multiple combinations:
+- Karpenter selects the optimal combination based on cost, availability, and zone balancing
+- This selection is done once at provisioning time (unlike dynamic pools, where evaluation occurs per provisioning event)
```
jmdeal, line 112: I don't grok the distinction you're making here. Concrete instance type selection is made at the same time (during the cloudprovider `Create` call). The difference is just the generation of the NodeClaim requirements. For dynamic provisioning we constrain the requirements based on the pods we schedule to each. For static provisioning, we'll just create `n` nodeclaims based on the NodeClaimTemplate on the NodePool

designs/static-capacity.md
sumukha-radhakrishna: Static Nodepool is nodepool that has replicas specified.

designs/static-capacity.md
sumukha-radhakrishna: Sync'ed up offline. 
Basically we need a way to bound the nodes that can be created for a cluster, during involuntary disruption we might see multiple nodes going down at the same time might cause the controller to create replacements, we might get into a situation where 2x nodes might be present on the cluster (extreme worst case), this might cause issues especially if the cluster is large. We could bound nodes by limiting them under limits section and only launch new NC if we are not violating limits

designs/static-capacity.md
ellistarn: workloads with static scale

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+```yaml
+spec:
+  # Makes NodePool static when specified
+  replicas: 5  # Number of nodes to maintain
+```
+The existence of replicas field will determine if NodePool is Static
+
+Key aspects:
+1. Static NodePools maintain fixed node count regardless of pod demand
+2. Will not be considered as consolidation candidate, so consolidation settings are not allowed other than defaults (at least in v1alpha1)
```
ellistarn, line 31: Do we want to prevent these settings from being set if replicas are set (i.e. validation) or just ignore them if they're set?

designs/static-capacity.md
```diff
@@ -0,0 +1,187 @@
...
+When `replicas` is specified, the `NodePool` enters a static provisioning mode where certain disruption-related fields become irrelevant or misleading. Specifically:
+
+- `limits` (used to influence consolidation and node selection) are not meaningful for static capacity.
+- `consolidationPolicy` and `consolidateAfter` (which control Karpenter’s consolidation logic) must not be used when the node count is fixed via `replicas`.
+
+#### Validation rules
+
+```yaml
+if replicas != nil:
+  - consolidationPolicy must be nil or default "WhenEmptyOrUnderutilized" or "Never"
```
ellistarn: Will we default this to "Never" for users? Or we can just leave this empty and let the controller decide based off of static/dynamic.

jmdeal (commented):
Checkpointing review

designs/static-capacity.md
sumukha-radhakrishna: will reword

designs/static-capacity.md
sumukha-radhakrishna: With the existing types for disruption.consolidateAfter and disruption.consolidationPolicy it is hard to determine if user explicity set the field to the default or it was defaulted so we decided to take in defaults and reject values that do not make sense (for eg consolidationPolicy can be a default value `WhenEmptyOrUnderutilized`, No value or `Never`) regardless of values we ignore them

designs/static-capacity.md
sumukha-radhakrishna: I think kubebuilder cannot conditionally add defaults based on the presence of another field. Right now we already default to `WhenEmptyOrUnderutilized` changing this would cause regression so we accept empty or default value or value that makes sense (Never)

designs/static-capacity.md
sumukha-radhakrishna: More like `Workloads that require predictable, always-available capacity` let me reword

designs/static-capacity.md
sumukha-radhakrishna: Some fixed node cases I can think of
* Capacity reservation/budgeting and infrastructure boundary
* Isolation (although this could be achieved with pod affinity/anti-affinity)
 
let me rephrase `Operational models that rely on a fixed number of nodes for budgeting, isolation, or infrastructure boundary`

designs/static-capacity.md
sumukha-radhakrishna: agree

designs/static-capacity.md
sumukha-radhakrishna: We will do forceful termination. Since this is a user driven action we will not wait until the node is drained and all volume attachments are deleted. If we do a graceful termination then nodes could be stuck in terminating causing us to have more nodes

designs/static-capacity.md
sumukha-radhakrishna: oops.

designs/static-capacity.md
sumukha-radhakrishna: Makes sense, just wanted to explicitly call-out that scaling down replicas will be forceful and wont respect disruption budgets.

designs/static-capacity.md
sumukha-radhakrishna: Good point. Yeah changing the Nodepool from dynamic to static would cause unintended consequences. If there are more Nodeclaims exist for a Nodepool that replicase after changing it to static, we might start terminating NodeClaims. Similarly there might be changes in consolidation behavior. However CEL validation wont work as it does not have access to old object, ideally we should not be accepting the change. Webhook perhaps?

designs/static-capacity.md
sumukha-radhakrishna: yes, you're right. the actual instance type selection happens during the cloud provider's Create() call in both static and dynamic provisioning. The distinction I was referring to is about when and how the NodeClaim requirements are determined:
*  In dynamic provisioning, we generate a new set of requirements for each provisioning event, based on the unschedulable pods at that time depending on needs.
*  In static provisioning, the NodeClaim requirements are directly derived from the NodeClaimTemplate on the NodePool. These are evaluated once per NodeClaim at creation, meaning the selection is based solely on what the template allows.

As a result, even though all NodeClaims come from the same static NodePool, they may still result in different instance types (shapes/flavors), depending on availability, since that decision happens during the Create() call.

designs/static-capacity.md
ellistarn: It could be as simple as just ignoring consolidationPolicy, regardless of what it's set to.

designs/static-capacity.md
jmdeal: Can we give an example of a type of workload that would be? The main reason I could see this being desired is application startup latency, so I'm trying to understand what distinguishes this item from item 1

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+- This selection is done once at provisioning time (unlike dynamic pools, where evaluation occurs per provisioning event)
+
+In static provisioning, the NodeClaim requirements are directly derived from the NodeClaimTemplate on the NodePool. These are evaluated once per NodeClaim at creation, meaning the selection is based solely on what the template allows.
+As a result, even though all NodeClaims come from the same static NodePool, they may still result in different instance types (shapes/flavors), depending on availability, since that decision happens during cloud-provider Create() call.
+
+To ensure high availability and fault tolerance, we automatically spread static nodes across AZs when the zone requirement includes multiple values.
+```yaml
+- key: topology.kubernetes.io/zone
+  operator: In
+  values: ["us-west-2a", "us-west-2b", "us-west-2c"]
```
ellistarn: If these aren't constrained by requirements, should we still balance from the zones discovered in offerings? I would expect so, but worth thinking about.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+  limits:
+    # Constrain the size of static Nodepool 
+    nodes: 20 # Karpenter will not scale up nodes or create replacements if limits are breaching
+```
+
+We will support scaling static node pools using a command such as:
+
+```sh
+kubectl scale nodepool <name> --replicas=10
+```
```
jmdeal: How would this work? Would we need to write a kubectl plugin or is this automatically supported for any resource that has a `spec.replicas` field? I don't think we want to start maintaining a plugin if that's required, I'd leave that to the community to implement if they see value.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+
+#### Scale-Up Controller
+- **Purpose**: Scale up nodeclaims to meet desired replicas
+- **Behavior**: Monitors the difference between desired replicas and current healthy NodeClaims, provisioning new nodes as needed respecting limits.nodes value
+- **Instance Selection**: Will find the cheapest instance possible during scale up operations based on the requirements
+- **Consolidation**: Static NodePools will not be consolidated to maintain predictable capacity
+
+#### Deprovisioning Controller  
+- **Purpose**: Scale down replicas when desired count is reduced
+- **Node Selection**: Will prioritize empty nodes for termination. In future iterations, we will add intelligent algorithms for selecting which nodes to scale down
```
jmdeal: I take it if there aren't non-empty nodes it will be random? Let's be explicit about the proposed selection logic.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+- **Termination Behavior**: Will forcefully terminate nodes to meet desired size without respecting PodDisruptionBudgets or terminationGracePeriod
```
jmdeal: My understanding from our discussion offline was we weren't forcibly terminating nodes, we were forcibly disrupting nodes - is that not correct? I'd be hesitant to do the former by default.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+
+**Important Note**: The scale-down of replicas represents scaling down capacity, not disruption. We don't want to end up in a situation where a user sets the replicas to a lower number and we would still be running significantly more capacity due to blocking pods, PDBs, and terminationGracePeriod constraints.
```
jmdeal: I disagree with this - by updating replicas you are setting a desired state. Karpenter should use it's existing termination mechanisms to get into that desired state. Customers have levers to pull if they want to ensure quick disruption, if they need to pull those levers they can. Making the behavior diverge between static and dynamic nodepools is surprising to me IMO.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+## Implementation Details
+
+### Controller Architecture
+
+We will be creating Static Capacity by adding 2 new controllers under a feature flag. This will be added to v1alpha1 release to get feedback from users.
+
+#### Scale-Up Controller
+- **Purpose**: Scale up nodeclaims to meet desired replicas
+- **Behavior**: Monitors the difference between desired replicas and current healthy NodeClaims, provisioning new nodes as needed respecting limits.nodes value
+- **Instance Selection**: Will find the cheapest instance possible during scale up operations based on the requirements
```
jmdeal: I don't think this controller will actually be responsible for instance type selection, will it? Is there any reason to not just create `n` NodeClaims from the NodePool's NodeClaimTemplate directly (with the required changes to support skew)?

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+- Lessen the churn and possibility of over-provisioning
+- Be accounted for during disruption/consolidation operations
+- Be considered during new dynamic capacity provisioning decisions
+
+This ensures that the scheduler is aware of both existing and planned static capacity when making provisioning decisions.
+
+### Testing Strategy
+
+For testing, we will add comprehensive integration tests to ensure the feature works correctly across different scenarios:
+- Scale-up operations with various instance types and requirements
```
jmdeal: If we're just creating standard NodeClaims I don't think there's any specific testing we need for static capacity. The NodeClaim lifecycle controller's behavior shouldn't differ for NodeClaims created by a dynamic or static nodepool. I think the only thing we really need to verify is that the created nodeclaims adhere to the topology skew requirements.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+- Scale-down operations with different node selection strategies  
+- Interaction between static and dynamic NodePools
```
jmdeal: What is this referring to? Just that we're considering the in-flight nodes that were created from our scheduling simulation?

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+
+We already have metrics from controller runtime that will provide observability into the static capacity controllers' performance and behavior.
```
jmdeal: I think we could expand more upon potential observability improvements? Do we want to add any status fields specific to static capacity? Are there any additional metrics we'd want to have (e.g. current replicas, desired replicas, etc)?

designs/static-capacity.md
jmdeal: +1 - I assume we'll spread across all available zones which, by default, is the following:
```yaml
  operator: Exists
```

We should define explicitly though if this is going to be a soft or hard constraint. Will we fail to provision if we can't satisfy the constraint or will we fallback? If we failed to maintain the spread at provisioning time, will we attempt to rebalance later?

designs/static-capacity.md
jmdeal: > the NodeClaim requirements are directly derived from the NodeClaimTemplate on the NodePool

I would word this bullet like this then, this is a lot less confusing to me.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+If has(self.replicas) != has(oldSelf.replicas)
+  => Validation Error
+```
+### Disruption
+
+Karpenter already has Disruption for nodes we can use the same mechanism. 
+Since Static Nodepool and NodeClaim is a variant of existing Nodepool/NodeClaim it will inherit Karpenter dynamic provider integration. Drift Detection can also be inherited to trigger replacement of drifted nodes. During this Karpenter will respect the Disruption Budget. If limits.nodes are specified then Karpenter would wait until Nodes are terminated and count is well within node limits before creating replacements.
+
+Replacement will be one-for-one i.e create replacements before terminating the old one to honor replicas if Nodepool node count do not exceed the node limits. 
+When a user scales down a static NodePool, Karpenter will forcefully terminate the corresponding NodeClaims/instances.
```
jmdeal: I had some other comment regarding this, but I disagree with this being the behavior. Customer's have levers to ensure that nodes are terminated in a set period of time, they can continue to leverage them with static NodePools. I don't see any reason to create this API divergence.

designs/static-capacity.md
```diff
@@ -0,0 +1,202 @@
...
+4. Nodes are spread based on the topology requirements
+5. Inherit existing Karpenter features
+6. Use existing NodeClaim infrastructure with owner references
+  
+
+### Modeling & Validation
+
+When `replicas` is specified, the `NodePool` enters a static provisioning mode where certain disruption-related fields become irrelevant or misleading. Specifically:
+
+- `limits.cpu` and `limits.memory` (used to influence consolidation and node selection) are not meaningful for static capacity.
```
jmdeal: We had discussed offline continuing to include these values since there isn't a good reason to introduce divergence between the two APIs

designs/static-capacity.md
sumukha-radhakrishna: Correct me if my understanding is wrong, with topology constrained in requirements we balance nodes across AZs as we keep track of count to avoid piling up on nodes in an AZ.
Without topology requirement, they will go to the cheapest without guaranteeing the spread. So if AZ balance is required then users need to explicitly define the constraint. There is no fallback logic as per the current implementation, best effort provisioning. I thought of exploring `nodeTopologySpread` as a follow-up as we might want to have a way to split capacity/nodes unevenly based on some weight.

designs/static-capacity.md
sumukha-radhakrishna: We get scale subresource for free in kubebuilder 
https://book.kubebuilder.io/reference/generating-crd.html#scale

designs/static-capacity.md
sumukha-radhakrishna: yeah I can add that clarification

designs/static-capacity.md
sumukha-radhakrishna: The more I think about it, more I feel like we should be force deleting nodes since users showed explicit scale-down intent.

designs/static-capacity.md
sumukha-radhakrishna: Continuing the discussion here - https://github.com/kubernetes-sigs/karpenter/pull/2309#discussion_r2255583201

designs/static-capacity.md
sumukha-radhakrishna: Yes, if we are considering in-flight Static NodeClaims. We already have UT covering these.

designs/static-capacity.md
sumukha-radhakrishna: We already have Spec.replicas with desiredCount and status.Nodes with current count. Imo we can keep the existing status infra now and add additional if needed. Yeah I can look into emitting desired and current count per NodePool

designs/static-capacity.md
sumukha-radhakrishna: For static NodePools having limits.cpu and limits.memory along with a fixed replicas count introduces complexity and potential confusion. 
Supporting both would require solving a bin‑packing problem to choose instance shapes that satisfies the replica count and  CPU/memory limits, which can be unsatisfiable in many scenarios (e.g., smallest shape already exceeds a limit). This also raises questions on precedence,  should we reduce replicas to meet limits, or choose smaller shapes and risk under-provisioning nodes? By ignoring these limits for now, static capacity remains predictable, the replicas field alone determines the exact number of nodes, with shapes chosen solely from the requirements. We can revisit limits in the future

designs/static-capacity.md
sumukha-radhakrishna: Okay, I will change this to be graceful termination since PDBs represent agreement to ensure availability during changes. This is the behavior across Kubernetes, even node lifecycle manager respects PDBS when the node is unhealthy/unready. 
Was debating myself wheather we should force kill due to explicit intent to scale down or eventually and gracefully move to desired state. We can in-future give a way to force scale-down so that we dont assume force kill intent by default.

designs/static-capacity.md
sumukha-radhakrishna: Here is the convo - https://github.com/kubernetes-sigs/karpenter/pull/2309#discussion_r2257573765
We will go with graceful termination and in-future give the ability to force delete with a setting.

designs/static-capacity.md
jmdeal: > Correct me if my understanding is wrong, with topology constrained in requirements we balance nodes across AZs as we keep track of count to avoid piling up on nodes in an AZ.

Are you asking if this is existing behavior for dynamic nodepools? This isn't correct, unless pods inject zonal requirements into the NodeClaim (e.g. via selectors, affinity, or TSC) all of the available zones are included on the NodeClaim template. Karpenter does not have any default AZ balance behavior - that needs to be enforced by application requirements.

> Without topology requirement, they will go to the cheapest without guaranteeing the spread. So if AZ balance is required then users need to explicitly define the constraint.

This is surprising to me, mainly because it gives customers no way to say "spread across all of my available AZs" which I expect will be what most customer want. On the other hand, your approach gives customers the option to say they don't care about spread, which I also expect will be the desired behavior for things like training jobs where the ability to get capacity is more important than maintaining zonal spread. 

Long-term, I expect we'll want some way to configure this explicitly. Either a new field like `nodeTopologySpread` like you've suggested, or I could see us exploring an extension to the existing requirements field:
```yaml
requirements:
  operator: Exists
  spreadPolicy: None | Preferred | Strict
```

If we were to include something like that though, we would also want to include support for dynamic nodepools to avoid behavior skew (FWIW there is demand for that). I'm on the fence about if we should block the initial version of this feature on explicit configuration of skew - on one hand I'm pretty convinced that this is not the long-term behavior we'll want, but on the other hand with the feature in alpha we can collect more definitive data. I'd like to hear others opinion on this.

If we do continue to go forward with the implicit spread behavior, I'd suggest the following:
- If a nodepool does not define a requirement for zone, we don't enforce any spread
- If a nodepool does define it with an operator, we spread against the zones that operator selects (e.g. Exists would be all zones, NotIn would be all available zones but those specified, etc)

Whatever we go forward with, we should define it in the RFC explicitly.

designs/static-capacity.md
jmdeal: > We can in-future give a way to force scale-down so that we dont assume force kill intent by default.

FWIW this exists today - you can set an annotation on the NodeClaim to set a forced termination time. This isn't documented since it's an implementation details of TerminationGracePeriod, but it wouldn't take much to flesh it out and make it user facing. This is what NodeRepair already uses today.

designs/static-capacity.md
jmdeal: > Supporting both would require solving a bin‑packing problem to choose instance shapes that satisfies the replica count and CPU/memory limits

I think this is far less complicated than you're making it out to be, and its already a solved problem in the dynamic provisioning controller. 

> This also raises questions on precedence, should we reduce replicas to meet limits, or choose smaller shapes and risk under-provisioning nodes?

I also don't think this is a complicated question to answer - we follow the same existing behavior as the existing controller. We don't need to try and ensure that the replica count is met, we continue to create NodeClaims and then start constraining the compatible instances once we start to reach limits. If the customers configuration prevents the desired replica count from being reached, that's expected behavior and we should surface to the customer via logs / events that provisioning is blocked by limits. 

I would agree that we wouldn't need to support this if we were creating a standalone CRD, but am against creating behavior skew within the same CRD except when absolutely necessary. I'm also not bought in on not supporting consolidation, but given it would be quite a bit more complicated and were committing to support it in the future I'm more willing to punt on it.

designs/static-capacity.md
sumukha-radhakrishna: Synced up offline. 
We will ensure that NodePool topology is tracked and used to inject appropriate zone-level requirements when creating NodeClaims. If customers want their nodes to be spread across all zones, they can explicitly list those zones in the requirements field. If they provide only one value, all NodeClaims will land in that specific zone. When no topology requirement is provided, we pick zones randomly in the cloud provider’s Create call. These are hard constraints, if capacity isn’t available in a specified zone, we won’t fallback to others, and provisioning will fail. We will explore a nodeSpreadConstraints field to give customers more control over zonal distribution, similar to how podTopologySpreadConstraints work.

designs/static-capacity.md
sumukha-radhakrishna: Since limits.cpu and limits.memory are already supported in the existing NodePool CRD, it makes sense to support them in static provisioning to avoid behavioral skew. 
Initially, I was concerned about the interaction between replicas and limits, especially when requirements permit multiple instance shapes. In those cases, we could see a situation where we successfully create all replicas with smaller instance types at first, but fail to do so later if only larger shapes are available, causing us to breach the limits. However, I agree that this is acceptable and consistent with dynamic provisioning behavior, where limits are hard constraints and replica count is best-effort. We can surface the blocking condition via logs and events if limits prevent further provisioning, just as we do today.

However, since limits would influence if we can create further NodeClaims/nodes during provisioning time by checking if we already exceeded limits so there is a possibility of us over-provisioning violating limits.
Also we might have to limit NodeClaims provisioned per reconcile loop or take in-flight NodeClaims into consideration, right now we create n nodeclaims at once since those nodes are not launched and registered, they are not considered during limits calculation. Either we need to do the calculation during provisioning time and take into account in-flight capacity or we let the current check as is and let it over-provision.

designs/static-capacity.md
sumukha-radhakrishna: After discussing offline. Having an implicit spread would diverge from existing NodePool requirements where it means to constrain and lack of it is unconstrained, we dont want to treat it any different for Static Nodepools. 
We plan to have nodeTopologySpread as fast follow and we already have a way for customers to spread across AZs by creating multiple nodepools each targeting a single AZ.

designs/static-capacity.md
jmdeal: > Also we might have to limit NodeClaims provisioned per reconcile loop or take in-flight NodeClaims into consideration, right now we create n nodeclaims at once since those nodes are not launched and registered, they are not considered during limits calculation. Either we need to do the calculation during provisioning time and take into account in-flight capacity or we let the current check as is and let it over-provision.

This is what we do in dynamic provisioning, we use roughly the following algorithhm:
- Create a NodeClaim template based on the NodePool
- Filter out instance types which would exceed limits from the NodeClaim template
- If the remaining set of instance types has len 0, fail to create the NodeClaim

In dynamic provision once we fail we would fall back to other, lower-weighted NodePools but here we would just fail for this provisioning loop.

sumukha-radhakrishna: While dynamic provisioning does filter NodeClaim templates based on limits [(code),](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/provisioning/scheduling/scheduler.go#L558) it operates under the assumption that pods are batched and packed efficiently. This is fundamentally different from static provisioning, where we attempt to create a fixed number of NodeClaims (replicas), and each is intended to represent an individual node.

If we were to apply limit filtering during static provisioning in the same way, we’d need to track and account for in-flight NodeClaims, since limits would affect not only what we can launch now, but also what we’ve already committed to launching. This would mean either:
* Calculating resource consumption based on the largest possible instance type to conservatively stay within limits, or
* Reconciling one NodeClaim per loop to give enough time for the fleet to allocate capacity and update accounting before provisioning the next.

In effect, limits would start to dictate instance shape, not just aggregate capacity. This creates a conflict between honoring replicas vs. satisfying hard limits.
Static provisioning and dynamic provisioning behave differently under limits constraints. I dont think its as simple as reusing the same filtering logic.

designs/static-capacity.md
ellistarn: > we already have a way for customers to spread across AZs by creating multiple nodepools each targeting a single AZ.

Wow I didn't think of this -- maybe this is a generic solve for static capacity

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+## Proposal
+Extend the existing NodePool resource to support static provisioning capabilities by adding new fields:
+
+```yaml
+spec:
+  # Makes NodePool static when specified
+  replicas: 5  # Number of nodes to maintain
+limits:
+  # Prevents Karpenter from creating new replicas once the limit is exceeded
+  nodes: 10 # Constrain the size of static Nodepool
```
jmdeal: We should add the status update here as well

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+limits:
+  # Prevents Karpenter from creating new replicas once the limit is exceeded
+  nodes: 10 # Constrain the size of static Nodepool 
+```
+The existence of replicas field will determine if NodePool is Static
+
+Key aspects:
+1. Static NodePools maintain fixed node count regardless of pod demand
+2. Will not be considered as consolidation candidate, so consolidation settings are not allowed other than defaults (at least in v1alpha1)
+3. Optional limits.nodes can be set to constrain the size of static Nodepool
```
jmdeal:
```suggestion
3. Optional limits.nodes can be set to constrain the maximum size of static Nodepool when bursting (e.g. during drift or expiration)
```

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+4. Nodes are spread based on the topology requirements
```
jmdeal: This is no longer true, right?

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+if replicas != nil:
+  - limits other than limits.nodes must not be set
+  => ValidationError if not
+```
+
+### Static NodePool limits
+
+We will not be supporting limits other than limits.node, while dynamic provisioning does filter NodeClaim templates based on limits (e.g., limits.cpu, limits.memory), it operates under the assumption that pods are batched and packed efficiently into a single NodeClaim. This is fundamentally different from static provisioning, where we attempt to create a fixed number of NodeClaims (replicas). Limits would start to dictate instance shape, not just aggregate capacity. This creates a conflict between honoring replicas vs. satisfying hard limits.
+
+However, we are introducing a new field, limits.nodes, to explicitly cap the maximum number of nodes a NodePool can provision.
```
jmdeal:
```suggestion
However, we are introducing a new field, limits.nodes, to explicitly cap the maximum number of nodes which can be provisioned for a given NodePool at a point in time.
```

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+
+### Disruption
+
+Karpenter already has Disruption for nodes we can use the same mechanism. 
+Since Static Nodepool and NodeClaim is a variant of existing Nodepool/NodeClaim it will inherit Karpenter dynamic provider integration. Drift Detection can also be inherited to trigger replacement of drifted nodes. During this Karpenter will respect the Disruption Budget. If limits.nodes are specified then Karpenter would wait until Nodes are terminated and count is well within node limits before creating replacements.
+
+Replacement will be one-for-one i.e create replacements before terminating the old one to honor replicas if Nodepool node count do not exceed the node limits. 
+When a user scales down a static NodePool, Karpenter will delete NodeClaims and gracefully terminate nodes.
+Key behaviors:
+- User-driven actions (scaling replicas) will not respect disruption budgets. The user explicitly intends to remove capacity, and we honor that request by deleting corresponding NodeClaims. However it will respect PDBs and terminationGracePeriod.
```
jmdeal: We should distinguish between termination and disruption. The disruption itself doesn't respect PDBs - it's only the drain / termination process that does. This is implied by the fact it's an automated forceful disruption method.
```suggestion
- User-driven actions (scaling replicas) will not respect disruption budgets. The user explicitly intends to remove capacity, and we honor that request by deleting corresponding NodeClaims. This will be classified as an [automated forceful disruption method](https://karpenter.sh/docs/concepts/disruption/#automated-forceful-methods), a la expiration.
```

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+- Supported Karpenter-driven actions (e.g., drift) respect disruption budgets and scheduling safety.
+
+Deletion of nodepool would delete all its NodeClaims. Garbage collection logic do not differ between static and dynamic NodePools.
```
jmdeal:
```suggestion
Deletion of nodepool would disrupt all of its NodeClaims. Garbage collection logic do not differ between static and dynamic NodePools.
```

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+
+**Important Note**: The scale-down of replicas in a static NodePool will follow standard graceful disruption semantics. This means that when the desired replica count is reduced, Karpenter will drain nodes and while respecting PodDisruptionBudgets (PDBs), terminationGracePeriod, and other workload disruption safeguards. We will not bypass these constraints, as PDBs represent an explicit agreement between the cluster operator and the controller to maintain application availability during changes. This aligns with Kubernetes conventions, where scaling actions are reconciled towards the desired state without ignoring safety guarantees, and ensures predictable behavior for workloads. While this may mean that scale-down is delayed when eviction blockers are present, it provides consistency with the broader Kubernetes ecosystem.
```
jmdeal: Graceful disruption refers to mechanisms like drift and consolidation which ensure PDBs and other constraints are respected **before** initiating the disruption action. This is a forceful disruption. 
```suggestion
**Important Note**: The scale-down of replicas in a static NodePool will follow standard forceful disruption semantics. This means that when the desired replica count is reduced, Karpenter will drain nodes and while respecting PodDisruptionBudgets (PDBs), terminationGracePeriod, and other workload disruption safeguards. We will not bypass these constraints, as PDBs represent an explicit agreement between the cluster operator and the controller to maintain application availability during changes. This aligns with Kubernetes conventions, where scaling actions are reconciled towards the desired state without ignoring safety guarantees, and ensures predictable behavior for workloads. While this may mean that scale-down is delayed when eviction blockers are present, it provides consistency with the broader Kubernetes ecosystem.
```

designs/static-capacity.md
```diff
@@ -0,0 +1,236 @@
...
+  operator: In
+  values: ["zone-2a", "zone-2b", "zone-2c"]
+```
+
+However, this introduces implicit behavior that conflicts with existing Karpenter semantics. In current behavior, absence of a requirement is treated as unconstrained, and specifying zones in requirements is interpreted as a hard constraint. Introducing automatic spreading implicitly redefines these semantics, which could have long-term consequences for other requirement types (e.g., capacity type).
+Given these concerns, we’re punting zonal spreading to a future enhancement. For now, we will continue to treat the requirements field as the single source of truth for topology decisions. Users who want to spread nodes across zones can do so explicitly by:
+- Specifying multiple zones in the topology.kubernetes.io/zone requirement, or
+- Creating multiple static NodePools, each pinned to a specific AZ.
+
+We plan to explore a nodeTopologySpread field in the future to offer users a more flexible and declarative way to express zone-spreading behavior, similar to podTopologySpreadConstraints.
```
jmdeal: I wouldn't prescribe a specific field, I'd just say we're going to explore a method to decoratively express intention to spread nodes across zones.

PR #2310, opened 2025-06-14
feat: Introduce NodePool repair policies with toleration durations

matanbaruch:
This pull request introduces a new feature that allows users to configure custom node repair toleration durations on a per-NodePool basis. Previously, node repair timing was dictated solely by CloudProvider defaults, but now users can override these defaults for specific node conditions or set a NodePool-wide default. This enables more granular control to match the needs of different workloads (e.g., stateful vs. stateless applications). The changes include updates to the NodePool API, CRDs, deep copy methods, documentation, and controller logic.

**NodePool Repair Configuration Feature:**

* Added a new `repair` block to the `NodePoolSpec` API, allowing users to specify custom repair policies and default toleration durations for node conditions. This includes the `RepairSpec` and `NodePoolRepairPolicy` types. [[1]](diffhunk://#diff-c1b722216cedf83c042b577c91a3d890a216d93f0fd5a85909683ae3b3365bd6R69-R74) [[2]](diffhunk://#diff-c1b722216cedf83c042b577c91a3d890a216d93f0fd5a85909683ae3b3365bd6R169-R208)
* Updated the CRDs (`karpenter.sh_nodepools.yaml`) to support the new `repair` configuration, including validation patterns and documentation for the new fields. [[1]](diffhunk://#diff-542ac25ed1857604758c133b5f49049a6ffac641ddd6202c3a34cb528f196a6eR175-R219) [[2]](diffhunk://#diff-4e0265201c4542984a38dca2ae7a5311bc50df00e0bd483148222b92726bfa37R175-R219)
* Implemented deep copy methods for the new `RepairSpec` and `NodePoolRepairPolicy` types to ensure correct runtime behavior. [[1]](diffhunk://#diff-f0bff30e893c06fbb3d8d69fd03e697b2b0c9a3e8d044accee7338c6ce202300R412-R427) [[2]](diffhunk://#diff-f0bff30e893c06fbb3d8d69fd03e697b2b0c9a3e8d044accee7338c6ce202300R445-R449) [[3]](diffhunk://#diff-f0bff30e893c06fbb3d8d69fd03e697b2b0c9a3e8d044accee7338c6ce202300R555-R579)

**Controller and Documentation Updates:**

* Modified the node health controller to use NodePool-specific repair toleration durations when available, falling back to CloudProvider defaults as needed.
* Added a new design document (`nodepool-repair-configuration.md`) detailing the motivation, API changes, usage examples, and design decisions for the NodePool repair configuration feature. Also updated the main node repair design doc to reference this new capability. [[1]](diffhunk://#diff-4a76e58e2289ae67e3c6ffb449739c423d8d075e2b81f6274a80c410d050cea5R1-R220) [[2]](diffhunk://#diff-997c287391c6ff3ac98a6fada129945da32e2c6de6d9c225352315b207527aeaR129)


designs/node-repair.md
```diff
@@ -18,99 +18,145 @@ The alpha implementation will not consider these features:
 
   - Disruption Budgets
   - Customer-Defined Conditions
-  - Customer-Defined Remediation Time
 
 The team does not have enough data to determine the right level of configuration that users will utilize. The opinionated mechanism would be responsible for defining unhealthy notes. The advantage of creating the mechanism would be to reduce the configuration burden for customers. **The feature will be gated under an alpha NodeRepair=true feature flag. This will  allow for additional feedback from customers. Additional feedback can support features that were originally considered out of scope for the Alpha stage.**
 
-## Recommendation: CloudProvider-Defined RepairPolicies
+## Recommendation: CloudProvider-Defined RepairPolicies with NodePool-Configurable TolerationDuration
```
engedaam: Can you write a separate section in this doc or move the changes you want to make to a separate design? It will make it hard to track in the future all the design choice and consideration we have made

designs/node-repair.md
```diff
@@ -18,99 +18,145 @@ The alpha implementation will not consider these features:
...
+    ConditionType metav1.ConditionType
+    // Status specifies the condition status that indicates unhealthy state
+    Status metav1.ConditionStatus
+    // TolerationDuration is the duration to wait before attempting to terminate 
+    // nodes that match this repair policy. If not specified, defaults to the 
+    // cloud provider's recommended duration for this condition type.
+    TolerationDuration *metav1.Duration
+}
+
+type NodePoolSpec struct {
```
engedaam: Can you describe your use case as to why we need to expose all of these details to users? Also, are you expecting cloudprovider to give default repair policies?

engedaam: Also should this be as part of the disruption stanza?

designs/node-repair.md
```diff
@@ -18,99 +18,145 @@ The alpha implementation will not consider these features:
...
+
+type RepairSpec struct {
+    // Policies defines the repair policies for different node conditions.
+    // These policies override the default tolerationDuration from the CloudProvider.
+    Policies []RepairPolicy
+    
+    // DefaultTolerationDuration is the default duration to wait before repairing
+    // nodes for any condition not explicitly configured in Policies.
+    // If not specified, uses the CloudProvider's default durations.
+    DefaultTolerationDuration *metav1.Duration
```
engedaam: Why define this if we give all the toleration duration for every policy? How will the interaction between customer defined polices?

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,29 @@
+{
```
engedaam: What's the purpose of the devcontonatiners? Why do we need this to merged into the project?

matanbaruch: I had difficulties setting up the environment for testing (golangci-lint version, go versioned installed)

Dev containers is the most efficient way to setup an environment regardless the developer environment.
Also support all of the IDE's.

https://code.visualstudio.com/docs/devcontainers/containers

designs/node-repair.md
matanbaruch: Done

designs/node-repair.md
matanbaruch: Done

designs/node-repair.md
matanbaruch: Fixed.

matanbaruch:
@engedaam Hey can you run tests please?

designs/node-repair.md
```diff
@@ -22,97 +22,176 @@ The alpha implementation will not consider these features:
```
engedaam, line 22: @matanbaruch instead of changing the node repair doc, can you separate out your use cases and proposal into their own section. It might be better to write this as it's own RFC, since this a pretty major change to the NodePool API. Main reason here is that I would good for us to keep the historical context for the choices that were made with the feature.

.devcontainer/devcontainer.json
engedaam: It might be better to separate out this change in a different PR

designs/node-repair.md
matanbaruch: You are right. Fixed

matanbaruch:
Resolve https://github.com/aws/karpenter-provider-aws/issues/7491

designs/node-repair.md
```diff
@@ -22,97 +22,11 @@ The alpha implementation will not consider these features:
 
 The team does not have enough data to determine the right level of configuration that users will utilize. The opinionated mechanism would be responsible for defining unhealthy notes. The advantage of creating the mechanism would be to reduce the configuration burden for customers. **The feature will be gated under an alpha NodeRepair=true feature flag. This will  allow for additional feedback from customers. Additional feedback can support features that were originally considered out of scope for the Alpha stage.**
 
-## Recommendation: CloudProvider-Defined RepairPolicies
```
engedaam, line 25: Why is this section being deleted? This is the historical context, we want to preserve, no?

designs/node-repair.md
```diff
@@ -125,5 +39,5 @@ There are additional features we will consider including after the initial itera
 * Disruption controls (budgets, terminationGracePeriod) for unhealthy nodes 
 * Node Reboot (instead of replacement)
 * Configuration surface for graceful vs forceful termination 
-* Additional consideration for the availability zone resiliency 
-
+* Additional consideration for the availability zone resiliency
+* Customer-defined node conditions for repair (beyond CloudProvider-defined conditions)
```
engedaam: Isn't this a duplicate of the first point?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Summary
+
+This RFC proposes adding a new `repair` configuration block to the NodePool API to give users fine-grained control over node repair behavior. This builds upon the existing node repair functionality by allowing per-condition toleration durations and workload-specific repair policies.
+
+## Motivation
+
+While cloud providers can define default repair policies, different workloads have different sensitivities to node issues:
+
+- **Critical, stateful applications** may require longer toleration durations to allow for manual intervention or graceful failover
```
engedaam: You bring up really powerful use cases, and I see the need for the controls. Can you help me understand for cases of stateful application? What would manual intervention look like generally? Is this something that you experienced? Would it make sense for Karpenter to automate these manual actions?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+Add a new `repair` block to the `NodePool` specification:
+
+```go
+type NodePoolSpec struct {
+    // ... existing fields ...
+    Disruption DisruptionSpec `json:"disruption"`
+
+    // Repair configures the repair policies for nodes in this NodePool.
+    // If not specified, repair policies will be disabled for this nodepool.
+    Repair *RepairSpec `json:"repair,omitempty"`
```
engedaam, line 54: Why have this as a separate stanza instead of the disruption spec?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+    // that matches a policy in this list, Karpenter will use the corresponding toleration duration.
+    // This provides fine-grained control over repair timing for different issues.
+    // These policies override the DefaultTolerationDuration.
+	Policies []RepairPolicy `json:"policies,omitempty"`
+
+    // DefaultTolerationDuration is the default duration to wait before repairing a node
+    // for any condition that is not explicitly covered by a policy in the 'policies' list.
+    // This acts as a fallback for any uncategorized or new node conditions.
+    // If not specified, Karpenter will use the CloudProvider's default duration.
+	DefaultTolerationDuration *metav1.Duration `json:"defaultTolerationDuration,omitempty"`
```
engedaam: Why have this field instead of using the cloud provider defaults? I'm not sure I understand the value of setting toleration duration for all policy? Would it not be better to set the policy for specific conditions?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+}
+
+type RepairPolicy struct {
```
engedaam: What about the status for the condition type? In the case for ready conditions, there are three states that the condition can be set `True`, `False`, and `Unknown`. For cases for DiskPressure, the case to repair on would done on the `True` condition?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+    // ... existing fields like consolidation, expireAfter ...
+}
+
+// RepairSpec defines the repair policies for nodes in this NodePool
+type RepairSpec struct {
+    // Policies defines a list of repair policies for specific node conditions. If a node has a condition
+    // that matches a policy in this list, Karpenter will use the corresponding toleration duration.
+    // This provides fine-grained control over repair timing for different issues.
+    // These policies override the DefaultTolerationDuration.
+	Policies []RepairPolicy `json:"policies,omitempty"`
```
engedaam: Are we going to allow users to define any status condition or just the once the Cloud Provider will look to repair?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+1. The `toleration` in a `RepairPolicy` within the NodePool's `repair.policies` for the specific condition type
+2. The NodePool's `repair.defaultTolerationDuration`
+3. CloudProvider's recommended default duration
+
+### Controller Workflow
+
+1. A diagnostic agent adds a status condition on a node
+2. Karpenter reconciles nodes and matches unhealthy conditions with repair statements from the CloudProvider
+3. Karpenter determines the `Toleration` using the `NodePool`'s repair configuration
+4. The Node Health controller forcefully terminates the NodeClaim once the node has been in an unhealthy state for the determined duration
```
engedaam: It might be good to callout that this design is not looking to make node repair a graceful disruption mechanism

.devcontainer/post-create.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
engedaam: Can you pull this out in a separate PR? Not sure it makes sense to have this change included here

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+    // to monitor on the node.
+    RepairPolicy() []RepairStatement
+  ...
+}
+
+type RepairStatement struct {
+    // Type of unhealthy state that is found on the node
+    Type metav1.ConditionType 
+    // Status condition of when a node is unhealthy
+    Status metav1.ConditionStatus
```
engedaam: Don't we still need the toleration duration field for the cloud provider?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+### Future Enhancements
+
+After the initial implementation, we will consider:
+- Disruption controls (budgets, terminationGracePeriod) for unhealthy nodes
+- Node reboot (instead of replacement)
+- Configuration surface for graceful vs. forceful termination
+- Availability zone resiliency considerations
+- Customer-defined node conditions for repair (beyond CloudProvider-defined conditions)
+
+## Alternative Designs Considered
```
engedaam, line 201: Isn't this the current design rather then an alternative?

designs/nodepool-repair-configuration.md
```diff
@@ -0,0 +1,198 @@
...
+
+```go
+type RepairPolicy struct {
+    Type metav1.ConditionType 
+    Status metav1.ConditionStatus
+    TolerationDuration metav1.Duration  // Fixed duration
+}
+```
+
+**Rejected because:** This approach lacks the flexibility users need to tailor repair behavior to different workload requirements. Different NodePools may need different toleration durations for the same condition type.
```
engedaam: It might be good to drop this part or move it at the begin of the doc, so that you can make it clear the motivation of this design

kwok/cloudprovider/cloudprovider.go
```diff
@@ -156,18 +156,16 @@ func (c CloudProvider) GetSupportedNodeClasses() []status.Object {
...
 
-func (c CloudProvider) RepairPolicies() []cloudprovider.RepairPolicy {
-	return []cloudprovider.RepairPolicy{
+func (c CloudProvider) RepairPolicies() []cloudprovider.RepairStatement {
+	return []cloudprovider.RepairStatement{
 		// Supported Kubelet Node Conditions
 		{
-			ConditionType:      corev1.NodeReady,
-			ConditionStatus:    corev1.ConditionFalse,
-			TolerationDuration: 10 * time.Minute,
```
engedaam, line 166: If a user does not set any policies, isn't the `TolerationDuration` provided by the cloud provider, what we need to determine when to repair?

pkg/cloudprovider/types.go
```diff
@@ -49,6 +49,14 @@ var (
 
 type DriftReason string
 
+type RepairStatement struct {
```
engedaam: Why do we need both the notion of a repair statement and repair policy?

pkg/cloudprovider/types.go
```diff
@@ -49,6 +49,14 @@ var (
...
+	// ConditionType of unhealthy state that is found on the node
+	ConditionType corev1.NodeConditionType
+	// ConditionStatus condition when a node is unhealthy
+	ConditionStatus corev1.ConditionStatus
+}
+
+// RepairPolicy is deprecated, use RepairStatement instead
 type RepairPolicy struct {
```
engedaam, line 62: Why can't drop this struct?

engedaam (commented):
I reviewed the design document, and did an initial pass on the code. Before doing a more through review on the code change, let's close on some of the design questions?

engedaam:
@matanbaruch Do you need another round of review or are still making changes?

ricardovj:
Any ideas when this might get released? Love this feature, thanks!

matanbaruch:
I will get working on this.

alikhil:
Hey @matanbaruch, this is a really great proposal. I'd love to see this feature merged 🚀

Is there anything I can help with to get it over the finish line? Happy to jump in!

kwok/cloudprovider/cloudprovider.go
matanbaruch: Fixed. The `TolerationDuration` is now preserved in the CloudProvider's `RepairPolicy` struct (unchanged from upstream). When a user does not set any NodePool repair policies, the controller falls back to the CloudProvider's `TolerationDuration` as the default. The resolution order is:

1. NodePool condition-specific policy override
2. NodePool `defaultTolerationDuration`
3. CloudProvider's `TolerationDuration` (the fallback)

See `resolveTolerationDuration()` in `pkg/controllers/node/health/controller.go` and the "Resolution Order" section in the design doc.

pkg/cloudprovider/types.go
matanbaruch: Fixed. The `RepairStatement` struct has been dropped entirely. Only `RepairPolicy` (the upstream struct with `ConditionType`, `ConditionStatus`, and `TolerationDuration`) remains. The CloudProvider interface is unchanged from upstream — `RepairPolicies() []RepairPolicy`.

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. The section has been restructured. The current design is now the main proposal in the doc. What was previously mislabeled has been moved to "Alternative Designs Considered" at the bottom, which now correctly documents two rejected alternatives:

1. **Placing Repair Under the Disruption Stanza** — rejected because repair is forceful and doesn't respect disruption budgets/TGP/do-not-disrupt
2. **Allowing User-Defined Conditions** — rejected because CloudProvider should be authoritative for which conditions indicate infrastructure failures

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. Added a dedicated section "Why a Separate `repair` Stanza (Not Under `disruption`)" in the design doc that explains:

- **Disruption** is a graceful lifecycle operation (consolidation, drift, expiration) that respects budgets, `terminationGracePeriod`, and `do-not-disrupt` annotations.
- **Repair** is a forceful response to unhealthy nodes that intentionally bypasses disruption controls, because an unhealthy node may be unable to gracefully drain pods or detach volumes.

Placing repair under `disruption` would mislead users into thinking these controls apply to repair operations, which they do not.

pkg/cloudprovider/types.go
matanbaruch: Fixed. `RepairStatement` has been removed. Only `RepairPolicy` exists now — the same struct from upstream with `ConditionType`, `ConditionStatus`, and `TolerationDuration`. No changes to `pkg/cloudprovider/types.go` from upstream.

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. The design doc now explicitly states in the Motivation section: "This design does not change the node repair mechanism to be a graceful disruption. Node repair remains a forceful termination. Unhealthy nodes will not respect customer-configured terminationGracePeriod or disruption budgets."

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. The CloudProvider interface is completely unchanged from upstream — `RepairPolicies() []RepairPolicy` with `TolerationDuration` preserved. The NodePool repair config only overrides the duration, never removes it. The CloudProvider's `TolerationDuration` serves as the final fallback in the resolution order.

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. Added a "Condition Status Handling" section with a table showing the three statuses (True, False, Unknown) and how they map to different condition types (Ready=False/Unknown, DiskPressure=True, etc.). The `ConditionStatus` field in `NodePoolRepairPolicy` must match the CloudProvider's `ConditionStatus` for the override to take effect.

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. The design doc now explicitly states: "Users cannot define new conditions to monitor through NodePool configuration. Only conditions defined by the CloudProvider's RepairPolicies() are monitored. The NodePool configuration only controls the toleration duration for those conditions."

designs/nodepool-repair-configuration.md
matanbaruch: Good point. `DefaultTolerationDuration` is an optional convenience for users who want a single override for all conditions without listing each one individually. The design doc now clarifies the resolution order — condition-specific policies take priority over the default, and the CloudProvider's duration is the final fallback when neither is set.

designs/nodepool-repair-configuration.md
matanbaruch: Fixed. Motivation is now the very first section of the design doc, before the Summary and Proposal sections.

designs/nodepool-repair-configuration.md
matanbaruch: Great question. The Motivation section now covers this — for stateful applications like databases and distributed storage, automated repair triggered too quickly can disrupt ongoing data replication or cause split-brain scenarios. Operators typically need time for manual verification (e.g., checking replication lag, ensuring quorum) before allowing a node to be replaced. The longer toleration duration gives them that window. This was motivated by real-world experience running stateful workloads on Kubernetes.

.devcontainer/post-create.sh
matanbaruch: Fixed. Devcontainer files have been removed from this PR.

designs/node-repair.md
matanbaruch: Fixed. The original `node-repair.md` is preserved with all historical context intact. Only added a link to the new RFC under "Future considerations". The proposal is in a separate `nodepool-repair-configuration.md` RFC document.

designs/node-repair.md
matanbaruch: Fixed. The original node-repair.md is no longer modified (historical context preserved). The new RFC is in a separate document.

matanbaruch:
@engedaam Ready for a re-review

matanbaruch:
@engedaam I think this presubmit is flaky.. 

Can you review it?

erezzarum:
Does this also support the reason behind the condition (event)?

matanbaruch:
Rebased onto latest main, conflicts are resolved and the CLA check is green again.

What changed since the last review:

1. Dropped `defaultTolerationDuration`. Your point in https://github.com/kubernetes-sigs/karpenter/pull/2310#discussion_r2195664282 was right, one duration applied to every policy is the wrong shape. Overrides are explicit per condition now, so a CloudProvider adding a new repair condition later keeps its own duration instead of inheriting an unrelated one.
2. Resolution is down to two steps: a NodePool policy matching `conditionType` and `conditionStatus`, otherwise the CloudProvider's `TolerationDuration`.
3. The CloudProvider interface is still untouched.
4. Squashed to a single commit.

@erezzarum not today. Matching is on `conditionType` and `conditionStatus` only. The condition `Reason` is not part of it, so you can't target a specific reason or the event behind it. If you have a concrete case for that let's do it as a follow up, I'd rather not widen this API further here.

@alikhil thanks for offering. Nothing to split up right now, it's only waiting on review.

@engedaam ready for another look.

PR #2312, opened 2025-06-15
refactor: stage consolidation for overwritable consolidationTTL

rschalo:
Fixes #N/A 

**Description**
Simplify code used in validation and set up consolidation to take in a customizable consolidationTTL which will eventually be used to remove `ExpectToWait` in consolidation tests.

`make presubmit`

### DIFF
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -74,7 +74,7 @@ func NewControllers(
 
 	controllers := []controller.Controller{
 		p, evictionQueue, disruptionQueue,
-		disruption.NewController(clock, kubeClient, p, cloudProvider, recorder, cluster, disruptionQueue),
+		disruption.NewController(disruption.MakeConsolidation(clock, cluster, kubeClient, p, cloudProvider, recorder, disruptionQueue)),
 		provisioning.NewPodController(kubeClient, p, cluster),
 		provisioning.NewNodeController(kubeClient, p),
 		nodepoolhash.NewController(kubeClient, cloudProvider),
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -61,19 +61,33 @@ type consolidation struct {
 	cloudProvider          cloudprovider.CloudProvider
 	recorder               events.Recorder
 	lastConsolidationState time.Time
+	consolidationTTL       time.Duration
+}
+
+type ConsolidationOption func(*consolidation)
+
+func WithConsolidationTTL(ttl time.Duration) ConsolidationOption {
+	return func(c *consolidation) {
+		c.consolidationTTL = ttl
+	}
 }
 
 func MakeConsolidation(clock clock.Clock, cluster *state.Cluster, kubeClient client.Client, provisioner *provisioning.Provisioner,
-	cloudProvider cloudprovider.CloudProvider, recorder events.Recorder, queue *orchestration.Queue) consolidation {
-	return consolidation{
-		queue:         queue,
-		clock:         clock,
-		cluster:       cluster,
-		kubeClient:    kubeClient,
-		provisioner:   provisioner,
-		cloudProvider: cloudProvider,
-		recorder:      recorder,
+	cloudProvider cloudprovider.CloudProvider, recorder events.Recorder, queue *orchestration.Queue, opts ...ConsolidationOption) consolidation {
+	c := &consolidation{
+		queue:            queue,
+		clock:            clock,
+		cluster:          cluster,
+		kubeClient:       kubeClient,
+		provisioner:      provisioner,
+		cloudProvider:    cloudProvider,
+		recorder:         recorder,
+		consolidationTTL: consolidationTTL,
 	}
+	for _, opt := range opts {
+		opt(c)
+	}
+	return *c
 }
 
 // IsConsolidated returns true if nothing has changed since markConsolidated was called.
--- a/pkg/controllers/disruption/drift.go
+++ b/pkg/controllers/disruption/drift.go
@@ -21,33 +21,20 @@ import (
 	"errors"
 	"sort"
 
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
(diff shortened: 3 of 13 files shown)


pkg/controllers/disruption/drift.go
```diff
@@ -21,33 +21,20 @@ import (
...
 
-func NewDrift(kubeClient client.Client, cluster *state.Cluster, provisioner *provisioning.Provisioner, recorder events.Recorder) *Drift {
-	return &Drift{
-		kubeClient:  kubeClient,
-		cluster:     cluster,
-		provisioner: provisioner,
-		recorder:    recorder,
-	}
+func NewDrift(c consolidation) *Drift {
+	return &Drift{consolidation: c}
```
jonathan-innis, line 37: Drift contains consolidation? That doesn't seem right to me. Perhaps we need to change the naming, but I'm also wondering what the need is to encapsulate all of the details of this for drift

pkg/controllers/disruption/validation.go
```diff
@@ -51,48 +45,33 @@ func IsValidationError(err error) bool {
...
+		consolidation:  c,
+		reason:         v1.DisruptionReasonEmpty,
 		filter:         e.ShouldDisrupt,
 		validationType: e.ConsolidationType(),
 	}
 }
 
-func (e *EmptinessValidator) Validate(ctx context.Context, cmd Command, _ time.Duration) (Command, error) {
+func (e *EmptinessValidator) Validate(ctx context.Context, cmd Command) (Command, error) {
+	if err := waitForValidationPeriod(ctx, e.clock, e.consolidationTTL); err != nil {
```
jonathan-innis, line 72: Nice! I'm glad we're properly encapsulating this in the validator now

pkg/controllers/disruption/validation.go
```diff
@@ -310,3 +270,14 @@ func (v *validation) validateCommand(ctx context.Context, cmd Command, candidate
 	// - our lifecycle command says to create a node with types {U_0, U_1, ..., U_n} where U is a subset of T
 	return nil
 }
+
+func waitForValidationPeriod(ctx context.Context, clock clock.Clock, validationPeriod time.Duration) error {
+	if validationPeriod > 0 {
+		select {
+		case <-ctx.Done():
+			return errors.New("context canceled")
```
jonathan-innis, line 278: nit 
```suggestion
			return ctx.Err()
```

pkg/controllers/disruption/consolidation.go
```diff
@@ -61,19 +61,33 @@ type consolidation struct {
...
+	cloudProvider cloudprovider.CloudProvider, recorder events.Recorder, queue *orchestration.Queue, opts ...ConsolidationOption) consolidation {
+	c := &consolidation{
+		queue:            queue,
+		clock:            clock,
+		cluster:          cluster,
+		kubeClient:       kubeClient,
+		provisioner:      provisioner,
+		cloudProvider:    cloudProvider,
+		recorder:         recorder,
+		consolidationTTL: consolidationTTL,
```
jonathan-innis, line 85: This is static, right? So is there a need to define this as a consolidation member variable

pkg/controllers/disruption/consolidation.go
```diff
@@ -61,19 +61,33 @@ type consolidation struct {
 	cloudProvider          cloudprovider.CloudProvider
 	recorder               events.Recorder
 	lastConsolidationState time.Time
+	consolidationTTL       time.Duration
+}
+
+type ConsolidationOption func(*consolidation)
+
+func WithConsolidationTTL(ttl time.Duration) ConsolidationOption {
```
jonathan-innis, line 69: Do you need this or can this just be a property of the validator directly?

pkg/controllers/controllers.go
```diff
@@ -74,7 +74,7 @@ func NewControllers(
 
 	controllers := []controller.Controller{
 		p, evictionQueue, disruptionQueue,
-		disruption.NewController(clock, kubeClient, p, cloudProvider, recorder, cluster, disruptionQueue),
+		disruption.NewController(disruption.MakeConsolidation(clock, cluster, kubeClient, p, cloudProvider, recorder, disruptionQueue)),
```
jonathan-innis, line 77: I'm curious why you want to encapsulate everything inside of consolidation and pass it to the constructor like this. If you are literally wrapping everything in the constructor, it sort of begs the question why you even need to wrap it in the first place

pkg/controllers/disruption/emptiness.go
```diff
@@ -89,16 +88,9 @@ func (e *Emptiness) ComputeCommand(ctx context.Context, disruptionBudgetMapping
...
-	select {
-	case <-ctx.Done():
-		return Command{}, scheduling.Results{}, errors.New("interrupted")
-	case <-e.clock.After(consolidationTTL):
-	}
-
-	validCmd, err := e.Validate(ctx, cmd, consolidationTTL)
+	// Emptiness doesn't need to validate with a scheduling simulation, as we can just take advantage of cluster.IsNodeNominated
+	// whether or not the node has zero reschedulable pods.
+	validCmd, err := e.Validate(ctx, cmd)
```
jonathan-innis, line 93: Really glad we are doing this now 🙌🏼

pkg/controllers/disruption/drift.go
rschalo: No need to encapsulate all of the details, we were passing in members of consolidation and this is looking to unify what is required for a new `Method`. If every `Method` takes in consolidation then it could just use what it needs - I think a change to accept a consolidation as a param but then change back to a struct that isn't just `consolidation` is reasonable.

pkg/controllers/controllers.go
rschalo: It's because when we make the disruption controller in our int tests, there isn't an opportunity to customize validator behavior. We could make `consolidationTTL` a member of `Validator` and export the Validator on each `Method` and then in the tests where relevant overwrite it there.

rschalo: Discussed offline - going to export Methods on the controller and overwrite methods individually as needed.

PR #2313, opened 2025-06-16
feat: Add configurable registration TTL

### LINKED ISSUE #357: Allow registration TTL to be user configurable

GnatorX:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/357

**Description**
Based on https://github.com/kubernetes-sigs/karpenter/pull/1989


pkg/apis/v1/nodeclaim.go
```diff
@@ -74,6 +74,14 @@ type NodeClaimSpec struct {
...
 	// +optional
 	ExpireAfter NillableDuration `json:"expireAfter,omitempty"`
+
+	// RegistrationTTL is the duration the controller will wait
+	// before terminating a node that hasn't registered during TTL time.
+	// If left undefined, the controller will use the default 15m for the node to register.
+	// +kubebuilder:validation:Pattern=`^([0-9]+(s|m|h))+$`
+	// +kubebuilder:validation:Type="string"
+	// +optional
+	RegistrationTTL *metav1.Duration `json:"registrationTTL,omitempty"`
```
engedaam: If we are not planning on drifting on this field, we need to add the `hash:"ignore"`

pkg/apis/v1/nodeclaim.go
```diff
@@ -74,6 +74,14 @@ type NodeClaimSpec struct {
 	// +kubebuilder:validation:Schemaless
 	// +optional
 	ExpireAfter NillableDuration `json:"expireAfter,omitempty"`
+
+	// RegistrationTTL is the duration the controller will wait
+	// before terminating a node that hasn't registered during TTL time.
+	// If left undefined, the controller will use the default 15m for the node to register.
+	// +kubebuilder:validation:Pattern=`^([0-9]+(s|m|h))+$`
+	// +kubebuilder:validation:Type="string"
```
engedaam, line 81: I think we should add the 15 min default as part of this field. That way the CRD field is the single source of truth

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -41,7 +41,7 @@ type Liveness struct {
 
 // registrationTTL is a heuristic time that we expect the node to register within
 // If we don't see the node within this time, then we should delete the NodeClaim and try again
-const registrationTTL = time.Minute * 15
+var registrationTTL = time.Minute * 15
```
engedaam: Consider dropping this variable for the nodeclaim spec value

engedaam (commented):
Few minor cleanup comments

pkg/apis/v1/nodeclaim.go
GnatorX: updated

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -51,9 +51,14 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 	if registered == nil {
 		return reconcile.Result{Requeue: true}, nil
 	}
+	nodeRegistrationTTL := registrationTTL
```
GnatorX: i am keeping the original registrationTTL because if users didn't apply the CRD first the default from the CRD won't be applied by apiserver. This makes sure we keep the 15 minute TTL regardless of is user updated the CRD

GnatorX:
@engedaam do you know why only 1.32.x fails?

pkg/apis/v1/nodepool.go
```diff
@@ -212,6 +212,14 @@ type NodeClaimTemplateSpec struct {
 	// +kubebuilder:validation:Schemaless
 	// +optional
 	ExpireAfter NillableDuration `json:"expireAfter,omitempty"`
+
+	// RegistrationTTL is the duration the controller will wait
+	// before terminating a node that hasn't registered during TTL time.
+	// If left undefined, the controller will use the default 15m for the node to register.
+	// +kubebuilder:validation:Pattern=`^([0-9]+(s|m|h))+$`
+	// +kubebuilder:validation:Type="string"
+	// +optional
```
GnatorX, line 222: i removed 15m default from here because API server would apply the default after the node pool is created causing issues with updating spec after nodepool is created

engedaam: Can you describe this more? The default should be getting add after nodepool creation. What are the issues with updating the nodepool?

pkg/controllers/nodeclaim/lifecycle/liveness.go
engedaam: Would it make sense to just have both then?

pkg/apis/v1/nodepool.go
GnatorX: I might have backed out nodepool changes when the problem is only nodeclaim. Let me double check.

pkg/controllers/nodeclaim/lifecycle/liveness.go
GnatorX: I think it make sense on the nodepool side but not nodeclaim because spec of nodeclaim is immutable and the default would fail to apply (ran into this during testing)

engedaam: Yeah, this makes sense to make the nodepool have the default, so new nodeclaims can have the value added

GnatorX:
https://github.com/kubernetes-sigs/karpenter/pull/2451

GnatorX:
https://github.com/kubernetes-sigs/karpenter/pull/2412

Bryce-Soghigian:
im playing with a poc for coreweave, we boot bare metal machines and have some latency on that where booting typically takes more than 20 minutes. In order to get my provider to work for coreweave I had to fork karpenter core and make that RegistrationTTL a public var that i configure with init() reassignment

vbhargav875:
Are we planning on merging this any time soon?

PR #2316, opened 2025-06-17
fix: update deletionTimestamp on terminating pods when after nodeDeletionTimestamp

cosimomeli:
**Description**
When a node receives the unreachable taint, the Kubernetes taint controller triggers the deletion of all pods after 5 minutes. When the Node Repair threshold is reached, Karpenter's drain procedure waits for all pods to be evicted or to be stuck on termination (when they have passed their deletionTimestamp), but if a Pod has a long termination grace period (RabbitMQ operator pods have 7 days, for example) the node will wait too long before being deleted.

To improve the forced termination, I added the terminating pods with a deletionTimestamp after the nodeTerminationTimestamp to be deleted again, so their deletionTimestamp can be aligned with the nodeTerminationTimestamp.

I added a unit test for this and also tested the change with both an Unhealthy Node on AWS (dead kubelet) and a simple node deletion.

### DIFF
--- a/pkg/controllers/node/termination/terminator/terminator.go
+++ b/pkg/controllers/node/termination/terminator/terminator.go
@@ -99,7 +99,7 @@ func (t *Terminator) Drain(ctx context.Context, node *corev1.Node, nodeGracePeri
 		return fmt.Errorf("listing pods on node, %w", err)
 	}
 	podsToDelete := lo.Filter(pods, func(p *corev1.Pod, _ int) bool {
-		return podutil.IsWaitingEviction(p, t.clock) && (!podutil.IsTerminating(p) || podutil.NeedsForcedEviction(p, nodeGracePeriodExpirationTime))
+		return podutil.IsWaitingEviction(p, t.clock) && (!podutil.IsTerminating(p) || podutil.IsPodEligibleForForcedEviction(p, nodeGracePeriodExpirationTime))
 	})
 	if err := t.DeleteExpiringPods(ctx, podsToDelete, nodeGracePeriodExpirationTime); err != nil {
 		return fmt.Errorf("deleting expiring pods, %w", err)
--- a/pkg/utils/pod/scheduling.go
+++ b/pkg/utils/pod/scheduling.go
@@ -82,10 +82,10 @@ func IsDrainable(pod *corev1.Pod, clk clock.Clock) bool {
 		!IsOwnedByNode(pod)
 }
 
-// NeedsForcedEviction checks if a pod needs to be deleted with a smaller grace period ensuring that the pod:
+// IsPodEligibleForForcedEviction checks if a pod needs to be deleted with a reduced grace period ensuring that the pod:
 // - Is terminating
 // - Has a deletion timestamp that is after the node's grace period expiration time
-func NeedsForcedEviction(pod *corev1.Pod, nodeGracePeriodExpirationTime *time.Time) bool {
+func IsPodEligibleForForcedEviction(pod *corev1.Pod, nodeGracePeriodExpirationTime *time.Time) bool {
 	return nodeGracePeriodExpirationTime != nil &&
 		IsTerminating(pod) &&
 		pod.DeletionTimestamp.Time.After(*nodeGracePeriodExpirationTime)


jonathan-innis:
/assign @engedaam 

Amanuel implemented Node Autorepair so assigning him since he's the relevant owner

chicco785:
hey @engedaam any estimated time for the review? thx!

engedaam:
> When the Node Repair threshold is reached, Karpenter's drain procedure waits for all pods to be evicted or to be stuck on termination (when they have passed their deletionTimestamp), but if a Pod has a long termination grace period (RabbitMQ operator pods have 7 days, for example) the node will wait too long before being deleted.

Currently, Karpenter does not immediately drain pods when initiating a Node Repair action. Instead, it relies on a tolerationDuration configured by the cloud provider. For example, in the AWS Provider, unreachable nodes are given a 30-minute toleration duration before Karpenter begins the process of deleting the node. During this termination period, Karpenter waits for pods to be terminated, which is handled by the drain logic implemented in the terminator.go file (specifically at this line: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/terminator/terminator.go#L140). The behavior you're describing in this PR aligns with our current expectations. To better understand any potential issues, could you provide a specific example where you've observed Karpenter taking longer than the configured toleration duration to terminate an unhealthy node? 

> To improve the forced termination, I added the terminating pods with a deletionTimestamp after the nodeTerminationTimestamp to be deleted again, so their deletionTimestamp can be aligned with the nodeTerminationTimestamp.

Can you help me understand why this would help here? We only really look at the deletionTimestamp for filtering pods, not for case when to force delete

chicco785:
> Currently, Karpenter does not immediately drain pods when initiating a Node Repair action. Instead, it relies on a tolerationDuration configured by the cloud provider. For example, in the AWS Provider, unreachable nodes are given a 30-minute toleration duration before Karpenter begins the process of deleting the node. During this termination period, Karpenter waits for pods to be terminated, which is handled by the drain logic implemented in the terminator.go file (specifically at this line: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/terminator/terminator.go#L140). The behavior you're describing in this PR aligns with our current expectations. To better understand any potential issues, could you provide a specific example where you've observed Karpenter taking longer than the configured toleration duration to terminate an unhealthy node?

As far as I understood, in case a Pod has a long termination grace period the node will not be removed at the end of the node toleration duration, but it will wait for the pod termination grace period. For example, in RabbitMQ operator pods have 7 days termination period, so the node won't be terminated before 7 days.

@cosimomeli can explain better.

cosimomeli:
> To better understand any potential issues, could you provide a specific example where you've observed Karpenter taking longer than the configured toleration duration to terminate an unhealthy node?

Hello @engedaam, thanks for the answer.
When a node becomes unreachable, Karpenter triggers Node Repair after 30 minutes (on AWS). Meanwhile, the Kubernetes taint controller starts evicting pods after 5 minutes when a node receives the `node.kubernetes.io/unreachable` taint.

Karpenter's terminator logic immediately drains every pod in the Node, as the node.health controller sets the node termination timestamp to the current timestamp, this is actually a forced shutdown, and the termination goes as expected, but there is one exception:
The `podsToDelete` [here](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/terminator/terminator.go#L101) filters out every pod that is already terminating, and has not passed its graceful termination period. This means that if I have a pod with a very long termination period (my example was RabbitMQ with 7 days), the termination controller will not touch it, and Karpenter will wait for its natural termination before deleting the node.



> > To improve the forced termination, I added the terminating pods with a deletionTimestamp after the nodeTerminationTimestamp to be deleted again, so their deletionTimestamp can be aligned with the nodeTerminationTimestamp.
> 
> Can you help me understand why this would help here? We only really look at the deletionTimestamp for filtering pods, not for case when to force delete

My change has the effect of adding inside `podsToDelete` the pods with a graceful shutdown period that ends after the deadline of the node termination, this way they get deleted again with a different graceful period, to be compatible with the node termination deadline.

cosimomeli:
Considering that after 30 minutes, excluding the ones with an explicit toleration of `node.kubernetes.io/unreachable`, all the pods are already deleted by the toleration controller as documented [here](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), it's not uncommon that a terminating pod needs another delete to change its default deletion time to speed up the draining process.

engedaam:
> Considering that after 30 minutes, excluding the ones with an explicit toleration of node.kubernetes.io/unreachable, all the pods are already deleted by the toleration controller as documented [here](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/), it's not uncommon that a terminating pod needs another delete to change its default deletion time to speed up the draining process.

This clears things up and thanks for the thorough explanation!

pkg/utils/pod/scheduling.go
```diff
@@ -82,6 +82,15 @@ func IsDrainable(pod *corev1.Pod, clk clock.Clock) bool {
 		!IsOwnedByNode(pod)
 }
 
+// NeedsForcedEviction checks if a pod needs to be deleted with a smaller grace period ensuring that the pod:
+// - Is terminating
+// - Has a deletion timestamp that is after the node's grace period expiration time
+func NeedsForcedEviction(pod *corev1.Pod, nodeGracePeriodExpirationTime *time.Time) bool {
```
engedaam:
```suggestion
// IsPodEligibleForForcedEviction checks if a pod needs to be deleted with a reduced grace period ensuring that the pod:
// - Is terminating
// - Has a deletion timestamp that is after the node's grace period expiration time
func IsPodEligibleForForcedEviction(pod *corev1.Pod, nodeGracePeriodExpirationTime *time.Time) bool {
```

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -99,7 +99,7 @@ func (t *Terminator) Drain(ctx context.Context, node *corev1.Node, nodeGracePeri
 		return fmt.Errorf("listing pods on node, %w", err)
 	}
 	podsToDelete := lo.Filter(pods, func(p *corev1.Pod, _ int) bool {
-		return podutil.IsWaitingEviction(p, t.clock) && !podutil.IsTerminating(p)
+		return podutil.IsWaitingEviction(p, t.clock) && (!podutil.IsTerminating(p) || podutil.NeedsForcedEviction(p, nodeGracePeriodExpirationTime))
```
cosimomeli:
```suggestion
		return podutil.IsWaitingEviction(p, t.clock) && (!podutil.IsTerminating(p) || podutil.IsPodEligibleForForcedEviction(p, nodeGracePeriodExpirationTime))
```

pkg/utils/pod/scheduling.go
cosimomeli: Thanks, I was out of good ideas for the name

PR #2324, opened 2025-06-20
perf: Reduce multiple patch calls in instance termination

jonathan-innis:
Fixes #N/A 

**Description**

This change updates our termination controller logic so that we ensure that we group all of our status condition patch calls together. Up to this point, we were updating them in separate calls, but the requeing that we were doing actually caused a lot of conflicts on large clusters which slowed down our ability to terminate nodes significantly.

This ensures that all patch operations are grouped together to reduce the chance of us hitting conflicts.

### Before PR (9m)

Screenshot 2025-06-23 at 11 55 59 AM

### After PR (3m10s)

Screenshot 2025-06-23 at 11 56 27 AM

- `make presubmit`
- Ran this on a 50,000 KWOK node cluster and compared the termination results


pkg/controllers/node/termination/controller.go
```diff
@@ -131,23 +130,58 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		if !lo.IsEmpty(result) || err != nil {
+			break
+		}
+	}
+	// If we don't have a NodeClaim, then there's nothing for us to patch here
+	if stored != nil && !equality.Semantic.DeepEqual(stored, nodeClaim) {
+		// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
+		// can cause races due to the fact that it fully replaces the list on a change
+		// Here, we are updating the status condition list
+		if err = c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); client.IgnoreNotFound(err) != nil {
```
jonathan-innis, line 153: This simplifies the terminationFunc code a lot so that we don't have to shove all of our checks down in there -- the trade-off is that we don't get a guarantee that this patch actually succeeds when we are tearing down until after the patch call succeeds

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -202,24 +201,34 @@ func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (rec
...
-		isInstanceTerminated, err := terminationutil.EnsureTerminated(ctx, c.kubeClient, nodeClaim, c.cloudProvider)
-		if err != nil {
-			// 404 = the nodeClaim no longer exists
-			if errors.IsNotFound(err) {
-				return reconcile.Result{}, nil
-			}
-			// 409 - The nodeClaim exists, but its status has already been modified
-			if errors.IsConflict(err) {
-				return reconcile.Result{Requeue: true}, nil
+		fmt.Printf("found provider id for nodeclaim %s\n", nodeClaim.Status.ProviderID)
```
DerekFrank: Did you intend to leave this in?

pkg/controllers/nodeclaim/lifecycle/controller.go
```diff
@@ -202,24 +201,34 @@ func (c *Controller) finalize(ctx context.Context, nodeClaim *v1.NodeClaim) (rec
...
+		deleteErr := c.cloudProvider.Delete(ctx, nodeClaim)
+		if cloudprovider.IgnoreNodeClaimNotFoundError(deleteErr) != nil {
+			return reconcile.Result{}, deleteErr
+		}
+		stored := nodeClaim.DeepCopy()
+		nodeClaim.StatusConditions().SetTrue(v1.ConditionTypeInstanceTerminating)
+		if !equality.Semantic.DeepEqual(stored, nodeClaim) {
```
DerekFrank, line 215: I know the azure folks have been looking at these `equality.Semantic.DeepEqual()` calls as a hotspot for CPU time. Do we need to check if the whole object is the same or can we check if just the statuses are not equal?

pkg/controllers/node/termination/controller.go
```diff
@@ -282,24 +262,21 @@ func (c *Controller) awaitVolumeDetachment(
...
+		return reconcile.Result{}, nil
 	}
-	isInstanceTerminated, err := termination.EnsureTerminated(ctx, c.kubeClient, nodeClaim, c.cloudProvider)
-	if client.IgnoreNotFound(err) != nil {
-		// 409 - The nodeClaim exists, but its status has already been modified
-		if errors.IsConflict(err) {
-			return &reconcile.Result{Requeue: true}, nil
-		}
-		return &reconcile.Result{}, fmt.Errorf("ensuring instance termination, %w", err)
+	deleteErr := c.cloudProvider.Delete(ctx, nodeClaim)
```
DerekFrank, line 271: We're deleting the nodeclaim here now?

pkg/controllers/nodeclaim/lifecycle/controller.go
jonathan-innis: > calls as a hotspot for CPU time

In general, these equality checks _do_ impact CPU time but it's super minimal comparatively -- I've seen them pop up in our CPU graphs here and there but most of the time the issue is more about the number of times that we are reconciling in a particular controller and reducing that over removing the deep-copy check -- I'd like to keep this pattern since we use it throughout

pkg/controllers/nodeclaim/lifecycle/controller.go
jonathan-innis: Nope! Thanks for reminding me!

pkg/controllers/node/termination/controller.go
jonathan-innis: Nope! This is deleting the node at the cloudprovider

pkg/controllers/nodeclaim/lifecycle/controller.go
DerekFrank: I think that makes sense. If we need to do a further optimization later we can do it all in one go across the codebase

pkg/controllers/nodeclaim/lifecycle/termination_test.go
```diff
@@ -368,4 +411,5 @@ var _ = Describe("Termination", func() {
 		ExpectExists(ctx, env.Client, node)
 		ExpectNotFound(ctx, env.Client, nodeClaim)
 	})
+	// TODO: Add the TerminationUtil testing in here so that we hold the same bar for Termination status condition
```
DerekFrank: Whats the reason for putting this off/what gap does this leave?

pkg/controllers/node/termination/controller.go
```diff
@@ -131,23 +130,58 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		result, err := f(ctx, nodeClaim, node, nodeTerminationTime)
-		if result != nil || err != nil {
-			return *result, err
+		result, err = f(ctx, nodeClaim, node, nodeTerminationTime)
+		if !lo.IsEmpty(result) || err != nil {
+			break
+		}
+	}
+	// If we don't have a NodeClaim, then there's nothing for us to patch here
+	if stored != nil && !equality.Semantic.DeepEqual(stored, nodeClaim) {
```
rschalo, line 149: Does DeepEqual not handle nil on its own?

pkg/controllers/node/termination/controller.go
```diff
@@ -131,23 +130,58 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
+		// can cause races due to the fact that it fully replaces the list on a change
+		// Here, we are updating the status condition list
+		if err = c.kubeClient.Status().Patch(ctx, nodeClaim, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); client.IgnoreNotFound(err) != nil {
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, fmt.Errorf("updating nodeclaim, %w", err)
 		}
+		// We only increment the drained metric after we have ensured that we have patched the status condition onto the NodeClaim
+		if !stored.StatusConditions().IsTrue(v1.ConditionTypeDrained) && nodeClaim.StatusConditions().IsTrue(v1.ConditionTypeDrained) {
```
rschalo, line 160: Shouldn't this be a check that both are true?

pkg/controllers/node/termination/controller.go
jonathan-innis: I don't want us to generate a patch if both of these are nil

pkg/controllers/node/termination/controller.go
jonathan-innis: No, I'm only trying to fire a metric when this status condition is changing from not true to true. Otherwise, we get duplicate metrics that are going to skew things

rschalo (approved):
/lgtm

PR #2334, opened 2025-06-25
test: allow e2e tests to output junit report

maxcao13:
Fixes #N/A 

**Description**
Adds the ability to specify `ARTIFACT_DIR` in the `Makefile` which allows e2e tests to output a `junit_report.xml` file to it, if `ARTIFACT_DIR` is specified.

Running
```bash
ARTIFACT_DIR=some_dir make e2etests
```
on a kind cluster with all prerequisite steps (`make install-kwok`, `make apply-with-kind`, etc.).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,12 +41,14 @@ apply-with-kind: verify build-with-kind ## Deploy the kwok controller from the c
 		--set-string controller.env[0].name=ENABLE_PROFILING \
 		--set-string controller.env[0].value=true
 
+JUNIT_REPORT := $(if $(ARTIFACT_DIR), --ginkgo.junit-report="$(ARTIFACT_DIR)/junit_report.xml")
 e2etests: ## Run the e2e suite against your local cluster
 	cd test && go test \
 		-count 1 \
 		-timeout 2h \
 		-v \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -41,12 +41,14 @@ apply-with-kind: verify build-with-kind ## Deploy the kwok controller from the c
 		--set-string controller.env[0].name=ENABLE_PROFILING \
 		--set-string controller.env[0].value=true
 
+JUNIT_REPORT := $(if $(ARTIFACT_DIR), --ginkgo.junit-report="$(ARTIFACT_DIR)/junit_report.xml")
```
engedaam, line 44: Can we make this conditional? Only get the report when we have an ENV

maxcao13: Thanks for the review. 

Sorry if the code is not clear, or I am misinterpreting you but I think the line is already conditional?

The `$(if ...)` function sets the `JUNIT_REPORT` variable to the `--ginkgo.junit-report` flag only when the `ARTIFACT_DIR` environment variable is defined, otherwise it will just be an empty string and not affect the full `go test` command.

PR #2337, opened 2025-06-30
fix: Fix re-retrieving object on retry

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that we re-grab the object when doing retries in-code rather than continually trying to patch with an old version of the object

`make presubmit`


engedaam (commented):
/lgtm

PR #2338, opened 2025-06-30
fix: Fix overriding error with patch call

jonathan-innis:
Fixes #N/A 

**Description**

This change ensures that we don't override the `err` that's returned from any of the `terminationFuncs` when we make a patch call. Currently, if the Patch call doesn't return an error, we can completely ignore an error that actually came from the CloudProvider

`make presubmit`


engedaam (commented):
/lgtm

PR #2341, opened 2025-07-01
fix: pod errors when nodepool requirements filter all instance types

rschalo:
Fixes #N/A 

**Description**
When NodeClaimTemplates are generated for the scheduling code, if NodePool requirements filter out all instance types then zero NodeClaimTemplates will be used for the `parallelizeUntil` call in the scheduler code. It's still important for Karpenter to ack the pods if the pods can schedule to existing or in-flight capacity, so after those checks if zero  NodeClaimTemplates exist, a default error is sent back for the pod via podErrors.

Expanded unit testing and make presubmit

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -444,7 +444,7 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
 		return nil
 	}
 	if len(s.nodeClaimTemplates) == 0 {
-		return serrors.Wrap(fmt.Errorf("nodepool requirements filtered out available instance types"), "Pod", klog.KRef("", pod.Name))
+		return fmt.Errorf("nodepool requirements filtered out all available instance types")
 	}
 	err := s.addToNewNodeClaim(ctx, pod)
 	if err == nil {


pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -4299,6 +4299,147 @@ var _ = Context("Scheduling", func() {
 			Expect(node.Labels).To(HaveKeyWithValue(v1.NodePoolLabelKey, nodePool.Name))
 		})
 	})
+	FDescribe("NodePool requirements instance filtering", func() {
```
jigisha620: Drop the focus from 4302 and 4303?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -442,6 +443,9 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
 	if err := s.addToInflightNode(ctx, pod); err == nil {
 		return nil
 	}
+	if len(s.nodeClaimTemplates) == 0 {
+		return serrors.Wrap(fmt.Errorf("nodepool requirements filtered out all compatible instance types"), "Pod", klog.KRef("", pod.Name))
```
jigisha620: Should this be "nodepool requirements filtered out all available instance types"? Compatible ITs in my opinion are the ones that we get after filtering by requirements.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -442,6 +443,9 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
 	if err := s.addToInflightNode(ctx, pod); err == nil {
 		return nil
 	}
+	if len(s.nodeClaimTemplates) == 0 {
+		return serrors.Wrap(fmt.Errorf("nodepool requirements filtered out available instance types"), "Pod", klog.KRef("", pod.Name))
```
jonathan-innis: The pod details are automatically injected when we actually log the pod failure so I don't think you need to include the reference to the pod here

pkg/test/expectations/expectations.go
```diff
@@ -358,6 +358,16 @@ func ExpectProvisionedNoBinding(ctx context.Context, c client.Client, cluster *s
 	return bindings
 }
 
+func ExpectProvisionedResults(ctx context.Context, c client.Client, cluster *state.Cluster, cloudProvider cloudprovider.CloudProvider, provisioner *provisioning.Provisioner, pods ...*corev1.Pod) scheduling.Results {
```
jonathan-innis, line 361: This is so specific to your use-case -- does it make sense to have this as an expectation at all or should we just put this directly in the test?

rschalo: I could see scenarios where we need to validate the results are what we expect. It's also used enough in these tests that a helper makes sense to me.

PR #2349, opened 2025-07-03
feat: (BREAKING) addition of launch timeout for nodeclaim lifecycle

rschalo:
Fixes #N/A 

**Description**
If Karpenter encounters issues launching instances then we should retry for a shorter amount of time than the full registration timeout. This unblocks provisioning decisions for pods that may be stuck waiting for compute that will not come up due to launch failures. This also allows Karpenter to make new decisions about what compute should be provisioned.

## BREAKING
This PR changes an alpha metric for nodeclaim disruptions. The metric reason `liveness` will change to `registration_timeout`.

Unit tests.

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -18,6 +18,7 @@ package lifecycle
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -39,10 +40,32 @@ type Liveness struct {
 	kubeClient client.Client
 }
 
-// registrationTTL is a heuristic time that we expect the node to register within
+// registrationTimeout is a heuristic time that we expect the node to register within
+// launchTimeout is a heuristic time that we expect to be able to launch within
 // If we don't see the node within this time, then we should delete the NodeClaim and try again
-const registrationTTL = time.Minute * 15
-const launchTTL = time.Minute * 5
+
+const (
+	registrationTimeout       = time.Minute * 15
+	registrationTimeoutReason = "registration_timeout"
+	launchTimeout             = time.Minute * 5
+	launchTimeoutReason       = "launch_timeout"
+)
+
+type NodeClaimTimeout struct {
+	duration time.Duration
+	reason   string
+}
+
+var (
+	RegistrationTimeout = NodeClaimTimeout{
+		duration: registrationTimeout,
+		reason:   registrationTimeoutReason,
+	}
+	LaunchTimeout = NodeClaimTimeout{
+		duration: launchTimeout,
+		reason:   launchTimeoutReason,
+	}
+)
 
 func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
 	registered := nodeClaim.StatusConditions().Get(v1.ConditionTypeRegistered)
@@ -53,18 +76,18 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 	if launched == nil {
 		return reconcile.Result{Requeue: true}, nil
 	}
-	if ttl := launchTTL - l.clock.Since(launched.LastTransitionTime.Time); ttl <= 0 {
-		if err := l.deleteNodeClaimForTTL(ctx, "launch_ttl", nodeClaim); err != nil {
+	if Timeout := launchTimeout - l.clock.Since(launched.LastTransitionTime.Time); Timeout <= 0 {
+		if err := l.deleteNodeClaimForTimeout(ctx, LaunchTimeout, nodeClaim); err != nil {
 			return reconcile.Result{}, err
 		}
 	}
 	if registered == nil {
 		return reconcile.Result{Requeue: true}, nil
 	}
-	// If the Registered statusCondition hasn't gone True during the TTL since we first updated it, we should terminate the NodeClaim
-	// NOTE: ttl has to be stored and checked in the same place since l.clock can advance after the check causing a race
-	if ttl := registrationTTL - l.clock.Since(registered.LastTransitionTime.Time); ttl > 0 {
-		return reconcile.Result{RequeueAfter: ttl}, nil
+	// If the Registered statusCondition hasn't gone True during the Timeout since we first updated it, we should terminate the NodeClaim
+	// NOTE: Timeout has to be stored and checked in the same place since l.clock can advance after the check causing a race
+	if timeout := registrationTimeout - l.clock.Since(registered.LastTransitionTime.Time); timeout > 0 {
+		return reconcile.Result{RequeueAfter: timeout}, nil
 	}
 	if err := l.updateNodePoolRegistrationHealth(ctx, nodeClaim); client.IgnoreNotFound(err) != nil {
 		if errors.IsConflict(err) {
@@ -73,7 +96,7 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 		return reconcile.Result{}, err
 	}
 	// Delete the NodeClaim if we believe the NodeClaim won't register since we haven't seen the node
-	if err := l.deleteNodeClaimForTTL(ctx, "liveness", nodeClaim); err != nil {
+	if err := l.deleteNodeClaimForTimeout(ctx, RegistrationTimeout, nodeClaim); err != nil {
 		return reconcile.Result{}, err
 	}
 	return reconcile.Result{}, nil
@@ -90,7 +113,7 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 		}
 		if nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown() {
 			stored := nodePool.DeepCopy()
-			// If the nodeClaim failed to register during the TTL set NodeRegistrationHealthy status condition on
+			// If the nodeClaim failed to register during the Timeout set NodeRegistrationHealthy status condition on
 			// NodePool to False. If the launch failed get the launch failure reason and message from nodeClaim.
 			if launchCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched); launchCondition.IsTrue() {
 				nodePool.StatusConditions().SetFalse(v1.ConditionTypeNodeRegistrationHealthy, "RegistrationFailed", "Failed to register node")
@@ -108,13 +131,13 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 	return nil
 }
 
-func (l *Liveness) deleteNodeClaimForTTL(ctx context.Context, reason string, nodeClaim *v1.NodeClaim) error {
+func (l *Liveness) deleteNodeClaimForTimeout(ctx context.Context, timeout NodeClaimTimeout, nodeClaim *v1.NodeClaim) error {
 	if err := l.kubeClient.Delete(ctx, nodeClaim); err != nil {
 		return client.IgnoreNotFound(err)
 	}
-	log.FromContext(ctx).V(1).WithValues("ttl", registrationTTL).Info("terminating due to registration ttl")
+	log.FromContext(ctx).V(1).WithValues("timeout", timeout.duration).Info(fmt.Sprintf("terminating due to %s timeout", timeout.reason))
 	metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
-		metrics.ReasonLabel:       reason,
+		metrics.ReasonLabel:       timeout.reason,
 		metrics.NodePoolLabel:     nodeClaim.Labels[v1.NodePoolLabelKey],
 		metrics.CapacityTypeLabel: nodeClaim.Labels[v1.CapacityTypeLabelKey],
 	})
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -75,7 +75,10 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 	if launched == nil {
 		return reconcile.Result{Requeue: true}, nil
 	}
-	if timeUntilTimeout := launchTimeout - l.clock.Since(launched.LastTransitionTime.Time); timeUntilTimeout <= 0 {
+	if !launched.IsTrue() {
+		if timeUntilTimeout := launchTimeout - l.clock.Since(launched.LastTransitionTime.Time); timeUntilTimeout > 0 {
+			return reconcile.Result{RequeueAfter: timeUntilTimeout}, nil
+		}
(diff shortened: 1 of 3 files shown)


pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -42,12 +42,22 @@ type Liveness struct {
...
 // If we don't see the node within this time, then we should delete the NodeClaim and try again
 const registrationTTL = time.Minute * 15
+const launchTTL = time.Minute * 5
 
 func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
 	registered := nodeClaim.StatusConditions().Get(v1.ConditionTypeRegistered)
 	if registered.IsTrue() {
 		return reconcile.Result{}, nil
 	}
+	launched := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched)
```
DerekFrank, line 75: If we're gonna functionally decompose the delete, why not add the status condition check? You could pass in the condition and the TTL to check for in addition to the reason.

I'm not sure they need to be functionally decomposed in the first place but if you want to go that route I think we should decompose it all. We could also define the types of conditions we check for in an array or something more explicit

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -103,3 +107,16 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 	}
 	return nil
 }
+
+func (l *Liveness) deleteNodeClaimForTTL(ctx context.Context, reason string, nodeClaim *v1.NodeClaim) error {
+	if err := l.kubeClient.Delete(ctx, nodeClaim); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+	log.FromContext(ctx).V(1).WithValues("ttl", registrationTTL).Info("terminating due to registration ttl")
```
DerekFrank: nit: we should update the log too

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -42,12 +42,22 @@ type Liveness struct {
...
 	registered := nodeClaim.StatusConditions().Get(v1.ConditionTypeRegistered)
 	if registered.IsTrue() {
 		return reconcile.Result{}, nil
 	}
+	launched := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched)
+	if launched == nil {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	if ttl := launchTTL - l.clock.Since(launched.LastTransitionTime.Time); ttl <= 0 {
+		if err := l.deleteNodeClaimForTTL(ctx, "launch_ttl", nodeClaim); err != nil {
```
DerekFrank: nit: might be worth adding the strings as consts instead of defining them on their own

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -63,15 +73,9 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 		return reconcile.Result{}, err
 	}
 	// Delete the NodeClaim if we believe the NodeClaim won't register since we haven't seen the node
-	if err := l.kubeClient.Delete(ctx, nodeClaim); err != nil {
-		return reconcile.Result{}, client.IgnoreNotFound(err)
+	if err := l.deleteNodeClaimForTTL(ctx, "liveness", nodeClaim); err != nil {
```
DerekFrank: nit: I'm not sure I like that one is liveness and the other is launch_ttl. Maybe `liveness` and `launch` instead?

rschalo: This should have been `registration_ttl` and was a miss the first time.

pkg/controllers/nodeclaim/lifecycle/liveness.go
rschalo: The status conditions and TTL aren't handled the same way and I think it's fine to functionally decompose the delete and metric because of that. I think somewhere in between is something like:
```
type NodeClaimTTL struct {
	duration time.Duration
	reason   string
}

var (
	RegistrationTTL = NodeClaimTTL{
		duration: registrationTTL,
		reason:   registrationTTLReason,
	}
	LaunchTTL = NodeClaimTTL{
		duration: launchTTL,
		reason:   launchTTLReason,
	}
)
```
WDYT?

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -103,3 +107,16 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 	}
 	return nil
 }
+
+func (l *Liveness) deleteNodeClaimForTTL(ctx context.Context, reason string, nodeClaim *v1.NodeClaim) error {
```
jonathan-innis: nit: Do we think we need this helper? Honestly I think we could probably just inline a lot of this

pkg/controllers/nodeclaim/lifecycle/liveness.go
jonathan-innis: I will say that we technically listed this as an alpha metric in our docs -- so we _do_ have the latitude to change some of the details of it -- realistically, I wish that we could double fire the metric so that we can keep the old behavior but, frankly, I'm fine with changing this to `registration_timeout` and calling this out as a breaking change to our metrics in our upgrade guide

pkg/controllers/nodeclaim/lifecycle/liveness.go
jonathan-innis:
```suggestion
		if err := l.deleteNodeClaimForTTL(ctx, "launch_timeout", nodeClaim); err != nil {
```

I think I prefer using the term timeout here over `ttl`

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -72,6 +72,8 @@ var _ = Describe("Liveness", func() {
 			}
```
jonathan-innis, line 73: Do we need to add a test to ensure that we are handling reading the status condition correctly now that we have this new timeout?

pkg/controllers/nodeclaim/lifecycle/liveness.go
rschalo: Need, no, but without it there's code duplication.

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
rschalo: Not sure how you mean - I've added a test that tests we don't delete the nodeclaim before the timeout so we have both before and after the timeout.

The other tests still validate that if we launched successfully that we don't delete the nodeclaim for the launch timeout if we haven't registered before the registration timeout.

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -103,3 +130,16 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 	}
 	return nil
 }
+
+func (l *Liveness) deleteNodeClaimForTimeout(ctx context.Context, timeout NodeClaimTimeout, nodeClaim *v1.NodeClaim) error {
+	if err := l.kubeClient.Delete(ctx, nodeClaim); err != nil {
+		return client.IgnoreNotFound(err)
+	}
+	log.FromContext(ctx).V(1).WithValues("timeout", timeout.duration).Info(fmt.Sprintf("terminating due to %s timeout", timeout.reason))
```
jonathan-innis: Since you are formatting the log line here, that sort-of implies that this could be a structured value -- can you put this value in the `WithValues()` as a `reason`?

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
jonathan-innis: What I meant was a test that validates that we are using the lastTransitionTime of the NodeClaim and not the creationTimestamp of the NodeClaim

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -39,22 +40,54 @@ type Liveness struct {
...
 func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
 	registered := nodeClaim.StatusConditions().Get(v1.ConditionTypeRegistered)
 	if registered.IsTrue() {
 		return reconcile.Result{}, nil
 	}
+	launched := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched)
+	if launched == nil {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	if Timeout := launchTimeout - l.clock.Since(launched.LastTransitionTime.Time); Timeout <= 0 {
```
jonathan-innis:
```suggestion
	if TimeUntilTimeout := launchTimeout - l.clock.Since(launched.LastTransitionTime.Time); TimeUntilTimeout <= 0 {
```

nit: This sort-of makes it a bit clearer what this variable stands for

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -39,22 +39,54 @@ type Liveness struct {
...
 func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
 	registered := nodeClaim.StatusConditions().Get(v1.ConditionTypeRegistered)
 	if registered.IsTrue() {
 		return reconcile.Result{}, nil
 	}
+	launched := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched)
+	if launched == nil {
+		return reconcile.Result{Requeue: true}, nil
+	}
+	if timeUntilTimeout := launchTimeout - l.clock.Since(launched.LastTransitionTime.Time); timeUntilTimeout <= 0 {
```
jonathan-innis: I'm realizing that when we write the logic like this, we won't requeue back again at the launch timeout -- the way we have it written right now, we will only requeue at the registration timeout and not at the launch timeout -- which _could_ mean that we miss the launch timeout event. 

I also think if you just check the launch timeout when you don't have the launched condition set and only check the registration timeout when you don't have the registered condition set, that might simplify your logic

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -147,18 +149,80 @@ var _ = Describe("Liveness", func() {
...
+		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
+		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
+
+		// try again a minute later but before the launch timeout
+		fakeClock.Step(time.Minute * 1)
+		_ = operatorpkg.ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
+		// expect that the nodeclaim was not deleted
+		ExpectExists(ctx, env.Client, nodeClaim)
+	})
+	It("should use the status condition transition times for timeouts, not the creation timestamp", func() {
```
jonathan-innis: We may want to have a test here that validates the RequeueAfter that we are getting both for the Registered and Launched timeouts

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -147,18 +149,80 @@ var _ = Describe("Liveness", func() {
...
+						corev1.ResourceMemory:   resource.MustParse("50Mi"),
+						corev1.ResourcePods:     resource.MustParse("5"),
+						fake.ResourceGPUVendorA: resource.MustParse("1"),
+					},
+				},
+			},
+		})
+		cloudProvider.AllowedCreateCalls = 0 // Don't allow Create() calls to succeed
+		ExpectApplied(ctx, env.Client, nodePool, nodeClaim)
+		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
```
jonathan-innis, line 209: Worth seeing what this result is and making sure that it matches the launch timeout value?

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -147,18 +149,80 @@ var _ = Describe("Liveness", func() {
...
+			condition.LastTransitionTime = metav1.NewTime(fakeClock.Now().Add(10 * time.Minute))
+			newConditions[i] = condition
+		}
+		nodeClaim.Status.Conditions = newConditions
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		// advance the clock to show that the timeout is not based on creation timestamp
+		fakeClock.Step(12 * time.Minute)
+		_ = ExpectObjectReconcileFailed(ctx, env.Client, nodeClaimController, nodeClaim)
+
+		// expect that the nodeclaim was not deleted after the timeout
```
jonathan-innis, line 224: This test validates this behavior for the launched condition -- should we also make sure that we have this same check for the registered condition if it doesn't already exist?

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -103,3 +129,16 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 	}
 	return nil
 }
+
+func (l *Liveness) deleteNodeClaimForTimeout(ctx context.Context, timeout NodeClaimTimeout, nodeClaim *v1.NodeClaim) error {
+	if err := l.kubeClient.Delete(ctx, nodeClaim); err != nil {
+		return client.IgnoreNotFound(err)
```
jonathan-innis: nit: It technically might be better for you to handle this error outside the function because then you can exit early -- if you swallow the error here, you will keep moving through your checks even though you know that the object doesn't exist already

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
rschalo: Needs a new expectation but I can add it.

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -147,18 +150,121 @@ var _ = Describe("Liveness", func() {
...
+			condition.LastTransitionTime = metav1.NewTime(fakeClock.Now().Add(10 * time.Minute))
+			newConditions[i] = condition
+		}
+		nodeClaim.Status.Conditions = newConditions
+		ExpectApplied(ctx, env.Client, nodeClaim)
+		// advance the clock to show that the timeout is not based on creation timestamp when considering registration timeout
+		fakeClock.Step(16 * time.Minute)
+		result := ExpectObjectReconciled(ctx, env.Client, nodeClaimController, nodeClaim)
+		Expect(result.RequeueAfter).To(Not(Equal(0 * time.Second)))
+		Expect(result.RequeueAfter > 0*time.Second && result.RequeueAfter < 15*time.Minute).To(BeTrue())
```
rschalo, line 262: Expect that the `requeueAfter` exists and is within the registration timeout window

jonathan-innis (approved):
/lgtm
/approve

PR #2354, opened 2025-07-07
perf: Add flag to disable costly metrics controllers

DerekFrank:
Fixes #N/A 

**Description**

Metrics controllers can cause metric scrape timeouts at scale, this adds an optional feature flag to disable them

`make presubmit`

### DIFF
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -98,7 +98,7 @@ func NewControllers(
 		nodehydration.NewController(kubeClient, cloudProvider),
 	}
 
-	if !options.FromContext(ctx).SimplifiedMetrics {
+	if !options.FromContext(ctx).DisableClusterStateObservability {
 		controllers = append(controllers,
 			metricspod.NewController(kubeClient, cluster),
 			metricsnodepool.NewController(kubeClient, cloudProvider),
--- a/pkg/operator/options/options.go
+++ b/pkg/operator/options/options.go
@@ -64,28 +64,28 @@ type FeatureGates struct {
 
 // Options contains all CLI flags / env vars for karpenter-core. It adheres to the options.Injectable interface.
 type Options struct {
-	ServiceName             string
-	MetricsPort             int
-	HealthProbePort         int
-	KubeClientQPS           int
-	KubeClientBurst         int
-	EnableProfiling         bool
-	DisableLeaderElection   bool
-	LeaderElectionName      string
-	LeaderElectionNamespace string
-	MemoryLimit             int64
-	CPURequests             int64
-	LogLevel                string
-	LogOutputPaths          string
-	LogErrorOutputPaths     string
-	BatchMaxDuration        time.Duration
-	BatchIdleDuration       time.Duration
-	preferencePolicyRaw     string
-	PreferencePolicy        PreferencePolicy
-	minValuesPolicyRaw      string
-	MinValuesPolicy         MinValuesPolicy
-	FeatureGates            FeatureGates
-	SimplifiedMetrics       bool
+	ServiceName                      string
+	MetricsPort                      int
+	HealthProbePort                  int
+	KubeClientQPS                    int
+	KubeClientBurst                  int
+	EnableProfiling                  bool
+	DisableLeaderElection            bool
+	DisableClusterStateObservability bool
+	LeaderElectionName               string
+	LeaderElectionNamespace          string
+	MemoryLimit                      int64
+	CPURequests                      int64
+	LogLevel                         string
+	LogOutputPaths                   string
+	LogErrorOutputPaths              string
+	BatchMaxDuration                 time.Duration
+	BatchIdleDuration                time.Duration
+	preferencePolicyRaw              string
+	PreferencePolicy                 PreferencePolicy
+	minValuesPolicyRaw               string
+	MinValuesPolicy                  MinValuesPolicy
+	FeatureGates                     FeatureGates
 }
 
 type FlagSet struct {
@@ -113,6 +113,7 @@ func (o *Options) AddFlags(fs *FlagSet) {
 	fs.IntVar(&o.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
 	fs.BoolVarWithEnv(&o.EnableProfiling, "enable-profiling", "ENABLE_PROFILING", false, "Enable the profiling on the metric endpoint")
 	fs.BoolVarWithEnv(&o.DisableLeaderElection, "disable-leader-election", "DISABLE_LEADER_ELECTION", false, "Disable the leader election client before executing the main loop. Disable when running replicated components for high availability is not desired.")
+	fs.BoolVarWithEnv(&o.DisableClusterStateObservability, "disable-cluster-state-observability", "DISABLE_CLUSTER_STATE_OBSERVABILITY", false, "Disable cluster state metrics and events")
 	fs.StringVar(&o.LeaderElectionName, "leader-election-name", env.WithDefaultString("LEADER_ELECTION_NAME", "karpenter-leader-election"), "Leader election name to create and monitor the lease if running outside the cluster")
 	fs.StringVar(&o.LeaderElectionNamespace, "leader-election-namespace", env.WithDefaultString("LEADER_ELECTION_NAMESPACE", ""), "Leader election namespace to create and monitor the lease if running outside the cluster")
 	fs.Int64Var(&o.MemoryLimit, "memory-limit", env.WithDefaultInt64("MEMORY_LIMIT", -1), "Memory limit on the container running the controller. The GC soft memory limit is set to 90% of this value.")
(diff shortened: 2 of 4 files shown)


pkg/controllers/controllers.go
```diff
@@ -99,25 +96,16 @@ func NewControllers(
...
-			status.WithHistogramBuckets(prometheus.ExponentialBuckets(0.5, 2, 15)), // 0.5, 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192
-			status.WithLabels(append(lo.Map(cloudProvider.GetSupportedNodeClasses(), func(obj status.Object, _ int) string { return v1.NodeClassLabelKey(object.GVK(obj).GroupKind()) }), v1.NodePoolLabelKey, v1.NodeInitializedLabelKey)...),
-		),
+	}
+
+	if !options.FromContext(ctx).FeatureGates.DisableMetricsControllers {
+		controllers = append(controllers, metricspod.NewController(kubeClient, cluster),
+			metricsnodepool.NewController(kubeClient, cloudProvider),
+			metricsnode.NewController(cluster),
+			status.NewController[*v1.NodeClaim](kubeClient, mgr.GetEventRecorderFor("karpenter"), status.EmitDeprecatedMetrics, status.WithLabels(append(lo.Map(cloudProvider.GetSupportedNodeClasses(), func(obj status.Object, _ int) string { return v1.NodeClassLabelKey(object.GVK(obj).GroupKind()) }), v1.NodePoolLabelKey)...)),
```
tallaxes: Is there a reason for removal of custom buckets?

pkg/controllers/controllers.go
```diff
@@ -99,25 +96,16 @@ func NewControllers(
 		nodeclaimdisruption.NewController(clock, kubeClient, cloudProvider),
 		nodeclaimhydration.NewController(kubeClient, cloudProvider),
 		nodehydration.NewController(kubeClient, cloudProvider),
-		status.NewController[*v1.NodeClaim](
```
tallaxes, line 102: Are these status controllers only responsible for metrics? Anything else would be missing by disabling them? (For example, looks like they also issue transition events ...)

pkg/controllers/controllers.go
```diff
@@ -99,25 +96,16 @@ func NewControllers(
...
-		),
-		status.NewGenericObjectController[*corev1.Node](
-			kubeClient,
-			mgr.GetEventRecorderFor("karpenter"),
-			status.WithHistogramBuckets(prometheus.ExponentialBuckets(0.5, 2, 15)), // 0.5, 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192
-			status.WithLabels(append(lo.Map(cloudProvider.GetSupportedNodeClasses(), func(obj status.Object, _ int) string { return v1.NodeClassLabelKey(object.GVK(obj).GroupKind()) }), v1.NodePoolLabelKey, v1.NodeInitializedLabelKey)...),
-		),
+	}
+
+	if !options.FromContext(ctx).FeatureGates.DisableMetricsControllers {
```
tallaxes: Is there a good criteria for when to use feature flag vs regular option?

tallaxes (commented):
Would be good to list exactly what would be missing/disabled when this flag is used.

pkg/controllers/controllers.go
DerekFrank: AFAIK we don't have a good set of guidelines for the difference. This might make a lot more sense as a regular option than a feature gate now that I think about it

engedaam: We typically use feature gates for stand alone feature. We have used feature gates for things like drift, spot to spot conditiondation, node repair and etc. This stacks me as a specific functionality change, where having an option/environment variable would be better

pkg/controllers/controllers.go
engedaam: +1

pkg/controllers/controllers.go
DerekFrank: They do also emit events for finalizers and status condition changes, but I think thats all they emit

pkg/controllers/controllers.go
DerekFrank: Ah whoops, I can add those back in

DerekFrank:
This option will disable all of the metrics that are just observing cluster state:

* `karpenter_node_allocatable`
* `karpenter_node_total_pod_requests`
* `karpenter_node_total_pod_limits`
* `karpenter_node_total_daemon_requests`
* `karpenter_node_total_daemon_limits`
* `karpenter_node_system_overhead`
* `karpenter_node_current_lifetime_seconds`
* `karpenter_nodepool_limit`
* `karpenter_nodepool_usage`
* `karpenter_pod_state`
* `karpenter_pod_startup_duration_seconds`
* `karpenter_pod_unstarted_time_seconds`
* `karpenter_pod_bound_duration_seconds`
* `karpenter_pod_unbound_time_seconds`
* `karpenter_pod_provisioning_bound_duration_seconds`
* `karpenter_pod_provisioning_unbound_time_seconds`
* `karpenter_pod_provisioning_startup_duration_seconds`
* `karpenter_pod_provisioning_unstarted_time_seconds`
* `karpenter_pod_provisioning_scheduling_undecided_time_seconds`
* `operator_nodeclaim_status_condition_transition_seconds`
* `operator_nodeclaim_status_condition_count`
* `operator_nodeclaim_status_condition_current_status_seconds`
* `operator_nodeclaim_status_condition_transitions_total`
* `operator_nodepool_status_condition_transition_seconds`
* `operator_nodepool_status_condition_count`
* `operator_nodepool_status_condition_current_status_seconds`
* `operator_nodepool_status_condition_transitions_total`
* `operator_nodeclass_status_condition_transition_seconds`
* `operator_nodeclass_status_condition_count`
* `operator_nodeclass_status_condition_current_status_seconds`
* `operator_nodeclass_status_condition_transitions_total`

pkg/operator/options/options.go
```diff
@@ -85,6 +85,7 @@ type Options struct {
 	minValuesPolicyRaw      string
 	MinValuesPolicy         MinValuesPolicy
 	FeatureGates            FeatureGates
+	SimplifiedMetrics       bool
```
jmdeal: We should figure out a more descriptive name than `SimplifiedMetrics`. Minimally, I think it should be clear from the name of the flag **which** metrics we're disabling. I'm thinking we could have something along the lines of `DisableClusterStateMetrics` and `DisableStatusConditionMetrics`.

DerekFrank: Talked offline, settled on DisableClusterStateObservability

pkg/controllers/controllers.go
```diff
@@ -85,9 +85,6 @@ func NewControllers(
 		informer.NewNodePoolController(kubeClient, cloudProvider, cluster),
 		informer.NewNodeClaimController(kubeClient, cloudProvider, cluster),
 		termination.NewController(clock, kubeClient, cloudProvider, terminator.NewTerminator(clock, kubeClient, evictionQueue, recorder), recorder),
-		metricspod.NewController(kubeClient, cluster),
-		metricsnodepool.NewController(kubeClient, cloudProvider),
-		metricsnode.NewController(cluster),
 		nodepoolreadiness.NewController(kubeClient, cloudProvider),
 		nodepoolregistrationhealth.NewController(kubeClient, cloudProvider),
 		nodepoolcounter.NewController(kubeClient, cloudProvider, cluster),
```
jmdeal, line 90: I'm wondering if it would make sense to include the NodePool Counter controller in this set? I could go either way  - since it's just propagating information from cluster state it should be relatively cheap, but the cost does scale with the number of nodepools. It is a "cluster state observability" controller though. Did it show up at all in your performance review?

pkg/operator/options/options.go
```diff
@@ -112,6 +113,7 @@ func (o *Options) AddFlags(fs *FlagSet) {
 	fs.IntVar(&o.KubeClientBurst, "kube-client-burst", env.WithDefaultInt("KUBE_CLIENT_BURST", 300), "The maximum allowed burst of queries to the kube-apiserver")
 	fs.BoolVarWithEnv(&o.EnableProfiling, "enable-profiling", "ENABLE_PROFILING", false, "Enable the profiling on the metric endpoint")
 	fs.BoolVarWithEnv(&o.DisableLeaderElection, "disable-leader-election", "DISABLE_LEADER_ELECTION", false, "Disable the leader election client before executing the main loop. Disable when running replicated components for high availability is not desired.")
+	fs.BoolVarWithEnv(&o.DisableClusterStateObservability, "disable-cluster-state-observability", "DISABLE_CLUSTER_STATE_OBSERVABILITY", false, "Optionally disable cluster state metrics and events")
```
jmdeal: nit: optional is implied
```suggestion
	fs.BoolVarWithEnv(&o.DisableClusterStateObservability, "disable-cluster-state-observability", "DISABLE_CLUSTER_STATE_OBSERVABILITY", false, "Disables cluster state metrics and events")
```

pkg/controllers/controllers.go
DerekFrank: We were using a relatively small number of nodepools (~12), we didn't see significant performance impact. One thing I want to explore is how exactly the volume of metrics changes as we scale the number of nodes in the cluster

jmdeal: It seems fine to include it for the time being, and we can re-evaluate if we see significant performance impact.

PR #2357, opened 2025-07-07
perf: Remove deep equals in favor of checking if the status has been updated

DerekFrank:
Fixes #N/A 

**Description**

This change optimizes the disruption controller's memory footprint by only comparing the the status conditions to verify if a nodeclaim has been drifted

`make presubmit`

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/consolidation.go
+++ b/pkg/controllers/nodeclaim/disruption/consolidation.go
@@ -34,6 +34,8 @@ type Consolidation struct {
 	clock      clock.Clock
 }
 
+// This function should only modify the nodeclaim's Status section
+//
 //nolint:gocyclo
 func (c *Consolidation) Reconcile(ctx context.Context, nodePool *v1.NodePool, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
 	hasConsolidatableCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeConsolidatable) != nil
--- a/pkg/controllers/nodeclaim/disruption/drift.go
+++ b/pkg/controllers/nodeclaim/disruption/drift.go
@@ -47,6 +47,7 @@ type Drift struct {
 	instanceTypeNotFoundCheckCache *cache.Cache
 }
 
+// This function should only modify the nodeclaim's Status section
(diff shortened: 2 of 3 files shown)


tallaxes (approved):

pkg/controllers/nodeclaim/disruption/consolidation.go
```diff
@@ -34,6 +34,8 @@ type Consolidation struct {
 	clock      clock.Clock
 }
 
+// This function should only modify the nodeclaim's Status section
```
jonathan-innis: nit: probably an unnecessary comment -- seems fine to just add the change -- it makes logical sense

pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -47,6 +47,7 @@ type Drift struct {
 	instanceTypeNotFoundCheckCache *cache.Cache
 }
 
+// This function should only modify the nodeclaim's Status section
```
jonathan-innis: nit: probably an unnecessary comment -- seems fine to just add the change -- it makes logical sense

jonathan-innis: FWIW: If someone were to update the spec or metadata today anyways, it wouldn't propagate so them updating other things would require them to have to know to call Patch differently anyways

pkg/controllers/nodeclaim/podevents/controller.go
```diff
@@ -89,7 +89,7 @@ func (c *Controller) Reconcile(ctx context.Context, pod *corev1.Pod) (reconcile.
 	// otherwise, set the pod event time to now
 	stored := nc.DeepCopy()
 	nc.Status.LastPodEventTime.Time = c.clock.Now()
-	if !equality.Semantic.DeepEqual(stored, nc) {
+	if !equality.Semantic.DeepEqual(stored.Status.LastPodEventTime, nc.Status.LastPodEventTime) {
```
ryan-mist, line 92: Is the reason we call `equality.Semantic.DeepEqual` here instead of always patching the status condition because of the following race condition when we're running with multiple concurrent reconciles?

> Before `stored := nc.DeepCopy()` (but after the ` c.clock.Since(nc.Status.LastPodEventTime.Time) < dedupeTimeout` check) another goroutine modifies `nc.Status.LastPodEventTime.Time` to time `t`. Then our goroutine has a deep copy with time `t` and when we call `c.clock.Now()` we also happen to get time `t`

PR #2366, opened 2025-07-11
fix: flakiness in expiration tests

saurav-agarwalla:
Fixes #N/A 

**Description**

`should expire the node after the expiration is reached` has been failing fairly frequently after https://github.com/kubernetes-sigs/karpenter/pull/2324 since the node and nodeclaim objects are deleted so quickly that the current validation doesn't work. This commit removes that validation since the eventual check of the node and nodeclaim being deleted should be good enough. I looked around to see if we can use any event to validate the expiration more precisely but couldn't find anything else.

Ran the test and it succeeds locally now. It would fail before this change.

### DIFF
--- a/test/suites/regression/expiration_test.go
+++ b/test/suites/regression/expiration_test.go
@@ -24,14 +24,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 
 	"sigs.k8s.io/karpenter/pkg/test"
 
 	. "github.com/onsi/ginkgo/v2"
-	. "github.com/onsi/gomega"
 )
 
 var _ = Describe("Expiration", func() {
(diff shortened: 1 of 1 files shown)


test/suites/regression/expiration_test.go
```diff
@@ -81,15 +81,6 @@ var _ = Describe("Expiration", func() {
 		env.EventuallyExpectHealthyPodCount(selector, numPods)
 		env.Monitor.Reset() // Reset the monitor so that we can expect a single node to be spun up after expiration
 
-		// Eventually the node will be tainted, which means its actively being disrupted
-		Eventually(func(g Gomega) {
```
engedaam, line 85: I think we still want to make sure the expired node is getting deleted. The next line is validating that we get a new node. When using KWOK, can we just opt to check if the node exists?

engedaam: We also need to do the same fix for the test below

saurav-agarwalla: We are validating that the node and node claim is deleted on L94 below.

saurav-agarwalla: Good point on the other test. For some reason I only saw this one being flaky maybe because it only has one pod so drains quickly. I'll change the other one too.

engedaam: > We are validating that the node and node claim is deleted on L94 below.


Ahh, missed that. sounds good

saurav-agarwalla (commented):
/retest

PR #2368, opened 2025-07-11
test: moving expectations to use operatorpkg when it exists over there

### LINKED ISSUE #1348: Migrate Ginkgo Expectations to operatorpkg

rlanhellas:
Fixes #1348

**Description**
Moving expectations from karpenter to use operatorpkg when it exists. There are some expectations which exists on operatorpkg but does not work properly in karpenter, so I changed some stuff on operatorpkg, here is the PR: https://github.com/awslabs/operatorpkg/pull/169. 

Locally on my Goland IDE.


pkg/controllers/disruption/consolidation_test.go
```diff
@@ -36,6 +38,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	. "github.com/awslabs/operatorpkg/test/expectations"
```
engedaam, line 39: Can we make the imported package on that is aliased as opposed to local package? To me it makes more sense imported function to be from the same package from a style perspective

rlanhellas: Did you mean adding something like operatorpkg alias instead of dot ?

engedaam: Yeah, switching to use an alias for the operatorpkg package and using . for the karpenter functions

rlanhellas: done @engedaam

pkg/test/expectations/expectations.go
```diff
@@ -136,126 +125,6 @@ func ExpectNotScheduled(ctx context.Context, c client.Client, pod *corev1.Pod) *
...
-		Expect(c.Delete(ctx, object)).To(Succeed())
-		DeferCleanup(func(obj client.Object) {
-			mergeFrom := client.MergeFrom(obj.DeepCopyObject().(client.Object))
-			obj.SetFinalizers([]string{})
-			Expect(client.IgnoreNotFound(c.Patch(ctx, obj, mergeFrom))).To(Succeed())
-		}, object)
-	}
-}
-
-func ExpectCleanedUp(ctx context.Context, c client.Client) {
```
engedaam, line 227: Why are we removing renaming/moving this function?

pkg/controllers/state/suite_test.go
```diff
@@ -95,25 +97,25 @@ var _ = BeforeEach(func() {
 	state.ClusterStateUnsyncedTimeSeconds.Reset()
 	cloudProvider.InstanceTypes = fake.InstanceTypesAssorted()
 	nodePool = test.NodePool(v1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: "default"}})
-	ExpectApplied(ctx, env.Client, nodePool)
+	operatorpkg.ExpectApplied(ctx, env.Client, nodePool)
 })
 var _ = AfterEach(func() {
-	ExpectCleanedUp(ctx, env.Client)
```
engedaam, line 101: Is this change necessary?

pkg/test/expectations/expectations.go
```diff
@@ -753,3 +622,43 @@ func ExpectParallelized(fs ...func()) {
...
+				defer GinkgoRecover()
+				Expect(c.DeleteAllOf(ctx, object, client.InNamespace(namespace),
+					&client.DeleteAllOfOptions{DeleteOptions: client.DeleteOptions{GracePeriodSeconds: lo.ToPtr(int64(0))}})).ToNot(HaveOccurred())
+			}(object, namespace.Name)
+		}
+	}
+	wg.Wait()
+}
+
+func ExpectObjectReconciledWithResult[T client.Object](ctx context.Context, c client.Client, reconciler reconcile.ObjectReconciler[T], object T) reconcile.Result {
```
engedaam: Why are we using this function? Why not the operatorpkg equivalent? https://github.com/awslabs/operatorpkg/blob/main/test/expectations/expectations.go#L34

rlanhellas: This expectations from operatorpkg is not compatible with karpenter because it returns `type.Assertion` but karpenter waits for `reconcile.Result`. In some places, like [this one](https://github.com/kubernetes-sigs/karpenter/blob/6891d210dba9e196d6f3b103b070fbc1cc8fb10b/pkg/controllers/node/termination/suite_test.go#L529), the `reconcile.Result` is essential because it's a argument for other method call.

pkg/controllers/state/suite_test.go
rlanhellas: you're  right, we don't need to change it. Just reverted that change.

pkg/test/expectations/expectations.go
rlanhellas: just reverted that change. We don't need change anything here.

pkg/controllers/disruption/consolidation_test.go
```diff
@@ -25,6 +25,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	. "sigs.k8s.io/karpenter/pkg/test/expectations"
```
jonathan-innis: What expectations do we need to drop from our current test expectations code so that we can dot-import the operatorpkg package? I'd like to avoid us having to prefix every expectation function that we use with operatorpkg with `operatorpkg` and, instead, be able to just swap-out the package that we are leveraging here

pkg/test/expectations/expectations.go
jonathan-innis: If we have specific places where we rely on the result and we need to reuse it, can we make those the exceptions to the rule and maybe just have specific handling for those cases with `ExpectReconcniled` rather than `ExpectObjectReconciled`? If we think that this is commonly enough used that we need to have a common function for it, I think we shoud consider adding `ExpectObjectReconciled(ctx context.Context, c client.Client, reconciler reconcile.ObjectReconciler[T], object T, opts ...reconcile.Options)` and then add an option or something to return the result rather than the expectation for the result -- I'm hoping that we don't have to do that and can instead just continually reuse the existing functions in operatorpkg

pkg/controllers/disruption/consolidation_test.go
rlanhellas: @jonathan-innis , we can swap the dot-import to operatorpkg instead of local expectations. So we will prefix all local expectations with something like  `localexp.Expect...` , and for operatorpkg we use dot-import. @engedaam  suggested to use dot-import in local expectations and named import for operatorpkg. please keep me honest here @engedaam . 

Another important point , we have a bunch of local expectations which rely on structs in karpenter project and we can't move it to operatorpkg, if we do that we'll need to move the karpenter structs too, here is an example: https://github.com/kubernetes-sigs/karpenter/blob/cf8d7b0c786208ada19f004c0294edea304b8533/pkg/test/expectations/expectations.go#L548.

pkg/test/expectations/expectations.go
rlanhellas: @jonathan-innis , I removed the `ExpectObjectReconciledWithResult` and I'm using the deprecated `ExpectReconciled` from operatorpkg, it's working fine, I just needed to add `//nolint: staticcheck` since it's a deprecated func and lint is complaining about that.

pkg/controllers/disruption/consolidation_test.go
engedaam: I believe the point the @jonathan-innis is making here is that we should drop the aliasing for both the Karpenter and operatorpkg. Seems like we need to enumerate all the places where we see a conflict between the two packages and resolve them. This could moving more generic function or possible constructing helper functions using once in the operatorpkg.

rlanhellas: got it, need more some time to same these fixes and come back here

rlanhellas: did all changes @engedaam  @jonathan-innis

engedaam (commented):
/lgtm

pkg/test/expectations/expectations.go
```diff
@@ -755,3 +623,9 @@ func ExpectParallelized(fs ...func()) {
 	}
 	wg.Wait()
 }
+
+func ExpectForceCleanedUpAll(ctx context.Context, c client.Client) {
+	ExpectForceCleanedUp(ctx, c, &corev1.NodeList{}, &v1.NodeClaimList{}, &corev1.PersistentVolumeClaimList{}, &corev1.PersistentVolumeList{},
```
engedaam: Optional nit.

Consider defining these values in a slice at the top of the file for improved readability

rlanhellas: Done

PR #2376, opened 2025-07-16
fix: nodeAffinity ignore restricted label

### LINKED ISSUE #2230: Pods using Restricted Node Labels don't scale up nodes
### Description

**Observed Behavior**:
Pods that contain reserved node labels in their selection affinity selection cause karpenter to not scale up new nodes for the workload and the pods are stuck in `Pending`

**Expected Behavior**:
Because the `nodeSelectorTerms` is an OR statement, the inclusion of a single reserved node label shouldn't prevent scaling up. In the example below we try to schedule to a node with `node-role.kubernetes.io/my-node-role: true` OR `kubernetes.io/arch: amd64` yes no scale up is observed. 

Deploy this yaml, observe that karpenter doesn't create a new nodeclaim
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: karpenter-test-deployment
  namespace: spinnaker
  labels:
    app.kubernetes.io/name: nginx
spec:
  replicas: 10
  selector:
    matchLabels:
      app.kubernetes.io/name: nginx
  template:
    metadata:
      labels:
        app.kubernetes.io/name: nginx
    spec:
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/my-node-role
                operator: In
                values:
                - "true"
            - matchExpressions:
              - key: kubernetes.io/arch
                operator: In
                values:
                - amd64
      containers:
      - name: nginx
        image: nginx:latest
        ports:
        - containerPort: 80
        resources:
          requests:
            cpu: "6"
            memory: "12Gi"
          limits:
            memory: "12Gi"
```
**Versions**:
- Chart Version: 1.3.3
- Kubernetes Version (`kubectl version`): 1.30

phuhung273:
Fixes #2230

**Description**
- Affinity validation passes if at least one of nodeSelectorTerm passes

Unit test

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -539,7 +539,11 @@ func validateAffinity(ctx context.Context, p *corev1.Pod) (errs error) {
 	if p.Spec.Affinity.NodeAffinity != nil {
 		if p.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution != nil {
 			for _, term := range p.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms {
-				errs = multierr.Append(errs, validateNodeSelectorTerm(ctx, term))
+				err := validateNodeSelectorTerm(ctx, term)
+				if err == nil {
+					return nil
+				}
+				errs = multierr.Append(errs, err)
 			}
 		}
 	}
--- a/pkg/controllers/provisioning/suite_test.go
+++ b/pkg/controllers/provisioning/suite_test.go
@@ -410,6 +410,54 @@ var _ = Describe("Provisioning", func() {
 			ExpectNotScheduled(ctx, env.Client, pod)
 		}
 	})
+	It("should provision nodes for pods with node affinities containing restricted and applicable label", func() {
+		nodePool := test.NodePool()
+		schedulable := []*corev1.Pod{
+			test.UnschedulablePod(test.PodOptions{NodeSelectorTerms: []corev1.NodeSelectorTerm{
+				{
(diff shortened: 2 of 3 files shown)


jonathan-innis:
@phuhung273 I'm curious if we want to allow this because, while having at least one valid nodeAffinity is a valid configuration, I will say that it sort of works against the user's intention and I'd argue that it's probably better for them to be aware that there is an issue vs. not

phuhung273:
Thanks for taking a look @jonathan-innis. I just feel that it is unreasonable for `nodeSelectorTerms`, which is known to be OR everywhere in k8s world, to be AND here.

This can help teams migrating from one autoscaling engine to another. Ex: my team migrating from ClusterAutoscaler to Karpenter, not flipping all logic from AND to OR would be 1 less thing to worry.

BojanZelic:
> @phuhung273 I'm curious if we want to allow this because, while having at least one valid nodeAffinity is a valid configuration, I will say that it sort of works against the user's intention and I'd argue that it's probably better for them to be aware that there is an issue vs. not

for the example in https://github.com/kubernetes-sigs/karpenter/issues/2230
```
      affinity:
        nodeAffinity:
            nodeSelectorTerms:
            - matchExpressions:
              - key: node-role.kubernetes.io/my-node-role
                operator: In
                values:
                - "true"
            - matchExpressions:
              - key: kubernetes.io/arch
                operator: In
                values:
                - amd64
```

If this is intentional, it's not very clear. There's no message indicating why a scale up isn't happening on Karpenter's end. Also this leads to very confusing behavior where the K8s scheduler will still schedule a pod if there's a node available, but it won't trigger a scale up with karpenter. So in some cases the pod does run and in other cases where it stays in `Pending` making it very confusing to debug.

PR #2379, opened 2025-07-17
fix: handle multiple PDBs for the same pod more gracefully

### LINKED ISSUE #2361: Karpenter gets stuck in an error loop when trying to evict a pod that has multiple PodDisruptionBudget
### Description

**Observed Behavior**:
If there are multiple PodDisruptionBudget resources that target the same pod (I wouldn't do this... but it can happen by mistake and appears to be allowed by k8s), then Karpenter will enter an error loop when it tries to evict that pod. It will be unable to perform any other operations during this time.

Example:
```
{
  "level": "ERROR",
  "time": "2025-07-10T02:45:55.200Z",
  "logger": "controller",
  "message": "failed evicting pod",
  "commit": "930b1f9",
  "controller": "eviction-queue",
  "namespace": "",
  "name": "",
  "reconcileID": "8f050e14-5026-4f73-ada0-c1b5454f1a38",
  "Pod": {
    "name": "somepod-7678668685-t2t7d",
    "namespace": "development"
  },
  "error": "This pod has more than one PodDisruptionBudget, which the eviction subresource does not support."
}
{
  "level": "ERROR",
  "time": "2025-07-10T02:46:05.216Z",
  "logger": "controller",
  "message": "failed evicting pod",
  "commit": "930b1f9",
  "controller": "eviction-queue",
  "namespace": "",
  "name": "",
  "reconcileID": "8904f387-00e4-4f7b-b22e-7b1952ee6b2c",
  "Pod": {
    "name": "somepod-7678668685-t2t7d,
    "namespace": "development"
  },
  "error": "This pod has more than one PodDisruptionBudget, which the eviction subresource does not support."
}
... [same error message continues forever....]
```

**Expected Behavior**:
Karpenter skips over the pod so it can continue making other scaling decisions. A single pod should not grind scaling to a halt.

1. Create two pdbs that have the same label selectors.
2. Create a pod with these labels that will be placed on a Karpenter-managed node.
3. Adjust pod resources so that the node it is on will be marked as underutilized and Karpenter will try to consolidate.
4. Karpenter will fail to consolidate because it will be unable to evict the pod.

**Versions**:
- Chart Version: 1.5.1
- Kubernetes Version (`kubectl version`): 1.30.13
- Karpenter Version: 1.5.1

saurav-agarwalla:
Fixes #2361 

**Description**

Eviction doesn't support multiple PDBs today so handle that more gracefully instead of letting the controller fail with an error. Upstream discussion around this: https://github.com/kubernetes/kubernetes/issues/75957

`make presubmit`


pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -55,6 +55,8 @@ import (
 const (
 	evictionQueueBaseDelay = 100 * time.Millisecond
 	evictionQueueMaxDelay  = 10 * time.Second
+
+	multiplePodDisruptionBudgetsError = "This pod has more than one PodDisruptionBudget, which the eviction subresource does not support."
```
rschalo, line 60: Is this available as an import from `corev1` or similar?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -184,12 +188,14 @@ func (q *Queue) Reconcile(ctx context.Context, pod *corev1.Pod) (reconcile.Resul
...
+		// 429 - PDB violation
+		if apierrors.IsTooManyRequests(err) || message == multiplePodDisruptionBudgetsError {
 			node, err2 := podutils.NodeForPod(ctx, q.kubeClient, pod)
 			if err2 != nil {
 				return reconcile.Result{}, err2
 			}
-			q.recorder.Publish(terminatorevents.NodeFailedToDrain(node, serrors.Wrap(fmt.Errorf("evicting pod violates a PDB"), "Pod", klog.KRef(pod.Namespace, pod.Name))))
+			errorMessage := lo.Ternary(message == multiplePodDisruptionBudgetsError, "eviction does not support multiple PDBs", "evicting pod violates a PDB")
+			q.recorder.Publish(terminatorevents.NodeFailedToDrain(node, serrors.Wrap(errors.New(errorMessage), "Pod", klog.KRef(pod.Namespace, pod.Name))))
 			return reconcile.Result{RequeueAfter: evictionQueueBaseDelay}, nil
```
rschalo: Does it make sense to use the same `requeue` for this misconfiguration? In the 429 case we could get to a point where the pod is evictable but that doesn't happen when multiple PDBs point to a pod.

pkg/controllers/node/termination/terminator/eviction.go
saurav-agarwalla: It is unfortunate but no: https://github.com/kubernetes/kubernetes/blob/84cacae7046df93c1f6f8ea97c912d948e1ad06a/pkg/registry/core/pod/storage/eviction.go#L226

pkg/controllers/node/termination/terminator/eviction.go
saurav-agarwalla: It is similar to PDBs as in when PDBs are misconfigured (e.g. set to 100% available or 0% unavailable), nothing can happen until the user takes an action. Same for this case. I could increase the requeue interval if you feel strongly about it but there's no "good" value for it as we don't know when the user is going to fix it.

rschalo: For the misconfiguration cases I think it makes sense to requeue slower. 

We don't know when the user may update things but in reality the difference between 100ms and 1s is probably unnoticeable to the user via kubectl but a 10x reduction in events for a single pod. Seems like a decent trade-off.

saurav-agarwalla: That's fair. 1 sec seems to be a reasonable trade-off. I am working on some more optimizations too so I'll put up a new revision along with those changes.

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -170,8 +172,10 @@ func (q *Queue) Reconcile(ctx context.Context, pod *corev1.Pod) (reconcile.Resul
 			},
 		}); err != nil {
 		var apiStatus apierrors.APIStatus
+		var message string
 		if errors.As(err, &apiStatus) {
 			code := apiStatus.Status().Code
+			message = apiStatus.Status().Message
```
jonathan-innis, line 179: Is this a different error code entirely? Rather than relying on the message (which probably won't change, but could), can we use the status code, particularly if this returns a 400

saurav-agarwalla: No, this particular error seems to have been handled very oddly upstream: https://github.com/kubernetes/kubernetes/blob/84cacae7046df93c1f6f8ea97c912d948e1ad06a/pkg/registry/core/pod/storage/eviction.go#L226. There's not even a traditional `IsXXXError` for this and I don't even understand why they made it a 500.

jonathan-innis: +1 to considering trying to push for this being a 400 upstream

saurav-agarwalla: Yup, will get this going upstream via a PR but probably going to be a tough battle changing the status code at this point. But hopefully I can at least move things around to make this a little easier for clients.

PR #2380, opened 2025-07-17
perf: Optimistically delete from the cache after launch

jonathan-innis:
Fixes #N/A 

**Description**

#2307 bumped the cache expiration timeout for the lifecycle controller to one hour. This increases the number of NodeClaim pointers that we can store in this cache. To reduce the amount of time that we hold things in this cache, we can optimistically remove entries from this cache by deleting the entry when we know we have stored all of the data after we have marked the node as launched

`make presubmit`


rschalo (approved):
/lgtm

Issue #2325, opened 2025-06-23
Support Pod Level Resources


toVersus:
### Description

**What problem are you trying to solve?**

The feature defined in [KEP-2837: Pod Level Resource Specifications](https://github.com/kubernetes/enhancements/issues/2837) is planned to graduate to beta and become enabled by default in Kubernetes 1.34. This feature allows users to specify resource requirements at the pod level, in addition to the existing container-level resource specifications. For more details, please refer to the [official documentation](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/) and [KEP-2837](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2837-pod-level-resource-spec).

KEP-2837 introduced another change to how resource requests are calculated. Prior to the implementation of this KEP, the logic for computing complex resource requests was moved into a shared library under component-helpers. As a result, components that need to calculate resource requirements can now use the [k8s.io/component-helpers/resource](https://pkg.go.dev/k8s.io/component-helpers/resource) package.

Currently, it appears that Karpenter uses [its own implementation](https://github.com/kubernetes-sigs/karpenter/blob/d0f1c47a99dda98ec482a810bc00e742ea7331fc/pkg/utils/resources/resources.go#L111-L144) for calculation of resource requirements. To support Pod Level Resources, the following updates will likely be necessary:

- Use [PodRequests](https://pkg.go.dev/k8s.io/component-helpers/resource#PodRequests) from k8s.io/component-helpers/resource.
- Update any custom logic to support Pod Level Resources.

Just for reference, Cluster Autoscaler has already been migrated to use `k8s.io/component-helpers/resource` in https://github.com/kubernetes/autoscaler/pull/8049.

**How important is this feature to you?**

Once Kubernetes 1.34 is released and users start using the Pod Level Resources feature, it’s likely that Karpenter will run into issues due to inaccurate node provisioning.

Also, as part of the beta graduation criteria for KEP-2837, we’re supposed to share this issue with the Karpenter folks, so I’ve created an issue. Please refer to [the beta criteria](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/2837-pod-level-resource-spec#phase-2--beta-target-134) and https://github.com/kubernetes/enhancements/pull/5362#discussion_r2144195575 for more details.

[...]

PR #2383, opened 2025-07-18
feat: Support Pod Level Resources

toVersus:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/2325

**Description**

This PR updates Karpenter to support Pod Level Resources. The Pod Level Resources feature is planned to graduate to beta in Kubernetes 1.34. Calculating pod resource requests and limits is complex, and to allow components like the scheduler and Node Autoscalers to share a common logic, the necessary functions have been published in [k8s.io/component-helpers/resources](https://pkg.go.dev/k8s.io/component-helpers/resource). While Karpenter currently implements its own custom logic, this PR transitions it to use the shared library.

There is one difference between the shared library and the existing custom logic. The current custom logic includes behavior that merges limits into requests when only limits are specified. However, this merging is actually handled by the API server's defaulting logic. Some existing tests rely on this merging logic, and without the second commit, those tests fail. Personally, I think Karpenter doesn't need to implement this logic explicitly, but WDYT?

```sh
make presubmit
```

### DIFF
--- a/pkg/controllers/provisioning/suite_test.go
+++ b/pkg/controllers/provisioning/suite_test.go
@@ -1016,7 +1016,7 @@ var _ = Describe("Provisioning", func() {
 				test.DaemonSetOptions{PodOptions: test.PodOptions{
 					ResourceRequirements: corev1.ResourceRequirements{
 						Limits:   corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("10000"), corev1.ResourceMemory: resource.MustParse("10000Gi")},
-						Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1")},
+						Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourceMemory: resource.MustParse("10000Gi")}, // simulate the API server’s defaulting from limits
 					},
 				}},
 			))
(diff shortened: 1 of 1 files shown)


jmdeal:
> Personally, I think Karpenter doesn't need to implement this logic explicitly, but WDYT?

So the issue is that the defaulting applies to the pods created for the daemonset, but not for the daemonset itself. Karpenter reads the daemonset spec to compute the "daemonset overhead" for a NodePool. We do this by creating fake daemonset pods based on the specs ([ref](https://github.com/kubernetes-sigs/karpenter/blob/6825a12e67c9c951a2b103120f164ce6c0f73281/pkg/controllers/provisioning/provisioner.go#L465-L490)) and then calling `RequestsForPods` against those pods ([ref](https://github.com/kubernetes-sigs/karpenter/blob/6825a12e67c9c951a2b103120f164ce6c0f73281/pkg/controllers/provisioning/scheduling/scheduler.go#L775-L781)).

That being said, I think it's reasonable for this interface to expect that the pods provided adhere to the API server's defaulting logic. It should be the responsibility of any component generating "fake" pods to ensure that they meet those requirements. In this case, we would want to update our `getDaemonsetPods` function to handle that defaulting. 

Other than that, this PR looks good to me and I'm happy to take on this shared dependency and drop our custom logic. Apologies that it took a while to get review, our bandwidth right now is really limited and we're trying to get more folks promoted to reviewer / approver. We are actually in a bit of a hurry to get this in for 1.34 - I was about to get a PR up but saw yours - so since the rest of the PR looks good I'm going to go ahead and approve then follow up with the change to `getDaemonsetPods`.

/lgtm
/approve

PR #2384, opened 2025-07-18
fix: Don't schedule a pod with DRA requirements

alimaazamat:
**Description**
Currently Karpenter creates nodes that can't satisfy DRA requirements and keep trying to schedule pod to the node.
Karpenter should recognize that pods are DRA enabled and not try to create nodes to fulfill them.

Unit tests and manually tested pod with resourceClaims in spec and scheduler correctly logs skipping message.

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -162,8 +162,8 @@ func NewScheduler(
 		nodeClaimTemplates:  templates,
 		topology:            topology,
 		cluster:             cluster,
-		daemonOverhead:      getDaemonOverhead(templates, daemonSetPods),
-		daemonHostPortUsage: getDaemonHostPortUsage(templates, daemonSetPods),
+		daemonOverhead:      getDaemonOverhead(ctx, templates, daemonSetPods),
+		daemonHostPortUsage: getDaemonHostPortUsage(ctx, templates, daemonSetPods),
 		cachedPodData:       map[types.UID]*PodData{}, // cache pod data to avoid having to continually recompute it
 		recorder:            recorder,
 		preferences:         &Preferences{ToleratePreferNoSchedule: toleratePreferNoSchedule},
@@ -177,14 +177,15 @@ func NewScheduler(
 		minValuesPolicy:         minValuesPolicy,
 		numConcurrentReconciles: lo.Ternary(option.Resolve(opts...).numConcurrentReconciles > 0, option.Resolve(opts...).numConcurrentReconciles, 1),
 	}
-	s.calculateExistingNodeClaims(stateNodes, daemonSetPods)
+	s.calculateExistingNodeClaims(ctx, stateNodes, daemonSetPods)
 	return s
 }
 
 type PodData struct {
-	Requests           corev1.ResourceList
-	Requirements       scheduling.Requirements
-	StrictRequirements scheduling.Requirements
+	Requests                 corev1.ResourceList
+	Requirements             scheduling.Requirements
+	StrictRequirements       scheduling.Requirements
+	HasResourceClaimRequests bool
 }
 
 type Scheduler struct {
@@ -209,6 +210,25 @@ type Scheduler struct {
 	numConcurrentReconciles int
 }
 
+// DRAError indicates a pod will not be attempted to be scheduled because it has Dynamic Resource Allocation requirements
+// that are not yet supported by Karpenter
+type DRAError struct {
+	error
+}
+
+func NewDRAError(err error) DRAError {
+	return DRAError{error: err}
+}
+
+func IsDRAError(err error) bool {
+	draErr := &DRAError{}
+	return errors.As(err, draErr)
+}
+
+func (e DRAError) Unwrap() error {
+	return e.error
+}
+
 // Results contains the results of the scheduling operation
 type Results struct {
 	NewNodeClaims []*NodeClaim
@@ -223,6 +243,11 @@ func (r Results) Record(ctx context.Context, recorder events.Recorder, cluster *
 	// Report failures and nominations
 	for p, err := range r.PodErrors {
 		if IsReservedOfferingError(err) {
+			continue
+		}
+		if IsDRAError(err) {
+			recorder.Publish(PodFailedToScheduleEvent(p, err))
+			log.FromContext(ctx).WithValues("Pod", klog.KObj(p)).Info("skipping pod with Dynamic Resource Allocation requirements, not yet supported by Karpenter")
 			continue
 		}
 		log.FromContext(ctx).WithValues("Pod", klog.KObj(p)).Error(err, "could not schedule pod")
@@ -261,6 +286,12 @@ func (r Results) Record(ctx context.Context, recorder events.Recorder, cluster *
 func (r Results) ReservedOfferingErrors() map[*corev1.Pod]error {
 	return lo.PickBy(r.PodErrors, func(_ *corev1.Pod, err error) bool {
 		return IsReservedOfferingError(err)
+	})
+}
+
+func (r Results) DRAErrors() map[*corev1.Pod]error {
+	return lo.PickBy(r.PodErrors, func(_ *corev1.Pod, err error) bool {
+		return IsDRAError(err)
 	})
 }
 
@@ -416,6 +447,11 @@ func (s *Scheduler) trySchedule(ctx context.Context, p *corev1.Pod) error {
 		if IsReservedOfferingError(err) {
 			return err
 		}
+		// DRA errors are permanent while the IgnoreDRARequests flag is enabled, so we shouldn't attempt to relax
+		// pod requirements as we don't want to schedule the pod.
+		if IsDRAError(err) {
+			return err
+		}
 		// Eventually we won't be able to relax anymore and this while loop will exit
 		if relaxed := s.preferences.Relax(ctx, p); !relaxed {
 			return err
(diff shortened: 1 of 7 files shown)


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -445,10 +446,17 @@ func (s *Scheduler) updateCachedPodData(p *corev1.Pod) {
...
 		Requirements:       requirements,
 		StrictRequirements: strictRequirements,
+		IsDRAEnabled:       pod.HasDRARequirements(p),
 	}
 }
 
 func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
+	// Check if pod has DRA requirements - if so, return immediately as DRA is not yet supported
+	if s.cachedPodData[pod.UID].IsDRAEnabled {
+		log.FromContext(ctx).V(1).WithValues("Pod", klog.KObj(pod)).Info("skipping pod scheduling, Dynamic Resource Allocation (DRA) is not yet supported by Karpenter")
```
jmdeal: Is there a reason we're just logging rather than returning an error here? No error suggests that we successfully scheduled a pod. Returning an error here wouldn't short-circuit the scheduler, and it has the value of ensuring that we include the error in the scheduling result. 

One thing we may want to consider is wrapping it in a special type of error (e.g. `DRAError`) and not inserting the pod back into the scheduling queue if the error is of that type. This would be similar to how we don't relax constraints if the error is of type `ReservedCapacityError`.

pkg/utils/pod/scheduling.go
```diff
@@ -207,3 +207,8 @@ func HasPodAntiAffinity(pod *corev1.Pod) bool {
 		(len(pod.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution) != 0 ||
 			len(pod.Spec.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution) != 0)
 }
+
+// HasDRARequirements returns true if ResourceClaims are defined in the pod spec
+func HasDRARequirements(pod *corev1.Pod) bool {
```
jmdeal, line 212: This is probably fine in practice, but technically we shouldn't fail scheduling just because `ResourceClaims` are defined for the pod. Really we only care if any of the containers consume them. For correctness sake, I'd check if any containers have requests for claims, rather than just checking if claims exist.

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal, line 1: We should make sure we add a test for this. Should be pretty straightforward - really all it would be is verifying that if we provide a DRA configured pod to the scheduler, that we don't create a NodeClaim for it and we get a pod error back.

jmdeal, line 1: I am wondering if we want to consider putting this behind a DRA feature gate? This provides the best of both worlds - if users want a backdoor to using DRA with Karpenter while the implementation is still in progress they can disable the native support feature gate. Enabling the feature gate will ensure that Karpenter is DRA compliant, even if at first that just means it doesn't take any action for pods with DRA requests

pkg/operator/options/options.go
```diff
@@ -60,6 +60,7 @@ type FeatureGates struct {
 	ReservedCapacity        bool
 	SpotToSpotConsolidation bool
 	NodeOverlay             bool
+	SkipDRAScheduling       bool
```
jmdeal: I wouldn't add this as a feature gate, I'd just add it as an additional top-level option. My rationale is that this is not a feature we're graduating. E.g. I'm thinking something like this:

`IGNORE_DRA_REQUESTS`: When set, Karpenter will ignore pods' DRA requests during scheduling simulations. NOTE: This flag will be removed once formal DRA support is GA in Karpenter.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -185,6 +185,7 @@ type PodData struct {
 	Requests           corev1.ResourceList
 	Requirements       scheduling.Requirements
 	StrictRequirements scheduling.Requirements
+	IsDRAEnabled       bool
```
jmdeal: nit: I'd consider renaming this to something like this - DRA itself is a global feature so I think this is a little clearer, but lmk what you think.
```suggestion
	HasResourceClaimRequests       bool
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -222,7 +223,7 @@ type Results struct {
 func (r Results) Record(ctx context.Context, recorder events.Recorder, cluster *state.Cluster) {
 	// Report failures and nominations
 	for p, err := range r.PodErrors {
-		if IsReservedOfferingError(err) {
+		if IsReservedOfferingError(err) || IsDRAError(err) {
```
jmdeal: I think we'd want to log / event for DRA errors, right? Otherwise it may be unclear why Karpenter is ignoring the pod.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -78,6 +78,25 @@ func (e ReservedOfferingError) Unwrap() error {
...
+}
+
+func IsDRAError(err error) bool {
+	draErr := &DRAError{}
+	return errors.As(err, draErr)
+}
+
+func (e DRAError) Unwrap() error {
+	return e.error
+}
```
jmdeal: I'd move this to scheduler.go since that's where we're actually constructing the error

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -409,11 +416,11 @@ func (s *Scheduler) trySchedule(ctx context.Context, p *corev1.Pod) error {
 		if err == nil {
 			return nil
 		}
-		// We should only relax the pod's requirements when the error is not a reserved offering error because the pod may be
+		// We should only relax the pod's requirements when the error is not a reserved offering error or DRA error because the pod may be
```
jmdeal: I wouldn't add it on in this sentence because the only part that applies is that we "shouldn't relax the pod's requirements". I'd either add another paragraph at the end or use a separate if statement with its own brief blurb.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -445,10 +452,17 @@ func (s *Scheduler) updateCachedPodData(p *corev1.Pod) {
...
 		Requirements:       requirements,
 		StrictRequirements: strictRequirements,
+		IsDRAEnabled:       pod.HasDRARequirements(p),
 	}
 }
 
 func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
+	// Check if pod has DRA requirements - if so, return DRA error when SkipDRAScheduling is enabled
+	if s.cachedPodData[pod.UID].IsDRAEnabled && karpopts.FromContext(ctx).FeatureGates.SkipDRAScheduling {
+		log.FromContext(ctx).V(1).WithValues("Pod", klog.KObj(pod)).Info("skipping pod scheduling, Dynamic Resource Allocation (DRA) is not yet supported by Karpenter")
```
jmdeal: Since we're returning the error we don't need to add the log here. This will be logged here: https://github.com/kubernetes-sigs/karpenter/blob/218cca8f9a96046b356addb2d91cc42bf205e275/pkg/controllers/provisioning/scheduling/scheduler.go#L228-L229

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -4597,6 +4597,115 @@ var _ = Context("Scheduling", func() {
...
 		})
 	})
+
+	Describe("Dynamic Resource Allocation (DRA)", func() {
+		It("should not create NodeClaims for DRA pods and return DRA error", func() {
+			nodePool := test.NodePool()
+			ExpectApplied(ctx, env.Client, nodePool)
+
+			// Create a pod with DRA requirements (container consumes ResourceClaims)
+			DRAEnabledPod := &corev1.Pod{
```
jmdeal: nit: variable naming should be camel case
```suggestion
			draEnabledPod := &corev1.Pod{
```

jmdeal: You should be able to use the `test.Pod` function to generate this, and just overwrite the properties that you need to. You may need to extend it to accept ResourceClaims.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -4597,6 +4597,115 @@ var _ = Context("Scheduling", func() {
 			}
 		})
 	})
+
+	Describe("Dynamic Resource Allocation (DRA)", func() {
```
jmdeal, line 4630: Let's also include a test to verify that we ignore pods which specify resource claim requests for an init container. You could use a `DescribeTable` to avoid duplication between the standard and init container tests.

pkg/controllers/provisioning/scheduling/scheduler.go
jmdeal, line 1: Do we also need to check if any daemonsets have requests for resource claims? AFAIK it's supported. We'd probably just want to exclude the daemonset when computing daemonset overhead if it has any requests.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -4597,6 +4597,115 @@ var _ = Context("Scheduling", func() {
...
+			Expect(results.PodErrors).To(HaveKey(DRAEnabledPod))
+			err, exists := results.PodErrors[DRAEnabledPod]
+			Expect(exists).To(BeTrue())
+			Expect(scheduling.IsDRAError(err)).To(BeTrue())
+			Expect(err.Error()).To(ContainSubstring("Dynamic Resource Allocation"))
+
+			// Verify DRAErrors() filtering works correctly
+			draErrors := results.DRAErrors()
+			Expect(draErrors).To(HaveLen(1))
+			Expect(draErrors).To(HaveKey(DRAEnabledPod))
```
jmdeal: Would it be simpler to just do this check rather than the check on PodErrors?

alimaazamat:
@jmdeal 
To use CPU for daemon overhead tests we have to wrap internal function `getDaemonOverhead` and expose it for testing. I dont think this is necessarily bad but I don't see another existing pattern of` suit_test.go` testing with internal functions.
pro: directly compares cpu
con: needs to have `getDaemonOverhead` function be public 

Instead of doing this we could do what you mentioned about checking to see if daemonset pod would land on small instance (IgnoreDRARequests enabled) or large instance (IgnoreDRARequests disabled). 
pro: doesn't need to expose `getDaemonOverhead` function
con: kinda roundabout way of comparing instances

I think keeping second option is better.

jmdeal:
Nice work! Looks like we need a rebase, but lgmt otherwise.
/lgtm
/approve

PR #2392, opened 2025-07-22
Overlay instance type in the provisioning and disruption scheduling path

engedaam:
Fixes #N/A 

**Description**

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -105,8 +105,7 @@ type InstanceType struct {
 	Capacity corev1.ResourceList
 	// Overhead is the amount of resource overhead expected to be used by kubelet and any other system daemons outside
 	// of Kubernetes.
-	Overhead *InstanceTypeOverhead
-
+	Overhead    *InstanceTypeOverhead
 	once        sync.Once
 	allocatable corev1.ResourceList
 }
@@ -117,6 +116,17 @@ type InstanceTypes []*InstanceType
 // and the operation is fairly expensive.
 func (i *InstanceType) precompute() {
 	i.allocatable = resources.Subtract(i.Capacity, i.Overhead.Total())
+}
+
+func (i *InstanceType) IsPricingOverlayApplied() bool {
+	_, found := lo.Find(i.Offerings, func(of *Offering) bool {
+		return of.IsAnOverlayApplied()
+	})
+	return found
+}
+
+func (i *InstanceType) ResetAllocatable() {
+	i.once = sync.Once{}
 }
 
 func (i *InstanceType) Allocatable() corev1.ResourceList {
@@ -261,6 +271,16 @@ type Offering struct {
 	Price               float64
 	Available           bool
 	ReservationCapacity int
+
+	overlayApplied bool
+}
+
+func (o *Offering) ApplyOverlay() {
+	o.overlayApplied = true
+}
+
+func (o *Offering) IsAnOverlayApplied() bool {
+	return o.overlayApplied
 }
 
 func (o *Offering) CapacityType() string {
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -40,6 +40,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	operatorlogging "sigs.k8s.io/karpenter/pkg/operator/logging"
 	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
+	"sigs.k8s.io/karpenter/pkg/utils/nodeoverlay"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
 	"sigs.k8s.io/karpenter/pkg/utils/pdb"
 )
@@ -200,7 +201,7 @@ func BuildNodePoolMap(ctx context.Context, kubeClient client.Client, cloudProvid
 	for _, np := range nodePools {
 		nodePoolMap[np.Name] = np
 
-		nodePoolInstanceTypes, err := cloudProvider.GetInstanceTypes(ctx, np)
+		nodePoolInstanceTypes, err := nodeoverlay.GetInstanceTypes(ctx, kubeClient, cloudProvider, np)
 		if err != nil {
 			// don't error out on building the node pool, we just won't be able to handle any nodes that
 			// were created by it
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -53,6 +53,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
+	"sigs.k8s.io/karpenter/pkg/utils/nodeoverlay"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
 	"sigs.k8s.io/karpenter/pkg/utils/pretty"
 )
@@ -245,7 +246,7 @@ func (p *Provisioner) NewScheduler(
 
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	for _, np := range nodePools {
-		its, err := p.cloudProvider.GetInstanceTypes(ctx, np)
+		its, err := nodeoverlay.GetInstanceTypes(ctx, p.kubeClient, p.cloudProvider, np)
 		if err != nil {
 			if errors.Is(err, context.DeadlineExceeded) {
 				return nil, fmt.Errorf("getting instance types, %w", err)
--- a/pkg/utils/nodeoverlay/nodeoverlay.go
+++ b/pkg/utils/nodeoverlay/nodeoverlay.go
@@ -1,0 +1,144 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nodeoverlay
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/mitchellh/hashstructure/v2"
+	"github.com/samber/lo"
+	lop "github.com/samber/lo/parallel"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+func GetInstanceTypes(ctx context.Context, kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, np *v1.NodePool) ([]*cloudprovider.InstanceType, error) {
+	its, err := pullInstanceTypes(ctx, cloudProvider, np)
+	if err != nil {
+		return []*cloudprovider.InstanceType{}, err
+	}
+	overlay := &v1alpha1.NodeOverlayList{}
+	err = kubeClient.List(ctx, overlay)
+	if err != nil {
+		return []*cloudprovider.InstanceType{}, err
+	}
+
+	return OverlayInstanceTypes(overlay.Items, its), nil
+}
+
+// Explain weights. That will be hard to understand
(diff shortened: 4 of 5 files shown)


pkg/cloudprovider/types.go
```diff
@@ -117,6 +118,39 @@ type InstanceTypes []*InstanceType
 // and the operation is fairly expensive.
 func (i *InstanceType) precompute() {
 	i.allocatable = resources.Subtract(i.Capacity, i.Overhead.Total())
+
+	// subtract the amount of memory that will be allocatable for
+	// application based on hugepage reservations
+	for name, quantity := range i.Capacity {
```
jonathan-innis, line 127: nit: Let's add a comment around this since this is a very special case that we are handling in a specifically different way when it comes to allocatable

pkg/cloudprovider/types.go
```diff
@@ -117,6 +118,39 @@ type InstanceTypes []*InstanceType
...
+		if strings.HasPrefix(string(name), corev1.ResourceHugePagesPrefix) {
+			i.allocatable.Memory().Sub(quantity)
```
jonathan-innis: Is this line necessary since you are doing the exact same things in the following lines? Does this update the value in-place?

pkg/cloudprovider/types.go
```diff
@@ -117,6 +118,39 @@ type InstanceTypes []*InstanceType
...
+			current := i.allocatable.Memory()
+			current.Sub(quantity)
+			i.allocatable[corev1.ResourceMemory] = lo.FromPtr(current)
+		}
+	}
+	// If there there are hugepages resource that consume all the memory on an instance
```
jonathan-innis:
```suggestion
	// If there are hugepages resources that consume all the memory on an instance
```

pkg/cloudprovider/types.go
```diff
@@ -117,6 +118,39 @@ type InstanceTypes []*InstanceType
...
+	// we will cap the memory used to being zero as negative memory does not make sense.
+	if i.allocatable.Memory().Sign() == -1 {
```
jonathan-innis: nit: You could probably do this with a conditional above with on L128 with

```
current.Sub(quantity)
if current.Sign() == -1 {
  current.Set(0)
}
```

pkg/utils/nodeoverlay/nodeoverlay.go
```diff
@@ -0,0 +1,186 @@
...
+		its, err = OverlayInstanceTypes(ctx, kubeClient, its)
+		if err != nil {
+			return []*cloudprovider.InstanceType{}, err
+		}
+		removeNodePoolRequirements(np, its)
+	}
+	return its, nil
+}
+
+func OverlayInstanceTypes(ctx context.Context, kubeClient client.Client, instanceTypes []*cloudprovider.InstanceType) ([]*cloudprovider.InstanceType, error) {
```
jonathan-innis: Given the number of times that we call GetInstanceTypes(), I'm not sure that we can afford to check compatability every single time (this is going to get really, really inefficient) -- is there a way that we can pre-compute this data (maybe in the controller) to have live compatibility information ready to go and then we can do a quick look-up to overlay the data?

In general, I think we're going to have to find a way to compute the overlay information out-of-band so that our calls to GetInstanceTypes() aren't so expensive that we grind everything to a halt -- this potentially may require us to have to cache on the upstream side of the codebase rather than the downstream side of the codebase as well

pkg/cloudprovider/types.go
```diff
@@ -117,6 +118,39 @@ type InstanceTypes []*InstanceType
 // and the operation is fairly expensive.
 func (i *InstanceType) precompute() {
```
jonathan-innis, line 119: Have we considered creating a decorator for `GetInstanceTypes()` that handles overlays? That will make it so any caller to `GetInstanceTypes()` doesn't have to think about applying the overlay every time

pkg/controllers/disruption/helpers.go
```diff
@@ -200,7 +201,7 @@ func BuildNodePoolMap(ctx context.Context, kubeClient client.Client, cloudProvid
 	for _, np := range nodePools {
 		nodePoolMap[np.Name] = np
 
-		nodePoolInstanceTypes, err := cloudProvider.GetInstanceTypes(ctx, np)
+		nodePoolInstanceTypes, err := nodeoverlay.GetInstanceTypes(ctx, kubeClient, cloudProvider, np)
```
jonathan-innis: IMO this is going to lead to bugs since you now have two packages that have the same function name -- it strikes me that someone is going to call the wrong package at some point and then we are going to get bad info

pkg/controllers/provisioning/provisioner.go
```diff
@@ -245,7 +246,7 @@ func (p *Provisioner) NewScheduler(
 
 	instanceTypes := map[string][]*cloudprovider.InstanceType{}
 	for _, np := range nodePools {
-		its, err := p.cloudProvider.GetInstanceTypes(ctx, np)
+		its, err := nodeoverlay.GetInstanceTypes(ctx, p.kubeClient, p.cloudProvider, np)
```
jonathan-innis: nit: same comment here about potentially introducing bugs with this setup

jonathan-innis:
Can we post performance benchmarking on these code changes in this PR as well? Given the critical path that this is touching, we need to be sure that this isn't regressing our scheduling code or drastically increasing our memory/CPU utilization

pkg/cloudprovider/types.go
engedaam: I like the idea of creating a decorator for the cloud provider and have switched to that approach

pkg/controllers/disruption/helpers.go
engedaam: dropping the function in favor of using a decorator

pkg/utils/nodeoverlay/nodeoverlay.go
engedaam: > In general, I think we're going to have to find a way to compute the overlay information out-of-band so that our calls to GetInstanceTypes() aren't so expensive that we grind everything to a halt -- this potentially may require us to have to cache on the upstream side of the codebase rather than the downstream side of the codebase as well

The idea of computing overlay information out-of-band was actually my initial approach (see PR [#2381](https://github.com/kubernetes-sigs/karpenter/pull/2381)). However, I moved away from upstream caching due to data freshness concerns. Caching instance type data upstream could lead to staleness issues, particularly for dynamic attributes like unavailable or reserved offerings. This staleness could trigger unnecessary churn and negatively impact scheduler performance.

> Given the number of times that we call GetInstanceTypes(), I'm not sure that we can afford to check compatibility every single time (this is going to get really, really inefficient) 

Regarding the performance impact of frequent GetInstanceTypes() calls with compatibility checks - while this approach is computationally expensive, I believe the downsides of stale state data would be more problematic in the long term. I'll run performance tests to gather comparative data. One potential compromise might be storing the requirements, though this would need careful validation to ensure we maintain data accuracy while improving performance.

pkg/cloudprovider/overlay/cloudprovider.go
```diff
@@ -0,0 +1,197 @@
...
+// Such that we only overlay currently defined set of overrides
+func pullInstanceTypes(ctx context.Context, cp cloudprovider.CloudProvider, nodePool *v1.NodePool) ([]*cloudprovider.InstanceType, error) {
+	its, err := cp.GetInstanceTypes(ctx, nodePool)
+	if err != nil {
+		return []*cloudprovider.InstanceType{}, err
+	}
+
+	// Since we are altering the instance type data we need to copy the instance types into a new struct
+	// The main things that have been copied are the offerings, capacity and overhead
+	return lo.Map(its, func(it *cloudprovider.InstanceType, _ int) *cloudprovider.InstanceType {
```
jmdeal, line 138: I'm a little concerned about the memory / performance overhead here - have you performed any scale testing yet? Do we need to consider a layer of caching here in addition to in the cloudprovider?

pkg/cloudprovider/overlay/cloudprovider.go
```diff
@@ -0,0 +1,197 @@
...
+		for _, it := range its {
+			if !it.Requirements.IsCompatible(overlaySelector) {
+				continue
+			}
+			if _, ok := overriddenInstanceType[it.Name]; !ok {
+				overriddenInstanceType[it.Name] = []string{}
+			}
+
+			for _, of := range it.Offerings {
+				offeringRequirementsHash := fmt.Sprint(lo.Must(hashstructure.Hash(of.Requirements.String(), hashstructure.FormatV2, &hashstructure.HashOptions{})))
```
jmdeal: nit: rather than indexing into the map with an instance type name and the hash, would it be better to just include the it name in the hash and have a single dimensional set?

pkg/cloudprovider/types.go
```diff
@@ -261,6 +293,16 @@ type Offering struct {
...
 	ReservationCapacity int
+
+	overlayApplied bool
+}
+
+func (o *Offering) ApplyOverlay() {
+	o.overlayApplied = true
+}
+
+func (o *Offering) IsAnOverlayApplied() bool {
```
jmdeal: nit: naming, similar suggestion applies elsewhere
```suggestion
func (o *Offering) IsOverlaid() bool {
```

pkg/cloudprovider/overlay/cloudprovider.go
```diff
@@ -0,0 +1,197 @@
...
+	sort.Slice(validOverlays, func(x int, y int) bool {
+		return lo.FromPtr(validOverlays[x].Spec.Weight) > lo.FromPtr(validOverlays[y].Spec.Weight)
+	})
+	work := []func(overlays []v1alpha1.NodeOverlay, its []*cloudprovider.InstanceType){
+		overlayCapacityOnInstanceTypes,
+		overlayPriceOnInstanceTypes,
+	}
+	lop.ForEach(work, func(f func(overlays []v1alpha1.NodeOverlay, its []*cloudprovider.InstanceType), i int) {
+		f(validOverlays, instanceTypes)
+	})
```
jmdeal: nit: we don't really use this `ForEach` helper elsewhere, and I'm not sure it provides much value over a native for each loop. What I've found in practice is it just makes debugging more difficult due to the additional closure
```suggestion
	for _, f := range []func(overlays []v1alpha1.NodeOverlay, its []*cloudprovider.InstanceType){
		overlayCapacityOnInstanceTypes,
		overlayPriceOnInstanceTypes,
	} {
	    f(validOverlays, instanceTypes)
	}
```

jmdeal: I would consider collapsing these two functions though since it requires we double our instance compatibility checks.

pkg/cloudprovider/overlay/cloudprovider.go
```diff
@@ -0,0 +1,197 @@
...
+				it.Capacity = lo.Assign(it.Capacity, overlay.Spec.Capacity)
+				it.ApplyResourceOverlay()
+			}
+		}
+	}
+}
+
+func overlayPriceOnInstanceTypes(overlays []v1alpha1.NodeOverlay, its []*cloudprovider.InstanceType) {
+	// we will only update instance types once
+	overriddenInstanceType := map[string][]string{}
```
jmdeal: Why do we need this check here, but not for capacity. In both cases shouldn't validation have ensured that there aren't any conflicts?

pkg/cloudprovider/overlay/cloudprovider.go
```diff
@@ -0,0 +1,197 @@
...
+	err := kubeClient.List(ctx, overlays)
+	if err != nil {
+		return []*cloudprovider.InstanceType{}, err
+	}
+
+	// Only apply overlays that have passed runtime validation and know there have not been
+	// any updates to the spec of the overlay field
+	validOverlays := lo.Filter(overlays.Items, func(o v1alpha1.NodeOverlay, _ int) bool {
+		condition := o.StatusConditions().Get(v1alpha1.ConditionTypeValidationSucceeded)
+		return condition.IsTrue() && condition.ObservedGeneration == o.Generation
```
jmdeal, line 78: I brought this up in the validation PR as well, but I'm wondering if we need some concept of "the observed generation of the group of overlays". I'm thinking about the scenario where we only successfully update the status for a subset of overlays, so there may be overlays with conflicts that we're unaware of. Without that group observed generation construct, we wouldn't be able to treat the conditions as a source of truth here (at least if I've thought it through correctly).

jmdeal (commented):
Checkpointing, haven't had a chance to go through tests yet

pkg/cloudprovider/overlay/cloudprovider.go
engedaam: This function is using `lop "github.com/samber/lo/parallel"` runs both function in parallel. You could do that in a for loop, but it made sense to parallelize  the call.  I do think it's fine to just combine these functions.

pkg/cloudprovider/overlay/cloudprovider.go
engedaam: Validation handles cases for clear conflicts where the weight is not defined, however for weight difference we will only apply at the overlays with the higher weight. I have updated this function to drop the overlay capacity function, so this does not matter anymore

pkg/cloudprovider/overlay/cloudprovider.go
engedaam: I like that idea and it would be cleaner. Updating the function

PR #2397, opened 2025-07-24
docs: Manual Node Provisioning RFC

raj-dhamsaniya:
Fixes
- Manual Node Provisioning for Karpenter - [kubernetes-sigs/karpenter/issues/749]

**Description**

This RFC plans to address Manual Node Provisioning feature request by proposing a new Overprovisioner API.

### DIFF
--- a/designs/manual-node-provisioning.md
+++ b/designs/manual-node-provisioning.md
@@ -88,6 +88,7 @@ The state field dictates how the overprovisioner controller manages the node lif
 
 This offers a balance between the speed of Running and the cost of Stopped, though it has more cloud-provider-specific prerequisites. When a pod needs capacity from a Stopped or Hibernated pool, the overprovisioner controller will start the instance, wait for it to join the cluster, and then make it available for scheduling.
 
+Since all the nodes are managed by NodeClaim CRD, we need to add a filed in the CRD to make sure it recognise which kind of state a node is in. We can add a lable to the nodeclaim which indicates wheather its hot, cold, or warm state node and based on that the consolidation or scaleup decisions can be made. Along with the label, we would be required to add some taints on the nodes so that during the initial startup, it does not schedule actual workload onto the same.
 
 ### Strategies Explained
 The strategy field provides high-level control over the behavior of the overprovisioning logic.
@@ -111,7 +112,7 @@ The controller will use `ceil()` to round up, ensuring at least one spare node i
 
 One of the alternatives suggested was to extend the NodePool API directly[1]. This was rejected because it conflates two distinct responsibilities: defining the properties of a pool of capacity (NodePool) and defining the proactive management strategy for that capacity (Overprovisioner). A separate CRD provides a cleaner separation of concerns, a more intuitive user experience, and avoids creating an overly complex, monolithic API object. The common "pause pod" workaround was also considered and rejected as it is an operationally burdensome hack and not cost effective solve for larger spike.
 
-### Possible implementation using CRDs
+### Possible implementation using current nodepool CRDs
 
 Here is the CRD mentioned which includes the overprovisioning in the existing definition. In this case, if there is a same overprovisioning config required for all the nodepools, the same config needs to be copied and managed as its part of the same CRD. This makes it a bit complex to manage them as well as it would be more error prone if the overprovisioning gets more complex with diff type of node state, stratagy, sizing and schedules. However, there is some upside in terms of some shared core functionality could be used, however, future enhancement could be hampered if this adds any other requirements.  
 ```apiVersion: karpenter.sh/v1beta1


engedaam:
Hi @raj-dhamsaniya, have you had a chance to look at this PR? It's looking to solve the same problem you meantion in your RFC: https://github.com/kubernetes-sigs/karpenter/pull/2309 Any chance you can take a look at that first and see if it fits your use case. This RFC propose to support manual provisioning without a new CRD

raj-dhamsaniya:
Yes @engedaam, I have gone through the proposal and mentioned the same in the ref as well. The proposal suggest to use the same CRD and just have a solve for the active nodes only which is not the only goal we have. IMHO, Having a separate CRD would help in improving it for future use cases and keep things simple. If you see the proposal in the CAS ([doc](https://docs.google.com/document/d/1bcct-luMPP51YAeUhVuFV7MIXud5wqHsVBDah9WuKeo/edit?tab=t.0)), it seems to be pointing to the similar behavior. 
If needed, we can still keep everything in the same CRD, improve upon the proposed solve in #2309 , but defining multiple over-provisioning like hot and cold for the same nodepool with diff config might make the nodepool definition more complex to maintain, which would not be the case for separate CRD.
Please go over the use case and suggest if we think its best to update the current nodepool definition, if so, I can raise some changes in that PR itself.

designs/manual-node-provisioning.md
```diff
@@ -0,0 +1,111 @@
...
+    - **Cost Efficiency:** Avoiding the cost of maintaining a buffer during periods of no activity. The buffer grows and shrinks with the cluster's usage.  
+
+When using KeepEmpty, the sizing.percentage field is most appropriate, as it creates a dynamic buffer that adapts to the cluster's scale.  
+The controller will use `ceil()` to round up, ensuring at least one spare node is created for any non-zero percentage in a running cluster.
+
+## Alternatives Considered
+
+One of the alternatives suggested was to extend the NodePool API directly[1]. This was rejected because it conflates two distinct responsibilities: defining the properties of a pool of capacity (NodePool) and defining the proactive management strategy for that capacity (Overprovisioner). A separate CRD provides a cleaner separation of concerns, a more intuitive user experience, and avoids creating an overly complex, monolithic API object. The common "pause pod" workaround was also considered and rejected as it is an operationally burdensome hack and not cost effective solve for larger spike.
+
```
Bryce-Soghigian, line 114: Was the nodeclaim CRD considered? You can provision nodes by just applying nodeclaim CRs to the cluster. Can you write up an option about that and making the nodeclaim crd manual provisioning a first class concept?  Or managing nodeclaim crds via some management crds?

raj-dhamsaniya: Thanks for the review and suggestion @Bryce-Soghigian. I have updated the implementation details for the same CRD approach. Please do mention if anything else you think might be required to be added.

zdymjx1998:
> Fixes
> 
> * Manual Node Provisioning for Karpenter - [[Mega Issue: Manual node provisioning #749](https://github.com/kubernetes-sigs/karpenter/issues/749)]
> 
> **Description**
> 
> This RFC plans to address Manual Node Provisioning feature request by proposing a new Overprovisioner API.
> 
> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Hello, I read your design and it looks to be helpful a lot to my user case. In my user case, I will collect the number of pods or total CPU/Memory consumption and pass to Karpenter, would like karpenter provides nodes based on the resource consumption/pod numbers. Is that possible that your design be adjusted somehow to consider number of pods or total resources (CPU+Memory) instead of nodes for overprovissioning?

zdymjx1998:
Hello, I read your design and it looks to be helpful a lot to my user case. In my user case, I will collect the number of pods or total CPU/Memory consumption and pass to Karpenter, would like karpenter to warm up/overprovision nodes based on the resource consumption/pod numbers. Is that possible that your design to be adjusted somehow to consider the number of pods or total resources (CPU+Memory) instead of nodes for overprovissioning?  Below is my suggestion:

```yaml
  # Overprovisioning definition in the same CRD. the fields represents same usage as mentioend fpr separate CRD.
  overprovision:
  - state: Running
    stratagy: KeepMinimum
    sizing:
      count: 10
    scheduledOverrides:
    - schedule: "0 9 * * mon-fri"
      duration: 8h
      # adding fields to enable cpu & memory consumption, or number of pods to be input\
      # user need to input either cpu & memory consumption, or number of pods for provision of nodes
      sizing:
        count: "0"
        pods: "10" # total number of pods, calculated by using the average pod cpu/memory in the cluster
        cpu: "2"  # total cpu consumption
        memory: "512MB"  # total memory
  - state: Stopped
    stratagy: KeepEmpty
    sizing:
      percentage: 20

```

designs/manual-node-provisioning.md
```diff
@@ -0,0 +1,194 @@
...
+Furthermore, the strategy for maintaining this buffer capacity varies. Some users need a constant, guaranteed number of spare nodes, while others prefer a dynamic buffer that scales proportionally with the cluster's total size. Attempting to shoehorn this rich set of requirements into the existing NodePool API would lead to an overly complex and bloated interface. A dedicated Overprovisioner API provides the necessary clarity and expressive power to manage these advanced strategies effectively.
+
+## Proposal: The Overprovisioner CRD
+We propose the creation of a new Custom Resource Definition (CRD) named Overprovisioner. This CRD will target an existing NodePool to define the type of capacity to be overprovisioned, while the Overprovisioner resource itself will define the quantity, state, and strategy for that spare capacity.
+### API DefinitionYAML
+```apiVersion: karpenter.sh/v1beta1
+kind: Overprovisioner
```
DerekFrank, line 32: I fail to understand what benefit a new API provides that can't be solved with Buffers, static capacity, and horizontal pod autoscaling

raj-dhamsaniya:
> Hello, I read your design and it looks to be helpful a lot to my user case. In my user case, I will collect the number of pods or total CPU/Memory consumption and pass to Karpenter, would like karpenter to warm up/overprovision nodes based on the resource consumption/pod numbers. Is that possible that your design to be adjusted somehow to consider the number of pods or total resources (CPU+Memory) instead of nodes for overprovissioning? Below is my suggestion:
> 
> ```yaml
>   # Overprovisioning definition in the same CRD. the fields represents same usage as mentioend fpr separate CRD.
>   overprovision:
>   - state: Running
>     stratagy: KeepMinimum
>     sizing:
>       count: 10
>     scheduledOverrides:
>     - schedule: "0 9 * * mon-fri"
>       duration: 8h
>       # adding fields to enable cpu & memory consumption, or number of pods to be input\
>       # user need to input either cpu & memory consumption, or number of pods for provision of nodes
>       sizing:
>         count: "0"
>         pods: "10" # total number of pods, calculated by using the average pod cpu/memory in the cluster
>         cpu: "2"  # total cpu consumption
>         memory: "512MB"  # total memory
>   - state: Stopped
>     stratagy: KeepEmpty
>     sizing:
>       percentage: 20
> ```

@zdymjx1998 seems some understanding/writing gap, but when I say 10% over provision, it means 10% of current nodepool resource count (cpu and memory). In that way, I believe the avg pod usage is already incorporated. But do provide if I'm understanding it wrong.

designs/manual-node-provisioning.md
raj-dhamsaniya: @DerekFrank Having a separate CRD would help in improving it for future use cases and keep things simple. If you see the proposal in the CAS ([doc](https://docs.google.com/document/d/1bcct-luMPP51YAeUhVuFV7MIXud5wqHsVBDah9WuKeo/edit?tab=t.0)), it seems to be pointing to the similar behavior.
If needed, we can still keep everything in the same CRD, improve upon the proposed solve in https://github.com/kubernetes-sigs/karpenter/pull/2309 , but defining multiple over-provisioning like hot and cold for the same nodepool with diff config might make the nodepool definition more complex to maintain, which would not be the case for separate CRD.

raj-dhamsaniya: @DerekFrank I'm not sure if Buffer is available(if its the case do lmk), the static capacity is just for the active additional capacity, that too on the basis of number of nodes and not dynamic. HPA not sure how it helps is large spikes as it can ask for scale up of pods but what we want is to increase the speed of scale up of nodes keeping cost in check so stopped instance do help there.

designs/manual-node-provisioning.md
raj-dhamsaniya: @Bryce-Soghigian added info on the nodeclaim crd now. https://github.com/kubernetes-sigs/karpenter/pull/2397/files#diff-69f6e618d321eef3aa550b3e1c9f6547345c899c539c3e45c7f8421758cc9045R91 

it would have a label by which we can identify the nodes to be consolidate or to use for next scaleup.

designs/manual-node-provisioning.md
DerekFrank: > improving it for future use cases

I'd rather not make new CRDs for usecases that don't exist yet.

> keep things simple 

Additional configuration surface is almost always the least simple option

> I'm not sure if Buffer is available

Its currently being proposed and implemented, I don't think we should create a new api simply because we are waiting for the creation of a different api, unless there are good reasons will want the overprovisioner api after buffers is supported

raj-dhamsaniya: @DerekFrank If buffer is being discussed, we can move this to buffer as well, its just a name change I believe, so if buffer provides the support for the stopped or hibernated nodes along with active nodes, it works for the faster scaleup setup. This would have same merit as buffer to be a separate API rather than being a part of the nodepool config. So no issues using the same or separate API.

If you have any thread of issue link where this is being discussed for karpenter, can you please share that? I would further comment there and propose the improvement there only!
I have gone through these https://github.com/kubernetes/autoscaler/pull/8151/files but those seems to be for the cluster autoscaler and not karpenter. Please correct me if my understanding is wrong.

DerekFrank: We are planning to support https://github.com/kubernetes/autoscaler/pull/8151/files in Karpenter. In fact, let me make an issue for that

DerekFrank: https://github.com/kubernetes-sigs/karpenter/issues/2571

raj-dhamsaniya: @DerekFrank But isn't it the same as mega issue https://github.com/kubernetes-sigs/karpenter/issues/749. The issue talks about having Over-provisioned/Buffer capacity that helps with faster scaling. it also talks about having warm pool of nodes if cloud provider supports it. Not sure on what is missing that is talked in mega issue and if anything is there, why can't we extend the buffer api concept to help with the closer.

DerekFrank: Nope, #749 is a combination of a couple independent features:

- Static Capacity
- Buffer Api
- Warm pooling inactive nodes

More of the usecases in that mega issue are solved by static capacity than are solved by buffers and warmpooling. We still want to do all three, but its good to disambiguate

PR #2402, opened 2025-07-25
better not use the common k8s label, or use it from nodeclaim

munger1985:
**Description**

I tested in my own env.

### DIFF
--- a/pkg/controllers/disruption/types.go
+++ b/pkg/controllers/disruption/types.go
@@ -106,7 +106,11 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 		return nil, serrors.Wrap(fmt.Errorf("nodepool not found"), "NodePool", klog.KRef("", nodePoolName))
 	}
 	// We only care if instanceType in non-empty consolidation to do price-comparison.
-	instanceType := instanceTypeMap[node.Labels()[corev1.LabelInstanceTypeStable]]
+	// becuz some other cloud k8s will use the label from the node, and can not change the label value,
+	// but the label in the nodeclaim is only controlled by karpenter, so we need to get this label from nodeclaim.
+	// or I think the best is to make use of a karpenter specific label
+	nodeClaimInstanceType := node.NodeClaim.Labels[corev1.LabelInstanceTypeStable]
+	instanceType := instanceTypeMap[node.Labels()[nodeClaimInstanceType]]
 	if pods, err = node.ValidatePodsDisruptable(ctx, kubeClient, pdbs); err != nil {
 		// If the NodeClaim has a TerminationGracePeriod set and the disruption class is eventual, the node should be
 		// considered a candidate even if there's a pod that will block eviction. Other error types should still cause


munger1985:
if changed to check the label value from nodeclaim, will have no impact to the existing projects.

munger1985:
just one line change

pkg/controllers/disruption/types.go
```diff
@@ -106,7 +106,11 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 		return nil, serrors.Wrap(fmt.Errorf("nodepool not found"), "NodePool", klog.KRef("", nodePoolName))
 	}
 	// We only care if instanceType in non-empty consolidation to do price-comparison.
-	instanceType := instanceTypeMap[node.Labels()[corev1.LabelInstanceTypeStable]]
+	// becuz some other cloud k8s will use the label from the node, and can not change the label value,
+	// but the label in the nodeclaim is only controlled by karpenter, so we need to get this label from nodeclaim.
+	// or I think the best is to make use of a karpenter specific label
+	nodeClaimInstanceType := node.NodeClaim.Labels[corev1.LabelInstanceTypeStable]
+	instanceType := instanceTypeMap[node.Labels()[nodeClaimInstanceType]]
```
Bryce-Soghigian, line 113: Can you provide more details what you are trying to accomplish and why you think this is the right way to solve this? 

If the nodes are created through karpenter are you unable in your cloudprovider to specify the node labels on creation? IMO seems reasonable to assume that nodes are created with `"node.kubernetes.io/instance-type"`.

Its a well known label and our provider provides it on creation: https://github.com/Azure/karpenter-provider-azure/blob/3926770a5221dd0d6292ad4287f495d92f266352/pkg/providers/instancetype/instancetype.go#L139

munger1985: hi, that is one issue we found in OCI implementation, there are flex shapes for vms in OCI, which means one shape like VM.E4.Flex can have different cpu and memory thus different prices, so we need to use a unique label like VM.E4.Flex-"cpu cores"-"memory" for the kind of shape, but unfortunatly for some reason, OCI cloud vendor will enforce the k8s worker node label node.kubernetes.io/instance-type to be like VM.E4.Flex instead of VM.E4.Flex-"cpu cores"-"memory" which cause the worker node randomly being killed as the price randomly changed due to different offerings with different cpus and memory.   the reason why it will not use unique label to select nodes is becuz current karpenter will try to use the common k8s label node.kubernetes.io/instance-type,  it is more open design if karpenter can use its own label like this one  in nodeclaim, the downstream vendor can avoid this conflict usage on the k8s common node label  node.kubernetes.io/instance-type, which is often used and controlled by cloud vendor

munger1985: nodeclaim is controlled and owned by karpenter, will not be modified and enforced by cloud vendors

munger1985: the label exist on node and nodeclaim, currently they have the same values, but the label on nodeclaim can be modified, while this label on node can not be modified on OCI, which caused issue when disrupting  node

munger1985: becuz the current issue is only about disruption, so we can use the label on nodeclaim to select the node to kill as vendor imlplementation can custmize the label if karpenter chose to use the label on nodeclaim(karpneter thing),  for adding new node, no issues at the moment, so we dont have to change the logic on adding new node logic in karpenter

munger1985: it is all about disruption, provision new worker node is not impacted

munger1985: the essence of this design is to do not conflict with the resouces that cloud provider would override, and open more pissibility for downstream implementation as it is sth that only defined and controlled by karpenter.

munger1985:
if we use the common label to select the nodes which needs to be removed, as it is the common label wellknown label for k8s, some cloud provider like oci will use it for keep consistency with their real shapes, which hence we can not modify it in karpenter downstream implementation

pkg/controllers/disruption/types.go
munger1985: > Can you provide more details what you are trying to accomplish and why you think this is the right way to solve this?
> 
> If the nodes are created through karpenter are you unable in your cloudprovider to specify the node labels on creation? IMO seems reasonable to assume that nodes are created with `"node.kubernetes.io/instance-type"`.
> 
> Its a well known label and our provider provides it on creation: https://github.com/Azure/karpenter-provider-azure/blob/3926770a5221dd0d6292ad4287f495d92f266352/pkg/providers/instancetype/instancetype.go#L139

it will have no problem to keep using the label from k8s node for creation, this issue only happens on OCI OKE node disruption.

munger1985 (commented):
pls, check if this design the most easy solution for all the downstream implementation? by default the label value are the same, no matter it is node.kubernetes.io/instance-type from node or from nodeclaim, but when it comes to disruption, it should avoid usage on well-known label node.kubernetes.io/instance-type on node, as it is well-known label, some cp will use it for some other cases and can not be modified.  and it is more clear when we use the label node.kubernetes.io/instance-type  on nodeclaim as nodeclaim is the think karpenter project can control, will not be controlled by cp or any other stuffs.

munger1985:
without this, downstream (karpenter-oci)has no way to  fix the bugs when oci has flex shape, karpenter will keep disrupting the nodes and create the same flex shape with the same cpu and ram which is not necessary.   and my change for this bug is the simplest , just do not use the common label in k8s , so that downstream can control the value and implement sth, and it did not affect other cloud implementations

PR #2406, opened 2025-07-29
perf: concurrent reconciles CPU-based scaling

ryan-mist:
Fixes #N/A

**Description**
* Scale `maxConcurrentReconciles` based on CPU resources.

* `make presubmit`

### DIFF
--- a/pkg/controllers/disruption/queue.go
+++ b/pkg/controllers/disruption/queue.go
@@ -21,7 +21,6 @@ import (
 	stderrors "errors"
 	"fmt"
 	"math"
-	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -48,6 +47,7 @@ import (
 
 	pscheduling "sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling"
 	operatorlogging "sigs.k8s.io/karpenter/pkg/operator/logging"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
@@ -110,10 +110,9 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
 	return queue
 }
 
-func (q *Queue) Register(_ context.Context, m manager.Manager) error {
-	cpuCount := runtime.GOMAXPROCS(0)
-	maxConcurrentReconciles := int(math.Ceil(float64(cpuCount)*15 + 85))
-	maxConcurrentReconciles = lo.Clamp(maxConcurrentReconciles, 100, 1000)
+func (q *Queue) Register(ctx context.Context, m manager.Manager) error {
+	cpuCount := int(math.Ceil(float64(options.FromContext(ctx).CPURequests) / 1000.0))
+	maxConcurrentReconciles := lo.Clamp(15*cpuCount+85, 100, 1000)
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("disruption.queue").
 		WatchesRawSource(source.Channel(q.source, &handler.TypedEnqueueRequestForObject[*v1.NodeClaim]{})).
--- a/pkg/controllers/disruption/queue.go
+++ b/pkg/controllers/disruption/queue.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	stderrors "errors"
 	"fmt"
+	"math"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
@@ -58,10 +60,9 @@ import (
 )
 
 const (
-	queueBaseDelay          = 1 * time.Second
-	queueMaxDelay           = 10 * time.Second
-	maxRetryDuration        = 10 * time.Minute
-	maxConcurrentReconciles = 100
+	queueBaseDelay   = 1 * time.Second
+	queueMaxDelay    = 10 * time.Second
+	maxRetryDuration = 10 * time.Minute
 )
 
 type UnrecoverableError struct {
@@ -110,6 +111,9 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
 }
 
 func (q *Queue) Register(_ context.Context, m manager.Manager) error {
+	cpuCount := runtime.GOMAXPROCS(0)
+	maxConcurrentReconciles := int(math.Ceil(float64(cpuCount)*15 + 85))
+	maxConcurrentReconciles = lo.Clamp(maxConcurrentReconciles, 100, 1000)
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("disruption.queue").
 		WatchesRawSource(source.Channel(q.source, &handler.TypedEnqueueRequestForObject[*v1.NodeClaim]{})).
--- a/pkg/controllers/node/hydration/controller.go
+++ b/pkg/controllers/node/hydration/controller.go
@@ -19,6 +19,8 @@ package hydration
 import (
 	"context"
 	"fmt"
+	"math"
+	"runtime"
 
 	"github.com/awslabs/operatorpkg/reasonable"
 	"github.com/samber/lo"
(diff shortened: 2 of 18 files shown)


rschalo:
Generally, is there a pattern where some values are set vs others? If we think there is, can we find a way to set it once and use it in those controllers? Also, how did we reach these values?

rschalo:
Would also be good to see a flame graph for the default 1cpu vs 4cpu

DerekFrank:
> Generally, is there a pattern where some values are set vs others?

Which values aren't set?

> Also, how did we reach these values?

These are the configs we tested for the 100k EKS testing: [https://aws.amazon.com/blogs/containers/amazon-eks-enables-ultra-scale-ai-ml-workloads-with-support-for-100k-nodes-per-cluster/](https://aws.amazon.com/blogs/containers/amazon-eks-enables-ultra-scale-ai-ml-workloads-with-support-for-100k-nodes-per-cluster/)

We'll need to fine tune these values over time likely as we have more data points, especially at mid scale

pkg/controllers/disruption/queue.go
```diff
@@ -58,10 +60,9 @@ import (
 )
 
 const (
-	queueBaseDelay          = 1 * time.Second
-	queueMaxDelay           = 10 * time.Second
-	maxRetryDuration        = 10 * time.Minute
-	maxConcurrentReconciles = 100
+	queueBaseDelay   = 1 * time.Second
```
DerekFrank, line 62: nit: we should keep the tab spacing from before

pkg/controllers/disruption/queue.go
```diff
@@ -109,7 +110,9 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
 	return queue
 }
 
-func (q *Queue) Register(_ context.Context, m manager.Manager) error {
+func (q *Queue) Register(ctx context.Context, m manager.Manager) error {
+	cpuCount := int(math.Ceil(float64(options.FromContext(ctx).CPURequests) / 1000.0))
+	maxConcurrentReconciles := lo.Clamp(15*cpuCount+85, 100, 1000)
```
DerekFrank: I'd love to see this code decomposed into a helper function in a util file and then re-used.

Similarly, we should avoid these magic numbers and have the min and max put as consts above

DerekFrank (changes requested):

pkg/controllers/disruption/queue.go
ryan-mist: Will do

pkg/utils/reconciles/reconciles.go
```diff
@@ -0,0 +1,41 @@
...
+	"sigs.k8s.io/karpenter/pkg/operator/options"
+)
+
+// cpuCount calculates CPU count in cores from context options (in millicores)
+func cpuCount(ctx context.Context) int {
+	return int(math.Ceil(float64(options.FromContext(ctx).CPURequests) / 1000.0))
+}
+
+// LinearScaleReconciles calculates maxConcurrentReconciles using linear scaling
+func LinearScaleReconciles(ctx context.Context, minReconciles int, maxReconciles int) int {
```
DerekFrank: Now that we have a function, we should have some unit tests that do basic validation

pkg/utils/reconciles/reconciles.go
rschalo, line 1: Let's change this to `controller/controller.go`

ryan-mist: Will do!

pkg/utils/controller/controller.go
```diff
@@ -0,0 +1,41 @@
...
+	"sigs.k8s.io/karpenter/pkg/operator/options"
+)
+
+// cpuCount calculates CPU count in cores from context options (in millicores)
+func cpuCount(ctx context.Context) int {
+	return int(math.Ceil(float64(options.FromContext(ctx).CPURequests) / 1000.0))
+}
+
+// LinearScaleReconciles calculates maxConcurrentReconciles using linear scaling
+func LinearScaleReconciles(ctx context.Context, minReconciles int, maxReconciles int) int {
```
DerekFrank: It might be cleaner to pass in the number of CPUs ahead of time rather than pass ctx down the whole way, but I don't think that's worth blocking on. It would make the unit tests cleaner

pkg/controllers/node/health/controller.go
```diff
@@ -65,10 +73,14 @@ func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudPr
...
 }
 
-func (c *Controller) Register(_ context.Context, m manager.Manager) error {
+func (c *Controller) Register(ctx context.Context, m manager.Manager) error {
 	return controllerruntime.NewControllerManagedBy(m).
 		Named("node.health").
 		For(&corev1.Node{}, builder.WithPredicates(nodeutils.IsManagedPredicateFuncs(c.cloudProvider))).
+		WithOptions(controller.Options{
+			RateLimiter:             reasonable.RateLimiter(),
+			MaxConcurrentReconciles: utilscontroller.LinearScaleReconciles(ctx, minReconciles, maxReconciles),
```
DerekFrank: How do we feel about logging the number of reconcilers chosen? I don't think its going to be super useful, but it'd only be logged once

pkg/utils/controller/suite_test.go
```diff
@@ -0,0 +1,85 @@
...
+		CPURequests: cpuRequests,
+	}
+	return opts.ToContext(context.Background())
+}
+
+var _ = Describe("Reconciles", func() {
+	minReconciles := 10
+	maxReconciles := 1000
+	Context("LinearScaleReconciles Calculations", func() {
+		It("should calculate minReconciles for 0.5 CPU core", func() {
```
DerekFrank: We could turn these into a table if we want to, it'd allow you to test more values in fewer lines. Not worth blocking on i don't think but if you want to!

pkg/controllers/node/health/controller.go
ryan-mist: Good point, added.

pkg/utils/controller/controller.go
ryan-mist: If we passed number of CPUs here each controller would have to call `cpuCount(ctx)` then `LinearScaleReconciles(cpuCount, minReconciles, maxReconciles)` instead of just one function call so it may be cleaner to leave as is right?

DerekFrank: You can inline the function call if you'd like to keep it on the same line. Its a question of 'do you pass the ctx var' or 'do you pass the CPU count'. I prefer the latter, but its 6 in one hand half dozen in the other really

ryan-mist: Makes sense, will do!

pkg/controllers/disruption/queue.go
```diff
@@ -109,7 +111,10 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
 	return queue
 }
 
-func (q *Queue) Register(_ context.Context, m manager.Manager) error {
+func (q *Queue) Register(ctx context.Context, m manager.Manager) error {
+	cpuCount := utilscontroller.CPUCount(ctx)
+	maxConcurrentReconciles := utilscontroller.LinearScaleReconciles(cpuCount, minReconciles, maxReconciles)
+	log.FromContext(ctx).Info("disruption.queue maxConcurrentReconciles set", "maxConcurrentReconciles", maxConcurrentReconciles)
```
jmdeal: nit: let's make this a debug log - it's not spammy but this is truly "debug" info that isn't relevant to most users. This comment applies everywhere we perform this log.
```suggestion
	log.FromContext(ctx).V(1).Info("disruption.queue maxConcurrentReconciles set", "maxConcurrentReconciles", maxConcurrentReconciles)
```

pkg/controllers/disruption/queue.go
```diff
@@ -54,14 +54,16 @@ import (
...
 const (
-	queueBaseDelay          = 1 * time.Second
-	queueMaxDelay           = 10 * time.Second
-	maxRetryDuration        = 10 * time.Minute
-	maxConcurrentReconciles = 100
+	queueBaseDelay   = 1 * time.Second
+	queueMaxDelay    = 10 * time.Second
+	maxRetryDuration = 10 * time.Minute
+	minReconciles    = 100
+	maxReconciles    = 1000
```
jmdeal: I'd consider just inlining these since they're only used by the registration function. Same comment applies elsewhere, but consider this non-blocking.

pkg/controllers/disruption/queue.go
```diff
@@ -109,7 +111,10 @@ func NewQueue(kubeClient client.Client, recorder events.Recorder, cluster *state
 	return queue
 }
 
-func (q *Queue) Register(_ context.Context, m manager.Manager) error {
+func (q *Queue) Register(ctx context.Context, m manager.Manager) error {
+	cpuCount := utilscontroller.CPUCount(ctx)
+	maxConcurrentReconciles := utilscontroller.LinearScaleReconciles(cpuCount, minReconciles, maxReconciles)
```
jmdeal: nit: I'd inline this call. This comment applies everywhere
```suggestion
	maxConcurrentReconciles := utilscontroller.LinearScaleReconciles(utilscontroller.CPUCount(ctx), minReconciles, maxReconciles)
```

pkg/controllers/disruption/queue.go
ryan-mist: Will do

PR #2411, opened 2025-07-29
perf: Disruption Queue Retry Duration Scaling

ryan-mist:
Fixes #N/A

**Description**
* Scale disruption queue retry duration based on queue length.

* `make presubmit`
* unit tests

### DIFF
--- a/pkg/controllers/disruption/queue.go
+++ b/pkg/controllers/disruption/queue.go
@@ -60,6 +60,8 @@ import (
 const (
 	queueBaseDelay          = 1 * time.Second
 	queueMaxDelay           = 10 * time.Second
+	minRetryDuration        = 10 * time.Minute
+	maxRetryDuration        = 1 * time.Hour
 	maxConcurrentReconciles = 100
 )
 
@@ -170,11 +172,10 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) (err error) {
 	q.RLock()
 	numCommands := len(q.providerIDToCommand)
 	q.RUnlock()
-	maxRetryDuration := 10*time.Minute + time.Duration(numCommands)*15*time.Second
-	maxRetryDuration = lo.Clamp(maxRetryDuration, 10*time.Minute, time.Hour)
+	retryDuration := lo.Clamp(time.Duration(numCommands)*30*time.Second, minRetryDuration, maxRetryDuration)
 	// Wrap an error in an unrecoverable error if it timed out
 	defer func() {
-		if q.clock.Since(cmd.CreationTimestamp) > maxRetryDuration {
+		if q.clock.Since(cmd.CreationTimestamp) > retryDuration {
 			err = NewUnrecoverableError(serrors.Wrap(fmt.Errorf("command reached timeout, %w", err), "duration", q.clock.Since(cmd.CreationTimestamp)))
 		}
 	}()
--- a/pkg/controllers/disruption/queue.go
+++ b/pkg/controllers/disruption/queue.go
@@ -60,7 +60,6 @@ import (
 const (
 	queueBaseDelay          = 1 * time.Second
 	queueMaxDelay           = 10 * time.Second
-	maxRetryDuration        = 10 * time.Minute
 	maxConcurrentReconciles = 100
 )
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/disruption/queue.go
```diff
@@ -80,6 +81,11 @@ func IsUnrecoverableError(err error) bool {
 	return stderrors.As(err, &unrecoverableError)
 }
 
+func CalculateRetryDuration(numCommands int) time.Duration {
+	retryDuration := time.Duration(numCommands) * 80 * time.Millisecond
```
rschalo: Why 80ms? Should this be a const?

ryan-mist: Will make a const for it. 

We set retry duration to 1hr when doing the EKS 100k testing for [this blog post](https://aws.amazon.com/blogs/containers/amazon-eks-enables-ultra-scale-ai-ml-workloads-with-support-for-100k-nodes-per-cluster/). 

At this 80ms scale, `retryDuration` will only start to scale up from the baseline of 10min at 7500 commands in the queue and hits the max retry duration of 1hr at 45000 commands.  By setting to 80ms, we scale to this number when queue gets large but ensure there isn't any scaling when running other workflows.

pkg/controllers/disruption/queue_test.go
```diff
@@ -413,5 +413,19 @@ var _ = Describe("Queue", func() {
 			// And expect the nodeClaim and node to be deleted
 			ExpectNotFound(ctx, env.Client, nodeClaim2, node2)
 		})
+		Context("CalculateRetryDuration", func() {
+			DescribeTable("should calculate correct timeout based on queue length",
+				func(numCommands int, expectedDuration time.Duration) {
+					actualDuration := disruption.CalculateRetryDuration(numCommands)
```
DerekFrank: Love that this bypasses the actual reconcile loop to test the function we actually care about

pkg/controllers/disruption/queue.go
```diff
@@ -168,9 +175,13 @@ func (q *Queue) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconci
 // Once the replacements are ready, it will terminate the candidates.
 // nolint:gocyclo
 func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) (err error) {
+	q.RLock()
+	numCommands := len(q.providerIDToCommand)
+	q.RUnlock()
+	retryDuration := CalculateRetryDuration(numCommands)
```
jmdeal: nit: can we just scope this into `CalculateRetryDuration` that way we're also scoping the mutex acquisition / release?

pkg/controllers/disruption/queue.go
```diff
@@ -80,6 +82,11 @@ func IsUnrecoverableError(err error) bool {
 	return stderrors.As(err, &unrecoverableError)
 }
 
+func CalculateRetryDuration(numCommands int) time.Duration {
```
jmdeal: nit: I'd consider a different name for this. `CalculateRetryDuration` sounds like this is computing our requeue period rather than a maximum attempt period. This is just my suggestion, but I'd consider something along these lines:
```suggestion
func GetMaxRetryDuration(numCommands int) time.Duration {
```

I imagine you avoided using "max" since the maximum is dynamic, but I still think it makes more sense because it is the maximum at that given point of time.

pkg/controllers/disruption/queue.go
ryan-mist: Currently, `CalculateRetryDuration` can be tested by plugging in numbers without adding thousands of elements to the queue. If we scope this into it, it would require us to have `CalculateRetryDuration` call another helper function to do what it currently does just for the purpose of testing. Because of this, do you think it is worth it to keep as is?

pkg/controllers/disruption/queue.go
ryan-mist: Will do!

pkg/controllers/disruption/queue.go
jmdeal: I would still opt to refactor and, rather than calling a helper function, update the capacity by enqueuing `n` items into the disruption queue. This more thoroughly tests the end-to-end behavior anyway and is still a sufficiently small "unit" in my opinion.

ryan-mist: I tried this testing in past commits (e.g. https://github.com/kubernetes-sigs/karpenter/pull/2411/commits/e7a55a4b614517777ab4779204ec55c30096e22c) but tests would timeout as we are enqueuing at least `5000` items in order to sufficiently test the scaling

ryan-mist: Refactored

pkg/controllers/disruption/queue.go
```diff
@@ -82,14 +82,17 @@ func IsUnrecoverableError(err error) bool {
 	return stderrors.As(err, &unrecoverableError)
 }
 
-func GetMaxRetryDuration(numCommands int) time.Duration {
+func GetMaxRetryDuration(q *Queue) time.Duration {
```
jmdeal: nit: any reason you opted to make this a standalone function instead of a method? When the only argument a function takes is the method receiver for a bunch of other methods, that stands out to me.

ryan-mist: Good point, IIRC theres not any reason. Will change

PR #2412, opened 2025-07-30
chore: increase NodeClaim lifecycle liveness registration timeout

ryan-mist:
Fixes #N/A

**Description**
*  Increase NodeClaim lifecycle liveness `registrationTimeout` from 15 minutes to 30 minutes

* `make presubmit`

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -44,7 +44,7 @@ type Liveness struct {
 // If we don't see the node within this time, then we should delete the NodeClaim and try again
 
 const (
-	registrationTimeout       = time.Minute * 15
+	registrationTimeout       = time.Minute * 30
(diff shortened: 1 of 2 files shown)


pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -44,7 +44,7 @@ type Liveness struct {
 // If we don't see the node within this time, then we should delete the NodeClaim and try again
 
 const (
-	registrationTimeout       = time.Minute * 15
+	registrationTimeout       = time.Minute * 30
```
DerekFrank, line 47: Might be worth bringing up at a working group to see if folks have a strong issue with this. I have no qualms

jmdeal: +1 to this - I think we can probably close on this at working group this week. We have several strong use-cases for pushing this out by default, so I don't think it will be too controversial.

PR #2415, opened 2025-07-31
chore: ignore pods that have unbound pvc with volumeBindingMode immediate

jigisha620:
Fixes #N/A 

**Description**
Karpenter should ignore pods that have immediate volumeBindingMode within the storageClass but has unbound pvc.

Added unit tests and tested locally by deploying on my dev cluster.


pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3243,7 +3249,6 @@ var _ = Context("Scheduling", func() {
 		})
 		It("should not launch nodes for pod with storageClass that uses an unsupported provisioner", func() {
 			scheduling.UnsupportedProvisioners = lo.Assign(scheduling.UnsupportedProvisioners, sets.New("other-provider"))
-			// Launch an initial pod onto a node and register the CSI Node with a volume count limit of 1
```
jigisha620, line 3246: Dropping this comment because it's not relevant here.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -204,5 +204,9 @@ func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassN
 	if UnsupportedProvisioners.Has(storageClass.Provisioner) {
 		return serrors.Wrap(fmt.Errorf("storageClass provisioner is not supported"), "StorageClass", klog.KRef("", storageClass.Provisioner))
 	}
+	// Ignore pods than have unbound pvc for voluemBindingMode immediate
+	if storageClass.VolumeBindingMode != nil && *storageClass.VolumeBindingMode == storagev1.VolumeBindingImmediate {
+		return serrors.Wrap(fmt.Errorf("persistentVolumeClaim is unbound"), "VolumeBindingMode", klog.KRef("", string(*storageClass.VolumeBindingMode)))
```
rschalo: I think this should reference the storage class and the error message expanded to the failure mode

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,57 @@ var _ = Context("Scheduling", func() {
 			// no nodes should be created as the storage class is using an unsupported provisioner
 			Expect(nodeList.Items).To(HaveLen(0))
 		})
+		It("should not launch nodes for pod with unbound volume for voluemBindingMode immediate", func() {
```
rschalo: Is it also possible to assert the metrics are what we expect?

jigisha620: Good catch, done.

pkg/controllers/provisioning/scheduling/volumetopology.go
jigisha620: Done

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,63 @@ var _ = Context("Scheduling", func() {
...
```
rschalo:
```suggestion
		It("should not launch nodes for pod with unbound volume for volumeBindingMode immediate", func() {
```

jigisha620: Done

rschalo (approved):
LGTM

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -204,5 +204,9 @@ func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassN
 	if UnsupportedProvisioners.Has(storageClass.Provisioner) {
 		return serrors.Wrap(fmt.Errorf("storageClass provisioner is not supported"), "StorageClass", klog.KRef("", storageClass.Provisioner))
 	}
+	// Ignore pods than have unbound pvc for voluemBindingMode immediate
```
jmdeal: I know this function is only ever called if the volume is unbound (checked via `pvc.Spec.VolumeName`), but this still strikes me as the wrong place to make this check. I would perform this check in `ValidatePersistentVolumeClaims` where we've already enforced that invariant. Really, I think we can just inline this function there.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -204,5 +204,9 @@ func (v *VolumeTopology) validateStorageClass(ctx context.Context, storageClassN
...
+	if storageClass.VolumeBindingMode != nil && *storageClass.VolumeBindingMode == storagev1.VolumeBindingImmediate {
+		return serrors.Wrap(fmt.Errorf("persistentVolumeClaim is unbound for volumeBindingMode: Immediate"), "StorageClass", klog.KRef("", storageClass.Provisioner))
```
jmdeal: Object names are typically all lowercase in our logs (see `nodeclaim`, `nodepool`, etc). Also, this log includes a key for StorageClass, but the value is provisioner. 

Note: this suggestion also only really applies if we inline this into `ValidatePersistentVolumeClaim`. Currently we're setting a key for `StorageClass` in both places.
```suggestion
		return serrors.Wrap(fmt.Errorf("unbound persistentvolumeclaim with immediate volume binding mode must be bound"), "StorageClass", klog.KObj(storageClass))
```

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,63 @@ var _ = Context("Scheduling", func() {
 			// no nodes should be created as the storage class is using an unsupported provisioner
 			Expect(nodeList.Items).To(HaveLen(0))
 		})
+		It("should not launch nodes for pod with unbound volume for volumeBindingMode immediate", func() {
+			sc := test.StorageClass(test.StorageClassOptions{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "default-storage-class",
+					Annotations: map[string]string{
```
jmdeal: Any reason we're including the default storage class annotation?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,63 @@ var _ = Context("Scheduling", func() {
...
+				Zones:             []string{"test-zone-1"}},
+			)
+
+			ExpectApplied(ctx, env.Client, sc)
+			volumeName := "tmp-ephemeral"
+			pod := test.UnschedulablePod(test.PodOptions{})
+			pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
+				Name: volumeName,
+				VolumeSource: corev1.VolumeSource{
+					Ephemeral: &corev1.EphemeralVolumeSource{
```
jmdeal: Why are we using an ephemeral volume source in this test? This is adding more complexity than necessary IMO, to test this behavior it seems more straightforward and less error prone to test with a pod that references the PVC directly rather than using an ephemeral volume claim template

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,63 @@ var _ = Context("Scheduling", func() {
...
+			cluster.AckPods(pod)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+			ExpectNotScheduled(ctx, env.Client, pod)
+
+			var nodeList corev1.NodeList
+			Expect(env.Client.List(ctx, &nodeList)).To(Succeed())
+			// no nodes should be created as the storage class is using volumeBindingMode immediate the pvc is unbound
+			Expect(nodeList.Items).To(HaveLen(0))
+			nn := client.ObjectKeyFromObject(pod)
+			Expect(cluster.PodSchedulingSuccessTime(nn).IsZero()).To(BeTrue())
```
jmdeal: Is there any particular reason why we're expanding the scope of this test to include updates for cluster state? We should have targeted tests for cluster state / metrics, I don't think we need to expand that into other tests unless absolutely necessary. The targeted tests should ensure that if we never scheduled a pod (e.g. there was no binding created by `ExpectProvisioned`) ,that these weren't updated.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3037,8 +3040,9 @@ var _ = Context("Scheduling", func() {
 						isDefaultStorageClassAnnotation: "true",
 					},
 				},
-				Provisioner: lo.ToPtr(csiProvider),
-				Zones:       []string{"test-zone-1"}})
+				Provisioner:       lo.ToPtr(csiProvider),
+				VolumeBindingMode: lo.ToPtr(storagev1.VolumeBindingWaitForFirstConsumer),
```
jmdeal, line 3044: Worth adding this as the default option in `test.StorageClass`?

pkg/controllers/provisioning/scheduling/volumetopology.go
jigisha620: Updated

pkg/controllers/provisioning/scheduling/volumetopology.go
jigisha620: Updated

pkg/controllers/provisioning/scheduling/suite_test.go
jigisha620: Dropped

pkg/controllers/provisioning/scheduling/suite_test.go
jigisha620: Updated

pkg/controllers/provisioning/scheduling/suite_test.go
jigisha620: Agreed, updated.

pkg/controllers/provisioning/scheduling/suite_test.go
jigisha620: Discussed this offline but I have currently updated each test to have volumeBindingMode as WaitForFirstConsumer, and in reality, when volumeBindingMode is not specified, Immediate is used. But if we change the default option in our test storageClass, it's not really imitating the real world behavior.

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,40 @@ var _ = Context("Scheduling", func() {
...
+				StorageClassName: lo.ToPtr(sc.Name),
+			})
+			pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{
+				Name: volumeName,
+				VolumeSource: corev1.VolumeSource{
+					PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+						ClaimName: pvc.Name,
+					},
+				},
+			})
```
jmdeal: tiny nit: can we inline this when creating the unschedulable pod?

pkg/controllers/provisioning/scheduling/suite_test.go
```diff
@@ -3293,6 +3298,40 @@ var _ = Context("Scheduling", func() {
...
+			ExpectApplied(ctx, env.Client, nodePool, sc, pvc, pod)
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+			ExpectNotScheduled(ctx, env.Client, pod)
+			var nodeList corev1.NodeList
+			Expect(env.Client.List(ctx, &nodeList)).To(Succeed())
+			// no nodes should be created as the storage class is using volumeBindingMode immediate the pvc is unbound
+			Expect(nodeList.Items).To(HaveLen(0))
```
jmdeal: nit: I think we can drop this - we have coverage elsewhere to ensure that no nodes are provisioned if no pods are scheduled.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -178,8 +178,16 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
 		if storageClassName == "" {
 			return serrors.Wrap(fmt.Errorf("unbound pvc must define a storage class"), "PersistentVolumeClaim", klog.KRef("", pvc.Name))
 		}
-		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
-			return serrors.Wrap(fmt.Errorf("failed to validate storage class, %w", err), "PersistentVolumeClaim", klog.KRef("", pvc.Name), "StorageClass", klog.KRef("", storageClassName))
+		storageClass := &storagev1.StorageClass{}
+		if err := v.kubeClient.Get(ctx, types.NamespacedName{Name: storageClassName}, storageClass); err != nil {
+			return serrors.Wrap(fmt.Errorf("failed to validate storage class, %w", err), "persistentVolumeClaim", klog.KRef("", pvc.Name), "storageClass", klog.KRef("", storageClassName))
```
jmdeal:
```suggestion
			return serrors.Wrap(fmt.Errorf("failed to validate pvc, failed to get storage class, %w", err), "PersistentVolumeClaim", klog.KRef("", pvc.Name), "StorageClass", klog.KRef("", storageClassName))
```

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -178,8 +178,16 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
...
+		}
+		if UnsupportedProvisioners.Has(storageClass.Provisioner) {
+			return serrors.Wrap(fmt.Errorf("storageClass provisioner is not supported"), "storageClass provisioner", klog.KRef("", storageClass.Provisioner))
```
jmdeal:
```suggestion
			return serrors.Wrap(fmt.Errorf("failed to validate pvc, provisioner is not supported"), "PersistentVolumeClaim", klog.KRef("", pvc.Name), "StorageClass", klog.KRef("", storageClass.Name, "Provisioner", klog.KRef("", storageClass.Provisioner))
```

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -178,8 +178,16 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
...
+		}
+		// Ignore pods than have unbound pvc for voluemBindingMode immediate
+		if storageClass.VolumeBindingMode != nil && *storageClass.VolumeBindingMode == storagev1.VolumeBindingImmediate {
+			return serrors.Wrap(fmt.Errorf("unbound persistentvolumeclaim with immediate volume binding mode must be bound"), "storageClass", klog.KObj(storageClass))
```
jmdeal: I know this is updating one of my original suggestions, just updating it for consistency with the other errors in this func.
```suggestion
			return serrors.Wrap(fmt.Errorf("failed to validate pvc, pvc with immediate volume binding mode must be bound"), "PersistentVolumeClaim", klog.KRef("", pvc.Name), "StorageClass", klog.KRef("", storageClass.Name))
```

jmdeal: One other comment I have here though is, I wonder if this error is going to be super noisy? Are these errors eventually being printed as logs? I assume so. We would expect there to be some delay between the PVC being created and it being bound, we don't want to flood our logs with errors during that period.

jigisha620: Discussed offline, this is a debug log so it shouldn't affect customers relying on alarming on errors.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -178,8 +178,16 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
 		if storageClassName == "" {
 			return serrors.Wrap(fmt.Errorf("unbound pvc must define a storage class"), "PersistentVolumeClaim", klog.KRef("", pvc.Name))
 		}
-		if err := v.validateStorageClass(ctx, storageClassName); err != nil {
-			return serrors.Wrap(fmt.Errorf("failed to validate storage class, %w", err), "PersistentVolumeClaim", klog.KRef("", pvc.Name), "StorageClass", klog.KRef("", storageClassName))
+		storageClass := &storagev1.StorageClass{}
+		if err := v.kubeClient.Get(ctx, types.NamespacedName{Name: storageClassName}, storageClass); err != nil {
+			return serrors.Wrap(fmt.Errorf("failed to validate pvc, failed to get storage class, %w", err), "PersistentVolumeClaim", klog.KObj(pvc), "StorageClass", storageClassName)
```
jmdeal: I'd still use a `klog.KRef` here. We don't have the object, but it **is** a k8s object and we want to have the same JSON path in the log for all storage classes:
```suggestion
			return serrors.Wrap(fmt.Errorf("failed to validate pvc, failed to get storage class, %w", err), "PersistentVolumeClaim", klog.KObj(pvc), "StorageClass", klog.KRef("", storageClassName))
```

Issue #2395, opened 2025-07-23
karpenter consolidate support oracle provider flex instance type


wy0824:
### Description

**What problem are you trying to solve?**
Hello team, we have an OCI provider impl of karpenter：https://github.com/zoom/karpenter-oci. For Oracle Cloud, there is the flex instance type, one type may have different CPU and memory sizes, and the price of the instance is calculated by CPU cores. Current, when karpenter consolidates, it will treat the same instance type as the same price, which may lead the consolidate unstable in karpenter-oci. Relate issue, please refer: https://github.com/zoom/karpenter-oci/issues/32

**How important is this feature to you?**
It's very important for karpenter-oci provider, currently the consolidate feature does not work for us. Changing the instance type name may take more effort, and it may cause a breaking change. 
We can help create a PR to fix this and make sure it is compatible with other cloud providers. The possible way is to change the `nodePoolToInstanceTypesMap` in `BuildNodePoolMap` to a `map[string]map[string][]*cloudprovider.InstanceType{}`, and filter the compatible instance type in `NewCandidate` func.
Please let us know if this fix method is acceptable. Thanks a lot.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jmdeal:
Just checking if I understand the issue correctly: for a given NodePool you can return multiple instance types with the same name, but different underlying capacity. Is that about right?  I expect this isn't the only place that this would cause an issue - name is expected to be a unique identifier for an instance type. What's preventing you from maintaining that invariant by, for example, appending the capacity as part of the instance type name (e.g. "instance-flex-1cpu-2Gi")

wy0824:
If we change the name of the instance type name, the user needs to change the requirement in the nodepool
eg, the older config:
```
        - key: karpenter.k8s.oracle/instance-shape-name
          operator: In
          values:
            - VM.Standard.E4.Flex
        - key: karpenter.k8s.oracle/instance-cpu
          operator: In
          values:
            - '4'
            - '8'
            - '16'
```
The new config becomes more complex(if we config the cpu/mem ratio to 1:2):
```
[...]

PR #2416, opened 2025-08-01
[feat #2395]support flex instance type

MrGirl:
feat #2395 

**Description**
For provider like oci, it supoort flex instance type like VM.E4.Flex,  so a instance name may has multiple instance types like
nodePoolToInstanceTypesMap[nodepoolname][VM.Standard.A2.Flex] -->
&{VM.Standard.A2.Flex karpenter.k8s.oracle/instance-cpu In [8], .....}
&{VM.Standard.A2.Flex karpenter.k8s.oracle/instance-cpu In [10], .....}
&{VM.Standard.A2.Flex karpenter.k8s.oracle/instance-cpu In [12], .....}

so we need to change NodePoolToInstanceTypesMap from 'map[string]map[string]*cloudprovider.InstanceType' to 'map[string]map[string][]*cloudprovider.InstanceType'

All the old uts have been tested, and we also did integration test in our own karpenter-oci 


wy0824:
@jmdeal please help review, check if it has the impact on capacity-reserved nodes.

wy0824:
hello, @jmdeal , is there any update about this PR, hope for your reply, thanks

bwagner5:
I don't think karpenter core needs to know that flex instance types map to multiple. Couldn't you pass flex variants (it looks like you have to set the vcpus and memory when launching) as individual instance types from the OCI Karpenter Cloud Provider and then provide scheduling labels to select all `VM.E4.Flex` and then have specific ones for `VM.E4.Flex-1vcpus-2GiB` or something like that?

wy0824:
hello @bwagner5 , we want to keep the label `node.kubernetes.io/instance-type` value the same as the CCM labeled. If we add another label, like `node.kubernetes.io/instance-type-wraped=VM.E4.Flex-1vcpus-2GiB`, it will not solve one instance type mapping multi capacities, and still has an impact on the consolidate logic.

munger1985:
as long as we can use the label on nodeclaim CR instead of node CR, downstream implementation can implement whatever we want ,  https://github.com/kubernetes-sigs/karpenter/pull/2402/files.

micsobczyk:
Hi @bwagner5, what do you think about the response on your question which was given by @wy0824 ?

PR #2417, opened 2025-08-01
fix: pod metrics when pod is terminal

### LINKED ISSUE #2403: Negative values on karpenter_pods_startup_duration_seconds
### Description

**Observed Behavior**:
I am getting negative values on 1.6.0 on metric `karpenter_pods_startup_duration_seconds`. This didn't happen on 1.5.0

```
# HELP karpenter_pods_startup_duration_seconds The time from pod creation until the pod is running.
# TYPE karpenter_pods_startup_duration_seconds summary
karpenter_pods_startup_duration_seconds{quantile="0"} -9.223372036854776e+09
karpenter_pods_startup_duration_seconds{quantile="0.5"} 16
karpenter_pods_startup_duration_seconds{quantile="0.9"} 96
karpenter_pods_startup_duration_seconds{quantile="0.99"} 171
karpenter_pods_startup_duration_seconds{quantile="1"} 186
karpenter_pods_startup_duration_seconds_sum -6.456359356198343e+10
karpenter_pods_startup_duration_seconds_count 339
```

When I calculate average as `sum/count` this huge negative number in `sum` - `-6.456359356198343e+10` makes the result in negative days.
**Expected Behavior**:
I can't see how this could be negative so seems like a bug. The pod age can only start at 0 and grow.

**Versions**:
- Chart Version: 1.6.0

jigisha620:
Fixes #2403 

**Description**
`karpenter_pods_startup_duration_seconds` is taking negative values for terminal pods because pod status condition Ready is set to False for terminal pods. 

Added unit tests. Deployed to my dev cluster.

### DIFF
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -319,9 +319,11 @@ func (c *Controller) recordPodStartupMetric(pod *corev1.Pod, schedulableTime tim
 				podName:      pod.Name,
 				podNamespace: pod.Namespace,
 			})
-			PodStartupDurationSeconds.Observe(cond.LastTransitionTime.Sub(pod.CreationTimestamp.Time).Seconds(), nil)
-			if !schedulableTime.IsZero() {
-				PodProvisioningStartupDurationSeconds.Observe(cond.LastTransitionTime.Sub(schedulableTime).Seconds(), nil)
+			if ready {
+				PodStartupDurationSeconds.Observe(cond.LastTransitionTime.Sub(pod.CreationTimestamp.Time).Seconds(), nil)
+				if !schedulableTime.IsZero() {
+					PodProvisioningStartupDurationSeconds.Observe(cond.LastTransitionTime.Sub(schedulableTime).Seconds(), nil)
+				}
 			}
 			c.pendingPods.Delete(key)
 			// Clear cluster state's representation of these pods as we don't need to keep track of them anymore
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -309,7 +309,7 @@ func (c *Controller) recordPodStartupMetric(pod *corev1.Pod, schedulableTime tim
 		return c.Type == corev1.PodReady && c.Status == corev1.ConditionTrue
 	})
 	if c.pendingPods.Has(key) {
-		if ready || podutils.IsTerminal(pod) {
+		if ready {
 			// Delete the unstarted metric since the pod is now started
 			PodUnstartedTimeSeconds.Delete(map[string]string{
 				podName:      pod.Name,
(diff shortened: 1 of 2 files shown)


pkg/controllers/metrics/pod/controller.go
```diff
@@ -319,9 +319,11 @@ func (c *Controller) recordPodStartupMetric(pod *corev1.Pod, schedulableTime tim
 				podName:      pod.Name,
 				podNamespace: pod.Namespace,
 			})
-			PodStartupDurationSeconds.Observe(cond.LastTransitionTime.Sub(pod.CreationTimestamp.Time).Seconds(), nil)
-			if !schedulableTime.IsZero() {
-				PodProvisioningStartupDurationSeconds.Observe(cond.LastTransitionTime.Sub(schedulableTime).Seconds(), nil)
+			if ready {
```
rschalo: Can we actually just handle this in one `if ready` and then perform the same logic in an `else if` for `podutils.IsTerminal`?

jmdeal: +1 - I don't mind the minimal duplication in this case since it keeps us explicit about what we're doing for each case. It's also worth leaving a comment regarding why we have to differentiate between ready and terminal pods. I don't think the updated last transition time on the status time is obvious unless you've looked into the issue.

pkg/controllers/metrics/pod/suite_test.go
```diff
@@ -287,9 +289,9 @@ var _ = Describe("Pod Metrics", func() {
 		Expect(found).To(BeFalse())
 
 		_, found = FindMetricWithLabelValues("karpenter_pods_startup_duration_seconds", nil)
-		Expect(found).To(BeTrue())
+		Expect(found).To(BeFalse())
```
jmdeal, line 292: The name of this test is no longer accurate, is it? We're no longer updating the startup time metrics, just the unstarted time metrics

pkg/controllers/metrics/pod/suite_test.go
```diff
@@ -330,9 +332,9 @@ var _ = Describe("Pod Metrics", func() {
...
```
jmdeal, line 335: Same comment on this test - we're no longer updating the started time metrics

pkg/controllers/metrics/pod/controller.go
jigisha620: Done

pkg/controllers/metrics/pod/suite_test.go
jigisha620: Updated

pkg/controllers/metrics/pod/suite_test.go
jigisha620: Updated

pkg/controllers/metrics/pod/controller.go
```diff
@@ -326,6 +326,22 @@ func (c *Controller) recordPodStartupMetric(pod *corev1.Pod, schedulableTime tim
 			c.pendingPods.Delete(key)
 			// Clear cluster state's representation of these pods as we don't need to keep track of them anymore
 			c.cluster.ClearPodSchedulingMappings(client.ObjectKeyFromObject(pod))
+		} else if podutils.IsTerminal(pod) {
+			// We do not emit the startup duration metric for pods that are terminal because such pods will have
+			// Ready status condition set to False which will cause the metric to take negative values.
```
jmdeal, line 331: I'd specifically call out that the last transition time was updated when it transitioned to false after the pod went terminal

jigisha620: But in this case more than that lastTransitionTime, it's the fact that `cond` stores Ready condition if it's set to true. If not then ideally there's no valid lastTransitionTime stored -
```
cond, ready := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
return c.Type == corev1.PodReady && c.Status == corev1.ConditionTrue
})
```

jmdeal: Ah got it, I misremembered the exact issue a bit. Sounds good then.

Issue #2387, opened 2025-07-21
Mega Issue: Enhancements to Disruption Budgets


saurav-agarwalla:
## Description

Disruption Budgets are often a common cause of confusion for Karpenter users. This stems from a bunch of reasons:
* Lacking documentation: the documentation could use some improvements to make this more intuitive for users and also add some more examples.
* Confusion around maintenance windows: there are a bunch of issues related to confusing behavior around maintenance windows and what that means for disruption.
* Reason-specific disruption: when disruption budgets were initially implemented, it didn't support any reasons so any form of disruption was counted in the budget. When reasons were added, this behavior remained unchanged but, semantically, users started expecting Karpenter to have a bucket-like behavior for different disruption reasons so if a node is disrupted due to expiration (for example), they would still want to allow disruptions for other reasons like drift. This has been a common cause of confusion for users.
* Lack of observability around disruption budgets: Karpenter could do better in exposing this information via status conditions or other mechanisms so that users are able to clearly tell how their budgets are enforced.

I am creating this mega issue to track improvements and consolidate the reported issues under a single umbrella.

saurav-agarwalla:
Some of the issues that I was able to track down which also includes some feature requests/enhancements to disruption budgets:
https://github.com/kubernetes-sigs/karpenter/issues/2374
https://github.com/kubernetes-sigs/karpenter/issues/2353
https://github.com/kubernetes-sigs/karpenter/issues/2344
https://github.com/kubernetes-sigs/karpenter/issues/2329
https://github.com/kubernetes-sigs/karpenter/issues/2218
https://github.com/kubernetes-sigs/karpenter/issues/2017
https://github.com/kubernetes-sigs/karpenter/issues/1902
https://github.com/kubernetes-sigs/karpenter/issues/1775
https://github.com/kubernetes-sigs/karpenter/issues/1719
https://github.com/kubernetes-sigs/karpenter/issues/1691

There are more in the Provider repo too: https://github.com/aws/karpenter-provider-aws/issues?q=is%3Aissue%20state%3Aopen%20disruption%20budget

PR #2423, opened 2025-08-08
fix: do not block drifted nodes from being terminated if consolidation is disabled

saurav-agarwalla:
Fixes #N/A 

**Description**

There was a regression introduced in https://github.com/kubernetes-sigs/karpenter/pull/2180:

```
        // Filter out empty candidates. If there was an empty node that wasn't consolidated before this, we should
        // assume that it was due to budgets. If we don't filter out budgets, users who set a budget for `empty`
        // can find their nodes disrupted here, which while that in itself isn't an issue for empty nodes, it could
        // constrain the `drift` budget.
        if len(candidate.reschedulablePods) == 0 {
            continue
        }
```

This doesn't account for cases where consolidation is disabled (or for that matter has a really long consolidation period). Drift shouldn't be blocked on consolidation since they are independent disruption methods.

So, what happens is that emptiness doesn't disrupt the node since consolidation is disabled. But drift also doesn't terminate the node since the node is empty and it incorrectly assumes that emptiness would always disrupt it.

The original intent makes sense though which is why I am also making a change to prioritize non-empty nodes over empty nodes for drift.

With this change, it is totally possible that nodes which aren't eligible to be consolidated at a moment in time get disrupted due to drift and consume a disruption budget but I think it is more important to still allow drift to happen so that we don't have stale nodes (either with outdated AMI or otherwise) even if they are empty because those empty nodes can still get pods scheduled on them which means that the user runs a risk of having their applications run on outdated AMIs.

This is also why I somehow feel that the original priority order of drift being the first disruption method makes sense since it should be more important for customers to be running a more up-to-date node than being more efficient with their workloads but as the original PR mentions, there are rough edges to that approach especially when customers drift nodes for non-critical reasons.

All of this could've been prevented in the first place if we actually had separate buckets for each disruption reason. Something that https://github.com/kubernetes-sigs/karpenter/issues/2387 intends to solve.

`make presubmit`

### DIFF
--- a/pkg/controllers/disruption/drift.go
+++ b/pkg/controllers/disruption/drift.go
@@ -19,8 +19,10 @@ package disruption
 import (
 	"context"
 	"errors"
+	"slices"
 	"sort"
 
+	"github.com/samber/lo"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/karpenter/pkg/utils/pretty"
@@ -61,14 +63,15 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
 			candidates[j].NodeClaim.StatusConditions().Get(string(d.Reason())).LastTransitionTime.Time)
 	})
 
-	for _, candidate := range candidates {
-		// Filter out empty candidates. If there was an empty node that wasn't consolidated before this, we should
-		// assume that it was due to budgets. If we don't filter out budgets, users who set a budget for `empty`
-		// can find their nodes disrupted here, which while that in itself isn't an issue for empty nodes, it could
-		// constrain the `drift` budget.
-		if len(candidate.reschedulablePods) == 0 {
-			continue
-		}
+	emptyCandidates := lo.Filter(candidates, func(c *Candidate, _ int) bool {
+		return len(c.reschedulablePods) == 0
+	})
+	nonEmptyCandidates := lo.Filter(candidates, func(c *Candidate, _ int) bool {
+		return len(c.reschedulablePods) > 0
+	})
+
+	// Prioritize non-empty candidates since we want them to get priority over empty candidates if the budget is constrained.
+	for _, candidate := range slices.Concat(nonEmptyCandidates, emptyCandidates) {
 		// If the disruption budget doesn't allow this candidate to be disrupted,
 		// continue to the next candidate. We don't need to decrement any budget
 		// counter since drift commands can only have one candidate.
--- a/pkg/controllers/disruption/drift_test.go
+++ b/pkg/controllers/disruption/drift_test.go
@@ -49,7 +49,7 @@ var _ = Describe("Drift", func() {
 		nodePool = test.NodePool(v1.NodePool{
 			Spec: v1.NodePoolSpec{
 				Disruption: v1.Disruption{
-					ConsolidateAfter: v1.MustParseNillableDuration("Never"),
+					ConsolidateAfter: v1.MustParseNillableDuration("1h"),
 					// Disrupt away!
 					Budgets: []v1.Budget{{
 						Nodes: "100%",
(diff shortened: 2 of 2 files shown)


pkg/controllers/disruption/drift_test.go
```diff
@@ -49,7 +49,7 @@ var _ = Describe("Drift", func() {
 		nodePool = test.NodePool(v1.NodePool{
 			Spec: v1.NodePoolSpec{
 				Disruption: v1.Disruption{
-					ConsolidateAfter: v1.MustParseNillableDuration("Never"),
+					ConsolidateAfter: v1.MustParseNillableDuration("1h"),
```
DerekFrank, line 52: Is there a reason to switch this to 1h?

pkg/controllers/disruption/drift_test.go
```diff
@@ -727,6 +728,48 @@ var _ = Describe("Drift", func() {
 			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(1))
 			ExpectNotFound(ctx, env.Client, nodeClaim, node)
 		})
+		It("should delete drifted nodes when they are empty and consolidation is disabled", func() {
```
rschalo, line 661: We should add a test that shows the ordering of disruption given non-empty nodes are being prioritized.

pkg/controllers/disruption/drift.go
```diff
@@ -61,14 +63,15 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
...
-		// assume that it was due to budgets. If we don't filter out budgets, users who set a budget for `empty`
-		// can find their nodes disrupted here, which while that in itself isn't an issue for empty nodes, it could
-		// constrain the `drift` budget.
-		if len(candidate.reschedulablePods) == 0 {
-			continue
-		}
+	emptyCandidates := lo.Filter(candidates, func(c *Candidate, _ int) bool {
+		return len(c.reschedulablePods) == 0
+	})
+	nonEmptyCandidates := lo.Filter(candidates, func(c *Candidate, _ int) bool {
```
jigisha620: Can we use `lo.FilterReject` for this operation with `len(c.reschedulablePods) == 0` as the predicate. That should give us two slices. One of the slice satisfies the predicate and the other slice doesn't. That way, you don't have to use `lo.Filter` twice.

saurav-agarwalla: TIL: I'll make the change.

pkg/controllers/disruption/drift_test.go
saurav-agarwalla: Sure thing. Updating in the next revision.

pkg/controllers/disruption/drift_test.go
saurav-agarwalla: Just making the default case more representative of real world. It didn't matter earlier since drift was prioritized over consolidation but it does now since emptiness runs first.

pkg/controllers/disruption/drift.go
```diff
@@ -61,14 +63,12 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
...
-		// can find their nodes disrupted here, which while that in itself isn't an issue for empty nodes, it could
-		// constrain the `drift` budget.
-		if len(candidate.reschedulablePods) == 0 {
-			continue
-		}
+	emptyCandidates, nonEmptyCandidates := lo.FilterReject(candidates, func(c *Candidate, _ int) bool {
+		return len(c.reschedulablePods) == 0
+	})
+
+	// Prioritize non-empty candidates since we want them to get priority over empty candidates if the budget is constrained.
```
saurav-agarwalla: @jmdeal pointed out offline that this might not be desirable since drifting non-empty nodes first might just mean that the pods from those nodes move to the empty nodes leading to more churn. I'll reverse the order in the next revision.

pkg/controllers/disruption/drift.go
```diff
@@ -61,14 +63,12 @@ func (d *Drift) ComputeCommand(ctx context.Context, disruptionBudgetMapping map[
...
-		// can find their nodes disrupted here, which while that in itself isn't an issue for empty nodes, it could
-		// constrain the `drift` budget.
-		if len(candidate.reschedulablePods) == 0 {
-			continue
-		}
+	emptyCandidates, nonEmptyCandidates := lo.FilterReject(candidates, func(c *Candidate, _ int) bool {
+		return len(c.reschedulablePods) == 0
+	})
+
+	// Prioritize empty candidates since we want them to get priority over non-empty candidates if the budget is constrained.
```
jmdeal, line 70: Can we update this comment to include why we want to prioritize the non-empty candidates? This is a bit tricky to reason about, so I think it's worth including.

saurav-agarwalla: Done.

PR #2430, opened 2025-08-12
chore: increase MaxInstanceTypes to give cloud-providers more control over instance type truncation

bwagner5:
Fixes #N/A 

**Description**

This change increases the instance types that are set in a NodeClaim from 60 to 600. The number 600 was chosen to be sufficiently high but still putting an upper-bound on the number of instance types set in the NodeClaim. 

This should not increase scheduling latency since truncation occurs after scheduling/bin-packing simulations. 

The AWS Cloud Provider is still truncating to 60 instance types, set [here](https://github.com/aws/karpenter-provider-aws/blob/main/pkg/providers/instance/instance.go#L62).

The Azure Cloud Provider sorts the NodeClaim's instance types and only passes 1 to the provisioning API [here](https://github.com/Azure/karpenter-provider-azure/blob/main/pkg/providers/instance/instance.go#L626).

Manual testing:
 1. Created a wide-open NodePool with AWS Cloud Provider 
 2. Created a small pod for Karpenter to provision a NodeClaim for
 3. Observed the NodeClaim contained 600 instance types
 4. Observed the CreateFleet request still contained 60 Instance Types from Cloud Provider truncation.

### DIFF
--- a/pkg/controllers/provisioning/scheduling/instance_selection_test.go
+++ b/pkg/controllers/provisioning/scheduling/instance_selection_test.go
@@ -1373,7 +1373,7 @@ var _ = Describe("Instance Type Selection", func() {
 			instanceTypes = append(instanceTypes, fake.NewInstanceType(opts2))
 			// We have the required InstanceTypes that meet the minValues requirement.
 			cloudProvider.InstanceTypes = instanceTypes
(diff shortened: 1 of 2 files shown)


jonathan-innis (approved):
/lgtm
/approve

PR #2433, opened 2025-08-12
feat: Create optional operator arguments to leverage leader lease functionality

DerekFrank:
Fixes #N/A 

**Description**

This allows cloud providers to leverage the controller runtime leader election label functionality

`make presubmit`


pkg/operator/operator.go
```diff
@@ -102,8 +103,14 @@ type Operator struct {
 	Clock               clock.Clock
 }
 
+type OperatorOpts struct {
```
jmdeal: Generally we're going to want to provide the helper functions to generate these opts. Something like this:

```go
// A descriptive description
func WithLeaderElectionLabels(labels map[string]string) option.Function[OperatorOpts] {
  return func(opts *OperatorOpts) {
    opts.LeaderElectionLabels = labels
  }
}
```

jmdeal: nit: this stutters a bit, I'd consider naming this `Options` that way the full name with the package is `operator.Options`

PR #2431, opened 2025-08-12
feat: Add static capacity controllers 

sumukha-radhakrishna:
Fixes #2309

**Description**

This adds a static controller that watches NodePool resources and creates NodeClaims based on the desired replica count specified in the NodePool spec. The changes include:
* Added new static provisioning controller that:
  * Watches NodePool resources for replica changes
  * Creates NodeClaims to match desired replica count
  * Handles scaling up 
* Scale down support deprovisioning controller
  * Watches NodePool resources for replica changes
  * Deletes NodeClaims to match desired replica count
  * Handles scaling down
* Drift support for static NodeClaims

**How was this change tested?**

The changes were tested through:
* Added comprehensive test suite 
* Ran a local NodePool with replicas and ensured NodeClaims/Nodes were created.
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: static-nodepool
spec:
  replicas: 2
  template:
    spec:
      requirements:
        - key: "karpenter.sh/capacity-type"
          operator: In
          values: ["on-demand"]
        - key: "node.kubernetes.io/instance-type"
          operator: In
          values: ["t3.medium"]
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: default
      expireAfter: 720h # 30 * 24h = 720h
  disruption:
    consolidateAfter: Never

```

Note this PR contains a dependent PR changes - #2345 

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2449, opened 2025-08-21
perf: Track NodeClaims state in NodePoolState

sumukha-radhakrishna:
Needed for #2431 

**Description**

This change introduces a NodePoolState to track NodeClaims state. The main changes include:
1. Introduction of a new NodePoolState structure that tracks:
   * NodeClaim states (Running and Deleting)
   * Mapping between NodeClaim names and NodePool names
   * NodePool limits

2. Modifications to existing code:
   * Updates to cluster.go to integrate NodePoolState
   * Added methods to track and manage NodeClaim states within the cluster

The primary goal of these changes appears to be improving the tracking and management of NodeClaims within NodePools, making the system more efficient in handling node states and limits.

Unit tests and `make presubmit`


pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+type NodeClaimState struct {
+	Running  sets.Set[string]
+	Deleting sets.Set[string]
+}
+
+// StateNodePool is a cached version of a NodePool in the cluster that maintains state which is expensive to compute every time it's needed.
+type NodePoolState struct {
+	npMutex sync.RWMutex
```
jonathan-innis:
```suggestion
	mu sync.RWMutex
```

nit: Just naming

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	"sync"
+	"sync/atomic"
+
+	"github.com/samber/lo"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+)
+
+// Tracks vended limits for a NodePool that are in-flight
```
jonathan-innis: This statement doesn't make sense to me -- what are "vended limits"?

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	defer n.npMutex.RUnlock()
+
+	return n.nodePoolNameToNodeClaimState[npName].Running.Has(ncName)
+}
+
+// Methods that expect the called to hold the lock
+
+// ReserveNodePoolNodeLimit attempts to reserve nodes against a NodePool's limit.
+// It ensures that the total of running nodes + deleting nodes + reserved nodes doesn't exceed the limit.
+func (n *NodePoolState) ReserveNodePoolNodeLimit(np string, limit int64, wantedLimit int64) int64 {
```
jonathan-innis: I'm not sure I get this function -- why do you need to explicitly reserve in memory like this? Why can't you just rely on the state tracking to sum up the total number of nodes in the resource counts?

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+)
+
+// Tracks vended limits for a NodePool that are in-flight
+type NodePoolLimit struct {
+	ReservedNodeLimit atomic.Int64
+}
+
+type NodeClaimState struct {
+	Running  sets.Set[string]
```
jonathan-innis: FWIW we chose to call this "Active" elsewhere in `statenode.go`

pkg/controllers/state/cluster.go
```diff
@@ -64,6 +64,8 @@ type Cluster struct {
 	nodePoolResources         map[string]corev1.ResourceList  // node pool name -> resource list
 	daemonSetPods             sync.Map                        // daemonSet -> existing pod
 
+	npState *NodePoolState
```
jonathan-innis:
```suggestion
	nodePoolState *NodePoolState
```

nit

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: vended limits are the limits that are in flight given to the caller

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: Discussed offline : We wanna specifically maintain limit reservation to ensure we dont burst over hard limits. We will reserve during launching new NodeClaims and release once we are done (successful or failed)

pkg/controllers/state/cluster.go
```diff
@@ -281,6 +285,9 @@ func (c *Cluster) UnmarkForDeletion(providerIDs ...string) {
 			oldNode := n.ShallowCopy()
 			n.markedForDeletion = false
 			c.updateNodePoolResources(oldNode, n)
+			if n.NodeClaim != nil {
+				c.npState.MarkNodeClaimRunning(n.NodeClaim.Labels[v1.NodePoolLabelKey], n.NodeClaim.Name)
```
jonathan-innis: You can't just rely on this, right? You have to also know that the NodeClaim doesn't have its DeletionTimestamp set, so you have to check that before you can move the NodeClaim into the "Active" or "Running" state

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	Deleting sets.Set[string]
+}
+
+// StateNodePool is a cached version of a NodePool in the cluster that maintains state which is expensive to compute every time it's needed.
+type NodePoolState struct {
+	mu sync.RWMutex
+
+	nodePoolNameToNodeClaimState map[string]NodeClaimState // node pool name -> node claim state (Running and Deleting node claim names)
+	nodeClaimNameToNodePoolName  map[string]string         // node claim name -> node pool name
+	nodePoolNameToNodePoolLimit  map[string]*NodePoolLimit // node pool -> nodepool limit
```
jonathan-innis: I don't think you need a struct for the NodePool Limit -- you should just be able to directly map between a string and the atomicInt

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+
+}
+
+func NewNodePoolState() *NodePoolState {
```
jonathan-innis, line 44: Do we delete entries in this NodePool state in response to a NodePool deletion? The current state will just leak entries from what I can tell

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+func (n *NodePoolState) ensureNodePoolEntry(np string) {
+	if _, ok := n.nodePoolNameToNodeClaimState[np]; !ok {
+		n.nodePoolNameToNodeClaimState[np] = NodeClaimState{
+			Running:  sets.New[string](),
+			Deleting: sets.New[string](),
+		}
+	}
+	if _, ok := n.nodePoolNameToNodePoolLimit[np]; !ok {
+		nl := &NodePoolLimit{}
+		nl.ReservedNodeLimit.Store(0)
```
jonathan-innis: AFAIK this should be set to 0 by default when it's first initialized

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	nodeClaimNameToNodePoolName  map[string]string         // node claim name -> node pool name
+	nodePoolNameToNodePoolLimit  map[string]*NodePoolLimit // node pool -> nodepool limit
+
+}
+
+func NewNodePoolState() *NodePoolState {
+	return &NodePoolState{
+		nodePoolNameToNodeClaimState: map[string]NodeClaimState{},
+		nodeClaimNameToNodePoolName:  map[string]string{},
+		nodePoolNameToNodePoolLimit:  map[string]*NodePoolLimit{},
```
jonathan-innis: I don't think we need to store the NodePool limit here in state -- let's just read it from the NodePool directly

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	n.ensureNodePoolEntry(npName)
+	n.nodeClaimNameToNodePoolName[ncName] = npName
+}
+
+// ReleaseNodePoolNodeLimit releases the NodePoolsNodeLimit
+func (n *NodePoolState) ReleaseNodePoolNodeLimit(npName string) {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	currentlyReserved := n.nodePoolNameToNodePoolLimit[npName].ReservedNodeLimit.Load()
```
jonathan-innis: IMO, you should be able to pass a numeric value for how many you want to release here and the functions should be called something like Add() and Subtract()

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	n.ensureNodePoolEntry(npName)
+	n.nodeClaimNameToNodePoolName[ncName] = npName
+}
+
+// ReleaseNodePoolNodeLimit releases the NodePoolsNodeLimit
+func (n *NodePoolState) ReleaseNodePoolNodeLimit(npName string) {
```
jonathan-innis: I think you can drop any function signature that contains the word `NodePool` -- this is fairly redundant when we are already talking about something called `NodePoolState`. You can imagine an external caller doing something like `cluster.NodePoolState.ReserveNodeLimit` rather than `cluster.NodePoolState.ReserveNodePoolNodeLimit`

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+
+// Methods that expect the called to hold the lock
+
+// ReserveNodePoolNodeLimit attempts to reserve nodes against a NodePool's limit.
+// It ensures that the total of running nodes + deleting nodes + reserved nodes doesn't exceed the limit.
+func (n *NodePoolState) ReserveNodePoolNodeLimit(np string, limit int64, wantedLimit int64) int64 {
+	n.ensureNodePoolEntry(np)
+	running, deleting := n.NodePoolNodeCounts(np)
+
+	for {
```
jonathan-innis, line 124: Can we add a function here to indicate why we are iterating multiple times here for this -- IMO: this pattern is a little strange compared to a mutex on an integer value. Haven't really seen this done, but I do recognize that it works

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+		)
+
+		if n.nodePoolNameToNodePoolLimit[np].ReservedNodeLimit.CompareAndSwap(currentlyReserved, currentlyReserved+grantedLimit) {
+			return grantedLimit
+		}
+	}
+}
+
+// Returns the current NodeClaims for a NodePool by its state (running, deleting)
+func (n *NodePoolState) NodePoolNodeCounts(npName string) (running, deleting int) {
```
jonathan-innis:
```suggestion
func (n *NodePoolState) NodeCounts(npName string) (running, deleting int) {
```

nit

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	// If our node is marked for deletion, we need to make sure that we delete it
+	if markedForDeletion {
+		n.MarkNodeClaimDeleting(npName, nodeClaim.Name)
+	} else {
+		n.MarkNodeClaimRunning(npName, nodeClaim.Name)
+
+	}
+
+	// if we transition from not running to running then we release the limit reservation
+	if !wasNodeClaimRunning && n.IsNodeClaimRunning(npName, nodeClaim.Name) {
```
jonathan-innis: I'm not sure that I get this -- why is this needed at all?

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,187 @@
...
+	// If we are launching/deleting a NodeClaim we need to track the state of the NodeClaim and its limits
+	npName := nodeClaim.Labels[v1.NodePoolLabelKey]
+	n.SetNodeClaimMapping(npName, nodeClaim.Name)
+	wasNodeClaimRunning := n.IsNodeClaimRunning(npName, nodeClaim.Name)
+
+	// If our node is marked for deletion, we need to make sure that we delete it
+	if markedForDeletion {
+		n.MarkNodeClaimDeleting(npName, nodeClaim.Name)
+	} else {
+		n.MarkNodeClaimRunning(npName, nodeClaim.Name)
```
jonathan-innis: You can't guarantee this because you still have to check the deletionTimestamp

pkg/controllers/state/cluster.go
```diff
@@ -873,3 +892,20 @@ func (c *Cluster) triggerConsolidationOnChange(old, new *StateNode) {
...
 }
+
+// We want to split cluster.go into smaller parts and give the ability to access/modify using helpers to avoid deep-copying
+// Any helpers on NodePoolState need to hold the lock
+
+func (c *Cluster) NodePoolNodeCounts(np string) (int, int) {
+	c.npState.mu.RLock()
+	defer c.npState.mu.RUnlock()
+
+	return c.npState.NodePoolNodeCounts(np)
```
jonathan-innis: Why does the NodeCount function itself not hold the lock?

pkg/controllers/state/cluster.go
```diff
@@ -873,3 +892,20 @@ func (c *Cluster) triggerConsolidationOnChange(old, new *StateNode) {
...
+}
+
+func (c *Cluster) ReserveNodePoolNodeLimit(np string, limit, wanted int64) int64 {
```
jonathan-innis:
```suggestion
func (c *Cluster) ReserveNodeCount(np string, limit, wanted int64) int64 {
```

nit, I also don't think you need to pass the limit value here

pkg/controllers/state/cluster.go
sumukha-radhakrishna: So we call `UnmarkForDeletion` from disruption queue controller. When it cannot successfully terminate the NodeClaim. 
Since `DeletionTimeStamp` will only be set when the api-server registers delete operation, if the call succeeds then we would not UnmarkForDeletion. So it is not strictly needed but doesnt hurt to check. Let me add

pkg/controllers/state/cluster.go
sumukha-radhakrishna: The NodeCount function is also called within ReserveNodeCount which has to hold the lock so it was cleaner for the caller to hold the lock.

pkg/controllers/state/cluster.go
sumukha-radhakrishna: We have to, we dont store NodePool obj anywhere.

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: Added incase we want to support other limits in the future.

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: Good point. I will add CleanUp logic if there are no NodePoolResources/NodeClaims left

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: This is the NodePool limit tracker that tracks the in-flight NodeClaims. Let me rename it as NodePoolTracker

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: Compare and Swap will not fail it returns false when the value we compared against is no longer current. Its a pattern while using CompareAndSwap

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: Its not, I removed this. We will directly call from provisioner

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: Added

pkg/controllers/state/statenodepool.go
jonathan-innis: Then let's update it then -- I don't see a reason to pro-actively build something like this (especially when it's internal and easily change-able)

pkg/controllers/state/statenodepool.go
jonathan-innis: Sorry, I meant a comment -- I don't know why I wrote function

pkg/controllers/state/cluster.go
jonathan-innis: And we can't pass the NP object into the function?

pkg/controllers/state/cluster.go
```diff
@@ -310,6 +320,10 @@ func (c *Cluster) UpdateNodeClaim(nodeClaim *v1.NodeClaim) {
 		n := c.newStateFromNodeClaim(nodeClaim, c.nodes[nodeClaim.Status.ProviderID])
 		c.nodes[nodeClaim.Status.ProviderID] = n
 	}
+
+	// Update nodepool state with NodeClaim
+	c.NodePoolState.UpdateNodeClaim(nodeClaim, c.IsNodeClaimMarkedForDeletion(nodeClaim.Status.ProviderID))
```
jonathan-innis: Don't we already have a `n.MarkedForDeletion()` call? If we don't have a stateNode here yet, we can know off the bat that it's not marked for deletion yet

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	n.mu.Lock()
+	defer n.mu.Unlock()
+
+	n.ensureNodePoolEntry(npName)
+	n.nodeClaimNameToNodePoolName[ncName] = npName
+}
+
+// Marks the given NodeClaim as running in NodePoolState
+func (n *NodePoolState) MarkNodeClaimRunning(npName, ncName string) {
```
jonathan-innis:
```suggestion
func (n *NodePoolState) MarkNodeClaimActive(npName, ncName string) {
```

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,203 @@
...
+		if npState.Active.Len() == 0 && npState.Deleting.Len() == 0 {
+			delete(n.nodePoolNameToNodeClaimState, npName)
+			delete(n.nodePoolNameToNodePoolLimit, npName)
+		}
+	}
+
+	delete(n.nodeClaimNameToNodePoolName, ncName)
+}
+
+func (n *NodePoolState) IsNodeClaimRunning(npName, ncName string) bool {
```
jonathan-innis:
```suggestion
func (n *NodePoolState) IsNodeClaimActive(npName, ncName string) bool {
```

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,203 @@
...
+	Deleting sets.Set[string]
+}
+
+// StateNodePool is a cached version of a NodePool in the cluster that maintains state which is expensive to compute every time it's needed.
+type NodePoolState struct {
+	mu sync.RWMutex
+
+	nodePoolNameToNodeClaimState map[string]NodeClaimState   // node pool name -> node claim state (Running and Deleting node claim names)
+	nodeClaimNameToNodePoolName  map[string]string           // node claim name -> node pool name
+	nodePoolNameToNodePoolLimit  map[string]*NodePooltracker // node pool -> nodepool limit
```
jonathan-innis:
```suggestion
	nodePoolNameToReservedNodeCount  map[string]atomic.Int64// node pool -> nodepool limit
```

IMO this is what we should use for the type and the naming

pkg/controllers/state/cluster.go
jonathan-innis: Given that the calling function already checks if the providerID is set, could you consider moving this up to L321

pkg/controllers/state/statenodepool_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/test"
+	. "sigs.k8s.io/karpenter/pkg/test/expectations"
+)
+
+var _ = Describe("NodePool Node Limits", func() {
+	var nodeClaim *v1.NodeClaim
+	var nodePool2 *v1.NodePool
```
jonathan-innis:
```suggestion
	var nodePool *v1.NodePool
```

pkg/controllers/state/statenodepool_test.go
```diff
@@ -0,0 +1,187 @@
...
+				Labels: map[string]string{
+					v1.NodePoolLabelKey: nodePool.Name,
+				},
+			},
+			Status: v1.NodeClaimStatus{
+				ProviderID: test.RandomProviderID(),
+			},
+		})
+		nodePool2 = test.NodePool(v1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: "nodepool-2"}})
+		ExpectApplied(ctx, env.Client, nodePool2)
```
jonathan-innis: When do we look up the NodePool? Do we even need to apply it to the cluster in this test?

pkg/controllers/state/statenodepool_test.go
```diff
@@ -0,0 +1,187 @@
...
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/test"
+	. "sigs.k8s.io/karpenter/pkg/test/expectations"
+)
+
+var _ = Describe("NodePool Node Limits", func() {
```
jonathan-innis:
```suggestion
var _ = Describe("NodePoolState", func() {
```

nit: style

pkg/controllers/state/statenodepool_test.go
```diff
@@ -0,0 +1,187 @@
...
+			Status: v1.NodeClaimStatus{
+				ProviderID: test.RandomProviderID(),
+			},
+		})
+		nodePool2 = test.NodePool(v1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: "nodepool-2"}})
+		ExpectApplied(ctx, env.Client, nodePool2)
+	})
+
+	// rsumukha@ todo : Add tests to verify Reserve and Release behavior during NodeClaim Create call
+	Context("NodePoolState.ReserveNodeCount", func() {
```
jonathan-innis:
```suggestion
	Context("ReserveNodeCount", func() {
```

pkg/controllers/state/statenodepool_test.go
```diff
@@ -0,0 +1,187 @@
...
+
+			wg.Wait()
+			// Should not exceed the limit
+			Expect(totalGranted).To(BeNumerically("<=", limit))
+			Expect(totalGranted).To(Equal(limit)) // Should grant exactly the limit
+		})
+
+	})
+
+	Context("NodePoolState.ReleaseNodeCount", func() {
```
jonathan-innis:
```suggestion
	Context("ReleaseNodeCount", func() {
```

pkg/controllers/state/suite_test.go
```diff
@@ -1241,6 +1241,9 @@ var _ = Describe("Cluster State Sync", func() {
 	})
 	It("should emit cluster_state_unsynced_time_seconds metric when cluster state is unsynced", func() {
 		nodeClaim := test.NodeClaim(v1.NodeClaim{
+			ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{
```
jonathan-innis: Does this mean that things fail now if there is no NodePool? Creating a NodeClaim without a NodePool is a valid setup, so we should make sure that this still works without this label

pkg/controllers/state/statenodepool_test.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+
+			wg.Wait()
+
+			// Should have full capacity available after releases
+			granted = cluster.NodePoolState.ReserveNodeCount(nodePool.Name, 10, 100)
+			Expect(granted).To(Equal(int64(10)))
+		})
+	})
+})
```
jonathan-innis, line 186: Should we add any testing for getting the count of nodes back: 1) When there is no NodePool tracked in our maps? 2) When there are NodeClaims tracked

pkg/controllers/state/statenodepool.go
jonathan-innis: Also, what's the need for this function -- feels kind of odd to have this here when this attribute for a NodeClaim doesn't relate to NodePools at all

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,203 @@
...
+	if st, ok := n.nodePoolNameToNodeClaimState[npName]; ok {
+		return len(st.Active), len(st.Deleting)
+	}
+	return 0, 0
+}
+
+// Updates the NodeClaim state and releases the Limit if NodeClaim transitions from not Running to Running
+func (n *NodePoolState) UpdateNodeClaim(nodeClaim *v1.NodeClaim, markedForDeletion bool) {
+	// If we are launching/deleting a NodeClaim we need to track the state of the NodeClaim and its limits
+	npName := nodeClaim.Labels[v1.NodePoolLabelKey]
```
jonathan-innis, line 169: If this value is empty, I think we should ignore updating for this NodeClaim, since that means that it isn't associated with any NodePool

pkg/controllers/state/suite_test.go
```diff
@@ -2430,6 +2466,381 @@ var _ = Describe("NodePool Resources", func() {
 	})
 })
 
+var _ = Describe("NodeClaim State Tracking", func() {
```
jonathan-innis:
```suggestion
var _ = Describe("NodePoolState Tracking", func() {
```

pkg/controllers/state/cluster.go
sumukha-radhakrishna: We can, thought its better to pass the limit.

pkg/controllers/state/statenodepool_test.go
sumukha-radhakrishna: We do apply nodeclaim to count for active nodeclaim

pkg/controllers/state/suite_test.go
sumukha-radhakrishna: I've removed this. it should work without this key. I will remove these

pkg/controllers/state/statenodepool_test.go
sumukha-radhakrishna: > 1) When there is no NodePool tracked in our maps? 

Will add this 

> 2) When there are NodeClaims tracked

We already have testing for this in suite_test.go under NodePool tracking

pkg/controllers/state/statenodepool.go
```diff
@@ -0,0 +1,200 @@
...
+		if npState.Active.Len() == 0 && npState.Deleting.Len() == 0 {
+			delete(n.nodePoolNameToNodeClaimState, npName)
+			delete(n.nodePoolNameToNodePoolLimit, npName)
+		}
+	}
+
+	delete(n.nodeClaimNameToNodePoolName, ncName)
+}
+
+func (n *NodePoolState) IsNodeClaimActive(npName, ncName string) bool {
```
jonathan-innis: I would shy away from adding functions that are exclusively used for testing -- is there a way that we can enable the ability to delete this?

pkg/controllers/state/suite_test.go
```diff
@@ -2430,6 +2434,401 @@ var _ = Describe("NodePool Resources", func() {
 	})
 })
 
+var _ = Describe("NodePoolState Tracking", func() {
+	var nodeClaim *v1.NodeClaim
+	var nodeClaim2 *v1.NodeClaim
+	var nodePool2 *v1.NodePool
+
+	BeforeEach(func() {
+		nodePool2 = test.NodePool(v1.NodePool{ObjectMeta: metav1.ObjectMeta{Name: "nodepool2"}})
```
jonathan-innis, line 2439: Why is it called nodePool2? Do we use nodePool1 in here?

pkg/controllers/state/suite_test.go
```diff
@@ -1777,6 +1777,9 @@ var _ = Describe("Data Races", func() {
 		// Call UpdateNodeClaim on 100 NodeClaims (enough to trigger a DATA RACE)
 		for i := 0; i < 100; i++ {
 			nodeClaim := test.NodeClaim(v1.NodeClaim{
+				ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{
```
jonathan-innis: Do we still need this?

pkg/controllers/state/suite_test.go
```diff
@@ -1795,6 +1798,7 @@ var _ = Describe("Taints", func() {
 		nodeClaim, node = test.NodeClaimAndNode(v1.NodeClaim{
 			ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{
 				corev1.LabelInstanceTypeStable: instanceType.Name,
+				v1.NodePoolLabelKey:            nodePool.Name,
```
jonathan-innis: And this?

pkg/controllers/state/suite_test.go
sumukha-radhakrishna: We just use nodepool and nodepool2. Its to track multiple nodepools

pkg/controllers/state/suite_test.go
```diff
@@ -1479,14 +1479,14 @@ var _ = Describe("Cluster State Sync", func() {
 		ExpectMetricGaugeValue(state.ClusterStateSynced, 0, nil)
 	})
 	It("shouldn't consider the cluster state synced if a nodeclaim is added manually with UpdateNodeClaim", func() {
-		nodeClaim := test.NodeClaim()
+		nodeClaim := test.NodeClaim(v1.NodeClaim{})
```
jonathan-innis: Why did you need to add these changes?

jonathan-innis:
/hold 

Wait for review of the nit comments

jonathan-innis:
Changing the title to "chore" classification, since this doesn't have to do with improving our performance really

jonathan-innis (approved):
/lgtm
/approve

PR #2471, opened 2025-08-30
fix(scheduling): conditionally inject matchLabelKeys

jmdeal:
Fixes #N/A 

**Description**
Starting in k8s 1.34, `matchLabelKeys` values are injected into match expressions by the API server. This matches the implementation for pod affinity. This PR adds a default version provider (overridable by the provider) which will allow us to conditionally disable our own injection on k8s versions 1.34 and greater.

`make test`


pkg/operator/operator.go
```diff
@@ -95,16 +97,22 @@ func init() {
 	})
 }
 
+type KubernetesVersionProvider interface {
```
jonathan-innis, line 100: Should this go in the version package? Seems a bit odd that this would land in the operator package to me

pkg/controllers/controllers.go
```diff
@@ -98,6 +101,10 @@ func NewControllers(
 		nodehydration.NewController(kubeClient, cloudProvider),
 	}
 
+	if p, ok := kubeVersionProvider.(*version.Provider); ok {
```
jonathan-innis, line 104: Is there a better way to handle this by registering the controller on the `operator.go` side. It feels sort of odd to pass this through and then do static type checking like this to enable the controller mechanism

pkg/controllers/disruption/suite_test.go
```diff
@@ -96,7 +96,7 @@ var _ = BeforeSuite(func() {
 	nodeStateController = informer.NewNodeController(env.Client, cluster)
 	nodeClaimStateController = informer.NewNodeClaimController(env.Client, cloudProvider, cluster)
 	recorder = test.NewEventRecorder()
-	prov = provisioning.NewProvisioner(env.Client, recorder, cloudProvider, cluster, fakeClock)
+	prov = provisioning.NewProvisioner(env.Client, env.KubernetesVersionProvider(), recorder, cloudProvider, cluster, fakeClock)
```
jonathan-innis, line 99: I could go either way, but why make this a method and not have this just be an interface that can be accessed directly

pkg/controllers/provisioning/scheduling/scheduling_benchmark_test.go
```diff
@@ -451,3 +453,11 @@ func randomCPU() resource.Quantity {
 	cpu := []int{100, 250, 500, 1000, 1500}
 	return resource.MustParse(fmt.Sprintf("%dm", cpu[r.Intn(len(cpu))]))
 }
+
+type staticVersionProvider struct {
```
jonathan-innis, line 457: Should static version provider just be something that we put in the version package? We could share this across packages, including in the test environment

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -431,15 +436,20 @@ func (t *Topology) newForTopologies(p *corev1.Pod) []*TopologyGroup {
...
+					tsc.LabelSelector.MatchExpressions = append(tsc.LabelSelector.MatchExpressions, metav1.LabelSelectorRequirement{
+						Key:      key,
+						Operator: metav1.LabelSelectorOpIn,
+						Values:   []string{value},
+					})
+				}
 			}
 		}
+		marshaled := string(lo.Must(json.Marshal(tsc.LabelSelector.MatchExpressions)))
+		fmt.Printf("%s\n", marshaled)
```
jonathan-innis, line 452: I think you need to drop this print statement and the marshaling

jmdeal:
Discussed offline, and agreed that we should figure out why the double-injected requirements are having the impact they are. In theory, it shouldn't matter if we've injected the same match expression twice into the selector term set. We've reverted 1.34 testing for the time being to unblock other PRs, and will reintroduce it once this PR is merged.

jmdeal:
Well this is interesting - it looks like we're probably running into https://github.com/mitchellh/hashstructure/issues/36. I've been able to verify that the issue is caused by us getting the same hash for a `TopologyGroup` constructed with the "value-a" selector and the "value-b" selector. As a result, we're only constructing a single TopologyGroup.

jmdeal:
I'm going to close this out in favor of https://github.com/kubernetes-sigs/karpenter/pull/2479 - the two PRs are significantly different in approach.

PR #2472, opened 2025-08-30
fix: Fix nil pointer exception for multiNodeConsolidation

jonathan-innis:
Fixes #N/A 

**Description**


pkg/controllers/disruption/consolidation_test.go
```diff
@@ -4672,11 +4773,6 @@ var _ = Describe("Consolidation", func() {
 			cloudProvider.InstanceTypes = []*cloudprovider.InstanceType{leastExpensiveInstance, mostExpensiveInstance}
 		})
 
-		AfterEach(func() {
```
DerekFrank, line 4675: Why is this no longer necessary?

pkg/controllers/disruption/consolidation.go
```diff
@@ -195,14 +195,13 @@ func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...
...
 	results.NewNodeClaims[0], err = results.NewNodeClaims[0].RemoveInstanceTypeOptionsByPriceAndMinValues(results.NewNodeClaims[0].Requirements, candidatePrice)
-
 	if err != nil {
 		if len(candidates) == 1 {
 			c.recorder.Publish(disruptionevents.Unconsolidatable(candidates[0].Node, candidates[0].NodeClaim, fmt.Sprintf("Filtering by price: %v", err))...)
 		}
 		return Command{}, nil
 	}
-	if len(results.NewNodeClaims[0].NodeClaimTemplate.InstanceTypeOptions) == 0 {
+	if len(results.NewNodeClaims[0].InstanceTypeOptions) == 0 {
```
DerekFrank, line 204: Was this always wrong?

jigisha620: This isn't necessarily wrong, right @jonathan-innis?

pkg/controllers/disruption/consolidation_test.go
jonathan-innis: It was never necessary since options are always reset to the default (which uses a strict policy) after every test anyways -- you can see this at the top-level where we set inject `test.Options()` back into the context, overriding it

pkg/controllers/disruption/consolidation.go
jonathan-innis: Yeah, it's not wrong -- it's just matching style elsewhere -- it's the same data structure, just shortening the call to it

DerekFrank (approved):

PR #2476, opened 2025-09-02
fix: support arbitrary reserved capacity labels for drift

jmdeal:
Fixes #N/A 

**Description**

This PR updates the instance type drift check to remove an arbitrary set of labels associated with reserved offerings, rather than just `cloudprovider.ReservationIDLabel`. This is to support the additional `karpenter.k8s.aws/capacity-reservation-type` label which is included as a requirement starting in 1.6 for the AWS provider. This wasn't previously caught due to the inclusion of caching for the `InstanceTypeNotFound` check.

This PR also fixes an issue with the caching for `InstanceTypeNotFound` where we would add a cache entry regardless of if we found the instance type or not. As a result, on the first iteration after the cache expires we would mark the NodeClaim as drifted and we would remove it on the second iteration. This change does mean that we will call the cloudprovider's `GetInstanceTypes` call each time when the instance type isn't found, but given this should be extremely rare (e.g. instance type retirement), this isn't a concern from a performance perspective.

`make presubmit`

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/drift.go
+++ b/pkg/controllers/nodeclaim/disruption/drift.go
@@ -99,6 +99,8 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1.NodePool, nodeClaim 
 		if reason := instanceTypeNotFound(its, nodeClaim); reason != "" {
 			return reason, nil
 		}
+		// Only add a cache entry once we've validated that an instance type exists. We only cache a successful check rather
+		// that the result to ensure we respond quickly to transient abnormalities in the GetInstanceTypes response.
 		d.instanceTypeNotFoundCheckCache.SetDefault(string(nodeClaim.UID), nil)
 	}
 	// Then check if it's drifted from the cloud provider side.


pkg/controllers/nodeclaim/disruption/drift.go
```diff
@@ -96,10 +96,10 @@ func (d *Drift) isDrifted(ctx context.Context, nodePool *v1.NodePool, nodeClaim
 		if err != nil {
 			return "", err
 		}
-		d.instanceTypeNotFoundCheckCache.SetDefault(string(nodeClaim.UID), nil)
 		if reason := instanceTypeNotFound(its, nodeClaim); reason != "" {
 			return reason, nil
 		}
+		d.instanceTypeNotFoundCheckCache.SetDefault(string(nodeClaim.UID), nil)
```
jonathan-innis, line 104: nit: maybe add a comment on top of this for why we are only setting the cache here

jonathan-innis (approved):
/lgtm
/approve

PR #2479, opened 2025-09-03
fix: avoid hash collisions with duplicate match expressions

jmdeal:
Fixes #N/A 

**Description**

When a topology spread constraint defines duplicate expressions in it's label selector, it can result in a hash collision with other `TopologyGroups` even if they have a different expression. This is a manifestation of https://github.com/mitchellh/hashstructure/issues/36. 

This issue surfaced itself when we enabled k8s 1.34 testing. This is due to https://github.com/kubernetes/kubernetes/pull/129874, which changed the behavior of `matchLabelKeys` for TSC to match that of pod affinity. As a result, both Karpenter and the API server are injecting expressions for `MatchLabelKeys`. Consider the following example from our tests:

```yaml
kind: Pod
apiVersion: v1
metadata:
  labels:
    test-label: "value-a"
spec:
  topologySpreadConstraints:
  - topologyKey: "kubernetes.io/hostname"
    whenUnsatisfiable: "DoNotSchedule"
    matchLabelKeys:
    - "test-label"
    maxSkew: 1
```

In memory, this results in the following label selector:
```yaml
labelSelector:
  matchExpressions:
  - key: "test-label"
    operator: "In"
    values:
    - "value-a"
  - key: "test-label"
    operator: "In"
    values:
    - "value-a"
```

Now if we have a second set of pods, where `test-label` is `value-b`, we would expect the same label selector except that instead of `value-a` it would be `value-b` in `values`. We would also expect to construct a separate topology group for pods with each of those label selectors. However, because of the aforementioned bug the two selectors have a hash collision and we use the same topology group for both.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/topologygroup.go
+++ b/pkg/controllers/provisioning/scheduling/topologygroup.go
@@ -182,20 +182,37 @@ func (t *TopologyGroup) IsOwnedBy(key types.UID) bool {
 // with self anti-affinity, we track that as a single topology with 100 owners instead of 100x topologies.
 func (t *TopologyGroup) Hash() uint64 {
 	return lo.Must(hashstructure.Hash(struct {
-		TopologyKey string
-		Type        TopologyType
-		Namespaces  sets.Set[string]
-		RawSelector *metav1.LabelSelector
-		MaxSkew     int32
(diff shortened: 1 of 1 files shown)


jonathan-innis (approved):
/lgtm
/approve

PR #2483, opened 2025-09-04
refactor: use iterator for iterating state nodes

jmdeal:
Fixes #N/A 

**Description**

Replaces the `cluster.ForEachNode` function with an iterator based implementation. This is the same behaviorally - the read lock is held until the scope of the for loop is escaped - but we can now iterate over our state nodes with standard for loop syntax.

`make presubmit`

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -19,6 +19,7 @@ package state
 import (
 	"context"
 	"fmt"
+	"iter"
 	"maps"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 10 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -229,22 +230,23 @@ func (c *Cluster) ForPodsWithAntiAffinity(fn func(p *corev1.Pod, n *corev1.Node)
...
-func (c *Cluster) ForEachNode(f func(n *StateNode) bool) {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
-
-	for _, node := range c.nodes {
-		if !f(node) {
-			return
+// Nodes returns an iterator which iterates over the state nodes in the cluster under a read-lock. It is not safe to
+// store the state.StateNode object and it should only be accessed while the iterator is active.
+func (c *Cluster) Nodes() iter.Seq[*StateNode] {
```
DerekFrank, line 235: Doesn't this mean we actually have an RLock while we're still looping?

jonathan-innis:
/hold

Acknowledging comments

pkg/controllers/state/cluster.go
jmdeal: Yes, that's the intention since we don't want to see a partial update

jmdeal:
Acked comments, /unhold

PR #2484, opened 2025-09-04
feat: add kwok tests for Static Capacity feature

sumukha-radhakrishna:
**Description**
Adds additional kwok tests to ensure Static Capacity is working as expected.

make presubmit


pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -80,7 +90,7 @@ func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.
 		return reconcile.Result{RequeueAfter: time.Minute}, nil
 	}
 
-	log.FromContext(ctx).WithValues("NodePool", klog.KObj(np), "current", runningNodeClaims, "desired", desiredReplicas, "deprovisionCount", nodeClaimsToDeprovision).
+	log.FromContext(ctx).WithValues("NodePool", klog.KObj(np), "current", runningNodeClaims, "desired", desiredReplicas, "deprovision-count", nodeClaimsToDeprovision).
```
jonathan-innis: The NodePool is automatically injected into the request context so you shouldn't have to add this to the log structured values yourself

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -159,3 +169,69 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
+// 2. If more nodes needed, nodes with lowest disruption cost (nodes with pods that have do-not-disrupt will have highest cost)
+func (c *Controller) getDeprovisioningCandidates(ctx context.Context, np *v1.NodePool, nodes []*state.StateNode, count int) []*state.StateNode {
+	// First get empty nodes
+	emptyNodes := lo.Filter(nodes, func(node *state.StateNode, _ int) bool {
+		pods, err := node.Pods(ctx, c.kubeClient)
+		if err != nil {
+			log.FromContext(ctx).WithValues("node", node.Name()).Error(err, "unable to list pods, treating as non-empty")
+			return false
+		}
+		return len(pods) == 0 || lo.EveryBy(pods, pod.IsOwnedByDaemonSet) && lo.NoneBy(pods, pod.HasDoNotDisrupt)
```
jonathan-innis, line 184: What was the rationale for deviating from the emptiness check logic that we have in the disruption controller? Is there a reason that this line had to specifically be different?

pkg/utils/nodepool/nodepool.go
```diff
@@ -76,15 +87,28 @@ func ListManaged(ctx context.Context, c client.Client, cloudProvider cloudprovid
 	}), nil
 }
 
-func NodeClaimEventHandler() handler.EventHandler {
-	return handler.EnqueueRequestsFromMapFunc(func(_ context.Context, o client.Object) []reconcile.Request {
+func NodeClaimEventHandler(c client.Client, staticOnly bool) handler.EventHandler {
```
jonathan-innis: Is this better handled with the `...options` variadic kind of handling? That way that callers that don't care about specifying False for this every time don't have to handle passing that in

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -41,7 +45,13 @@ import (
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	disruptionutils "sigs.k8s.io/karpenter/pkg/utils/disruption"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
+	"sigs.k8s.io/karpenter/pkg/utils/pod"
+)
+
+const (
+	TerminationReason = "deprovisioned"
```
jonathan-innis: I'm not in love with this termination reason name that we have given -- can we do something like `scale_down` or `overprovisioned` or something else which better fits a case where we have a static NodePool that has too many nodes against the desired replicas

pkg/controllers/static/deprovisioning/controller.go
sumukha-radhakrishna: Ideally we should have common logic for fetching deprovisioning candidates and disruption candidates.
Right now candidate selection logic GetCandidates() is written in such a way that it can only be called from within disruption controller. (Expects queue to be passed into GetCandidate func, we have to pass in queue to deprovisioning controller just to call the func).
It requires little bit of refactor and we wanted to keep the diff minimal. Ideally we should have a common queue controller that creates and deletes all NodeClaims so we just add cmds from disruption/provisioning/deprovisioning controller.

pkg/controllers/static/deprovisioning/controller.go
sumukha-radhakrishna: `provisioned` and `overprovisioned` sounds right to me

Makefile
```diff
@@ -7,7 +7,8 @@ HELM_OPTS ?= --set logLevel=debug \
 			--set controller.resources.requests.memory=1Gi \
 			--set controller.resources.limits.cpu=1 \
 			--set controller.resources.limits.memory=1Gi \
-			--set settings.featureGates.nodeRepair=true
+			--set settings.featureGates.nodeRepair=true \
+			--set settings.featureGates.staticCapacity=true
```
DerekFrank, line 11: Does this enable it by default?

pkg/controllers/disruption/staticdrift.go
```diff
@@ -84,7 +84,7 @@ func (d *StaticDrift) ComputeCommands(ctx context.Context, disruptionBudgetMappi
 		maxAllowedDrifts := d.cluster.NodePoolState.ReserveNodeCount(npName, nodeLimit, maxDrifts)
 
 		if maxAllowedDrifts <= 0 {
-			return []Command{}, nil
```
DerekFrank, line 87: Shouldn't we panic if we get a <0 answer?

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -65,7 +75,7 @@ func NewController(kubeClient client.Client, cluster *state.Cluster, cloudProvid
 // Note/todo : We have race bw queue and deprovisioning controller, while handling drift we create new replacements and then mark old as deleting
 // Although the delta is small, we need to track replacing NodeClaims to solve it the right way
 func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.Result, error) {
-	ctx = injection.WithControllerName(ctx, "deprovisioning.static")
+	ctx = injection.WithControllerName(ctx, "static.deprovisioning")
```
DerekFrank, line 78: Should this mirror the termination reason? static.overprovisioning? Or are we trying to establish that "overprovisioned" is the noun and deprovisioning is the verb?

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -159,3 +169,69 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
+
+	candidates := lo.Slice(emptyNodes, 0, count)
+	remaining := count - len(candidates)
+
+	if remaining == 0 {
+		return candidates
+	}
+
+	// Get non-empty nodes with their costs
+	type NonEmptyNodes struct {
```
DerekFrank: nit: `NonEmptyNode`

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -159,3 +169,69 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
+		}, true
+	})
+
+	slices.SortFunc(nonEmptyNodes, func(i, j NonEmptyNodes) int {
+		// If one node has do-not-disrupt pods and the other doesn't, the one without should come first
+		if i.hasDoNotDisrupt != j.hasDoNotDisrupt {
+			return lo.Ternary(i.hasDoNotDisrupt, 1, -1)
+		}
+		// If neither has do-not-disrupt pods, compare their costs
+		return cmp.Compare(disruptionutils.ReschedulingCost(ctx, i.pods)*disruptionutils.LifetimeRemaining(c.clock, np, i.node.NodeClaim),
```
DerekFrank, line 226: Do we want to compose this out and unit test it?

Makefile
sumukha-radhakrishna: We dont intend to, let me check again. Afaiu, these opts are used just for kwok tests, correct me if I am wrong

pkg/controllers/disruption/staticdrift.go
sumukha-radhakrishna: We wont get <0

pkg/controllers/static/deprovisioning/controller.go
sumukha-radhakrishna: static.deprovisioning controller terminated nodeclaims due to overprovisioned made sense.

pkg/controllers/static/deprovisioning/controller.go
sumukha-radhakrishna: We already have UT covering these cases. Do you want us testing anything specifically?

test/suites/regression/staticcapacity_test.go
```diff
@@ -0,0 +1,442 @@
...
+						},
+					},
+				})
+			}
+		})
+
+		It("should create static NodeClaims to meet desired replicas", func() {
+			nodePool.Spec.Replicas = lo.ToPtr(int64(10))
+			env.ExpectCreated(nodeClass, nodePool)
+			env.EventuallyExpectInitializedNodeCount("==", 10)
```
DerekFrank, line 58: Idk who wrote this original function, but shouldn't the signature be something like `env.EventuallyExpectInitializedNodeCount().ToEqual(10)`?

test/suites/regression/staticcapacity_test.go
```diff
@@ -0,0 +1,442 @@
...
+			env.EventuallyExpectNodeClaimsReady(nodeClaims...)
+
+			nodePool.Spec.Replicas = lo.ToPtr(int64(15))
+			env.ExpectUpdated(nodePool)
+			nodes := env.EventuallyExpectInitializedNodeCount("==", 15)
+			nodeClaims = env.EventuallyExpectCreatedNodeClaimCount("==", 15)
+			env.EventuallyExpectNodeClaimsReady(nodeClaims...)
+
+			for _, node := range nodes {
+				Expect(node.Labels).To(HaveKeyWithValue(v1.CapacityTypeLabelKey, v1.CapacityTypeOnDemand))
```
DerekFrank: Why are we checking if they're on demand? Shouldn't we be checking that they're from the static nodepool instead?

test/suites/regression/staticcapacity_test.go
```diff
@@ -0,0 +1,442 @@
...
+
+		It("should respect node limits when provisioning", func() {
+			nodePool.Spec.Replicas = lo.ToPtr(int64(10))
+			nodePool.Spec.Limits = v1.Limits{
+				corev1.ResourceName("nodes"): resource.MustParse("5"),
+			}
+			env.ExpectCreated(nodeClass, nodePool)
+
+			// Should only create 5 nodes due to limit
+			env.EventuallyExpectInitializedNodeCount("==", 5)
```
DerekFrank: This is a little odd. Shouldn't we be checking that we never make more than 5?

test/suites/regression/staticcapacity_test.go
```diff
@@ -0,0 +1,442 @@
...
+			env.ExpectUpdated(nodePool)
+
+			// Verify the drifted node was replaced
+			env.EventuallyExpectDrifted(nodeClaims...)
+
+			// Should create a replacement node and then remove the drifted one
+			env.ConsistentlyExpectDisruptionsUntilNoneLeft(5, 5, 5*time.Minute)
+		})
+
+		It("should handle drift with node limits", func() {
```
DerekFrank: Maybe we should also check that limits and budgets are respected, whichever is lower?

test/suites/regression/staticcapacity_test.go
```diff
@@ -0,0 +1,442 @@
...
+							Requirements: nodePool.Spec.Template.Spec.Requirements,
+							NodeClassRef: nodePool.Spec.Template.Spec.NodeClassRef,
+						},
+					},
+				},
+			})
+			label = map[string]string{"app": "large-app"}
+		})
+
+		It("should not affect dynamic NodeClaim creation when static NodePool is present", func() {
```
DerekFrank: Can you write a test that verifies that the dynamic nodepool provisions capacity once the static one runs out?

test/suites/regression/staticcapacity_test.go
DerekFrank: Well outside the scope of this PR of course

sumukha-radhakrishna: I think comparator and value makes sense

test/suites/regression/staticcapacity_test.go
sumukha-radhakrishna: Yeah I can check that we dont go over the limit count

test/suites/regression/staticcapacity_test.go
sumukha-radhakrishna: We already have that covered in the existing test, when static nodeclaims cant satisfy pod constraints create dynamic nodeclaims

Issue #715, opened 2023-03-22
DaemonSets not being correctly calculated when choosing a node


kfirsch:
### Version

__Karpenter Version:__ v0.24.0

__Kubernetes Version:__ v1.21.0

### Context

Due to the significant resource usage of certain Daemonsets, particularly when operating on larger machines, we have chosen to divide these Daemonsets based on affinity rules that use Karpenter's labels such as `karpenter.k8s.aws/instance-cpu` or `karpenter.k8s.aws/instance-size`.

### Expected Behavior

When selecting a node for provisioning, Karpenter must only consider the appropriate Daemonsets that will run on that node. 

### Actual Behavior

It appears that Karpenter is wrongly including all of the split Daemonsets instead of only the appropriate one, which can result in poor instance selection when provisioning new nodes or inaccurate consolidation actions.

### Steps to Reproduce the Problem

+ Create a fresh cluster with Karpenter deployed and a default provisioner:
```
apiVersion: karpenter.sh/v1alpha5
kind: Provisioner
metadata:
  name: default
spec:
  providerRef:
    name: default
  requirements:
    - key: "node.kubernetes.io/instance-type"
      operator: In
      values: ["c6i"]
    - key: "karpenter.sh/capacity-type"
      operator: In
      values: ["on-demand"]
  consolidation:
    enabled: true
```
+ Duplicate one of your Daemonsets and split them into small/large machines using the following settings:
```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - key: karpenter.k8s.aws/instance-cpu
          operator: Lt
          values:
          - "31"

  resources:
    requests:
      cpu: 1
```
```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
      - matchExpressions:
        - key: karpenter.k8s.aws/instance-cpu
          operator: Gt
          values:
          - "30"

  resources:
    requests:
      cpu: 10
```
+ Create a simple Pod with 1 CPU request, Karpenter Should provision a 2 or max 4 cpu Instance but will instead provision a large >10 cpu machine due wrongly include the bigger Daemonset in the 2\4\8 cpu evaluation. 

+ Same behavior when using `karpenter.k8s.aws/instance-size` or even `podAntiAffinity` rules in the Daemonset affinities.

Thank you for your help in addressing this issue.

[...]

PR #2486, opened 2025-09-05
feat: account for instance type requirements for daemonset scheduling

saurav-agarwalla:
Fixes  https://github.com/kubernetes-sigs/karpenter/issues/715

**Description**

*feat: account for instance type requirements for daemonset scheduling*

The current scheduling logic doesn't take instance type requirements into account when calculating the daemonset overhead. Some real world examples of how this has affected users are documented in https://github.com/kubernetes-sigs/karpenter/issues/715 but in the worst case, it is possible that scheduling completely fails when daemonset overheads are included which would never be satisfied in a given list of instance types for a specific nodepool. 

For example, say a daemonset needs GPU resources but the nodepool requirements filter out GPU instances. What will happen is that the GPU resources will be added to the overhead but since none of the instance types can support that overhead, scheduling will never succeed.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/scheduling/nodeclaim.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaim.go
@@ -43,8 +43,8 @@ type NodeClaim struct {
 	Pods               []*corev1.Pod
 	reservationManager *ReservationManager
 	topology           *Topology
-	hostPortUsage      *scheduling.HostPortUsage
-	daemonResources    corev1.ResourceList
+	hostPortUsage      map[*cloudprovider.InstanceType]*scheduling.HostPortUsage
+	daemonResources    map[*cloudprovider.InstanceType]corev1.ResourceList
 	hostname           string
 
 	// We store the reserved offerings rather than appending reservation ID labels for two reasons:
@@ -83,8 +83,8 @@ var nodeID int64
 func NewNodeClaim(
 	nodeClaimTemplate *NodeClaimTemplate,
 	topology *Topology,
-	daemonResources corev1.ResourceList,
-	hostPortUsage *scheduling.HostPortUsage,
+	daemonResources map[*cloudprovider.InstanceType]corev1.ResourceList,
+	hostPortUsage map[*cloudprovider.InstanceType]*scheduling.HostPortUsage,
 	instanceTypes []*cloudprovider.InstanceType,
 	reservationManager *ReservationManager,
 	reservedOfferingMode ReservedOfferingMode,
@@ -95,7 +95,7 @@ func NewNodeClaim(
 	template.Requirements.Add(nodeClaimTemplate.Requirements.Values()...)
 	template.Requirements.Add(scheduling.NewRequirement(corev1.LabelHostname, corev1.NodeSelectorOpIn, hostname))
 	template.InstanceTypeOptions = instanceTypes
-	template.Spec.Resources.Requests = daemonResources
+	template.Spec.Resources.Requests = corev1.ResourceList{}
 	return &NodeClaim{
 		NodeClaimTemplate:    template,
 		hostPortUsage:        hostPortUsage,
@@ -117,11 +117,6 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 		return nil, nil, nil, err
 	}
 
-	// exposed host ports on the node
-	hostPorts := scheduling.GetHostPorts(pod)
-	if err := n.hostPortUsage.Conflicts(pod, hostPorts); err != nil {
-		return nil, nil, nil, fmt.Errorf("checking host port usage, %w", err)
-	}
 	nodeClaimRequirements := scheduling.NewRequirements(n.Requirements.Values()...)
 
 	// Check NodeClaim Affinity Requirements
@@ -143,7 +138,7 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 	// Check instance type combinations
 	requests := resources.Merge(n.Spec.Resources.Requests, podData.Requests)
 
-	remaining, unsatisfiableKeys, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, podData.Requests, n.daemonResources, requests, relaxMinValues)
+	remaining, unsatisfiableKeys, err := filterInstanceTypesByRequirements(n.InstanceTypeOptions, nodeClaimRequirements, pod, podData.Requests, n.daemonResources, n.hostPortUsage, requests, relaxMinValues)
 	if relaxMinValues {
 		// Update min values on the requirements if they are relaxed
 		for key, minValues := range unsatisfiableKeys {
@@ -169,11 +164,15 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements
 	// Update node
 	n.Pods = append(n.Pods, pod)
 	n.InstanceTypeOptions = instanceTypes
-	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests)
+	// At this point, all the instance types which won't fit the daemonsets have been filtered out so we can add the daemonset overhead to the nodeclaim spec.
+	// It is possible that none of the daemonsets fit on the nodeclaim due to instance type requirements so don't add any overhead in that case.
+	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests, lo.ValueOr(n.daemonResources, lo.FirstOrEmpty(instanceTypes), corev1.ResourceList{}))
 	n.Requirements = nodeClaimRequirements
 	n.topology.Register(corev1.LabelHostname, n.hostname)
 	n.topology.Record(pod, n.Spec.Taints, nodeClaimRequirements, scheduling.AllowUndefinedWellKnownLabels)
-	n.hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	lo.ForEach(lo.Values(n.hostPortUsage), func(hostPortUsage *scheduling.HostPortUsage, _ int) {
+		hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	})
 	n.reservationManager.Reserve(n.hostname, offeringsToReserve...)
 	n.releaseReservedOfferings(n.reservedOfferings, offeringsToReserve)
 	n.reservedOfferings = offeringsToReserve
@@ -370,7 +369,7 @@ func (e InstanceTypeFilterError) Error() string {
 }
 
 //nolint:gocyclo
-func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceType, requirements scheduling.Requirements, podRequests, daemonRequests, totalRequests corev1.ResourceList, relaxMinValues bool) (cloudprovider.InstanceTypes, map[string]int, error) {
+func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceType, requirements scheduling.Requirements, pod *corev1.Pod, podRequests corev1.ResourceList, daemonRequests map[*cloudprovider.InstanceType]corev1.ResourceList, hostPortUsage map[*cloudprovider.InstanceType]*scheduling.HostPortUsage, totalRequests corev1.ResourceList, relaxMinValues bool) (cloudprovider.InstanceTypes, map[string]int, error) {
 	unsatisfiableKeys := map[string]int{}
 	// We hold the results of our scheduling simulation inside of this InstanceTypeFilterError struct
 	// to reduce the CPU load of having to generate the error string for a failed scheduling simulation
@@ -383,17 +382,33 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
 		requirementsAndOffering: false,
 		fitsAndOffering:         false,
 
-		requirements:   requirements,
-		podRequests:    podRequests,
-		daemonRequests: daemonRequests,
+		requirements: requirements,
+		podRequests:  podRequests,
 	}
 	remaining := cloudprovider.InstanceTypes{}
 
 	for _, it := range instanceTypes {
+		hostPortUsageForInstanceType := lo.ValueOr(hostPortUsage, it, &scheduling.HostPortUsage{})
+		// exposed host ports on the pod
+		hostPorts := scheduling.GetHostPorts(pod)
+		if err := hostPortUsageForInstanceType.Conflicts(pod, hostPorts); err != nil {
+			continue
+		}
+
+		var totalRequestsForInstanceType corev1.ResourceList
+		daemonRequestsForInstanceType := daemonRequests[it]
+		if len(daemonRequestsForInstanceType) != 0 {
+			err.daemonRequests = daemonRequestsForInstanceType
+			totalRequestsForInstanceType = resources.MergeInto(totalRequestsForInstanceType, totalRequests)
+			totalRequestsForInstanceType = resources.MergeInto(totalRequestsForInstanceType, daemonRequestsForInstanceType)
+		} else {
+			totalRequestsForInstanceType = totalRequests
+		}
+
 		// the tradeoff to not short-circuiting on the filtering is that we can report much better error messages
 		// about why scheduling failed
 		itCompat := compatible(it, requirements)
-		itFits := fits(it, totalRequests)
+		itFits := fits(it, totalRequestsForInstanceType)
 
 		// By using this iterative approach vs. the Available() function it prevents allocations
 		// which have to be garbage collected and slow down Karpenter's scheduling algorithm
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -143,7 +143,7 @@ func NewScheduler(
 	templates := lo.FilterMap(nodePools, func(np *v1.NodePool, _ int) (*NodeClaimTemplate, bool) {
 		var err error
 		nct := NewNodeClaimTemplate(np)
-		nct.InstanceTypeOptions, _, err = filterInstanceTypesByRequirements(instanceTypes[np.Name], nct.Requirements, corev1.ResourceList{}, corev1.ResourceList{}, corev1.ResourceList{}, minValuesPolicy == karpopts.MinValuesPolicyBestEffort)
+		nct.InstanceTypeOptions, _, err = filterInstanceTypesByRequirements(instanceTypes[np.Name], nct.Requirements, &corev1.Pod{}, corev1.ResourceList{}, map[*cloudprovider.InstanceType]corev1.ResourceList{}, map[*cloudprovider.InstanceType]*scheduling.HostPortUsage{}, corev1.ResourceList{}, minValuesPolicy == karpopts.MinValuesPolicyBestEffort)
 		if len(nct.InstanceTypeOptions) == 0 {
 			if instanceTypeFilterErr, ok := lo.ErrorsAs[InstanceTypeFilterError](err); ok && instanceTypeFilterErr.minValuesIncompatibleErr != nil {
 				recorder.Publish(NoCompatibleInstanceTypes(np, true))
@@ -194,8 +194,8 @@ type Scheduler struct {
 	existingNodes           []*ExistingNode
 	nodeClaimTemplates      []*NodeClaimTemplate
 	remainingResources      map[string]corev1.ResourceList // (NodePool name) -> remaining resources for that NodePool
-	daemonOverhead          map[*NodeClaimTemplate]corev1.ResourceList
-	daemonHostPortUsage     map[*NodeClaimTemplate]*scheduling.HostPortUsage
+	daemonOverhead          map[*NodeClaimTemplate]map[*cloudprovider.InstanceType]corev1.ResourceList
+	daemonHostPortUsage     map[*NodeClaimTemplate]map[*cloudprovider.InstanceType]*scheduling.HostPortUsage
 	cachedPodData           map[types.UID]*PodData // (Pod Namespace/Name) -> pre-computed data for pods to avoid re-computation and memory usage
 	preferences             *Preferences
 	topology                *Topology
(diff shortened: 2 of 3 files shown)


pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -117,11 +117,6 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 		return nil, nil, nil, err
 	}
 
-	// exposed host ports on the node
-	hostPorts := scheduling.GetHostPorts(pod)
-	if err := n.hostPortUsage.Conflicts(pod, hostPorts); err != nil {
-		return nil, nil, nil, fmt.Errorf("checking host port usage, %w", err)
-	}
```
saurav-agarwalla, line 124: I need some feedback on this particular change: the previous behavior skipped a pod when the host port usage of that pod conflicts with daemonset pods. However, now that we consider instance type requirements for daemonsets, daemonsets might not be able to schedule on all instance types. As a result, I've moved this logic to where we filter out instance types instead. 

I want to confirm that this doesn't unknowingly break any other use-cases or assumptions.

logyball: you could add a test for this behavior:

```go
		It("should not account for daemonset hostports when daemonset would not schedule on selected node", func() {
			ExpectApplied(ctx, env.Client, test.NodePool(), test.DaemonSet(
				test.DaemonSetOptions{PodOptions: test.PodOptions{
					ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("4"), corev1.ResourceMemory: resource.MustParse("4Gi")}},
					NodeRequirements:     []corev1.NodeSelectorRequirement{{Key: fake.LabelInstanceSize, Operator: corev1.NodeSelectorOpIn, Values: []string{"large"}}},
					HostPorts:            []int32{8080},
				}},
			))
			ExpectApplied(ctx, env.Client, test.NodePool(), test.DaemonSet(
				test.DaemonSetOptions{PodOptions: test.PodOptions{
					ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("2"), corev1.ResourceMemory: resource.MustParse("2Gi")}},
					NodeRequirements:     []corev1.NodeSelectorRequirement{{Key: fake.LabelInstanceSize, Operator: corev1.NodeSelectorOpIn, Values: []string{"small"}}},
					HostPorts:            []int32{8081},
				}},
			))
			pod := test.UnschedulablePod(
				test.PodOptions{
					ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceCPU: resource.MustParse("1"), corev1.ResourceMemory: resource.MustParse("1Gi")}},
					NodeRequirements:     []corev1.NodeSelectorRequirement{{Key: fake.LabelInstanceSize, Operator: corev1.NodeSelectorOpIn, Values: []string{"small"}}},
					HostPorts:            []int32{8080},
				},
			)
			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
			node := ExpectScheduled(ctx, env.Client, pod)

			allocatable := instanceTypeMap[node.Labels[corev1.LabelInstanceTypeStable]].Capacity
			Expect(node.Labels[fake.LabelInstanceSize]).To(Equal("small"))
		})
```

In my quick experimentation, the above test ^ fails on the `main` branch but succeeds on your branch.

logyball:
One other thing that I thought of, which I'll definitely want a Karpenter expert to weigh in on, is the relative performance penalty introduced by this.  When running the benchmarking tests, it does look like a regression (expected, as the scheduling code is now more complicated).  The question is whether this performance penalty is blocking or not. 

```sh
# run with go test -tags=test_performance -run=XXX -bench=. ./pkg/controllers/provisioning/scheduling/ -count=10 -timeout=30m 

goos: darwin
goarch: arm64
cpu: Apple M1 Pro
                      │  /tmp/old   │               /tmp/new                │
                      │   sec/op    │    sec/op     vs base                 │
Scheduling1-10          189.8µ ± 2%    313.6µ ± 3%   +65.25% (p=0.000 n=10)
Scheduling50-10         45.56m ± 9%    61.18m ± 6%   +34.29% (p=0.000 n=10)
Scheduling100-10        86.72m ± 9%   123.43m ± 3%   +42.33% (p=0.000 n=10)
Scheduling500-10        424.4m ± 6%    603.0m ± 4%   +42.08% (p=0.000 n=10)
Scheduling1000-10       824.7m ± 4%   1218.2m ± 6%   +47.73% (p=0.000 n=10)
Scheduling2000-10        1.725 ± 5%     2.459 ± 7%   +42.53% (p=0.000 n=10)
Scheduling5000-10        4.420 ± 7%     6.536 ± 3%   +47.88% (p=0.000 n=10)
Scheduling10000-10       9.651 ± 6%    13.731 ± 2%   +42.28% (p=0.000 n=10)
Scheduling20000-10       23.25 ± 1%     30.87 ± 2%   +32.76% (p=0.000 n=10)
RespectPreferences-10    15.91 ± 5%     16.44 ± 1%         ~ (p=0.353 n=10)
IgnorePreferences-10    775.0m ± 4%   1572.7m ± 2%  +102.94% (p=0.000 n=10)
geomean                 623.5m         898.1m        +44.04%

                      │   /tmp/old   │                /tmp/new                │
                      │    nodes     │     nodes      vs base                 │
Scheduling1-10           1.000 ±  0%    1.000 ±   0%       ~ (p=1.000 n=10) ¹
Scheduling50-10          10.00 ±  0%    10.00 ±   0%       ~ (p=1.000 n=10) ¹
Scheduling100-10         20.00 ±  0%    20.00 ±   0%       ~ (p=1.000 n=10) ¹
Scheduling500-10         100.0 ±  0%    100.0 ±   0%       ~ (p=1.000 n=10) ¹
Scheduling1000-10        200.0 ±  0%    200.0 ±   0%       ~ (p=1.000 n=10) ¹
Scheduling2000-10        400.0 ±  0%    400.0 ±   0%       ~ (p=1.000 n=10) ¹
Scheduling5000-10       1.000k ±  0%   1.000k ±   0%       ~ (p=1.000 n=10) ¹
Scheduling10000-10      2.000k ±  0%   2.000k ±   0%       ~ (p=1.000 n=10) ¹
Scheduling20000-10      4.000k ±  0%   4.000k ±   0%       ~ (p=1.000 n=10) ¹
RespectPreferences-10   4.000k ±  0%   4.000k ±   0%       ~ (p=1.000 n=10) ¹
IgnorePreferences-10     8.500 ± 88%    6.000 ± 167%       ~ (p=0.922 n=10)
geomean                  140.9          136.5         -3.12%
¹ all samples are equal

                      │  /tmp/old   │               /tmp/new               │
                      │    pods     │    pods      vs base                 │
Scheduling1-10           1.000 ± 0%    1.000 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling50-10          50.00 ± 0%    50.00 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling100-10         100.0 ± 0%    100.0 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling500-10         500.0 ± 0%    500.0 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling1000-10       1.000k ± 0%   1.000k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling2000-10       2.000k ± 0%   2.000k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling5000-10       5.000k ± 0%   5.000k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling10000-10      10.00k ± 0%   10.00k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling20000-10      20.00k ± 0%   20.00k ± 0%       ~ (p=1.000 n=10) ¹
RespectPreferences-10   4.000k ± 0%   4.000k ± 0%       ~ (p=1.000 n=10) ¹
IgnorePreferences-10    4.000k ± 0%   4.000k ± 0%       ~ (p=1.000 n=10) ¹
geomean                  794.8         794.8       +0.00%
¹ all samples are equal

                      │   /tmp/old   │              /tmp/new               │
                      │   pods/sec   │  pods/sec    vs base                │
Scheduling1-10          5.269k ±  2%   3.188k ± 3%  -39.49% (p=0.000 n=10)
Scheduling50-10         1098.0 ±  8%    817.5 ± 6%  -25.54% (p=0.000 n=10)
Scheduling100-10        1153.0 ± 10%    810.2 ± 3%  -29.74% (p=0.000 n=10)
Scheduling500-10        1178.5 ±  6%    829.2 ± 4%  -29.64% (p=0.000 n=10)
Scheduling1000-10       1212.5 ±  4%    820.9 ± 6%  -32.30% (p=0.000 n=10)
Scheduling2000-10       1159.5 ±  5%    813.3 ± 6%  -29.86% (p=0.000 n=10)
Scheduling5000-10       1131.0 ±  7%    765.0 ± 3%  -32.37% (p=0.000 n=10)
Scheduling10000-10      1036.0 ±  5%    728.2 ± 2%  -29.71% (p=0.000 n=10)
Scheduling20000-10       860.0 ±  1%    647.8 ± 2%  -24.68% (p=0.000 n=10)
RespectPreferences-10    251.5 ±  5%    243.2 ± 1%        ~ (p=0.325 n=10)
IgnorePreferences-10    5.162k ±  4%   2.543k ± 2%  -50.73% (p=0.000 n=10)
geomean                 1.275k          885.1       -30.58%
```

pkg/controllers/provisioning/scheduling/nodeclaim.go
saurav-agarwalla: Thanks for your review, @logyball. 

Good call out on adding a separate test for this. I don't think this change in behavior should break anything but precisely why I wanted to make sure that someone more familiar with it can confirm. :)

saurav-agarwalla:
@logyball good catch on the performance regression. I am not yet sure how representative that benchmark is of real-world usage and if it needs to block this given the benefits but I am going to take a stab to see if I can make some optimizations to get rid of some of the performance penalty.

saurav-agarwalla:
@logyball made some performance optimizations, added the test and also ensured that the (now) minor performance penalty is only for when customers use daemon set pods.

New results are below and I think they look reasonable:
```
➜  ~ benchstat /tmp/old /tmp/new
goos: darwin
goarch: arm64
cpu: Apple M3 Pro
                      │   /tmp/old   │               /tmp/new                │
                      │    sec/op    │    sec/op      vs base                │
Scheduling1-12          164.3µ ±  8%    190.5µ ±  4%  +15.94% (p=0.000 n=10)
Scheduling50-12         37.51m ±  7%    40.54m ± 15%        ~ (p=0.143 n=10)
Scheduling100-12        76.69m ±  9%    79.59m ± 10%        ~ (p=0.393 n=10)
Scheduling500-12        388.3m ± 27%    404.6m ±  7%        ~ (p=0.684 n=10)
Scheduling1000-12       776.8m ± 13%    898.3m ± 22%        ~ (p=0.075 n=10)
Scheduling2000-12        1.540 ± 12%     1.726 ± 24%  +12.06% (p=0.001 n=10)
Scheduling5000-12        3.738 ± 11%     4.197 ± 12%  +12.30% (p=0.001 n=10)
Scheduling10000-12       8.113 ±  9%    10.153 ± 17%  +25.15% (p=0.000 n=10)
Scheduling20000-12       20.19 ± 25%     22.03 ± 11%        ~ (p=0.105 n=10)
RespectPreferences-12    11.66 ± 15%     10.85 ±  5%   -6.95% (p=0.000 n=10)
IgnorePreferences-12    649.5m ±  6%   1006.5m ±  8%  +54.96% (p=0.000 n=10)
geomean                 534.6m          604.7m        +13.12%

                      │   /tmp/old   │               /tmp/new                │
                      │    nodes     │    nodes      vs base                 │
Scheduling1-12           1.000 ±  0%    1.000 ±  0%       ~ (p=1.000 n=10) ¹
Scheduling50-12          10.00 ±  0%    10.00 ±  0%       ~ (p=1.000 n=10) ¹
Scheduling100-12         20.00 ±  0%    20.00 ±  0%       ~ (p=1.000 n=10) ¹
Scheduling500-12         100.0 ±  0%    100.0 ±  0%       ~ (p=1.000 n=10) ¹
Scheduling1000-12        200.0 ±  0%    200.0 ±  0%       ~ (p=1.000 n=10) ¹
Scheduling2000-12        400.0 ±  0%    400.0 ±  0%       ~ (p=1.000 n=10) ¹
Scheduling5000-12       1.000k ±  0%   1.000k ±  0%       ~ (p=1.000 n=10) ¹
Scheduling10000-12      2.000k ±  0%   2.000k ±  0%       ~ (p=1.000 n=10) ¹
Scheduling20000-12      4.000k ±  0%   4.000k ±  0%       ~ (p=1.000 n=10) ¹
RespectPreferences-12   4.000k ±  0%   4.000k ±  0%       ~ (p=1.000 n=10) ¹
IgnorePreferences-12    11.000 ± 45%    8.500 ± 88%       ~ (p=0.444 n=10)
geomean                  144.3          140.9        -2.32%
¹ all samples are equal

                      │  /tmp/old   │               /tmp/new               │
                      │    pods     │    pods      vs base                 │
Scheduling1-12           1.000 ± 0%    1.000 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling50-12          50.00 ± 0%    50.00 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling100-12         100.0 ± 0%    100.0 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling500-12         500.0 ± 0%    500.0 ± 0%       ~ (p=1.000 n=10) ¹
Scheduling1000-12       1.000k ± 0%   1.000k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling2000-12       2.000k ± 0%   2.000k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling5000-12       5.000k ± 0%   5.000k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling10000-12      10.00k ± 0%   10.00k ± 0%       ~ (p=1.000 n=10) ¹
Scheduling20000-12      20.00k ± 0%   20.00k ± 0%       ~ (p=1.000 n=10) ¹
RespectPreferences-12   4.000k ± 0%   4.000k ± 0%       ~ (p=1.000 n=10) ¹
IgnorePreferences-12    4.000k ± 0%   4.000k ± 0%       ~ (p=1.000 n=10) ¹
geomean                  794.8         794.8       +0.00%
¹ all samples are equal

                      │   /tmp/old   │               /tmp/new               │
                      │   pods/sec   │   pods/sec    vs base                │
Scheduling1-12          6.086k ±  8%   5.250k ±  4%  -13.74% (p=0.000 n=10)
Scheduling50-12         1.333k ±  7%   1.234k ± 13%        ~ (p=0.138 n=10)
Scheduling100-12        1.304k ±  9%   1.256k ± 11%        ~ (p=0.393 n=10)
Scheduling500-12        1.290k ± 21%   1.236k ±  7%        ~ (p=0.644 n=10)
Scheduling1000-12       1.288k ± 11%   1.113k ± 18%        ~ (p=0.072 n=10)
Scheduling2000-12       1.298k ± 14%   1.159k ± 20%  -10.74% (p=0.001 n=10)
Scheduling5000-12       1.338k ± 10%   1.192k ± 11%  -10.95% (p=0.001 n=10)
Scheduling10000-12      1233.0 ±  8%    985.0 ± 14%  -20.12% (p=0.000 n=10)
Scheduling20000-12       990.3 ± 20%    908.6 ± 10%        ~ (p=0.105 n=10)
RespectPreferences-12    343.1 ± 13%    368.7 ±  5%   +7.48% (p=0.000 n=10)
IgnorePreferences-12    6.159k ±  6%   3.974k ±  8%  -35.47% (p=0.000 n=10)
geomean                 1.487k         1.315k        -11.60%
```

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -169,11 +164,15 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements
...
+	// At this point, all the instance types which won't fit the daemonsets have been filtered out so we can add the daemonset overhead to the nodeclaim spec.
+	// It is possible that none of the daemonsets fit on the nodeclaim due to instance type requirements so don't add any overhead in that case.
+	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests, lo.ValueOr(n.daemonResources, lo.FirstOrEmpty(instanceTypes), corev1.ResourceList{}))
 	n.Requirements = nodeClaimRequirements
 	n.topology.Register(corev1.LabelHostname, n.hostname)
 	n.topology.Record(pod, n.Spec.Taints, nodeClaimRequirements, scheduling.AllowUndefinedWellKnownLabels)
-	n.hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	lo.ForEach(lo.Values(n.hostPortUsage), func(hostPortUsage *scheduling.HostPortUsage, _ int) {
+		hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	})
```
jmdeal, line 175: nit: I'd avoid the uses of the lo functions here, we don't use `lo.ForEach` generally (I'd only use it when you're passing an l-value functor) and `lo.Values` just causes an extra allocation. It probably also makes sense to only update the hostPortUsage for compatible instance types rather than updating all instance types all of the time.
```suggestion
	hostPorts := scheduling.GetHostPorts(pod)
	for _, it := range instanceTypes {
		hostPortUsage[it].Add(pod, hostPorts)
	}
```

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -169,11 +164,15 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements
 	// Update node
 	n.Pods = append(n.Pods, pod)
 	n.InstanceTypeOptions = instanceTypes
-	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests)
+	// At this point, all the instance types which won't fit the daemonsets have been filtered out so we can add the daemonset overhead to the nodeclaim spec.
+	// It is possible that none of the daemonsets fit on the nodeclaim due to instance type requirements so don't add any overhead in that case.
+	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests, lo.ValueOr(n.daemonResources, lo.FirstOrEmpty(instanceTypes), corev1.ResourceList{}))
```
jmdeal, line 169: By adding this here, won't we be counting the daemon requests multiple times (each time we add a pod)? Also, do we want to be adding the resources for a random instance type? If we add anything here, I think we'd want to make sure we're adding the smallest set of daemon resources. I think there's an argument that we just don't add daemon resources at all though.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -383,17 +382,33 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
+		// exposed host ports on the pod
+		hostPorts := scheduling.GetHostPorts(pod)
+		if err := hostPortUsageForInstanceType.Conflicts(pod, hostPorts); err != nil {
+			continue
+		}
+
+		var totalRequestsForInstanceType corev1.ResourceList
+		daemonRequestsForInstanceType := daemonRequests[it]
+		if len(daemonRequestsForInstanceType) != 0 {
+			err.daemonRequests = daemonRequestsForInstanceType
```
jmdeal, line 401: I think the behavior of this error is now a little strange, because we're only going to see the daemon requests for the last instance type we evaluated. I don't think it's feasible for us to enumerate the daemonset resources for each  instance type in the error, but there are probably some other options:
- Display the lower and upper bound requests
- Display groups of instance types and their offset (still going to be a lot of info):
```json
{
    "daemonSetOverhead": [
        {
            "instanceTypes": [...],
            "resources": ...
        }
    ]
}

File: pkg/controllers/provisioning/scheduling/nodeclaim.go [hunk 5]
```diff
@@ -43,8 +43,8 @@ type NodeClaim struct {
 	Pods               []*corev1.Pod
 	reservationManager *ReservationManager
 	topology           *Topology
-	hostPortUsage      *scheduling.HostPortUsage
-	daemonResources    corev1.ResourceList
+	hostPortUsage      map[*cloudprovider.InstanceType]*scheduling.HostPortUsage
+	daemonResources    map[*cloudprovider.InstanceType]corev1.ResourceList
```
jmdeal, line 47: While it works, I wouldn't use pointers as map keys here. If, for whatever reason, we need to make a copy of an instance type and try to use it as a key in the future this will fail. I'd just use the instance type's name as the unique identifier.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -770,49 +770,62 @@ func parallelizeUntil(workers, pieces int, doWorkPiece func(int) bool) {
...
 	_ = preferences.toleratePreferNoScheduleTaints(pod)
 	if err := scheduling.Taints(nodeClaimTemplate.Spec.Taints).ToleratesPod(pod); err != nil {
 		return false
 	}
 	for {
+		podRequirements := scheduling.NewStrictPodRequirements(pod)
 		// We don't consider pod preferences for scheduling requirements since we know that pod preferences won't matter with Daemonset scheduling
-		if nodeClaimTemplate.Requirements.IsCompatible(scheduling.NewStrictPodRequirements(pod), scheduling.AllowUndefinedWellKnownLabels) {
+		if nodeClaimTemplate.Requirements.IsCompatible(podRequirements, scheduling.AllowUndefinedWellKnownLabels) &&
+			// We use Intersects instead of IsCompatible for instance type requirements since we want to ignore any custom keys on the daemonset pod since they
```
jmdeal, line 826: Can we add a test for this? IIUC the behavior we're trying to support is a DaemonSet with a node selector on a custom label that's defined on the NodePool, but is not a well-known label. I think we probably don't fully support this type of requirement though - consider a NodePool with multiple possible values for that custom label which could be constrained by the requirements of other pods we're scheduling to it.

jmdeal: It's probably still reasonable for us to support this partially for the more common case where you have a single label value (e.g. OS)

logyball:
@saurav-agarwalla any chance you've had a chance to look at @jmdeal 's review?

logyball:
@saurav-agarwalla - another ping here, this would be super useful for our use cases.  I can take a stab at taking over this PR if you're no longer available to work on it.

saurav-agarwalla:
Apologies for missing the notification on this. I am not sure if I'll be able to get this past the finish line anytime soon so happy to hand this over if you are interested.

PR #2487, opened 2025-09-05
fix: Handle unbound volumes with volumeName defined

jonathan-innis:
Fixes #N/A 

**Description**

A code change was made for Karpenter to filter out Immediate volumes that had not been bound yet by checking for volumes without a `volumeName` and considering if they are using bindingMode `Immediate` or not. There is an additional check that the kube-scheduler does when volumeName is defined, which includes checking if the PVC has the `pv.kubernetes.io/bind-completed` included in its annotations.

See: https://github.com/kubernetes/kubernetes/blob/ecf2c52f756461cfb7ffd5469975ecd635e5feeb/pkg/scheduler/framework/plugins/volumebinding/binder.go#L770

`make presubmit`


pkg/test/storage.go
```diff
@@ -109,6 +109,9 @@ func PersistentVolumeClaim(overrides ...PersistentVolumeClaimOptions) *v1.Persis
 	if len(options.Resources.Requests) == 0 {
 		options.Resources = v1.VolumeResourceRequirements{Requests: v1.ResourceList{v1.ResourceStorage: resource.MustParse("1Gi")}}
 	}
+	if options.VolumeName != "" {
+		options.ObjectMeta.Annotations = lo.Assign(options.ObjectMeta.Annotations, map[string]string{"pv.kubernetes.io/bind-completed": "yes"})
```
saurav-agarwalla: nit: we should use the constant from upstream `volume.AnnBindCompleted`.

saurav-agarwalla (approved):
/lgtm

pkg/test/storage.go
jonathan-innis: Sure 👍🏼

PR #2489, opened 2025-09-05
chore: Change appName from const to var

AndrewMitchell25:
Fixes #N/A 

**Description**
This PR changes the `appName` from a `const` to a `var` so that it can be overwritten by consumers who embed Karpenter as a library. Currently, `appName` is hardcoded as `karpenter`, which means that downstream projects that import Karpenter's operator package can't customize the application name for their own use cases, affecting both the user agent and event recording. 

`make presubmit`


pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -168,6 +168,14 @@ func (v *VolumeTopology) ValidatePersistentVolumeClaims(ctx context.Context, pod
 		if pvc == nil {
 			continue
 		}
+		// Handle cases specifically rejected by kube-scheduler
+		// https://github.com/kubernetes/kubernetes/blob/56f6358c11b78e8e3d39e8cd8ff016ff7c70c56b/pkg/scheduler/framework/plugins/volumebinding/volume_binding.go#L333
+		if !pvc.DeletionTimestamp.IsZero() {
```
jonathan-innis: I think you need a rebase after pulling the latest changes

AndrewMitchell25: Ok I think I fixed it

pkg/operator/operator.go
```diff
@@ -63,10 +63,9 @@ import (
 	"sigs.k8s.io/karpenter/pkg/utils/env"
 )
 
-const (
-	appName   = "karpenter"
-	component = "controller"
-)
+const component = "controller"
```
jonathan-innis: nit: I wonder if we can drop the component constant entirely since we only use it in one spot

jonathan-innis (commented):
/lgtm
/approve

jonathan-innis:
/hold

Small comment, please take a look before /unhold

PR #2510, opened 2025-09-16
fix: drift and deprovisioning race condition

sumukha-radhakrishna:
fix: drift and deprovisioning race condition

More conversation in the PR here - https://github.com/kubernetes-sigs/karpenter/pull/2431

**Description**


pkg/controllers/state/statenodepool.go
```diff
@@ -26,8 +26,13 @@ import (
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 )
 
+// Currently NodeClaims be in one of these states
 type NodeClaimState struct {
-	Active   sets.Set[string]
+	// NodeClaims that have been launched
+	Active sets.Set[string]
+	// NodeClaims that are actively drifting
+	Drifting sets.Set[string]
```
jmdeal: For naming, I think it would be more accurate to call this something along the lines of "pending disruption". I have two main reasons for thinking this:
- Eventually this is going to expand to cover more than just drift
- I wouldn't consider a node to be finished "drifting" once we mark it for deletion. I would consider it to be drifting from the time we create the drift decision to the time the node is terminated.

I think pending disruption, or something along those lines, is a good way to frame it since the disruption process begins when we successfully create the replacement NodeClaims. Once that succeeds, we'd mark the NodeClaim as deleting (in-memory) and no longer consider it pending disruption.

pkg/controllers/disruption/staticdrift.go
```diff
@@ -99,6 +99,8 @@ func (d *StaticDrift) ComputeCommands(ctx context.Context, disruptionBudgetMappi
 				Replacements: replacementsFromNodeClaims(result.NewNodeClaims...),
 				Results:      result,
 			})
+			// Mark the Candidate as Drifting in statenodepool
+			d.cluster.NodePoolState.MarkNodeClaimDrifting(npName, c.NodeClaim.Name)
```
jmdeal: If we shift to the "pending disruption" semantic, we should probably mark this either right before we enqueue commands into the disruption queue, or make that the first thing we do when we start processing a command. I'd lean towards the latter since it keeps the mutation of cluster state in one place.

jmdeal: Discussed offline, we're going to change this to be marked in the orchestration queue. That change will probably be up by the time this comment is published.

pkg/controllers/static/provisioning/controller.go
```diff
@@ -73,10 +73,11 @@ func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
-	runningNodeClaims, _ := c.cluster.NodePoolState.GetNodeCount(np.Name)
+	runningNodeClaims, _, driftingNodeClaims := c.cluster.NodePoolState.GetNodeCount(np.Name)
 	desiredReplicas := lo.FromPtr(np.Spec.Replicas)
 	// Size down of replicas will be handled in deprovisioning controller to drain nodes and delete NodeClaims
-	if int64(runningNodeClaims) >= desiredReplicas {
+	// If there are drifting NodeClaims we need to count them as Active as drift controller is in the process of creating replacements
```
jmdeal:
```suggestion
	// If there are drifting NodeClaims we need to count them as Active as disruption controller is in the process of creating replacements
```

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -81,8 +81,10 @@ func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
-	runningNodeClaims, _ := c.cluster.NodePoolState.GetNodeCount(np.Name)
+	runningNodeClaims, _, _ := c.cluster.NodePoolState.GetNodeCount(np.Name)
 	desiredReplicas := lo.FromPtr(np.Spec.Replicas)
+	// We dont wanna run into deprovisioning and drift controller races so we
+	// Only count running NodeClaims for deprovisioning as drifting NodeClaims will have replacements
```
jmdeal: nit: I think it's a good idea to add a few more details on the race here, it took me a moment to think through the exact conditions which this could cause an issue. It would probably take a fair bit longer for someone unfamiliar.
```suggestion
	// We don't want to run into deprovisioning and disruption controller races so we
	// only count running NodeClaims for deprovisioning. If we counted both active and 
	// NodeClaims pending disruption here, we could burst above the desired replica count 
	// while the replacements are being created.
```

pkg/controllers/state/statenodepool.go
sumukha-radhakrishna: umm makes sense. Will mark it as `PendingDisruption`

pkg/controllers/disruption/queue.go
```diff
@@ -319,12 +319,19 @@ func (q *Queue) StartCommand(ctx context.Context, cmd *Command) error {
 	// Update the command to only consider the successfully MarkDisrupted candidates
 	cmd.Candidates = markedCandidates
 
+	// Before creating replacements we need to mark the nodeclaims as drifting
+	// We are doing this here since we are actively creating replacements
+	lo.ForEach(cmd.Candidates, func(c *Candidate, _ int) {
+		if c.OwnedByStaticNodePool() {
+			q.cluster.NodePoolState.MarkNodeClaimPendingDisruption(c.NodePool.Name, c.NodeClaim.Name)
+		}
+	})
```
jmdeal: nit: we pretty much never use `lo.ForEach`. If I'm being honest, I'm not sure why it exists since go has for each loops 🤷 
```suggestion
	for _, c := range cmd.Candidates {
		if c.OwnedByStaticNodePool() {
			q.cluster.NodePoolState.MarkNodeClaimPendingDisruption(c.NodePool.Name, c.NodeClaim.Name)
		}
	}
```

pkg/controllers/disruption/staticdrift.go
```diff
@@ -68,10 +68,10 @@ func (d *StaticDrift) ComputeCommands(ctx context.Context, disruptionBudgetMappi
 		limit, ok := np.Spec.Limits[resources.Node]
 		nodeLimit := lo.Ternary(ok, limit.Value(), int64(math.MaxInt64))
 		// Current nodes (includes in‑flight per your cluster state)
-		runningNodes, _ := d.cluster.NodePoolState.GetNodeCount(npName)
+		runningNodes, _, pendingdisruption := d.cluster.NodePoolState.GetNodeCount(npName)
```
jmdeal: nit: consistent naming
```suggestion
		nodesRunningCount, _, nodesPendingDisruptionCount := d.cluster.NodePoolState.GetNodeCount(npName)
```

PR #2516, opened 2025-09-17
fix: default CPU_REQUESTS when non-positive value is provided

jmdeal:
Fixes https://github.com/aws/karpenter-provider-aws/issues/8495

**Description**

This change eases configuration by allowing helm charts to specify this env var as a function of the container's resource requests. When a container doesn't specify resource requests, or specifies a value of 0, we'll fall back to the default (1000).

`make presubmit`


jonathan-innis (approved):
/lgtm
/approve

PR #2519, opened 2025-09-18
perf: add method for DeepCopyNodesWithReject

ryan-mist:
Fixes #N/A 

**Description**
* add method for cluster state to deep copy nodes with a rejection filter
   * decreases number of expensive DeepCopy operations

* `make presubmit`

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -232,25 +232,22 @@ func BuildDisruptionBudgetMapping(ctx context.Context, cluster *state.Cluster, c
 	disruptionBudgetMapping := map[string]int{}
 	numNodes := map[string]int{}   // map[nodepool] -> node count in nodepool
 	disrupting := map[string]int{} // map[nodepool] -> nodes undergoing disruption
-	for _, node := range cluster.DeepCopyNodes() {
-		// We only consider nodes that we own and are initialized towards the total.
-		// If a node is launched/registered, but not initialized, pods aren't scheduled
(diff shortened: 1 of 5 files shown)


jmdeal:
Rather than adding another custom iteration call, does it make more sense to build up the list this way:
```go
var nodes []*state.Node
for node := range cluster.Nodes() {
  if predicate(node) {
    nodes = append(nodes, node.DeepCopy())
  }
}
```

IMO it's a good idea to minimize the surface of each object and just compose on top of it.

pkg/controllers/disruption/helpers.go
```diff
@@ -232,25 +232,25 @@ func BuildDisruptionBudgetMapping(ctx context.Context, cluster *state.Cluster, c
...
+	// healthy before continuing.
+	// Additionally, don't consider nodeclaims that have the terminating condition. A nodeclaim should have
+	// the Terminating condition only when the node is drained and cloudprovider.Delete() was successful
+	// on the underlying cloud provider machine.
+	var nodes state.StateNodes
+	for n := range cluster.Nodes() {
+		if n.Managed() && n.Initialized() && !n.NodeClaim.StatusConditions().Get(v1.ConditionTypeInstanceTerminating).IsTrue() {
+			nodes = append(nodes, n.DeepCopy())
 		}
+	}
```
jmdeal: nit: I'd consider keeping the same structure we had previously, i.e. do a check, continue, do another check, continue, and only add the the copied nodes if all of the checks passed. This helps readability imo.

ryan-mist: Makes sense

PR #2520, opened 2025-09-18
chore: Update NodeRegistrationHealthy SC to use a buffer mechanism

jigisha620:
Fixes #N/A 

**Description**
This change is intended to allow `NodeRegistrationHealthy` SC on the nodePool to be updated to `False` even if it was previously set to `True` depending on the values stored in the boolean `NodePoolNodeRegistrationBuffer`. In this implementation, the size of this buffer is `4` and if there are >=2 registration failures, we update the `NodeRegistrationHealthy` SC to `False` even if it was `True` previously. This wasn't possible in the existing implementation because once the SC was set to `True`, it could only be set to `Unknown` on updates to nodePool/nodeClass. This change allows it to go back to `False`.

Deployed to my dev cluster and tested. Also added unit tests.

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -129,9 +129,9 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 		if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
 			return err
 		}
-		l.npState.Update(string(nodePool.UID), false)
+		l.npState.Update(nodePool.UID, false)
 		stored := nodePool.DeepCopy()
-		if l.npState.Status(string(nodePool.UID)) == nodepoolhealth.StatusUnhealthy && !nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsFalse() {
+		if l.npState.Status(nodePool.UID) == nodepoolhealth.StatusUnhealthy && !nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsFalse() {
 			// If the nodeClaim failed to register during the timeout set NodeRegistrationHealthy status condition on
 			// NodePool to False. If the launch failed get the launch failure reason and message from nodeClaim.
 			if launchCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched); launchCondition.IsTrue() {
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -32,11 +32,13 @@ import (
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/metrics"
+	"sigs.k8s.io/karpenter/pkg/state/nodepoolhealth"
 )
 
 type Liveness struct {
 	clock      clock.Clock
 	kubeClient client.Client
+	npState    nodepoolhealth.State
 }
 
 // registrationTimeout is a heuristic time that we expect the node to register within
@@ -81,6 +83,12 @@ func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reco
 			// This should never occur because if we failed to launch we requeue the object with error instead of this requeueAfter
 			return reconcile.Result{RequeueAfter: timeUntilTimeout}, nil
 		}
+		if err := l.updateNodePoolRegistrationHealth(ctx, nodeClaim); client.IgnoreNotFound(err) != nil {
+			if errors.IsConflict(err) {
+				return reconcile.Result{Requeue: true}, nil
+			}
+			return reconcile.Result{}, err
+		}
 		if err := l.deleteNodeClaimForTimeout(ctx, LaunchTimeout, nodeClaim); err != nil {
 			if client.IgnoreNotFound(err) != nil {
 				return reconcile.Result{}, err
@@ -121,8 +129,9 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 		if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
 			return err
 		}
-		if nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown() {
-			stored := nodePool.DeepCopy()
+		l.npState.Update(string(nodePool.UID), false)
+		stored := nodePool.DeepCopy()
+		if l.npState.Status(string(nodePool.UID)) == nodepoolhealth.StatusUnhealthy && !nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsFalse() {
 			// If the nodeClaim failed to register during the timeout set NodeRegistrationHealthy status condition on
 			// NodePool to False. If the launch failed get the launch failure reason and message from nodeClaim.
 			if launchCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched); launchCondition.IsTrue() {
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -131,28 +131,28 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 		if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
 			return err
 		}
-		_, found := lo.Find(nodeClaim.GetOwnerReferences(), func(o metav1.OwnerReference) bool {
+		if _, found := lo.Find(nodeClaim.GetOwnerReferences(), func(o metav1.OwnerReference) bool {
 			return o.UID == nodePool.UID
-		})
-		if found {
-			stored := nodePool.DeepCopy()
-			if l.npState.DryRun(nodePool.UID, false).Status() == nodepoolhealth.StatusUnhealthy && !nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsFalse() {
-				// If the nodeClaim failed to register during the timeout set NodeRegistrationHealthy status condition on
-				// NodePool to False. If the launch failed get the launch failure reason and message from nodeClaim.
-				if launchCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched); launchCondition.IsTrue() {
-					nodePool.StatusConditions().SetFalse(v1.ConditionTypeNodeRegistrationHealthy, "RegistrationFailed", "Failed to register node")
-				} else {
-					nodePool.StatusConditions().SetFalse(v1.ConditionTypeNodeRegistrationHealthy, launchCondition.Reason, launchCondition.Message)
-				}
-				// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
-				// can cause races due to the fact that it fully replaces the list on a change
-				// Here, we are updating the status condition list
-				if err := l.kubeClient.Status().Patch(ctx, nodePool, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); client.IgnoreNotFound(err) != nil {
-					return err
-				}
-			}
-			l.npState.Update(nodePool.UID, false)
+		}); !found {
+			return nil
 		}
+		stored := nodePool.DeepCopy()
+		if l.npState.DryRun(nodePool.UID, false).Status() == nodepoolhealth.StatusUnhealthy && !nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsFalse() {
+			// If the nodeClaim failed to register during the timeout set NodeRegistrationHealthy status condition on
+			// NodePool to False. If the launch failed get the launch failure reason and message from nodeClaim.
+			if launchCondition := nodeClaim.StatusConditions().Get(v1.ConditionTypeLaunched); launchCondition.IsTrue() {
+				nodePool.StatusConditions().SetFalse(v1.ConditionTypeNodeRegistrationHealthy, "RegistrationFailed", "Failed to register node")
+			} else {
+				nodePool.StatusConditions().SetFalse(v1.ConditionTypeNodeRegistrationHealthy, launchCondition.Reason, launchCondition.Message)
+			}
+			// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
+			// can cause races due to the fact that it fully replaces the list on a change
+			// Here, we are updating the status condition list
+			if err := l.kubeClient.Status().Patch(ctx, nodePool, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); client.IgnoreNotFound(err) != nil {
+				return err
+			}
+		}
+		l.npState.Update(nodePool.UID, false)
 	}
 	return nil
 }
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/awslabs/operatorpkg/object"
 	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/api/errors"
 
@@ -132,7 +133,7 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 			return err
 		}
 		if _, found := lo.Find(nodeClaim.GetOwnerReferences(), func(o metav1.OwnerReference) bool {
-			return o.UID == nodePool.UID
+			return o.Kind == object.GVK(nodePool).Kind && o.UID == nodePool.UID
 		}); !found {
 			return nil
 		}
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"time"
 
+	"github.com/samber/lo"
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	"k8s.io/apimachinery/pkg/types"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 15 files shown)


pkg/utils/nodepool/registrationhealth/buffer.go
```diff
@@ -0,0 +1,84 @@
...
+func New() *NodeRegistrationBuffer {
+	return &NodeRegistrationBuffer{
+		values:       make([]bool, 0, BufferSize),
+		currentIndex: 0,
+	}
+}
+
+func (b *NodeRegistrationBuffer) AddHealth(value bool) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
```
ryan-mist: Would it be cleaner to initialize values as `make([]bool, BufferSize, BufferSize)` and `currentIndex` as -1.

Then this function could just be:
```
func (b *NodeRegistrationBuffer) AddHealth(value bool) {
     b.mu.Lock()
	 defer b.mu.Unlock()
     b.currentIndex = (b.currentIndex + 1) % BufferSize
     b.values[b.currentIndex] = value
}
```
and the check for unknown in IsHealthy() could be if b.currentIndex == -1

pkg/utils/nodepool/registrationhealth/buffer.go
```diff
@@ -0,0 +1,84 @@
...
+	if len(b.values) < BufferSize {
+		b.values = append(b.values, value)
+	} else {
+		// If buffer is full, replace the oldest entry
+		b.values[b.currentIndex] = value
+		b.currentIndex = (b.currentIndex + 1) % BufferSize
+	}
+}
+
+func (b *NodeRegistrationBuffer) IsHealthy() int {
```
ryan-mist: nit: I don't know if this is more common but it seems like we usually have boolean return values and 3 functions of `IsUnhealthy`, `IsHealthy` and `IsUnknown`. -1, 0, 1 makes sense but it becomes a little hard to read.

pkg/controllers/state/cluster.go
```diff
@@ -58,12 +59,13 @@ type Cluster struct {
...
-	nodeNameToProviderID      map[string]string               // node name -> provider id
-	nodeClaimNameToProviderID map[string]string               // node claim name -> provider id
-	nodePoolResources         map[string]corev1.ResourceList  // node pool name -> resource list
-	daemonSetPods             sync.Map                        // daemonSet -> existing pod
+	nodes                     map[string]*StateNode                                 // provider id -> cached node
+	bindings                  map[types.NamespacedName]string                       // pod namespaced named -> node name
+	nodeNameToProviderID      map[string]string                                     // node name -> provider id
+	nodeClaimNameToProviderID map[string]string                                     // node claim name -> provider id
+	nodePoolResources         map[string]corev1.ResourceList                        // node pool name -> resource list
+	nodePoolToAZlaunched      map[string]*registrationhealth.NodeRegistrationBuffer // nodePool name -> [availabilityZone, buffer]
```
ryan-mist: Is the comment / naming off? It seems that its more of a map of node pool to registration buffer (i.e. each node pool has its own buffer)

jigisha620: Oh no, forgot to update this. Will make a change.

pkg/utils/nodepool/registrationhealth/buffer.go
jigisha620: I think the thing with doing `make([]bool, BufferSize, BufferSize)` is that it will initialize all the values as `False`.  And since I want to be able to set it explicitly depending on a registration success or failure, I went for an empty buffer instead.

pkg/utils/ringbuffer/buffer.go
```diff
@@ -0,0 +1,56 @@
...
+func (b *Buffer[T]) Reset() {
+	b.values = b.values[:0]
+	b.currentIndex = 0
+}
+
+func (b *Buffer[T]) GetItems() []T {
+	return b.values
+}
+
+func (b *Buffer[T]) Capacity() int {
```
jmdeal: The capacity is immutable, is there a reason to add this to the public interface?

pkg/utils/ringbuffer/buffer.go
```diff
@@ -0,0 +1,56 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ringbuffer
+
+type Buffer[T any] struct {
+	values       []T
+	currentIndex int
```
jmdeal: nit: since there's no other index than the `currentIndex`, should we just name it something like `index` or `head`? I have a slight preference for `head` since queues built on circular buffers typically have a `head` and `tail` pointer.

pkg/utils/ringbuffer/buffer.go
```diff
@@ -0,0 +1,56 @@
...
+func (b *Buffer[T]) Len() int {
+	return len(b.values)
+}
+
+func (b *Buffer[T]) Reset() {
+	b.values = b.values[:0]
+	b.currentIndex = 0
+}
+
+func (b *Buffer[T]) GetItems() []T {
```
jmdeal: nit: typically we don't include `Get` in method names
```suggestion
func (b *Buffer[T]) Items() []T {
```

pkg/utils/ringbuffer/buffer.go
```diff
@@ -0,0 +1,56 @@
...
+*/
+
+package ringbuffer
+
+type Buffer[T any] struct {
+	values       []T
+	currentIndex int
+}
+
+func NewBuffer[T any](capacity int) *Buffer[T] {
```
jmdeal: nit: since the Buffer is the only exported type in this package (and it should stay that way), we can just name this `New`
```suggestion
func New[T any](capacity int) *Buffer[T] {
```

pkg/utils/ringbuffer/buffer.go
```diff
@@ -0,0 +1,56 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ringbuffer
+
+type Buffer[T any] struct {
```
jmdeal: nit: I'd consider calling this `RingBuffer` to align with the package name
```suggestion
type RingBuffer[T any] struct {
```

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,122 @@
...
+func (t *Tracker) Reset() {
+	t.Lock()
+	defer t.Unlock()
+
+	t.buffer.Reset()
+}
+
+func (t *Tracker) Status() Status {
+	t.Lock()
+	defer t.Unlock()
```
jmdeal: We shouldn't need a write lock here, right?

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (t *Tracker) SetStatus(status Status) {
+	switch status {
+	case StatusUnknown:
+		t.buffer.Reset()
+	case StatusHealthy:
+		t.Update(true)
+	case StatusUnhealthy:
+		t.Update(false)
```
jmdeal: Since we have `ThresholdFalse` as a constant, I'd rely on that to figure out how many items we need to insert into the queue.

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (t *Tracker) Status() Status {
+	t.Lock()
+	defer t.Unlock()
+
+	if t.buffer.Len() == 0 {
+		return StatusUnknown
+	}
+	// Count number of true and false
+	unhealthyCount := 0
```
jmdeal: tiny nit: typically prefer this style for zero-value initialization
```suggestion
	var unhealthyCount int
```

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (s State) Update(nodePool string, launchStatus bool) {
+	s.nodePoolNodeRegistration(nodePool).Update(launchStatus)
+}
+
+func (s State) SetStatus(nodePool string, status Status) {
+	s.nodePoolNodeRegistration(nodePool).SetStatus(status)
+}
+
+func (s State) ResetStatus(nodePool string) {
```
jmdeal: nit: do we need a `ResetStatus` method when this should be the equivalent of calling `SetStatus(StatusUnknown)`?

pkg/state/nodepoolhealth/suite_test.go
```diff
@@ -0,0 +1,76 @@
...
+})
+
+func TestAPIs(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "NodePoolHealthState")
+}
+
+var _ = Describe("NodePoolHealthState", func() {
+	It("should expect status unknown for a new nodePool with empty buffer", func() {
+		Expect(npState.Status(npUUID)).To(BeEquivalentTo(nodepoolhealth.StatusUnknown))
```
jmdeal: Why do we need to use `BeEquivalentTo` here (as opposed to `Equal`)? Shouldn't this be the same concrete type?

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -121,8 +129,9 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 		if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
```
jmdeal, line 132: Not from this PR, but can we not get the NodePool's UID from the owner reference on the NodeClaim? This would ensure that if a NodePool is created, deleted, and recreated we're actually tracking the NodePool that this NodeClaim is bound to, not the replacement.

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,122 @@
...
+	case StatusUnhealthy:
+		t.Update(false)
+		t.Update(false)
+	}
+}
+
+type State map[string]*Tracker
+
+func (s State) nodePoolNodeRegistration(nodepool string) *Tracker {
+	tracker, exists := s[nodepool]
```
jmdeal: We should probably have some way to garbage collect the entries in this tracker, but I'm not sure it's worth the added complexity given the small size. The right way to do it would probably be to block any calls to the tracker until we've created entries for all existing NodePools and fail closed if we make any calls for non-existent NodePools. That being said, I think we need a better story in general for how we ensure that shared state is hydrated before beginning controller reconciliation - I don't think we need to tackle it in this PR.

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,122 @@
...
+	if !exists {
+		tracker = NewTracker(BufferSize)
+		s[nodepool] = tracker
+	}
+	return tracker
+}
+
+func (s State) Status(nodePool string) Status {
```
jmdeal: Since we're always using UIDs for these methods, let's enforce it with the type system. Otherwise it's not clear as a caller if you should be providing a UID or the name. I'd honestly assume it's the name.
```suggestion
func (s State) Status(nodePoolUID types.UID) Status {
```

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -121,8 +129,9 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
...
 			return err
 		}
-		if nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown() {
-			stored := nodePool.DeepCopy()
+		l.npState.Update(string(nodePool.UID), false)
```
jmdeal: Shouldn't we update this after we persist the status condition so we don't double count it when there are errors?

jigisha620: Good point. The only concern is that we update the buffer for the nodePool and then validate if the buffer state is unhealthy to update the status condition. So, to handle the error case, we will have to remove the entry that we just added.

pkg/utils/ringbuffer/buffer.go
jigisha620: Can drop it.

pkg/controllers/nodeclaim/lifecycle/liveness.go
jigisha620: If the nodePool is deleted and recreated, does it even make sense to track the registration for the old nodePool? We do registration tracking so that we can update the status condition on the nodePool. If a nodePool with UID `abc` is deleted then we are not going to be able to update it's status condition anyway?

pkg/state/nodepoolhealth/tracker.go
jigisha620: I agree. I did not spend much time thinking how we would tackle garbage collecting the entries in this tracker.

pkg/controllers/nodeclaim/lifecycle/liveness.go
jmdeal: The issue right now is that since we're doing a lookup by the NodePool's name, if `nodepool-foo` (UID 1) is deleted and `nodepool-foo` (UID 2) is created, we'll get the second one here rather than the first even if the NodeClaim was created for the first.

pkg/controllers/nodeclaim/lifecycle/liveness.go
jmdeal: Yeah, you're right. Maybe we add a "dry run" option to the tracker? For a dry run we just create a copy of the tracker, make the update, return the result. We then persist it after. If the tracker were any larger I don't think I'd want to make a copy, but it's a small struct and we're not in a performance critical path.

pkg/controllers/nodeclaim/lifecycle/liveness.go
jigisha620: Discussed offline and updated.

pkg/controllers/nodeclaim/lifecycle/liveness.go
jigisha620: Added a dry run method

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -129,22 +131,27 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
...
-				nodePool.StatusConditions().SetFalse(v1.ConditionTypeNodeRegistrationHealthy, launchCondition.Reason, launchCondition.Message)
-			}
-			// We use client.MergeFromWithOptimisticLock because patching a list with a JSON merge patch
-			// can cause races due to the fact that it fully replaces the list on a change
-			// Here, we are updating the status condition list
-			if err := l.kubeClient.Status().Patch(ctx, nodePool, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); client.IgnoreNotFound(err) != nil {
-				return err
+		_, found := lo.Find(nodeClaim.GetOwnerReferences(), func(o metav1.OwnerReference) bool {
+			return o.UID == nodePool.UID
+		})
```
jmdeal: nit: to improve clarity by reducing nesting, I'd return early if it's not found

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -116,3 +116,10 @@ func (s State) Update(nodePoolUID types.UID, launchStatus bool) {
 func (s State) SetStatus(nodePoolUID types.UID, status Status) {
 	s.nodePoolNodeRegistration(nodePoolUID).SetStatus(status)
 }
+
+func (s State) DryRun(nodePoolUID types.UID, launchStatus bool) *Tracker {
+	trackerCopy := NewTracker(BufferSize)
+	trackerCopy.buffer = s.nodePoolNodeRegistration(nodePoolUID).buffer
```
jmdeal: We shouldn't perform a shallow-copy of the buffer here, it should be a deep copy. Right now we're still mutating the same underlying memory. The test passes because the length isn't being mutated by the original buffer, but this could result in a race condition if we attempt to do a dry-run mutation and real mutation in parallel.

pkg/state/nodepoolhealth/suite_test.go
```diff
@@ -74,4 +74,8 @@ var _ = Describe("NodePoolHealthState", func() {
 		npState.SetStatus(npUUID2, nodepoolhealth.StatusUnknown)
 		Expect(npState.Status(npUUID2)).To(Equal(nodepoolhealth.StatusUnknown))
 	})
+	It("should create a copy of the tracker and update it and not the original tracker in case of Dry Run", func() {
```
jmdeal, line 77: We should update this test to not perform a dry-run against an empty buffer. This test would pass if we were constructing an empty buffer in the dry run call rather than copying the existing buffer.

pkg/state/nodepoolhealth/tracker.go
jigisha620: My bad, that was the whole point of doing a dry-run. Fixed it.

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	// Determine health status based on threshold
+	if (float64(unhealthyCount) / float64(BufferSize)) >= ThresholdFalse {
+		return StatusUnhealthy
+	} else {
+		return StatusHealthy
+	}
+}
+
+func (t *Tracker) SetStatus(status Status) {
```
jmdeal, line 84: This implementation only works on an empty buffer, right? We need to ensure we're always resetting the buffer first. Let's also make sure we add a unit test to catch this.

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -121,8 +131,13 @@ func (l *Liveness) updateNodePoolRegistrationHealth(ctx context.Context, nodeCla
 		if err := l.kubeClient.Get(ctx, types.NamespacedName{Name: nodePoolName}, nodePool); err != nil {
 			return err
 		}
-		if nodePool.StatusConditions().Get(v1.ConditionTypeNodeRegistrationHealthy).IsUnknown() {
-			stored := nodePool.DeepCopy()
+		if _, found := lo.Find(nodeClaim.GetOwnerReferences(), func(o metav1.OwnerReference) bool {
+			return o.UID == nodePool.UID
```
jmdeal: We should also check that the owner reference is a NodePool. I don't believe UIDs are guaranteed to be unique across different kinds. Although we don't add any other kind of owner reference, I can imagine users might build different ownership constructs on top of NodeClaims for their platforms.

pkg/state/nodepoolhealth/tracker.go
jigisha620: Makes sense, could reset the buffer outside the switch but then we wouldn't need the `StatusUnknown` case, but keeping it and doing a reset inside each case seems more readable.

Issue #732, opened 2023-01-12
Limitation of node number per provisioner


sergkondr:
### Tell us about your request

It would be nice to have the ability to limit the number of nodes created by certain provisioner. For example:
```
  limits:
    nodes: "6"
```

### Tell us about the problem you're trying to solve. What are you trying to do, and why is it hard?

Let's say our application spawns pods for some tasks, and each pod requires a separate node. Now it limits by the max size of the node group. All these nodes are Spot instances, and there is a problem that sometimes there are no instances of the current family in the region, so we use different instance families: m4, m5, m5a, r5, r5a, etc. These instances could have a different amount of CPU and mem.

It would be nice to have the ability to limit the number of nodes by their count, not by their resources. It is clear that we have max pods in our app, but it is a bit of a synthetic example.

### Are you currently working around this issue?

We use `limits.resources.cpu` with approx number of CPUs, but it is not accurate and not transparent a little bit.

### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

yr8sdk:
Max nodes could benefit the total cluster resource utilization. Ideally, I would like to schedule most of my pods on the least number of nodes ( depends of HA of course) to have simple control over the pods memory limit and since there isn't CPU limit already, more pods could use the unutilized CPU.

cest-pas-faux:
https://github.com/kubernetes-sigs/karpenter/issues/745

cest-pas-faux:
Better explanation here : https://github.com/kubernetes-sigs/karpenter/issues/745

jukie:
I've started on this with #1151 but still need to test

jukie:
RFC here: https://github.com/kubernetes-sigs/karpenter/pull/1160

benben:
similar use case here: limited set of preallocated IPs and I want to make sure we never provision more instances then we have IPs. These IPs are static in the sense that they are communicated to customer upfront so they can allow our service in their firewalls.

I'd expect karpenter to honor this limit and if more pods are coming, it would consolidate to larger instances to keep the number at the limit.

elmiko:
[...]

Issue #2385, opened 2025-07-21
Max nodeClaims cap


enxebre:
### Description

**What problem are you trying to solve?**
As a Service Provider I want to enforce a maximum number of Nodes in the cluster so I can guarantee my SLAs.
This includes nodes managed by karpenter. So I would like to have the ability for karpenter to short circuit when a given number of nodeClaims have been met.
Some not necessarily exclusive options to express intent for this might be:
- Via flag
- Global API knob
- NodePool API knob

**How important is this feature to you?**

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

enxebre:
cc @jkyros cc @elmiko cc @jonathan-innis

elmiko:
thanks @enxebre !

this is along the same lines as we were discussing at the working group meeting last week. from the meeting, it sounded like the use cases we talked about (limited network IP space, control plane density concerns) are well received and we should have some method for limiting karpenter.

due to the way this would work internally, it /may/ be complex to introduce but there was also the thought that "node" might be able to be used as a resource type in the node pool resource limits. this needs further investigation.

there is some further followup investigation to do and report back to the working group (and here).

elmiko:
this issue is also related to https://github.com/kubernetes-sigs/karpenter/issues/732

rschalo:
Hi all - thanks for opening the issue. Not sure if it needs an RFC but could be easier to discuss changes to limits. It seems like a more natural place to discuss this than a PR.

Issue #1747, opened 2024-10-11
Add NodePoolGroupLimit CRD for limits that span NodePools


JacobHenner:
### Description

**What problem are you trying to solve?**

Today, limits can only be specified on individual NodePools. While this is fine for simple situations, it is insufficient when multiple NodePools comprise a logical grouping of compute that should share a limit. This happens most often when there are important variations in a NodePool's properties beyond its requirements that mandate the use of multiple NodePools, but when they are otherwise related in some way relevant to limits (e.g. same department, team, application, budget line-item).

For example, an organization might group limits by team. A team might require nodes labelled in two distinct ways, necessitating the use of two NodePools. Splitting the team's limit in half for each NodePool might not be sufficient if the balance of nodes between the NodePools varies over time.

I propose a `NodePoolGroupLimit` CRD (or a similar appropriate name) that would allow a defined limit to apply to NodePools chosen by a selector. If multiple `NodePoolGroupLimit` objects select the same NodePool, the most prohibitive limit should take precedence.

It might look something like this:

```yaml
apiVersion: karpenter.sh/v1
kind: NodePoolGroupLimit
metadata:
  name: frontend-bingo
  labels:
    team: frontend
    service: bingo
spec:
  selector:
    # label selector for NodePool labels
    team: frontend
    service: bingo
  limits:
    cpu: "100"
    memory: 128Gi
```

**How important is this feature to you?**

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

njtran:
Based on your request, it seems like it's common for teams to have multiple NodePools, where there's overarching org-wide constraints across the cluster. It seems like you also don't want a necessarily global limit across the cluster too, you want something in between. 

Are you willing to open up an RFC to talk about your proposed solution and alternatives to the solution that you've explored?

JacobHenner:
> Based on your request, it seems like it's common for teams to have multiple NodePools, where there's overarching org-wide constraints across the cluster.

[...]

Issue #1177, opened 2024-04-09
Ability to Scale Karpenter Provisioned Nodes To 0 On Demand Or By Schedule During Off Hours


ronberna:
### Description

**What problem are you trying to solve?**
We've recently begun the migration from using ASG's (AutoScaling Groups) and  CAS (Cluster AutoScaler) to Karpenter.  With ASG's, as part of cost saving measures, our EKS clusters are scaled down during off hours and weekends in lower environments, and then scaled back up during office hours.  This was performed by running a lambda at a scheduled time to set the min/max/desired settings of the ASG to 0.  The current values of the min/max/desired settings before the update to 0 are captured and stored in ssm. For the scale up, the lambda reads this ssm parameter to set the ASG min/max/desired values. With Karpenter, this is not possible.

As a workaround, we have a lambda that will patch the cpu limit of the nodepool and set it to 0 so that no new Karpenter nodes will be provisioned.  The lambda will then take care of deleting the previously provisioned Karpenter nodes.  We have a mix of workloads running in the cluster with some using HPA and some not, so trying to scale down all of the deployments to remove the Karpenter provisioned nodes will not work.  It has also been suggested to delete the nodepool and reapply it via a cronjob.  This option will also not work since some of our clusters are in a controlled environment.

The ask here is to introduce a feature in Karpenter that will handle scaling down/up all Karpenter provisioned nodes on-demand via a flag or possibly with the update of the cpu limit, Karpenter will not provision any new nodes and will also clean up previously provisioned nodes without having to introduce additional cronjobs, lambdas, or deleting nodepools.

**How important is this feature to you?**
This feature is important as it will help with AWS cost savings by not having EC2 instances running during off hours and not having to add additional components (lambdas, cronjobs, etc...) to aid with scaling Karpenter provisioned instances.

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

jonathan-innis:
> Karpenter will not provision any new nodes and will also clean up previously provisioned nodes without having to introduce additional cronjobs, lambdas, or deleting nodepools

[...]

PR #2525, opened 2025-09-20
docs: RFC for node limits per NodePool

maxcao13:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/732
Related: https://github.com/kubernetes-sigs/karpenter/issues/2385

**Description**
RFC for formally supporting nodes limits per `NodePool`.

### DIFF
--- a/designs/nodepool-node-limits.md
+++ b/designs/nodepool-node-limits.md
@@ -1,0 +1,161 @@
+# NodePool Node Limits in Karpenter
+
+## Motivation
+
+Service providers and cluster administrators should have the ability to enforce maximum node counts per `NodePool` to address several critical use cases:
+
+* **Licensing constraints**: Respect software licensing limits (e.g., monitoring agents with per-node licensing)
+* **IP address management**: With fixed IP address pools, physically prevent node creation beyond the available IP capacity
+* **Service Provider SLAs**: Guarantee service level agreements by preventing unlimited node provisioning
+* **Resource limits**: Respect cloud provider account limits or organizational quotas with Karpenter-level control rather than relying on infrastructure-level restrictions
+
+## Notes
+
+- Enforcing global maximum Node limits is not a goal of this design.
+- This design does not involve nodes managed outside of Karpenter.
+- The Node limits discussed in this design are slightly different from the Node limits discussed in the [Karpenter - Static Capacity](static-capacity.md) design.
+    - See the [Relationship to Static Capacity Design](#relationship-to-static-capacity-design) section for more details.
+
+## Design Options
+
+There are three potential approaches for implementing `NodePool` node limits, each with different trade-offs:
+
+### Option 1: Hard Limits
+
+**Description**: Node limits are enforced as hard constraints that block both provisioning and disruption when the limit is reached. Disruption is blocked because Karpenter requires the new nodes to be schedulable and ready before the old node can be terminated, which can lead to more nodes than the limit at one state.
+
+**Benefits**:
+- Guarantees the invariant that node count never exceeds the limit
+- Consistent behavior across all operations
+
+**Drawbacks**:
+- Can block disruption operations when at the limit, which may lead to suboptimal resource utilization
+
+### Option 2: Soft Limits
+
+**Description**: Node limits are enforced only during provisioning, allowing disruption operations to continue even when at the limit.
+
+**Benefits**:
+- Disruption functions continue to work, allowing consolidation and drift even at the limit.
+- In steady state, nodes will stay under the limit (assuming the cloud provider allows the node limit to be exceeded)
+
+**Drawbacks**:
+- Intermittent states may temporarily exceed the node limit during disruption operations
+- Potential undefined behavior if Karpenter allows more nodes at once, but the cloud provider blocks disruption at the infrastructure level (e.g., due to exhaustion of IP addresses)
+    - This may lead to node provisioning timeout failures, and the user running into undesired node churn if  disruption occurs.
+
+### Option 3: Configurable Hard/Soft Node Limits
+
+**Description**: Allow users to configure whether limits are enforced as hard or soft constraints per NodePool. This can be configured via a new field in the NodePool spec `nodepool.spec.disruption.nodeLimitMode`.
+The values can be `Hard` or `Soft`, defaulting to `Hard`.
+
+**Benefits**:
+- Flexibility to choose the appropriate enforcement model per use case
+- Can use hard limits for critical constraints (IP addresses) and soft limits for cost optimization
+
+**Drawbacks**:
+- Additional configuration complexity for users
+- More complex implementation
+
+<!-- ### Recommended Approach: Hard Limits
+
+Based on the critical use cases (e.g., IP address management), we recommend **Option 1: Hard Limits** as the initial implementation.
+
+TODO(maxcao13): Want to hear from the community.
+-->
+
+## Relationship to Static Capacity Design
+
+This design builds upon the infrastructure established in the [Karpenter - Static Capacity](static-capacity.md) design, which already formally introduced node limits, but specifically for static capacity:
+
+- `nodepool.spec.limits.nodes`: `NodePool` limits for constraining maximum node count
+- `nodepool.status.nodes`: Current node count tracking per NodePool
+- Validation rules for handling node limits in NodePool specs
+
+This design utilizes those existing fields for non-static `NodePool`s where node count can be limited by a hard or soft limit.
+
+## Implementation Details
+
+### NodePool Limits Integration
+
+NodePool `nodes` limits have already been softly defined in the `NodePool` spec as a supported resource type, as per the [Static Capacity design](static-capacity.md). The current implementation leverages this existing structure to enforce node limits for dynamic `NodePool`s.
+
+```go
+// In pkg/apis/v1/nodepool.go
+type NodePoolSpec struct {
+    // ... other fields
+    Limits Limits `json:"limits,omitempty"`
+}
+
+type Limits v1.ResourceList
+```
+
+```yaml
+spec:
+  limits:
+    # Prevents Karpenter from creating new nodes once the limit is exceeded
+    nodes: 10
+```
+
+### Provisioning
+- Node limits are enforced in the simulated scheduling logic.
+- Before creating a new `NodeClaim`, check if the `NodePool` has remaining node capacity.
+    - This requires the `StateNode.Capacity()` implementation to include a `node` resource field of `1`, indicating that a single node is being requested.
+    - This will allow the scheduler to correctly subtract the node capacity when updating the remaining resources for the `NodePool`.
+- When node limits have been exhausted, new `NodeClaim` creation is blocked.
+
+### Disruption
+
+An easy way to implement the node limit functionality into the disruption logic is by integrating it into the disruption budget calculation:
+
+```go
+// In pkg/controllers/disruption/helpers.go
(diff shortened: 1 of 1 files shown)


designs/nodepool-node-limits.md
```diff
@@ -0,0 +1,143 @@
...
+**Benefits**:
+- Guarantees the invariant that node count never exceeds the limit
+- Consistent behavior across all operations
+
+**Drawbacks**:
+- Can block disruption operations when at the limit, which may lead to suboptimal resource utilization
+
+### Option 2: Soft Limits
+
+**Description**: Node limits are enforced only during provisioning, allowing disruption operations to continue even when at the limit.
```
ellistarn: We've discussed a strategy in the past like "launchBeforeTerminate" or "terminateBeforeLaunch", which could interact with this proposal.

maxcao13: Thanks for the response. Is there an issue or thread for this, or was it an internal discussion?

Ideally, if that feature was implemented, we probably wouldn't need hard/soft limits and simply change the strategy depending if we are at the limit or not? But that probably requires a bit more detail on that part as well.

designs/nodepool-node-limits.md
```diff
@@ -0,0 +1,143 @@
...
+- The Node limits discussed in this design are different from the Node limits discussed in the [Karpenter - Static Capacity](static-capacity.md) design, but involve similar concepts.
+    - See the [Relationship to Static Capacity Design](#relationship-to-static-capacity-design) section for more details.
+
+## Design Options
+
+There are three potential approaches for implementing `NodePool` node limits, each with different trade-offs:
+
+### Option 1: Hard Limits
+
+**Description**: Node limits are enforced as hard constraints that block both provisioning and disruption when the limit is reached. Disruption is blocked because Karpenter requires the new nodes to be schedulable and ready before the old node can be terminated, which can lead to more nodes than the limit at one state.
```
GnatorX: I think its worth calling out that not all disruption requires new node replacement. For example, empty or simply consolidating into existing nodes. So the final implementation may be turning off paths that may cause a new node launch while letting some disruption to still occur.

In addition, Expiration is "Forceful" where it doesn't wait for a replacement to be launch before terminating a node. 

Is locking both provision and disruption ideal? I am wondering if there is something else that Karpenter does instead

maxcao13: Sorry you are right, I meant to write disruption that requires new node replacement (drift + consolidation, not expiration). I can clarify that, thanks.

designs/nodepool-node-limits.md
ellistarn: Just a coffee chat, I think. Hard vs soft is more generally applicable to limits, not just node limits -- I could see us pushing that aspect of the problem into the future and starting simple with just the node key. 

The reason we haven't supported this in the past is because of the interaction with the scheduling algorithm. For a trivial example, imagine a node limit of 1, and 33 pods with a cpu request of 1. You could achieve this with a 32 core machine and a 1 core machine. You can imagine how this gets more complicated with more complicated scenarios. This is a variant of the knapsack problem.

With our limit, we would be forced into a single 64 core machine, which comes with significant waste. Further, our scheduler would need to "look-ahead" in the binpacking process, and start to factor in limits across multiple placement decisions. The complexity of this explodes pretty quick. In practice, our current implementation is relatively naïve and might not be impacted by this today, but adding this will make it more challenging to implement better heuristics.

Not trying to say we won't do this, just highlighting something that's not covered in the doc. 

---

I am curious what is most motivating for you for this RFC. Is it solely IP Address management or 3p licensing? Or the others?

maxcao13: That makes a lot of sense. I can totally see the problem here, where adding limits on the number of knapsacks/bins will add another layer of complexity if eventually we are to improve the scheduling algorithm.

I think you are right, as Karpenter is today, implementing this will not affect the complexity (simply just block any scheduling simulations that create more nodes than the limit), but it could force suboptimal resource usage as to your example (depending on the size of the instance types). Although, I would assume the cluster admin would have to live with that usage if they are enforcing strict node limits.

> I am curious what is most motivating for you for this RFC. Is it solely IP Address management or 3p licensing? Or the others?

Our particular use case is that we provide clusters as a service (like eks) and we want to integrate a managed Karpenter offering. However we cannot guarantee that our clusters will function well over a certain amount of nodes as per a service level agreement. Pretty much what is detailed here: https://github.com/kubernetes-sigs/karpenter/issues/2385.

I am aware that what I am proposing is per NodePool and not global, but given the other issue/RFC that tried to solve this in the past, I think the best path for us would to implement this particular RFC, and then support something like https://github.com/kubernetes-sigs/karpenter/issues/1747 which could serve as a global limit.

designs/nodepool-node-limits.md
```diff
@@ -0,0 +1,144 @@
...
+    // ... other fields
+    Limits Limits `json:"limits,omitempty"`
+}
+
+type Limits v1.ResourceList
+```
+
+```yaml
+spec:
+  limits:
```
GnatorX, line 105: I am not sure we should rely on the existing limit especially if users already use this for soft node constraint. I think other "limits" within here are also soft so I don't want nodes to behave differently, nodes is hard and others are soft. I am wondering if we should have a different spec variable instead where we can introduce all hard limits.

Alternatively, we can introduce a new variable that specify if limits is hard or soft however I am unsure if there are use cases were user would want a mix of both.

Personally I am interested in introducing a hard limit variable for example
```yaml
spec:
   limits:
      pods: 10
   hardLimits:
      nodes: 10
```

maxcao13: Makes sense to me, although if Karpenter plans in the future to support something like "terminateBeforeLaunch", I'm wondering if it would still be useful to have separate fields for this.

I would imagine if we did later support a mechanism like that, that we would just switch to terminateBeforeLaunch while we are at the limit. But then I'm wondering if we would just change the behaviour of the regular `spec.limits` to be "hard" and then deprecate `spec.hardLimits` entirely.

I think how we have it right now, where we have a separate field `spec.hardLimits` totally makes sense (and only supporting nodes for this particular RFC) since the current limits are technically soft and wouldn't have the same behaviour across resource types. But just wondering about the future if we ever decide to do terminateBeforeLaunch, because that would solve part of the issues here in this RFC with Drifting and Consolidating at the hard limit, and may even remove the need to have separate limits. But I would need some input from maintainers about that.

jmdeal: > although if Karpenter plans in the future to support something like "terminateBeforeLaunch", I'm wondering if it would still be useful to have separate fields for this.

I don't think having "terminateBeforeLaunch" would remove the need for a mix of soft and hard limits. That feature would come with tradeoffs, the main one being that there is no guarantee that you could get optimal capacity after the initial termination (e.g. due to InsufficientCapacityErrors). This makes it a poor choice for consolidation IMO. Additionally, the evicted pods will be stuck pending for a longer period of time. While this shouldn't cause availability issues with properly configured PDBs, it's still not desirable.

Soft limits are nice because they give you the ability to opt into hard limits without relying on "terminateBeforeLaunch" and still get voluntary disruption.

designs/nodepool-node-limits.md
```diff
@@ -0,0 +1,144 @@
...
+### Option 2: Soft Limits
+
+**Description**: Node limits are enforced only during provisioning, allowing graceful disruption operations to continue even when at the limit.
+
+**Benefits**:
+- Graceful disruption functions continue to work, allowing Consolidation and Drift even at the limit.
+- In steady state, nodes will stay under the limit (assuming the cloud provider allows the node limit to be exceeded)
+
+**Drawbacks**:
+- Intermittent states may temporarily exceed the node limit during disruption operations
```
jmdeal: I think this drawback makes this a non-option for the goals listed in the motivation section. It's worth having the proposal since this is the most straightforward extension of Karpenter's existing limits mechanism, but I'd explicitly call out that it doesn't meet the requirements of the RFC.

designs/nodepool-node-limits.md
```diff
@@ -0,0 +1,144 @@
...
+- Potential undefined behavior if Karpenter allows more nodes at once, but the cloud provider blocks disruption at the infrastructure level (e.g., due to exhaustion of IP addresses)
+    - This may lead to node provisioning timeout failures, and the user running into undesired node churn if  disruption occurs.
+
+### Option 3: Configurable Hard/Soft Node Limits
+
+**Description**: Allow users to configure whether limits are enforced as hard or soft constraints per NodePool. This can be configured via a new field in the NodePool spec `nodepool.spec.disruption.nodeLimitMode`.
```
jmdeal: I'd rather see us stage support for generic hard limits, since I think that is the natural extension of this feature. It's pretty easy to come up with hypothetical examples of why you'd want a hard limit on specific resources (e.g. GPUs). This could be something like what @GnatorX proposed in this [comment](https://github.com/kubernetes-sigs/karpenter/pull/2525/files#r2417124678) or we could consider something like this:

```yaml
limits:
  soft:
    nodes: 10
  hard:
    nodes: 12
  # Soft limits can still be specified at the top-level for backwards compatibility, 
  # but this approach would no longer be documented.
  cpu: 10
```

I feel like this is the cleanest solution, but I'm not sure how I feel about the backwards compatibility strategy.

designs/nodepool-node-limits.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Option 1: Hard Limits
+
+**Description**: Node limits are enforced as hard constraints that block both provisioning and graceful disruption (Consolidation and Drift) when the limit is reached. Graceful disruption is blocked because Karpenter requires the new nodes to be schedulable and ready before the old node can be terminated, which can lead to more nodes than the limit at one state.
+
+**Benefits**:
+- Guarantees the invariant that node count never exceeds the limit
+- Consistent behavior across all operations
+
+**Drawbacks**:
```
jmdeal, line 31: Another drawback is that this doesn't leave us with a path to introducing soft limits for nodes in the future, short of introducing a separate field to toggle the mode like suggested in option 3. I don't love the separate field option because I do think that's more difficult to reason about. I also think different resources having different disruption semantics is difficult to reason about in this approach.

maxcao13:
I've update the doc according to all the comments so far, thanks for the reviews everyone :-)

jmdeal:
Ran across this [comment](https://github.com/kubernetes-sigs/karpenter/issues/1177#issuecomment-2068571016) on a different issue as well, and I think it brings up a good point about hard limits: What should Karpenter's behavior be when a NodePools limits are tightened and there are now more nodes than the allowed limits? This is worth considering.

maxcao13:
> Ran across this https://github.com/kubernetes-sigs/karpenter/issues/1177#issuecomment-2068571016 on a different issue as well, and I think it brings up a good point about hard limits: What should Karpenter's behavior be when a NodePools limits are tightened and there are now more nodes than the allowed limits? This is worth considering.

I talk about it here:
https://github.com/kubernetes-sigs/karpenter/pull/2525/files#diff-5eac97882a24e1c56d7ac0dc9cd56c6c5d7ca182f5e1344bfe644eee898a5132R207-R215

I think its makes sense to enforce the limit if they are applied lower than the current number of nodes in the nodepool, and to reuse the existing deprovisioning logic that already exists in the static capacity code.

maxcao13:
I'm closing this RFC since "hard" limits are not feasible to be kept consistent in a distributed system. Limits today will continue to be best-effort, including node limits.

See this thread for more information: https://github.com/kubernetes/autoscaler/pull/8702

PR #2526, opened 2025-09-20
feat(limits): support NodePool node limits

maxcao13:
Fixes: https://github.com/issues/created?issue=kubernetes-sigs%7Ckarpenter%7C2657, #732 

**Description**
This PR attempts to allow NodePool Node limits to work in the same behaviour as CPU and memory nodepool limits, and properly limit the resource from being created over the limit _specifically_ during provisioning.

See the original issue for more details.

### DIFF
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -27,6 +27,7 @@ import (
 	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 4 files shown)


pkg/controllers/state/statenode.go
```diff
@@ -349,20 +350,25 @@ func (in *StateNode) Initialized() bool {
...
-					ret[resourceName] = quantity
+				if resources.IsZero(capacity[resourceName]) {
+					capacity[resourceName] = quantity
 				}
 			}
-			return ret
+		} else {
+			capacity = lo.Assign(in.NodeClaim.Status.Capacity)
 		}
-		return in.NodeClaim.Status.Capacity
```
ryan-mist, line 364: nit: The if/else makes it a bit confusing to look at. Since we only need to add the extra `resources.Node` when there is a NodeClaim without a Node yet, I wonder if we can just change that return statement?
e.g.
```
func (in *StateNode) Capacity() corev1.ResourceList {
	if !in.Initialized() && in.NodeClaim != nil {
		// Override any zero quantity values in the node status
		if in.Node != nil {
			...
		}
		return lo.Assign(in.NodeClaim.Status.Capacity, corev1.ResourceList{resources.Node: resource.MustParse("1")})
	}
	...
}
```

maxcao13:
As talked about in the recent karpenter working group, I can get rid of the flag and just make this the default behaviour if we are bringing functionality parity for node limits in-line the other resource limits (cpu, memory), but not exactly sure if that's the case or not.

EDIT: based on https://github.com/kubernetes-sigs/karpenter/issues/2657#issuecomment-3770492532, I believe that node limits are not in line with the other resource limits. I will remove the feature flag in the next round of reviews if a maintainer agrees with me.

DerekFrank:
I'm aligned, lets move forward with a flagless approach

maxcao13:
@DerekFrank
Thanks! Fixed in latest push, along with a unit test case that covers the use case. If there needs to be more validation here, please let me know.

DerekFrank:
Sorry I haven't had a chance to take a look at this, I'm hoping to take a look sometime this week

PR #2534, opened 2025-09-24
fix: over provisioning static nodeclaims during controller crashes

sumukha-radhakrishna:
Fixes over provisioning of NodeClaims during controller crashes.

**Description**

Since we use local Cluster NodePoolState to account for limits, if during provisioning controller dies then the next leader would need to wait until cluster state is populated to avoid over-provisioning of resources. 

We dont need to sync on cluster state for deprovisioning as we gradually populate NodePoolState and look at the replicas to delete NC when over-provisioned.

UT and tested by rolling and restarting karpenter to check if we overprovision resources.
Even tested deprovisioning case.

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -897,3 +897,8 @@ func (c *Cluster) triggerConsolidationOnChange(old, new *StateNode) {
 		return
 	}
 }
+
+// HasSynced returns whether the cluster state has been synchronized at least once.
+func (c *Cluster) HasSynced() bool {
+	return c.hasSynced.Load()
+}
--- a/pkg/controllers/static/provisioning/controller.go
+++ b/pkg/controllers/static/provisioning/controller.go
@@ -73,6 +73,13 @@ func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	// We need to wait until our representation of cluster is populated accurately with what we have in api-server or else
+	// we would end up over provisioning due to misrepresentation of NodePoolState in our cluster state.
+	// This usually happens when there is controller crash, so we check if the cluster has synced atleast once.
+	if !c.cluster.HasSynced() && !c.cluster.Synced(ctx) {
+		return reconcile.Result{RequeueAfter: time.Second}, nil
+	}
+
 	runningNodeClaims, _, nodesPendingDisruptionCount := c.cluster.NodePoolState.GetNodeCount(np.Name)
 	desiredReplicas := lo.FromPtr(np.Spec.Replicas)
 	// Size down of replicas will be handled in deprovisioning controller to drain nodes and delete NodeClaims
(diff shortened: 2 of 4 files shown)


pkg/controllers/static/provisioning/controller.go
```diff
@@ -73,6 +73,13 @@ func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	// We need to wait until our representation of cluster is populated accurately with what we have in api-server or else
+	// we would end up over provisioning due to misrepresentation of NodePoolState in our cluster state.
+	// This usually happens when there are controller crashes
+	if !c.cluster.Synced(ctx) {
+		return reconcile.Result{RequeueAfter: time.Second}, nil
+	}
```
jmdeal, line 81: One of the advantages of our current approach is that we don't need to await cluster sync every time we do a provisioning action. I think we'd like to maintain that if possible. I think the most straightforward approach to fix this issue and achieve this property would be to introduce a separate `Synced` method on NodePoolState and that is a "sticky" wrapper around `cluster.Synced()`. Once it returns true once, it will always return true without rechecking.

jmdeal: I think we can probably do better than this, but I can see us punting that for the longer term effort to break up cluster state.

sumukha-radhakrishna: Cluster State has `HasSynced` to check if we have synced atleast once, gonna use that. 
HasSynced will be true and Synced will be false in the case where we have nodeclaims without providerID, we wait until NodeClaims resolve providerID values to mark as Synced, in this case we dont really care as we mark the NodeClaim as Active in NodePool state regardless of weather we have resolved NodeClaim providerIDs

pkg/controllers/state/cluster.go
```diff
@@ -55,7 +55,8 @@ type Cluster struct {
 	kubeClient    client.Client
 	cloudProvider cloudprovider.CloudProvider
 	clock         clock.Clock
-	hasSynced     atomic.Bool
+	// We use this value to determine if we have synced the cluster state atleast once
+	HasSynced atomic.Bool
```
jmdeal: I'd expose this through a `Getter` rather than the raw atomic - we don't want any other part of the codebase updating this.

pkg/controllers/static/provisioning/controller.go
```diff
@@ -73,6 +73,15 @@ func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.
...
 	}
 
+	// We need to wait until our representation of cluster is populated accurately with what we have in api-server or else
+	// we would end up over provisioning due to misrepresentation of NodePoolState in our cluster state.
+	// This usually happens when there is controller crash, so we check if the cluster has synced atleast once.
+	if !c.cluster.HasSynced.Load() {
+		if !c.cluster.Synced(ctx) {
+			return reconcile.Result{RequeueAfter: time.Second}, nil
+		}
+	}
```
jmdeal: tiny nit: 
```suggestion
	if !c.cluster.HasSynced.Load() && !c.cluster.Synced(ctx) {
		return reconcile.Result{RequeueAfter: time.Second}, nil
	}
```

PR #2485, opened 2025-09-05
fix: make toolchain failing due to deletion of asciicheck

sumukha-radhakrishna:
Fixes: Failing CI and make toolchain

* Upgrades golangci-lint
* upgrades the config 
* Ran presubmit to auto lints/corrects based on golangci-lint v2

Currently make toolchain fails with 
```
/Users/rsumukha/go/pkg/mod/github.com/golangci/golangci-lint@v1.64.8/pkg/golinters/asciicheck/asciicheck.go:4:2: reading github.com/tdakkota/asciicheck/go.mod at revision v0.4.1: git ls-remote -q origin in /Users/rsumukha/go/pkg/mod/cache/vcs/e91413ff76cb95901df8e2d834855d07c617befa58b5014bc6cd6fd1148034dd: exit status 128:
	remote: Repository not found.
	fatal: repository 'https://github.com/tdakkota/asciicheck/' not found
make: *** [toolchain] Error 1
```

asciicheck is a dependency of golangci-lint, asciicheck has been deleted 🫤 https://github.com/tdakkota/asciicheck 
CI will start failing once the cached package expire. 

golangci started maintaining a forked version of asciicheck - https://github.com/golangci/asciicheck

They have updated their go.mod to use the forked version but havent released yet. We will use main until they release. 
https://github.com/golangci/golangci-lint/pull/6018

SInce we need to upgrade to v2 of golangci-lint, we need to migrate the config that changes/corrects based off v2.

**How was this change tested?**

make toolchain succeeds and make presubmit

PR #2539, opened 2025-09-25
fix: allow non-churn empty nodes to be disrupted (#2206)

AndrewMitchell25:
Fixes #N/A 

**Description**
Cherry picks "fix: allow non-churn empty nodes to be disrupted (#2206)" and upgrades golangci-lint (similar to #2485)
`make toolchain` and `make presubmit`

### DIFF
--- a/pkg/controllers/disruption/singlenodeconsolidation.go
+++ b/pkg/controllers/disruption/singlenodeconsolidation.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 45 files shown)


pkg/controllers/disruption/singlenodeconsolidation.go
```diff
@@ -33,10 +34,17 @@ const SingleNodeConsolidationType = "single"
 // SingleNodeConsolidation is the consolidation controller that performs single-node consolidation.
 type SingleNodeConsolidation struct {
 	consolidation
+	PreviouslyUnseenNodePools sets.Set[string]
```
jmdeal: It looks like this is vestigial from the change we dropped from this backport to shuffle nodepools. We can drop this right?

AndrewMitchell25: Looks like it I'll update this

PR #2550, opened 2025-09-29
fix: Skip NodePools that cannot provide a set of instance types

### LINKED ISSUE #2524: nodeoverlay controller blocks all NodePools if one isn't ready
### Description

**Observed Behavior**:

We had a broken nodepool. (Accidentally deployed the definition to a cluster that didn't have the appropriate subnets)

All NodePools stopped being able to provision Nodes/NodeClaims

**Expected Behavior**:

Only the NodePool that's missing it's subnets should fail to deploy - this should not break all other NodePools.

Enable the NodeOverlay feature gate
Create a broken NodePool

```
---
apiVersion: karpenter.sh/v1
kind: NodePool
spec:
  template:
    spec:
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: worker-vpn
---
apiVersion: karpenter.k8s.aws/v1
kind: EC2NodeClass
metadata:
  name: worker-vpn
spec:
  subnetSelectorTerms:
  - tags:
      karpenter-discovery-name: doesn't-exist
```

**Versions**:
- Chart Version: 1.7.0
- Kubernetes Version (`kubectl version`): v1.33.3-eks-b707fbb

engedaam:
Fixes #2524 

**Description**
- allow node overlay controller to continue when one node pool fails to retive the instance tpes

- N/A


moko-poi (commented):
Should we add a check for when all NodePools fail?

This would allow us to return early instead of continuing with an empty map.
```go
if len(nodePoolToInstanceTypes) == 0 && len(nodePoolList.Items) > 0 {
    return reconcile.Result{}, fmt.Errorf("unable to get instance types for all NodePools")
}
```

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -85,10 +89,14 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	for i := range nodePoolList.Items {
 		its, err := c.cloudProvider.GetInstanceTypes(ctx, &nodePoolList.Items[i])
 		if err != nil {
-			return reconcile.Result{}, fmt.Errorf("listing instance types, %w", err)
+			log.FromContext(ctx).WithValues("NodePool", klog.KObj(&nodePoolList.Items[i])).Error(err, "skipping, unable to resolve instance types")
```
jmdeal, line 92: We should requeue more aggressively if we fail to get instance types for any NodePools, otherwise we might get resolution failures until we poll again.

engedaam: Are you think we requeue if we do not reslove instance for a single nodepool? Also, how fast are you thinking?

matthchr: Doesn't returning `fmt.Errorf("unable to get instance types for all NodePools")` below already trigger backoff according to the ratelimiter below, up to a delay of 1m?

So we'll already retry every minute in the worst case. Do you want to be more aggressive than that?

matthchr (approved):

ryan-mist:
Closing this one out as https://github.com/kubernetes-sigs/karpenter/pull/3002 was merged

PR #2553, opened 2025-09-29
test: configmap case for dra kwok driver

alimaazamat:
Contributes to #[2545](https://github.com/kubernetes-sigs/karpenter/issues/2545) 

**Overall Changes**

```
```
Explanation of architecture is in readme.

Unit tests + e2e integration tests

### DIFF
--- a/dra-kwok-driver/pkg/controllers/resourceslice.go
+++ b/dra-kwok-driver/pkg/controllers/resourceslice.go
@@ -27,6 +27,7 @@ import (
 	"github.com/awslabs/operatorpkg/singleton"
 	corev1 "k8s.io/api/core/v1"
 	resourcev1 "k8s.io/api/resource/v1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -429,9 +430,11 @@ func (r *ResourceSliceController) nodeMatchesFieldExpression(node *corev1.Node, 
 }
 
 // resourceSliceNeedsUpdate checks if an existing ResourceSlice needs to be updated
-// based on device count or slice count changes
 func (r *ResourceSliceController) resourceSliceNeedsUpdate(existing *resourcev1.ResourceSlice, desiredDevices []resourcev1.Device, desiredSliceCount int64) bool {
-	return len(existing.Spec.Devices) != len(desiredDevices) || existing.Spec.Pool.ResourceSliceCount != desiredSliceCount
+	if existing.Spec.Pool.ResourceSliceCount != desiredSliceCount || len(existing.Spec.Devices) != len(desiredDevices) {
+		return true
+	}
+	return !apiequality.Semantic.DeepEqual(existing.Spec.Devices, desiredDevices)
 }
 
 // cleanupOrphanedResourceSlices removes ALL ResourceSlices managed by this driver


dra-kwok-driver/pkg/config/types.go
```diff
@@ -0,0 +1,234 @@
...
+	"regexp"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Config represents the DRA KWOK driver configuration loaded from ConfigMap
+type Config struct {
+	// Driver specifies the DRA driver name for ResourceSlices
+	Driver string `yaml:"driver" json:"driver"`
```
jmdeal: nit: we shouldn't need to declare yaml and json names for these fields, we typically only specify json.

dra-kwok-driver/pkg/config/types.go
```diff
@@ -0,0 +1,234 @@
...
+}
+
+// Mapping defines a mapping from node selector to ResourceSlice configuration
+type Mapping struct {
+	// Name is a human-readable identifier for this mapping
+	Name string `yaml:"name" json:"name"`
+	// NodeSelector determines which nodes this mapping applies to
+	NodeSelector metav1.LabelSelector `yaml:"nodeSelector" json:"nodeSelector"`
+	// ResourceSlice defines the devices to create for matching nodes
+	ResourceSlice ResourceSliceConfig `yaml:"resourceSlice" json:"resourceSlice"`
```
jmdeal: Any reason you want to use a custom type here vs embedding the upstream type directly? I know I proposed the opposite in the NodeOverlay RFC, but I think it makes sense in this case where we'll want to verify that Karpenter behaves correctly for non-customer facing fields (e.g. pools).

dra-kwok-driver/main.go
```diff
@@ -0,0 +1,122 @@
...
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		logger.Error(err, "unable to set up ready check")
+		panic(err)
+	}
+
+	// Initialize ResourceSlice controller first
+	resourceSliceController := controllers.NewResourceSliceController(
+		mgr.GetClient(),
+		"karpenter.sh.dra-kwok-driver",
+		nil, // Will be set after ConfigMap controller is created
```
jmdeal: Self TODO: come back to this

Makefile
```diff
@@ -81,6 +96,15 @@ test: ## Run tests
 		--ginkgo.v \
 		-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./...
 
+test-dra: ## Run DRA tests
```
jmdeal:
```suggestion
test-dra-driver: ## Run DRA tests
```

Since this is specifically running the driver's test suite, rather than Karpenter's let's clarify

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,302 @@
...
+		Client:           client,
+		driverName:       driverName,
+		configController: configController,
+	}
+}
+
+// SetConfigController sets the config controller reference (used when controllers are initialized in sequence)
+func (r *ResourceSliceController) SetConfigController(configController *ConfigMapController) {
+	r.configController = configController
+}
```
jmdeal: To break this circular dependency, I'd consider using a shard config store rather than injecting the configmap controller directly.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,302 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/karpenter/dra-kwok-driver/pkg/config"
+)
+
+// ResourceSliceController manages ResourceSlice lifecycle based on Node events and driver configuration
+type ResourceSliceController struct {
+	client.Client
```
jmdeal: nit: I wouldn't embed this type, I don't think there's a reason we would need `ResourceSliceController` to satisfy the same interfaces as `client.Client`. Typically we just call this member `kubeClient` in all of our other controllers.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// Reconcile handles Node changes and manages corresponding ResourceSlices
+func (r *ResourceSliceController) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	// Get the node
+	node := &corev1.Node{}
+	err := r.Get(ctx, req.NamespacedName, node)
+	if err != nil {
```
jmdeal: nit(style): prefer to inline when possible
```suggestion
	if err := r.Get(ctx, req.NamespacedName, node); err != nil {
```

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,302 @@
...
+		if errors.IsNotFound(err) {
+			// Node was deleted - clean up ResourceSlices
+			log.Info("node deleted, cleaning up resourceslices", "node", req.Name)
+			return r.cleanupResourceSlicesForNode(ctx, req.Name)
+		}
+		log.Error(err, "failed to fetch node")
+		return reconcile.Result{}, err
```
jmdeal: You should be able to drop this error, controller-runtime will print errors when you return an error from a reconciler so long as the logger is configured correctly. We will typically wrap the error with a bit of additional context though:

```suggestion
		log.Error(err, "failed to fetch node")
		return reconcile.Result{}, serrors.Wrap(fmt.Errorf("getting node, %w"), "node", klog.KRef("", req.Name))
```

jmdeal: This same comment applies anywhere else where an error is printed before we return from a reconciler.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,302 @@
...
+	if !r.isKWOKNode(node) {
+		log.V(2).Info("skipping non-KWOK node", "node", node.Name)
+		return reconcile.Result{}, nil
+	}
+
+	// Get current driver configuration
+	cfg := r.configController.GetConfig()
+	if cfg == nil {
+		log.V(1).Info("no driver configuration available, skipping node", "node", node.Name)
+		return reconcile.Result{}, nil
```
jmdeal: There might not be driver information now, but there could be in the future right? Maybe there's just a race concerning when the informer caches are updated, or we created a config after the node has already joined the cluster. Either way, I think we want to react.

There are a bunch of ways we could do this, but I think the most straightforward method would be to evaluate all nodes in the cluster periodically. So rather than reconciling on individual nodes, we list all of the nodes, compute the expected state for each of those nodes (which ResourceSlices need to be created or updated), and then update the cluster.

edit: I see you achieved this with a callback, that approach works to. I don't think there are any race conditions with that approach. Personally, I'd lean towards the polling approach though for simplicity but let me know what you think.

dra-kwok-driver/pkg/config/types.go
```diff
@@ -0,0 +1,173 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	resourcev1 "k8s.io/api/resource/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Config represents the DRA KWOK driver configuration loaded from ConfigMap
+type Config struct {
```
jmdeal: Since we're creating a strongly typed config anyway, I wonder if it makes more sense to just create a CRD (e.g. `test.karpenter.sh/v1 - DRAConfig`). This would allow you to move the validation to the API server (using CEL) and ensure types are satisfied at the API server layer. I think this is a little more relevant now since the CRD based alternative (NodeOverlay) has been put on hiatus, but I think it's a good idea regardless.

dra-kwok-driver/pkg/controllers/configmap.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+// Register registers the controller with the manager
+func (r *ConfigMapController) Register(ctx context.Context, mgr manager.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		Named("configmap").
+		For(&corev1.ConfigMap{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(obj client.Object) bool {
+			// Only watch our specific ConfigMap
+			return obj.GetName() == r.configMapName && obj.GetNamespace() == r.configMapNamespace
```
jmdeal: We're going to want to be able to support multiple configs in a cluster at any given time. We'll definitely want to be able to validate behavior with multiple NodePools and different drivers, as well as multiple drivers within the same NodePool

dra-kwok-driver/pkg/config/types.go
```diff
@@ -0,0 +1,173 @@
...
+	// Mappings defines how to map node labels to device configurations
+	Mappings []Mapping `yaml:"mappings" json:"mappings"`
+}
+
+// Mapping defines a mapping from node selector to ResourceSlice configuration
+type Mapping struct {
+	// Name is a human-readable identifier for this mapping
+	Name string `yaml:"name" json:"name"`
+	// NodeSelector determines which nodes this mapping applies to
+	NodeSelector metav1.LabelSelector `yaml:"nodeSelector" json:"nodeSelector"`
```
jmdeal: I'd consider expanding this to NodeSelectorTerms, rather than a single NodeSelector. This way we can apply the same mapping to disjoint sets of nodes.

dra-kwok-driver/dra-kwok-driver
jmdeal, line 1: I assume this is a compiled binary? Is this output by a makefile, or just from your own development? We should drop it regardless, but ideally our Makefile shouldn't produce any local artifacts - it should just upload them to the configured container registry.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,302 @@
...
+			existing.Spec = resourceSlice.Spec
+			existing.Labels = resourceSlice.Labels
+
+			log.Info("updating resourceslice",
+				"resourceslice", resourceSliceName,
+				"node", node.Name,
+				"mapping", mapping.Name,
+				"devices", len(resourceSlice.Spec.Devices),
+			)
+			if err := r.Update(ctx, existing); err != nil {
```
jmdeal: Rather than updating the ResourceSlices directly, we could do the following:
- Create the new ResourceSlices in the same pool, but bump the generation
- Delete the ResourceSlices from the previous generation

This would allow our E2E cases to test the "incomplete" pool case, albeit not consistently. We'll have consistent tests for those scenarios in functional tests.

dra-kwok-driver/pkg/controllers/resourceslice.go
alimaazamat: Yeah I am fine with polling approach, is 30 sec loop okay?
Also faster to batch operations for updates so for now I grouped ResourceSlices by node for updates.
One thing to note is that if we have node failures it will just linger and repeatedly fail and keep polling on it? I think for now we keep this but we can revisit improving this later.

alimaazamat: I also added cleanUpOrphanedResourcesSlices to remove ResourceSlices when config doesn't exist just to ensure cluster state always matches config. Requests run sequentially for right now to avoid race conditions when reconciling, I dont think there wil be a scenario where cycles would queue up and cause issues for right now.

dra-kwok-driver/pkg/controllers/resourceslice.go
alimaazamat: I decided on in place updates, and to test incomplete pool with ConfigMap in E2E tests. Then by bumping generations in place it will tell us that the pool is being updated, and for pools with multiple ResourceSlices it will demonstrate that the pool is in progress of updating when the generation of all the ResourceSlices is not matching the desired.
For example we start with two ResourceSlices at gen 0. We update each ResourceSlice one by one. So we have a period of time where only one ResourceSlice is bumped to gen 1 and the other still at gen 0 which demonstrates the incomplete pool case.

dra-kwok-driver/pkg/controllers/configmap.go
alimaazamat: CRD would solve this.

alimaazamat: Actually I've been thinking this dra KWOK driver as just one driver for now. Will implement CRD with just karpenter.sh dra mock driver and later think about multiple driver configuration later

dra-kwok-driver/pkg/config/types.go
alimaazamat: Okay implementing CRD with just one ResourceSlice controller rather than having two controllers.

Makefile
```diff
@@ -1,6 +1,7 @@
 # This is the format of an AWS ECR Public Repo as an example.
 export KWOK_REPO ?= ${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_DEFAULT_REGION}.amazonaws.com
 export KARPENTER_NAMESPACE=kube-system
+export KIND_CLUSTER_NAME ?= test-cluster
```
alimaazamat, line 4: Will remove this later, left it because it was faster to run e2e tests with this env var

dra-kwok-driver/pkg/apis/v1alpha1/draconfig_types.go
```diff
@@ -0,0 +1,182 @@
...
+	NodeSelectorTerms []corev1.NodeSelectorTerm `json:"nodeSelectorTerms,omitempty"`
+
+	// resourceSlice defines the ResourceSlice template to create for matching nodes.
+	// The driver will create a ResourceSlice using this template, filling in:
+	// - metadata.name (generated from node name and mapping name)
+	// - spec.nodeName (from the matching node)
+	// - spec.driver (from DRAConfigSpec.Driver)
+	//
+	// +required
+	ResourceSlice ResourceSliceTemplate `json:"resourceSlice,omitzero"`
```
jmdeal: How difficult is it to expand this to a slice to support the multi-slice per pool use-case?

dra-kwok-driver/pkg/apis/crds/examples/sample-configs.yaml
```diff
@@ -0,0 +1,68 @@
...
+      nodeSelectorTerms:
+        - matchExpressions:
+            - key: node.kubernetes.io/instance-type
+              operator: In
+              values:
+                - g5.48xlarge
+                - p4d.24xlarge
+      resourceSlice:
+        pool:
+          name: gpu-pool
```
jmdeal: We'll need a way to dynamically name pools to prevent overlap. The standard convention that I've seen is `${DRIVER_NAME}/${NODE_NAME}`. I think that covers our use-case. 

What I'd consider is allowing multiple `ResourceSlices` per mapping and renaming `mappings` to `pools` to codify this concept. Then we can dynamically generate the pool info when we create the ResourceSlices rather than specifying it manually.

Alternatively we could allow templating in the config (e.g. `name: gpu-pool-${node.metadata.name}`), but I think that's overkill for our use-case.

dra-kwok-driver/main.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+		logger.Error(err, "unable to set up ready check")
+		panic(err)
+	}
+
+	// Initialize ResourceSlice controller (single controller manages all drivers)
+	resourceSliceController := controllers.NewResourceSliceController(
+		mgr.GetClient(),
+		"karpenter", // Just namespace
```
jmdeal: I don't think there's a need to namespace these resources. I could see us namespacing them if the objects they're acting against (nodes) were also namespaced, but since nodes are global I think it makes sense for DRAConfigs to be as well.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,514 @@
...
+		kubeClient: kubeClient,
+		namespace:  namespace,
+	}
+}
+
+// Register registers the controller with the manager and starts the polling loop
+func (r *ResourceSliceController) Register(ctx context.Context, mgr manager.Manager) error {
+	// Start a goroutine for periodic reconciliation
+	// This runs outside of controller-runtime's reconciliation loop
+	go r.startPollingLoop(ctx)
```
jmdeal: I'd take a look at the singleton reconciler pattern we use elsewhere in Karpenter. It's pretty simple in practice: it's a channel-based reconciler that gets kicked off with a single event and `reconcile.Result{RequeueAfter: controllerPeriod}` keeps it iterating. The advantage of this approach is that you still get controller-runtime's error handling behavior.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,514 @@
...
+}
+
+// groupConfigsByDriver groups DRAConfigs by their driver name, warning about duplicates
+func (r *ResourceSliceController) groupConfigsByDriver(logger logr.Logger, configs []v1alpha1.DRAConfig) map[string]v1alpha1.DRAConfig {
+	configsByDriver := make(map[string]v1alpha1.DRAConfig)
+	for _, cfg := range configs {
+		driverName := cfg.Spec.Driver
+		if existingCfg, exists := configsByDriver[driverName]; exists {
+			// Duplicate driver! Log warning and skip
+			logger.Info("WARNING: Multiple DRAConfigs found for same driver, using first one",
```
jmdeal: Are there any use-cases that we'd want to support where you define multiple configs per driver? I think since we support multiple mappings, the answer is no. I could see a future where we'd want to since we're hitting object size limits, but I don't envision us hitting those limits in test scenarios. 

That being said, I'm not sure if there's a reason we need this restriction. Does it simplify logic significantly?

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,514 @@
...
+		} else if errors.IsNotFound(err) {
+			// ResourceSlice doesn't exist - create it
+			desired := &resourcev1.ResourceSlice{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: resourceSliceName,
+					Labels: map[string]string{
+						"kwok.x-k8s.io/managed-by": "dra-kwok-driver",
+						"kwok.x-k8s.io/node":       node.Name,
+						"kwok.x-k8s.io/mapping":    mapping.Name,
+						"kwok.x-k8s.io/driver":     driverName,
```
jmdeal: Are these standard kwok labels? Does kwok do anything with these? If they're custom labels we're defining ourselves, we typically use the `test.karpenter.sh` domain. Of course this doesn't functionally matter, it's just convention.

dra-kwok-driver/pkg/controllers/resourceslice.go
```diff
@@ -0,0 +1,514 @@
...
+					},
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion: "v1",
+							Kind:       "Node",
+							Name:       node.Name,
+							UID:        node.UID,
+						},
+					},
+				},
```
jmdeal: Just curious, are we relying on this? I don't have an issue including it since it will ensure resources will be cleaned up appropriately even if the test driver stops functioning and it matches what you would expect the lifecycle of a ResourceSlice to be.

dra-kwok-driver/pkg/apis/v1alpha1/draconfig_types.go
alimaazamat: Updated to ResourceSlices, multiple slices per pool https://github.com/kubernetes-sigs/karpenter/pull/2553/changes/114bff854765959f8c6652cb0dd58be21afe4391

dra-kwok-driver/pkg/controllers/resourceslice.go
alimaazamat: Not relying on it, controller has it's own cleanup: `cleanupUnexpectedResourceSlices`

dra-kwok-driver/pkg/controllers/resourceslice.go
alimaazamat: Yes those `kwok.x-k8s.io` are real KWOK labels, changed to use `test.karpenter.sh` https://github.com/kubernetes-sigs/karpenter/pull/2553/commits/9c6ec7319bee01a96ce49aac0314519acb60ce91

dra-kwok-driver/pkg/apis/crds/examples/sample-configs.yaml
alimaazamat: Done in https://github.com/kubernetes-sigs/karpenter/pull/2553/changes/114bff854765959f8c6652cb0dd58be21afe4391

dra-kwok-driver/pkg/controllers/resourceslice.go
alimaazamat: No it doesn't simplify that much I was just making sure we don't have multiple DRAConfigs per driver. But since we won't have that in the first place we can just return all pools per driver. https://github.com/kubernetes-sigs/karpenter/pull/2553/commits/ba2065958e5c1c1235d29ee2c6f0128e8b49be55

dra-kwok-driver/pkg/controllers/resourceslice.go
alimaazamat: Done here https://github.com/kubernetes-sigs/karpenter/pull/2553/changes/e9428480fcb2fe7c6ccf44c3a6194deb7b745a70

jmdeal:
Nice work, appreciate your patience on reviews!

/lgtm
/approve

PR #2562, opened 2025-10-05
docs: RFC for consolidation price improvement factor

jukie:
**Description**
This is a proposal to introduce a cost savings threshold when making consolidation decisions in order to reduce churn.

Related issue: https://github.com/aws/karpenter-provider-aws/issues/7146 and https://github.com/kubernetes-sigs/karpenter/issues/1440
RFC only but draft PR opened at https://github.com/kubernetes-sigs/karpenter/pull/2561


designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
...
+Introduce a **Price Improvement Factor** configuration option that allows users to require replacement nodes to offer **meaningful cost savings** before consolidation occurs. This factor acts as a threshold: only when `replacement_price < (current_price × factor)` will consolidation proceed.
+
+For example, a factor of **0.8** would require at least **20% cost savings**, while the default of **1.0** maintains current behavior (consolidate for any savings).
+
+### Configuration Levels
+
+The price improvement factor will be configurable at **two levels** with the following precedence:
+
+1. **NodePool-level** (highest priority) - Per-workload control
+2. **Operator-level** (fallback) - Cluster-wide default
```
ellistarn, line 29: We generally avoid global configuration of Karpenter and recommend things like Kyverno to control this type of behavior.

designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
...
+    // for consolidation to occur. Only consolidate when replacement nodes cost
+    // less than (current_price × factor).
+    //
+    // If not specified, falls back to the operator-level setting.
+    //
+    // Examples:
+    //   1.0 = Consolidate for any cost savings (legacy behavior)
+    //   0.8 = Require 20% cost savings
+    //   0.5 = Require 50% cost savings (very conservative)
+    //   0.0 = Disable price-based consolidation
```
ellistarn, line 55: Sugg: reorient this around percentage, where it's an integer from [0, 100]. The kubernetes ecosystem avoids floats in APIs due to precision serialization challenges.

designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
...
+The price improvement factor filters instance type options during consolidation decisions:
+
+1. **Resolve configuration value** (precedence order):
+   - If `NodePool.spec.disruption.consolidationPriceImprovementFactor` is set, use it
+   - Else, use operator-level `--consolidation-price-improvement-factor`
+   - Else, use default `1.0` which maintains current behavior.
+
+2. **Calculate threshold**: `maxAllowedPrice = currentPrice × priceImprovementFactor`
+
+3. **Filter instances**: Only consider instances where `launchPrice < maxAllowedPrice`
```
ellistarn, line 101: Make sure you're factoring multi node consolidation into this. There are a few variants:

1 -> 0 (remove)
1 -> 1 (replace)
n -> 1 (multi node)

We don't currently do this today

1 -> 2 (split)

But if we ever did, we'd need to make sure we were handling price improvement appropriately. I can't think of any reason that this would be particularly difficult, though.

designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
...
+  disruption:
+    consolidationPriceImprovementFactor: 0.8  # Test with 20% threshold
+
+# Phase 3: Roll out cluster-wide
+# (change operator default to 0.8, remove NodePool overrides)
+```
+
+## Backward Compatibility
+
+- **No breaking changes**: Default value of `1.0` maintains existing behavior
```
ellistarn, line 233: I am curious if we should explore a better default (e.g. .05%). I'd love to see some analysis on consolidations across a cluster over time, with a histogram of price improvement for each change.

There is a reasonable argument that Karpenter's consolidation algorithm is a heuristic and that we could launch with a low value for this setting without it being a breaking change. We make implementation changes to consolidation all the time and don't guarantee any deterministic decision making around cost/disruption tradeoffs.

designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
+# Consolidation Price Improvement Factor
```
ellistarn, line 1: I'm sure you're aware of this: https://github.com/kubernetes-sigs/karpenter/blob/main/designs/spot-consolidation.md#2-price-improvement-factor, but might be worth referencing as prior art.

designs/consolidation-price-improvement-factor.md
tzneal: One complication is we need to define how multi-node works with nodes from different nodepools with different price thresholds configured.


2 -> 1 Multi-Node
Node A = from nodepool with 10% threshold
Node B = from nodepool with 5% threshold


Replacement Node is 4% cheaper than A+B, should clearly not replace
Replacement Node is 6% cheaper than A+B, ??
Replacement Node is 10% cheaper than A+B, should clearly replace

I would guess it should take the most conservative value and not replace for the 6% cheaper case.

designs/consolidation-price-improvement-factor.md
jukie: Okay no problem. By this do you mean drop the operator-level config and keep it NodePool only?

designs/consolidation-price-improvement-factor.md
jukie: Yep I would agree with the approach of taking the most conservative. Said another way, each Node's improvement threshold must be met in order for the replacement to proceed.

designs/consolidation-price-improvement-factor.md
jukie: I would love to apply a different default but initially assumed the maintainers would prefer matching existing behavior upon initial release. Starting with .05% seems like it'd be a safe initial value and after user feedback it could be tuned higher.

designs/consolidation-price-improvement-factor.md
ellistarn: yep.

ellistarn:
@jukie, I like the thinking here, but I want to make sure we have exhausted "zero config" options. One question I always ask myself when adding new API surface is "how would a customer decide how much to configure this". 

1. If the answer is "vibes", then we probably shouldn't expose it and should just take a heuristic opinion. 
2. If the answer is "do XYZ calculation", we should explore automating that calculation
3. If the answer is "there is a fundamental tradeoff that only the customer can know, and this is how they know", then we add the config.

jukie:
Thanks @ellistarn. For workloads that don't have any negative experience associated with node churn and the resulting pod interruptions (long startup times, cache warmups, etc) users might prefer the current state of consolidating upon any form of savings but for workloads that do I think it is a tradeoff between cost savings and interruptions so I think it's justified to expose it as a configuration option.

ellistarn:
> I think it is a tradeoff between cost savings and interruptions so I think it's justified to expose it as a configuration option.

How will you decide to set this for your organization?

jukie:
An org would have the option of running multiple node pools each with different values based on the workloads that target them. However that has the end result and tradeoff of probably running more nodes than actually necessary so the next consideration might be to run a single (or some fewer number of NodePools) instead to still get the savings and adjust the improvement factor accordingly.

The decision would be a balance between cost savings and causing interruptions to workloads. Things like pod-deletion-cost or do-not-disrupt annotations could help here too.

ellistarn:
So you would expect organizations to run multiple load tests with different values and measure their overall costs in each version?

jukie:
If an org wants it perfectly tailored to their needs that might be a reasonable approach. However I suspect users who are seeking to tweak this are coming from the other side of the spectrum where the stronger desire is to reduce node churn and workload interruption. Node cost is a factor and should still be minimized which Karpenter solves for but node churn and workload interruption incurs another form of cost.

An org could directly measure all 3 parameters and use whatever works best for them.

I don't think there's a single value that's guaranteed to work for all use cases (even within the same org) so similar to how [ClusterAutoscaler](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#how-does-scale-down-work) handles it, I think it should be exposed.

ellistarn:
I'm open minded. I'm going to see if we can gather some data on this and see if it can help inform our decisions. Will get back soon.

jukie:
I've had this running locally for about a month and have some takeaways. The savings threshold does work as expected and lead to the desired behavior and avoids constant node churn.
I also wanted to callout some of the nuances different levels where it would actually take action:
  - A threshold of 10-20% will in most cases only prevent instance type generation consolidations (e.g. old workload requested m7 but was consolidated to a different node or scaled down, and the pods on the node would all fit on an m6).
  - A threshold of 100% leads to almost never consolidating unless the node can be fully deleted (e.g. won't consolidate from 4xlarge --> 2xlarge) or it's a multi-node consolidation action that leads to the same result. For many users, this might be a desirable configuration that strikes a balance between cost savings and disruption.
  - Generally speaking this threshold is directionally correct but doesn't map 1:1 with cluster resource utilization (nor should it in my opinion)

I think this is a great feature to include and provides an additional knob for controlling disruption

designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
...
+
+**Reference Issue**: https://github.com/aws/karpenter-provider-aws/issues/7146
+
+The core problem is that Karpenter consolidates nodes whenever a cheaper alternative exists, without requiring **meaningful cost improvement**. A node running at $0.95/hr will be immediately replaced by one at $0.90/hr (5% savings), then again by one at $0.85/hr (5.5% savings), creating a cascade of disruptions for minimal benefit.
+
+## Proposed Solution
+
+### Overview
+
+Introduce a **Price Improvement Factor** configuration option that allows users to require replacement nodes to offer **meaningful cost savings** before consolidation occurs. This factor acts as a threshold: only when `replacement_price < (current_price × factor)` will consolidation proceed.
```
DerekFrank, line 20: This has some interesting side effects with reserved capacity. Because reserved capacity has their price decimal shifted a few places downward to ensure they are always cheaper than other offerings, they are always valid to consolidation onto from OD or Spot unless you set the knob to 0. However, the knob affects reserved to reserved consolidation just fine.

I think that behavior is expected and ideal for the single cluster case, but I wonder if thats a sharp edge for users sharing reservations across clusters.

designs/consolidation-price-improvement-factor.md
```diff
@@ -0,0 +1,350 @@
...
+
+For example, a factor of **0.8** would require at least **20% cost savings**, while the default of **1.0** maintains current behavior (consolidate for any savings).
```
DerekFrank, line 22: I think you're right, a percentage is the correct bar to use, but I think there is one other option to at least discuss.

The other option that comes to mind is a flat value: i.e. 1$/hour

The two surfaces are identical if nodes are homogeneously priced. If a nodepool contains expensive and cheap nodes, the flat value would be more likely to affect cheap nodes and less likely to affect expensive ones. An example:

Expensive Node: 1.0$/hour
Cheap node: .1$/hour
Threshold: .05$/hour

In this case, the cheap node needs a replacement that reduces its price by 50% or more to be valid, but the expensive node needs a node thats only 5% cheaper. 

I think this would lead to Karpenter preferring small nodes and generally prevent multi node consolidation of very cheap nodes.

jukie: I suspect a flat value will lead to a lot of unexpected behavior such as what you've explained so would very much prefer something percentage-based or something else that scales fairly across instance types.

designs/consolidation-price-improvement-factor.md
jukie: Could you elaborate on this scenario a bit? I think the same logic would exist today where reserved capacity instance types are preferred. I don't think karpenter handles the cross-cluster reservations at all today so I am curious how users handle this besides carving up the reservation between clusters.

PR #2565, opened 2025-10-08
chore: Add `AddToScheme` pattern to v1 API version

zamai:
**Description**
Many Kubernetes controllers and integration tests rely on AddToScheme helpers to register CRDs. Adding  this function to Karpenter simplifies setup, and aligns with standard Kubernetes API patterns.

This PR adds AddToScheme support for Karpenter API types, enabling them to be registered with a runtime.Scheme. This makes it straightforward for external controllers, tests, and tools to use Karpenter’s types with controller-runtime clients, decoders, and webhooks without manual type registration.

Example how currently I register these types: 
```go
      // Manually register Karpenter types since it doesn't export AddToScheme
	karpenterGV := schema.GroupVersion{Group: "karpenter.sh", Version: "v1"}
	metav1.AddToGroupVersion(s, karpenterGV)
	s.AddKnownTypes(karpenterGV,
		&karpenterv1.NodePool{},
		&karpenterv1.NodePoolList{},
		&karpenterv1.NodeClaim{},
		&karpenterv1.NodeClaimList{},
	)
```

After Fix: 
```go
	s := runtime.NewScheme()
        karpenterv1.AddToScheme(s)

```

Manually, and I've generated some tests locally, but I feel that they might be unnecessary in the repo. 

Please let me know if adding tests for this variable would a requirement, to merge. 


pkg/apis/v1/register.go
```diff
@@ -0,0 +1,47 @@
...
+	GroupVersion = schema.GroupVersion{Group: apis.Group, Version: "v1"}
+
+	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+
+	// AddToScheme adds the types in this group-version to the given scheme.
+	AddToScheme = SchemeBuilder.AddToScheme
+)
+
+func addKnownTypes(s *runtime.Scheme) error {
```
DerekFrank, line 38: nit: How do you feel about:

```suggestion
func (s *runtime.Scheme) addKnownTypes() {
```

DerekFrank (commented):
Do you have any examples of controllers that follow this type of pattern that could be refactored?

pkg/apis/v1/register.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"sigs.k8s.io/karpenter/pkg/apis"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: apis.Group, Version: "v1"}
+
+	// SchemeBuilder is used to add go types to the GroupVersionKind scheme
+	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
```
DerekFrank, line 32: If i'm not mistaken, this includes several different options for adding known types, as you could call:

```
v1.SchemeBuilder.Build()
```
as well. Is this intentional?

zamai: It was intentionally copping this registration from go-client:
https://github.com/kubernetes/client-go/blob/d9e3e97bc805a0a2936112b8a33b78be27b09eec/scale/scheme/register.go

pkg/apis/v1/register.go
zamai: I feel okay about it, but I used `runtime.NewSchemeBuilder` which accepts function with this signature:
```
NewSchemeBuilder(funcs ...func(*Scheme) error) SchemeBuilder 
```

I was following the same pattern I saw other packages implement, example: 
https://github.com/argoproj/argo-rollouts/blob/92f4801475175abc888f7a29fee63fd965e15c14/pkg/apis/rollouts/v1alpha1/register.go#L35

moko-poi (approved):

pkg/apis/v1/register.go
zamai: @DerekFrank bump, maybe we could merge this tiny improvement?

pkg/apis/v1/register.go
```diff
@@ -0,0 +1,47 @@
...
+	// AddToScheme adds the types in this group-version to the given scheme.
+	AddToScheme = SchemeBuilder.AddToScheme
+)
+
+func addKnownTypes(s *runtime.Scheme) error {
+	s.AddKnownTypes(GroupVersion,
+		&NodePool{},
+		&NodePoolList{},
+		&NodeClaim{},
+		&NodeClaimList{},
```
jmdeal, line 43: We need `NodeOverlay` and `NodeOverlayList` as well, but once those are added this should be good to merge.

PR #2584, opened 2025-10-22
feat: add a nodepool cost metric

DerekFrank:
Fixes #N/A 

**Description**

Adds a new object that tracks cluster cost.

Adds a controller that uses this new object to emit a per-nodepool cost metric

Added unit tests for the clusterCost object. I also ran it in my cluster:

```
# HELP karpenter_nodepools_cost_total Total cost of the nodepool
# TYPE karpenter_nodepools_cost_total gauge
karpenter_nodepools_cost_total{nodepool="default-4cpu"} 0.153
```

```
NAME                 TYPE        CAPACITY    ZONE         NODE                                           READY     AGE     IMAGEID                 ID                                      NODEPOOL       NODECLASS
default-4cpu-dbwbj   c6a.large   on-demand   us-west-2a                                                  Unknown   7m20s   ami-05046f6e7a37d62b5   aws:///us-west-2a/i-09a93b046e7b584a3   default-4cpu   default
default-4cpu-vm9ch   c6a.large   on-demand   us-west-2c   ip-192-168-64-158.us-west-2.compute.internal   True      6d19h   ami-05046f6e7a37d62b5   aws:///us-west-2c/i-0ab9cc564f36b1621   default-4cpu   default
```

The OD pricing in the console is `0.0765 USD per Hour`

### DIFF
--- a/pkg/state/cost/cost.go
+++ b/pkg/state/cost/cost.go
@@ -291,6 +291,9 @@ func (cc *ClusterCost) DeleteNodeClaim(ctx context.Context, nodeClaim *v1.NodeCl
 	cc.nodeClaimSet.Delete(client.ObjectKeyFromObject(nodeClaim).String())
 }
 
+// internalAddOffering updates the internal clusterCost state to include a new offering for a given nodepool.
+// It is used to increment the overall cost when a node joins the cluster. It is only called by UpdateNodeClaim
+// after that function has determined if a nodeclaim is new.
 func (cc *ClusterCost) internalAddOffering(ctx context.Context, np *v1.NodePool, instanceName, capacityType, zone string, firstTry bool) error {
 	_, exists := cc.npCostMap[np.UID]
 	if !exists {
@@ -336,6 +339,9 @@ func (cc *ClusterCost) internalAddOffering(ctx context.Context, np *v1.NodePool,
 	return nil
 }
 
+// internalRemoveOffering updates the internal clusterCost state to remove an existing offering for a given nodepool.
+// It is used to decrement the overall cost when a node leeaves the cluster. It is only called by DeleteNodeClaim
+// after that function has determined if a nodeclaim is already being accounted for.
 func (cc *ClusterCost) internalRemoveOffering(np *v1.NodePool, instanceName, capacityType, zone string) error {
 	npc, exists := cc.npCostMap[np.UID]
 	if !exists {


pkg/controllers/nodeoverlay/controller.go
```diff
@@ -115,6 +119,28 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 
 	c.instanceTypeStore.UpdateStore(temporaryStore)
 	c.clusterState.MarkUnconsolidated()
+
+	var errs error
+	for _, np := range nodePoolList.Items {
```
engedaam: Instead of pulling the instance types from nodePoolList, I would suggest `nodePoolToInstanceTypes` as that already contains the mapping for the instance types. Keep in my this is just the raw instance type data from the cloud provider.

pkg/controllers/metrics/cost/controller.go
```diff
@@ -0,0 +1,93 @@
...
+	"sigs.k8s.io/karpenter/pkg/metrics"
+)
+
+// These are alpha metrics, they may not stay. Do not rely on them.
+var (
+	ClusterCost opmetrics.GaugeMetric
+)
+
+// Initialize metrics at runtime
+func initializeMetrics() {
```
engedaam: Why initilze the metric this way? We typically decalre the metric as a `var`. This will just condence the metric initialization, and the metric will be declaered on dependency import

pkg/controllers/metrics/cost/controller.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Update cost metrics for each nodepool
+	for _, np := range nodepools.Items {
+		cost := c.clusterCost.GetNodepoolCost(&np)
+		ClusterCost.Set(cost, map[string]string{
+			metrics.NodePoolLabel: np.Name,
+		})
+	}
+
+	return reconciler.Result{RequeueAfter: time.Second * 30}, nil
```
engedaam: Consider reducing the 30-second wait time to 5 seconds since this operation has low computational overhead - this would provide more accurate pricing data without significant resource impact.

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,367 @@
...
+		cost, err := npCost.UpdateCost()
+		if err != nil {
+			return fmt.Errorf("failed to update cost for nodepool %q: %w", np.Name, err)
+		}
+		cc.npCostMap[np.Name].cost = cost
+	}
+	return nil
+}
+
+func (npc *NodePoolCost) UpdateCost() (float64, error) {
```
engedaam: nit. 

Can we make this fucntion private? It does not seem like we are using it anywhere else

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -115,6 +119,28 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
...
+		if err != nil {
+			errs = multierr.Append(errs, err)
+			continue
+		}
+		overlayedInstanceTypes, err := c.instanceTypeStore.ApplyAll(np.Name, instanceTypes)
+		if err != nil {
+			errs = multierr.Append(errs, err)
+			continue
+		}
+		err = c.clusterCost.UpdateOfferings(ctx, &np, overlayedInstanceTypes)
```
engedaam: The node overlay functionality, currently behind a feature flag, handles cluster cost updates based on cloud provider pricing. Without this feature enabled, we can't guarantee accurate cost tracking, especially for spot instances where prices change dynamically. Suggest separating the cluster cost update logic from this controller to ensure reliable pricing updates regardless of node overlay status. What's the strategy for handling price updates when the feature is disabled?

pkg/controllers/state/cost_test.go
```diff
@@ -0,0 +1,574 @@
...
+
+var _ = Describe("ClusterCost", func() {
+	var clusterCost *state.ClusterCost
+	var testNodePool *v1.NodePool
+	var testNodePool2 *v1.NodePool
+	var testInstanceType *cloudprovider.InstanceType
+	var spotOffering *cloudprovider.Offering
+	var onDemandOffering *cloudprovider.Offering
+	BeforeEach(func() {
+		//log.SetLogger(log.FromContext(ctx))
```
engedaam: nit. 

Drop?

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -49,23 +49,27 @@ type Controller struct {
...
 		cloudProvider:     cp,
 		instanceTypeStore: instanceTypeStore,
 		clusterState:      clusterState,
+		clusterCost:       clusterCost,
 	}
 }
 
 // Reconcile validates that all node overlays don't have conflicting requirements
+//
+//nolint:gocyclo
```
engedaam: Can we just pull out the pricing updates into a seprate function? I think that will make it easier to read the reconile function anyways

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,367 @@
...
+		np := &v1.NodePool{}
+		err := cc.client.Get(ctx, client.ObjectKey{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, np)
+		if err != nil {
+			log.FromContext(ctx).Error(fmt.Errorf("failed to get nodepool %q for nodeclaim %q: %w", nodeClaim.Labels[v1.NodePoolLabelKey], nodeClaim.Name, err), "failed to process nodeclaim for cost tracking")
+			CostTrackingErrorsTotal.Inc(map[string]string{
+				metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
+			})
+			return
+		}
+		cc.Lock()
```
engedaam: Considered locking inside of the `internalAddOffering`? since that is where the write is taking place

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,367 @@
...
+		log.FromContext(ctx).Error(fmt.Errorf("nodeclaim %q is missing required labels: %v", nodeClaim.Name, missingLabels), "failed to remove nodeclaim from cost tracking")
+		CostTrackingErrorsTotal.Inc(map[string]string{
+			metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
+		})
+		return
+	}
+
+	nodePoolName := nodeClaim.Labels[v1.NodePoolLabelKey]
+	np := &v1.NodePool{}
+	err := cc.client.Get(ctx, client.ObjectKey{Name: nodePoolName}, np)
```
engedaam, line 270: How would we handle NodePool deletion? Don't we need to remove those entires from cluster cost to avoid a memory leak?

engedaam (commented):
Looks good for the most part! My main worry is many around using the node overlay controller as the location for instance type pricing updates, since that is behind a feature flag

pkg/controllers/metrics/cost/controller.go
DerekFrank: We do this elsewhere https://github.com/kubernetes-sigs/karpenter/blob/9327a891f525d39f67890a8a624fddbbfa065f56/pkg/controllers/metrics/node/controller.go#L63

I don't know if it matters much either way, we can do it as a var instead if you prefer

pkg/controllers/nodeoverlay/controller.go
DerekFrank: We could, but I generally err against functional decomposition that doesn't get re-used. I'm not sure if moving it out makes it more readable, or if it just means you need to read two different functions

pkg/controllers/nodeoverlay/controller.go
DerekFrank: I don't want to rely on `nodePoolToInstanceTypes` unless we are comfortable saying that `validateAndUpdateInstanceTypeOverrides` will never modify that map. I don't think there's a significant computational cost to re-get the instance types from the provider

pkg/controllers/nodeoverlay/controller.go
DerekFrank: You raise a good point. I was thinking we can tie the clusterCost object into the CP itself to call updateInstanceTypes() when it updates pricing\instance type information, but I should have made that clear.

That would handle both nodeoverlay disablement as well as spot pricing updates.

pkg/controllers/state/cost.go
DerekFrank: Because `internalAddOffering` is recursive, handling locking inside it is messy. This solution was cleaner IMO

pkg/state/cost/cost.go
DerekFrank: Good point, we do want to delete nodepools that get deleted. I think the solve is to delete the npc objects when we notice that the node we are deleting is the last instance offering in that nodepool

pkg/controllers/metrics/cost/controller.go
engedaam: Yeah, I would prefer to use the `var` method. For the example you showed, we are a unique failure for us to use initialization method. Here is the context https://github.com/kubernetes-sigs/karpenter/pull/2127

pkg/controllers/nodeoverlay/controller.go
engedaam: Are you thinking of just using a decorator? I guess need to make sure that cloudprovider apply it, or else we would still see the same issue.

DerekFrank: I was going to just do it in the providers themselves, but a decorator might be cleaner

DerekFrank: On second thought, I don't think a decorator makes sense. I think it should be up to the cloud providers to call the function when they update the pricing. We don't want to do it on a decorator, because we don't want to make an O(n) call whenever we call GetInstanceTypes(), and we also don't want to have to build in behavior to tell when the pricing has changed.

pkg/controllers/metrics/cost/controller.go
DerekFrank: I'll cut it down to 10, I'm not sure we expect nodes to change much faster than that

pkg/controllers/nodeoverlay/controller.go
engedaam: > I don't want to rely on nodePoolToInstanceTypes unless we are comfortable saying that validateAndUpdateInstanceTypeOverrides will never modify that map.

`validateAndUpdateInstanceTypeOverrides` must never modify instance types since these are direct pointers from the cloud provider. Modifying these would corrupt both upstream and cloud provider data simultaneously. This would compromise Karpenter's state since all components, including the overlay controller, provider-supplied instance type data.

> I don't think there's a significant computational cost to re-get the instance types from the provider

We can't make assumptions about `GetInstanceTypes()` performance as it's entirely cloud provider dependent. A better approach would be to fetch the instance types once and maintain an immutable copy of the unmodified list.

pkg/controllers/nodeoverlay/controller.go
engedaam: I was thinking more along the lines of having a controller (metrics controller could work) that would be responsible for updating the price information periodically, that way we don't ask the cloud provider to ever need to think about adding this information

DerekFrank: To summarize our offline discussion, we have a couple different options:

1. Make a controller that updates pricing ~once and hour
2. Allow the pricing provider in the CPs to notify clusterCost when there is an update

Either is fine, @jmdeal which would you prefer?

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+		if !equal(oldIts, newIts) {
+			shouldUpdate = true
+			break
+		}
+	}
+
+	if shouldUpdate {
+		newNpItMap := make(map[string]map[string]*cloudprovider.InstanceType)
+		for _, np := range npl.Items {
+			newIts, err := c.cloudProvider.GetInstanceTypes(ctx, &np)
```
engedaam: Can we just make this call once? May create a map that tracks all the new nodePools

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+			shouldUpdate = true
+			break
+		}
+		newIts, err := c.cloudProvider.GetInstanceTypes(ctx, &np)
+		if err != nil {
+			return reconciler.Result{}, err
+		}
+		if !equal(oldIts, newIts) {
+			shouldUpdate = true
+			break
```
engedaam: Don't we want to make sure we are check all the nodePools that are different first instead of going and go over all of those nodePool only?

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	}
+
+	if shouldUpdate {
+		newNpItMap := make(map[string]map[string]*cloudprovider.InstanceType)
+		for _, np := range npl.Items {
```
engedaam: Why do we need to loop through the nodePools again? Can we not just use the loop from above and update the map that stores the instance type information?

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+		c.npItMap = newNpItMap
+
+	}
+
+	return reconciler.Result{RequeueAfter: 1 * time.Hour}, nil
+}
+
+func equal(oldIts map[string]*cloudprovider.InstanceType, newIts []*cloudprovider.InstanceType) bool {
+	for _, it := range newIts {
```
engedaam, line 97: How do you handle cases where the instance types or offering is not provided back from `GetInstanceTypes` in the new set of Instance types? Since that case should also trigger an update, but I'm not sure this will be handled here

pkg/controllers/controllers.go
```diff
@@ -90,7 +92,7 @@ func NewControllers(
 		informer.NewNodeController(kubeClient, cluster),
 		informer.NewPodController(kubeClient, cluster),
 		informer.NewNodePoolController(kubeClient, cloudProvider, cluster),
-		informer.NewNodeClaimController(kubeClient, cloudProvider, cluster),
+		informer.NewNodeClaimController(kubeClient, cloudProvider, cluster, clusterCost),
```
engedaam, line 96: Don't we need to also add the pricing informer here as well?

pkg/controllers/metrics/cost/controller.go
```diff
@@ -0,0 +1,87 @@
...
+
+// These are alpha metrics, they may not stay. Do not rely on them.
+var (
+	ClusterCost = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodePoolSubsystem,
+			Name:      "cost_total",
+			Help:      "Total cost of the nodepool",
```
engedaam: What's your though on how we communicate currency for this metric? What about units, since cloud providers can give us any unit they want to provide such as per hour or per min (Not sure if any do, but is just more to get your thoughts on it) ? Also, should we add a disclaimer here that this only factors in the cost of instance and not additional things like volumes or networking?

pkg/state/cost/cost.go
engedaam: Can you point me to where you made this change? Curious to see the solution

engedaam (commented):
Looking good!

pkg/controllers/state/informer/pricing.go
DerekFrank: Offerings for a given instancetype no longer existing is handled on L111, the function doesn't currently handle instance types no longer being returned. I can add that

pkg/controllers/metrics/cost/controller.go
DerekFrank: I think we leave it unit-less upstream. Karpenter doesn't care about units for price and I don't know why it should start for this change. We probably should add a disclaimer here, but the more important place is the docs IMO.

pkg/controllers/state/informer/pricing.go
DerekFrank: I'm going to collapse the three comments on this function into one thread (this one). 

We're discussing two different approaches here to the same problem. They both work, and I don't particularly see a major difference between them. Either way, we have a O(nodepool * instance types) function that is called infrequently.

My implementation goes a little something like:
```
Figure out if anything has changed, and if so, continue to next step
Construct a new map using the CP. If you get an error, requeue
Overwrite the old data with the new data, and trigger an update
```

You're suggesting an implementation that goes a little something like:

```
For each nodepool, take the following steps:
1. Get the new instance types.
2. Overwrite the old data with the new data. (check if they've changed first?)
3. Trigger an update
```

My question is what are you trying to optimize or what are you worried about with my approach?

engedaam (commented):
/lgtm

pkg/controllers/metrics/cluster/controller.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func NewController(client client.Client, clusterCost *state.ClusterCost) *Controller {
+	return &Controller{
+		client:      client,
+		clusterCost: clusterCost,
+	}
+}
+
+func (c *Controller) Reconcile(ctx context.Context) (reconciler.Result, error) {
```
jmdeal, line 66: We should make sure we delete the metric when the NodePool is deleted, otherwise this can result in our metric list growing unbounded. Given that this is NodePools and Nodes it probably won't blow up in practice, but unless there's a strong reason not to we should ensure we're garbage collecting.

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -115,6 +119,28 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 
 	c.instanceTypeStore.UpdateStore(temporaryStore)
 	c.clusterState.MarkUnconsolidated()
+
+	var errs error
+	for _, np := range nodePoolList.Items {
+		instanceTypes, exists := nodePoolToInstanceTypes[np.Name]
+		if !exists {
+			errs = multierr.Append(errs, fmt.Errorf("frinding instance types for nodepool: %s", np.Name))
```
jmdeal:
```suggestion
			errs = multierr.Append(errs, fmt.Errorf("finding instance types for nodepool, %s", np.Name))
```

pkg/state/cost/cost.go
jmdeal, line 1: nit: I'd move this to the top level `state` package rather than colocating the state and the controllers which update it.

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+// change or when cloud provider pricing information is refreshed.
+//
+// Returns an error if instance type information cannot be updated or if cost
+// recalculation fails.
+func (cc *ClusterCost) UpdateOfferings(ctx context.Context, np *v1.NodePool, instanceTypes []*cloudprovider.InstanceType) error {
+	cc.Lock()
+	defer cc.Unlock()
+	err := cc.internalUpdateOfferings(np, instanceTypes)
+	if err != nil {
+		return fmt.Errorf("failed to update offerings for nodepool %q: %w", np.Name, err)
```
jmdeal: The standard we chose for Karpenter was to use commas to delimit errors rather than colons:
```suggestion
		return fmt.Errorf("failed to update offerings for nodepool %q, %w", np.Name, err)
```

This applies for all wrapped errors, I won't leave a comment on every one. It looks like you were consistent with the use of colons in this PR.

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+		if exists {
+			if equal(oldIts, newIts) {
+				continue
+			}
+		}
+		err = c.clusterCost.UpdateOfferings(ctx, &np, newIts)
+		if err != nil {
+			errs = multierr.Append(errs, err)
+			continue
+		}
```
jmdeal: small style nits (inlining):
```suggestion
		if exists && equal(oldIts, newIts) {
			continue
		}
		if err := c.clusterCost.UpdateOfferings(ctx, &np, newIts); err != nil {
			errs = multierr.Append(errs, err)
			continue
		}
```

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+		newNpItMap[client.ObjectKeyFromObject(&np).String()] = lo.SliceToMap(newIts, func(it *cloudprovider.InstanceType) (string, *cloudprovider.InstanceType) {
+			return it.Name, it
+		})
+	}
+	if errs != nil {
+		return reconciler.Result{}, err
+	}
+	c.npItMap = newNpItMap
```
jmdeal, line 88: Do we need to do this atomic swap of the maps? Why can't we update each individual entry when we successfully update the offerings? I don't think this is too big of a deal, but since we have the equality check to deduplicate operations it seems like we should be able to make use of it for partial applications.

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+		if err != nil {
+			errs = multierr.Append(errs, err)
+			continue
+		}
+		newNpItMap[client.ObjectKeyFromObject(&np).String()] = lo.SliceToMap(newIts, func(it *cloudprovider.InstanceType) (string, *cloudprovider.InstanceType) {
+			return it.Name, it
+		})
+	}
+	if errs != nil {
+		return reconciler.Result{}, err
```
jmdeal: nit
```suggestion
		return reconciler.Result{}, fmt.Errorf("refreshing pricing info, %w", err)
```

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+//
+// The ClusterCost maintains real-time cost information by:
+// - Tracking NodeClaim additions and removals
+// - Managing instance type offerings and their prices
+// - Calculating aggregate costs per NodePool and cluster-wide
+//
+// All operations are thread-safe through internal locking mechanisms.
+type ClusterCost struct {
+	sync.RWMutex
+	npCostMap map[string]*NodePoolCost // nodepool.Name -> NodePoolCost
```
jmdeal: We shouldn't use resource names as a key in this map since it can introduce failures when resources are recreated. I'd use the NodePool UUID as the main key and if we need the name as well it can be stored alongside it. This applies anywhere we're using NodePool as an index.

(We definitely still use names elsewhere as indexes, but I'd rather defend against this going forward than punt until we refactor everything else)

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,370 @@
...
+// - Tracking NodeClaim additions and removals
+// - Managing instance type offerings and their prices
+// - Calculating aggregate costs per NodePool and cluster-wide
+//
+// All operations are thread-safe through internal locking mechanisms.
+type ClusterCost struct {
+	sync.RWMutex
+	npCostMap map[string]*NodePoolCost // nodepool.Name -> NodePoolCost
+	// nodeClaimMap tracks which NodeClaims are currently being monitored for cost
+	nodeClaimMap map[string]bool
```
jmdeal: nit: I'd use `sets.Set` if the intention is to use this as a set and not a tri-state tracker (e.g. differentiating between an element not existing and false).

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+
+	cloudProvider cloudprovider.CloudProvider
+	client        client.Client
+}
+
+// NodePoolCost represents the cost tracking information for a single NodePool.
+// It maintains the current cost, available instance types, and count of active offerings.
+type NodePoolCost struct {
+	cost                 float64
+	overlayedInstanceMap map[string]*cloudprovider.InstanceType // instance name -> instance
```
jmdeal: I'd consider just calling this `instanceTypeMap` since `overlayInstanceMap` suggests that it's only the instances types which have overlays applied. I don't feel too strongly about this, so let me know what you think. If you want to keep the overlay prefix though, I'd suggest `overlaidInstanceTypeMap` (spelling + instance vs instance type).

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+	return nil
+}
+
+func (cc *ClusterCost) internalUpdateOfferings(np *v1.NodePool, instanceTypes []*cloudprovider.InstanceType) error {
+	npCost, exists := cc.npCostMap[np.Name]
+
+	if !exists {
+		cc.createNewNodePoolCost(np, instanceTypes)
+	} else {
+		npCost.overlayedInstanceMap = lo.FilterSliceToMap(instanceTypes, func(it *cloudprovider.InstanceType) (string, *cloudprovider.InstanceType, bool) {
```
jmdeal: I'm indifferent on the filter, we don't have it elsewhere because it would be breaking an invariant to mix in a nil instance type. It doesn't hurt though.

edit: it looks like we're also not filtering in createNewNodePoolCost - I don't think it matters what we do but we should be consistent.

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+			return fmt.Errorf("failed to update cost for nodepool %q: %w", np.Name, err)
+		}
+		cc.npCostMap[np.Name].cost = cost
+	}
+	return nil
+}
+
+func (npc *NodePoolCost) updateCost() (float64, error) {
+	cost := 0.0
+	for ok, oc := range npc.offeringCounts {
```
jmdeal: nit: i'd not use `ok` here since that's the standard variable name in go for an existence check. It's like using `i` for something other than a for loop index.

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+	cost                 float64
+	overlayedInstanceMap map[string]*cloudprovider.InstanceType // instance name -> instance
+	// offeringCounts tracks how many instances of each offering type are currently active
+	offeringCounts map[OfferingKey]OfferingCount
+}
+
+// OfferingKey uniquely identifies a specific compute offering by its zone,
+// capacity type (e.g., spot/on-demand), and instance type name.
+type OfferingKey struct {
+	Zone, Capacity, InstanceName string
```
jmdeal: nit: I'd call this `CapacityType` to differentiate from the capacity of the offering (reserved offerings have a fixed capacity)

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+		// First lets check if the right labels are there
+		var missingLabels []string
+		for _, key := range NecessaryLabels {
+			_, exists := nodeClaim.Labels[key]
+			if !exists {
+				missingLabels = append(missingLabels, key)
+			}
+		}
+		if len(missingLabels) > 0 {
+			log.FromContext(ctx).Error(fmt.Errorf("nodeclaim %q is missing required labels: %v", nodeClaim.Name, missingLabels), "failed to process nodeclaim for cost tracking")
```
jmdeal: nit: prefer structured errors rather than embedding the value into the error message. 

```suggestion
			log.FromContext(ctx).Error(serrors.Wrap(fmt.Errorf("nodeclaim is missing required labels", "NodeClaim", klog.KObj(nodeClaim), "missing-labels", missingLabels)), "failed to process nodeclaim for cost tracking")
```

The same comment applies anywhere we're constructing an error with embedded values.

pkg/controllers/state/informer/nodeclaim.go
```diff
@@ -59,13 +61,15 @@ func (c *NodeClaimController) Reconcile(ctx context.Context, req reconcile.Reque
...
 			c.cluster.DeleteNodeClaim(req.Name)
+			c.clusterCost.DeleteNodeClaim(ctx, nodeClaim)
 		}
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
 	if !nodeclaimutils.IsManaged(nodeClaim, c.cloudProvider) {
 		return reconcile.Result{}, nil
 	}
 	c.cluster.UpdateNodeClaim(nodeClaim)
+	c.clusterCost.UpdateNodeClaim(ctx, nodeClaim)
```
jmdeal, line 73: We should make sure we're only processing launched NodeClaims here, otherwise we won't have resolved the concrete instance type.

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+}
+
+// UpdateNodeClaim adds a NodeClaim to cost tracking. The NodeClaim must have
+// all required labels or it will be ignored and logged as an error.
+func (cc *ClusterCost) UpdateNodeClaim(ctx context.Context, nodeClaim *v1.NodeClaim) {
+	cc.RLock()
+	_, exists := cc.nodeClaimMap[client.ObjectKeyFromObject(nodeClaim).String()]
+	cc.RUnlock()
+	if !exists {
```
jmdeal: nit: early return to reduce nesting
```suggestion
	if exists {
      return
	}
```

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+		if len(missingLabels) > 0 {
+			log.FromContext(ctx).Error(fmt.Errorf("nodeclaim %q is missing required labels: %v", nodeClaim.Name, missingLabels), "failed to process nodeclaim for cost tracking")
+			CostTrackingErrorsTotal.Inc(map[string]string{
+				metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
+			})
+			return
+		}
+
+		np := &v1.NodePool{}
+		err := cc.client.Get(ctx, client.ObjectKey{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, np)
```
jmdeal: If you use the owner reference you can get the UUID of the owning NodePool and use that for the lookup.

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,370 @@
...
+			return
+		}
+
+		np := &v1.NodePool{}
+		err := cc.client.Get(ctx, client.ObjectKey{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, np)
+		if err != nil {
+			log.FromContext(ctx).Error(fmt.Errorf("failed to get nodepool %q for nodeclaim %q: %w", nodeClaim.Labels[v1.NodePoolLabelKey], nodeClaim.Name, err), "failed to process nodeclaim for cost tracking")
+			CostTrackingErrorsTotal.Inc(map[string]string{
+				metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
+			})
```
jmdeal, line 208: Since we increment this whenever we return an error I'd consider evaluating this in a defer call:
```go
func UpdateNodeClaim(...) (err error) {
  defer func() {
    if err != nil {
      // Increment err metric
    }
  }
}
```

I have the same comment for DeleteNodeClaim

pkg/controllers/state/cost.go
```diff
@@ -0,0 +1,370 @@
...
+			missingLabels = append(missingLabels, key)
+		}
+	}
+	if len(missingLabels) > 0 {
+		log.FromContext(ctx).Error(fmt.Errorf("nodeclaim %q is missing required labels: %v", nodeClaim.Name, missingLabels), "failed to remove nodeclaim from cost tracking")
+		CostTrackingErrorsTotal.Inc(map[string]string{
+			metrics.NodePoolLabel: nodeClaim.Labels[v1.NodePoolLabelKey],
+		})
+		return
+	}
```
jmdeal: nit: I'd consider separating this out into a "validate nodeclaim" util since we use it in a few places. That would also help ensure the error messages never skew.

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,370 @@
...
+		log.FromContext(ctx).Error(fmt.Errorf("failed to remove offering for nodeclaim %q in nodepool %q: %w", nodeClaim.Name, np.Name, err), "failed to remove nodeclaim from cost tracking")
+		CostTrackingErrorsTotal.Inc(map[string]string{
+			metrics.NodePoolLabel: np.Name,
+		})
+		return
+	}
+	delete(cc.nodeClaimMap, client.ObjectKeyFromObject(nodeClaim).String())
+}
+
+func (cc *ClusterCost) internalAddOffering(ctx context.Context, np *v1.NodePool, instanceName, capacityType, zone string, firstTry bool) error {
```
jmdeal, line 297: The naming add and remove offering suggests that we're updating the tracked offering types / pricing rather than the NodeClaim count that matches an offering. I'd consider just calling these add / remove NodeClaim

pkg/controllers/state/informer/nodeclaim.go
DerekFrank: Isn't that handled in the cluster cost calculator itself? We check for the required labels including instance type

pkg/controllers/metrics/cluster/controller.go
```diff
@@ -0,0 +1,87 @@
...
+	// List all nodepools in the cluster
+	var nodepools v1.NodePoolList
+	if err := c.client.List(ctx, &nodepools); err != nil {
+		return reconciler.Result{}, err
+	}
+
+	// Update cost metrics for each nodepool
+	for _, np := range nodepools.Items {
+		cost := c.clusterCost.GetNodepoolCost(&np)
```
jmdeal, line 87: Is there a reason we want to poll the cluster cost state when updating this metric instead of just updating it each time state is updated? I think we still need this controller to handle garbage collection so it likely doesn't matter in practice, but since these are pull based rather than push based metrics the frequent updates should be fine.

pkg/state/cost/cost.go
DerekFrank: I think we actually are adding an offering to the overall cluster cost

pkg/controllers/state/informer/pricing.go
DerekFrank: I did this so we don't have to delete offerings that no longer become available. I thought it was cleanest to just recreate the map each time we update than try to maintain it. Especially because we're updating it single threaded

pkg/controllers/metrics/cluster/controller.go
DerekFrank: The clusterCost object is intended to surface a 'GetClusterCost()' function for decision tracking, since that will be a poll model I figured I would build the metrics controller on top of that model instead of having to build two models

pkg/state/cost/cost.go
DerekFrank: The UpdateNodeClaim and DeleteNodeClaim functions are adding and removing a nodeclaim, but the internal*Offering functions do not care about nodeclaims, they are on a offering level

pkg/controllers/state/informer/nodeclaim.go
jmdeal: You're right, it might be a little defensive to perform that filtering here. Personally I'd prefer it, but I won't block on it.

pkg/state/cost/cost.go
jmdeal: The distinction I'm drawing is that an offering is a "type" rather than a concrete instantiation. At least that's my mental model for offerings. Here we're adding a NodeClaim which adheres to the type definition of an offering - we're just parsing that type info higher up the stack. I don't feel too strongly about this, but it's probably worth a godoc comment just defining what this function does. Even though it's not in the public interface, it still helps when you're reasoning about what this function does when it's called elswhere.

jmdeal:
Neither of my comments are blocking, approving with a hold. If you want to go ahead and merge you can remove the hold.

/approve
/lgtm
/hold

pkg/state/cost/cost.go
DerekFrank: I can add a comment clarifying the purpose

pkg/controllers/state/informer/nodeclaim.go
DerekFrank: I tend to prefer having the functions be defensive and let anybody call them with the knowledge it may be a no-op. I think it leads to cleaner business logic code. I'm going to leave it

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+	} else {
+		oc.Count += 1
+	}
+	cc.npCostMap[np.UID].offeringCounts[offeringKey] = oc
+	cc.npCostMap[np.UID].cost += oc.Cost
+	return nil
+}
+
+// internalAddOffering updates the internal clusterCost state to remove an existing offering for a given nodepool.
```
jmdeal:
```suggestion
// internalRemoveOffering updates the internal clusterCost state to remove an existing offering for a given nodepool. 
```

PR #2586, opened 2025-10-23
fix(labels): simplify RestrictedLabel filtering and relax cel restrictions

maxcao13:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/1046

**Description**
~nodeClaim Labels are the source for core karpenter to sync Node Labels in a centralized fashion.
Some bootstrap userdata provider implementations also consume these labels and pass them through kubelet self setting.~

~That results in a coupling between a centralized and a kubelet self setting approach.
This coupling results in conflicting criteria for validations and degraded UX.
This PR removes the coupling.~

~As a consequence, Node Labels are not unncessarily restricted for the centralized sync anymore.
This better empowers administrators reducing the reliance on self setting kubelet and minimizing the risk of a Node steering privileged workloads to itself.~

~A subset of labels, excluding those disallowed by the node restriction admission, is now stored in json within the "karpenter.sh/node-restricted-labels" NodeClaim annotation. This enables bootstrap userdata providers to continue utilizing the labels if needed.~

After some discussion, this PR has been reworked to below description:

This PR does the following:

- Drop CEL restrictions for kubernetes.io and k8s.io domains
- Remove kubernetes.io and k8s.io from RestrictedLabelDomains
- Remove LabelDomainExceptions (no longer needed)
- Remove IsRestrictedNodeLabel function (use WellKnownLabels directly)
- Move non-well-known label resolution from NodeClaim launch to creation time
- Remove Requirements.Labels() function
  - Cloud providers can re-implement and filter additional domains if needed
  - No need to filter for kubelet syncability - Karpenter syncs all labels

Karpenter now syncs all labels to the Node directly, and no longer relies on kubelet to sync labels. Cloud providers can still continue to pass labels to kubelet if they wish. However providers will need to filter out labels that would be forbidden to pass to kubelet labels on the provider side, since this is now no longer done in the core code.

See the proposal at the bottom of https://github.com/kubernetes-sigs/karpenter/pull/2586#issuecomment-3747509240

### Note, this is a continuation of https://github.com/kubernetes-sigs/karpenter/pull/1908 which has gone stale. I've gotten permission from the original author and I am trying to push this PR forward. A lot of the discussion and review already happened in that PR.


maxcao13:
I'm assigning all the people assigned to the original PR.

/assign @elmiko
/assign @jmdeal
/assign @jonathan-innis

pkg/controllers/nodeclaim/lifecycle/launch.go
```diff
@@ -128,10 +130,19 @@ func PopulateNodeClaimDetails(nodeClaim, retrieved *v1.NodeClaim) *v1.NodeClaim
...
 		nodeClaim.Labels, // User-defined labels
 	)
-	nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, retrieved.Annotations)
+
+	// We store labels compliant with the node restriction admission as an annotation on the NodeClaim
+	// so that bootstrap provider implementation can use them for kubelet labels.
+	kubeletLabels := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).KubeletLabels()
+	json, err := json.Marshal(kubeletLabels)
+	if err != nil {
+		panic(err)
```
moko-poi: Is this panic intentional? I think returning an error would be better here, as a panic could crash the entire controller even if only one NodeClaim has an issue.

maxcao13: Thanks, fixed :-)

pkg/apis/v1/labels.go
```diff
@@ -135,13 +129,21 @@ var (
...
 func IsRestrictedLabel(key string) error {
 	if WellKnownLabels.Has(key) {
 		return nil
 	}
-	if IsRestrictedNodeLabel(key) {
-		return serrors.Wrap(fmt.Errorf("label is restricted; specify a well known label or a custom label that does not use a restricted domain"), "label", key, "well-known-labels", sets.List(WellKnownLabels), "restricted-labels", sets.List(RestrictedLabelDomains))
+	labelDomain := GetLabelDomain(key)
+	for restrictedLabelDomain := range RestrictedLabelDomains {
+		if labelDomain == restrictedLabelDomain || strings.HasSuffix(labelDomain, "."+restrictedLabelDomain) {
+			return fmt.Errorf("using label %s is not allowed as it might interfere with the internal provisioning logic; specify a well known label: %v, or a custom label that does not use a restricted domain: %v", key, sets.List(WellKnownLabels), sets.List(RestrictedLabelDomains))
```
jmdeal: We should continue to use the serrror formatting from the source, I think we migrated to using structured errors after the original PR was created.

pkg/controllers/nodeclaim/lifecycle/launch.go
```diff
@@ -123,20 +128,29 @@ func (l *Launch) launchNodeClaim(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+
+	// We store labels compliant with the node restriction admission as an annotation on the NodeClaim
+	// so that bootstrap provider implementation can use them for kubelet labels.
+	kubeletLabels := scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).KubeletLabels()
+	json, err := json.Marshal(kubeletLabels)
+	if err != nil {
+		return nil, err
+	}
+	kubeletLabelsAnnotation := map[string]string{apis.Group + "/node-restricted-labels": string(json)}
+	nodeClaim.Annotations = lo.Assign(nodeClaim.Annotations, retrieved.Annotations, kubeletLabelsAnnotation)
```
jmdeal: I'm not sure there's any reason to persist this to the annotation. The library functions to perform this filtering are available to cloud providers, if cloud providers choose to pass labels to kubelet they can apply this filtering themself. Unless there's a strong need, I'd rather not make this implementation detail visible to end-users.

This also is going to take place after the underlying instance has already been launched for the node, meaning the kubelet config was already generated. It wouldn't be useable by cloud providers at this stage. If we were to persist this info to the NodeClaim object directly, we'd want to do that when we create the NodeClaim in provisioning, but I don't think we want to anyway.

pkg/scheduling/requirements.go
```diff
@@ -267,13 +267,30 @@ func (r Requirements) Intersects(requirements Requirements) (errs error) {
 	return errs
 }
 
+// Labels filters out labels from the requirements that are not allowed to be synced centrally by Karpenter to the Node.
```
jmdeal: I wouldn't update `Labels` to handle this filtering implicitly. Either way I think we're making a breaking change:
- Not filtering implicitly means that if `Labels()` is used to determine the labels to pass to kubelet, updates would be required
- Filtering means that if `Labels` are used to determine scheduling requirements, it would return an incomplete set.

I'd opt to not filter and allow the cloud providers to perform the filtering explicitly. We should mark this PR as BREAKING though to make it clear to cloud providers that updates may be required.

moko-poi:
@maxcao13 
rebase please.

elmiko:
this is making sense to me, thanks @maxcao13 

/lgtm

jmdeal:
I'm just getting back from OOO and getting caught up, it looks like my previous review hadn't published so I'm publishing it now. That being said, I've spent a while digging into how we're currently using labels in the project, and I think we have a pretty big opportunity to simplify our approach to label management. Right now it's honestly a bit of a fractured mess, the result of organic growth, and I'd like to see us address that directly rather than adding to the complexity. Apologies for the long post, but I wanted to get the context I gathered written down and posted publicly.

TLDR: My take is that we don't need to do kubelet based label syncing, and if cloudproviders want to they can take on the filtering complexity. The value is extremely limited for the level of complexity it introduces. Given the criticality of labels being set accurately on the NodeClaim I don't want to overcomplicate that logic. If anything, it would be great if we can take this opportunity to simplify the existing logic.

---

## What are NodeClaim labels for?

NodeClaim labels serve two purpose: they are an interim scheduling representation while nodes are registering with the cluster, and they are an observability feature for users. For both of these purposes, we do not care about any of these restricted. In fact it's the opposite - it is critical that all labels that were used to compute scheduling decisions are present on the NodeClaim once it's launched.

## What is a restricted label?

Frankly, the concept of a "restricted label" is a bit of a mess at the moment. There are two sets of restricted labels which are used for slightly different purposes. The first set, restricted node labels, are used when determining which requirements we can't just select a random value from when selecting labels:
- Well-known labels
  - Examples: `node.kubernetes.io/instance-type`, `topology.kubernetes.io/zone`, `karpenter.k8s.aws/instance-family`, etc
- Labels which can't be applied by Kubelet due to domain restrictions
  - Domains: `kubernetes.io`, `k8s.io`
  - Exceptions were added for the following domains: `kops.k8s.io`, `node.kubernetes.io`, and `node-restriction.kubernetes.io`
    - Both `kops.k8s.io` and `node-restriction.kubernetes.io` are still restricted by kubelet, providers must perform additional filtering. The net result of this is that you can specify these labels on your Karpenter NodePool, but they will only be synced by Karpenter (not Kubelet)
- Labels with the `karpenter.sh` domain (or other restricted domains injected by providers like `karpenter.k8s.aws`)
  - This appears to be an artifact of using the `RestrictedLabelDomain` set for both kubelet restrictions and NodePool / NodeClaim validation where we want to reject any non-well-known labels with this domain. 
- `node.kubernetes.io/hostname` is also restricted

Standard "restricted labels" are.a subset of "restricted node labels" which just excludes well-known labels. This set is used for runtime validation of NodePool/Claim requirements and labels.

I'll have a proposal at the end to simplify this, but it's best to walk through how "restricted node labels" are actually used first.

## The lifecycle of NodeClaim labels

### NodeClaim Creation

This is the stage where Karpenter first applies NodeClaims to the cluster. These NodeClaims are the output of either a scheduling simulation or a static NodePool scale-up. In either case, the following labels are included on the NodeClaim (merged, listed from lowest to highest precedence):
- User-specified labels (NodePool's `metadata.labels`)
- NodePool name (`karpenter.sh/nodepool`)
- NodeClass name (Cloudprovider dependent, ex: `karpenter.k8s.aws/ec2nodeclass`)

While these are merged, both the NodePool and NodeClass names should be restricted by CEL and runtime validation, resulting in no conflicts in practice. All of these labels are essential to sync at this time:
- User-specified labels need to be synced from the NodeClaim to the Node at registration time. They need to be persisted to the NodeClaim when it's created, otherwise the NodePool could be edited between creation and registration.
- The NodePool and NodeClass labels are used for identification in metrics.

Another important property of these labels is that their values are known concretely. For a well-formed provider, there should be no scenario where the properties of the instance launched does not adhere to these labels. For example, if the label `karpenter.k8s.aws/instance-family: c8i` is applied, whichever instance type is selected must belong to this family. This is enforced by Karpenter's scheduler since each label is translated into a single-value requirement. There is **no** scenario where we should apply a label to a NodeClaim when it has not yet been concretely resolved.

**Note:** We don't include any other labels because they are assumed to be dependent on the instance type selected for the NodeClaim by the CloudProvider. Theoretically, we could also include any labels generated from requirements with a single possible value, but this isn't done today.

### NodeClaim Launch

This stage occurs after an instance is successfully launched for a NodeClaim. At this point we've resolved the concrete instance type and those details are returned back via the `cloudprovider.Create` call. The NodeClaim's labels are updated, merging the following values (again from lowest to highest precedence):
- Labels returned via the `cloudprovider.Create` call
- A randomly selected value from each requirement without a restricted key
- The labels which were applied at creation

The first set of labels should include all well-known labels (e.g. `topology.kuberentes.io/zone`, `node.kubernetes.io/instance-type`, etc). A Karpenter provider must be able to consistently provide all well-known labels for a given instance type given this is used 

The next set labels is a little more confusing - for any requirements that don't have a restricted key we will select a random value. What this really boils down to in practice is any non-well known label. The current "restricted node check" is more complex than it needs to be given that most of the labels it filters out can't exist due to CEL and runtime validation on requirements. In my opinion, it would also make sense to shift this selection to NodeClaim creation - we can resolve the value without input from the CloudProvider so we should resolve it to a concrete value as soon as possible.

The final set of labels are just the labels we already applied at creation.

## Node Registration

This is the stage where Kubelet has registered the node with the cluster and Karpenter marks the NodeClaim as registered. At this point, Karpenter also syncs the following labels:
- Existing node labels
- NodeClaim labels (set during launch)
- The `karpenter.sh/registered` label

## Why pass any labels to Kubelet?

The main reason to register labels via Kubelet rather than via Karpenter when the node joins the cluster is to prevent scheduling race conditions. The following race condition can occur if we do not have this behavior:
- Create a pod with a NodeAffinity term with a NotIn requirement on a **user-specified label** (any well-known labels aren't currently synced)
- The pod tolerates Karpenter's unregistered taint (or all taints, common for daemonsets)
- The pod schedules to the node before labels are synced by Karpenter

I don't think this use case has significant value. A single value being used to reject pods from a node is the purpose of a taint, and taints will continue to be synced by Kubelet (if the cloudprovider configures it). Additionally, the behavior today is inconsistent since only certain labels are synced (e.g. well-known labels are ignored). I'd prefer a consistent model where all labels are synced to the node via registration, rather than a subset.

## Proposal
- Simplify the current restricted labels concepts
  - For standard "restricted labels", this simplifies to to any label with the `karpenter.sh` domain (and any injected by the cloudprovider) and any individual restricted labels (currently just `kubernetes.io/hostname`)
  - For "restricted node labels", this just becomes an alias for well known labels. I would drop the alias and just use well known labels.
- Shift concrete label resolution for non-well-known labels to NodeClaim creation (also filter these out of requirements).
- Drop any considerations for kubelet "syncability". If a cloudprovider chooses to implement it, they can perform the filtering on their side. They likely will want to anyway (e.g. the AWS doesn't use as many labels as possible to increase the compability of launch templates across launches).

elmiko:
> My take is that we don't need to do kubelet based label syncing, and if cloudproviders want to they can take on the filtering complexity. The value is extremely limited for the level of complexity it introduces. Given the criticality of labels being set accurately on the NodeClaim I don't want to overcomplicate that logic. If anything, it would be great if we can take this opportunity to simplify the existing logic.

i tend to agree with this sentiment, i'm still reading through the comments. i'm all for making it simpler, but i want to make sure that any labels we choose to evaluate or persist to nodes outside of the kubelet will be allowed to be processed.

pkg/controllers/nodeclaim/lifecycle/launch.go
maxcao13: With your proposal to let the cloudprovider to do the filtering, yes we wouldn't need this. And thanks for catching the other part about this. I can remove this logic.

pkg/scheduling/requirements.go
maxcao13: I'm not so clear on this part, I don't necessarily want this to be breaking in that all cloud providers would have to explicitly filter out the restrictedlabeldomains, I feel like it would make more sense to keep that in the core. Am I missing something?

If `Labels()` is used in [core once during NodeClaimLaunch](https://github.com/kubernetes-sigs/karpenter/blob/bf49008c1f8ea4864841d523e2fe0e0702dcc80b/pkg/controllers/nodeclaim/lifecycle/launch.go#L131), and once potentially in each provider (in AWS provider I believe it's used [here to pass to the launchTemplate to be passe to the kubelet](https://github.com/aws/karpenter-provider-aws/blob/44c308875455fb50d011552cda0bc0b1c74d9aa5/pkg/providers/launchtemplate/launchtemplate.go#L145)), can we just make a new function in core for core to use which only filters out well-known labels (your new definition of `IsRestrictedNodeLabel`, but keep `Labels()` in the scheduling/requirements interface with it's original definition?
```go
func (r Requirements) Labels() map[string]string {
	labels := map[string]string{}
	for key, requirement := range r {
		if !v1.IsOriginalRestrictedNodeLabel(key) {
			if value := requirement.Any(); value != "" {
				labels[key] = value
			}
		}
	}
	return labels
}
```
In other words for the cloudprovider contract `Labels()`, would use something like `IsOriginalRestrictedNodeLabel` and this would continue to filter out restricted label domains + well known labels, and let through exceptions.

For the core usage, we make a new function, and we don't restrict anything except well-known labels (your new definition of `RestrictedNodeLabels`).

Or if I can try to make it a bit clearer:
1. we have a function that's consumable by cloudproviders (in this case it's `scheduling.Labels()`, and it's purpose is that it potentially will be fed to kubelet as `--kubelet-labels`. We know for sure that the kubelet will reject some domains, so we can just filter it once in core karpenter.
```go
// in karpenter-provider-aws/pkg/providers/launchtemplate/launchtemplate.go
func (p *DefaultProvider) EnsureAll(
...
// STILL called Labels()
scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).Labels(), // Inject single-value requirements into userData
		map[string]string{karpv1.CapacityTypeLabelKey: capacityType},
	), tags)
```
3. On the other hand, we have a separate function that is just to get all the labels from requirements during NodeClaimLaunch and we are going to sync these onto the incoming Node (which now include the restricted domains but NOT well-known labels since those are resolved from the `cloudProvider.create()` call).
```go
// in karpenter/pkg/controllers/nodeclaim/lifecycle/launch.go
...
// Called something like RequirementLabels now
	nodeClaim.Labels = lo.Assign(
		retrieved.Labels, // CloudProvider-resolved labels
		scheduling.NewNodeSelectorRequirementsWithMinValues(nodeClaim.Spec.Requirements...).RequirementLabels(), // Single-value requirement resolved labels
		nodeClaim.Labels, // User-defined labels
	)
```

jmdeal: I should have clarified, my mega comment superceeds these comments. I just published these for posterity. My current thought is that we should drop the `Labels` method outright. This really exists for one-purpose - selecting a random value for a non-well known label that's specified in requirements. I think we should shift that selection to NodeClaim creation and persist them to the metadata. Then we don't need this function in either the cloudprovider or the lifecycle controller. 

If we follow that proposal, we'll still need similar filtering logic but I don't think it should live on requirements. It's always been a bit of a misleading function since it obfuscated which labels were being included. You're original change does help by simplifying the filtering logic, but I think it's best to just scope it to the NodeClaim creation code.

> don't necessarily want this to be breaking in that all cloud providers would have to explicitly filter out the restrictedlabeldomains

My thought for this is that providers will no longer need to perform filtration, I can't link to a section of my [comment](https://github.com/kubernetes-sigs/karpenter/pull/2586), but the "Why pass any labels to Kubelet?" section states why.

jmdeal:
> I can re-work this PR to your proposal, but I want to make sure we are on the same page about keeping the relaxed CEL/runtime validation changes here.

Yeah, we're on the same page. The changes you've proposed to validation sounds good to me. We should continue to restrict the `karpenter.sh` domain (and any domains passed by cloudproviders), but we can open up the `kubernetes.io` and `k8s.io` domains (other than hostname).

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -76,6 +76,21 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 	return nct
 }
 
+// SingleValueRequirementLabels resolves non-well-known requirements to labels.
+// Well-known labels and restricted labels (e.g., hostname) are skipped.
```
elmiko: just out of curiosity, do we document this behavior where multiple labels would overwrite themselves?

i guess i'm wondering if users need to know this.

maxcao13: Good question, I don't know if that's really documented anywhere. It does seem like like if you specify this:
```
spec:
  template:
    metadata:
      labels:
        "label1": "2"
        "label2": "1"
```
Karpenter will always just pick the label key most furthest down the label list to propagate to the resulting Node labels, and NodeClaim labels + requirements, and this is on the main branch, so not a regression or anything.

elmiko: ok, sounds like this might not be a big concern for users. appreciate the discussion.

pkg/apis/v1/nodeclaim_validation_cel_test.go
jmdeal, line 1: Let's add negative tests for both NodePools and NodeClasses for `RestrictedLabels`. This should have been there in the first place, but it was previously covered by the restricted domains test since hostname is covered there.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
jmdeal: Quick clarification @elmiko, when you say multiple labels overwrite themselves do you mean a situation like this?

```yaml
spec:
  template:
    metadata:
      labels:
        test: foo
        test: bar
```

Since the type of `labels` on the CRD is an `object`, only the last key is persisted to the API server - Karpenter isn't doing any prioritization.

In @maxcao13's example, since the labels have different keys both would be persisted and Karpenter would apply both.

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -108,6 +123,9 @@ func (i *NodeClaimTemplate) ToNodeClaim() *v1.NodeClaim {
 		}
 	}
 
+	// Resolve non-well-known requirements to labels after pod requirements have been intersected
+	i.Labels = lo.Assign(i.Labels, SingleValueRequirementLabels(i.Requirements))
+
```
jmdeal: Let's add a bit more context about why we can add these labels here since it is a bit difficult to reason about in a vacuum:

```suggestion
	// We'll assign any labels with known, concrete values at NodeClaim creation time. This includes any labels from the
	// NodeClaimTemplate (since there's a single possible value), and any resolved values for custom labels in the
	// NodeClaimTemplate's requirements. The latter **cannot** be instance type dependent (like well-known labels) since
	// Karpenter can't reason about which label domains would belong to each instance type.
	i.Labels = lo.Assign(i.Labels, i.resolveCustomLabelsFromRequirements())
```

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
```diff
@@ -76,6 +76,21 @@ func NewNodeClaimTemplate(nodePool *v1.NodePool) *NodeClaimTemplate {
 	return nct
 }
 
+// SingleValueRequirementLabels resolves non-well-known requirements to labels.
+// Well-known labels and restricted labels (e.g., hostname) are skipped.
+func SingleValueRequirementLabels(r scheduling.Requirements) map[string]string {
```
jmdeal: nit: I'd decrease the scope of this, at least to the scheduling package but it makes sense to just make it a member function of the NodeClaimTemplate IMO.

```suggestion
// resolveCustomLabelsFromRequirements resolves the concrete values for user-defined labels from a NodeClaimTemplate's
// requirements.
func (i *NodeClaimTemplate) resolveCustomLabelsFromRequirements() map[string]string {
```

pkg/controllers/provisioning/suite_test.go
```diff
@@ -1466,6 +1466,26 @@ var _ = Describe("Provisioning", func() {
 				Expect(node.Labels).To(HaveKeyWithValue("subdomain."+domain+"/test", "test-value"))
 			}
 		})
+		It("should not label nodes with well known labels", func() {
```
jmdeal: This statement isn't strictly true today, since we **will** include well-known labels that are included in the NodePool's `spec.template.metadata.labels` field (which is legal). It's probably not very common for user's to actually do so, but any specified labels are essentially the same as the following requirement:
```yaml
key: $label_key
operator: In
values:
- $label_value
```

If we wanted to, we could include any well-known labels from requirements that had a single value as well since that's guaranteed to be the concrete value regardless of the instance type the cloudprovider provisions. I'm indifferent on if we want to do that - if we choose to I'd update this test to reflect that, otherwise I'd just update the name.
```suggestion
		It("should not label nodes with well-known labels from requirements", func() {
```

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
maxcao13: Oh sorry, I thought my keys in my example were the same, :facepalm:  but that's what I meant, yes thanks for the clarification.
```yaml
spec:
  template:
    metadata:
      labels:
        "label1": "2"
        "label1": "1"
```
It will persist the label with value `1` and not `2`.

pkg/controllers/provisioning/suite_test.go
maxcao13: I'll just rename the test. Don't want to add unnecessary complexity in the code, since we will resolve the value during nodeclaim launch anyways. Thanks!

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
elmiko: right, this was my thinking, but i think the danger of making a mistake when all the labels are in the same manifest is very low.

i was trying to figure out if there was a place where the user might have the same label on multiple manifests and then not realize that the last seen key would be the final value. i don't think it's a big concern though.

jmdeal:
I'm running some final sanity checking in the Karpenter provider, once I complete that I'll plan on approving: https://github.com/aws/karpenter-provider-aws/pull/8888

pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go
jmdeal: While I think it is a concern from a UX standpoint, it's not possible to apply that configuration to a cluster in practice - the extra fields would be dropped by the API server and that's consistent behavior across k8s. At that point it's more of a question about k8s behavior than Karpenter behavior. It is a different story when you're talking cross-field though. The one place I could see this being confusing is if you're adding the same labels to both `spec.template.metadata.labels` and `spec.template.spec.requirements`. For example:
```yaml
spec:
  template:
    metadata:
      custom-label: foo
    spec:
      requirements:
      - key: custom-label
        operator: In
        values: ['foo', 'bar']
```
The behavior today is that the requirements from the labels and requirements are ANDed together so the final requirement would be `custom-label IN foo`. We could consider adding validation to prevent this type of configuration, but I don't think we need to scope it into this PR.

jmdeal:
Looks good on my end 🚀 
/lgtm
/approve

diranged:
👏 🥳

PR #2594, opened 2025-10-27
test: setup CI for performance testing

jigisha620:
Fixes #N/A 

**Description**
Setup CI for performance testing -
1. Moved common test setup to a different workflow file
2. Added `karpenter perf` as the approval comment to trigger perf tests
3. Added a basic test for now to ensure that the workflow works

https://github.com/jigisha620/karpenter/actions/runs/18859537758/job/53814815799

Tested in my fork

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -55,25 +55,6 @@ jobs:
         kubectl get nodepools
         kubectl get pods -A
         kubectl describe nodes
-    # TEMPORARILY DISABLED: Karpenter KPI Analysis Package
-    # This package analyzes key performance indicators (KPIs) for integration tests.
-    # Currently non-functional due to changes in core Karpenter metrics.
-    # Disabled due to test flakiness until underlying metric changes are addressed.
-    # Reference: https://github.com/nathangeology/karpenter_evaluate/blob/main/main.py
-    # - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8
-    #   with:
-    #     repository: nathangeology/karpenter_evaluate
-    #     path: ./karpenter_eval/ # Installs to a folder in the Karpenter repo for the test
-    #     ref: "1130af927302e6913a4947952112f793eeafc564"
-    #     fetch-depth: 0
-    # - name: install KPI report dependencies
-    #   shell: bash
-    #   run: |
-    #     pip install pandas==2.2.2
-    #     pip install pyarrow==16.1.0 
-    #     pip install tabulate==0.9.0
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,96 @@
...
+        make apply-with-kind
+    - name: ping cluster
+      shell: bash
+      run: | 
+        sleep 15
+        kubectl get pods -n kube-system | grep karpenter 
+        kubectl get nodepools
+        kubectl get pods -A
+        kubectl describe nodes
+    # TEMPORARILY DISABLED: Karpenter KPI Analysis Package
```
DerekFrank: Can we remove this as its clear we don't have plans to fix it?

test/suites/performance/provisioning_test.go
```diff
@@ -0,0 +1,51 @@
...
+					},
+					ResourceRequirements: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceCPU: resource.MustParse("1"),
+						},
+					},
+				}})
+			env.ExpectCreated(deployment)
+			env.ExpectCreated(nodePool, nodeClass)
+			env.EventuallyExpectHealthyPodCountWithTimeout(10*time.Minute, labelSelector, replicas)
```
DerekFrank, line 48: So in this example, this is a test to provision a node for a pod, and we think it should take <10 minutes?

jigisha620: Honestly it shouldn't even taken 10 minutes but this is just a basic test to get started. The main point of this PR is to set-up the CI for performance testing. When we add more tests, we can update this as well. I don't mind updating it now either.

.github/workflows/e2e.yaml
jigisha620: Wasn't sure if we plan on fixing this. Since I was moving the reusable code to a different workflow, I wanted to ensure that I don't drop something that we intend to fix at some point.

DerekFrank: It'll still be in the commit history for the old version, I think its safe to remove it unless we want to fix it in the next couple months

test/suites/performance/provisioning_test.go
DerekFrank: Its fine at 10 minutes for now, I'm just confirming the test case we're starting with

jigisha620: Got it. Yes, that's the test case - provision a node for a pending pod.

.github/workflows/e2e.yaml
jigisha620: Makes sense, will drop for now.

PR #2602, opened 2025-10-30
test: add GetNodePoolCost func for perf testing

jigisha620:
Fixes #N/A 

**Description**
This PR adds a function to calculate nodepool cost. This can be used to measure performance. For example - when Karpenter performs consolidation, this function can be used to compare cost before and after consolidation.

Also added a step in the workflow for `ko` installation as this PR https://github.com/kubernetes-sigs/karpenter/pull/2673 accidentally dropped `ko` causing test failures.

Ran the tests in my fork - https://github.com/jigisha620/karpenter/actions/runs/19941766749/job/57181075613


test/pkg/environment/common/expectations.go
```diff
@@ -1262,3 +1262,55 @@ func (env *Environment) GetDaemonSetOverhead(np *v1.NodePool) corev1.ResourceLis
 		return p, true
 	})...)
 }
+
+// GetClusterCost calculates the cost of all nodes in the cluster by grouping nodes
+// with identical instance configurations (type, zone, OS, arch, capacity type) and matching
+// them against pricing data from env.InstanceTypes. Returns 0.0 if no nodes or pricing data exists.
+//
+//nolint:gocyclo
+func (env *Environment) GetClusterCost() float64 {
```
DerekFrank: I wonder if a metric scraping approach is more extensible. Karpenter should emit cost metrics soon, and if we want to set other performance SLAs we might want to do it via metrics scraping. What are your thoughts?

jigisha620: We could move to metrics once we have them, but until then I think this is the most straightforward way to get the cluster cost.

kwok/cloudprovider/helpers.go
```diff
@@ -63,21 +64,31 @@ type InstanceTypeOptions struct {
...
-			return nil, fmt.Errorf("could not read custom instance types file: %w", err)
+	paths := []string{lo.FromPtr(defaultRawInstanceTypesPath)}
+	if opts := options.FromContext(ctx); opts != nil {
+		paths = append(paths, opts.InstanceTypesFilePath)
+	}
+
+	for _, path := range paths {
+		if path != "" {
+			if data, err := os.ReadFile(path); err != nil {
+				return nil, fmt.Errorf("could not read instance types file %s: %w", path, err)
```
jmdeal: nit:
```suggestion
				return nil, serrors.Wrap(fmt.Errorf("could not read instance types file, %w", err), "file", path)
```

test/pkg/environment/common/environment.go
```diff
@@ -92,6 +95,15 @@ func NewEnvironment(t *testing.T) *Environment {
 	// Get the output dir if it's set
 	outputDir, _ := os.LookupEnv("OUTPUT_DIR")
 
+	instanceTypes, err := kwok.ConstructInstanceTypes(ctx)
+	if err != nil {
+		log.Default().Printf("failed constructing instance types: %v", err)
```
jmdeal: Shouldn't this also be a fatal error?
```suggestion
		log.Default().Printf("failed constructing instance types, %w", err)
```

test/pkg/environment/common/environment.go
```diff
@@ -92,6 +95,15 @@ func NewEnvironment(t *testing.T) *Environment {
 	// Get the output dir if it's set
 	outputDir, _ := os.LookupEnv("OUTPUT_DIR")
 
+	instanceTypes, err := kwok.ConstructInstanceTypes(ctx)
```
jmdeal: I don't like the idea of us taking a direct dependency on the kwok provider implementation inside our common testing package. If we were to have this type of util, I would constrain it to a dedicated kwok testing package so that kwok details don't leak into our common packages. This problem goes away though if we just wait to rely on the upstream metric.

test/pkg/environment/common/expectations.go
```diff
@@ -1262,3 +1262,56 @@ func (env *Environment) GetDaemonSetOverhead(np *v1.NodePool) corev1.ResourceLis
...
+	GinkgoHelper()
+
+	By("Calculating cluster cost")
+	nodes := &corev1.NodeList{}
+	Expect(env.Client.List(env, nodes, client.HasLabels{test.DiscoveryLabel})).To(Succeed())
```
jmdeal: If we go with this approach, we should scrape NodeClaims rather than nodes since they exist for the instance's entire lifetime wheras nodes only exist for a subset of it.

test/pkg/environment/common/expectations.go
```diff
@@ -1262,3 +1262,56 @@ func (env *Environment) GetDaemonSetOverhead(np *v1.NodePool) corev1.ResourceLis
...
+		CapacityType string
+	}
+
+	uniqueInstanceOffering := make(map[NodeKey]int)
+	for _, node := range nodes.Items {
+		key := NodeKey{
+			InstanceType: node.Labels["node.kubernetes.io/instance-type"],
+			Zone:         node.Labels["topology.kubernetes.io/zone"],
+			OS:           node.Labels["kubernetes.io/os"],
+			Arch:         node.Labels["kubernetes.io/arch"],
```
jmdeal: Why do we care about OS and arch?

kwok/cloudprovider/helpers.go
jigisha620: Updated the PR to use cost metrics instead, applicable for all the other comments.

test/suites/performance/provisioning_test.go
```diff
@@ -33,6 +34,10 @@ var _ = Describe("Performance", func() {
 			env.ExpectCreated(deployment)
 			env.ExpectCreated(nodePool, nodeClass)
 			env.EventuallyExpectHealthyPodCountWithTimeout(10*time.Minute, labelSelector, replicas)
+			Eventually(func(g Gomega) {
```
DerekFrank, line 37: Do we need to do the same for Consolidation?

jigisha620: We need to do this for all the other performance tests but I think we should do it in a different PR. I only meant to add the reusable function and an example of how it should be used.

jigisha620: @nathangeology, will you be able to update the perf tests to use the reusable function that I added?

PR #2606, opened 2025-10-30
test: lock unsynced time mutex during cluster reset for raceless testing

DerekFrank:
Fixes #N/A 

**Description**

Closes known race in static capacity tests: https://github.com/kubernetes-sigs/karpenter/actions/runs/18947767733/job/54103817138

make presubmit

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -571,6 +571,8 @@ func (c *Cluster) NodePoolResourcesFor(nodePoolName string) corev1.ResourceList 
 
 // Reset the cluster state for unit testing
 func (c *Cluster) Reset() {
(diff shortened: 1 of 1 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -571,6 +571,8 @@ func (c *Cluster) NodePoolResourcesFor(nodePoolName string) corev1.ResourceList
 
 // Reset the cluster state for unit testing
 func (c *Cluster) Reset() {
+	c.unsyncedTimeMu.Lock()
+	defer c.unsyncedTimeMu.Unlock()
```
sumukha-radhakrishna, line 575: We should not be holding the cluster lock and calling Reset so it shouldnt be a problem

PR #2609, opened 2025-10-31
test: Added perf-test e2e benchmark tests

nathangeology:
**Description**
Added 5 performance benchmark e2e tests to the 'performance' test suite. These test run via Ginkgo and gomega against a KIND + KWOK cluster in the test environment. Currently the results of the test are simply logged out by ginko and written out locally into a text file. These are not yet collected anywhere, but will work to change this. The goal is to track these benchmarks going forward. Will add two more benchmark e2e tests next week. 

**Test 1 Simple scale to 1000, then to 700. :**
Called 'large_scale_test' right now, will rename this.
Scale two deployments to 500 pods each (1000 total) and then down to 350 each (700 total). Config as follows:
```
smallPodResources := corev1.ResourceRequirements{
		Requests: corev1.ResourceList{
			corev1.ResourceCPU:    resource.MustParse("950m"),   // 0.95 vCPU
			corev1.ResourceMemory: resource.MustParse("3900Mi"), // 3900 MB
		},
	}

	// Define resource requirements for large pods (500 pods)
	// 3.8 vCPU and 31 GB memory each
	largePodResources := corev1.ResourceRequirements{
		Requests: corev1.ResourceList{
			corev1.ResourceCPU:    resource.MustParse("3800m"), // 3.8 vCPU
			corev1.ResourceMemory: resource.MustParse("31Gi"),  // 31 GB
		},
	}
```
This test measures the time to scale out and have the pods ready and the resulting avg utilization of CPU and memory. After this, the test scales in to 70% of max. The time to scale-in the capacity and the resulting utilization is measured. 

**Test 2: Same as test 1, but the 1st deployment has Host Name topology spreading enabled. Same measurements and scaling pattern. Called "Host Name Spreading" right now.**

**Test 3. Same as test 2, but the pod count is doubled.  Called "Host Name Spreading XL" right now.**

**Test 4. Provisions capacity for 600 pods in 2 deployments and then adds a annotation to the node pool to trigger drift. The time to drift and replace all nodes is measured as well as the difference in resource utilization. Called "Drift Performance" right now**

**Test 5. Scales to 1000 pods similiar to tests 1 and 2, but has 30 deployments instead of 2. Scales in to 700 after the scale out test. Measures time to scale and utilization. Called "Wide Deployments".**

I ran the tests in Github actions. All tests passing as of writing, will be adding two more tests and will verify they work as well. 


jamesmthompson:
> The time to scale-in the capacity and the resulting utilization is measured.

I didn't fully understand this. Is some kind of decision made with the measurement? Is that decision a pass/fail? If so, what's the difference between pass and a fail, and how did we arrive at particular thresholds for making decisions?

nathangeology:
> > The time to scale-in the capacity and the resulting utilization is measured.
> 
> I didn't fully understand this. Is some kind of decision made with the measurement? Is that decision a pass/fail? If so, what's the difference between pass and a fail, and how did we arrive at particular thresholds for making decisions?

We will eventually establish some baseline criteria (e.g. with some deploment config, we should be able to scale to 1000 pods in under 1 minute and 30 seconds). For now we just want to collect the results to establish a good baseline and avoid having a flaky set of tests.

test/pkg/environment/common/cleanup.go
```diff
@@ -0,0 +1,270 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/test"
+)
+
+// CleanupOptions configures cleanup behavior for test resources
+type CleanupOptions struct {
```
DerekFrank: Do we need this when we have env.Cleanup()? https://github.com/kubernetes-sigs/karpenter/blob/9e54967f712c7b4aee3a223f05c5472e708b9583/test/suites/regression/suite_test.go#L84

test/suites/performance/basic_test.go
```diff
@@ -0,0 +1,468 @@
...
+
+	"sigs.k8s.io/karpenter/pkg/test"
+	"sigs.k8s.io/karpenter/test/pkg/environment/common"
+)
+
+var _ = Describe("Performance", func() {
+	Context("Basic Deployment", func() {
+		It("should efficiently scale two deployments with different resource profiles", func() {
+			By("Setting up performance test with 1000 pods across two resource profiles")
+			GinkgoWriter.Printf("\n" + strings.Repeat("=", 70) + "\n")
```
DerekFrank: I don't think we usually have printfs in our test logs. If you want to store some information, you should write it to a file like we do for the scheduling benchmarking test and that can be accessed via automation

test/suites/performance/basic_test.go
```diff
@@ -0,0 +1,468 @@
...
+				"Should not require more than 50 nodes for 1000 pods")
+
+			Expect(avgCPUUtil).To(BeNumerically(">", 0.4),
+				"Average CPU utilization should be greater than 70%")
+
+			Expect(avgMemUtil).To(BeNumerically(">", 0.4),
+				"Average memory utilization should be greater than 70%")
+
+			// Verify all pods are actually running
+			env.EventuallyExpectHealthyPodCount(smallPodSelector, 500)
```
DerekFrank: I think this is a duplicate of the above statement prior to report generation

test/suites/performance/basic_test.go
```diff
@@ -0,0 +1,468 @@
...
+				GinkgoWriter.Printf("   All performance targets met successfully!\n")
+			} else {
+				GinkgoWriter.Printf("\n⚠️  PERFORMANCE TEST: ATTENTION NEEDED\n")
+				for _, warning := range report.Warnings {
+					GinkgoWriter.Printf("   • %s\n", warning)
+				}
+			}
+
+			// Add scale-in metrics to the final report if scale-in was performed
+			if report.ScaleInEnabled {
```
DerekFrank: I think we could refactor this to be much more ergonomic by cutting out the printf statements. We could also re-use a lot of this skeleton across all of the tests since they have the same basic setup

DerekFrank (changes requested):
Have some high level feedback before I go through each individual test

test/pkg/environment/common/cleanup.go
nathangeology: Fixed this

test/suites/performance/basic_test.go
nathangeology: Got most of these out, I might need to look through actions.go however

test/suites/performance/basic_test.go
nathangeology: Did a big refactor today, we should look through it soon and see what you think

nathangeology:
Example performance report for the 'host-name-spreading' test
[performance-results-Host Name Spreading-19090412420.zip](https://github.com/user-attachments/files/23371550/performance-results-Host.Name.Spreading-19090412420.zip)

test/suites/performance/actions.go
```diff
@@ -0,0 +1,951 @@
...
+	Labels                    map[string]string
+	Annotations               map[string]string
+	TopologySpreadConstraints []corev1.TopologySpreadConstraint
+	PodAntiAffinity           *corev1.PodAntiAffinity
+	NodeAffinity              *corev1.NodeAffinity
+	deployment                *appsv1.Deployment // Store created deployment for cleanup
+}
+
+// NewCreateDeploymentAction creates a new deployment action
+func NewCreateDeploymentAction(name string, replicas int32, profile ResourceProfile) *CreateDeploymentAction {
```
DerekFrank: I think we can borrow the implementation of https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/test/deployment.go#L36 to create the deployments

test/suites/performance/actions.go
```diff
@@ -0,0 +1,951 @@
...
+	}
+}
+
+// SetNodePool sets the nodePool reference (called by the execution engine)
+func (a *TriggerDriftAction) SetNodePool(np interface{}) {
+	a.nodePool = np
+}
+
+// Execute triggers the drift scenario
+func (a *TriggerDriftAction) Execute(env *common.Environment) error {
```
DerekFrank: Yea alright I think I liked it the previous way better :( 

The amount of error handling and switching on these Execute actions makes them not super worth functionally decomposing

test/suites/performance/do_no_disrupt.go
```diff
@@ -0,0 +1,126 @@
...
+			})
+			doNotDisruptPods := env.Monitor.RunningPods(doNotDisruptPodSelector)
+			nodesWithProtectedPods := make(map[string]bool)
+			for _, pod := range doNotDisruptPods {
+				if pod.Spec.NodeName != "" {
+					nodesWithProtectedPods[pod.Spec.NodeName] = true
+				}
+			}
+
+			GinkgoWriter.Printf("Nodes with do-not-disrupt pods: %d\n", len(nodesWithProtectedPods))
```
DerekFrank: nit: lets not have any `Printf`s. These should be the error messages when we check values for their correctness

test/suites/performance/actions.go
```diff
@@ -0,0 +1,951 @@
...
+		TotalReservedMemoryUtil: avgMemUtil,
+		ResourceEfficiencyScore: resourceEfficiencyScore,
+		PodsPerNode:             podsPerNode,
+		Rounds:                  len(consolidationRounds),
+		Timestamp:               time.Now(),
+	}, nil
+}
+
+// MonitorDrift monitors drift operations with replacement rounds
+func MonitorDrift(env *common.Environment, expectedPods int, timeout time.Duration) (*PerformanceReport, error) {
```
DerekFrank: Can we phrase these functions in the 'Expect' syntax? Or does that make it much uglier

test/suites/performance/actions.go
```diff
@@ -0,0 +1,951 @@
...
+}
+
+// SetNodeAffinity sets node affinity on the action
+func (a *CreateDeploymentAction) SetNodeAffinity(nodeAffinity *corev1.NodeAffinity) *CreateDeploymentAction {
+	a.NodeAffinity = nodeAffinity
+	return a
+}
+
+// Execute creates the deployment
+func (a *CreateDeploymentAction) Execute(env *common.Environment) error {
```
DerekFrank: I think I was wrong before, refactoring these into actions has complicated the code more significantly than I thought it would.

test/suites/performance/actions.go
```diff
@@ -0,0 +1,951 @@
...
+		PodsPerNode:             podsPerNode,
+		Rounds:                  driftRounds,
+		Timestamp:               time.Now(),
+	}, nil
+}
+
+// getDeploymentNames returns a list of deployment names for debugging
+func getDeploymentNames(deploymentMap map[string]*appsv1.Deployment) []string {
+	names := make([]string, 0, len(deploymentMap))
+	for name := range deploymentMap {
```
DerekFrank: You should be able to use lo.Keys() or lo.Map() to get this slice

nathangeology: With the refactor, this is moot. :)

test/suites/performance/actions.go
nathangeology: Per our discussion yesterday, I was able to refactor things a bit and it seems a lot more simple now.

test/suites/performance/actions.go
nathangeology: This is now moved to be part of the report struct and we can use expects to check for deployment creation and then to check output conditions in the report.

test/suites/performance/do_no_disrupt.go
nathangeology: I think I got these out, I'll check again to see if I missed any. The only printfs left are during the report file writing, it dumps the json to log as a backup to file write. We can do something different there at some point if we want.

test/suites/performance/actions.go
nathangeology: I switched to Test.Deployment, seems like it has all we need. The only thing I added is the DeploymentOptions "templates" file. This seems to really cut down on the deployment options boiler plate at the top of the tests.

test/pkg/environment/common/deployments_templates.go
```diff
@@ -0,0 +1,235 @@
...
+	// Apply all modifiers
+	for _, opt := range opts {
+		opt(&deploymentOpts)
+	}
+
+	return deploymentOpts
+}
+
+// WithLabels adds or merges labels to the pod template
+func WithLabels(labels map[string]string) DeploymentOptionModifier {
```
DerekFrank: I do like this format, its a builder pattern. We could just refactor the original test package to have a builder pattern and include all of this in there instead of making a new file. I think that will be cleaner.

test/suites/regression/termination_test.go
```diff
@@ -273,56 +268,41 @@ var _ = Describe("Termination", func() {
...
-				},
-				TerminationGracePeriodSeconds: lo.ToPtr(int64(60)),
-				Image:                         "alpine:3.20.2",
-				Command:                       []string{"/bin/sh", "-c", "sleep 1000"},
-				PreStopSleep:                  lo.ToPtr(int64(60)),
-				PriorityClassName:             "system-cluster-critical",
-				ResourceRequirements:          corev1.ResourceRequirements{Limits: corev1.ResourceList{corev1.ResourceMemory: resource.MustParse("1Gi")}},
-			},
-		})
+		deployment := test.Deployment(test.CreateDeploymentOptions("deployment", 1, "100m", "128Mi",
```
DerekFrank: I'm not sure if this one needs to be refactored, the complicated deployments still make sense to do with overrides I think

nathangeology: Yeah, it certainly didn't make it less verbose haha

nathangeology: Ok I reverted this part of the refactor.

DerekFrank (approved):
/lgtm

jmdeal:
/approve

Just need to address the merge conflict

Issue #2571, opened 2025-10-13
Support K8s Buffer Api


DerekFrank:
### Description

**What problem are you trying to solve?**

SIG Autoscaling has been working on an API for users to declare a buffer\headroom: https://github.com/kubernetes/autoscaler/pull/8151

This would solve the following problems for Karpenter users:
* Kueue Integration
* Scheduling latency improvements for sensitive applications
* pod headroom

Karpenter should support this Buffer API by accommodating it in its scheduling and consolidation algorithms.

**How important is this feature to you?**

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

moko-poi:
I've created an RFC for this feature: https://github.com/kubernetes-sigs/karpenter/pull/2611
The proposal uses virtual pods to integrate Buffer API with Karpenter's scheduling and consolidation.

Zhenshan-Jin:
hi team, curious is there a timeline about this integration ? Thank you

jicowan:
The `active-capacity` strategy from the buffer API works well for steady-state headroom, but it
doesn't serve **gang / batch workloads** — the use case this issue lists first ("Kueue
Integration"). An `active-capacity` buffer sized for a gang job never stops: after the gang runs
and completes, it refills and keeps provisioning capacity for a job that already ran. There's no
way to say "provision X once, for this workload, then stop."

The CapacityBuffer design doc already scopes the fix as future work:
> "Ephemeral capacity strategy: One-time capacity request pattern for batch systems like Kueue
> (deferred to future work)" … "ephemeral strategy provides deterministic one-time capacity with
> completion semantics."

**Use case:** a user submits an 8-pod all-or-nothing training Job via Kueue. They want nodes
pre-warmed so the gang schedules immediately on admission (no reactive cold-start), but they do
**not** want standing capacity after the run. Same shape for Volcano/YuniKorn PodGroups and
periodic batch jobs.

**Sketch of the approach** (details for an RFC, not here): a new
`provisioningStrategy: buffer.x-k8s.io/ephemeral-capacity` that provisions once, detects when the
[...]

PR #2611, opened 2025-11-03
docs: RFC for Capacity Buffer API Support

moko-poi:
## Description
  This is a proposal to add support for the standard Kubernetes CapacityBuffer API (autoscaling.x-k8s.io/v1alpha1) to enable pre-provisioned spare capacity in
  Karpenter clusters.

  The RFC introduces a virtual pod approach that integrates buffer capacity into Karpenter's scheduling and consolidation algorithms while maintaining compatibility
  with the existing Cluster Autoscaler Buffer API.

  **Related issue:** kubernetes-sigs/karpenter#2571

  RFC only - implementation will follow in subsequent PRs

  ## Key Features
  - Standard CapacityBuffer CRD support
  - Virtual pod generation for buffer capacity
  - Integration with Karpenter's NodeClaim-based architecture
  - Consolidation protection for buffer capacity
  - Cross-autoscaler compatibility (Cluster Autoscaler → Karpenter migration)

### DIFF
--- a/designs/capacity-buffer.md
+++ b/designs/capacity-buffer.md
@@ -53,20 +53,6 @@ Extend Karpenter to support the standard CapacityBuffer API by treating buffer c
 
 **Goal:** Karpenter respects configured CapacityBuffers, maintaining additional capacity as if they were pods scheduled in the cluster.
 
-### Modeling & Validation
-
-When a `CapacityBuffer` is created, Karpenter generates virtual pods that participate in scheduling decisions but are never actually created in the cluster. These virtual pods:
-
-- Are derived from either `scalableRef` (existing workload) or `podTemplateRef` (custom template)
-- Include special labels for identification: `karpenter.sh/capacity-type: buffer`
-- Have lower priority than real workloads to ensure user pods take precedence
-- Are considered during both provisioning and consolidation decisions
-
-Buffer capacity persists until:
-- The CapacityBuffer resource is deleted
-- Real workloads consume the buffered capacity
-- Consolidation occurs that preserves equivalent buffer capacity elsewhere
-
 ### API Specification
 
 Karpenter will support the standard CapacityBuffer API:
@@ -127,24 +113,46 @@ spec:
 3. **Field types**: `replicas` and `percentage` are `int32` (not `int`)
 4. **Minimum values**: Both `replicas` and `percentage` have `+kubebuilder:validation:Minimum=0`
 
-### Virtual Pod Generation
+### Virtual Pod Generation Workflows
 
-Following Cluster Autoscaler's pattern:
+The buffer controller generates virtual pods in response to these events:
 
-1. **ScalableRef Resolution**: Query the referenced workload's pod template
-2. **PodTemplateRef Resolution**: Directly use the referenced template
-3. **Replica Calculation**:
-   - If `percentage` is set: calculate `percentage * currentReplicas / 100` (integer division)
-   - If `replicas` is set: use the specified value
-   - If both are set: use `min(percentageResult, replicas)` - see Design Decisions
-4. **Virtual Pod Creation**: Generate N in-memory pods with special labels:
-   ```yaml
-   labels:
-     karpenter.sh/capacity-type: buffer
-     autoscaling.x-k8s.io/buffer: <buffer-name>
-     autoscaling.x-k8s.io/buffer-namespace: <buffer-namespace>
-   ```
-5. **Priority Assignment**: Use system priority class or lowest priority to ensure user pods take precedence
+**1. Initial CapacityBuffer Creation**
+- User creates CapacityBuffer resource
+- Controller validates and translates spec
+- Generates initial set of virtual pods
+- Updates status to ReadyForProvisioning
+
+**2. CapacityBuffer Spec Update**
+- User modifies replicas, percentage, or references
+- Controller detects spec change
+- Regenerates virtual pods with new configuration
+- Updates status with new replica count
+
+**3. Referenced Workload Scale (for percentage-based buffers)**
+- Referenced deployment/statefulset scales
+- Controller detects scale event
+- Recalculates buffer size based on new percentage
+- Regenerates virtual pods to match
+
+**4. PodTemplate Generation Change**
+- Referenced workload updates pod template
+- Controller detects template change
+- Regenerates virtual pods with new template
+- Updates podTemplateGeneration in status
+
+**Virtual Pod Properties:**
+```yaml
+metadata:
+  labels:
+    karpenter.sh/capacity-type: buffer
+    autoscaling.x-k8s.io/buffer: <buffer-name>
+    autoscaling.x-k8s.io/buffer-namespace: <buffer-namespace>
+spec:
+  # Derived from scalableRef or podTemplateRef
+  # No priority modification - Karpenter provisions for all pods equally
+  # kube-scheduler handles actual pod placement based on priority
+```
 
 ## Design Decisions
 
@@ -243,26 +251,63 @@ spec:
 
 ## Implementation Details
 
+### Behavior Model
+
+Karpenter maintains buffer capacity using virtual pods that participate in scheduling decisions:
+
+**Virtual Pod Lifecycle:**
+1. **Creation**: When a CapacityBuffer is created, generate virtual pods representing the buffer
+2. **Continuous Maintenance**: As real pods are scheduled, maintain the buffer by keeping virtual pods in the scheduling pipeline
+3. **Updates**: When CapacityBuffer spec changes (replicas, percentage), regenerate virtual pods
+4. **Deletion**: When CapacityBuffer is deleted, remove virtual pods from consideration
+
+**Key Behaviors:**
+- Virtual pods are derived from `scalableRef` (existing workload) or `podTemplateRef` (custom template)
+- Virtual pods include labels for identification: `karpenter.sh/capacity-type: buffer`
+- Virtual pods participate in provisioning (trigger NodeClaim creation) and consolidation (prevent capacity removal)
+- Buffer capacity is maintained continuously - Karpenter provisions additional capacity when real pods consume buffer capacity
+
+**Example Flow:**
+```
+1. User creates CapacityBuffer with replicas: 5
+2. Karpenter generates 5 virtual buffer pods
+3. Provisioner includes these pods in scheduling → creates NodeClaims
+4. Real pods are scheduled on this capacity
+5. Virtual buffer pods remain in scheduling queue → trigger more provisioning
+6. Buffer is maintained as long as CapacityBuffer exists
+```
+
 ### Controller Architecture
-
-We will implement capacity buffer support by adding a new controller and integrating with existing Karpenter components. 
-
-**⚠️ Critical Integration Considerations**: This implementation must account for Karpenter's unique architecture that operates at the NodeClaim level rather than individual pod scheduling, and integrates deeply with Karpenter's state management system.
 
 #### Buffer Controller (New)
 **Location**: `pkg/controllers/capacitybuffer/`
 
-**Purpose**: 
-- Watch CapacityBuffer CRDs and translate them into virtual pods
-- Maintain an in-memory cache of buffer pods for scheduling integration
-- Handle buffer lifecycle and status updates
-
-**Key Components**:
+**Controller API:**
 ```go
-type Controller struct {
+type BufferController interface {
+    // GetVirtualPods returns all active buffer pods for scheduling
+    GetVirtualPods(ctx context.Context) ([]*corev1.Pod, error)
+    
+    // GetDisruptionCost returns cost for disrupting capacity on a node
+    // Buffer virtual pods have lower disruption cost than real workloads
+    GetDisruptionCost(ctx context.Context, node *corev1.Node) (float64, error)
+    
+    // Run starts the reconciliation loop
+    Run(ctx context.Context) error
+}
+```
+
+**Responsibilities:**
+- Watch CapacityBuffer CRDs and translate them into virtual pods
+- Maintain virtual pods in memory for scheduling integration
+- Handle buffer lifecycle and status updates
+- Provide virtual pods to provisioner and disruption controllers
+
+**Internal Components**:
+```go
+type bufferController struct {
     kubeClient      client.Client
-    cluster         *state.Cluster        // Karpenter's cluster state management
-    provisioner     *provisioning.Provisioner
+    bufferClient    *capacitybuffer.Client
     
     // Buffer-specific components
     bufferCache     map[types.NamespacedName]*BufferState
@@ -274,23 +319,18 @@ type Controller struct {
 type BufferState struct {
     Buffer          *v1alpha1.CapacityBuffer
     VirtualPods     []*corev1.Pod
-    NodeClaims      []*scheduler.NodeClaim  // Track at NodeClaim level
-    LastScheduled   time.Time
+    LastUpdate      time.Time
 }
 
 type BufferTranslator struct {
     // Combines multiple translators:
     // - PodTemplateTranslator for podTemplateRef
     // - ScalableObjectsTranslator for scalableRef
-    // - ResourceLimitsTranslator for limits-based sizing
 }
 ```
 
 #### Scheduling Integration
 **Location**: `pkg/controllers/provisioning/`
-
-**Critical Architecture Consideration**: 
-Karpenter's scheduling operates at the `NodeClaim` level, not individual pods. The `provisioner.Schedule()` method creates `NodeClaim` objects through complex constraint solving.
 
 **Correct Integration Approach**:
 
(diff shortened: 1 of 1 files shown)


designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
+# Capacity Buffer API Support
```
DerekFrank, line 1: Thanks for the design! Couple high level notes:
* I don't think the capacity buffer api exists yet. The sig-autoscaling RFC was merged but I don't think the api itself has been released. Doesn't mean we can't think ahead, but implementation for this will have to wait until the api exists.
* Overall I think the doc is a bit messy. I think it would be a stronger proposal if you started from the CX and derived implementation from that. Similarly, I think the implementation section could be much stronger if it started from the requirements of the existing controllers and worked backwards to the API's that the capacity buffer controller should provide.

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+2. **Burst workloads** that need immediate scheduling for CI/CD, batch jobs, or event-driven applications
+3. **High-availability services** that require buffer capacity to handle traffic spikes or node failures
+4. **Consistent user experience** across different autoscaling solutions in the Kubernetes ecosystem
+
+Currently, users attempt to achieve buffer capacity through workarounds:
+- Deploying "balloon pods" or placeholder workloads to maintain minimum capacity
+- Using pause containers with resource requests to reserve capacity
+- Over-provisioning through static NodePools
+
+The Kubernetes SIG Autoscaling has standardized a CapacityBuffer API to declare spare capacity/headroom in clusters. Cluster Autoscaler supports this API (autoscaling.x-k8s.io/v1alpha1), providing a vendor-agnostic way to express capacity requirements.
```
DerekFrank: I'm not sure CAS has support for that API yet

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
+# Capacity Buffer API Support
+
+## Background
+
+Karpenter currently operates as a dynamic cluster autoscaler, provisioning nodes just-in-time based on pending pod demand. However, several important use cases require maintaining pre-provisioned spare capacity:
+
+1. **Performance-critical applications** where just-in-time provisioning latency is unacceptable
+2. **Burst workloads** that need immediate scheduling for CI/CD, batch jobs, or event-driven applications
+3. **High-availability services** that require buffer capacity to handle traffic spikes or node failures
+4. **Consistent user experience** across different autoscaling solutions in the Kubernetes ecosystem
```
DerekFrank: I don't think we care all that much about consistent UX. In fact, the two autoscaling solutions work very differently. I do think we could say we care about intent driven configuration though

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+Currently, users attempt to achieve buffer capacity through workarounds:
+- Deploying "balloon pods" or placeholder workloads to maintain minimum capacity
+- Using pause containers with resource requests to reserve capacity
+- Over-provisioning through static NodePools
+
+The Kubernetes SIG Autoscaling has standardized a CapacityBuffer API to declare spare capacity/headroom in clusters. Cluster Autoscaler supports this API (autoscaling.x-k8s.io/v1alpha1), providing a vendor-agnostic way to express capacity requirements.
+
+## Proposal
+
+Extend Karpenter to support the standard CapacityBuffer API (autoscaling.x-k8s.io/v1alpha1) by integrating buffer capacity into scheduling and consolidation algorithms.
```
DerekFrank: Given that the API is alpha, whatever design we create should include the standard set of alpha protections we use. I don't see where you've discussed feature gating this and the opt-in opt-out behavior, but the RFC should include details on that

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+  # Reference existing workload to derive capacity
+  scalableRef:
+    apiGroup: apps
+    kind: Deployment
+    name: my-app
+  replicas: 5  # Maintain buffer for 5 additional replicas
+```
+
+Key aspects:
+1. **Virtual Pod Approach**: Follow Cluster Autoscaler's pattern using in-memory virtual pods
```
DerekFrank: I think we can reduce this to a single goal, something along the lines of 'Karpenter respects configured CapacityBuffers, maintaining additional capacity as if they were pods'. 1-3 in this list are implementation details that get us towards that goal

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+2. **Scheduling Integration**: Buffer pods participate in Karpenter's bin-packing algorithms
+3. **Consolidation Protection**: Ensure buffer capacity is preserved during node optimization
+4. **Standard Compatibility**: Use the same CRD as Cluster Autoscaler for ecosystem consistency
+
+### Modeling & Validation
+
+When a `CapacityBuffer` is created, Karpenter generates virtual pods that participate in scheduling decisions but are never actually created in the cluster. These virtual pods:
```
DerekFrank: I think you've skipped passed UX and right into implementation here.

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+  replicas: 5  # Maintain buffer for 5 additional replicas
+```
+
+Key aspects:
+1. **Virtual Pod Approach**: Follow Cluster Autoscaler's pattern using in-memory virtual pods
+2. **Scheduling Integration**: Buffer pods participate in Karpenter's bin-packing algorithms
+3. **Consolidation Protection**: Ensure buffer capacity is preserved during node optimization
+4. **Standard Compatibility**: Use the same CRD as Cluster Autoscaler for ecosystem consistency
+
+### Modeling & Validation
```
DerekFrank: Are you proposing validation on these buffers that Karpenter will need to perform individually?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+When a `CapacityBuffer` is created, Karpenter generates virtual pods that participate in scheduling decisions but are never actually created in the cluster. These virtual pods:
+
+- Are derived from either `scalableRef` (existing workload) or `podTemplateRef` (custom template)
+- Include special labels for identification: `karpenter.sh/capacity-type: buffer`
+- Have lower priority than real workloads to ensure user pods take precedence
+- Are considered during both provisioning and consolidation decisions
+
+Buffer capacity persists until:
+- The CapacityBuffer resource is deleted
+- Real workloads consume the buffered capacity
```
DerekFrank: I'm not sure this is the intended experience of the buffer api. When new pods fit into the buffered capacity, I think we should scale up to maintain that buffer: https://github.com/kubernetes/autoscaler/pull/8151/files#diff-8ea88c4ccba1b6b78409a11d4ffbeda163a396f77a906a78647a344e572de5a8R71

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+- `apps/v1/ReplicaSet` 
+- `apps/v1/StatefulSet`
+- `core/v1/PodTemplate` (via PodTemplateRef)
+- Any resource with scale subresource
+
+### Virtual Pod Generation
+
+Following Cluster Autoscaler's pattern:
+
+1. **ScalableRef Resolution**: Query the referenced workload's pod template
```
DerekFrank: I think this section would be better off enumerating the workflows that would require the generation of new virtual pods in addition to how big they are

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+2. **PodTemplateRef Resolution**: Directly use the referenced template
+3. **Replica Calculation**: Compute absolute replicas from relative percentages
+4. **Virtual Pod Creation**: Generate N in-memory pods with special labels:
+   ```yaml
+   labels:
+     karpenter.sh/capacity-type: buffer
+     autoscaling.x-k8s.io/buffer: <buffer-name>
+     autoscaling.x-k8s.io/buffer-namespace: <buffer-namespace>
+   ```
+5. **Priority Assignment**: Use system priority class or lowest priority to ensure user pods take precedence
```
DerekFrank: I don't think Karpenter cares about this, we should always maintain capacity for both pods and buffers, and kube-scheduler will ensure that the pods actually schedule. Did you mean to refer to disruption cost?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+
+## Implementation Details
+
+### Controller Architecture
+
+We will implement capacity buffer support by adding a new controller and integrating with existing Karpenter components. 
+
+**⚠️ Critical Integration Considerations**: This implementation must account for Karpenter's unique architecture that operates at the NodeClaim level rather than individual pod scheduling, and integrates deeply with Karpenter's state management system.
```
DerekFrank: I think you can remove this, it'd be more useful to enumerate the controller's APIs and how they get leveraged by other controllers

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+
+#### Buffer Controller (New)
+**Location**: `pkg/controllers/capacitybuffer/`
+
+**Purpose**:
```
DerekFrank: What controller APIs correspond to these purposes?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+    // Existing logic with buffer awareness
+    // Buffer pods participate in topology constraints and affinity calculations
+    // But are treated as "evictable" during constraint solving
+}
+```
+
+#### Disruption Integration  
+**Location**: `pkg/controllers/disruption/`
+
+**Behavior**:
```
DerekFrank: Disruption cost modifications are missing here

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+- Buffer pods can be evicted during consolidation (they're virtual and easily recreated)
```
DerekFrank: nit: Buffer pods don't need to be evicted (they don't exist)

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+
+3. **Parallel Execution**: Disruption commands are executed in parallel with careful orchestration
+
+**Revised Integration Approach**:
+
+```go
+// In disruption/consolidation.go
+func (c *consolidation) computeConsolidation(ctx context.Context, candidates ...*Candidate) (Command, error) {
+    // 1. Get buffer requirements for affected nodes
+    bufferRequirements := c.bufferController.GetBufferRequirementsForNodes(ctx, candidates)
```
DerekFrank: I don't think I understand what this is doing. Is this attempting to ensure we maintain capacity for buffer pods?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+        return Command{}, fmt.Errorf("consolidation would violate buffer capacity requirements")
+    }
+    
+    return c.buildConsolidationCommand(results), nil
+}
+
+// New method: Buffer-aware scheduling simulation
+func (c *consolidation) simulateSchedulingWithBuffers(
+    ctx context.Context, 
+    realPods, bufferPods []*corev1.Pod,
```
DerekFrank: Why do we care if the pods are buffered or real?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+    requirements *BufferRequirements,
+) bool {
+    // Ensure equivalent buffer capacity is available after consolidation
+    return c.bufferController.ValidateCapacityPreservation(results, requirements)
+}
+```
+
+**State Management Integration**:
+```go
+// Integration with state.Cluster
```
DerekFrank: If we can avoid using state.Cluster, we should avoid using state.Cluster.

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+        if bc.isBufferCapacityInsufficient(bufferState) {
+            bc.requestBufferProvisioning(ctx, bufferState)
+        }
+    }
+    
+    return nil
+}
+```
+
+**Revised Protection Strategy**:
```
DerekFrank: I'm not sure what this section is saying. Is this an alternative set of requirements or are you re-iterating?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+
+2. **State Synchronization**: Buffer controller maintains consistency with Karpenter's `state.Cluster`
+
+3. **Disruption Budget Integration**: Buffer requirements participate in Karpenter's disruption budget calculations
+
+4. **Capacity Preservation Validation**: Before any consolidation, validate that equivalent buffer capacity will remain available
+
+5. **Graceful Degradation**: If buffer capacity cannot be maintained, prioritize user workloads and log buffer capacity warnings
+
+### API Integration
```
DerekFrank, line 435: I think this section is repeated

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+        }
+    }
+    
+    return nil
+}
+```
+
+**Revised Protection Strategy**:
+
+1. **NodeClaim-Level Tracking**: Buffer capacity is tracked at the NodeClaim level, not just pod level
```
DerekFrank: I don't think this is correct. What are you trying to say with this?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+2. **Filter by provisioning strategy** (buffer.x-k8s.io/active-capacity)
+3. **Translate buffer specs**:
+   - ScalableRef → query target workload's pod template
+   - PodTemplateRef → use referenced template directly
+   - Calculate replica count (absolute, percentage, or resource-limited)
+4. **Generate virtual pods** with buffer-specific labels
+5. **Update buffer status** with translation results
+6. **Inject virtual pods** into Karpenter's scheduling pipeline
+
+### Implementation Phases
```
DerekFrank, line 507: Could this section enumerate what requirements must be met before implementation can begin, and then what all functionality is required for the alpha release of buffer support?

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+- Delete CapacityBuffer → Nodes deprovisioned (if unused)
+- User pods schedule immediately on buffer capacity
+- Consolidation does not remove buffer nodes
+
+### Performance Tests
+- Impact on scheduling latency
+- Memory overhead of virtual pods
+- Watch performance with many buffers
+
+## Migration & Compatibility
```
DerekFrank, line 723: I think this section is missing a feature flag discussion

designs/capacity-buffer.md
```diff
@@ -0,0 +1,594 @@
...
+## Open Questions & Karpenter-Specific Considerations
+
+1. **Q**: Should we support buffer priorities?
+   **A**: Defer to future KEP - not in initial scope
+
+2. **Q**: How to handle buffer with no matching nodes?
+   **A**: Follow Karpenter's provisioning behavior - create suitable NodeClaims through scheduler constraint solving
+
+3. **Q**: How does buffer capacity interact with NodePool limits?
+   **A**: Buffer NodeClaims must respect NodePool resource limits and budget constraints
```
DerekFrank: Again, what is a buffer nodeclaim?

DerekFrank (changes requested):

designs/capacity-buffer.md
moko-poi: @DerekFrank 

> I don't think the capacity buffer api exists yet. The sig-autoscaling RFC was merged but I don't think the api itself has been released. Doesn't mean we can't think ahead, but implementation for this will have to wait until the api exists.

I've verified that the Capacity Buffer API was released in cluster-autoscaler 1.34.0 (Sep 27, 2025):
- #8508: Add capacity buffer CRD  
- #8521: Add Capacity Buffer controller logic
- #8825: CRD scope fix (1.34.2)
https://github.com/kubernetes/autoscaler/releases?q=capacity+buffer&expanded=true

moko-poi: > Overall I think the doc is a bit messy. I think it would be a stronger proposal if you started from the CX and derived implementation from that. Similarly, I think the implementation section could be much stronger if it started from the requirements of the existing controllers and worked backwards to the API's that the capacity buffer controller should provide.

Great feedback. I'll restructure the document to:
1. Start with user-facing customer experience (CX) - what users want to achieve
2. Derive implementation requirements from CX needs
3. Restructure implementation section to work backwards from existing controller 
   requirements to the APIs that buffer controller should provide

This will make the design rationale clearer.

designs/capacity-buffer.md
moko-poi: The API is now available at autoscaling.x-k8s.io/v1alpha1.
https://github.com/kubernetes-sigs/karpenter/pull/2611#discussion_r2630239753

designs/capacity-buffer.md
moko-poi: Fair point - Karpenter and CAS work quite differently by design. I'll reframe this as "intent-driven configuration" instead. The goal is enabling users to express "I want X amount of buffer capacity" in a declarative way, not necessarily matching CAS behavior.

designs/capacity-buffer.md
moko-poi: Absolutely - I'll add a feature gating section covering:
1. Feature flag: `CapacityBuffer` (disabled by default)
2. Opt-in behavior via feature gate flag
3. Stability levels and graduation criteria
4. Alpha protection: no SLA, may have bugs, disabled by default

This follows Karpenter's standard alpha feature practices.

designs/capacity-buffer.md
moko-poi: Good simplification. I'll consolidate the key aspects to:
"Karpenter respects configured CapacityBuffers, maintaining additional capacity as if they were pods scheduled in the cluster."

The current numbered list (virtual pods, scheduling, consolidation) are implementation details that should move to the implementation section.

designs/capacity-buffer.md
moko-poi: You're right - I have API specification details in multiple places.

designs/capacity-buffer.md
moko-poi: @DerekFrank 
I think I was confusing concepts here. Let me clarify what I meant:

Buffer capacity is represented as virtual pods (not NodeClaims). These virtual pods participate in scheduling just like real pods, which may result in new NodeClaims being created. The buffer controller needs to track which capacity was provisioned for buffer purposes.

Should I clarify this as "tracking which capacity was provisioned for buffer pods" instead? Or is the entire concept unclear?

designs/capacity-buffer.md
moko-poi: Good point. I'll restructure Implementation Phases.

designs/capacity-buffer.md
moko-poi: I'll add a dedicated Feature Flag section covering:
- Flag name and default state
- Opt-in/opt-out behavior  
- CRD installation requirements
- Graduation criteria for beta/stable

designs/capacity-buffer.md
moko-poi: This is unclear terminology on my part. There's no such thing as a "buffer nodeclaim" 
- that's mixing concepts.

What actually happens:
1. Buffer controller creates virtual pods
2. Virtual pods participate in scheduling (like real pods)
3. If capacity is needed, Karpenter's provisioner creates NodeClaims (as normal)
4. These NodeClaims aren't "buffer NodeClaims", they're regular NodeClaims that 
   happen to accommodate buffer pods

I'll clarify this throughout the document and remove confusing terminology.

designs/capacity-buffer.md
moko-poi: No, validation is handled by the CRD's built-in validation rules (kubebuilder tags). 
Karpenter doesn't need to perform additional validation - we just watch and process valid CapacityBuffer resources. The "Modeling & Validation" section is describing the behavior model, not proposing Karpenter-side validation. I'll clarify this section or remove it if it's confusing.

designs/capacity-buffer.md
moko-poi: You're absolutely right - I misunderstood the behavior. The buffer should be maintained continuously. When real pods consume buffer capacity, Karpenter should provision additional capacity to restore the buffer. I'll update the document to reflect this:
- Buffer capacity is maintained as long as the CapacityBuffer resource exists
- When real pods are scheduled on capacity that was provisioned for buffers, 
  Karpenter provisions additional capacity to maintain the buffer
- The buffer acts like a "target spare capacity" rather than a one-time provision

designs/capacity-buffer.md
moko-poi: Good point. Instead of focusing on how we resolve templates, I should describe when and why we generate virtual pods. I'll restructure this section to enumerate:

1. Initial CapacityBuffer creation → generate virtual pods
2. CapacityBuffer spec update (replicas/percentage change) → regenerate virtual pods
3. Referenced workload scale (for percentage-based buffers) → regenerate virtual pods
4. PodTemplate generation change → regenerate virtual pods

Then explain how these virtual pods flow into the provisioner.

designs/capacity-buffer.md
moko-poi: You're right - I was thinking about kube-scheduler behavior, not Karpenter behavior. Karpenter provisions capacity for all pods (real and virtual), and kube-scheduler handles priority during actual pod placement. I'll remove the priority assignment discussion from the virtual pod generation section since it's not relevant to Karpenter's provisioning logic.

designs/capacity-buffer.md
moko-poi: Agreed. Instead of warning about architecture complexity, I should clearly define:
- What APIs the buffer controller provides to other controllers
- How those APIs are consumed (provisioner, disruption)
- The contract/interface between components

I'll restructure the Controller Architecture section to focus on APIs and contracts.

designs/capacity-buffer.md
moko-poi: I should define explicit interfaces:

```go
type BufferController interface {
    // GetVirtualPods returns all active buffer pods for scheduling
    GetVirtualPods(ctx context.Context) ([]*corev1.Pod, error)
    
    // GetDisruptionCost returns cost for disrupting pods on a node
    GetDisruptionCost(ctx context.Context, node *corev1.Node) (float64, error)
    
    // Run starts the reconciliation loop
    Run(ctx context.Context) error
}

designs/capacity-buffer.md
moko-poi: Good catch. Buffer pods should have lower disruption cost than real workloads.
I'll add:

- Virtual buffer pods should have disruption cost of 0 or very low value
- This allows consolidation to prefer moving buffer capacity over real workloads
- Disruption controller needs to query buffer controller for these costs

I'll add this to the Disruption Integration section.

designs/capacity-buffer.md
moko-poi: Exactly right - poor word choice. Virtual pods don't need eviction, they just needto be removed from the virtual pod set. I'll change the language to:
"Buffer virtual pods can be removed from consideration during consolidation (theydon't require actual eviction)"

designs/capacity-buffer.md
moko-poi: @DerekFrank 
Yes, but I think I'm overcomplicating it. The simpler approach:
- Treat virtual buffer pods like any other pods during consolidation simulation
- If consolidation would remove capacity that buffer pods need, it's invalid
- No special "buffer requirements" tracking needed - the pods themselves represent the requirement

Should I simplify this to just include buffer pods in the consolidation simulation?

designs/capacity-buffer.md
moko-poi: We probably don't need to distinguish. Both should participate in scheduling simulation equally. The only difference might be disruption cost. I'll removethe distinction in the simulation logic and just pass all pods together.

designs/capacity-buffer.md
moko-poi: Understood. I was assuming we'd need state.Cluster integration, but if we can avoid it, that's simpler. The buffer controller should:

- Watch CapacityBuffer CRDs
- Generate virtual pods
- Provide those pods to the provisioner via a simple interface

No need for deep state.Cluster integration. I'll remove those references.

designs/capacity-buffer.md
moko-poi: It was meant to be clarification but ended up confusing. I'll remove this
"Revised Protection Strategy" section since the actual implementation approach should be clear from the Controller Architecture and Integration sections.

designs/capacity-buffer.md
moko-poi: You're absolutely right. I've restructured the document to address this:

1. Added a "User Experience" section that shows what users expect when they 
   create a CapacityBuffer
2. The "Modeling & Validation" section has been moved/merged into implementation 
   details
3. Document now flows: Background → User Experience → Proposal → Design Decisions 
   → Implementation

The structure now starts from user intent and derives implementation from that, rather than jumping straight to technical details.

moko-poi:
@DerekFrank 

Thank you for the detailed review. I've addressed all of your feedback comments and restructured the RFC accordingly. 

Please review again when you have time.

PR #2620, opened 2025-11-10
chore: Add locking mechanism to nodePoolRegistration State

jigisha620:
Fixes #N/A 

**Description**

The `nodepoolhealth.State` was implemented as an unsynchronized `map[types.UID]*Tracker`, causing race conditions and potential panics when multiple goroutines accessed NodePool health state concurrently.

Changes in this PR -
- Converted State to thread-safe struct by changing from map type to struct with sync.RWMutex and trackers map
- Added NewState() constructor for proper initialization
- Implemented double-checked locking in nodePoolNodeRegistration()
- Added proper synchronization i.e. RLock() protection for Status() and DryRun() read operations
- Fixed SetStatus() deadlock by using direct buffer.Insert() instead of nested Update() calls

Now we have consistent lock hierarchy: State → Tracker

Added tests. Tested on dev cluster.

### DIFF
--- a/pkg/state/nodepoolhealth/tracker.go
+++ b/pkg/state/nodepoolhealth/tracker.go
@@ -63,6 +63,9 @@ func (t *Tracker) Reset() {
 }
 
 func (t *Tracker) Status() Status {
+	t.RLock()
+	defer t.RUnlock()
+
 	if t.buffer.Len() == 0 {
 		return StatusUnknown
 	}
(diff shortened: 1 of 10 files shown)


pkg/state/nodepoolhealth/tracker.go
```diff
@@ -82,49 +85,73 @@ func (t *Tracker) Status() Status {
...
+	s.RLock()
+	tracker, exists := s.trackers[nodePoolUID]
+	s.RUnlock()
 
-func (s State) nodePoolNodeRegistration(nodePoolUID types.UID) *Tracker {
-	tracker, exists := s[nodePoolUID]
 	if !exists {
-		tracker = NewTracker(BufferSize)
-		s[nodePoolUID] = tracker
+		s.Lock()
```
ryan-mist, line 122: I think this results in a deadlock because RWMutex doesn't support recursive locking. IIRC, if a goroutine has the read lock and tries to acquire the write lock it results in a deadlock

ryan-mist: Nevermind just saw the s.RUnlock()

jigisha620: But we are releasing the read lock before acquiring the write lock.

pkg/state/nodepoolhealth/tracker.go
```diff
@@ -82,49 +85,73 @@ func (t *Tracker) Status() Status {
 }
 
 func (t *Tracker) SetStatus(status Status) {
+	t.Lock()
+	defer t.Unlock()
+
 	switch status {
 	case StatusUnknown:
 		t.buffer.Reset()
```
jmdeal, line 93: Don't we now have a deadlock issue here since `Reset` also acquires a lock? I'd consider just adding private methods (e.g. `reset` and `update`) which don't acquire the lock and call them directly here. Then the public versions of those methods are just locked wrappers.

jigisha620: `Reset` doesn't acquire a lock though.

jmdeal: I was looking at `(*Tracker) Reset` and got my wires crossed, lgtm then 🚀

PR #2624, opened 2025-11-12
fix: Exclude DaemonSet pods from disruption cost calculation

### LINKED ISSUE #2084: Karpenter consolidates node with pod over empty node
### Description

**Observed Behavior**:

In a nodepool with four nodes, and four pods — one pod per node — experiences a downscale to three pods, the now empty node is not disrupted and instead a node with a pod is disrupted instead. 

**Expected Behavior**:

The empty node is disrupted.

**Versions**:
- Chart Version: Karpenter 1.0.1
- Kubernetes Version (`kubectl version`): 1.30

**Notes**

I chased through the code to try and understand why this was happening. As best as I can tell, this seems to happen if the empty node is older than other nodes. This makes sense as the disruption cost is [set to be the multiple of the sum of the cost of rescheduling the pods and the lifetime remaining of the node](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/types.go#L113-L115). This means that this bug only effects nodes with `ExpireAfter` set. Due to the pod cost calculation, if a nodepool has a large number of daemonsets compared to the number of "workload pods" on the nodes, then the pod eviction cost is dominated by the daemonsets, leading to the sorting being approximately the same as just sorting the nodes by age.

I think this is a bug because the empty nodes should always be disrupted first, even if it has 99% lifetime remaining and the other has 5%. I believe the root cause is due to daemonset pods being part of the eviction cost of pods on a node. All pods in a nodepool should have the same set of daemonset pods, so removing them should provide a more intuitive sorting order.

I am willing to work on a fix for this issue.

- [node disruption cost multiplication](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/disruption/types.go#L113-L115)
- [pods eviction cost](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/utils/disruption/disruption.go#L49-L78)
- [node lifetime cost calculation](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/utils/disruption/disruption.go#L39)

The PR that [added this code does not explain why we are counting daemonsets.](https://github.com/kubernetes-sigs/karpenter/pull/789/files#diff-2a7cf721cf6d408a7025db62063f6d774475dc1f9acf69ff7957c802f52b39f3R141)

moko-poi:
Fixes #2084

**Description**
Fixes an issue where Karpenter consolidates nodes with workload pods instead of empty nodes when the empty node has lower remaining lifetime.

The root cause was that disruption cost calculation included DaemonSet pods, even though:
- DaemonSet pods are not rescheduled (automatically recreated by DaemonSet controller)
- All nodes have the same DaemonSet pods
- This caused the cost to be dominated by node age rather than actual workload

**Changes:**
- Use `reschedulablePods` (excluding DaemonSet pods) instead of all `pods` for disruption cost calculation in `pkg/controllers/disruption/types.go`
- This ensures empty nodes (with only DaemonSet pods) have zero rescheduling cost and are prioritized for consolidation

- All existing disruption tests pass (232 tests)
- Existing test `emptiness_test.go:555-603` validates DaemonSet-only nodes are treated as empty
- No regressions introduced

### DIFF
--- a/pkg/controllers/disruption/types.go
+++ b/pkg/controllers/disruption/types.go
@@ -121,17 +121,15 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 			return nil, err
 		}
 	}
-	reschedulablePods := lo.Filter(pods, func(p *corev1.Pod, _ int) bool { return pod.IsReschedulable(p) })
 	return &Candidate{
 		StateNode:         node,
 		instanceType:      instanceType,
 		NodePool:          nodePool,
 		capacityType:      node.Labels()[v1.CapacityTypeLabelKey],
 		zone:              node.Labels()[corev1.LabelTopologyZone],
-		reschedulablePods: reschedulablePods,
-		// Only include reschedulable pods in disruption cost calculation.
-		// DaemonSet pods are excluded because they will be automatically recreated on any new node.
-		DisruptionCost: disruptionutils.ReschedulingCost(ctx, reschedulablePods) * disruptionutils.LifetimeRemaining(clk, nodePool, node.NodeClaim),
+		reschedulablePods: lo.Filter(pods, func(p *corev1.Pod, _ int) bool { return pod.IsReschedulable(p) }),
+		// We get the disruption cost from all pods in the candidate, not just the reschedulable pods
+		DisruptionCost: disruptionutils.ReschedulingCost(ctx, pods) * disruptionutils.LifetimeRemaining(clk, nodePool, node.NodeClaim),
 	}, nil
 }
 


pkg/controllers/disruption/types.go
```diff
@@ -121,15 +121,17 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 			return nil, err
 		}
 	}
+	reschedulablePods := lo.Filter(pods, func(p *corev1.Pod, _ int) bool { return pod.IsReschedulable(p) })
```
DerekFrank: I think this change makes sense logically, but I'm not sure it actually solves #2084. @rschalo's comment in the original issue is correct, emptiness should be considered prior to single node consolidation. Also, won't this only affect cases where some nodes have more daemonsets than others? Is there a common case for that?

moko-poi: @DerekFrank Great question! Let me address both points.

On DaemonSet count differences:

No, this fix doesn't require nodes to have different DaemonSet counts. The issue occurs even when all nodes have the same number of DaemonSets, which is the typical case.

The bug isn't about differences in DaemonSet counts between nodes. It's about DaemonSets inflating the DisruptionCost, which causes node age to dominate the sorting instead of actual workload.

Example: All nodes have the SAME DaemonSet count (5 DaemonSets each, `ExpireAfter=90d`):

- Empty node (80d old): 0 workload pods + 5 DaemonSets
- Node with pod (80d old): 1 workload pod + 5 DaemonSets

Current behavior:

```
Empty node:      Cost = (5 DaemonSets + 0 pods) × 0.11 = 0.55
Node with pod:   Cost = (5 DaemonSets + 1 pod)  × 0.11 = 0.66
Difference: only 0.11
```

The DaemonSets dominate the cost (5 out of 5 vs 5 out of 6), making the difference tiny. Node age becomes the primary sorting factor.

After this PR:

```
Empty node:      Cost = 0 pods × 0.11 = 0.00
Node with pod:   Cost = 1 pod  × 0.11 = 0.11
```

Empty nodes are now clearly prioritized regardless of age.

On whether this solves #2084:

You're absolutely right that Emptiness runs before Single-node consolidation. However, this DisruptionCost bug affects both phases, not just Single-node consolidation.

The Emptiness phase also uses DisruptionCost for sorting
https://github.com/kubernetes-sigs/karpenter/blob/7de3cedace196e510992fef411615d8b8b7c73e4/pkg/controllers/disruption/emptiness.go#L58-L62

This means that even in the Emptiness phase, when there are multiple empty nodes with the same DaemonSet count but different ages, the current bug causes them to be sorted by age (due to different `LifetimeRemaining` values) rather than all having the same cost of zero.

Additionally, there are scenarios where empty nodes aren't deleted in the Emptiness phase and proceed to Single-node consolidation:

1. Cluster already marked as consolidated 
https://github.com/kubernetes-sigs/karpenter/blob/7de3cedace196e510992fef411615d8b8b7c73e4/pkg/controllers/disruption/emptiness.go#L59
2. Disruption budget constraints
https://github.com/kubernetes-sigs/karpenter/blob/7de3cedace196e510992fef411615d8b8b7c73e4/pkg/controllers/disruption/emptiness.go#L70
3. Validation errors during command execution

In these cases, the node proceeds to Single-node consolidation, where the incorrect DisruptionCost causes the wrong node to be selected.

This PR fixes the DisruptionCost calculation in both Emptiness and Single-node consolidation phases, addressing the root cause described by @JeremyBolster in [#2084](https://github.com/kubernetes-sigs/karpenter/issues/2084):

> if a nodepool has a large number of daemonsets compared to the number of "workload pods" on the nodes, then the pod eviction cost is dominated by the daemonsets, leading to the sorting being approximately the same as just sorting the nodes by age.

DerekFrank: Disruption Cost is just used to sort the nodes, the relative difference between two nodes shouldn't matter. We only use the value ordinally, we don't care about the difference.

There is also the unintended side effect of this preventing empty nodes from having a meaningful metric to sort by. If you have two empty nodes, their lifetime will no longer be taken into account, because 0*anything is still 0. I don't know if that is an issue, I haven't fully considered it

As for the ways that nodes can be skipped in Emptiness:
1. If the cluster is marked as consolidated for emptiness, it should also be marked as consolidated for Single Node. That could be the bug if there is a race condition there, but then that would be for the fix.
2. If the disruption budgets for a nodepool are blocking emptiness, they should also block single node for the same nodepool
3. If the emptiness command is invalid, then we couldn't consolidate that node. If there is a bug there that is invalidating the command when it shouldn't, that would be a bug and we should fix that.

If you can reproduce this issue regularly and demonstrate that the disruption cost is indeed to blame, I would love to see that reproduction, but as it stands I fail to see how this change will solve the linked issue.

moko-poi: You are absolutely right that we should first confirm whether we can reproduce this issue, and then consider the actual response.

PR #2629, opened 2025-11-14
chore: adding custom config into leader election

MaximilianoUribe:
Fixes #N/A 

**Description**
Need custom configuration for leader election

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -104,6 +104,7 @@ type Operator struct {
 
 type Options struct {
 	LeaderElectionLabels map[string]string
+	LeaderElectionConfig *rest.Config  // Optional separate config for leader election
 }
 
 // Adds LeaderElectionLabels to the underlying manager's LeaderElectionOptions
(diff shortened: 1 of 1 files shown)


DerekFrank:
Hello @MaximilianoUribe, could you detail what config you need to add and why?

MaximilianoUribe:
Hi @DerekFrank,
We need a custom configuration for leader election in Karpenter for Azure, as it requires using `rest.InClusterConfig()` to function correctly. Since the client is utilized in the core operator, we’re proposing this approach as a solution. Interested to hear your thoughts or any concerns you might have about this direction.

matthchr:
Context here is that (in Azure) Karpenter is hosted in cluster A but manages cluster B. So the kubeclient for nodes/CRDs/etc is cluster B, but leader election should be managed in cluster A. This is already achievable with Karpenter configuration today for controllers we own (we just register them with whichever kubeclient is correct), but since leader election is managed by core we need a hook to configure it to point at Cluster A rather than B.

MaximilianoUribe:
@DerekFrank can we get an update on this?

PR #2639, opened 2025-11-18
fix: topology spread constraint with nodeAffinityPolicy: Honor

### LINKED ISSUE #2623: Karpenter doesn’t honor nodeAffinityPolicy in topologySpreadConstrains when taintless NodePools exist
### Description

**Observed Behavior**:
Karpenter refuses to provision new nodes for a deployment that:
1. has a strict nodeSelector and tolerations targeting a specific tainted NodePool (nodepool A), and
2. uses `topologySpreadConstraints` with `nodeAffinityPolicy: Honor` and `nodeTaintsPolicy: Honor` on a topology key- [capacity-spread](https://karpenter.sh/docs/concepts/scheduling/#on-demandspot-ratio-split).

Pods from Deployment remain in Pending, and Karpenter logs errors like:
```
unsatisfiable topology constraint for topology spread, key=capacity-spread
(counts = 1: 1, 10: 0, 2: 0, 3: 0, 4: 0, 5: 0, 6: 0, 7: 0, 8: 0, 9: 0,
 podDomains = capacity-spread Exists,
 nodeDomains = capacity-spread In [1])
```

Key facts:
- NodePool A:
   - has label & taint of `nodepool_a=foo`
   - has requirement of `capacity-spread In ["1"]` **ONLY**
   - is the **only** NodePool whose nodes match deployment's **nodeSelector + tolerations**

- NodePool B:
   - has different labels that do not match the nodeSelector
   - **has no taints**
   - defines a wider capacity-spread range - `["2","3","4","5","6","7","8","9","10"]`

Even though there is exactly one node matching the nodeSelector + taint (in NodePool A, with capacity-spread="1"), Karpenter still sees the constrain from NodePool B and failing to create a new node from NodePool A.

**Expected Behavior**:
Given:
- Deployment's nodeSelector only matches NodePool A, and
- topologySpreadConstraints on **`nodeAffinityPolicy: Honor`** (which is default) and `nodeTaintsPolicy: Honor`,

The expected behavior is:
- Since `nodeAffinityPolicy` is set to `Honor`, the only relevant nodepool should be A, which would create node with capacity-spread of 1 instead of pods being stuck in Pending. But Karpenter takes a look at Nodepool B and sees it's capacity-spread values as well,
- For this workload, there should effectively be a single topology domain (NodePool A with capacity-spread="1"), so maxSkew: 1 is satisfied.
- When Deployment is scaled up and pods are Pending, Karpenter should provision additional nodes in NodePool A instead of treating the constraint as unsatisfiable and leaving pods Pending.

1. Create Nodepool A with relevant labels & taint:
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: workload-nodepool-a
spec:
  template:
    metadata:
      labels:
        # Used by the deployment's nodeSelector
        workload-group: isolated-app
    spec:
      # Example, adjust to your environment
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: workload-nodeclass
      expireAfter: Never
      taints:
      - key: workload-group
        value: isolated-app
        effect: NoSchedule
      requirements:
      - key: kubernetes.io/os
        operator: In
        values:
        - linux
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
      # Example instance constraints – can be anything reasonable
        operator: In
        values:
        - m
        - c
        operator: Gt
        values:
        - "4"
        operator: In
        values:
        - on-demand
      # This is the important topology domain constraint
      - key: topology-spread-domain
        operator: In
        values:
        - "1"
```
2. Create nodepool B without any taints:
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: shared-nodepool-b
spec:
  template:
    metadata:
      labels:
        # Generic shared apps group, DOES NOT match test-bug nodeSelector
        workload-group: shared-apps
    spec:
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: shared-apps-nodeclass
      # No taints – this is part of the bug scenario (taintless NodePool)
      expireAfter: Never
      requirements:
      - key: kubernetes.io/os
        operator: In
        values:
        - linux
      - key: kubernetes.io/arch
        operator: In
        values:
        - amd64
        operator: In
        values:
        - m
        - c
        operator: Gt
        values:
        - "4"
        operator: In
        values:
        - on-demand
        - spot
      # Important: this NodePool advertises MANY topology domains
      - key: topology-spread-domain
        operator: In
        values:
        - "2"
        - "3"
        - "4"
        - "5"
        - "6"
        - "7"
        - "8"
        - "9"
        - "10"
```

3. Create a deploymeny with relevant Tolerations & NodeSelector:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-bug
  namespace: test-bug-namespace
spec:
  selector:
    matchLabels:
      app: test-bug
  template:
    metadata:
      labels:
        app: test-bug
    spec:
      # Restrict to NodePool A only
      nodeSelector:
        workload-group: isolated-app
      # Tolerate the taint on NodePool A
      tolerations:
      - key: workload-group
        value: isolated-app
        effect: NoSchedule
      # Topology spread
      topologySpreadConstraints:
      - topologyKey: capacity-spread
        maxSkew: 1
        labelSelector:
          matchLabels:
            app: test-bug
        nodeAffinityPolicy: Honor
        nodeTaintsPolicy: Honor
      containers:
      - name: test-bug
        image: nginx:1.27
```

4. Scale up the deployment so that it requires creating at least two new nodes:
`kubectl scale deployment test-bug --replicas=50`

5. Observe that pods are stuck in Pending due to Karpenter simulation of breaking Skew

**Versions**:
- Chart Version: `1.8.2`
```
Client Version: v1.33.0
Kustomize Version: v5.6.0
Server Version: v1.33.5-eks-113cf36
```

tal19987:
## Description

This PR fixes issue #2623 where topology spread constraints with `nodeAffinityPolicy: Honor` were not correctly considering node affinity requirements when calculating topology domains.

## Problem

When a pod has topologySpreadConstraints with `nodeAffinityPolicy: Honor`, Karpenter was incorrectly including ALL topology domains from all nodepools instead of only those matching the pod's nodeSelector. This caused 'unsatisfiable topology constraint' errors even when valid nodes could be provisioned which caused pods to be stuck in Pending for ever.

## Solution

- Updated `TopologyDomainGroup` to track both taints AND requirements for each topology domain
- Modified `ForEachDomain` to use `TopologyNodeFilter` that considers both `nodeTaintsPolicy` and `nodeAffinityPolicy`
- Added `matchesRequirements` check to filter domains by node requirements

## Example Scenario (from issue #2623)

**Before this fix:**
- NodePool A: capacity-spread=1, taint workload-group=app:NoSchedule
- NodePool B: capacity-spread=2-10, no taint
- Pod: nodeSelector workload-group=app, tolerates taint, topology spread with nodeAffinityPolicy: Honor

Karpenter considered domains 1-10, causing unsatisfiable constraint errors.

**After this fix:**
Karpenter correctly considers only domain 1, allowing proper provisioning.

## Testing

Tested locally with a reproduction scenario matching the issue. Verified:
- Zero 'unsatisfiable topology constraint' errors
- Pods successfully scheduled to correct nodes
- Topology domains correctly filtered by node affinity

Fixes #2623

### DIFF
--- a/pkg/controllers/provisioning/scheduling/topology_test.go
+++ b/pkg/controllers/provisioning/scheduling/topology_test.go
@@ -1680,17 +1680,13 @@ var _ = Describe("Topology", func() {
 							}},
 							Requirements: []v1.NodeSelectorRequirementWithMinValues{
 								{
-									NodeSelectorRequirement: corev1.NodeSelectorRequirement{
-										Key:      v1.CapacityTypeLabelKey,
-										Operator: corev1.NodeSelectorOpExists,
-									},
+									Key:      v1.CapacityTypeLabelKey,
+									Operator: corev1.NodeSelectorOpExists,
 								},
 								{
-									NodeSelectorRequirement: corev1.NodeSelectorRequirement{
-										Key:      spreadLabel,
-										Operator: corev1.NodeSelectorOpIn,
-										Values:   []string{spreadDomain},
-									},
+									Key:      spreadLabel,
+									Operator: corev1.NodeSelectorOpIn,
+									Values:   []string{spreadDomain},
 								},
 							},
 						},
@@ -1709,17 +1705,13 @@ var _ = Describe("Topology", func() {
 						Spec: v1.NodeClaimTemplateSpec{
 							Requirements: []v1.NodeSelectorRequirementWithMinValues{
 								{
-									NodeSelectorRequirement: corev1.NodeSelectorRequirement{
-										Key:      v1.CapacityTypeLabelKey,
-										Operator: corev1.NodeSelectorOpExists,
-									},
+									Key:      v1.CapacityTypeLabelKey,
+									Operator: corev1.NodeSelectorOpExists,
 								},
 								{
-									NodeSelectorRequirement: corev1.NodeSelectorRequirement{
-										Key:      spreadLabel,
-										Operator: corev1.NodeSelectorOpIn,
-										Values:   []string{spreadDomain, "2", "3"},
-									},
+									Key:      spreadLabel,
+									Operator: corev1.NodeSelectorOpIn,
+									Values:   []string{spreadDomain, "2", "3"},
 								},
 							},
 						},


pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -22,51 +22,76 @@ import (
...
 			}
 		}
+		if !taintMatched {
+			return false
+		}
 	}
+
+	if filter.AffinityPolicy == v1.NodeInclusionPolicyHonor {
+		for _, reqs := range d.requirements {
+			if filter.matchesRequirements(reqs) {
```
DerekFrank, line 90: If I am reading this right, this means that if the pod matches any of the requirements, we should call f. Should this be if the pod matches all of the requirements?

tal19987: In this struct, each entry in `d.requirements` is a full `scheduling.Requirements` set that we recorded while building the domain group, effectively one possible node configuration that can realize this topology domain.
With `AffinityPolicy=Honor` we want to keep a domain if the pod is compatible with any one of those configurations, which is why matches returns `true` on the first match.
If we required them all to match, we’d drop domains that are shared between a compatible and an incompatible `NodePool` even though there is schedulable capacity in that domain.

DerekFrank: Makes sense. I think I buy the fix, let me run it by someone else quick to sanity check the logic

DerekFrank:
Alright, I've chatted very briefly with @jmdeal and we're gonna review the perf implications sometime next week. Scheduling is *the* performance hotspot for Karpenter so we're going to be cautious. 

If you have the time, supplementing the scheduling benchmarking would help us gain confidence in the change

tal19987:
@DerekFrank 
I added a benchmark specifically for this use case (`BenchmarkTopologySpreadHonor`) and ran it on both `main` branch (baseline) and my fix branch (`fix-topology-spread-nodeaffinitypolicy`).

To reduce noise and get statistically meaningful results, I ran the full benchmark (that includes my benchmark as well) suite 20 times with:
```
# baseline (main)
go test -tags=test_performance -run=1 -bench=. -count=20 -cpuprofile output/cpu_main.out -memprofile output/mem_main.out ./pkg/controllers/provisioning/scheduling/ > output/main_full_20.txt

# fix branch
go test -tags=test_performance -run=1 -bench=. -count=20 -cpuprofile output/cpu_fix.out -memprofile output/mem_fix.out ./pkg/controllers/provisioning/scheduling/ > output/fix_full_20.txt
```

I then used `benchstat` to compare `main` vs `fix-topology-spread-nodeaffinitypolicy`:

```
export PATH=$PATH:$(go env GOPATH)/bin && benchstat output/main_full_20.txt output/fix_full_20.txt > benchtest.txt
```

**Conclusion:** No statistically significant performance regression was observed.
You can see the result here: (I can attach screenshots as well)
[benchtest.txt](https://github.com/user-attachments/files/23943889/benchtest.txt)

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -22,51 +22,76 @@ import (
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 )
 
+type topologyDomain struct {
```
jmdeal: I think we should call this something like `topologyDomainAttributes`, the topology domain is the string in the original map (e.g. `us-west-2a` would be a domain for a `topology.kubernetes.io/zone` topology).

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
jmdeal: That logic sounds right to me, it's the same reason we store a set of taint sets rather than anding all of them together. 👍

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
tal19987: Done. Renamed to `topologyDomainAttributes` and updated in relevant places

PR #2642, opened 2025-11-18
fix: Add capacity type req to nodeclaim based on offerings capacity types

### LINKED ISSUE #2638: Consolidation controller forces spot capacity requirement breaking providers without guaranteed spot availability
### Description

**Observed Behavior**:

We have a custom cloud provider where spot capacity availability is not guaranteed. Consolidation from an expensive on-demand instance to a cheaper on-demand instance fails unless users explicitly specify capacity-type requirements in their NodePool.

When no capacity-type requirement is specified in the NodePool:

1. Karpenter correctly identifies an underutilized node and generates a replacement NodeClaim with a smaller instance type
2. The consolidation controller adds a spot capacity-type requirement to the replacement NodeClaim (see code snippet below)
3. Our provider rejects/deletes the NodeClaim because spot is not available for that instance type
4. Consolidation fails with error: `replacement was deleted, NodeClaim "..." not found`

The [code in `disruption/consolidation.go`](https://github.com/kubernetes-sigs/karpenter/blob/a1357ee7fbf3fcc814c89e06a73df21c36453dae/pkg/controllers/disruption/consolidation.go#L215-L230) forces the replacement NodeClaim to be spot capacity-type since it assumes Spot capacity is allowed (even if there's no valid offerings)

```
// We are consolidating a node from OD -> [OD,Spot] but have filtered the instance types by cost based on the  
// assumption, that the spot variant will launch. We also need to add a requirement to the node to ensure that if  
// spot capacity is insufficient we don't replace the node with a more expensive on-demand node. Instead the launch  
// should fail and we'll just leave the node alone. We don't need to do the same for reserved since the requirements  
// are injected on by the scheduler.  
ctReq := results.NewNodeClaims[0].Requirements.Get(v1.CapacityTypeLabelKey)  
if ctReq.Has(v1.CapacityTypeSpot) && ctReq.Has(v1.CapacityTypeOnDemand) {  
results.NewNodeClaims[0].Requirements.Add(scheduling.NewRequirement(v1.CapacityTypeLabelKey, corev1.NodeSelectorOpIn, v1.CapacityTypeSpot))  
}  
  
return Command{  
Candidates: candidates,  
Replacements: replacementsFromNodeClaims(results.NewNodeClaims...),  
Results: results,  
}, nil  
}
```

**Logs showing the issue**:

```
{"level":"INFO","message":"disrupting node(s)","reason":"underutilized","decision":"replace","disrupted-nodes":[{"capacity-type":"on-demand","instance-type":"m5.24xlarge"}],"replacement-nodes":[{"capacity-type":"spot","instance-types":"m5.16xlarge"}]}

{"level":"ERROR","message":"failed terminating nodes while executing a disruption command","error":"replacement was deleted, NodeClaim \"defaultnodepool-h5mwh\" not found"}
```

**Workaround**: Consolidation works when customers explicitly specify capacity-type in NodePool:

```
requirements:
	  operator: In  
	  values: ["on-demand"]
```

**Expected Behavior**:

Consolidation should work without requiring user to specify capacity-type requirements when:

- The original node is on-demand
- The provider only supports on-demand capacity for specific instance types
- The replacement instance type is cheaper

The consolidation controller should respect the capacity-type offerings available from the cloud provider and not force spot requirements when spot is not available or guaranteed.

**Reproduction Steps** (Please include YAML):

1. Use a custom cloud provider with dynamic capacity configuration where spot availability is not guaranteed for all instance types
2. Create a NodePool without capacity-type requirements:

```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: defaultnodepool
spec:
  template:
    spec:
      requirements:
          operator: Exists
      nodeClassRef:
        group: custom.provider.io
        kind: CustomNodeClass
        name: defaultnodeclass
```

3. Scale up workload that creates an on-demand node (e.g., m5.24xlarge) where only on-demand capacity is available

4. Configure provider to offer a smaller, cheaper instance type (e.g., m5.12xlarge) with only on-demand capacity
   
5. Reduce workload to make the large node underutilized

6. Observe consolidation attempts to create spot instance even though spot is not available, causing the NodeClaim to be deleted and consolidation to fail

**Root Cause Analysis**:

The consolidation logic makes two problematic assumptions:

1. **Spot capacity is always available**: The code assumes that if spot appears in the initial NodeClaim requirements, it will be available for the replacement instance type
2. **Spot is always cheaper**: The logic defaults to spot without performing actual cost comparison across capacity types for the specific instance being consolidated

This works for providers where spot capacity is universally available, but fails for providers with dynamic capacity models where instance types can be configured with different capacity type offerings at runtime.

**Proposed Solution**:

The fix should involve two main components: (1. MVP, 2. would fix any edge cases as well)

1. Enhance NodeClaim Creation

During node claim creation, inject capacity type requirements based on actually available offerings:

- Examine available offerings for compatible instance types
- Add the intersection of available capacity types as requirements on the NodeClaim
- This ensures the NodeClaim only includes capacity types that the provider can actually fulfill

2. Improve Consolidation Decision Logic

Before forcing spot requirement, validate:

- Is spot capacity actually available for the replacement instance type?
- Is spot actually cheaper than on-demand for this specific consolidation?
- If the original node was on-demand and no capacity preference was specified, maintain on-demand unless spot is confirmed available and cheaper

We're willing to contribute an upstream fix via PR.

**Versions**:

- Karpenter Version: v1.0+ (consolidation controller)
- Kubernetes Version: v1.31.13-eks-3cfe0ce
- Cloud Provider: Custom provider with dynamic capacity configuration

AshwinS27:
Fixes #2638

**Description**
This change checks all the offerings for available capacityTypes. If there's only 1 capacity-type offerings available it adds that as a requirement to the nodeclaim.

This change would allow for OD --> OD consolidation when spot instance types are not available.

N/A 
Need more guidance for this part.


AshwinS27:
/check-cla
`/check-cla`

AshwinS27:
the presubmit tests might be flaky? Running `K8S_VERSION=1.32.x make presubmit` locally worked for me

PR #2644, opened 2025-11-19
fix: initialize npMap in metrics cost controller to prevent nil map panic

jamesmt-aws:
fix: Initialize npMap in metrics cost controller to prevent nil map panic. Fixes issue introduced in commit a1357ee7 (PR #2584)

**Description**
- Add unit test to reproduce nil map panic in metrics.cost controller
- Initialize npMap field in NewController constructor
- Fixes crash on every reconciliation attempt

Simulations and local unit tests crashed before the fix, and unit tests pass after the fix.


jmdeal:
Bypassing merge-blocker - codeql check has been fixed on main.

PR #2645, opened 2025-11-20
test: Add multi node consolidation metrics

clbar-aws:
**Description**

This PR adds three new Prometheus metrics to monitor the efficiency and performance of the binary search algorithm used in multi-node consolidation. These metrics provide visibility into:

- How many iterations the binary search takes to find optimal consolidation batches
- The actual size of consolidation batches achieved
- Time wasted on failed consolidation attempts

Validated using KWOK provider by scaling a deployment from 50→5 replicas, which triggered multi-node consolidation that successfully consolidated 2 nodes. All three metrics correctly reported values showing 1 iteration per batch, average batch size of 2 nodes, and zero failed simulations.

## Example Output

```javascript
# HELP karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations Alpha metric prone to change. Number of binary search iterations to find optimal multi-node consolidation batch.
# TYPE karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations histogram
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="1"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="2"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="3"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="4"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="5"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="6"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="7"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_bucket{le="+Inf"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_sum 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_iterations_count 2

# HELP karpenter_voluntary_disruption_multi_node_consolidation_batch_size Alpha metric prone to change. Number of nodes in successful multi-node consolidation batch, labeled with decision type.
# TYPE karpenter_voluntary_disruption_multi_node_consolidation_batch_size histogram
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="2"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="3"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="4"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="5"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="7"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="10"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="15"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="20"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="25"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="30"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="40"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="50"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_bucket{decision="delete",le="+Inf"} 2
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_sum{decision="delete"} 4
karpenter_voluntary_disruption_multi_node_consolidation_batch_size_count{decision="delete"} 2

```


pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -124,10 +124,15 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
 	}
 
 	lastSavedCommand := Command{}
+	iterationCount := 0
+	var failedIterationTime time.Duration
+
```
jamesmthompson, line 130: Imagine a case where the binary search succeeds after two rounds. So the whole process could be encoded as FFS, for Fail Fail Succeed. The time for F and S are tracked separately under the current design, so FFS increments both timers even though the aggregate operation eventually succeeded. 

So FFS today, there's allocation of time to both the F timer and the S timer. But I could imagine an argument where the metrics aren't per-call, but are per-search. So FFFFFFF would be a fail, but S, FS, FFS, etc would all be successes. What's the right thing to do from an analysis point of view with FFS?

jamesmthompson: I wonder if we want something more like total_failed_iteration_duration, and then have a second metric called something like wasted_search_duration that we only increment when lastSavedCommand.Decision() == NoOpDecision.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -162,9 +167,28 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
+	}
+	if lastSavedCommand.Decision() != NoOpDecision {
+		batchSize := len(lastSavedCommand.Candidates)
+		decision := "delete"
+		if lastSavedCommand.Decision() == ReplaceDecision {
+			decision = "replace"
+		}
+		MultiNodeConsolidationBatchSize.Set(float64(batchSize), map[string]string{
+			decisionLabel: decision,
+		})
```
jamesmthompson: Right now we won't set the batchsize if there's a NoOpDecision. Is that the right thing? Or should we still emit a zero for batchsize in that case?

jamesmthompson (commented):
Thanks for sending this out! I had a couple of questions about how we're thinking about using these.

pkg/controllers/disruption/metrics.go
```diff
@@ -108,4 +108,36 @@ var (
...
+		prometheus.HistogramOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: voluntaryDisruptionSubsystem,
+			Name:      "multi_node_consolidation_iterations",
+			Help:      "Number of binary search iterations to find optimal multi-node consolidation batch.",
+			Buckets:   []float64{1, 2, 3, 4, 5, 6, 7},
+		},
+		[]string{},
+	)
+	MultiNodeConsolidationBatchSize = opmetrics.NewPrometheusGauge(
```
DerekFrank: I think we want this to be a histogram instead of a gauge

pkg/controllers/disruption/multinodeconsolidation.go
Clabes: I think we only want to record successful non-zero batch sizes so as to better pick a starting point for the binary search (we obviously wouldn't pick zero), I could see an argument for tracking the zeros though...

pkg/controllers/disruption/multinodeconsolidation.go
Clabes: We could add a label for the Op decision "success" or "failure" as the final decision to allow decomposing both of the metrics. So we'd have (FFS time in F, "success") and (FFF time, "failure"). In some sense for FFS the time spent FF is still time wasted.

clbar-aws (commented):
Changed from gauge to histogram for successful batch size metric. Added indicator of simulation success/failure with the time spent in failed simulations

PR #2646, opened 2025-11-20
fix: Update label for deleted node

SwechchhaSinha:
fix: Change the label added on the deleted node

Description

When a node is marked to be deleted by Karpenter, the label value for Kubernetes exclude from loadbalancers is set to karpenter, instead of usual value of true. This causes the node to become  a part of the load balancer group.

How was this change tested?

Update the change in our forked repository and soaked and tested.


jmdeal:
I haven't found any documentation that indicates this needs to be true, and the implementations I've checked validate the label exists rather than the value. Do you have an example of where this doesn't work as expected?

/triage needs-information

PR #2656, opened 2025-11-23
docs(rfc): add grace period support for do-not-disrupt pod annotation

### LINKED ISSUE #752: Add a gracePeriod for the `do-not-evict` pod annotation

moko-poi:
Fixes #752

**Overview**
This RFC proposes adding a grace period mechanism to the `karpenter.sh/do-not-disrupt` pod annotation to address the challenge of pods indefinitely blocking node disruption while still
protecting long-running workloads during their expected execution time.

**Problem Statement**
Currently, the `karpenter.sh/do-not-disrupt: "true"` annotation provides indefinite protection, which can:
- Block critical cluster maintenance operations
- Prevent timely security updates and node replacements
- Leave cluster administrators with limited control over disruption windows

**Proposed Solution**
Add a new annotation `karpenter.sh/do-not-disrupt-grace-period` (specified in seconds) that:
- Protects pods for a specified duration from creation time
- Automatically allows disruption after the grace period expires
- Maintains backward compatibility (no grace period = indefinite protection)

Example:
```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    karpenter.sh/do-not-disrupt: "4h"  # 4 hours
```

### DIFF
--- a/designs/do-not-disrupt-grace-period.md
+++ b/designs/do-not-disrupt-grace-period.md
@@ -26,7 +26,7 @@ While this might appear to be a concern for higher-level abstractions (PodDisrup
 2. **Separation of Concerns**: This feature addresses **two distinct disruption scenarios** ([#752](https://github.com/kubernetes-sigs/karpenter/issues/752#issuecomment-2109882683)):
    - **Cost-optimization disruptions** (consolidation): Workload owners should have full control to block these
    - **Mandatory maintenance disruptions** (security updates, node replacements): Cluster administrators need eventual override capability
-   
+
    The grace period provides workload owners with a way to declare maximum protection time, preventing indefinite blocking while still allowing necessary protection during normal execution.
 
 3. **Beyond Job Abstractions**: Many workloads requiring protection are **not Kubernetes Jobs**:
@@ -79,7 +79,7 @@ Extend the existing `karpenter.sh/do-not-disrupt` annotation to accept duration 
 #### Annotation Values
 
 - `"true"`: Indefinite protection (existing behavior, backward compatible)
-- Duration string: Protection for the specified duration from pod creation time (e.g., `"4h"`, `"30m"`, `"1h30m"`)
+- Duration string: Protection for the specified duration from pod binding time (e.g., `"4h"`, `"30m"`, `"1h30m"`)
   - Follows Go's `time.Duration` format
   - Common examples: `"30m"`, `"1h"`, `"4h"`, `"24h"`, `"1h30m"`
 
@@ -100,7 +100,7 @@ spec:
 
 1. **Indefinite Protection**: If `karpenter.sh/do-not-disrupt: "true"` is set, the behavior remains the same as today - indefinite protection (backward compatible)
 2. **Time-Limited Protection**: If set to a duration value (e.g., `"4h"`):
-   - The pod is protected from disruption for the specified duration starting from the pod's creation time
+   - The pod is protected from disruption for the specified duration starting from the pod's binding time
    - After the grace period expires, the pod is treated as if it doesn't have the do-not-disrupt annotation
    - The node becomes eligible for disruption if no other constraints prevent it
 3. **Invalid Values**: If the value cannot be parsed as either `"true"` or a valid duration, it is treated as indefinite protection (backward compatible, fail-safe behavior)
@@ -109,10 +109,10 @@ spec:
 
 The grace period expiration time is calculated as:
 ```
-expiration_time = pod.CreationTimestamp + parsed_duration
+expiration_time = pod.BoundTimestamp + parsed_duration
 ```
 
-For example, if a pod is created at `2024-01-01T10:00:00Z` with `karpenter.sh/do-not-disrupt: "4h"`, it will be protected until `2024-01-01T14:00:00Z`.
+For example, if a pod is bound at `2024-01-01T10:00:00Z` with `karpenter.sh/do-not-disrupt: "4h"`, it will be protected until `2024-01-01T14:00:00Z`.
 
 ### Implementation Details
 
@@ -124,18 +124,18 @@ func parseDoNotDisrupt(value string) (indefinite bool, duration time.Duration, e
     if value == "true" {
         return true, 0, nil
     }
-    
+
     d, err := time.ParseDuration(value)
     if err != nil {
         // Invalid format - treat as indefinite (fail-safe)
         return true, 0, nil
     }
-    
+
     if d <= 0 {
         // Zero or negative - treat as indefinite (fail-safe)
         return true, 0, nil
     }
-    
+
     return false, d, nil
 }
 ```
@@ -175,7 +175,7 @@ spec:
 ```
 
 **Behavior**:
-- For the first 2 hours after pod creation, the node hosting this job will not be disrupted
+- For the first 2 hours after pod binding, the node hosting this job will not be disrupted
 - After 2 hours, if the job hasn't completed, the node becomes eligible for disruption
 - This provides protection for normal execution while ensuring the node can eventually be disrupted
 
@@ -249,11 +249,11 @@ This feature is complementary to, but distinct from, NodeClaim's `spec.terminati
 | Feature | Scope | Purpose | Set By | Behavior |
 |---------|-------|---------|--------|----------|
 | `karpenter.sh/do-not-disrupt: "4h"` | **Pod-level** | Self-expiring protection | Workload owner | "Protect this specific pod for 4 hours, then allow disruption" |
-| `spec.terminationGracePeriod` | **NodeClaim-level** | Eventual disruption guarantee | Cluster admin | "Disrupt this node after N hours regardless of pods" |
+| `spec.terminationGracePeriod` | **NodeClaim-level** | Eventual disruption guarantee | Cluster admin | "Terminate this node after N hours regardless of pods" |
 
 **Key Differences**:
 
-1. **Granularity**: 
+1. **Granularity**:
    - `do-not-disrupt` duration: Per-pod protection with automatic expiration
    - `terminationGracePeriod`: Global override affecting all pods on a node
 
@@ -270,6 +270,18 @@ This feature is complementary to, but distinct from, NodeClaim's `spec.terminati
 When both mechanisms are active:
 1. Pod-level protection expires first → Node becomes eligible for disruption if no other blocking pods
 2. NodeClaim termination grace period expires → Node is forcefully disrupted even with blocking pods (eventual disruption mode)
+
+**Eventual Disruption Interaction:**
+
+To ensure drift operations are not indefinitely blocked by application constraints, drift is considered an "eventual
+disruption type". What this means is that nodes with blocking PDBs or `do-not-disrupt` annotations **will** be
+considered valid disruption candidates if `terminationGracePeriod` is configured. This behavior will remain unchanged
+by this RFC.
+
+If a pod with a time-bound `do-not-disrupt` annotation is present on a draining node, that pod will not be drained until
+one of two conditions are met:
+- The pod's `do-not-disrupt` period elapses
+- The node's `terminationGracePeriod` elapses
 
 **Example Scenario**:
 ```yaml
@@ -289,9 +301,13 @@ metadata:
 ```
 
 Timeline:
-- **0-4 hours**: Pod is protected, node cannot be disrupted
-- **4-72 hours**: Pod protection expired, node eligible for disruption (respects PDBs in graceful mode)
-- **After 72 hours**: Node forcefully disrupted regardless of pods (eventual disruption mode)
+
+| t (hours) | Description|
+|-|-|
+|0|Pod is scheduled, pod will be protected until t=4.|
+|2|The node is disrupted by drift.|
+|4|The pod's grace period elapses and the pod is drained.|
+|4-72|Remaining pods are drained and the node is terminated. If all pods can not be gracefully drained by t=72, they are forcibly drained and the node is terminated.|
 
 This layered approach provides:
 - **Workload owners**: Ability to protect their workloads during expected execution
@@ -316,7 +332,7 @@ The annotation controls **when** a pod can be disrupted. Pod's `terminationGrace
 
 Users can determine the grace period status by:
 1. Checking pod annotations for the grace period value
-2. Comparing current time against `pod.CreationTimestamp + grace_period`
+2. Comparing current time against `pod.BindingTimestamp + grace_period`
 3. Observing Karpenter events when nodes become eligible for disruption
 
 Potential future enhancements:
@@ -329,7 +345,7 @@ Potential future enhancements:
 
 Set grace period at the node level instead of pod level.
 
-**Rejected because**: 
+**Rejected because**:
 - Less granular control
 - Doesn't align with the pod-level nature of do-not-disrupt annotation
 - Multiple pods on a node may have different requirements
@@ -362,6 +378,8 @@ Have a controller automatically remove the do-not-disrupt annotation after a per
 - Modifying user-defined annotations can be confusing
 - Doesn't preserve the original intent in pod spec
 - More complex implementation
+
+Of course, users are still able to perform this type of orchestration if they have more complex disruption requirements.
 
 ### Alternative 5: Rely on Higher-Level Abstractions
 


designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,249 @@
...
+
+Currently, when a pod has the `karpenter.sh/do-not-disrupt: "true"` annotation, it indefinitely prevents Karpenter from disrupting the node where the pod is running. This creates several issues:
+
+1. **Cluster Maintenance Challenges**: Workload owners can inadvertently or intentionally block critical cluster maintenance operations (security updates, node replacements, etc.)
+2. **No Upper Bound**: Long-running jobs (CI/CD pipelines, ML training) need protection during execution, but there's no mechanism to automatically lift this protection after a reasonable time
+3. **Administrative Control**: Cluster administrators have limited control over how long a workload can block node disruption
+
+### Use Cases
+
+1. **CI/CD Pipelines**: Jobs that should not be interrupted but have a known maximum runtime (e.g., 4 hours)
```
ellistarn, line 49: Does it make more sense to use `activeDeadlineSeconds` for this https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup?

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,249 @@
...
+2. **ML Training Jobs**: Long-running training that should complete uninterrupted but has an expected duration
+3. **Batch Processing**: Jobs that should run to completion but have predictable execution times
+4. **Cluster Maintenance**: Ensuring that even with do-not-disrupt pods, nodes can eventually be disrupted for mandatory updates
```
ellistarn: The way drift works today, your node will be tainted, allow work to complete, and eventually the node will be removed. 

See:
> When configured, a node may be disrupted via drift even if there are pods with blocking PDBs or the karpenter.sh/do-not-disrupt annotation scheduled to it

https://karpenter.sh/docs/concepts/disruption/#terminationgraceperiod

cc: @jmdeal

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,249 @@
+# Do-Not-Disrupt Grace Period
```
ellistarn, line 1: Can we rename this PR to `RFC: ...`

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,249 @@
...
+1. **CI/CD Pipelines**: Jobs that should not be interrupted but have a known maximum runtime (e.g., 4 hours)
+2. **ML Training Jobs**: Long-running training that should complete uninterrupted but has an expected duration
+3. **Batch Processing**: Jobs that should run to completion but have predictable execution times
+4. **Cluster Maintenance**: Ensuring that even with do-not-disrupt pods, nodes can eventually be disrupted for mandatory updates
+
+## Proposal
+
+### API Design
+
+Add a new annotation `karpenter.sh/do-not-disrupt-grace-period` that specifies the duration (in seconds) for which the `karpenter.sh/do-not-disrupt` protection should remain active.
```
ellistarn: This strikes me as somewhat backwards, as jobs get more expensive to disrupt the longer they run. More work will need to be redone. I could see a use case for allowing disruption for a period at the beginning of a job, but this doesn't appear to be your problem.

designs/do-not-disrupt-grace-period.md
moko-poi: This question was also raised in #752 by @njtran. @stevehipwell clarified the use case in https://github.com/kubernetes-sigs/karpenter/issues/752#issuecomment-1790110717:

> "for the use case I'm referring to the pods are explicitly not jobs and are expensive to re-run, for example a CI/CD system running E2E tests where terminating the pod would have a high
>  cost."

activeDeadlineSeconds is a Job-specific feature and cannot be applied to regular Pods. Many CI/CD systems (Jenkins agents, Tekton tasks, custom test runners) run as
standalone Pods rather than Jobs due to:
- Integration with existing orchestration systems
- Need for custom lifecycle management
- Requirements for persistent connections or stateful operations

These workloads still need temporary protection from disruption but shouldn't be able to block node maintenance indefinitely.

designs/do-not-disrupt-grace-period.md
moko-poi: You're correct that terminationGracePeriod provides eventual disruption. However, this proposal addresses a different concern raised in #752:

The key distinction:
- terminationGracePeriod (NodeClaim-level): Global setting - "disrupt this node after N hours, no matter what"
- This proposal (Pod-level): Self-expiring protection - "protect this pod for N hours, then auto-release"

Why both are needed:

terminationGracePeriod requires cluster admins to set a fixed timeout for all disruption scenarios. This proposal allows workload owners to declare protection that automatically
expires, preventing indefinite blocking without requiring:
- Manual intervention to remove stuck annotations
- External policy engines (OPA/Kyverno)
- Trust that applications will clean up properly

As noted in https://github.com/kubernetes-sigs/karpenter/issues/752#issuecomment-1790110688:
"having pods stuck due to user error and blocking node maintenance is a common problem when running a platform"

The grace period annotation provides declarative, time-bound protection at the Pod level, complementing the NodeClaim-level safety net of terminationGracePeriod.

Would it help to add a section in the design doc clarifying how these features work together?

ellistarn: I understand the proposed mechanism. I don't fully understand the requirements that lead to this solution. Curious to your thoughts on my comment here: https://github.com/kubernetes-sigs/karpenter/pull/2656#discussion_r2554263500. 

Per https://github.com/kubernetes-sigs/karpenter/pull/2656#discussion_r2554486933, wouldn't it follow that these systems should support something similar? If they are making the pods manually, they will also be required to implement the appropriate retries. Shimming a soft deadline concept into Karpenter smells like misplaced responsibility to me. I understand the competing layers and how Karpenter ends up being a pragmatic point of leverage for platform engineering teams. Realistically, I wish this were just a feature of `PodDisruptionBudget`. 

To be clear, this is in no way a hard no, I just want to make sure I understand the use case before we decide this is the right solution.

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,249 @@
...
+#### Annotation
+
+```yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: long-running-job
+  annotations:
+    karpenter.sh/do-not-disrupt: "true"
+    karpenter.sh/do-not-disrupt-grace-period: "14400"  # 4 hours in seconds
```
ellistarn: If we do support this, I think this might be better modeled as a go duration in the same field.

```
karpenter.sh/do-not-disrupt: 1d
```

designs/do-not-disrupt-grace-period.md
moko-poi: @ellistarn Thank you for the detailed feedback. I've added a comprehensive comparison section in the RFC that clarifies how these features work together. The key insight is that they serve different purposes:

terminationGracePeriod: Admin-controlled "ultimate deadline" - ensures nodes can eventually be disrupted regardless of workload state
This proposal: Workload-controlled "self-expiring protection" - allows workload owners to declare reasonable protection windows that automatically expire

The scenario we're addressing from #752:
1. Developer adds `do-not-disrupt: "true"` to protect a 2-hour CI job
2. Job hangs/crashes but pod doesn't terminate (not using Job controller)
3. Node is now indefinitely blocked from disruption
4. Platform team must manually intervene to find and remove the annotation

With grace period: `do-not-disrupt: "4h"` - the developer declares "this should never run longer than 4 hours," preventing indefinite blocking while still protecting normal execution.

The RFC now includes a detailed table and timeline showing how both mechanisms complement each other. Would appreciate your thoughts on whether this clarifies the distinction.

designs/do-not-disrupt-grace-period.md
moko-poi: @ellistarn Excellent point - I initially framed this incorrectly. This isn't about jobs becoming cheaper to disrupt over time. I've reframed the Use Cases section to clarify:

This is about establishing a maximum protection duration, not a cost-based calculation.

The real problem we're solving:
- Stuck workloads: A workload owner sets `do-not-disrupt: "true"`, then the pod hangs, crashes, or they simply forget to remove the annotation → node is blocked indefinitely
- No upper bound: Currently there's no way for a workload owner to say "protect this for 4 hours maximum, then allow disruption regardless"

As noted in #752: *"having pods stuck due to user error and blocking node maintenance is a common problem when running a platform"*

The grace period provides a declarative way to say "this workload should never need protection longer than X hours" - it's a safety valve against indefinite blocking, not a statement about disruption cost over time.

The updated RFC now emphasizes:
1. Preventing stuck workloads - automatic cleanup without manual intervention  
2. Maximum protection duration - "this should never run longer than 4 hours"
3. Maintenance planning - cluster admins can plan maintenance windows with confidence

Does this better explain the motivation?

designs/do-not-disrupt-grace-period.md
moko-poi: @ellistarn Great suggestion! I've updated the RFC to use a single annotation with Go duration format as you recommended.

ellistarn:
Hey @moko-poi, I am increasingly convinced that this is a valuable and well scoped feature to help with the Cost/Availability tradeoff. I want to get a nod from @jmdeal and @jamesmt-aws as well.

stevehipwell:
@moko-poi thanks for picking this up, you've done a great job on the RFC. I think this covers my primary requirements from #752 and is actually a better solution.

May I add a suggestion that Karpenter adds a label to the nodes that are being protected by this feature to enable context aware scheduling?

moko-poi:
@stevehipwell Thanks for the suggestion! 

Node labeling is an interesting idea for scheduler integration, but I think it's a separate concern from the core grace period feature. The protection mechanism works entirely within Karpenter's disruption controller by checking pod annotations.

Could we treat this as a potential follow-up enhancement? I'd like to see real-world usage patterns first to understand if existing mechanisms (custom admission controllers, scheduler extenders) are insufficient.

stevehipwell:
@moko-poi my requirement/suggestion is to allow end users to avoid scheduling on nodes that are going to be removed when their grace period expires. This probably requires a label to function, but there could also be a dynamic taint to stop accidental scheduling.

ellistarn:
> @moko-poi my requirement/suggestion is to allow end users to avoid scheduling on nodes that are going to be removed when their grace period expires. This probably requires a label to function, but there could also be a dynamic taint to stop accidental scheduling.

This is an interesting idea, but let's scope this to another discussion.

There's a whole category of labels in this space like:
- age (don't schedule if older than 30d)
- disruption-blocked (don't schedule if can't be terminated)
- drifted (don't schedule if drifted, but not yet tainted)

Unclear to me if we want to model this as a single label, multiple labels, single label with multiple possible values, etc. 

@stevehipwell, would you be open to an RFC?

jamesmthompson:
This is a thoughtful RFC addressing a real problem. One question: would a time-varying PodDisruptionBudget be a better fit architecturally? Something that could express "high disruption cost for hours 0-4, low cost after that."

If we wanted to go full galaxy brain, we could even have a negative cost (also known as a benefit) to catch cases where, based only on running time of the pod, it's much more likely that something has gone totally awry and we'd be better off with a replacement.

I realize PDBs are a K8s concept that is bigger than just Karpenter. I'm curious if this has been considered or if there's appetite to propose it upstream.

ellistarn:
> I realize PDBs are a K8s concept that is bigger than just Karpenter. I'm curious if this has been considered or if there's appetite to propose it upstream.

I do think that this annotation is essentially papering over gaps in PDB. There are number of gaps in PDB (i.e., you cannot model cost, time, or launch-before-terminate, etc, etc)

Historically, we've built this into Karpenter because its simply faster and easier. As these capabilities stack up, I do think we might want to revisit some work upstream. For now, I'm inclined to introduce the feature as proposed.

JeremyBolster:
> I do think that this annotation is essentially papering over gaps in PDB. There are number of gaps in PDB (i.e., you cannot model cost, time, or launch-before-terminate, etc, etc)

> Historically, we've built this into Karpenter because its simply faster and easier. As these capabilities stack up, I do think we might want to revisit some work upstream. For now, I'm inclined to introduce the feature as proposed.

We have had a similar issue with limitations in PDBs, and are trying to work around them by defining a meta-object that supports denying evictions for: Guaranteed / Minimum Pod Lifespan, red-green/scheduled budget sizes, disruption throughput / anti-thrashing. I am just mentioning in case anyone wants to reach out to collaborate on ideas :)

ellistarn:
> > I do think that this annotation is essentially papering over gaps in PDB. There are number of gaps in PDB (i.e., you cannot model cost, time, or launch-before-terminate, etc, etc)
> 
> > Historically, we've built this into Karpenter because its simply faster and easier. As these capabilities stack up, I do think we might want to revisit some work upstream. For now, I'm inclined to introduce the feature as proposed.
> 
> We have had a similar issue with limitations in PDBs, and are trying to work around them by defining a meta-object that supports denying evictions for: Guaranteed / Minimum Pod Lifespan, red-green/scheduled budget sizes, disruption throughput / anti-thrashing. I am just mentioning in case anyone wants to reach out to collaborate on ideas :)

Very interested! I think @jamesmt-aws might be as well. Are you imagining writing a KEP?

JeremyBolster:
I should write a KEP, I have a company internal document right now. I can try to get out a draft that I can share on slack by the end of the week for those interested.

ellistarn:
> I should write a KEP, I have a company internal document right now. I can try to get out a draft that I can share on slack by the end of the week for those interested.

Happy to review. I'm on the kubernetes slack if you need to reach me.

moko-poi:
Thanks everyone for the thorough discussion — this has been incredibly helpful.

Based on the feedback so far, I believe the core scope of this RFC is now clear and stable:
- Pod-level, self-expiring protection using a Go duration
- Explicitly complementary to NodeClaim terminationGracePeriod
- No node labeling / tainting in scope (to be discussed separately)
- Acknowledged overlap with PDB gaps, but pragmatic to land in Karpenter for now

Unless there are remaining blocking concerns, I’d like to move this RFC toward approval.

@jmdeal @jamesmt-aws — I’d really appreciate your thoughts or approval when you have a chance.

jamesmthompson:
> @jmdeal @jamesmt-aws — I’d really appreciate your thoughts or approval when you have a chance.

Apologies for missing this for a bit. This one has my approval to move forward (for whatever that is worth). Very excited to see where this goes!

moko-poi:
Hi @jmdeal @tallaxes @ellistarn, 

I'd appreciate your review when you have a chance. @jamesmt-aws and @ellistarn have approved the RFC, and I'm ready to move forward with implementation once we have your feedback.

Thanks!

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,423 @@
...
+- **Job limitation**: Many workloads requiring protection are not Jobs (CI/CD pods, custom runners)
+- **System diversity**: Different CI/CD systems have different architectures; requiring all to implement deadline logic is impractical
+- **Platform engineering reality**: Platform teams need defense mechanisms against misconfigured or stuck workloads
+- **PDB scope**: PodDisruptionBudget controls disruption policy but doesn't provide time-based expiration
+- **Pragmatic leverage point**: Karpenter is the natural place for node lifecycle controls to prevent indefinite resource blocking
+- As reviewer noted: "I understand the competing layers and how Karpenter ends up being a pragmatic point of leverage for platform engineering teams"
+
+While ideally this would be a Kubernetes-native PDB feature, the immediate problem space (platform teams managing node lifecycle with reasonable bounds on user-controlled blocking) makes Karpenter a suitable implementation point.
+
+## Testing
```
ellistarn: I'm not sure this section is necessary for the RFC. We can iron this out in the implementation phase

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,423 @@
+# Do-Not-Disrupt Grace Period
```
ellistarn, line 1: I am good to go with this RFC, but want a nod from @tzneal and @jmdeal .

ellistarn:
Your description is out of date with the RFC
```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    karpenter.sh/do-not-disrupt: "true"
    karpenter.sh/do-not-disrupt-grace-period: "3600"  # 1 hour
```

designs/do-not-disrupt-grace-period.md
moko-poi: @ellistarn Agreed. I've removed the Testing section from the RFC. 
We can determine the specific test coverage during implementation.
[eb4bca8](https://github.com/kubernetes-sigs/karpenter/pull/2656/commits/eb4bca874c637bbfbbc83517418c0438e119ba3e)

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,408 @@
...
+# Pod with 4-hour protection
+metadata:
+  annotations:
+    karpenter.sh/do-not-disrupt: "4h"
+```
+
+Timeline:
+- **0-4 hours**: Pod is protected, node cannot be disrupted
+- **4-72 hours**: Pod protection expired, node eligible for disruption (respects PDBs in graceful mode)
+- **After 72 hours**: Node forcefully disrupted regardless of pods (eventual disruption mode)
```
jmdeal: `terminationGracePeriod` only becomes active once the node has been disrupted, in this example would that happen at t=0 (simultaneously with the protected pod scheduling)? Also, could you elaborate on the interaction with the "eventual disruption" mode for drift? The behavior I'm referring to is where we'll initiate disruption due to drift even when there are blocking PDBs or `do-not-disrupt` annotations on bound pods. 

I think we want to maintain this behavior, even if it could result in the minimum lifetime set by the `do-not-disrupt` annotation being ignored. This is important to ensure cluster admins maintain controller over node lifecycle management. It would be good to inform application owners that their configuration may not be respected (maybe an event) though.

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,408 @@
...
+3. **Invalid Values**: If the value cannot be parsed as either `"true"` or a valid duration, it is treated as indefinite protection (backward compatible, fail-safe behavior)
+
+### Time Calculation
+
+The grace period expiration time is calculated as:
+```
+expiration_time = pod.CreationTimestamp + parsed_duration
+```
+
+For example, if a pod is created at `2024-01-01T10:00:00Z` with `karpenter.sh/do-not-disrupt: "4h"`, it will be protected until `2024-01-01T14:00:00Z`.
```
jmdeal: Do we want to use the pod creation timestamp, or do we want to use the time the pod was bound as the start time? I think we probably want to do the latter since that actually enforces a minimum execution time for the pod. By using the creation timestamp, if pods remain pending for an extended duration of time this could result in an unexpectedly low protection time. Of course we'd like pod pending times to remain low with Karpenter, but they could go up due to NodePool misconfigurations or insufficient capacity for compatible instance types.

designs/do-not-disrupt-grace-period.md
```diff
@@ -0,0 +1,408 @@
...
+### Alternative 3: Absolute Timestamp Annotation
+
+Use an absolute timestamp instead of duration.
+
+**Rejected because**:
+- Requires users to calculate absolute timestamps
+- More error-prone
+- Less intuitive than duration-based specification
+
+### Alternative 4: Controller-Managed Expiration
```
jmdeal, line 373: Worth noting that this is the current state of the world and this proposal won't prevent users from doing it if they have more complex disruption control requirements.

designs/do-not-disrupt-grace-period.md
ellistarn: Strong agree with the UX being around "execution time". This may be slightly surprising for users looking for a "max job lifetime" since it'll exceed that, but we'll just need to be clear in the docs.

jmdeal:
@moko-poi any updates? I think the only outstanding "major" change is this comment here: https://github.com/kubernetes-sigs/karpenter/pull/2656#discussion_r2709304421. If you don't have the bandwidth to update this but agree with the direction we can get it updated and move to merge this.

moko-poi:
Thank you so much @jmdeal for picking up the remaining updates and getting this merged\! I really appreciate you addressing the creation → binding time change and adding the eventual disruption interaction details. Sorry for the delay on my end — I'll make sure to be more responsive going forward. Looking forward to working on the implementation\!

PR #2663, opened 2025-11-25
test: fix flaking static provisioning test

DerekFrank:
Fixes https://github.com/kubernetes-sigs/karpenter/actions/runs/19683853291/job/56384371037

**Description**

If we call cluster.Reset() we get rid of our internal state, which causes race conditions. Since deleting cluster state is not indicative of how the code will function, I have switched it to instead delete nodeclaims regularly.

Ran the tests quite a few times locally, no flakes

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -592,6 +592,13 @@ func (c *Cluster) Reset() {
 	c.podsSchedulableTimes = sync.Map{}
 }
 
+// sets the cluster to be synced or unsynced for unit testing
+func (c *Cluster) SetSynced(state bool) {
+	c.unsyncedTimeMu.Lock()
+	defer c.unsyncedTimeMu.Unlock()
+	c.hasSynced.Store(state)
+}
+
 func (c *Cluster) GetDaemonSetPod(daemonset *appsv1.DaemonSet) *corev1.Pod {
 	if pod, ok := c.daemonSetPods.Load(client.ObjectKeyFromObject(daemonset)); ok {
 		return pod.(*corev1.Pod).DeepCopy()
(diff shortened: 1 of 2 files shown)


pkg/controllers/static/provisioning/suite_test.go
```diff
@@ -541,7 +541,11 @@ var _ = Describe("Static Provisioning Controller", func() {
 				go func(i int) {
 					defer GinkgoRecover()
 					if i%4 == 0 {
-						cluster.Reset()
+						// Delete the nodeclaims every so often
+						nodeClaimList := &v1.NodeClaimList{}
```
jmdeal: I don't think deleting NodeClaims is really in the spirit of this test either. Really I think, rather than trying to induce a race this way, it makes more sense to just force the cluster into an unsynced state and verify that the reconciler requeues without creating any claims.

DerekFrank: That makes sense, I can add a unit testing function to force the cluster to unsync

pkg/controllers/state/cluster.go
```diff
@@ -592,6 +592,13 @@ func (c *Cluster) Reset() {
 	c.podsSchedulableTimes = sync.Map{}
 }
 
+// sets the cluster to be unsynced for unit testing
+func (c *Cluster) Unsynced() {
```
ellistarn: Unsynced is a noun, which looks like a read-only api. Maybe change it to a verb `SetSynced(bool)`?

pkg/controllers/static/provisioning/suite_test.go
```diff
@@ -558,6 +558,8 @@ var _ = Describe("Static Provisioning Controller", func() {
 				return len(list.Items)
 			}, 5*time.Second).Should(BeNumerically("<=", 10))
 
+			_, e := controller.Reconcile(ctx, nodePool)
```
ellistarn: Isn't there a helper we can use like ExpectObjectReconciled or someth?

DerekFrank: Fair point, that makes a lot more sense

PR #2670, opened 2025-11-30
feat: Add detailed logging for consolidation blocking reasons

### LINKED ISSUE #2238: Add more context to "removing consolidatable status condition" message
### Description

we want to track down why a certain node was not considered for consolidation,
for example "pod xyz was not disruptable"

for example we have a 16xl node that is running almost empty but karpenter only says
```
"marking consolidatable"
"removing consolidatable status condition"
... loop forever
```
and no indication of what is going on
... because there is a pod with a bad pdb sitting on it

Medium: would allow us to track down why nodes are wasting so much resources

moko-poi:
Fixes #2238

**Description**

Adds detailed structured logging to `ValidatePodsDisruptable()` to help operators understand why nodes cannot be consolidated.

Previously, Karpenter would loop with generic "marking/removing consolidatable" messages when a node couldn't be consolidated. Now it logs specific details:
- Which pod is blocking consolidation
- The blocking reason (do-not-disrupt annotation or PDB violation)
- Which PDB is preventing eviction
- List of affected pods for PDB violations

**Example output:**
```
cannot consolidate node, PDB prevents eviction
  node="ip-10-0-1-100"
  pods=["default/mysql-0"]
  pdb="default/mysql-pdb"
  reason="pdb-blocks-eviction"
```

Verified compilation and existing tests pass. The change only adds logging without modifying behavior.

### DIFF
--- a/pkg/controllers/state/statenode.go
+++ b/pkg/controllers/state/statenode.go
@@ -241,13 +241,38 @@ func (in *StateNode) ValidatePodsDisruptable(ctx context.Context, kubeClient cli
 		// We only consider pods that are actively running for "karpenter.sh/do-not-disrupt"
 		// This means that we will allow Mirror Pods and DaemonSets to block disruption using this annotation
 		if !podutils.IsDisruptable(po) {
+			klog.FromContext(ctx).Info("cannot consolidate node, pod has do-not-disrupt annotation",
+				"node", in.Name(),
+				"nodeClaim", in.NodeClaim.Name,
+				"pod", klog.KObj(po),
+				"reason", "do-not-disrupt")
 			return pods, NewPodBlockEvictionError(serrors.Wrap(fmt.Errorf(`pod has "karpenter.sh/do-not-disrupt" annotation`), "Pod", klog.KObj(po)))
 		}
 	}
 	if pdbKeys, ok := pdbs.CanEvictPods(pods); !ok {
+		// Identify which specific pods are blocked by PDBs
+		blockedPods := []string{}
+		for _, po := range pods {
+			if _, evictable := pdbs.CanEvictPods([]*corev1.Pod{po}); !evictable {
+				blockedPods = append(blockedPods, fmt.Sprintf("%s/%s", po.Namespace, po.Name))
+			}
+		}
+
 		if len(pdbKeys) > 1 {
+			klog.FromContext(ctx).Info("cannot consolidate node, multiple PDBs block eviction",
+				"node", in.Name(),
+				"nodeClaim", in.NodeClaim.Name,
+				"pods", blockedPods,
+				"pdbs", pdbKeys,
+				"reason", "multiple-pdbs")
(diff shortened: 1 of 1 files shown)


moko-poi:
Hi @mwielgus @tallaxes,
Could you please take a look at this PR when you have a chance?

moko-poi:
Hi @mwielgus @tallaxes @ellistarn @grosser,

Friendly ping on this PR. It's been about a month since my last message.

This is a small logging-only change (+25 lines) that helps operators diagnose consolidation issues (fixes #2238). All tests are passing with 100% coverage.

Would appreciate a review when you have a moment. Thanks!

pkg/controllers/state/statenode.go
```diff
@@ -241,13 +241,38 @@ func (in *StateNode) ValidatePodsDisruptable(ctx context.Context, kubeClient cli
 		// We only consider pods that are actively running for "karpenter.sh/do-not-disrupt"
 		// This means that we will allow Mirror Pods and DaemonSets to block disruption using this annotation
 		if !podutils.IsDisruptable(po) {
+			klog.FromContext(ctx).Info("cannot consolidate node, pod has do-not-disrupt annotation",
+				"node", in.Name(),
+				"nodeClaim", in.NodeClaim.Name,
+				"pod", klog.KObj(po),
+				"reason", "do-not-disrupt")
 			return pods, NewPodBlockEvictionError(serrors.Wrap(fmt.Errorf(`pod has "karpenter.sh/do-not-disrupt" annotation`), "Pod", klog.KObj(po)))
```
ellistarn, line 249: We currently emit events when disruption is blocked at the caller context: https://github.com/kubernetes-sigs/karpenter/blob/c4559b16/pkg/controllers/disruption/types.go#L119. But we intentionally filter out these errors if they are "eventual" disruption candidates.

The general concern is that we want to balance user visibility with user spam, which reduces visibility.

moko-poi: @ellistarn 
Thank you for the detailed feedback. I understand your concern about balancing visibility with avoiding log spam, and I appreciate you pointing out the existing event publishing mechanism.

You're right that my current implementation logs all blocking cases at Info level, which doesn't respect the filtering logic for eventual disruption candidates (where TerminationGracePeriod is set). This could create excessive logs in large clusters, especially for transient PDB blocks that will eventually resolve.

I see a few possible approaches to address this:
1. Change the log level from Info to V(1) (debug level) so operators can enable it when troubleshooting specific consolidation issues, without having it on by default.
2. Instead of adding logs in ValidatePodsDisruptable(), enhance the existing Kubernetes event messages that are already published via recorder.Publish() in [types.go](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) to include more details about which pod and PDB is blocking.
3. Pass the disruptionClass context down to ValidatePodsDisruptable() and only log when it's not an eventual disruption candidate, matching the filtering logic that already exists.

Which approach would you prefer? Or if you have a different direction in mind, I'm happy to adjust accordingly.

PR #2671, opened 2025-11-30
fix: Filter topology domains by NodePool compatibility with pod requirements

### LINKED ISSUE #2227: All AZ domains included in topology spread constraints evaluation even when nodepool restricts to specific AZs
### Description

# All AZ domains included in topology spread constraints evaluation even when nodepool restricts to specific AZs

**Observed Behavior**:
When scheduling pods with topology spread constraints, Karpenter evaluates constraints against all availability zones known to the cluster, not just the ones defined in the nodepool being used. In my case, I have a nodepool "x" that explicitly restricts to 3 AZs (us-east-1a, us-east-1b, us-east-1c), and I use nodeSelector in my pods to explicitly target only this nodepool. However, the error message shows that Karpenter is considering all 6 AZs (a-f) in its topology calculations:

```
"incompatible with nodepool \"x\", daemonset overhead={\"cpu\":\"330m\",\"memory\":\"420Mi\",\"pods\":\"8\"}, unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = us-east-1a: 2, us-east-1b: 2, us-east-1c: 2, us-east-1d: 0, us-east-1e: 0, us-east-1f: 0, podDomains = topology.kubernetes.io/zone Exists, nodeDomains = topology.kubernetes.io/zone In [us-east-1a us-east-1b us-east-1c]"
```

The issue specifically occurs because I have other nodepools in the same cluster that do NOT have zone constraints. Karpenter aggregates domain information from all nodepools, so even though my target nodepool only allows 3 zones, the domain group includes all 6 zones from the unconstrained nodepools.

**Expected Behavior**:
When pods use nodeSelector to target a specific nodepool, Karpenter should only consider the AZs defined in that nodepool for topology spread constraint calculations. In this case, only us-east-1a, us-east-1b, and us-east-1c should be considered, since those are the only zones allowed by the nodepool "x" that the pod is explicitly targeting.

1. Have multiple nodepools in a cluster, with at least one nodepool that does not restrict AZs (this is key to reproducing the issue)
2. Create a nodepool with specific AZ restrictions:

```yaml
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: x
spec:
  template:
    metadata:
      labels:
        node.pubg.io/management-class: general
        node.pubg.io/node-type: x
        node.pubg.io/platform: linux
        nodeType: x
    spec:
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: default
      requirements:
        - key: karpenter.k8s.aws/instance-family
          operator: In
          values:
            - c7gn
          operator: In
          values:
            - us-east-1a
            - us-east-1b
            - us-east-1c
        - key: kubernetes.io/os
          operator: In
          values:
            - linux
          operator: In
          values:
            - on-demand
      taints:
        - effect: NoSchedule
          key: nodeType
          value: x
```

3. Deploy pods with topology spread constraints and nodeSelector targeting this nodepool:

```yaml
nodeSelector:
  nodeType: x
```

**Versions**:
- Chart Version: 1.3.2
- Kubernetes Version (`kubectl version`): v1.31

Looking at the code in `pkg/controllers/provisioning/scheduling/topology.go`, it appears that the `buildDomainGroups` function combines domain information from all nodepools into a single shared domain group. This causes topology spread constraints to be evaluated against all domains known to the cluster, not just the ones defined in the specific nodepool being used.

moko-poi:
Fixes #2227

**Description**

This PR fixes an issue where topology spread constraints incorrectly evaluated all cluster zones instead of only zones available in NodePools that match the pod's nodeSelector or nodeAffinity requirements.

**Problem:**
When a pod specifies a nodeSelector or nodeAffinity targeting a specific NodePool with limited zones, Karpenter's topology spread evaluation considered all zones across all NodePools in the cluster. This caused unnecessary scheduling failures when the topology spread constraint couldn't be satisfied within the pod's targeted NodePool's zone scope.

For example:
- NodePool A: zones [us-east-1a, us-east-1b]
- NodePool B: zones [us-east-1a, us-east-1b, us-east-1c]
- Pod with nodeSelector targeting NodePool A and maxSkew=1

Before this fix, Karpenter would try to spread across all 3 zones (a, b, c), even though the pod can only schedule to NodePool A which has 2 zones. This would cause the pod to fail scheduling when a 3-pod distribution couldn't be satisfied.

**Solution:**
Track NodePool requirements (labels + requirements) with each topology domain and filter domains based on compatibility with pod requirements during topology spread evaluation. This ensures that only zones from NodePools compatible with the pod's nodeSelector/nodeAffinity are considered.

**Changes:**
1. Added `DomainSource` struct to track NodePool requirements and taints per domain
2. Modified `TopologyDomainGroup.Insert()` to accept NodePool requirements
3. Modified `TopologyDomainGroup.ForEachDomain()` to filter domains by compatibility with pod requirements
4. Updated topology building logic to pass NodePool requirements through the stack

- Added 4 comprehensive test cases covering:
  - Pods with nodeSelector targeting specific zones
  - Pods with nodeAffinity targeting zone ranges
  - Pods without NodePool restrictions using all available zones
  - The exact scenario from Issue #2227 (maxSkew=1, 3 pods, 2-zone NodePool)
- All existing topology tests pass
- No compilation errors

### DIFF
--- a/pkg/controllers/provisioning/scheduling/topologydomaingroup.go
+++ b/pkg/controllers/provisioning/scheduling/topologydomaingroup.go
@@ -116,7 +116,7 @@ func (t TopologyDomainGroup) isDomainCompatible(pod *v1.Pod, sources []DomainSou
 		if affinityPolicy == v1.NodeInclusionPolicyHonor {
 			requirementsMatch = len(podRequirementSets) == 0
 			for _, podReqs := range podRequirementSets {
-				if err := source.NodePoolRequirements.Intersects(podReqs); err == nil {
+				if err := source.NodePoolRequirements.Compatible(podReqs, scheduling.AllowUndefinedWellKnownLabels); err == nil {
 					requirementsMatch = true
 					break
 				}


tzneal:
This doesn't seem correct to me. There is a mechanism to restrict topology domains, but its via node selector or required node affinity (see https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#interaction-with-node-affinity-and-node-selectors).

The scheduler isn't going to be aware of this special logic.

Taking your example:

> For example:
> NodePool A: zones [us-east-1a, us-east-1b]
> NodePool B: zones [us-east-1a, us-east-1b, us-east-1c]
> Pod with nodeSelector targeting NodePool A and maxSkew=1

Your pods are part of a deployment targeting NodePool A. Assume other node exist in the cluster for other reasons in all zones so the scheduler is aware that the zones exist. 

If Karpenter launches nodes for a required spread over 1a/1b, the scheduler will refuse to schedule them if it can't also schedule enough Pods to 1c.  There's nothing in the deployment spec to inform the scheduler of your new constraint.

moko-poi:
Thank you for the detailed feedback, @tzneal. You raise a critical point about the Kubernetes scheduler interaction that I need to think through more carefully.

Let me make sure I understand your concern correctly: Even if Karpenter filters domains based on NodePool compatibility, the Kubernetes scheduler will independently evaluate topology spread constraints against all domains visible from existing nodes in the cluster. This could create a mismatch where:

1. Karpenter provisions nodes only in zones [a, b, c] based on NodePool constraints
2. The scheduler sees existing nodes in zones [a, b, c, d, e, f] and expects pods to spread across all 6 zones
3. The scheduler rejects pod placements that Karpenter considers valid

My assumption was that when a pod has `nodeSelector: {nodeType: x}`, the Kubernetes scheduler would only consider nodes matching that selector when calculating available topology domains, per the [interaction with node selectors documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#interaction-with-node-affinity-and-node-selectors). However, I may be misunderstanding how this works in practice.

Questions:
1. Does the scheduler calculate topology domains globally first (from all nodes), then filter candidates? Or does it only consider nodes matching the pod's nodeSelector/nodeAffinity from the start?
2. If the former, should users always explicitly specify zone constraints in the pod's nodeAffinity (e.g., `topology.kubernetes.io/zone In [us-east-1a, us-east-1b, us-east-1c]`) in addition to NodePool selectors?
3. Would the correct solution be to document this requirement rather than changing Karpenter's behavior?

I'm happy to test this in a real cluster scenario with:
- Multiple NodePools with different zone constraints
- Existing nodes in various zones
- Pods with topology spread constraints + nodeSelector

This would help verify whether my implementation creates scheduler conflicts. Would that be a valuable next step, or do you already know this approach won't work based on the scheduler's design?

I appreciate your guidance on the right direction here.

tzneal:
Testing would be ideal, its quite possible I'm mis-remembering how this worked.

moko-poi:
Thank you for being open to testing! I'll set up a test scenario to verify the actual scheduler behavior and report back with findings.

I'll test with:
1. Multiple NodePools with different zone constraints
2. Pre-existing nodes in various zones (to ensure the scheduler sees all zones globally)
3. Pods with `nodeSelector` targeting specific NodePools + topology spread constraints
4. Monitor both Karpenter's provisioning decisions and the actual scheduler's placement decisions

moko-poi:
As an additional note, the Kubernetes documentation describes how topology domains are derived when nodeSelector or required nodeAffinity is used:

> "The scheduler will skip the non-matching nodes from the skew calculations if the incoming Pod has spec.nodeSelector or spec.affinity.nodeAffinity defined."

https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/#interaction-with-node-affinity-and-node-selectors

My interpretation is that the scheduler forms topology domains only from nodes that match the pod’s selector or required affinity, and zones containing only non-matching nodes are not included in the skew evaluation. I will verify this behavior in a real cluster and share the results so we can confirm how the scheduler handles this case in practice.

moko-poi:
@tzneal @engedaam @tallaxes Thank you for your feedback. I've completed testing with KWOK (Kubernetes Without Kubelet) to verify the actual behavior comparing before and after this fix.

Test Setup

Environment:
- NodePool A: zones [test-zone-a, test-zone-b] only
- NodePool B: zones [test-zone-c] only
- Test scenario:
  - Step 1: Deploy 1 existing pod to zone-c (using NodePool B)
  - Step 2: Deploy 3 new pods targeting zones a & b via nodeAffinity (using NodePool A)
  - Constraint: maxSkew=1 across topology.kubernetes.io/zone

Results

**Before Fix (main branch)**

Nodes provisioned: 2
├─ test-zone-a: 1 node (nodepool-a)
├─ test-zone-b: 0 nodes ← Problem!
└─ test-zone-c: 1 node (nodepool-b)

Pod distribution:
├─ test-zone-a: 3 new-pods ← Violates maxSkew=1!
├─ test-zone-b: 0 new-pods
└─ test-zone-c: 1 existing-pod

Karpenter logs (main branch):
{"message":"computed new nodeclaim(s) to fit pod(s)","nodeclaims":2,"pods":4}
{"message":"launched nodeclaim","zone":"test-zone-a"}
// Zone-b node was NOT created

Issue: All 3 new pods concentrated in zone-a, failing to satisfy the topology spread constraint properly.

**After Fix (this PR)**

Nodes provisioned: 3
├─ test-zone-a: 1 node (nodepool-a)
├─ test-zone-b: 1 node (nodepool-a) ← Fixed!
└─ test-zone-c: 1 node (nodepool-b)

Pod distribution:
├─ test-zone-a: 2 new-pods
├─ test-zone-b: 1 new-pod ← Properly distributed!
└─ test-zone-c: 1 existing-pod

Skew: 2-1 (satisfies maxSkew=1) ✅

Karpenter logs (this PR):
{"message":"computed new nodeclaim(s) to fit pod(s)","nodeclaims":2,"pods":3}
{"message":"launched nodeclaim","zone":"test-zone-a"}
{"message":"launched nodeclaim","zone":"test-zone-b"} ← Now creates zone-b node!

Findings

The fix correctly addresses the issue:

1. Before: Karpenter evaluated topology domains from all NodePools (including zone-c from NodePool B), causing suboptimal node placement
2. After: Karpenter only considers zones from NodePools compatible with pod requirements (zones a & b from NodePool A)
3. Result: Topology spread constraint is properly satisfied with 2-1 distribution across available zones

This aligns with the Kubernetes documentation you referenced: "The scheduler will skip the non-matching nodes from the skew calculations if the incoming Pod has spec.nodeSelector or spec.affinity.nodeAffinity defined."

The fix ensures Karpenter's provisioning logic matches the scheduler's domain filtering behavior.

moko-poi:
@engedaam @tallaxes 

Could you please review this PR when you have a chance? 

I've completed KWOK testing that demonstrates the fix correctly addresses the topology spread issue. The test results show that after the fix, pods are properly distributed (2-1 across zones a & b) instead of concentrating all 3 pods in zone-a.

This aligns with the Kubernetes scheduler behavior documented in the topology spread constraints guide regarding nodeSelector/nodeAffinity interaction.

Thank you!

jmdeal:
I think this is a duplicate of https://github.com/kubernetes-sigs/karpenter/pull/2639, right? We did have to revert that PR due to https://github.com/kubernetes-sigs/karpenter/issues/2785, but we were able to root cause the issue [here](https://github.com/kubernetes-sigs/karpenter/issues/2785#issuecomment-3755823578). I had found two issues, the main one it looks like you're PR is also susceptible to (handling of NodePool / NodeClass affinity). I need to take a closer look, but it looks like your PR does address the other issue in the original where taints and requirements were treated as independent.

@tal19987, as author of the original PR would you be able to verify if this one also meets your use case? Since this handles the requirement-taint dependency issue (which was the bigger change), I think this one is pretty close to merge. I need to take a deeper review pass still though.

tal19987:
@jmdeal @moko-poi Thanks for asking me to verify this PR.

I've reviewed this approach and it seems to solve the core issue from #2623:

Diffs:
1. Taints and requirements are now evaluated together per-DomainSource (fixes the independent evaluation bug @jmdeal identified)
2. Uses base NodePool requirements from labels + spec rather than instance type requirements
3. KWOK testing demonstrates correct 2-1 distribution across 2 zones

Question/Suggestion:
The original Issue #2623 used `nodeSelector` targeting a custom NodePool label (`workload-group=app`), not zone-based selection. Could we add a test case that explicitly covers:
- NodePool A: has label `workload-group=app` + taint + `capacity-spread=1`
- NodePool B: different labels + `capacity-spread=2-10` + no taint (and already has a few nodes up so they have the `capacity-spread` label)
- Pod: `nodeSelector: { workload-group: app }` + tolerations + TSC with Honor policies
- Result: The pod should see only one domain of `capacity-spread=1` and schedule pods without issues

This would directly validate against my original regression scenario.

tal19987:
@jmdeal @moko-poi
Any updates on this?

moko-poi:
@tal19987 Sorry for the late reply, and thanks for your thorough review and the concrete test scenario.

I've rebased the branch onto the latest main and am now working on the following:

1. Adding the test case you suggested — NodePool A with custom label
   (`workload-group=app`) + taint + single `capacity-spread` domain, and
   NodePool B with different labels + multiple `capacity-spread` domains +
   no taint. This directly validates against the regression scenario from #2623.

2. Adding a regression test for the #2785 revert cause — Specifically
   covering the case where a NodePool has zero active nodes with `minDomains`
   set, to ensure this PR does not reintroduce the issue that caused #2639
   to be reverted.

3. Investigating a fix for `NodeSelectorTerms` handling — Currently the
   code only uses the first `NodeSelectorTerm` for domain filtering. Since
   terms are OR'd, this could incorrectly exclude valid domains when multiple
   terms are specified. I'm evaluating how to handle this properly.

Will push updates soon.

moko-poi:
@tal19987 @jmdeal I've pushed all the changes mentioned in my previous comment.

For the test cases, I added two new tests in the "NodePool Domain Filtering (Issue #2227)" context. The first covers the scenario from #2623: a tainted NodePool with a custom label (workload-group=isolated-app) restricted to a single zone, alongside an untainted NodePool with different labels covering other zones. A pod targeting the isolated NodePool via nodeSelector with
tolerations and TSC using Honor policies should only see the single zone from the compatible NodePool. The second is a regression test for #2785 to ensure that NodePools with zero active nodes still register their domains correctly when minDomains is set, which was the root cause of the #2639 revert.

For the bug fix, I found that the domain filtering logic was only using the first NodeSelectorTerm when building pod requirements. Since terms are OR'd together, this could incorrectly exclude valid domains when multiple terms are specified. I fixed this by reusing TopologyNodeFilter.Requirements, which already handles all terms correctly, instead of duplicating the logic in
NewTopologyGroup.

All 342 scheduling tests pass locally.

Could an org member run /ok-to-test when you get a chance?

talayalon (commented):
Thanks for addressing this — overall the approach is solid. Taints + requirements are correctly evaluated together per-DomainSource, `nodeFilter.Requirements` properly handles all `NodeSelectorTerms`, and there's no `requirements.Clone()` so the OOM from #2639 shouldn't recur. Left a few inline comments.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -109,32 +109,41 @@ func buildDomainGroups(nodePools []*v1.NodePool, instanceTypes map[string][]*clo
 	domainGroups := map[string]TopologyDomainGroup{}
 	for npName, its := range instanceTypes {
 		np := nodePoolIndex[npName]
+		// This represents what a pod needs to match to schedule on this NodePool
+		// We exclude instance type requirements here because:
+		// 1. Instance type requirements are specific to individual instance types (capacity-type, instance-family, etc.)
+		// 2. Pod nodeSelector/nodeAffinity typically targets NodePool-level constraints (labels, zones, etc.)
+		// 3. Instance type requirements are evaluated separately during scheduling
```
tal19987: Note: `karpenter.sh/nodepool` and the NodeClass label (e.g. `karpenter.k8s.aws/ec2nodeclass`) are not part of `np.Spec.Template.Spec.Requirements` or `np.Spec.Template.ObjectMeta.Labels` - they are injected later in [`nodeclaimtemplate.go:70-72`](https://github.com/kubernetes-sigs/karpenter/blob/eca8d1b01f8b807a530f708e08a92617a0cb8ad9/pkg/controllers/provisioning/scheduling/nodeclaimtemplate.go#L70-L72).

Since `Intersects()` only checks keys present in both sets, a pod with `nodeSelector: {karpenter.sh/nodepool: X}` will pass the check for ALL NodePools - making filtering a no-op for this selector.

This is safe (no regression), but it means the fix does not cover NodePool-name-based selection. @jmdeal - is this acceptable scope, or should `karpenter.sh/nodepool` be injected into `nodePoolRequirements` here?

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -17,56 +17,120 @@ limitations under the License.
...
+	}
+
+	// Create a new domain source for this NodePool
+	source := DomainSource{
+		NodePoolRequirements: nodePoolRequirements,
+		Taints:               taints,
+	}
+
+	// NOTE: We must not optimize/override sources based solely on taints.
+	// Even if a NodePool has no taints, it may be incompatible with a pod's nodeSelector/nodeAffinity.
```
tal19987, line 82: The old taint optimization ("if any source has no taints, override all others") kept the `sources` list short. Removing it means more `DomainSource` entries per domain and more iterations in `isDomainCompatible`. This is correct for the fix, but could you confirm with a `-benchmem` run of the scheduling benchmarks that there's no significant regression?

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -1922,6 +1974,357 @@ var _ = Describe("Topology", func() {
...
+			ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+
+			// All pods should successfully schedule
+			for _, pod := range pods {
+				ExpectScheduled(ctx, env.Client, pod)
+			}
+
+			// Pods should be spread across available zones
+			ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(2, 1, 1))
+		})
```
tal19987, line 2399: This regression test verifies that zero-node NodePools register domains, which is good. However, the actual #2785 failure path involved `karpenter.sh/nodepool` selection + taints + `minDomains` + 0 nodes. Could this test be strengthened by using a tainted NodePool with a custom label (similar to the #2623 test above) combined with `minDomains` and 0 active nodes? That would get closer to the real failure path.

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
tal19987: You can check out what I did with `benchmem` here: https://github.com/kubernetes-sigs/karpenter/pull/2639#issuecomment-3614381141

moko-poi: Ran the full scheduling benchmark suite with `-benchmem` (`-count=6`) comparing `main` vs this branch.

**Environment**
```
goos: darwin
goarch: amd64
cpu: VirtualApple @ 2.50GHz
```

**Commands**
```bash
# Baseline (main branch)
git checkout main
cd pkg/controllers/provisioning/scheduling
go test -tags=test_performance -run=XXX -bench=. -count=6 -benchmem | tee /tmp/bench-main.txt

# This branch
git checkout fix/topology-spread-nodepool-zone-filtering
go test -tags=test_performance -run=XXX -bench=. -count=6 -benchmem | tee /tmp/bench-fix.txt

# Compare
benchstat main=/tmp/bench-main.txt fix=/tmp/bench-fix.txt
```

**Key metrics (geomean across all benchmarks)**
| Metric | main | fix | Change |
|---|---|---|---|
| sec/op | 618.8m | 603.6m | -2.45% |
| B/op | 195.9Mi | 192.5Mi | -1.75% |
| allocs/op | 4.805M | 4.685M | -2.48% |

<details>
<summary>Full benchstat output</summary>

```
goos: darwin
goarch: amd64
cpu: VirtualApple @ 2.50GHz
                      │     main     │                fix                 │
                      │    sec/op    │    sec/op     vs base              │
Scheduling1-12          198.7µ ±  5%   199.1µ ±  2%       ~ (p=0.589 n=6)
Scheduling50-12         47.64m ± 11%   45.31m ±  8%       ~ (p=0.132 n=6)
Scheduling100-12        87.48m ± 17%   79.59m ± 11%  -9.02% (p=0.041 n=6)
Scheduling500-12        442.9m ±  9%   420.1m ±  5%       ~ (p=0.132 n=6)
Scheduling1000-12       904.5m ± 10%   867.2m ±  6%       ~ (p=0.310 n=6)
Scheduling2000-12        1.701 ±  7%    1.756 ±  7%       ~ (p=0.310 n=6)
Scheduling5000-12        4.509 ±  9%    4.555 ±  7%       ~ (p=1.000 n=6)
Scheduling10000-12       9.904 ±  4%    9.667 ±  4%  -2.40% (p=0.041 n=6)
Scheduling20000-12       21.76 ±  4%    21.36 ± 10%       ~ (p=1.000 n=6)
RespectPreferences-12    11.98 ±  6%    11.76 ±  5%       ~ (p=0.132 n=6)
IgnorePreferences-12    775.2m ±  3%   763.3m ±  3%       ~ (p=0.485 n=6)
geomean                 618.8m         603.6m        -2.45%

                      │      main      │                 fix                 │
                      │      B/op      │     B/op       vs base              │
Scheduling1-12           46.01Ki ± 12%   46.65Ki ± 12%       ~ (p=0.818 n=6)
Scheduling50-12          10.60Mi ± 16%   11.49Mi ± 17%       ~ (p=0.937 n=6)
Scheduling100-12         19.15Mi ± 17%   19.03Mi ± 10%       ~ (p=0.485 n=6)
Scheduling500-12        101.15Mi ± 16%   93.97Mi ± 10%       ~ (p=0.485 n=6)
Scheduling1000-12        227.1Mi ± 19%   210.9Mi ±  8%       ~ (p=0.310 n=6)
Scheduling2000-12        420.8Mi ±  9%   424.6Mi ±  7%       ~ (p=0.818 n=6)
Scheduling5000-12        1.316Gi ± 10%   1.360Gi ±  9%       ~ (p=0.485 n=6)
Scheduling10000-12       3.591Gi ± 11%   3.602Gi ±  4%       ~ (p=0.699 n=6)
Scheduling20000-12       11.25Gi ±  2%   11.10Gi ±  6%       ~ (p=0.310 n=6)
RespectPreferences-12    20.57Gi ±  0%   20.57Gi ±  0%  +0.01% (p=0.026 n=6)
IgnorePreferences-12     154.1Mi ± 62%   130.4Mi ± 55%       ~ (p=0.513 n=6)
geomean                  195.9Mi         192.5Mi        -1.75%

                      │     main     │                fix                 │
                      │  allocs/op   │  allocs/op    vs base              │
Scheduling1-12           577.5 ± 10%    583.5 ± 10%       ~ (p=0.853 n=6)
Scheduling50-12         359.4k ± 26%   368.7k ± 29%       ~ (p=0.699 n=6)
Scheduling100-12        618.5k ± 30%   600.4k ± 13%       ~ (p=0.394 n=6)
Scheduling500-12        3.476M ± 25%   3.146M ± 15%       ~ (p=0.485 n=6)
Scheduling1000-12       7.956M ± 29%   7.099M ± 13%       ~ (p=0.310 n=6)
Scheduling2000-12       14.09M ± 14%   14.30M ± 11%       ~ (p=0.818 n=6)
Scheduling5000-12       39.83M ± 17%   42.20M ± 15%       ~ (p=0.394 n=6)
Scheduling10000-12      94.95M ± 22%   95.72M ±  9%       ~ (p=0.699 n=6)
Scheduling20000-12      264.4M ±  4%   256.4M ± 15%       ~ (p=0.310 n=6)
RespectPreferences-12   315.1M ±  0%   315.1M ±  0%       ~ (p=0.310 n=6)
IgnorePreferences-12    1.998M ± 49%   1.775M ± 38%       ~ (p=0.558 n=6)
geomean                 4.805M         4.685M        -2.48%
```

</details>

All differences are within noise (p > 0.05 for nearly all benchmarks). The removal of the taint optimization does not cause measurable regression in scheduling performance or memory usage.

pkg/controllers/provisioning/scheduling/topology_test.go
moko-poi: Strengthened the regression test in 763ec069. The test now mirrors the actual #2785 failure path:

- Tainted NodePool (`gha-linux=true:NoSchedule`) with custom label (`workload-group: gha-runners`) and all 3 zones
- Second untainted NodePool (`shared`) with different label and zone-3 only
- Pod with `nodeSelector: {workload-group: gha-runners}` + toleration for the taint
- `minDomains: 3` with `NodeTaintsPolicy: Honor` and `NodeAffinityPolicy: Honor`
- Zero active nodes initially for both NodePools

This follows the #2623 test pattern (tainted NodePool + custom label + tolerations) while adding the zero-node + minDomains combination that caused the actual #2785 failure.

Note: We use a custom label (`workload-group`) rather than `karpenter.sh/nodepool` because, as noted in the other comment, that label is injected later in `nodeclaimtemplate.go:70-72` and isn't available during topology domain filtering. The custom label exercises the same filtering code path effectively.

pkg/controllers/provisioning/scheduling/topology.go
jmdeal: I don't think this is actually true. Users will configure pod-level scheduling constraints that aren't specified on the NodePool. There's two main schools of though for Karpenter NodePool configuration: specific NodePools per application (or group of applications) or broad NodePools where instance-type selection is driven by node affinity.

Consider a pod which requires a specific instance family, but that family is only available in a subset of zones for the NodePool. If we only consider NodePool requirements, we'll assume that the pod can spread across all available zones for the NodePool, rather than the intersection of available zones in the NodePool and the zones for that instance type. We would then fail to scale-up because there wouldn't be any available offerings which could satisfy the topology spread constraint. Granted, a user could get around this by also specifying a node affinity term on zone, but that's not ideal.

That being said, I'm concerned about the memory / performance overhead of storing requirements for each compatible instance type * the number of NodePools. A naive solution could result in thousands of domain sources per domain group. If we wanted to support this, I think we'd need to figure out a better optimization for storing requirements which heavily overlap - maybe a trie-like data structure for requirements? I think we can scope this out of this PR since it's still a net improvement over Karpenter's current behavior. Let's document the current limitations of this approach though.

jmdeal: Yeah, I agree that we should inject the NodePool and NodeClass labels here.

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -17,56 +17,120 @@ limitations under the License.
...
 
-// TopologyDomainGroup tracks the domains for a single topology. Additionally, it tracks the taints associated with
-// each of these domains. This enables us to determine which domains should be considered by a pod if its
-// NodeTaintPolicy is honor.
-type TopologyDomainGroup map[string][][]v1.Taint
+// DomainSource tracks the requirements and taints for a specific domain provided by a NodePool.
+// This allows us to determine if a pod can use this domain based on its nodeSelector and nodeAffinity.
+type DomainSource struct {
+	// NodePoolRequirements contains the combined requirements from the NodePool's labels and requirements.
+	// This includes both user-defined labels and instance type requirements (e.g., karpenter.k8s.aws/instance-size).
```
jmdeal: nit: this suggests we're handling each instance-type as a standalone domain source (which we would need to do to be strictly correct)
```suggestion
	// This includes both user-defined and well-known labels.
```

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -17,56 +17,120 @@ limitations under the License.
...
+		requirementsMatch := len(podRequirementSets) == 0
+		for _, podReqs := range podRequirementSets {
+			if err := source.NodePoolRequirements.Intersects(podReqs); err == nil {
+				requirementsMatch = true
 				break
 			}
 		}
+		if !requirementsMatch {
+			continue
+		}
```
jmdeal: We should only perform this check if the affinity policy is set to honor, right?

pkg/controllers/provisioning/scheduling/topology.go
moko-poi: Thanks for the detailed explanation. That makes sense — a pod requiring a specific instance family available only in a subset of zones would still see the full NodePool zone set with this approach.

Addressed in fbf3a1db. I added a comment documenting this limitation in `buildDomainGroups()` and injected `karpenter.sh/nodepool` and NodeClass labels into `nodePoolRequirements`. The limitation comment notes that per-instance-type requirements are not considered for compatibility checking, and that full per-instance-type filtering happens later during scheduling.

moko-poi: Done in fbf3a1db. Injected both `v1.NodePoolLabelKey` and `v1.NodeClassLabelKey()` into `nodePoolRequirements` in `buildDomainGroups()`. Added a test that verifies pods selecting a NodePool via `karpenter.sh/nodepool` correctly filter topology domains to that NodePool's zones only.

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
moko-poi: Fixed in fbf3a1db.

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
moko-poi: Good catch. Fixed in fbf3a1db. I added an `affinityPolicy` parameter to `ForEachDomain` and `isDomainCompatible`, and the requirements check is now only performed when `affinityPolicy == NodeInclusionPolicyHonor`, consistent with `TopologyNodeFilter.Matches()`.

For topology spread, the default `AffinityPolicy` is `Honor` so the requirements check runs as before. When `NodeAffinityPolicy` is set to `Ignore`, the check is skipped and all domains are considered. For affinity/anti-affinity, the `nodeFilter` is zero-value so `AffinityPolicy` is `""` which doesn't equal `"Honor"`, meaning the check is also skipped — this is correct since affinity/anti-affinity counts all nodes regardless.

I also added a test verifying that `NodeAffinityPolicy: Ignore` causes all domains to be considered regardless of nodeSelector.

pkg/controllers/provisioning/scheduling/topology.go
```diff
@@ -109,32 +109,54 @@ func buildDomainGroups(nodePools []*v1.NodePool, instanceTypes map[string][]*clo
...
+		// This represents what a pod needs to match to schedule on this NodePool.
+		// We use NodePool-level requirements only, not per-instance-type requirements, because:
+		// 1. Instance type requirements are specific to individual instance types (capacity-type, instance-family, etc.)
+		// 2. Pod nodeSelector/nodeAffinity typically targets NodePool-level constraints (labels, zones, etc.)
+		// 3. Instance type requirements are evaluated separately during scheduling
+		//
+		// Limitation: Per-instance-type requirements (e.g., zones available for specific instance families)
+		// are not considered for compatibility checking. A pod may be matched to a NodePool even if no
+		// individual instance type satisfies all of the pod's constraints. Full per-instance-type filtering
+		// happens later during scheduling.
```
jmdeal: Updated the wording to better reflect why we need to track NodePool requirements and the limitations
```suggestion
		// We need to associate available domains with the requirements of the NodePool that provides them. Consider a
		// scenario where one NodePool provides a single zone and another provides three. If a pod has a scheduling
		// constraint which restricts it to the single zone NodePool, the only zone it can spread across is that zone. We
		// should not consider all zones eligible domains (assuming NodeAffinityPolicy is set to honor).
		//
		// Note that this approach still has limitations. Eligible domains are only tracked on a NodePool level rather than
		// an instance type level. If an instance type has it's own domain restrictions and the application is only
		// compatible with that instance type, we should only consider those zones. This is an extremely large combinatoric
		// space, so this is currently not supported. Applications can work around this limitation by specifying the eligible 
		// domains for the instance type in a nodeSelector.

```

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
```diff
@@ -17,56 +17,124 @@ limitations under the License.
...
+	for _, source := range sources {
+		// Only check requirements compatibility when the affinity policy is Honor,
+		// consistent with TopologyNodeFilter.Matches() behavior (topologynodefilter.go).
+		// When the policy is Ignore (or unset for affinity/anti-affinity), all domains
+		// are considered compatible from a requirements perspective.
+		requirementsMatch := true
+		if affinityPolicy == v1.NodeInclusionPolicyHonor {
+			requirementsMatch = len(podRequirementSets) == 0
+			for _, podReqs := range podRequirementSets {
+				if err := source.NodePoolRequirements.Intersects(podReqs); err == nil {
```
jmdeal: I think `Intersects` is actually subtly wrong. If a pod has a requirement for a non-well known label that isn't defined on a NodePool, we shouldn't consider that NodePools domains as eligible for the pod. `Intersects` only checks that the common requirements are compatible (i.e. shared keys between `NodePoolRequirements` and `podReqs`). `Compatible` with `AllowUndefinedWellKnownLabels` should be the right check.

pkg/controllers/provisioning/scheduling/topology.go
moko-poi: Thanks for the suggestion, I've updated the comment wording in 12089dca. The new text better explains the motivation with the concrete scenario and clarifies the limitation around instance-type-level domain tracking.

pkg/controllers/provisioning/scheduling/topologydomaingroup.go
moko-poi: Good catch, thanks. You're right that Intersects silently skips keys only present in one side, so a custom label like workload-group=app on the pod would never be checked against a NodePool that doesn't define it. Changed to Compatible with AllowUndefinedWellKnownLabels in 33a4bce2.

PR #2674, opened 2025-12-01
feat: Add Gte and Lte operators for requirements

ellistarn:
Adds `Gte` (>=) and `Lte` (<=) operators for NodePool / NodeOverlay requirements.

## Review Focus

- `pkg/apis/v1/` - New operator constants and validation
- `pkg/scheduling/requirement.go` - Core intersection logic

## Design Notes

**Canonicalization:** Gt/Lt are normalized to Gte/Lte internally (Gt N → Gte N+1, Lt N → Lte N-1). Jason and I went back and forth on this, but decided it was cleaner to canonicalize. The tradeoff: users who specify `Gt 5` in their NodePool will see `Gte 6` on the resulting NodeClaim and in logs. The values are mathematically equivalent, but this may be surprising. We accepted this for simpler internal logic (one bound type instead of two). We also expect users will begin using Gte/Lte much more commonly anyways.

**API struct change:** Apologies for the mechanical churn in the API struct changes. This was necessary to allow controller-gen to add new fields to the enum validation, which had the side effect of touching many files.

**CRD enum duplication:** The generated CRD has duplicate enum values (e.g., `In` appears twice). This is a controller-gen bug that unions the kubebuilder annotation with the underlying type. Harmless, and fixing requires a larger refactor.

---

## Bugfix

Fixed a pre-existing bug where `NodeSelectorRequirements()` would only return one bound when both upper and lower bounds existed on the same key. For example, if a NodePool specified both `Gte 8` and `Lte 8` on `instance-cpu`, the resulting NodeClaim would only have the `Gte 8` requirement, making it broader than intended. Added test coverage for this case.

---

## Testing

1. Unit
2. Manual
```
❯ k get nodeclaims -oyaml
apiVersion: v1
items:
- apiVersion: karpenter.sh/v1
  kind: NodeClaim
  metadata:
    annotations:
      compatibility.karpenter.k8s.aws/cluster-name-tagged: "true"
      karpenter.k8s.aws/ec2nodeclass-hash: "1655830214716096975"
      karpenter.k8s.aws/ec2nodeclass-hash-version: v4
      karpenter.k8s.aws/instance-profile-name: etarn-karpenter-dev_8344205451026389402
      karpenter.k8s.aws/tagged: "true"
      karpenter.sh/nodeclaim-min-values-relaxed: "false"
      karpenter.sh/nodepool-hash: "6821555240594823858"
      karpenter.sh/nodepool-hash-version: v3
    creationTimestamp: "2025-12-09T00:48:28Z"
    finalizers:
    - karpenter.sh/termination
    generateName: default-4cpu-
    generation: 1
    labels:
      karpenter.k8s.aws/ec2nodeclass: default
      karpenter.k8s.aws/instance-capability-flex: "false"
      karpenter.k8s.aws/instance-category: c
      karpenter.k8s.aws/instance-cpu: "8"
      karpenter.k8s.aws/instance-cpu-manufacturer: amd
      karpenter.k8s.aws/instance-cpu-sustained-clock-speed-mhz: "3600"
      karpenter.k8s.aws/instance-ebs-bandwidth: "10000"
      karpenter.k8s.aws/instance-encryption-in-transit-supported: "true"
      karpenter.k8s.aws/instance-family: c6a
      karpenter.k8s.aws/instance-generation: "6"
      karpenter.k8s.aws/instance-hypervisor: nitro
      karpenter.k8s.aws/instance-memory: "16384"
      karpenter.k8s.aws/instance-network-bandwidth: "3125"
      karpenter.k8s.aws/instance-size: 2xlarge
      karpenter.k8s.aws/instance-tenancy: default
      karpenter.sh/capacity-type: on-demand
      karpenter.sh/nodepool: default-4cpu
      kubernetes.io/arch: amd64
      kubernetes.io/os: linux
      node.kubernetes.io/instance-type: c6a.2xlarge
      topology.k8s.aws/zone-id: usw2-az4
      topology.kubernetes.io/region: us-west-2
      topology.kubernetes.io/zone: us-west-2d
    name: default-4cpu-wxz8x
    ownerReferences:
    - apiVersion: karpenter.sh/v1
      blockOwnerDeletion: true
      kind: NodePool
      name: default-4cpu
      uid: d213818a-b328-4e89-a981-684582a18ef8
    resourceVersion: "1776680"
    uid: 7c7875fb-9951-4bd8-be31-349617705a19
  spec:
    expireAfter: 720h
    nodeClassRef:
      group: karpenter.k8s.aws
      kind: EC2NodeClass
      name: default
    requirements:
      operator: Gte
      values:
      - "8"
      operator: Lte
      values:
      - "8"
    - key: karpenter.k8s.aws/ec2nodeclass
      operator: In
      values:
      - default
    - key: kubernetes.io/os
      operator: In
      values:
      - linux
    - key: karpenter.sh/nodepool
      operator: In
      values:
      - default-4cpu
      operator: In
      values:
      - on-demand
    - key: kubernetes.io/arch
      operator: In
      values:
      - amd64
      operator: Gte
      values:
      - "3"
      operator: In
      values:
      - c
      - m
      - r
      operator: In
      values:
      - c3.2xlarge
      - c4.2xlarge
      - c5.2xlarge
      - c5a.2xlarge
      - c5ad.2xlarge
      - c5d.2xlarge
      - c5n.2xlarge
      - c6a.2xlarge
      - c6i.2xlarge
      - c6id.2xlarge
      - c6in.2xlarge
      - c7a.2xlarge
      - c7i-flex.2xlarge
      - c7i.2xlarge
      - c8a.2xlarge
      - c8i-flex.2xlarge
      - c8i.2xlarge
      - m3.2xlarge
      - m4.2xlarge
      - m5.2xlarge
      - m5a.2xlarge
      - m5ad.2xlarge
      - m5d.2xlarge
      - m5dn.2xlarge
      - m5n.2xlarge
      - m5zn.2xlarge
      - m6a.2xlarge
      - m6i.2xlarge
      - m6id.2xlarge
      - m6idn.2xlarge
      - m6in.2xlarge
      - m7a.2xlarge
      - m7i-flex.2xlarge
      - m7i.2xlarge
      - m8a.2xlarge
      - m8i-flex.2xlarge
      - m8i.2xlarge
      - r3.2xlarge
      - r4.2xlarge
      - r5.2xlarge
      - r5a.2xlarge
      - r5ad.2xlarge
      - r5b.2xlarge
      - r5d.2xlarge
      - r5dn.2xlarge
      - r5n.2xlarge
      - r6a.2xlarge
      - r6i.2xlarge
      - r6id.2xlarge
      - r6idn.2xlarge
      - r6in.2xlarge
      - r7a.2xlarge
      - r7i.2xlarge
      - r7iz.2xlarge
      - r8a.2xlarge
      - r8i-flex.2xlarge
      - r8i.2xlarge
    resources:
      requests:
        cpu: 6150m
        pods: "7"
  status:
    allocatable:
      cpu: 7910m
      ephemeral-storage: 17Gi
      memory: 14974928Ki
      pods: "58"
      vpc.amazonaws.com/pod-eni: "38"
    capacity:
      cpu: "8"
      ephemeral-storage: 20Gi
      memory: 15991760Ki
      pods: "58"
      vpc.amazonaws.com/pod-eni: "38"
    conditions:
    - lastTransitionTime: "2025-12-09T00:48:31Z"
      message: ""
      observedGeneration: 1
      reason: Launched
      status: "True"
      type: Launched
    - lastTransitionTime: "2025-12-09T00:48:47Z"
      message: ""
      observedGeneration: 1
      reason: Registered
      status: "True"
      type: Registered
    - lastTransitionTime: "2025-12-09T00:48:57Z"
      message: ""
      observedGeneration: 1
      reason: Initialized
      status: "True"
      type: Initialized
    - lastTransitionTime: "2025-12-09T00:48:57Z"
      message: ""
      observedGeneration: 1
      reason: Ready
      status: "True"
      type: Ready
    imageID: ami-0719a54ea5c286ca8
    lastPodEventTime: "2025-12-09T00:48:56Z"
    nodeName: ip-192-168-77-165.us-west-2.compute.internal
    providerID: aws:///us-west-2d/i-080a0ed2bb109cd11
kind: List
metadata:
  resourceVersion: ""
```

### DIFF
--- a/pkg/apis/v1/nodeclaim.go
+++ b/pkg/apis/v1/nodeclaim.go
@@ -91,7 +91,7 @@ const (
 // and minValues that represent the requirement to have at least that many values.
 type NodeSelectorRequirementWithMinValues struct {
 	//nolint:kubeapilinter
-	// Key is the label key that the selector applies to.
+	// The label key that the selector applies to.
 	// +required
 	Key string `json:"key"`
 
@@ -99,13 +99,13 @@ type NodeSelectorRequirementWithMinValues struct {
 	// enum values into the ones already defined on v1.NodeSelectorOperator
 
 	//nolint:kubeapilinter
-	// Operator represents a key's relationship to a set of values.
-	// Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+	// Represents a key's relationship to a set of values.
+	// Valid operators are In, NotIn, Exists, DoesNotExist. Gt, Lt, Gte, and Lte.
 	// +kubebuilder:validation:Enum:=Gte;Lte
 	// +required
 	Operator v1.NodeSelectorOperator `json:"operator,omitempty"`
 	//nolint:kubeapilinter
-	// Values is an array of string values. If the operator is In or NotIn,
+	// An array of string values. If the operator is In or NotIn,
 	// the values array must be non-empty. If the operator is Exists or DoesNotExist,
 	// the values array must be empty. If the operator is Gt, Lt, Gte, or Lte, the values
 	// array must have a single element, which will be interpreted as an integer.
--- a/pkg/apis/v1alpha1/nodeoverlay.go
+++ b/pkg/apis/v1alpha1/nodeoverlay.go
@@ -24,11 +24,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// NodeSelectorRequirement is a node selector requirement that extends the standard
-// Kubernetes operators with Gte and Lte for NodeOverlay use cases.
+// NodeSelectorRequirement extends corev1.NodeSelectorRequirement with Gte and Lte operators.
 type NodeSelectorRequirement struct {
 	//nolint:kubeapilinter
-	// Key is the label key that the selector applies to.
+	// The label key that the selector applies to.
 	// +required
 	Key string `json:"key"`
 
@@ -36,13 +35,13 @@ type NodeSelectorRequirement struct {
 	// enum values into the ones already defined on v1.NodeSelectorOperator
 
 	//nolint:kubeapilinter
-	// Operator represents a key's relationship to a set of values.
-	// Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+	// Represents a key's relationship to a set of values.
+	// Valid operators are In, NotIn, Exists, DoesNotExist. Gt, Lt, Gte, and Lte.
 	// +kubebuilder:validation:Enum:=Gte;Lte
 	// +required
 	Operator corev1.NodeSelectorOperator `json:"operator,omitempty"`
 	//nolint:kubeapilinter
-	// Values is an array of string values. If the operator is In or NotIn,
+	// An array of string values. If the operator is In or NotIn,
 	// the values array must be non-empty. If the operator is Exists or DoesNotExist,
 	// the values array must be empty. If the operator is Gt, Lt, Gte, or Lte, the values
 	// array must have a single element, which will be interpreted as an integer.
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -124,7 +124,7 @@ spec:
                       and minValues that represent the requirement to have at least that many values.
                     properties:
                       key:
-                        description: Key is the label key that the selector applies to.
+                        description: The label key that the selector applies to.
                         type: string
                         maxLength: 316
                         pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(\/))?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$
@@ -148,8 +148,8 @@ spec:
                         type: integer
                       operator:
                         description: |-
-                          Operator represents a key's relationship to a set of values.
-                          Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+                          Represents a key's relationship to a set of values.
+                          Valid operators are In, NotIn, Exists, DoesNotExist. Gt, Lt, Gte, and Lte.
                         enum:
                           - Gte
                           - Lte
@@ -162,7 +162,7 @@ spec:
                         type: string
                       values:
                         description: |-
-                          Values is an array of string values. If the operator is In or NotIn,
+                          An array of string values. If the operator is In or NotIn,
                           the values array must be non-empty. If the operator is Exists or DoesNotExist,
                           the values array must be empty. If the operator is Gt, Lt, Gte, or Lte, the values
                           array must have a single element, which will be interpreted as an integer.
--- a/kwok/charts/crds/karpenter.sh_nodeoverlays.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeoverlays.yaml
@@ -85,12 +85,10 @@ spec:
                     - Well-known labels (e.g., node.kubernetes.io/instance-type, karpenter.sh/nodepool)
                     - Custom labels from NodePool's spec.template.labels
                   items:
-                    description: |-
-                      NodeSelectorRequirement is a node selector requirement that extends the standard
-                      Kubernetes operators with Gte and Lte for NodeOverlay use cases.
+                    description: NodeSelectorRequirement extends corev1.NodeSelectorRequirement with Gte and Lte operators.
                     properties:
                       key:
-                        description: Key is the label key that the selector applies to.
+                        description: The label key that the selector applies to.
                         type: string
                         maxLength: 316
                         pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(\/))?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$
@@ -105,8 +103,8 @@ spec:
                             rule: self != "kubernetes.io/hostname"
                       operator:
                         description: |-
-                          Operator represents a key's relationship to a set of values.
-                          Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+                          Represents a key's relationship to a set of values.
+                          Valid operators are In, NotIn, Exists, DoesNotExist. Gt, Lt, Gte, and Lte.
                         enum:
                           - Gte
                           - Lte
@@ -119,7 +117,7 @@ spec:
                         type: string
                       values:
                         description: |-
-                          Values is an array of string values. If the operator is In or NotIn,
+                          An array of string values. If the operator is In or NotIn,
                           the values array must be non-empty. If the operator is Exists or DoesNotExist,
                           the values array must be empty. If the operator is Gt, Lt, Gte, or Lte, the values
                           array must have a single element, which will be interpreted as an integer.
--- a/kwok/charts/crds/karpenter.sh_nodepools.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodepools.yaml
@@ -281,7 +281,7 @@ spec:
                               and minValues that represent the requirement to have at least that many values.
                             properties:
                               key:
-                                description: Key is the label key that the selector applies to.
+                                description: The label key that the selector applies to.
                                 type: string
                                 maxLength: 316
                                 pattern: ^([a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*(\/))?([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$
@@ -307,8 +307,8 @@ spec:
                                 type: integer
                               operator:
                                 description: |-
-                                  Operator represents a key's relationship to a set of values.
-                                  Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+                                  Represents a key's relationship to a set of values.
+                                  Valid operators are In, NotIn, Exists, DoesNotExist. Gt, Lt, Gte, and Lte.
                                 enum:
                                   - Gte
                                   - Lte
@@ -321,7 +321,7 @@ spec:
                                 type: string
                               values:
                                 description: |-
-                                  Values is an array of string values. If the operator is In or NotIn,
+                                  An array of string values. If the operator is In or NotIn,
                                   the values array must be non-empty. If the operator is Exists or DoesNotExist,
                                   the values array must be empty. If the operator is Gt, Lt, Gte, or Lte, the values
                                   array must have a single element, which will be interpreted as an integer.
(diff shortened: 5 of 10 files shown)


pkg/apis/v1/nodeclaim.go
```diff
@@ -82,10 +82,37 @@ type NodeClaimSpec struct {
...
+	Key string `json:"key"`
+
+	// NOTE below: The code generator works strangely, and will union these
+	// enum values into the ones already defined on v1.NodeSelectorOperator
+
+	// operator represents a key's relationship to a set of values.
+	// Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+	// +kubebuilder:validation:Enum:=Gte;Lte
+	// +required
+	Operator v1.NodeSelectorOperator `json:"operator,omitempty"`
```
ryan-mist, line 106: nit: should we remove omitempty since its a required field?

pkg/apis/v1/nodeclaim.go
```diff
@@ -82,10 +82,37 @@ type NodeClaimSpec struct {
...
+	// values is an array of string values. If the operator is In or NotIn,
+	// the values array must be non-empty. If the operator is Exists or DoesNotExist,
+	// the values array must be empty. If the operator is Gt, Lt, Gte, or Lte, the values
+	// array must have a single element, which will be interpreted as an integer.
+	// This array is replaced during a strategic merge patch.
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:items:MaxLength:=63
```
ryan-mist: OOC: Why are we adding validation to .Values? it looks like the upstream `corev1.NodeSelectorRequirement` struct does not have validation on max length or the pattern

pkg/apis/v1/nodeclaim.go
ellistarn: Nice catch!

ellistarn:
> Any particular reason to normalize Gt/Lt to Gte/Lte rather than the other way around to keep existing behavior in logs/NodeClaim? The one thing I can think of as why not to do it would be because >= 0 would show up as > -1, but wondering if there was anything else

Great question! This PR asserts that GTE/LTE is easier to reason about than GT/LT, which is why we're building these operators in the first place. It would be odd to assert this and then choose the alternative as the canonical form. The -1 is a nice callout I didn't consider, but definitely is part of the "harder to reason about" rationale.

pkg/apis/v1/nodeclaim.go
ellistarn: Nice catch! Kiro snuck this in.

designs/gte-lte-operators.md
```diff
@@ -0,0 +1,55 @@
...
+spec:
+  template:
+    spec:
+      requirements:
+        - key: karpenter.sh/capacity
+          operator: Gte  # New operator
+          values: ["2"]
+        - key: node.kubernetes.io/instance-memory
+          operator: Lte  # New operator
+          values: ["64"]
```
jmdeal: nit: use some real labels
```suggestion
        - key: karpenter.k8s.aws/instance-cpu
          operator: Gte  # New operator
          values: ["2"]
        - key: karpenter.k8s.io/instance-memory
          operator: Lte  # New operator
          values: ["64"]
```

pkg/apis/v1/nodeclaim.go
```diff
@@ -82,10 +82,35 @@ type NodeClaimSpec struct {
...
+	NodeSelectorOpGte v1.NodeSelectorOperator = "Gte"
+	NodeSelectorOpLte v1.NodeSelectorOperator = "Lte"
+)
+
 // A node selector requirement with min values is a selector that contains values, a key, an operator that relates the key and values
 // and minValues that represent the requirement to have at least that many values.
 type NodeSelectorRequirementWithMinValues struct {
-	v1.NodeSelectorRequirement `json:",inline"`
+	//nolint:kubeapilinter
+	// key is the label key that the selector applies to.
```
jmdeal: nit: elsewhere we capitalize the name of the field in the comment, I'd do it here as well for consistency
```suggestion
	// Key is the label key that the selector applies to.
```

pkg/apis/v1/nodeclaim.go
```diff
@@ -82,10 +82,35 @@ type NodeClaimSpec struct {
...
+	// +required
+	Key string `json:"key"`
+
+	// NOTE below: The code generator works strangely, and will union these
+	// enum values into the ones already defined on v1.NodeSelectorOperator
+
+	// operator represents a key's relationship to a set of values.
```
jmdeal:
```suggestion
	// Operator represents a key's relationship to a set of values.
```

pkg/apis/v1/nodeclaim.go
```diff
@@ -82,10 +82,35 @@ type NodeClaimSpec struct {
...
+	// Valid operators are In, NotIn, Exists, DoesNotExist, Gt, Lt, Gte, and Lte.
+	// +kubebuilder:validation:Enum:=Gte;Lte
+	// +required
+	Operator v1.NodeSelectorOperator `json:"operator,omitempty"`
+	// values is an array of string values. If the operator is In or NotIn,
```
jmdeal:
```suggestion
	// Values is an array of string values. If the operator is In or NotIn,
```

pkg/apis/v1alpha1/nodeoverlay.go
```diff
@@ -32,7 +32,7 @@ type NodeOverlaySpec struct {
 	// - Custom labels from NodePool's spec.template.labels
 	// +kubebuilder:validation:XValidation:message="requirements with operator 'NotIn' must have a value defined",rule="self.all(x, x.operator == 'NotIn' ? x.values.size() != 0 : true)"
 	// +kubebuilder:validation:XValidation:message="requirements with operator 'In' must have a value defined",rule="self.all(x, x.operator == 'In' ? x.values.size() != 0 : true)"
-	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt' or 'Lt' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
+	// +kubebuilder:validation:XValidation:message="requirements operator 'Gt', 'Lt', 'Gte', or 'Lte' must have a single positive integer value",rule="self.all(x, (x.operator == 'Gt' || x.operator == 'Lt' || x.operator == 'Gte' || x.operator == 'Lte') ? (x.values.size() == 1 && int(x.values[0]) >= 0) : true)"
 	// +kubebuilder:validation:MaxItems:=100
 	// +required
 	Requirements []v1.NodeSelectorRequirement `json:"requirements,omitempty"`
```
jmdeal: Should we use a new type here rather than continuing to use the upstream type? It's not strictly necessary, but it strikes me as strange to allow values which wouldn't normally be valid for the upstream type. Encoding it in a different type makes it clear that it accepts all valid Karpenter operators, rather than all valid upstream operators.

jmdeal: Minimally, even if we didn't use a different type we'd need to modify the inherited CEL validation - right now it's inheriting the valid operators from the upstream type so you wouldn't be able to specify Gte or Lte. We should probably add tests to all of the CEL validation suites to exercise that we can specify all valid operators.

pkg/scheduling/requirement.go
```diff
@@ -77,11 +78,21 @@ func NewRequirementWithFlexibility(key string, operator corev1.NodeSelectorOpera
 	}
 	if operator == corev1.NodeSelectorOpGt {
 		value, _ := strconv.Atoi(values[0]) // prevalidated
-		r.greaterThan = &value
+		value++                             // canonicalize GT N to GTE N+1
```
jmdeal: We should probably use CEL to bound the LT and GT values such that this doesn't result in an overflow (or an underflow for other operations). FWIW though I can't really imagine a use case where this would be an issue considering the requirement would be nonsense (e.g. GT MAX_INT or LT MIN_INT).

pkg/scheduling/requirement.go
```diff
@@ -90,63 +101,63 @@ func NewRequirement(key string, operator corev1.NodeSelectorOperator, values ...
 	return NewRequirementWithFlexibility(key, operator, nil, values...)
 }
 
+// BoundedNodeSelectorRequirements handles the case where both gte and lte exist.
+// Unlike other operators which intersect into a single values set, bounds are stored
+// separately and must be serialized as two distinct NodeSelectorRequirements.
+// This method is separate from NodeSelectorRequirement() to avoid slice allocations
+// in the common case where only one bound exists.
+func (r *Requirement) BoundedNodeSelectorRequirements() []v1.NodeSelectorRequirementWithMinValues {
```
jmdeal, line 113: I'm surprised to see we don't actually handle this case correctly today. I guess it never surfaced as a bug because we inject the concrete set of instance types that we're compatible with when we launch, so we're not relying on the serialized requirements to get the correct set of instance types.

designs/gte-lte-operators.md
ellistarn: Good catch, will update to use real labels.

pkg/apis/v1/nodeclaim.go
ellistarn: Fixed

pkg/apis/v1/nodeclaim.go
ellistarn: Fixed

pkg/apis/v1/nodeclaim.go
ellistarn: Fixed

pkg/apis/v1alpha1/nodeoverlay.go
ellistarn: I explored using a new type but it led to massive changes and imports across the test base. I'll update the CEL validation to allow Gte/Lte and add test coverage for all valid operators.

pkg/scheduling/requirement.go
ellistarn: Good callout. I'll add CEL validation to bound the GT and LT values to prevent overflow/underflow during canonicalization. This keeps the runtime code clean and catches invalid values at admission time.

ellistarn: Updated to handle this in code instead of CEL validation - Gt MaxInt now returns a DoesNotExist requirement (matches nothing). Lt doesn't need special handling since the math works correctly (Lt 0 → Lte -1, which won't match any non-negative values).

pkg/apis/v1alpha1/nodeoverlay.go
ellistarn: Misunderstood this to be a comment re: overflow and our validation. I get that the issue is specific to overlay. Fixing.

pkg/apis/v1alpha1/nodeoverlay_validation_test.go
```diff
@@ -211,7 +213,23 @@ var _ = Describe("CEL/Validation", func() {
 				{Key: corev1.LabelTopologyZone, Operator: corev1.NodeSelectorOpLt, Values: []string{"a"}},
 				{Key: corev1.LabelTopologyZone, Operator: corev1.NodeSelectorOpLt, Values: []string{"-1"}},
 			} {
-				nodeOverlay.Spec.Requirements = []corev1.NodeSelectorRequirement{requirement}
+				nodeOverlay.Spec.Requirements = []NodeSelectorRequirement{requirement}
+				Expect(env.Client.Create(ctx, nodeOverlay)).ToNot(Succeed())
+				Expect(nodeOverlay.RuntimeValidate(ctx)).ToNot(Succeed())
+			}
+		})
+		It("should fail with invalid GTE or LTE values", func() {
```
jmdeal, line 221: Could we add this same validation test for NodePool and NodeClaim?

ellistarn: Done

PR #2675, opened 2025-12-01
fix: Improve Karpenter memory when using Node Overlay with copy-on-write

### LINKED ISSUE #2655: Karpenter node overlay causing increased in memory usage
### Description

**Observed Behavior**:
I have many node pools but all node pools have finite selection of instance types (7th or 6th gen, only certain family etc).

Recently, I started testing node overlay for spot pricing (we have some discounts I wish to reflect) and for the node overlay I had a broad requirement matching all spot capacity type and any instance types (no restrictions) such that any spot instance type considered by Karpenter has the appropriate pricing to compare to our on-demand pricing + discounts.

We saw a significant increase in memory (20-30%) which translated to ~500MB or so.

pprof before NodeOverlay=true:
(Sorry I don't think git likes svg but here is a png)
Image

pprof after NodeOverlay=true with Spot node overlay:
Image

You can see in instance type, the cumulated memory of all subfunctions is around  204MB before Node overlay and 1148 MB after.

I think the huge jump in memory is because the instance types we store in offering cache. Specifically, because of my overlay's broad range of instance type is causing us to now store all instance types in the offering cache vs a very small set previously.

**Expected Behavior**:
Since Node overlay should only be applying adjustments or overlays on instance types that Node Pools can launch, in an instance type constrained cluster Node Overlay should only augment the existing instance type offering cache not extend it. 

I haven't dived into the code yet. This is still speculation but my earlier investigation into Node Overlay I believe creates a separate instance type cache in the core library rather than the one in the cloud provider(?). I am wondering if this is the reason why 

- Create 1 node pool with very constrained instance types.
- Measure memory
- Create a very permissive Node Overlay
- Measure Memory

**Versions**:

- Kubernetes Version (`kubectl version`): 1.8.2 AWS Karpenter provider

GnatorX:
**Fixes #2655**

## Problem

Full deep copying of instance types caused excessive memory usage in deployments with many instance types and node pools (~1GB+ for 200 types × 5 pools).

## Solution

Replace full deep copy with selective copying:
- **Share** immutable fields (Requirements, Overhead)
- **Copy** only modified fields (Capacity, affected Offerings)
- **New** `applyPriceOverlays()` for granular offering copying

### Before
```go
overriddenInstanceType := it.DeepCopy()  // Always copies everything
```

### After
```go
overriddenInstanceType := &cloudprovider.InstanceType{
    Name:         it.Name,
    Requirements: it.Requirements, // Shared
    Overhead:     it.Overhead,     // Shared
}
// Only copy fields that need modification
```

## Results
Before: red line, After: Orange
Screenshot 2025-12-02 at 2 13 51 PM

Before:
```
      === BEFORE (Deep Copy) - Memory Tests ===

      TestMemoryUsage_OverlayScenarios:
        no_overlays:             Total Allocs: 71,514     | Bytes: 12.75 MB    | Heap: 3.97 MB
        price_overlays_only:     Total Allocs: 2,184,576  | Bytes: 164.56 MB   | Heap: 3.99 MB
        capacity_overlays_only:  Total Allocs: 1,140,409  | Bytes: 297.21 MB   | Heap: 3.99 MB
        mixed_overlays:          Total Allocs: 11,137,342 | Bytes: 1,108.82 MB | Heap: 2.76 MB

      TestMemoryUsage_ScaleWithNodePools:
        nodepools_1:   Total Allocs: 2,213,176   | Bytes: 1,271.41 MB  | Heap: 4.00 MB
        nodepools_5:   Total Allocs: 11,137,216  | Bytes: 2,083.00 MB  | Heap: 4.00 MB
        nodepools_10:  Total Allocs: 22,274,324  | Bytes: 3,716.96 MB  | Heap: 4.00 MB
        nodepools_20:  Total Allocs: 44,547,761  | Bytes: 7,028.17 MB  | Heap: 4.01 MB
        nodepools_50:  Total Allocs: 111,365,788 | Bytes: 15,697.76 MB | Heap: 4.01 MB

      TestMemoryUsage_ScaleWithInstanceTypes:
        instances_50:  Total Allocs: 3,893,618  | Bytes: 15,981.94 MB | Heap: 2.77 MB
        instances_100: Total Allocs: 7,788,094  | Bytes: 16,550.38 MB | Heap: 2.77 MB
        instances_200: Total Allocs: 11,137,399 | Bytes: 17,363.31 MB | Heap: 2.78 MB
        instances_500: Total Allocs: 11,137,410 | Bytes: 18,176.25 MB | Heap: 2.78 MB

      === BEFORE (Deep Copy) - Benchmarks ===

      BenchmarkStoreApply/selective-copy-apply    568,892 ns/op    943,506 B/op    12,475 allocs/op
      BenchmarkStoreApplyNoOverlay                  5,525 ns/op          0 B/op         0 allocs/op
      BenchmarkStoreApplyPriceOnly              1,297,943 ns/op  2,191,611 B/op    30,403 allocs/op
      BenchmarkStoreApplyCapacityOnly             818,722 ns/op  1,912,016 B/op    15,800 allocs/op
      BenchmarkStoreApplyBothOverlays           1,376,852 ns/op  2,329,316 B/op    30,803 allocs/op
      BenchmarkStoreApplyAllScenario            7,406,884 ns/op 12,321,060 B/op   162,776 allocs/op
      BenchmarkNodeOverlayControllerScenario       75,832 ns/op     86,400 B/op     1,080 allocs/op
```
After:

```
=== AFTER (Selective Copy-on-Write) - Memory Tests ===
      TestMemoryUsage_OverlayScenarios:
      No overlays: Total Allocs: 71509 | Bytes: 12.75 MB | Heap: 3.97 MB | Objects: 27144
      price_overlays_only: Total Allocs: 1129755 | Bytes: 44.64 MB | Heap: 3.98 MB | Objects: 27160
      capacity_overlays_only: Total Allocs: 71501 | Bytes: 64.87 MB | Heap: 3.98 MB | Objects: 27157
      mixed_overlays:  Total Allocs: 6006405 | Bytes: 323.89 MB | Heap: 2.75 MB | Objects: 15904
     
      TestMemoryUsage_ScaleWithNodePools:
      nodepools_1: Total Allocs: 1186975 | Bytes: 375.95 MB | Heap: 3.99 MB | Objects: 27192
      nodepools_5: Total Allocs: 6006364 | Bytes: 634.95 MB | Heap: 3.99 MB | Objects: 27216
      nodepools_10:  Total Allocs: 12012637 | Bytes: 1163.79 MB | Heap: 3.99 MB | Objects: 27245
      nodepools_20:  Total Allocs: 24025139 | Bytes: 2265.00 MB | Heap: 4.00 MB | Objects: 27289
      nodepools_50:  Total Allocs: 60061914 | Bytes: 5410.40 MB | Heap: 4.00 MB | Objects: 27292
     
      TestMemoryUsage_ScaleWithInstanceTypes:
      instances_50: Total Allocs: 2100180 | Bytes: 5501.46 MB | Heap: 2.76 MB | Objects: 16011
      instances_100: Total Allocs: 4200300 | Bytes: 5683.51 MB | Heap: 2.77 MB | Objects: 16035
      instances_200: Total Allocs: 6006410 | Bytes: 5943.84 MB | Heap: 2.76 MB | Objects: 15985
      instances_500: Total Allocs: 6006397 | Bytes: 6204.17 MB | Heap: 2.76 MB | Objects: 16007
    
    === AFTER (Selective Copy-on-Write) - Benchmarks ===
 
     BenchmarkStoreApply/selective-copy-apply 	   	273716 ns/op	  299578 B/op	    6723 allocs/op
     BenchmarkStoreApplyNoOverlay                5619 ns/op	       0 B/op	       0 allocs/op
     BenchmarkStoreApplyPriceOnly	       593105 ns/op	  493055 B/op	   16400 allocs/op
     BenchmarkStoreApplyCapacityOnly	  	102446 ns/op	  294401 B/op	    1000 allocs/op
     BenchmarkStoreApplyBothOverlays            	    693806 ns/op	  768427 B/op	   17201 allocs/op
     BenchmarkStoreApplyAllScenario            	       3802529 ns/op	 4127444 B/op	   91365 allocs/op
     BenchmarkNodeOverlayControllerScenario  	 	77965 ns/op	   86400 B/op	    1080 allocs/op

```

## Testing

```bash
make test-memory    # Run memory tests
make benchmark      # Run benchmarks
```

All tests pass. Comprehensive memory and benchmark tests added.

## Files

- `pkg/controllers/nodeoverlay/store.go` - Implementation
- `pkg/controllers/nodeoverlay/store_memory_test.go` - Memory tests
- `pkg/controllers/nodeoverlay/store_bench_test.go` - Benchmarks
- `Makefile` - New test targets

## Reviewers

Key areas:
1. Field sharing logic in `apply()` method
2. Selective copying in `applyPriceOverlays()`
3. Memory test validation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -92,6 +92,7 @@ require (
 )
 
 retract (
+	v1.8.1 // has a bug that prevents Karpenter from scheduling pods with specific topologySpreadConstraint configurations - https://github.com/kubernetes-sigs/karpenter/issues/2785
 	v0.100.101-test // accidentally published testing version
 	v0.35.3 // accidentally published incomplete patch release
 	v0.34.4 // accidentally published incomplete patch release


grosser:
ran this and can confirm it works 🤞 
image

GnatorX:
Thanks for testing too!

GnatorX:
A comment that came from working group meeting, @jmdeal, was that we want to ensure we won't affect the instance type cache of any provider. This code shouldn't because we are still making a copy (similar to the original implementation) but instead of a deep copy, we are doing a shallow copy. Because of this, we aren't changing any of the underlying objects directly but rather the copy of it. 

The main difference here is that rather than recreating all parts of the object, we are copying the reference to parts of the object and only change the parts that have been change. This should not affect the underlying object in anyway.

Makefile
```diff
@@ -81,6 +81,12 @@ test: ## Run tests
 		--ginkgo.v \
 		-cover -coverprofile=coverage.out -outputdir=. -coverpkg=./...
 
+test-memory: ## Run memory usage tests for node overlay store
```
DerekFrank: This raises an interesting question. We have benchmarks in other hotspots of the code, but no Makefile targets for those. I don't think thats a reason not to have these targets, I think thats a good reason to have targets for the other benchmarks.

@jmdeal what do you think?

pkg/controllers/nodeoverlay/store.go
```diff
@@ -132,21 +132,56 @@ func (s *internalInstanceTypeStore) apply(nodePoolName string, it *cloudprovider
...
-	if instanceTypeUpdate.Price != nil {
-		for _, of := range overriddenInstanceType.Offerings {
-			if overlay, ok := instanceTypeUpdate.Price[of.Requirements.String()]; ok {
-				of.ApplyPriceOverlay(lo.FromPtr(overlay.OverlayUpdate))
-			}
-		}
+	// Create a shallow copy of the instance type, sharing immutable fields
+	overriddenInstanceType := &cloudprovider.InstanceType{
+		Name:         it.Name,
+		Requirements: it.Requirements, // Shared - never modified
```
DerekFrank: If I am understanding this correctly, it does mean that cloud providers *can't* modify these, which is true for the aws provider but I don't know if it holds for the other ones

pkg/controllers/nodeoverlay/store_test.go
```diff
@@ -0,0 +1,514 @@
...
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider/fake"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+)
+
+//nolint:gocyclo
+func TestStoreApplySelectiveCopy(t *testing.T) {
```
jmdeal: We don't for benchmark tests, but everywhere else we use Ginkgo / Gomega for unit and integration testing. For consistency we should stick to that. I don't think that should be a significant refactor.

Makefile
```diff
@@ -81,6 +81,12 @@ test: ## Run tests
...
+	go test -v ./pkg/controllers/nodeoverlay/... -run TestMemoryUsage
+
+benchmark: ## Run benchmark tests for node overlay store
```
engedaam, line 87: Instead of collecting data, what do you of having hard limits and failing tests based on those limits? My thinking is that contributors won't have the data in hand so they will not have a frame of reference when the memory usage increases

GnatorX: that feels reasonable let me give it a shot

pkg/controllers/nodeoverlay/store_test.go
GnatorX: sounds good

Makefile
GnatorX: To clarify, the benchmark test is only for showing users memory usage. The store_memory_test will ensure memory usage is kept around what we expect it to be

GnatorX:
Updated based on comments. Let me know what you think.

GnatorX:
Wanted to give a nudge @jmdeal and @engedaam to see if there is anything else we wanted to updated with the PR.

jmdeal:
The only outstanding thing from me is using the Ginkgo spec framework for the test. This shouldn't be a large refactor, it's just a matter of consistency across the rest of the project. Instead of defining multiple `TestXXX(t *testing.T)` functions, we define a single test function which serves as the entry point for Ginkgo ([example](https://github.com/kubernetes-sigs/karpenter/blob/bc64ee6341b20eed08979c10c32736c90f92a0c1/pkg/events/suite_test.go#L73-L76)). I should have been more clear in my original comment that we don't just use ginkgo / gomega for assertions.

GnatorX:
Sorry for the amount of commits... my worklaptop doesn't let me run etcd binary so running tests has been a pain

GnatorX:
sorry in an attempt revert changes i screwed up the history. https://github.com/kubernetes-sigs/karpenter/pull/2790 is the new PR

PR #2677, opened 2025-12-01
test: fix flaking provisioning large replica test

DerekFrank:
Fixes: https://github.com/kubernetes-sigs/karpenter/actions/runs/19681142380/job/56375210769?pr=2626

**Description**

Occasionally, the server side nodeclaim name generation will fail to create a unique name. Ordinarily, this would be retried. In our tests, it isn't. This retries the reconciler like it would in practice to prevent flakes

make presubmit a bunch

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/karpenter
 
-go 1.25.3
+go 1.25.5
 
 require (
 	github.com/Pallinder/go-randomdata v1.2.0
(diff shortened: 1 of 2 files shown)


ellistarn:
> Occasionally, the server side nodeclaim name generation will fail to create a unique name.

What?! Can we fix this?

DerekFrank:
> What?! Can we fix this? 

Yes I think thats the obvious followup. This is mostly a patch to get the presubmits working. We can hold it if we prefer to fix the core issue.

I think solving the core issue will be tricky, as we currently generate unique names via this function: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/test/metadata.go#L41

We could up the random alphanumeric to higher than 10 to try and lower the chance of collisions, but if we want to prevent them we probably need to store unique names and handle collisions properly

ellistarn:
Were you able to reliably make this fail without the change, and then reliably make it succeed with the change?

DerekFrank:
I have discovered that the reason this was difficult to replicate locally was that the tests were running post 1.32 https://github.com/kubernetes/enhancements/issues/4420

This flake only affects < 1.32 versions.

ellistarn (approved):
/lgtm

PR #2679, opened 2025-12-02
fix: Maintain allocatable <= capacity invariant in KWOK cloudprovider

jamesmt-aws:
# Fix: Revert to using instanceType.Capacity/Allocatable() in KWOK

## Summary
Fixes a bug where allocatable could exceed capacity by reverting to the original behavior before commit 99057233.

## Prompts

| Prompt | Resolution |
|--------|------------|
| Does commit 99057233 violate the allocatable ≤ capacity invariant? | Yes - confirmed the bug exists |
| Why not just set both to requests? | Tried that, but reviewer asked why not use instanceType |
| Why not use instanceType.Capacity/Allocatable()? | Good point! That was the original code - let's revert to it |
| Did you run the full test suite? | Yes - all tests pass ✅ |

## Changes

**kwok/cloudprovider/cloudprovider.go**
- Reverted to original behavior: `Capacity: instanceType.Capacity, Allocatable: instanceType.Allocatable()`
- Removed the buggy `lo.Assign()` logic from commit 99057233
- Removed the comment about NodeOverlay (not relevant to this fix)

**kwok/cloudprovider/cloudprovider_test.go**
- Simple test demonstrating the bug in commit 99057233
- Verifies that instanceType approach maintains the invariant

## Benefits

- **Correctness**: instanceType.Allocatable() is already ≤ instanceType.Capacity
- **Simplicity**: No special logic needed - just use instance type values
- **Consistency**: Matches real node behavior more closely
- **Tested**: Full test suite passes (all pkg/... tests)

## Testing

**Unit test** shows:
- Original approach: capacity=4 CPU/8Gi, allocatable=3.9 CPU/7.5Gi ✅
- Buggy approach: capacity=2 CPU/4Gi, allocatable=3.9 CPU/7.5Gi ❌

**Full test suite:** All tests in `pkg/...` pass

## Root Cause

Commit 99057233 tried to be clever by merging `nodeClaim.Spec.Resources.Requests` with `instanceType.Allocatable()`:

```go
Capacity:    nodeClaim.Spec.Resources.Requests,  // 2 CPU, 4Gi (pod requests)
Allocatable: lo.Assign(requests, instanceType.Allocatable()),  // 3.9 CPU, 7.5Gi (instance type)
```

This violated the invariant because:
1. Capacity was set to pod requests (2 CPU)
2. Allocatable came from instance type (3.9 CPU)
3. Result: allocatable > capacity ❌

**The fix:** Revert to the original simple approach that just uses instance type values for both.


kwok/cloudprovider/cloudprovider.go
```diff
@@ -229,12 +229,26 @@ func (c CloudProvider) toNode(nodeClaim *v1.NodeClaim) (*corev1.Node, error) {
 			// so we only apply resource requests, since NodeOverlay will not apply.
 			// If this changes in the future, we'll need to update capacity and allocatable values for KWOK nodes.
 			Capacity:    nodeClaim.Spec.Resources.Requests,
-			Allocatable: lo.Assign(nodeClaim.Spec.Resources.Requests, instanceType.Allocatable()),
+			Allocatable: minResourceList(nodeClaim.Spec.Resources.Requests, instanceType.Allocatable()),
```
ellistarn: I'm not deep on this code, but it strikes me as odd that we're doing different thinks for Capacity and Allocatable here. Why is the logic different?

jamesmthompson: After a bit of conversation, I've negotiated to something like this. Would you like this better? 

```
Capacity:    nodeClaim.Spec.Resources.Requests,
Allocatable: nodeClaim.Spec.Resources.Requests,
```

PR #2684, opened 2025-12-03
test: Added an scaling interference benchmark test 

nathangeology:
**Description**
Added an 'interference' scaling benchmark where you start by scaling a deployment to 500 pods with host name tse. Then once the cluster stabilizes, you reduce that deployment by 50% to 250 pods and create a new deployment with 500 new pods at the same time. This benchmark tests the ability of the system to stabilize and work out capacity when there is both pod scale in and scale out actions occurring at the same time. 

### DIFF
--- a/.github/actions/install-pyroscope/action.yaml
+++ b/.github/actions/install-pyroscope/action.yaml
@@ -3,7 +3,6 @@ description: 'Installs Pyroscope'
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
     - name: download pyroscope
       shell: bash
       run: | 


.github/workflows/kind-perf-e2e.yaml
```diff
@@ -31,6 +31,8 @@ jobs:
             focus: "Do Not Disrupt Performance Test"
           - name: "Provisioning"
             focus: "Provisioning"
+          - name: "Host Name Spreading Interference"
+            focus: "Host Name Spreading Deployment Interference"
```
ellistarn: What's leading us to test host name spreading? Hostname topology in an autoscaled cluster is more or less an antipattern because the number of nodes is dynamic.

nathangeology: The fast answer is that its an easy button for making provisioning and consolidation harder. The longer answer is that its a straightforward way to force more topology domains and that seems to slow a lot of Karpenter processes down in general. There are other more complicated ways to force more topology domains (anti-affinity, spreading on labels, having lots of taints/tolerations to worry about, etc), but we'll see if we need more benchmarks on those over time based on issues/feedback. In any case, its worth tracking our performance with lots of topology domains and we may make changes in the future that improve performance for those that use it despite us telling them its unwise.

ellistarn: I think we should focus on test cases that mirror common customer usage patterns. Happy to chat on the patterns I'm most familiar with. The pathological case I've seen is a configuring pod hostname anti-affinity for a large deployment. It's similar to what you have, but much more commonly used in practice.

nathangeology: Ok, switched the test to use hostname anti-affinity. I will see about doing a separate PR for the existing tests for this.

test/suites/performance/interference_test.go
```diff
@@ -0,0 +1,129 @@
...
+				"Total scale-out time should be less than 10 minutes")
+			Expect(scaleOutReport.TotalNodes).To(BeNumerically("<", 650),
+				"Should not require more than 1000 nodes for 1000 pods")
+			Expect(scaleOutReport.TotalReservedCPUUtil).To(BeNumerically(">", 0.55),
+				"Average CPU utilization should be greater than 55%")
+			Expect(scaleOutReport.TotalReservedMemoryUtil).To(BeNumerically(">", 0.75),
+				"Average memory utilization should be greater than 75%")
+
+			// ========== PHASE 2: Interference Scale Out TEST ==========
+			By("Net scaling out scaling out interference test")
```
ellistarn:
```suggestion
			By("Net scaling out interference test")
```

test/suites/performance/interference_test.go
```diff
@@ -0,0 +1,129 @@
...
+			scaleOutReport, err := ReportScaleOutWithOutput(env, "Self Anti-Affinity Performance Test", 500, 15*time.Minute, "self_antiaffinity_scale_out_small")
+			Expect(err).ToNot(HaveOccurred(), "Scale-out should execute successfully")
+
+			By("Validating scale-out performance with self anti-affinity")
+			Expect(scaleOutReport.TestType).To(Equal("scale-out"), "Should be detected as scale-out test")
+			Expect(scaleOutReport.TotalPods).To(Equal(500), "Should have 500 total pods")
+
+			// Performance assertions - self anti-affinity requires one pod per node
+			Expect(scaleOutReport.TotalTime).To(BeNumerically("<", 5*time.Minute),
+				"Total scale-out time should be less than 10 minutes")
```
ellistarn: These times aren't making sense to me. 15min above, 5 min assertion here, 10m error message. Seems like we let it run for 15 and then fail, so we can catch failures that are between 5 and 15 mins. That's reasonable. May just need a fresh error message.

ellistarn:
Thanks! LGTM other than the callouts.

test/suites/performance/interference_test.go
nathangeology: Yep, just missed updating the string to align with the expectation here. The 15 min above is a timeout to stop the test and throw an exception if the test hangs for some reason. I fixed the string mismatch you called out and double checked the others for good measure.

ellistarn (approved):
/ok-to-test

.github/actions/install-prometheus/action.yaml
```diff
@@ -3,6 +3,7 @@ description: 'Installs prometheus'
 runs:
   using: "composite"
   steps:
+    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
```
DerekFrank: Did you mean to grab these from main?

nathangeology: Nope! I hadn't notice that these were in there.

nathangeology: Removed them, let's see if the presubmits flake or not

DerekFrank (approved):
/lgtm

PR #2685, opened 2025-12-03
feat: add configurable additional labels to NodePool metrics

### LINKED ISSUE #2221: Allow setting additional metric labels
### Description

we have custom labels on our node-pools that we'd like to report in the metrics as well,
so we can graph for example "version", "spot", "node-type", "customer" etc custom labels

medium: we have to read and re-emit karpenters metrics atm to add the metadata from the `nodepool` it reports on

moko-poi:
## What problem does this PR solve?

Closes #2221

Allows users to customize which NodePool labels are included in NodePool metrics (`karpenter_nodepools_limit` and `karpenter_nodepools_usage`), enabling better observability for custom attributes like capacity type, availability zone, and architecture.

### Implementation approach

Added a CLI flag `--additional-nodepool-metric-labels` (environment variable: `ADDITIONAL_NODEPOOL_METRIC_LABELS`) that accepts a comma-separated list of label keys to include in NodePool metrics.

**Default behavior**: No additional labels (empty by default). Users must explicitly configure which labels to include.

### Key design decisions

1. **CLI flag instead of annotations**: Satisfies Prometheus constraint that all time series for a metric must have identical label keys. Label keys are fixed at startup when metrics are registered.

2. **Deferred initialization**: Follows the same pattern as Node metrics controller - metrics are declared as package variables but initialized in `initializeMetrics()` called from `NewController()`. This ensures options are available when building the label key set.

3. **Empty string for missing labels**: If a NodePool doesn't have a specified label, the metric uses an empty string `""` for that label value. This maintains consistent label keys across all NodePools while allowing flexibility.

4. **No defaults**: By not providing default labels, this PR stays focused on providing the capability without making assumptions about which labels exist on users' NodePools.

### Example usage

```bash
# No additional labels (default)
karpenter

# Add custom labels for capacity type, zone, and architecture
karpenter --additional-nodepool-metric-labels=capacity_type,zone,architecture

# Or use environment variable
export ADDITIONAL_NODEPOOL_METRIC_LABELS=environment,team,version
karpenter
```

With a NodePool:
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: production-spot
  labels:
    capacity_type: spot
    zone: us-east-1a
    architecture: amd64
spec:
  limits:
    cpu: "1000"
```

When configured with --additional-nodepool-metric-labels=capacity_type,zone,architecture, produces metrics:
```
karpenter_nodepools_limit{nodepool="production-spot",resource_type="cpu",capacity_type="spot",zone="us-east-1a",architecture="amd64"} 1000
```

Without the flag, produces metrics:
```
karpenter_nodepools_limit{nodepool="production-spot",resource_type="cpu"} 1000
```

## Testing

- Added comprehensive test coverage in suite_test.go:
  - Verifies additional labels are included in metrics when configured
  - Validates empty string behavior for missing labels
  - All 6 tests passing

## Checklist

- Added CLI flag with empty default
- Implemented deferred metric initialization
- Updated controller wiring to pass context
- Added test helper support
- Comprehensive test coverage (6/6 passing)
- Follows existing Node metrics pattern for consistency

### DIFF
--- a/pkg/controllers/metrics/nodepool/controller.go
+++ b/pkg/controllers/metrics/nodepool/controller.go
@@ -19,6 +19,7 @@ package nodepool
 import (
 	"context"
 	"strings"
+	"sync"
 	"time"
 
 	opmetrics "github.com/awslabs/operatorpkg/metrics"
@@ -37,37 +38,19 @@ import (
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 	"sigs.k8s.io/karpenter/pkg/state/cost"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
 )
 
+const (
+	resourceTypeLabel = "resource_type"
+	nodePoolNameLabel = "nodepool"
+)
+
 var (
-	Limit = opmetrics.NewPrometheusGauge(
-		crmetrics.Registry,
-		prometheus.GaugeOpts{
-			Namespace: metrics.Namespace,
-			Subsystem: metrics.NodePoolSubsystem,
-			Name:      "limit",
-			Help:      "Limits specified on the nodepool that restrict the quantity of resources provisioned. Labeled by nodepool name and resource type.",
-		},
-		[]string{
-			metrics.ResourceTypeLabel,
-			metrics.NodePoolLabel,
-		},
-	)
-	Usage = opmetrics.NewPrometheusGauge(
-		crmetrics.Registry,
-		prometheus.GaugeOpts{
-			Namespace: metrics.Namespace,
-			Subsystem: metrics.NodePoolSubsystem,
-			Name:      "usage",
-			Help:      "The amount of resources that have been provisioned for a nodepool. Labeled by nodepool name and resource type.",
-		},
-		[]string{
-			metrics.ResourceTypeLabel,
-			metrics.NodePoolLabel,
-		},
-	)
+	Limit       opmetrics.GaugeMetric
+	Usage       opmetrics.GaugeMetric
 	ClusterCost = opmetrics.NewPrometheusGauge(
 		crmetrics.Registry,
 		prometheus.GaugeOpts{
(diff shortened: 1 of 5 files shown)


moko-poi:
The test is failing, but we've addressed it in https://github.com/kubernetes-sigs/karpenter/pull/2687#issuecomment-3606518843 .

pkg/controllers/metrics/nodepool/controller.go
```diff
@@ -131,11 +146,23 @@ func getLimits(nodePool *v1.NodePool) corev1.ResourceList {
...
 		nodePoolNameLabel: nodePool.Name,
 	}
+
+	// Add values for additional labels from CLI flags
+	for _, labelKey := range opts.AdditionalNodePoolMetricLabels {
+		if value, ok := nodePool.Labels[labelKey]; ok {
+			labels[labelKey] = value
+		} else {
+			labels[labelKey] = ""
+		}
```
grosser: already produces empty string on missing
```suggestion
		labels[labelKey] = nodePool.Labels[labelKey]
```

pkg/controllers/metrics/nodepool/suite_test.go
```diff
@@ -160,4 +168,51 @@ var _ = Describe("Metrics", func() {
...
 		}
 	})
+
+	It("should include additional labels from CLI flags", func() {
+		// Set NodePool with additional labels
+		nodePool.Labels = map[string]string{
+			"capacity_type": "spot",
+			"zone":          "us-east-1a",
+			"architecture":  "amd64",
+		}
```
grosser, line 184: do these labels even commonly exist ?
I was hoping zone would translate to `topology.kubernetes.io/zone`
... maybe have that as the only default since it is a well known label ?
(alternatively have no defaults so this is noop for users that don't care)

pkg/operator/options/options.go
```diff
@@ -157,6 +161,16 @@ func (o *Options) Parse(fs *FlagSet, args ...string) error {
...
 	o.MinValuesPolicy = MinValuesPolicy(o.minValuesPolicyRaw)
+
+	// Parse comma-separated additional nodepool metric labels
+	if o.rawAdditionalNodePoolMetricLabels != "" {
+		o.AdditionalNodePoolMetricLabels = lo.Map(strings.Split(o.rawAdditionalNodePoolMetricLabels, ","), func(s string, _ int) string {
+			return strings.TrimSpace(s)
+		})
+	} else {
+		o.AdditionalNodePoolMetricLabels = []string{}
+	}
```
grosser: nit: less code paths 🤷 
```suggestion
	o.AdditionalNodePoolMetricLabels = lo.Map(strings.Split(o.rawAdditionalNodePoolMetricLabels, ","), func(s string, _ int) string {
		return strings.TrimSpace(s)
	})
```

grosser (commented):
logic makes sense
idk about the defaults

pkg/controllers/metrics/nodepool/controller.go
moko-poi: Good catch! Fixed in bbe404f

pkg/operator/options/options.go
moko-poi: [3377cac](https://github.com/kubernetes-sigs/karpenter/pull/2685/commits/3377cac3e29a246b8f504f3de5251e2425e578d0)
Updated to use lo.FilterMap to reduce code paths as suggested. I also added filtering for empty strings since strings.Split("", ",") returns []string{""} rather than an empty slice, which would cause invalid empty label keys in the metrics controller.

pkg/controllers/metrics/nodepool/suite_test.go
moko-poi: @grosser [f461d94](https://github.com/kubernetes-sigs/karpenter/pull/2685/commits/f461d946d27fb042e51c9d4c54daf09ebebb6922)
I've removed the default values entirely. The default is now an empty string, making this a no-op unless users explicitly configure it. This keeps the PR focused on providing the capability for users to configure additional metric labels, without making assumptions about which labels exist on their NodePools.

moko-poi:
@grosser 
I've updated the implementation to remove the default labels (`capacity_type`, `zone`, `architecture`). The default is now an empty string, making this feature a no-op unless users explicitly configure it.

This keeps the PR scope focused on **providing the capability** for users to add custom labels to their metrics, rather than imposing specific defaults that may not match their label schemas.

grosser (approved):
✨

moko-poi:
@engedaam @tallaxes 

Could you please review this PR when you have a chance? This feature adds configurable additional labels to NodePool metrics, which has been approved by @grosser. Looking forward to your feedback.

Thank you!

grosser:
@jmdeal could you please take a look at this PR, waiting for a review for very long already 😞

PR #2688, opened 2025-12-03
fix: do not log on missing labels

DerekFrank:
Fixes #N/A 

**Description**

The cluster cost calculator, designed to track Karpenter's perspective of the total cost of the cluster. This allows users to validate the effects of configurations such as NodeOverlay's Price Adjustment, NodePool Selectors, NodePool Consolidation configuration, etc.  These errors are expected before labels are propagated up to the NodeClaim from the Node, we should not log an error on them.

`make presubmit`


pkg/state/cost/cost.go
ellistarn, line 209: This metric will increase, right? Is this even a problem?

DerekFrank: Great catch

PR #2691, opened 2025-12-04
test: Fix benchmark tests after PR #2673 broke them due to toolchain building changes

nathangeology:
**Description**
PR #2673 broke the perf-e2e benchmark tests in a few different ways. One, it broke the ko package which has a dependency conflict with the go container registry package. Two, the github actions version hashes weren't all updated together which resulted in header errors due to versioning conflicts. Lastly, some of the build-with-kind type makefile actions and setup procedures needed some updates due to make toolchain going away. 

Running perf-e2e-tests and presubmit tests on the changes. 

### DIFF
--- a/.github/actions/install-deps/action.yaml
+++ b/.github/actions/install-deps/action.yaml
@@ -28,4 +28,6 @@ runs:
         go tool setup-envtest use -p path ${{ inputs.k8sVersion }} | xargs -I {} sudo mv {} /usr/local/kubebuilder/bin
     - name: Install ko
       shell: bash
-      run: go install github.com/google/ko@v0.14.1
+      run: |
+        curl -sSfL https://github.com/ko-build/ko/releases/download/v0.15.4/ko_0.15.4_Linux_x86_64.tar.gz | sudo tar xzf - -C /usr/local/bin ko
+        sudo chmod +x /usr/local/bin/ko


.github/actions/install-deps/action.yaml
```diff
@@ -26,3 +26,8 @@ runs:
       run: |
         sudo mkdir -p /usr/local/kubebuilder
         go tool setup-envtest use -p path ${{ inputs.k8sVersion }} | xargs -I {} sudo mv {} /usr/local/kubebuilder/bin
+    - name: Install ko
+      shell: bash
```
ellistarn, line 30: Can we install ko with go tool?

Makefile
```diff
@@ -110,7 +110,7 @@ verify: ## Verify code. Includes codegen, docgen, dependencies, linting, formatt
 	@# Use perl instead of sed due to https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux
 	@# We need to do this "sed replace" until controller-tools fixes this parameterized types issue: https://github.com/kubernetes-sigs/controller-tools/issues/756
 	@perl -i -pe 's/sets.Set/sets.Set[string]/g' pkg/scheduling/zz_generated.deepcopy.go
-	go tool nwa config -c add
```
ellistarn, line 113: Why did we remove this one?

.github/workflows/e2e.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
       run: |
         kubectl config current-context
         kubectl get nodes
-        kubectl taint nodes chart-testing-control-plane CriticalAddonsOnly:NoSchedule
+        kubectl taint nodes --all CriticalAddonsOnly:NoSchedule
```
ellistarn, line 46: Why this change?

.github/actions/install-deps/action.yaml
nathangeology: There was an issue with doing that related to this: https://github.com/ko-build/ko/issues/1550. It's a dependency conflict between `ko` and `go-containerregistry` and the only way around it was to install the binary separately outside of go tool

.github/workflows/e2e.yaml
nathangeology: It just makes it more robust to how the KIND cluster is named

Makefile
nathangeology: Mainly because I switched back to the hack/boilerplate.sh approach. We could go with the go tool nwa approach or the boilerplate approach, but it needs to be consistent either way

.github/workflows/e2e.yaml
ellistarn: Got it -- was trying to figure out how my toolchain change broke this.

Makefile
ellistarn: I intentionally moved us to nwa. It's in use in a couple of other projects. Let's use it!

nathangeology:
I'm likely going to close this pull request since the issues covered here are going to be resolved in https://github.com/kubernetes-sigs/karpenter/pull/2602

nathangeology:
Issues addressed by this PR addressed in: https://github.com/kubernetes-sigs/karpenter/pull/2602

PR #2695, opened 2025-12-06
chore: Reduce pod scheduling error log spam with deduplication

### LINKED ISSUE #1904: Karpenter runtime logs showing tons of record: incompatible with nodepool , daemonset overhead
### Description

**Version**:
Karpenter Version: v1.0.6
Kubernetes Version: v1.31

**Context:**

 There are some pods (not daemonset) weren't scheduled to the dedicated nodes successfully according to karpenter logs as below:

`"incompatible with nodepool \"gpu\", daemonset overhead={\"cpu\":\"605m\",\"memory\":\"1288Mi\",\"pods\":\"12\"}, did not tolerate nvidia.com/gpu=1:NoSchedule;` 

`incompatible with nodepool \"app\", daemonset overhead={\"cpu\":\"605m\",\"memory\":\"1288Mi\",\"pods\":\"12\"}, incompatible requirements, label \"eks.amazonaws.com/nodegroup\" does not have known values"`

As per design we don't want these pods are scheduled to the dedicated nodes by using node taints, pod toleration/nodeSelector, whereas it did actually, the good thing is it's failed.

As of now It don't impact our business, everything looks good due to above failure, but those tons of error message in karpenter logs. **We'd like to know how to avoid it happen then clear those error messages.**

**Other Information**:

**We have two nodepools, these are gpu and app.**

gpu has its own taints as below  

`key    = "nvidia.com/gpu"`
`value  = "1"`
`effect = "NoSchedule"`

**app nodepool doesn't have taints, but it has startup_taints as below** 

`key    = "node.cilium.io/agent-not-ready"`
`value  = "true"`
`effect = "NoExecute"`
      

**We also have two node groups managed by AWS ASG, one is for karpenter, the other is for infrastructure addons.** 

karpenter node group is a dedicated node, only accept karpenter pods.

Infrastructure node group only accept infrastructure addons, which has taints and label 

Taint:

`key    = "node-group"`
`value  = "infra"`
`effect = "NO_SCHEDULE"`

node label:

`eks.amazonaws.com/nodegroup=non-prod-uw2-blue-infra-nodegroup`

 **all infra addons should be running at Infrastructure node group** 

-------------

The one of infra addon, Istio pod should be scheduled to infra node group instead of gpu / app nodepool, but it did actually according to karpenter logs, Istio has toleration and nodeselector as below:

`tolerations:`  

   `\- key: "node-group"`

​    `operator: "Exists"`

   `nodeSelector:`

​    `eks.amazonaws.com/nodegroup: ${CLUSTER_NAME}-infra-nodegroup`

-------------------

We spent some time to investigate it, but no luck, still can't find root cause, so have to raise issue here. 

**We'd like to know how to avoid it happen then clear those error messages if possible.**

moko-poi:
## Description
Implements log deduplication for pod scheduling errors to address excessive logging when pods cannot be scheduled.

## Problem
Fixes #1904

When pods cannot be scheduled (e.g., due to missing NodePool, taint intolerance, or resource constraints), the scheduler logs an error on every scheduling attempt. This results in 100k+ log entries per hour for a single unschedulable pod, causing:
- Log storage bloat
- Difficulty in identifying actual issues
- Performance overhead from excessive I/O

## Solution
Added cache-based log deduplication using a 5-minute timeout window:
- Logs are deduplicated per pod UID + error message combination
- Different errors for the same pod are logged separately
- Same error is logged again after 5 minutes if still occurring
- Matches the existing event deduplication timeout for consistency

## Changes
- Modified `pkg/controllers/provisioning/scheduling/scheduler.go`:
  - Added `logErrorCache` to `Results` struct
  - Implemented `shouldLogPodError()` method with composite cache key
  - Integrated cache check before logging scheduling errors
- Added `pkg/controllers/provisioning/scheduling/log_deduplication_test.go`:
  - Comprehensive test coverage for deduplication scenarios
  - Validates different pods, same errors, different errors, and timeout expiration

## Testing
- All existing tests pass (336/337 specs)
- New deduplication tests validate correct behavior
- Verified different error types for same pod are logged separately
- Confirmed timeout allows re-logging after 5 minutes

### DIFF
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/awslabs/operatorpkg/option"
 	"github.com/awslabs/operatorpkg/serrors"
+	"github.com/patrickmn/go-cache"
 	"github.com/samber/lo"
 	"go.uber.org/multierr"
 	corev1 "k8s.io/api/core/v1"
@@ -48,6 +49,12 @@ import (
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 	"sigs.k8s.io/karpenter/pkg/utils/pod"
 	"sigs.k8s.io/karpenter/pkg/utils/resources"
+)
+
+const (
+	// podSchedulingErrorLogTimeout defines how long to wait before logging the same pod scheduling error again.
+	// This prevents log spam when a pod repeatedly fails to schedule for expected reasons (e.g., incompatible NodePool).
+	podSchedulingErrorLogTimeout = 5 * time.Minute
 )
 
 type ReservedOfferingMode int
@@ -234,6 +241,7 @@ type Results struct {
 	NewNodeClaims []*NodeClaim
 	ExistingNodes []*ExistingNode
 	PodErrors     map[*corev1.Pod]error
+	logErrorCache *cache.Cache // cache to deduplicate pod scheduling error logs
 }
 
 // Record sends eventing and log messages back for the results that were produced from a scheduling run
@@ -250,7 +258,12 @@ func (r Results) Record(ctx context.Context, recorder events.Recorder, cluster *
 			log.FromContext(ctx).WithValues("Pod", klog.KObj(p)).Info("skipping pod with Dynamic Resource Allocation requirements, not yet supported by Karpenter")
 			continue
 		}
-		log.FromContext(ctx).WithValues("Pod", klog.KObj(p)).Error(err, "could not schedule pod")
+		// Deduplicate log messages for the same pod+error combination to reduce log spam
+		// Different errors for the same pod will still be logged
+		// Events are already deduplicated by the PodFailedToScheduleEvent DedupeTimeout
+		if r.shouldLogPodError(p, err) {
+			log.FromContext(ctx).WithValues("Pod", klog.KObj(p)).Error(err, "could not schedule pod")
+		}
 		recorder.Publish(PodFailedToScheduleEvent(p, err))
 	}
 	for _, existing := range r.ExistingNodes {
@@ -293,6 +306,24 @@ func (r Results) DRAErrors() map[*corev1.Pod]error {
 	return lo.PickBy(r.PodErrors, func(_ *corev1.Pod, err error) bool {
 		return IsDRAError(err)
 	})
+}
+
+// shouldLogPodError checks if we should log an error for this pod.
+// Returns true if this is the first time seeing this specific pod+error combination,
+// or if enough time has passed since the last log for this combination.
+// This allows different errors for the same pod to be logged, while deduplicating repeated identical errors.
+func (r Results) shouldLogPodError(p *corev1.Pod, err error) bool {
+	if r.logErrorCache == nil {
+		return true
+	}
+	// Include both pod UID and error message in the cache key to allow different errors
+	// for the same pod to be logged separately
+	key := fmt.Sprintf("%s:%s", p.UID, err.Error())
+	if _, found := r.logErrorCache.Get(key); found {
+		return false
+	}
+	r.logErrorCache.Set(key, nil, podSchedulingErrorLogTimeout)
+	return true
 }
 
 func (r Results) NodePoolToPodMapping() map[string][]*corev1.Pod {
@@ -428,6 +459,7 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
 		NewNodeClaims: s.newNodeClaims,
 		ExistingNodes: s.existingNodes,
 		PodErrors:     podErrors,
+		logErrorCache: cache.New(podSchedulingErrorLogTimeout, 10*time.Minute),
 	}, ctx.Err()
 }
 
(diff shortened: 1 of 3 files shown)


moko-poi:
Hi @njtran and @tallaxes,

I'd appreciate your review on this PR when you have a moment.

**Overview:**
This PR addresses issue #1904 by implementing log deduplication for pod scheduling errors. Currently, unschedulable pods generate 100k+ log entries per hour, causing log storage bloat and making it difficult to identify real issues.

**Key Changes:**
- Added cache-based deduplication with a 5-minute timeout window (consistent with existing event deduplication)
- Logs are deduplicated per pod UID + error message combination
- Different errors for the same pod are still logged separately
- Refactored `Results.Record()` method to satisfy gocyclo requirements

**Testing:**
- All existing tests pass (336/337 specs)
- Added comprehensive unit tests for deduplication scenarios
- Coverage: 70.83% for changed lines

I'd love to hear your feedback on the approach, especially regarding:
1. The 5-minute timeout duration
2. The cache key structure (pod UID + error message)
3. Any edge cases I might have missed

Thank you for taking the time to review! 🙏

GnatorX:
Let me take a look
/assign gnatorx

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -428,6 +463,7 @@ func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, err
 		NewNodeClaims: s.newNodeClaims,
 		ExistingNodes: s.existingNodes,
 		PodErrors:     podErrors,
+		logErrorCache: cache.New(podSchedulingErrorLogTimeout, 10*time.Minute),
```
GnatorX: I think this won't work, happy to take over, but this is new for every Solve which means the cache will be tossed out and you won't actually get any dedup

moko-poi:
@GnatorX Thanks for the review and the offer to take it over. I have moved the dedup cache to the long-lived Provisioner and wired it through scheduler.WithLogErrorCache so it now actually deduplicates across Solve calls, and intentionally omitted it from the disruption simulation path to avoid the NopLogger-suppressed errors polluting the shared cache and silently hiding real provisioning logs. I verified end to end on a kind+kwok cluster that the same 2 unschedulable pods went from 29 error lines over 15 reconciles down to 2.
/remove-lifecycle rotten

moko-poi:
For reference, here are the raw logs from a kind+kwok cluster running the same workload (one NodePool with a NoSchedule taint, two pods that do not tolerate it) for sixty seconds on both branches.

Without this PR (main, commit 86026aa) the same pod+error pair was logged on every reconcile, producing 29 ERROR lines across 15 reconciles:

```
{"level":"ERROR","time":"2026-06-10T08:48:20.630Z","message":"could not schedule pod","reconcileID":"27c9d54f-d830-403e-8296-ea9c32f6f487","Pod":{"name":"unsched-2"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
{"level":"ERROR","time":"2026-06-10T08:48:32.139Z","message":"could not schedule pod","reconcileID":"312ec859-3c29-45c0-bda7-b74d933dc221","Pod":{"name":"unsched-2"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
{"level":"ERROR","time":"2026-06-10T08:48:32.139Z","message":"could not schedule pod","reconcileID":"312ec859-3c29-45c0-bda7-b74d933dc221","Pod":{"name":"unsched-1"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
{"level":"ERROR","time":"2026-06-10T08:48:42.138Z","message":"could not schedule pod","reconcileID":"67bd7833-0d62-46aa-bef1-66175848ecf2","Pod":{"name":"unsched-2"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
{"level":"ERROR","time":"2026-06-10T08:48:42.138Z","message":"could not schedule pod","reconcileID":"67bd7833-0d62-46aa-bef1-66175848ecf2","Pod":{"name":"unsched-1"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
...24 more identical lines across 13 more reconciles
```

With this PR (commit a3da649 plus the fix), the same workload over 14 reconciles produces only 2 ERROR lines, fired on the first reconcile and suppressed for the rest of the 5-minute window:

```
{"level":"ERROR","time":"2026-06-10T08:51:42.667Z","message":"could not schedule pod","reconcileID":"b0a1ac6a-b6d1-43f5-a332-820cc9f926cf","Pod":{"name":"unsched-2"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
{"level":"ERROR","time":"2026-06-10T08:51:42.668Z","message":"could not schedule pod","reconcileID":"b0a1ac6a-b6d1-43f5-a332-820cc9f926cf","Pod":{"name":"unsched-1"},"error":"did not tolerate taint (taint=special-workload=true:NoSchedule)"}
```

jmdeal:
I'd be very hesitant to introduce this type of deduplication on the controller side. If anything, we're looking at increasing the verbosity of a lot of our logs (particularly at debug). A common pattern we use today to reduce log count is to group impacted resources into a single log line and truncate the list of resources, but this makes it difficult to trace out this history of actions performed against a resource. For this particular log line, this could make it difficult to understand if a pod wasn't scheduled due to an error, or if a bug in Karpenter caused it not to be evaluated during that simulation.

The tenant I would take is that we shouldn't discard information at the application level. Operators have the option to discard logs that they don't need to retain at ingestion time (e.g. via fluent bit) or via some other post-processing pipeline. 

We will still want to take an opinionated on some logs being too expensive or too verbose to be on by default. That's what we should use DEBUG logs for. Of course in this case it's an error, but I think it's actually indicative of a different problem — the problem isn't that we're logging too many times, it's that we're retrying the pod over and over again when nothing has changed in cluster state that would indicate it's now schedulable. For this particular problem we should address it at it's source rather than patching over it with a log cache.

PR #2706, opened 2025-12-09
chore: ensure serrors wrap with even number of keys and values

ryan-mist:
Fixes #N/A 

**Description**
* ensure serrors.Wrap has equal number of keys and values 

* make presubmit

### DIFF
--- a/pkg/state/cost/suite_test.go
+++ b/pkg/state/cost/suite_test.go
@@ -166,7 +166,7 @@ var _ = Describe("ClusterCost", func() {
 		It("should update costs correctly for duplicate spot offering to increment count", func() {
 			// Setup: Add initial spot offering to create baseline
 			nodeClaim := createTestNodeClaim(testNodePool, testInstanceType.Name, spotOffering.CapacityType(), spotOffering.Zone())
-			clusterCost.UpdateNodeClaim(ctx, nodeClaim)
+			Expect(clusterCost.UpdateNodeClaim(ctx, nodeClaim)).To(Succeed())
 
 			RunUpdateNodeClaimTest(
 				clusterCost,
@@ -182,7 +182,7 @@ var _ = Describe("ClusterCost", func() {
 		It("should update costs correctly for different offering types to same nodepool", func() {
 			// Setup: Add spot offering first, then test will add on-demand offering
 			nodeClaim := createTestNodeClaim(testNodePool, testInstanceType.Name, spotOffering.CapacityType(), spotOffering.Zone())
-			clusterCost.UpdateNodeClaim(ctx, nodeClaim)
+			Expect(clusterCost.UpdateNodeClaim(ctx, nodeClaim)).To(Succeed())
 
 			RunUpdateNodeClaimTest(
 				clusterCost,
@@ -198,7 +198,7 @@ var _ = Describe("ClusterCost", func() {
 		It("should update costs correctly for offering to different nodepool", func() {
 			// Setup: Add spot offering to first nodepool, then test will add on-demand to second nodepool
 			nodeClaim := createTestNodeClaim(testNodePool, testInstanceType.Name, spotOffering.CapacityType(), spotOffering.Zone())
-			clusterCost.UpdateNodeClaim(ctx, nodeClaim)
+			Expect(clusterCost.UpdateNodeClaim(ctx, nodeClaim)).To(Succeed())
 
 			RunUpdateNodeClaimTest(
 				clusterCost,
@@ -216,7 +216,7 @@ var _ = Describe("ClusterCost", func() {
 		It("should update costs correctly for single spot offering from nodepool with one offering", func() {
 			// Setup: Add one spot offering to remove
 			nodeClaim := createTestNodeClaim(testNodePool, testInstanceType.Name, spotOffering.CapacityType(), spotOffering.Zone())
-			clusterCost.UpdateNodeClaim(ctx, nodeClaim)
+			Expect(clusterCost.UpdateNodeClaim(ctx, nodeClaim)).To(Succeed())
 
 			RunDeleteNodeClaimTest(
 				clusterCost,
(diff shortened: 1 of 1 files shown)


pkg/state/cost/cost.go
```diff
@@ -233,7 +232,7 @@ func (cc *ClusterCost) UpdateNodeClaim(ctx context.Context, nodeClaim *v1.NodeCl
 	defer cc.Unlock()
 	err := cc.internalAddOffering(ctx, np, nodeClaim.Labels[corev1.LabelInstanceTypeStable], nodeClaim.Labels[v1.CapacityTypeLabelKey], nodeClaim.Labels[corev1.LabelTopologyZone], true)
 	if err != nil {
-		log.FromContext(ctx).Error(serrors.Wrap(err, "failed to add offering for nodeclaim", "nodeclaim", klog.KObj(nodeClaim), "nodepool", klog.KObj(np)), "failed to process nodeclaim for cost tracking")
+		log.FromContext(ctx).Error(serrors.Wrap(fmt.Errorf("failed to add offering for nodeclaim, %w", err), "nodeclaim", klog.KObj(nodeClaim), "nodepool", klog.KObj(np)), "failed to process nodeclaim for cost tracking")
```
ellistarn: This pattern doesn't look like idiomatic use of serr, which is what led to this confusion in the first place. I'd expect one error message from this line. We have:
```
failed to add offering for nodeclaim
```
and 
```
ailed to process nodeclaim for cost tracking
```

sugg collapsing these messages into something like:
```
log.FromContext(ctx).Error(serrors.Wrap(err, "nodeclaim", klog.KObj(nodeClaim), "nodepool", klog.KObj(np), "failed to add offering during nodeclaim cost tracking")
```

---

Thinking further, the entire point of wrapping errors is to propagate them upwards so that their context can be handled/logged appropriately from the caller. This entire pattern seems wrong to me. Why are we both wrapping and logging, when we could just pass the values directly into the logger.


---

Thinking even further, why are we not returning the error and letting the caller context handler like we do in the rest of our reconcilers? It's very uncommon to swallow an error in an internal helper like this rather than propagating it upwards. Maybe @jmdeal can provide some context on the bigger picture here.

DerekFrank: I have the context here. In the PR for this code, @jmdeal suggested handling the error in the parent reconcilers instead of here. I argued since the function calls `DeleteNodeClaim` and `UpdateNodeClaim` are only being used in one place each and that one place would want to swallow the errors, its fine to swallow them here, at least for now. 

Its a fairly easy refactor to pass the errors up to the parent reconciler to swallow there instead if you feel strongly

ellistarn: I don't feel strongly. Just not the typical pattern I see in reconcilers.

Return-on-error is a common error handling pattern that allows me to turn my brain off. It makes it really easy to guarantee that your control system doesn't inadvertently continue with logic after an error when you shouldn't. 

There are cases where this isn't the right behavior and you want to continue after an error, which is where you typically see the "swallowing" logic.

If you still see the value in your approach (I don't), feel free to keep it. I care more about a fix to the logging issues.

pkg/controllers/state/informer/nodeclaim.go
```diff
@@ -62,15 +63,19 @@ func (c *NodeClaimController) Reconcile(ctx context.Context, req reconcile.Reque
...
 		}
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
 	if !nodeclaimutils.IsManaged(nodeClaim, c.cloudProvider) {
 		return reconcile.Result{}, nil
 	}
 	c.cluster.UpdateNodeClaim(nodeClaim)
-	c.clusterCost.UpdateNodeClaim(ctx, nodeClaim)
+	if err := c.clusterCost.UpdateNodeClaim(ctx, nodeClaim); err != nil {
+		log.FromContext(ctx).Error(err, "failed to process nodeclaim for cost tracking")
```
ellistarn, line 77: Do we want to return the error all the way out to controller runtime so the loop gets backoffretried and the state gets fixed as quickly as possible? This is typically true for all control loops, barring a few exceptions.

ellistarn (approved):
Thanks for this. One more minor note about retry behavior. Feel free to merge if no changes from there.

/lgtm

pkg/state/cost/cost.go
```diff
@@ -213,42 +212,40 @@ func (cc *ClusterCost) UpdateNodeClaim(ctx context.Context, nodeClaim *v1.NodeCl
...
-		return
+		return nil
 	}
 
 	np := &v1.NodePool{}
 	if err := cc.client.Get(ctx, types.NamespacedName{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, np); err != nil {
-		log.FromContext(ctx).Error(serrors.Wrap(err, "nodepool", nodeClaim.Labels[v1.NodePoolLabelKey], "nodeclaim", klog.KObj(nodeClaim)), "failed to process nodeclaim for cost tracking")
 		failed = true
-		return
+		return serrors.Wrap(err, "nodepool", nodeClaim.Labels[v1.NodePoolLabelKey], "nodeclaim", klog.KObj(nodeClaim))
```
ellistarn, line 221: when did we start using klog to get namespaced names? I am more familiar with client.ObjectKeyFromObject or https://github.com/awslabs/operatorpkg/blob/main/object/object.go#L34.

pkg/controllers/state/informer/nodeclaim.go
ryan-mist: Yup, Ill put out a PR to change this. 

As an added benefit, a lot of the setup for unit tests will need to be changed to ensure the NodeClaim has a OwnerRef set. So, we'll stop seeing these logs in all the unit tests:
```
    logger.go:146: 2025-12-18T10:19:13.916-0800	ERROR	cost/cost.go:228	failed to get nodepool for nodeclaim	{"error": "nodepool not found for nodeclaim (nodeclaim=frillprism-125-lqnhpcbytl, nodepool=nodepool-3)"}
```

PR #2707, opened 2025-12-11
feat: Add prometheus metrics for nodepool disruption decisions performed, active disruptions

cnmcavoy:
Fixes N/A

**Description**

Adds two new prometheus metrics:

* `karpenter_nodepools_decisions`
* `karpenter_nodepools_active_disruptions`

`karpenter_nodepools_decisions` tracks the same actions as `karpenter_voluntary_disruptions_decisions_total`, but with an extra dimension for the nodepool. As a result, the total count of `karpenter_nodepools_decisions` may be greater than `karpenter_voluntary_disruptions_decisions_total` when 1 decision spans more than 1 nodepool. 

`karpenter_nodepools_active_disruptions` tracks the active number of terminating nodeclaims in each nodepool, which is useful to compare against the existing metric `karpenter_nodepools_allowed_disruptions`. 

This change allows us to track how disrupted individual nodepools are, and measure the consolidation in individual nodepools over time.

`make presubmit` and tested in our clusters.

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -271,6 +271,9 @@ func BuildDisruptionBudgetMapping(ctx context.Context, cluster *state.Cluster, c
 		NodePoolAllowedDisruptions.Set(float64(allowedDisruptions), map[string]string{
 			metrics.NodePoolLabel: nodePool.Name, metrics.ReasonLabel: string(reason),
 		})
+		NodePoolNodesConsumingBudgets.Set(float64(disrupting[nodePool.Name]), map[string]string{
+			metrics.NodePoolLabel: nodePool.Name, metrics.ReasonLabel: string(reason),
+		})
 		if numNodes[nodePool.Name] != 0 && allowedDisruptions == 0 {
 			recorder.Publish(disruptionevents.NodePoolBlockedForDisruptionReason(nodePool, reason))
 		}
--- a/pkg/controllers/disruption/metrics.go
+++ b/pkg/controllers/disruption/metrics.go
@@ -58,6 +58,16 @@ var (
 		},
 		[]string{decisionLabel, metrics.ReasonLabel, ConsolidationTypeLabel},
 	)
+	NodepoolDecisionsPerformed = opmetrics.NewPrometheusCounter(
+		crmetrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: voluntaryDisruptionSubsystem,
+			Name:      "decisions_by_nodepool_total",
+			Help:      "Number of disruption decisions performed by nodepool. Labeled by nodepool name, disruption decision, reason, and consolidation type.",
+		},
+		[]string{metrics.NodePoolLabel, decisionLabel, metrics.ReasonLabel, ConsolidationTypeLabel},
+	)
 	EligibleNodes = opmetrics.NewPrometheusGauge(
 		crmetrics.Registry,
 		prometheus.GaugeOpts{
@@ -95,6 +105,16 @@ var (
 			Subsystem: metrics.NodePoolSubsystem,
 			Name:      "allowed_disruptions",
 			Help:      "The number of nodes for a given NodePool that can be concurrently disrupting at a point in time. Labeled by NodePool. Note that allowed disruptions can change very rapidly, as new nodes may be created and others may be deleted at any point.",
+		},
+		[]string{metrics.NodePoolLabel, metrics.ReasonLabel},
+	)
+	NodePoolNodesConsumingBudgets = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodePoolSubsystem,
+			Name:      "nodes_consuming_budgets",
+			Help:      "The number of nodes consuming the budget of a nodepool at a point in time. Labeled by NodePool.",
 		},
 		[]string{metrics.NodePoolLabel, metrics.ReasonLabel},
 	)
--- a/pkg/controllers/disruption/queue.go
+++ b/pkg/controllers/disruption/queue.go
@@ -360,6 +360,17 @@ func (q *Queue) StartCommand(ctx context.Context, cmd *Command) error {
 	q.Unlock()
 
 	// An action is only performed and pods/nodes are only disrupted after a successful add to the queue
+	nodePools := lo.Uniq(lo.Map(cmd.Candidates, func(c *Candidate, _ int) string {
+		return c.NodePool.Name
+	}))
+	for _, nodePool := range nodePools {
+		NodepoolDecisionsPerformed.Inc(map[string]string{
+			metrics.NodePoolLabel:  nodePool,
+			decisionLabel:          string(cmd.Decision()),
+			metrics.ReasonLabel:    strings.ToLower(string(cmd.Reason())),
+			ConsolidationTypeLabel: cmd.ConsolidationType(),
+		})
+	}
 	DecisionsPerformedTotal.Inc(map[string]string{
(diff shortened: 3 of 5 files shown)


pkg/controllers/disruption/queue.go
```diff
@@ -360,6 +360,17 @@ func (q *Queue) StartCommand(ctx context.Context, cmd *Command) error {
...
+	}))
+	for _, nodePool := range nodePools {
+		NodepoolDecisionsPerformed.Inc(map[string]string{
+			metrics.NodePoolLabel:  nodePool,
+			decisionLabel:          string(cmd.Decision()),
+			metrics.ReasonLabel:    strings.ToLower(string(cmd.Reason())),
+			ConsolidationTypeLabel: cmd.ConsolidationType(),
+		})
+	}
 	DecisionsPerformedTotal.Inc(map[string]string{
```
DerekFrank, line 374: Should we just add a label to this metric instead of adding a new metric?

pkg/controllers/disruption/helpers.go
```diff
@@ -271,6 +271,9 @@ func BuildDisruptionBudgetMapping(ctx context.Context, cluster *state.Cluster, c
 		NodePoolAllowedDisruptions.Set(float64(allowedDisruptions), map[string]string{
 			metrics.NodePoolLabel: nodePool.Name, metrics.ReasonLabel: string(reason),
 		})
+		NodePoolActiveDisruptions.Set(float64(disrupting[nodePool.Name]), map[string]string{
```
DerekFrank: This does technically include not ready nodes as well as ones that are disrupting. Does this number have use outside of knowing the end result of:

`lo.Max([]int{allowedDisruptions - disrupting[nodePool.Name], 0})`?

pkg/controllers/disruption/queue.go
cnmcavoy: I'm not sure I follow - are you thinking of using a label on `DecisionsPerformedTotal`? What about when a command includes multiple nodepools? The signature of the tags is `map[string]string`, it doesn't take an array of tags. Do you want to concat the nodepool names together? I don't think there's a clean way to add this to the metric, but I'm open to ideas.

pkg/controllers/disruption/helpers.go
cnmcavoy: That is intentional, the goal of this PR is to expose Karpenter's state to help operators understand it's decision making process.

pkg/controllers/disruption/helpers.go
jmdeal, line 1: > karpenter_nodepools_active_disruptions tracks the active number of terminating nodeclaims in each nodepool, which is useful to compare against the existing metric karpenter_nodepools_allowed_disruptions.

Could you give an example of how you might compare these metrics? I can come up with some use-cases for the metric myself, I just want to make sure we're working backwards from the problem when designing these metrics. In it's current state though, I'd consider changing the name. By including `disrupting` in the name it implies that this only includes nodes that are actually being disrupted by Karpenter. It's really a superset of that because it also includes nodes that went NotReady for other reasons (e.g. kubelet OOMed). IIUC what we're trying to capture is the number of nodes that our counting against our budget, so something like `karpenter_nodepools_nodes_consuming_budgets` might be more appropriate? I'm not in love with that name FWIW, we should workshop it, but that's the intent we want to convey.

pkg/controllers/disruption/metrics.go
```diff
@@ -58,6 +58,16 @@ var (
 		},
 		[]string{decisionLabel, metrics.ReasonLabel, ConsolidationTypeLabel},
 	)
+	NodepoolDecisionsPerformed = opmetrics.NewPrometheusCounter(
+		crmetrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodePoolSubsystem,
```
jmdeal: Should this be in the nodepool subsystem or the voluntary disruptions subsystem? I'd lean towards the latter since it's a parallel metric to `karpenter_voluntary_disruption_decisions_total`. We could name it something along the lines of `karpenter_voluntary_disruption_decisions_total_by_nodepool_total` as well which is better for discoverability IMO.

GnatorX: Is it possible to add nodepool label to the existing metric and add a new "nodepool" label value that tracks deduplicated decision across all nodepool (matching previous metric)

cnmcavoy: @jmdeal Sure, I only put it in the nodepool metrics because `voluntary_disruption` metrics didn't have a nodepool dimension on them. I don't have a strong opinion on which subsystem the metric is under.

@GnatorX already asked and answered: https://github.com/kubernetes-sigs/karpenter/pull/2707#discussion_r2651960469

pkg/controllers/disruption/helpers.go
cnmcavoy: We're using it so that we can show teams how many nodes can be disrupted at any given point in time in their nodepool:

image

I'm not picky about the metric name, so I asked an AI to bikeshed the name a bit, it came up with these:

- `karpenter_nodes_consuming_disruption_budgets`
- `karpenter_nodes_in_disruption_budget`
- `karpenter_nodes_under_disruption_budget`

`karpenter_nodes_in_disruption_budget` might be the shortest and clearest? Let me know if you prefer `karpenter_nodepools_nodes_consuming_budgets` or one of the others.

pkg/controllers/disruption/queue.go
GnatorX: I guess (at least what I am thinking) is adding `metrics.NodePoolLabel:  nodePool` into the map of decisionsperformed. Is that an issue? Sorry I am just looking at the code without testing it

cnmcavoy: Yeah that is a breaking change for the metric. The metric currently is 1:1 with "command" decisions karpenter makes. A "command" can span many nodepools, it's 1 to many. So you would need the metric to increment for each nodepool, not for each decision. Which is what the new metric I implemented does.

Unless I'm misunderstanding - give me an example of how you would implement this?

pkg/controllers/disruption/helpers.go
jmdeal: We're definitely bikeshedding, but I'd err on the side of verbosity. My preference would be `karpenter_nodepools_nodes_consuming_disruption_budgets`. I'd want to keep the family consistent with `karpenter_nodepools_active_disruptions` and I prefer consuming to in because "in" is ambiguous to me - I could see it being either nodes that are consuming the budget or nodes that are within the bounds of the budget (available to disrupt). If that sounds good to you, let's go forward with that.

pkg/controllers/disruption/metrics.go
jmdeal: Ok, are we agreed on `karpenter_voluntary_disruption_decisions_by_nodepool_total` then?

pkg/controllers/disruption/queue.go
GnatorX: Got it ok that's what i was wondering. That seems fine, although I would like to eventually converge to a single metric. Fine with the current implementation

pkg/controllers/disruption/metrics.go
GnatorX: I am good with it

pkg/controllers/disruption/metrics.go
```diff
@@ -98,6 +108,16 @@ var (
 		},
 		[]string{metrics.NodePoolLabel, metrics.ReasonLabel},
 	)
+	NodePoolActiveDisruptions = opmetrics.NewPrometheusGauge(
```
jmdeal: Looks like this metric still needs to be updated per this [discussion](https://github.com/kubernetes-sigs/karpenter/pull/2707/changes#r2710193699), but looks good to me otherwise.

cnmcavoy: Whoops, I missed this. Fixed.

PR #2709, opened 2025-12-11
chore: add minDrainTime

ryan-mist:
Fixes #N/A 

**Description**
* add `minDrainTime` to termination flow to ensure pods scheduled immediately after the taint is applied will still be drained

* `make presubmit`
* manual testing to ensure minDrainTime has passed before node termination

### DIFF
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -212,6 +212,8 @@ func (c *Controller) awaitDrain(
 	}
 
 	// If the nodeclaim exists, check if minDrainTime has elapsed. If it hasn't we should requeue.
+	// This check helps to ensure that we drain pods scheduled to the Node immediately after we taint it, which
+	// can occur when the scheduler has not seen the taint yet.
 	if nodeClaim != nil {
 		cond := nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained)
 		if cond == nil || (cond.IsUnknown() && c.clock.Since(cond.LastTransitionTime.Time) < MinDrainTime) {


pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -52,6 +53,7 @@ func NewTerminator(clk clock.Clock, kubeClient client.Client, eq *Queue, recorde
 }
 
 // Taint idempotently adds a given taint to a node with a NodeClaim
+// If the taint is karpenter.sh/disruption:NoSchedule, we add the karpenter.sh/disrupted-taint-time annotation.
```
ryan-mist: This feels a bit off to me that the behavior would change depending on the taint. This function is only used for adding the disruption taint, so I almost wonder if changing this function to `TaintForDisruption` or something along those lines would be more appropriate.

pkg/controllers/node/termination/controller.go
```diff
@@ -190,25 +191,34 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
-		if nodeClaim != nil {
-			nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining")
-		}
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+
+	// If the nodeclaim exists, check if minDrainTime has elapsed. If it hasn't we should requeue.
+	if nodeClaim != nil {
+		cond := nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained)
+		if cond == nil || (cond.IsUnknown() && c.clock.Since(cond.LastTransitionTime.Time) < MinDrainTime) {
```
jmdeal, line 219: Sanity check:  there's no scenario where this would be nil, right? I don't see any issue with keeping the check, just making sure.

pkg/controllers/node/termination/controller.go
```diff
@@ -190,25 +191,34 @@ func (c *Controller) finalize(ctx context.Context, node *corev1.Node) (reconcile
...
 			return reconcile.Result{}, fmt.Errorf("draining node, %w", err)
 		}
 		c.recorder.Publish(terminatorevents.NodeFailedToDrain(node, err))
-		if nodeClaim != nil {
-			nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining")
-		}
 		return reconcile.Result{RequeueAfter: 1 * time.Second}, nil
 	}
+
+	// If the nodeclaim exists, check if minDrainTime has elapsed. If it hasn't we should requeue.
```
jmdeal, line 214: I'd add some details here about why we have the minDrainTime

pkg/controllers/node/termination/controller.go
ryan-mist: There could be if the providerID on the NodeClaim wasn't updated yet or theres multiple NodeClaims with the same providerID. Not sure how the latter will happen (unless someone manually does this) but I know we've seen the former before.

The note is here - https://github.com/kubernetes-sigs/karpenter/blob/8fe7bfee8c1eb5079377a2b5606c274e4fa5883e/pkg/controllers/node/termination/controller.go#L99

PR #2717, opened 2025-12-17
test: speed up tests by using fake clock for sleeps

ellistarn:
## Problem
Tests were slow due to real `time.Sleep(time.Second)` calls in controller code that get executed during test reconciliation.

## Solution
Use `clock.Sleep()` instead of `time.Sleep()` in lifecycle and expiration controllers, allowing tests to skip real sleeps via fake clock.

## Results
| Package | Before | After | Speedup |
|---------|--------|-------|---------|
| nodeclaim/lifecycle | 175s | 60s | 66% faster |


tzneal (approved):
Nice!

/lgtm

PR #2724, opened 2025-12-19
feat: move marking for disruption before validation

ryan-mist:
Fixes #N/A 

**Description**
We see the following race condition with todays consolidation workflow:
* Karpenter taints a consolidatable node with `karpenter.sh/disrupted:NoSchedule`
* Before kube-scheduler's informer caches are up-to-date with the taint on the node, it binds a pod with `do-not-disrupt` annotation to the node.

This PR moves tainting before validation to catch this race condition.

This PR shifts the main intention of validation from detecting cluster churn to catching this race condition. The idea is that `consolidateAfter` works to similar affects as the original intention of validation since the Node cannot be consolidated for a period of time after a pod is removed (or added). For this reason, we also drop validation of re-running the scheduling simulation.

Old Disruption Order:
```
Get Candidates for disruption
If Validating (i.e. {singleNode, mulitNode, emptiness}Consolidation)
     Wait for 15 seconds
     Validate Candidates
     Revalidate Command [if not emptinessConsolidation]
Mark Candidates for disruption (add taint to Node and add condition to NodeClaim)
Continue ...
```

New Disruption Order:
```
Get Candidates for disruption
Mark Candidates for disruption (add taint to Node and add condition to NodeClaim)
If Validating (i.e. {singleNode, mulitNode, emptiness}Consolidation)
     Wait for 5 seconds
     Validate Candidates
Continue ...
```

* `make presubmit`
* manual testing


jonathan-innis:
I'm curious if there's a larger proposal about this since this seems like a fairly significant change -- won't this cause a bunch of overlaunching and sort of nullifies the validation behavior of consolidation (since no more pods will schedule to the node if it's already marked with a NoSchedule taint). 

I require there being conversation about using a PreferNoSchedule for this here but NoSchedule feels like it removes the ability to do proper checking that the consolidation should actually be performed. 

FWIW, I feel like if we are making this change, we should just remove the time-based validation altogether since it will be doing effectively nothing at this point.

jonathan-innis:
Can you link the race condition you are referring to from an issue?

ryan-mist:
It shifts the main intention of validation from detecting cluster churn to catching this race condition. It does still "catch" cluster churn between our decision and taint application, but this is a smaller window than before. The idea is that `consolidateAfter` works to similar affects as the original intention of validation since the Node cannot be conolidated for `n` time after a pod is removed (or added). This makes this original validation extra and not a strictly necessary check.

> FWIW, I feel like if we are making this change, we should just remove the time-based validation altogether since it will be doing effectively nothing at this point.

We may not need to revalidate the command, since the window for pods to be scheduled on the node from the original simulation to tainting the Node is small. We do need the time-based validation to check that no pods with the `do-not-disrupt` annotation have scheduled to the Node after we apply the taint.

PR #2726, opened 2025-12-19
chore: add min drain time

ryan-mist:
Fixes #N/A 

**Description**
* add option `MIN_NODE_DRAIN_TIME` to allow configuration of minimum time we spend draining the Node during termination
   * the check is done by adding an annotation on the Node to indicate when draining started

* manual testing to ensure annotation is added to the Node and minDrainTime is respected
```
Annotations:        alpha.kubernetes.io/provided-node-ip: 192.168.123.182
...
                    karpenter.sh/disrupted-taint-time: 2025-12-19T13:28:10-08:00
...
CreationTimestamp:  Fri, 19 Dec 2025 10:57:55 -0800
Taints:             karpenter.sh/disrupted:NoSchedule
```


pkg/operator/options/options.go
```diff
@@ -111,6 +112,7 @@ func (o *Options) AddFlags(fs *FlagSet) {
 	fs.StringVar(&o.LogErrorOutputPaths, "log-error-output-paths", env.WithDefaultString("LOG_ERROR_OUTPUT_PATHS", "stderr"), "Optional comma separated paths for logging error output")
 	fs.DurationVar(&o.BatchMaxDuration, "batch-max-duration", env.WithDefaultDuration("BATCH_MAX_DURATION", 10*time.Second), "The maximum length of a batch window. The longer this is, the more pods we can consider for provisioning at one time which usually results in fewer but larger nodes.")
 	fs.DurationVar(&o.BatchIdleDuration, "batch-idle-duration", env.WithDefaultDuration("BATCH_IDLE_DURATION", time.Second), "The maximum amount of time with no new pending pods that if exceeded ends the current batching window. If pods arrive faster than this time, the batching window will be extended up to the maxDuration. If they arrive slower, the pods will be batched separately.")
+	fs.DurationVar(&o.MinNodeDrainTime, "min-node-drain-time", env.WithDefaultDuration("MIN_NODE_DRAIN_TIME", 0*time.Second), "Minimum time to wait after tainting a node before draining pods to ensure the scheduler sees the taint")
```
jmdeal: I'd default this to 5 seconds

ryan-mist: done

pkg/controllers/node/termination/terminator/terminator.go
```diff
@@ -51,8 +52,10 @@ func NewTerminator(clk clock.Clock, kubeClient client.Client, eq *Queue, recorde
 	}
 }
 
-// Taint idempotently adds a given taint to a node with a NodeClaim
-func (t *Terminator) Taint(ctx context.Context, node *corev1.Node, taint corev1.Taint) error {
+// TaintForDisruption idempotently adds the karpenter.sh/disruption:NoSchedule taint
+// and the karpenter.sh/disrupted-taint-time annotation to a node with a NodeClaim
+func (t *Terminator) TaintForDisruption(ctx context.Context, node *corev1.Node) error {
```
moko-poi, line 57: You are changing your Taint() method to TaintForDisruption(), but the Taint() method is still being used elsewhere.
ref: pretty.go

Wouldn't it be better to keep the Taint() method and add TaintForDisruption() as a separate method?

moko-poi (commented):
Implement unit tests.

jmdeal:
Going to close this out for the time being - we're actively investing in changes to consolidation logic that may conflict and have found that the change made in the termination controller is good enough to solve the burning issues we were seeing due to cache inconsistency between Karpenter and the scheduler. We can revisit this based on the outcome of some of those ongoing discussions around consolidation.

PR #2727, opened 2025-12-19
chore: add pod readiness dimension to pod state metrics

DerekFrank:
Fixes #N/A 

**Description**

Add a dimension to podState about pod readiness.

Adds a metric to count the number of pods deleted.

`make presubmit`

### DIFF
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -54,6 +54,7 @@ const (
 	podHostInstanceType = "instance_type"
 	podPhase            = "phase"
 	podScheduled        = "scheduled"
+	podReady            = "ready"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/controllers/metrics/pod/controller.go
```diff
@@ -166,6 +166,17 @@ var (
 		},
 		[]string{podName, podNamespace},
 	)
+	// Stage: alpha
+	PodPendingPercentage = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.PodSubsystem,
+			Name:      "binding_percent",
```
jonathan-innis:
```suggestion
			Name:      "bound_percent",
```

Is it binding_percent or bound_percent since these pods are already bound to nodes?

DerekFrank: Fair point!

pkg/controllers/metrics/pod/controller.go
```diff
@@ -166,6 +167,17 @@ var (
 		},
 		[]string{podName, podNamespace},
 	)
+	// Stage: alpha
+	PodDeletionsCount = opmetrics.NewPrometheusCounter(
```
ellistarn: Do we want to try to delineate deletion from eviction? API server has
```
apiserver_request_total{code="201",component="apiserver",dry_run="",group="",resource="pods",scope="resource",subresource="eviction",verb="POST",version="v1"} 7
```

ellistarn: Though capturing the (successful) evictions triggered by Karpenter in the termination controller is probably enough.

DerekFrank: It makes sense to drop this and rely on that: https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/node/termination/terminator/eviction.go#L215

PR #2728, opened 2025-12-20
fix: update disruption metrics even without candidates

### LINKED ISSUE #2344: karpenter_nodepools_allowed_disruptions metrics is mismatching actual budget schedule or schedule is not working
### Description

**Observed Behavior**:
I have:
```
spec:
  disruption:
    budgets:
    - duration: 12h00m0s
      nodes: "0"
      reasons:
      - Drifted
      - Empty
      - Underutilized
      schedule: 0 10 * * mon-fri
    - nodes: "2"
    consolidateAfter: 30s
```
 but if I look graphs or query metrics endpoint:

Image
metrics endpoint returns:
```
curl localhost:8080/metrics -s | grep karpenter_nodepools_allowed_disruptions | grep opensearch
karpenter_nodepools_allowed_disruptions{nodepool="opensearch-amd64-on-demand",reason="Empty"} 2
karpenter_nodepools_allowed_disruptions{nodepool="opensearch-amd64-on-demand",reason="Underutilized"} 0
```

In my understanding it should be almost the same  and forbidding disruption according to schedule (so showing 0 for this period of time) but graph show budget is different for a long time.
There is no consolidation happening during this query.
  

**Expected Behavior**:
Metric should reflect actual schedule, but the problem is that I' not sure that schedule even works as it should. 

There is another example for another nodepool:

Image
```
spec:
  disruption:
    budgets:
    - nodes: "3"
      reasons:
      - Empty
      - Underutilized
    - nodes: "2"
      reasons:
      - Drifted
    - duration: 12h0m0s
      nodes: "0"
      reasons:
      - Drifted
      schedule: 0 12 * * mon-fri
    consolidateAfter: 30s

```
Graph shows that only UnderUtilized is active all the time but not Empty
also Drifted doesn't consistent it was allowing 2 nodes but after weekend
 it reports 0.

```
curl localhost:8080/metrics -s | grep karpenter_nodepools_allowed_disruptions | grep bottle    
karpenter_nodepools_allowed_disruptions{nodepool="bottlerocket-nodepool",reason="Empty"} 3
karpenter_nodepools_allowed_disruptions{nodepool="bottlerocket-nodepool",reason="Underutilized"} 3
```

no "Drifted" is reported at all, which is fine but I'd like to have 0 reported.

**Versions**:
- Chart Version: 1.3.3

```
#kubectl version                                                                                                                               
Client Version: v1.33.2
Kustomize Version: v5.6.0
Server Version: v1.30.13-eks-5d4a308
WARNING: version difference between client (1.33) and server (1.30) exceeds the supported minor version skew of +/-1
Kubecolor Version: v0.5.0

```

moko-poi:
fix: update disruption metrics even without candidates

Fixes #2344

**Description**

The `karpenter_nodepools_allowed_disruptions` metric was only being updated when disruption candidates existed. This caused the metric to report stale values in several scenarios:

- When no nodes are eligible for disruption
- When schedule-based budgets change (e.g., entering/exiting disruption windows)
- When NodePools are deleted but their metrics remain

**Root Cause:**
The `BuildDisruptionBudgetMapping()` function was called after checking if candidates exist. When `len(candidates) == 0`, the function returned early without updating metrics.

**Solution:**
Move `BuildDisruptionBudgetMapping()` before the candidate check. This ensures metrics are updated on every reconcile loop (every 10 seconds), regardless of whether disruption candidates exist.

**Impact:**
- Metrics now accurately reflect schedule-based budget restrictions in real-time
- Deleted NodePools naturally stop being reported (only active pools are updated)
- No breaking changes - existing behavior with candidates is preserved
- Minimal performance impact - same function calls, just reordered

- Code compiles successfully: `go build ./pkg/controllers/disruption/...`
- No new test failures - existing tests pass as candidate behavior is unchanged
- Manual verification: metric update logic is triggered for all disruption reasons during reconcile, independent of candidate existence
- Logical validation: `BuildDisruptionBudgetMapping()` is side-effect free except for metrics/events, so moving it earlier is safe

**Example Scenario Fixed:**
```yaml
spec:
  disruption:
    budgets:
    - duration: 12h00m0s
      nodes: "0"
      schedule: 0 10 * * mon-fri  # Block disruption 10:00-22:00 weekdays
    - nodes: "2"
```

Before: Metric shows `2` even during blocked window  
After: Metric correctly shows `0` during 10:00-22:00 and `2` outside that window

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -192,13 +192,16 @@ func (c *Controller) disrupt(ctx context.Context, disruption Method) (bool, erro
 		metrics.ReasonLabel: strings.ToLower(string(disruption.Reason())),
 	})
 
-	// If there are no candidates, move to the next disruption
-	if len(candidates) == 0 {
-		return false, nil
-	}
+	// Always build disruption budget mapping to ensure metrics are up-to-date,
+	// even when there are no candidates for disruption
 	disruptionBudgetMapping, err := BuildDisruptionBudgetMapping(ctx, c.cluster, c.clock, c.kubeClient, c.cloudProvider, c.recorder, disruption.Reason())
 	if err != nil {
 		return false, fmt.Errorf("building disruption budgets, %w", err)
+	}
+
+	// If there are no candidates, move to the next disruption
+	if len(candidates) == 0 {
+		return false, nil
 	}
 	// Determine the disruption action
 	cmds, err := disruption.ComputeCommands(ctx, disruptionBudgetMapping, candidates...)
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -111,7 +111,7 @@ func NewControllers(
 	if !options.FromContext(ctx).DisableClusterStateObservability {
 		controllers = append(controllers,
 			metricspod.NewController(kubeClient, cluster),
-			metricsnodepool.NewController(kubeClient, cloudProvider, clusterCost),
+			metricsnodepool.NewController(kubeClient, cloudProvider, clusterCost, cluster, clock),
 			metricsnode.NewController(cluster),
 			status.NewController[*v1.NodeClaim](
 				kubeClient,
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -39,7 +39,6 @@ import (
 	"sigs.k8s.io/karpenter/pkg/controllers/provisioning/scheduling"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/events"
-	"sigs.k8s.io/karpenter/pkg/metrics"
 	operatorlogging "sigs.k8s.io/karpenter/pkg/operator/logging"
 	nodeutils "sigs.k8s.io/karpenter/pkg/utils/node"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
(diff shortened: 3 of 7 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -192,14 +192,17 @@ func (c *Controller) disrupt(ctx context.Context, disruption Method) (bool, erro
 		metrics.ReasonLabel: strings.ToLower(string(disruption.Reason())),
 	})
 
-	// If there are no candidates, move to the next disruption
-	if len(candidates) == 0 {
-		return false, nil
-	}
+	// Always build disruption budget mapping to ensure metrics are up-to-date,
```
DerekFrank, line 195: I'm not sure this is the right refactor here. Because the metrics are entirely based on cluster state, I think I would prefer if we move the metrics outside of BuildDisruptionBudgetMapping and into the nodepool or node metrics controller. That ensures the metrics will always be emitted no matter the changes to consolidation or disruption.

moko-poi: @DerekFrank 
You're absolutely right. I've implemented the refactor following your suggestion. I moved the metrics emission out of BuildDisruptionBudgetMapping and into the nodepool metrics controller, which ensures the metrics are always updated regardless of whether disruption candidates exist or consolidation logic changes.

Specifically, I added cluster state and clock dependencies to the nodepool metrics controller and implemented an updateDisruptionBudgetMetrics method. This method is called for each NodePool during the Reconcile loop, counts nodes directly from cluster state, evaluates schedule-based budgets using the clock, and sets the metrics accordingly. Since the controller requeues every 5 minutes, the metrics are updated periodically independent of the disruption controller's execution.

This change makes the metrics logic completely independent from the disruption controller logic, making it more robust. Thank you for the architectural improvement suggestion.

[b8227fb](https://github.com/kubernetes-sigs/karpenter/pull/2728/commits/b8227fb572d7ed64edceebdbcbb24519f29189c4)

PR #2729, opened 2025-12-20
fix: prevent DaemonSet pods from circling in Failed/Pending during disruption

### LINKED ISSUE #2009: DaemonSet pods circle in FAILED state due to rescheduling to node which is terminated
### Description

**Observed Behavior**:
DaemonSet(DS) pods circle in `Failed` -> `Pending` state on node which is consolidate and terminated later. 
DS pods (like aws-node, kube-proxy) usually tolerates all Taints. Karpenter put Taint `karpenter.sh/disrupted:NoSchedule` on node, now DS pods get into `Failed` state and new pods in Pending state appear, extra `not-ready` taints are also added so newly scheduled pods are rejected. Repeats until node is terminated and removed from cluster. 
When large amount of nodes is rotated in same time, our alerting is fired with false positive alerts about multiple pods swapping between Failed/Pending state. 

See attached log files and screenshot.

Image
[eks-controller-logs.csv](https://github.com/user-attachments/files/18871177/eks-controller-logs.csv)
[karpenter-controller.log](https://github.com/user-attachments/files/18871187/karpenter-controller.log)

We observer same issue on older versions, but not on versions 0.37 and below. 

**Expected Behavior**:
We do not observe flipping state of DaemonSet pod between Failure/Pending during node consolidation/termination. 

Simply consolidate some node controlled by Karpenter and watch k8s cluster Events or EKS controller logs. You will see karpenter remove all pods from node expect DS + put there `karpenter` taint, DS pods get into Failed state -> transition to Pending state, `not-ready` taints are added to node, after ~ 1 minute node is terminated. 

**Versions**:
- Chart Version: 1.2.1
- Kubernetes Version:
```
Client Version: v1.30.0
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.30.8-eks-2d5f260
````

moko-poi:
Fixes #2009

**Description**

This PR fixes an issue where DaemonSet pods (such as `aws-node` and `kube-proxy`) enter a Failed/Pending cycle during node disruption (consolidation/termination).

**Root Cause:**
When Karpenter marks a node for disruption, it applies the `karpenter.sh/disrupted:NoSchedule` taint to prevent new pods from scheduling. However, DaemonSet pods were being evicted during this process. When the DaemonSet controller attempts to recreate these pods, the NoSchedule taint prevents them from being scheduled back to the node, causing them to remain in Pending state. This creates a continuous cycle of pod failures and rescheduling attempts until the node is fully terminated.

**Solution:**
Modified the `IsEvictable()` and `IsDrainable()` functions in `pkg/utils/pod/scheduling.go` to explicitly exclude DaemonSet pods from eviction during node disruption. This approach aligns with `kubectl drain` behavior, where DaemonSet pods remain running until the node is actually terminated, allowing the DaemonSet controller to naturally manage pod recreation on other available nodes.

**Changes:**
- Added `!IsOwnedByDaemonSet(pod)` check to `IsEvictable()` function
- Added `!IsOwnedByDaemonSet(pod)` check to `IsDrainable()` function  
- Added test case to verify DaemonSet pods are not evicted during disruption
- Updated existing test case for DaemonSet pods with PDBs to reflect the new behavior

1. **Unit Tests**: Added comprehensive test cases in both disruption and termination controllers:
   - `should not evict daemonset pods during node disruption` - Verifies DaemonSet pods remain running when disruption taint is applied
   - `should consider candidates with only daemonset pods` - Verifies nodes with only DaemonSet pods can be disrupted
   - Updated `should consider candidates that have fully blocking PDBs on daemonset pods` - Verifies PDBs don't block disruption when only DaemonSet pods are present

2. **Test Execution**: All existing tests pass with these changes:
   ```
   make test FOCUS="daemonset"
   ```
   - 3 DaemonSet-related tests: PASSED
   - All pkg tests: PASSED with no failures

3. **Race Detection**: Tests executed with `-race` flag to ensure no data races introduced

### DIFF
--- a/pkg/controllers/disruption/suite_test.go
+++ b/pkg/controllers/disruption/suite_test.go
@@ -1386,7 +1386,7 @@ var _ = Describe("Candidate Filtering", func() {
 		Expect(err.Error()).To(Equal(fmt.Sprintf(`pdb prevents pod evictions (PodDisruptionBudget=[%s])`, client.ObjectKeyFromObject(budget))))
 		Expect(recorder.DetectedEvent(fmt.Sprintf(`Pdb prevents pod evictions (PodDisruptionBudget=[%s])`, client.ObjectKeyFromObject(budget)))).To(BeTrue())
 	})
-	It("should not consider candidates that have fully blocking PDBs on daemonset pods", func() {
+	It("should consider candidates that have fully blocking PDBs on daemonset pods since daemonset pods are not evicted", func() {
 		daemonSet := test.DaemonSet()
 		nodeClaim, node := test.NodeClaimAndNode(v1.NodeClaim{
 			ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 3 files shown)


jmdeal:
> When the DaemonSet controller attempts to recreate these pods, the NoSchedule taint prevents them from being scheduled back to the node, causing them to remain in Pending state.

If I understand correctly this is the core issue, right? The issue isn't that daemonsets are being evicted, it's that some daemonset pods are being recreated and entering a pending / failed state. What's not clear to me is why they're being recreated - the daemonset controller should only create a pod for a node if it tolerates the taints. If it does tolerate the taint, Karpenter shouldn't have disrupted the pod in the first place due to the existing `IsEvictable` check.

I think there are probably use-cases where we want to drain daemonsets. Some daemonsets perform resource cleanup which may not be possible once the node is terminating. An example that comes to mind is the EBS CSI driver which cleans up VolumeAttachment objects during termination. For this reason I don't think we'd want to exclude daemonsets from the drain process altogether, but we should identify the root cause for these pods being recreated.

moko-poi:
Closing this PR based on the review feedback from @jmdeal and further investigation.

After reproducing the issue with tests on the latest `main` branch, I found that the root cause is different from what this PR addresses:

**Scenario A** (DaemonSet pods that do NOT tolerate `karpenter.sh/disrupted:NoSchedule`):
- Karpenter does evict these pods during drain (in priority group 2, after non-daemon pods)
- However, as @jmdeal pointed out, the DaemonSet controller should not recreate pods on nodes with taints the pod template doesn't tolerate, so this path is unlikely to cause the reported cycling behavior

**Scenario B** (DaemonSet pods that tolerate ALL taints via `operator: Exists`, e.g. `aws-node`, `kube-proxy`):
- Karpenter correctly does NOT evict these pods (`ToleratesDisruptedNoScheduleTaint` returns true → `IsEvictable` returns false)
- The Failed/Pending cycle occurs because: instance termination → kubelet kills DS pod → pod enters Failed → DS controller recreates (pod tolerates all taints, so the dying node is still eligible) → new pod schedules to dying node → fails again → cycle repeats until the node object is removed from the API server
- **This PR's fix (`!IsOwnedByDaemonSet` in `IsEvictable`/`IsDrainable`) has no effect on this scenario** since Karpenter was never evicting these pods in the first place

The actual root cause for the issue reporter's case (aws-node, kube-proxy) is the Kubernetes DaemonSet controller recreating pods on a dying node that still exists in the API server, not Karpenter's eviction behavior. I'll leave a detailed comment on the issue with these findings.

PR #2735, opened 2025-12-22
feat: add pod resource requests and count metrics

DerekFrank:
Fixes #N/A 

**Description**

Separating this out from https://github.com/kubernetes-sigs/karpenter/pull/2626/ for ease of review.

make presubmit

### DIFF
--- a/pkg/controllers/metrics/cluster/controller.go
+++ b/pkg/controllers/metrics/cluster/controller.go
@@ -34,7 +34,10 @@ import (
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	"sigs.k8s.io/karpenter/pkg/state/cost"
+	"sigs.k8s.io/karpenter/pkg/state/podresources"
 )
+
+const resourceType = "resource_type"
 
 // These are alpha metrics, they may not stay. Do not rely on them.
 var (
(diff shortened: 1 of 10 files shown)


pkg/controllers/state/informer/pod.go
```diff
@@ -65,9 +68,11 @@ func (c *PodController) Reconcile(ctx context.Context, req reconcile.Request) (r
 		if errors.IsNotFound(err) {
 			// notify cluster state of the node deletion
 			c.cluster.DeletePod(req.NamespacedName)
+			c.podResources.DeletePod(pod)
```
DerekFrank, line 71: Moving @ellistarn's comment over:

> Would this be better encapsulated by cluster state?

We could write this into cluster.go, the reason it was made a separate object was for mutex contention purposes. We've been trying to separate cluster state out into different data stores where it make sense. I think this is one of those cases

ellistarn: Cool. FWIW I really meant "state" controllers in general. I see them as a bunch of efficient indexes, and strongly agree with factoring them into multiple modules.

pkg/controllers/metrics/cluster/controller.go
```diff
@@ -48,19 +51,43 @@ var (
...
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.PodSubsystem,
+			Name:      "resources_total",
+			Help:      "Total resource requests of all pods.",
+		},
+		[]string{resourceType},
+	)
+	// Stage: alpha
+	PodCount = opmetrics.NewPrometheusGauge(
```
ellistarn, line 66: I think we do want to be daemonset aware. WDYT about a label that includes the first ownerreference gvk?

pkg/controllers/state/informer/pod.go
ellistarn: But, we could optionally do all of that delegation in the cluster object -- might be able to help us coordinate mutexes. Informers are also a reasonable place to do this. I don't have a mental model of this entire state gomod right now, but I suspect we may need to refactor at some point.

DerekFrank:
> Why not use https://github.com/ellistarn/karpenter/blob/72387e317dd1868733631e34f7919611f08ce77f/pkg/controllers/metrics/node/controller.go#L74 ?

Fair point, I must have missed that when I was looking for this metric. That metric does only include bound pods, but I think that is fine for now.

For https://github.com/kubernetes-sigs/karpenter/pull/2626, I can surface the information through cluster.go to the decision tracker, but the node controller will need to still recompute the information to get it to have the correct labels.

PR #2739, opened 2025-12-23
feat: add configurable node repair unhealthy threshold

### LINKED ISSUE #2134: Auto Repair: Make % of unhealthy nodes configurable
### Description

We have node auto repair activated. The node pool has a very dynamic size but usually is around 3 large nodes.

When a single node becomes unhealthy karpenter won't auto repair it as 1 node is 33% of the node pool:

```
  Warning  NodeRepairBlocked        88s (x19 over 142m)  karpenter        more then 20% nodes are unhealthy in the nodepool
```

It would be great to either have this % configurable, have an absolute value or even be able to disable this safeguard.

clark42:
## What does this PR do?

Adds a new `nodeRepairUnhealthyThreshold` field to the `NodePool.spec.disruption` configuration, allowing operators to configure the maximum percentage or count of unhealthy nodes before node auto repair is blocked.

## Why is this needed?

The current hardcoded 20% threshold is too restrictive for small NodePools (3-5 nodes). When 2 out of 4 nodes become unhealthy (50%), node auto repair is blocked with the message "more than 20% nodes are unhealthy in the nodepool", leaving the cluster in a degraded state requiring manual intervention.

### Real-world use case

In our EKS cluster with Karpenter, we experienced an incident where 2 out of 4 nodes went into NotReady state simultaneously. Despite having Node Auto Repair enabled (`nodeRepair: true`), Karpenter refused to repair these nodes because 50% > 20% threshold.

The event showed:
```
NodeRepairBlocked: more than 20% nodes are unhealthy in the nodepool
```

This left us with a degraded cluster for hours until manual intervention.

### New field in NodePool spec

```yaml
apiVersion: karpenter.sh/v1
kind: NodePool
spec:
  disruption:
    nodeRepairUnhealthyThreshold: "50%"  # or absolute count like "2"
```

### Accepted values

- Percentage: `"20%"`, `"50%"`, `"100%"` (string with % suffix)
- Absolute count: `"1"`, `"2"`, `"5"` (string representing integer)

### Default behavior

Default remains `"20%"` for backward compatibility. Existing NodePools without this field will behave exactly as before.

## Comparison with AWS Managed Node Groups

AWS EKS Managed Node Groups offer similar configurability:
- `maxUnhealthyNodeThresholdCount`
- `maxUnhealthyNodeThresholdPercentage`

See: https://docs.aws.amazon.com/eks/latest/userguide/node-health.html

This PR brings similar flexibility to Karpenter.

## Changes

- Added `NodeRepairUnhealthyThreshold` field to `Disruption` struct in `pkg/apis/v1/nodepool.go`
- Modified `pkg/controllers/node/health/controller.go` to use the NodePool's configured threshold instead of the hardcoded value
- Updated event messages to show the actual configured threshold

## Testing

- [ ] Unit tests (to be added)
- [ ] Manual testing on EKS cluster

## Related Issues

Fixes #2134

## Checklist

- [x] Code compiles correctly
- [x] Backward compatible (default is still 20%)
- [ ] Documentation updates
- [ ] Unit tests

### DIFF
--- a/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
+++ b/kwok/charts/crds/karpenter.sh_nodeclaims.yaml
@@ -9,389 +9,372 @@ spec:
   group: karpenter.sh
   names:
     categories:
-      - karpenter
+    - karpenter
     kind: NodeClaim
     listKind: NodeClaimList
     plural: nodeclaims
     singular: nodeclaim
   scope: Cluster
   versions:
-    - additionalPrinterColumns:
-        - jsonPath: .metadata.labels.node\.kubernetes\.io/instance-type
-          name: Type
-          type: string
-        - jsonPath: .metadata.labels.karpenter\.sh/capacity-type
-          name: Capacity
-          type: string
-        - jsonPath: .metadata.labels.topology\.kubernetes\.io/zone
-          name: Zone
-          type: string
-        - jsonPath: .status.nodeName
-          name: Node
-          type: string
-        - jsonPath: .status.conditions[?(@.type=="Ready")].status
-          name: Ready
-          type: string
-        - jsonPath: .metadata.creationTimestamp
-          name: Age
-          type: date
-        - jsonPath: .status.imageID
-          name: ImageID
-          priority: 1
-          type: string
-        - jsonPath: .status.providerID
-          name: ID
-          priority: 1
-          type: string
-        - jsonPath: .metadata.labels.karpenter\.sh/nodepool
-          name: NodePool
-          priority: 1
-          type: string
-        - jsonPath: .spec.nodeClassRef.name
-          name: NodeClass
-          priority: 1
-          type: string
-        - jsonPath: .status.conditions[?(@.type=="Drifted")].status
-          name: Drifted
-          priority: 1
-          type: string
-      name: v1
-      schema:
-        openAPIV3Schema:
-          description: NodeClaim is the Schema for the NodeClaims API
-          properties:
-            apiVersion:
-              description: |-
-                APIVersion defines the versioned schema of this representation of an object.
-                Servers should convert recognized schemas to the latest internal value, and
-                may reject unrecognized values.
-                More info: https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#resources
-              type: string
-            kind:
-              description: |-
-                Kind is a string value representing the REST resource this object represents.
-                Servers may infer this from the endpoint the client submits requests to.
-                Cannot be updated.
-                In CamelCase.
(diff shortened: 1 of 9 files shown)


engedaam:
Instead of introducing additional configuration surface area, would it make sense to provide opinionated default unhealthy thresholds based on cluster size, for example, small (<10 nodes), medium (<50 nodes), and large (<100 nodes) clusters? The specific node counts here are illustrative rather than prescriptive; the underlying idea is that different cluster scales may warrant different threshold categories.

Related to that, should this behavior be defined at the cluster level rather than per-node-pool? 

These thresholds were originally introduced in node repair as a safety mechanism to reactively block repairs during large-scale outages. Exposing this value as a configurable knob, especially at the node-pool level, would require users to reason about and tune it for each pool independently, which could make overall node repair behavior harder to understand and reason about across the cluster.

clark42:
Thanks for the feedback @engedaam!

I completely understand the safety concern - the 20% threshold exists as a guardrail to prevent cascading failures during large-scale outages, and exposing it as a configurable knob could lead users to inadvertently weaken this protection.

**Addressing your points:**

**1. The threshold is already per-NodePool, not cluster-level**

Looking at the current implementation in `controller.go`, the 20% is already evaluated per-NodePool (via `isNodePoolHealthy`), not at cluster level. This PR doesn't change the scope - it just makes the hardcoded value configurable.

**2. Documentation to mitigate the risk**

I'd be happy to add documentation explaining:
- The purpose of this threshold as a safety mechanism
- The risk of cascading failures when setting a higher value
- Recommended values based on NodePool size and criticality
- A clear warning that increasing this value should only be done with full understanding of the implications

This way, users who need flexibility (small pools, specific use-cases) can tune it, while the documentation ensures they understand the trade-offs.

**3. Context: small NodePools and AWS approach**

Our use-case: a 4-node pool where 2 nodes became unhealthy. With 20% (rounding to 0-1 node), auto-repair was blocked for 3+ hours. The safety mechanism designed for large-scale outages prevented recovery on a small pool.

AWS recently introduced similar configurability for managed node groups with `maxUnhealthyNodeThresholdCount` and `maxUnhealthyNodeThresholdPercentage` ([docs](https://docs.aws.amazon.com/eks/latest/userguide/node-health.html)), suggesting this need for tunability is recognized.

**Alternative if full configurability is still a concern:**

I'm also open to implementing automatic thresholds based on NodePool size (as you suggested), for example:
- A minimum floor like `max(20%, 2 nodes)` 
- Or tiered thresholds: small pools (<10 nodes) → 50%, medium (<50) → 30%, large → 20%

This would preserve safety for large pools while allowing small pools to recover, without requiring user configuration.

Let me know how you'd like to proceed!

PR #2740, opened 2025-12-23
feat: Add option to enabled warmup when leader election is used

GnatorX:
Fixes #N/A 

**Description**
- Add option to enable controller runtime warm up when leader election is used. Warm up allows non-leader pods to warm up k8s cache and watch for events hydrating the informer cache even when it isn't the leader. This allows faster fail over (don't have to hydrate informer after leader changes).

- Ran with the change. Follower memory didn't increase dramatically (note that this is simply the informer cache, any subsequent caches that is built after pod is created isn't ran until it becomes leader)
- Attempted to force a leader election by killing the leader and making sure follower comes up normally

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -184,7 +184,7 @@ func NewOperator(o ...option.Function[Options]) (context.Context, *Operator) {
 			// EnableWarmup allows controllers to start their sources (watches/informers) before leader election
 			// is won. This pre-populates caches and improves leader failover time. Only effective when leader
 			// election is enabled, so we only set it when both conditions are true.
-			EnableWarmup: lo.ToPtr(!options.FromContext(ctx).DisableLeaderElection && options.FromContext(ctx).EnableControllerWarmup),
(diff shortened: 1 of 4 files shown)


ellistarn:
Any reason to not have this on by default? 

There's a network tradeoff, but it seems very worthy given the risk of coordinated lists across controllers during zonal fail over.

GnatorX:
Same question was asked during working group. I think it make sense to turn it on by default but I think its likely safer to go first as not default then maybe flip it on as default after a release or two?

ellistarn:
Sounds good. If our intended goal state is on by default, wdyt about flipping the flag to "disable" and making it true by default (for now). 

The general idea being that the natural state of a Boolean is false. We haven't been perfect with this, but I think it generally helps with readability (short of something more sophisticated like feature gates). 

@jmdeal I'm noticing that feature gates has a comment to deprecate the field. What's the history there?

GnatorX:
That make sense to me, I can update it to do that. I also needed to test one more code path then will ask for another review

PR #2743, opened 2025-12-24
fix: support volume topology scheduling alternatives

### LINKED ISSUE #2742: Support multi-zone topology scheduling for persistent volumes (and storage class)
### Description

Karpenter's volume topology scheduling logic currently only considers the first allowed topology term from StorageClass and PersistentVolume definitions. When volumes define multiple topology terms (which are ORed together), Karpenter fails to properly schedule pods across all available options. This is particularly problematic with CSI drivers that generate PersistentVolumes with multiple allowed topologies - for example, the Azure CSI driver for ZRS (Zone-Redundant Storage) disks creates PVs with separate terms for each available zone.

**Observed Behavior**:

- When a StorageClass or PersistentVolume specifies multiple topology terms (e.g., multiple zones, regions, or other topology keys), only the first term's constraints are applied to node affinity. E.g. with the following affinity in PV, Karpenter will only consider `westus3-1` as allowed:

```yaml
nodeAffinity:
  required:
    nodeSelectorTerms:
    - matchExpressions: [{key: topology.disk.csi.azure.com/zone, operator: In, values: [westus3-1]}]
    - matchExpressions: [{key: topology.disk.csi.azure.com/zone, operator: In, values: [westus3-2]}]
    - matchExpressions: [{key: topology.disk.csi.azure.com/zone, operator: In, values: [westus3-3]}]
    - matchExpressions: [{key: topology.disk.csi.azure.com/zone, operator: In, values: [westus3-4]}]
```

Note that the equivalent alternative representation would have worked:

```yaml
nodeAffinity:
  required:
    nodeSelectorTerms:
    - matchExpressions: [{key: topology.disk.csi.azure.com/zone, operator: In, values: [westus3-1, westus3-2, westus3-3, westus3-4]}]
```

... but this is not what CSI driver generates. It looks like it does not have a choice, the driver [produces](https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/43b5cf0db35abd90764555e57dc91c5784ff4074/pkg/azuredisk/controllerserver.go#L381-L395) "accessible topology" segment maps, and external-provisioner [translates](https://github.com/kubernetes-csi/external-provisioner/blob/2b5ff2d86c3de548de753eb5e15ca99d0a99583e/pkg/controller/topology.go#L53-L75) each segment map into a nodeSelectorTerm, and this OR-of-ANDs with single-valued keys is there for a good reason, most likely can't be converted into a shorter form in a generic case.

- Pod anti-affinity rules combined with multi-term volumes cannot properly distribute pods as intended
- The OR semantics of multiple topology terms are lost, and the cross-product of requirements is not computed
- Workloads using CSI-provisioned volumes with multiple allowed topologies (e.g. Azure ZRS) fail to schedule correctly

**Expected Behavior**:

- All topology terms specified in StorageClass `AllowedTopologies` and PersistentVolume `NodeAffinity.Required` should be respected
- When volumes specify multiple topology terms (whether for zones, regions, or other topology keys), the scheduler should treat these as ORed options
- Scheduling simulation should properly compute the cartesian product of node affinity terms to maintain correct AND/OR semantics
- Pods should be able to schedule across all available topology options when combined with pod affinity/anti-affinity constraints

For this to manifest on PV, the following needs to be true:
* PV has more than one term (e.g. more than one allowed zone)
* PV exist when Pod is being scheduled
* Pod has constraint other than the first PV constraint (e.g. references second zone)

<details>
<summary>Example for Azure, using StorageClass with ZRS and immediate provisioning:</summary>

<P>
Create StorageClass for ZRS and immediate binding:

```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: managed-csi-zrs
provisioner: disk.csi.azure.com
parameters:
  skuName: Premium_ZRS
volumeBindingMode: Immediate
```

Create a PersistentVolumeClaim referencing that StorageClass:

```yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: zrs-pvc
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: managed-csi-zrs
  resources:
    requests:
      storage: 10Gi
```

At this point CSI driver will provision PV that looks something like this (note multiple nodeAffinity terms for zones):

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  annotations:
    pv.kubernetes.io/provisioned-by: disk.csi.azure.com
    volume.kubernetes.io/provisioner-deletion-secret-name: ""
    volume.kubernetes.io/provisioner-deletion-secret-namespace: ""
  creationTimestamp: "2025-12-24T03:33:21Z"
  finalizers:
  - external-provisioner.volume.kubernetes.io/finalizer
  - kubernetes.io/pv-protection
  name: pvc-4e13e628-ef1b-4d32-98f5-5735b5147886
  resourceVersion: "1242403"
  uid: 02ba67e0-2b40-4f78-8d72-294e4e61191e
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 10Gi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: zrs-pvc
    namespace: default
    resourceVersion: "1242387"
    uid: 4e13e628-ef1b-4d32-98f5-5735b5147886
  csi:
    driver: disk.csi.azure.com
    volumeAttributes:
      csi.storage.k8s.io/pv/name: pvc-4e13e628-ef1b-4d32-98f5-5735b5147886
      csi.storage.k8s.io/pvc/name: zrs-pvc
      csi.storage.k8s.io/pvc/namespace: default
      requestedsizegib: "10"
      skuName: Premium_ZRS
      storage.kubernetes.io/csiProvisionerIdentity: <snip>
    volumeHandle: <snip>
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: topology.disk.csi.azure.com/zone
          operator: In
          values:
          - westus3-1
      - matchExpressions:
        - key: topology.disk.csi.azure.com/zone
          operator: In
          values:
          - westus3-2
      - matchExpressions:
        - key: topology.disk.csi.azure.com/zone
          operator: In
          values:
          - westus3-3
      - matchExpressions:
        - key: topology.disk.csi.azure.com/zone
          operator: In
          values:
          - westus3-4
      - matchExpressions:
        - key: topology.disk.csi.azure.com/zone
          operator: In
          values:
          - ""
  persistentVolumeReclaimPolicy: Delete
  storageClassName: managed-csi-zrs
  volumeMode: Filesystem
status:
  lastPhaseTransitionTime: "2025-12-24T03:33:21Z"
  phase: Bound
```

Create pod referencing that PVC, but requesting the second zone:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: pod-in-zone-2
spec:
  affinity:
    nodeAffinity:
        nodeSelectorTerms:
        - matchExpressions:
            operator: In
            values:
            - westus3-2  # requesting second zone
  containers:
  - name: app
    image: nginx
    volumeMounts:
    - name: data
      mountPath: /data
  volumes:
  - name: data
    persistentVolumeClaim:
      claimName: zrs-pvc
```

Expected: Pod should schedule successfully in westus3-2 (the second allowed zone).
Actual: Pod fails to schedule because Karpenter only considers the first zone (westus3-1). The event on the pod will include requirement "topology.kubernetes.io/zone DoesNotExist".

</details>

**Versions**:
- Chart Version: 1.6.x (but should be the same in later versions)
- Kubernetes Version (`kubectl version`): tested on 1.33.5 (but should be the same on others)

tallaxes:
Fixes #2742

**Description**

Fixes volume topology scheduling to correctly handle multiple topology terms in PersistentVolumes and StorageClasses.

This PR updates volume topology handling to preserve OR semantics across PV `NodeSelectorTerms` and StorageClass `AllowedTopologies`, while keeping volume-derived topology constraints separate from pod affinity.

- Return volume topology requirements as a list of scheduling alternatives instead of flattening them into one requirement set
- Process all StorageClass `AllowedTopologies`, not just the first term
- Process all PersistentVolume node affinity `NodeSelectorTerms`, not just the first term
- Compute the cross product of topology alternatives when a pod references multiple PVC-backed volumes
- Add volume requirements only to node requirements, not pod affinity, so topology spread counting uses the pod's original constraints
- Prune incompatible cross-product branches when at least one valid volume topology alternative remains
- Preserve existing behavior when every branch is incompatible; treating that case as unschedulable requires separate validation/metrics handling and will be handled independently
- Preserve hostname-only Local and HostPath PV topology terms as unconstrained alternatives after hostname affinity is ignored
- Add explicit test helper fields for multi-term PV and StorageClass topology fixtures without changing existing `Zones` helper semantics
- Add regression coverage for multi-zone PVs, multi-zone StorageClasses, and mixed valid/invalid multi-volume topology alternatives

* Unit tests


pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -49,17 +50,6 @@ type VolumeTopology struct {
 }
 
 func (v *VolumeTopology) Inject(ctx context.Context, pod *v1.Pod) error {
```
jmdeal: We'll need a rebase in light of https://github.com/kubernetes-sigs/karpenter/pull/2780, but I think the logic stays largely the same. The main difference is that rather than relying on the existing logic to iterate through different pod requirements, we'll need to add additional logic in the scheduler to iterate over volume requirements. 

Right now we iterate on pod constraints in the outer loop of the scheduler - each time a pod fails to schedule, we iterate to the next set of constraints and add it back to the queue. Now we need to consider the cross product of two sets of requirements - volume requirements and pod requirements. Given the latter is normally a small set (typically 1), I'd probably do this within the NodeClaim to keep things simple, but we could also do it in the outer loop.

tallaxes: Ok, done, PTAL @jmdeal 

- **`volumetopology.go`** — Changed `GetRequirements()` to return `[]scheduling.Requirements` (list of alternatives). Each StorageClass `AllowedTopology` and PV `NodeSelectorTerm` becomes a separate alternative. Cross-product is computed when a pod has multiple volumes.
- **`provisioner.go`** — Updated `getVolumeTopologyRequirements()` return type from `map[UID]Requirements` to `map[UID][]Requirements`.
- **`scheduler.go`** — Updated `PodData.VolumeRequirements` and `Scheduler.volumeReqsByPod` types to use `[]scheduling.Requirements`.
- **`nodeclaim.go`** — Restructured `CanAdd()` to iterate over volume alternatives. Each alternative is tried with full scheduling checks (topology, instance types, offerings). If one zone fails (e.g., due to pod anti-affinity), the next zone is tried. Removed the standalone `addVolumeRequirements` helper.
- **`existingnode.go`** — Same iteration pattern as nodeclaim for existing node scheduling.

tallaxes:
Vulnerabilities flagged here and in other PRs addressed by https://github.com/kubernetes-sigs/karpenter/pull/2844

pkg/test/storage.go
jmdeal, line 1: Rather than updating the defaults and affecting all existing tests, can we just add a specific test for this format? That way we don't affect our existing test coverage for multi-value requirement tests, but we still get the necessary additional coverage.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -48,30 +48,55 @@ type VolumeTopology struct {
...
+			for _, volReq := range volAlts {
+				merged := scheduling.NewRequirements()
+				if existing != nil {
+					merged.Add(existing.Values()...)
+				}
+				merged.Add(volReq.Values()...)
+				newAlts = append(newAlts, merged)
+			}
+		}
+		alternatives = newAlts
```
jmdeal: There's a subtle issue here. Consider a case where we have two volumes, each associated with a storage class that has two AZs in it's allowed topologies.
```yaml
allowedTopologies:
  - matchLabelExpressions:
        values:
          - us-east-1a
  - matchLabelExpressions:
        values:
          - us-east-1b
```
We'll end up with the following requirements when we take the cross product:
- {zone IN us-east-1a} ∩ {zone IN us-east-1a} = {zone IN us-east-1a}
- {zone IN us-east-1a} ∩ {zone IN us-east-1b} = {zone DOES_NOT_EXIST}
- {zone IN us-east-1b} ∩ {zone IN us-east-1a} = {zone DOES_NOT_EXIST}
- {zone IN us-east-1b} ∩ {zone IN us-east-1b} = {zone IN us-east-1b}

In practice this should be fine since we don't expect there to be any instance types compatible with the `topology.kubernetes.io/zone DoesNotExist` requirement. There will be some no-op iterations, but that's just a performance concern. 

You could imagine this being an issue for a label which isn't present on every instance type. Since this would result in some DoesNotExist operators, we could select an instance that doesn't have a value defined for that label when in reality it's not compatible with any instance since it has a requirement on a disjoint set of values.

I don't think there's a method which creates this exact intersection with our requirements code today. We basically want to say take the intersection for any overlapping keys and pass through any non-overlapping keys. If the overlapping keys have disjoint values, error and discard the term. Here's a contrived example I came up with:

```yaml
# Volume A SC:
allowedTopologies:
  - matchLabelExpressions:
      - key: topology.custom.csi/rack
        values: [rack-1]
  - matchLabelExpressions:
      - key: topology.custom.csi/rack
        values: [rack-2]

# Volume B SC:
allowedTopologies:
  - matchLabelExpressions:
        values: [us-east-1a]
      - key: topology.custom.csi/rack
        values: [rack-1]
  - matchLabelExpressions:
        values: [us-east-1b]
      - key: topology.custom.csi/rack
        values: [rack-2]
```

- {rack IN rack-1} ∩ {zone IN 1a, rack IN rack-1} = {zone IN 1a, rack IN rack-1}
- {rack IN rack-1} ∩ {zone IN 1b, rack IN rack-2} = {zone IN 1b, rack IN DOES_NOT_EXIST}
  - Discarded because the rack requirements are disjoint
- {rack IN rack-2} ∩ {zone IN 1a, rack IN rack-1} = {zone IN 1a, rack IN DOES_NOT_EXIST}
  - Discarded because the rack requirements are disjoint
- {zone IN rackl-2} ∩ {zone IN 1b, rack IN rack-2} = {zone IN 1b, rack IN rack-2}

pkg/controllers/provisioning/scheduling/existingnode.go
```diff
@@ -91,14 +91,41 @@ func (n *ExistingNode) CanAdd(pod *v1.Pod, podData *PodData, volumes scheduling.
...
+	baseRequirements := scheduling.NewRequirements(n.requirements.Values()...)
+	baseRequirements.Add(podData.Requirements.Values()...)
+
+	// Build the list of volume requirement alternatives to try.
+	volumeAlternatives := podData.VolumeRequirements
+	if len(volumeAlternatives) == 0 {
+		volumeAlternatives = []scheduling.Requirements{nil}
+	}
+
+	// Try each volume alternative. Choosing a zone for volumes affects topology checks.
```
jmdeal: I've noticed quite a few comments are making the assumption that topology will only be zones. Since we support arbitrary CSI drivers, we shouldn't make that assumption. I know at least one counter-example used by Karpenter today: storage classes for EKS Auto Mode's EBS CSI Driver should be configured with a `eks.amazonaws.com/compute-type: auto` selector in allowed topologies.

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: Really, I think this is an issue with the intersection of disjoint sets resulting in a DoesNotExist requirement. It's not an issue in practice since wherever we do intersections we do compatibility checks first, but it leads to subtle issues like this. I'm not sure if this behavior is actually relied on anywhere, if not I think it's worth updating. That doesn't need to be scoped to this PR though.

pkg/test/storage.go
tallaxes: Done; added a way to pass selector terms into options https://github.com/kubernetes-sigs/karpenter/pull/2743/commits/fce197f976d3ddd90723b7e8556f1e3c76e34c80

pkg/controllers/provisioning/scheduling/existingnode.go
tallaxes: Addressed https://github.com/kubernetes-sigs/karpenter/pull/2743/commits/15e740424b3634f65338fd39307f46c4c930b844

pkg/controllers/provisioning/scheduling/volumetopology.go
tallaxes: Yeah, good point, and disjoint alternatives could survive in the new codepath. I pulled the fix into stacked PR https://github.com/tallaxes/karpenter-core/pull/1 for a smaller review; it prunes incompatible branches before merging, threads the filtered pod set through provisioning, and adds regression coverage. @jmdeal, please let me know if this makes sense / is directionally correct, and I will merge it here. (Alternatively, could also handle this as a follow-up PR, just to land the core fix)

tallaxes: As discussed in [tallaxes/karpenter-core#1](https://github.com/tallaxes/karpenter-core/pull/1#discussion_r3237176832), I pulled the mixed valid/invalid disjoint-branch handling back into this PR, in  https://github.com/kubernetes-sigs/karpenter/pull/2743/commits/2dae1125994d6ee1e68fd2b01739d1d1b1d477c2. This covers the original example by pruning incompatible cross-product branches when at least one valid volume topology alternative remains. Leaving the all-pruned case unchanged, since treating that as unschedulable needs separate validation/metrics handling in a follow-up.

tallaxes: Also note a small additional fix/test in https://github.com/kubernetes-sigs/karpenter/pull/2743/commits/c922c37ddae44bdb9c4864ae07bfa2c26efd6893. It extends the multi-term OR semantics to local/hostPath PVs: when Karpenter strips hostname-keyed expressions from a PV's NodeAffinity (since it can't satisfy a specific hostname on new capacity), a term whose only expression was the hostname becomes unconstrained and, as an OR branch, makes the PV satisfiable regardless of the other terms. Added a test covering local and hostPath volumes.

I am thinking it belongs here rather than a follow-up because it's the same OR-of-NodeSelectorTerms code path being changed - just the local/hostPath flavor of it - but let me know if you disagree.

PR #2745, opened 2025-12-24
feat: Add Pod Disruption Schedule

### LINKED ISSUE #1719: Pod disruption window
### Description

I have some workloads that are sensitive to interruptions at certain points of the day and thus are using the `karpenter.sh/do-not-disrupt` annotation. I'd like the ability to allow disruptions to these pods at specific points via cron format schedule.

In order to allow reclaiming nodes for expiration or underutilization I'm currently running my own controller that watches DisruptionBlocked events and then removes the do-not-disrupt annotation if the pods are marked with another one indicating the schedule for when disruptions are allowed. I'd like to add this upstream and get rid of my own controller.

* karpenter.sh/disruption-schedule - cron format of when disruptions are allowed (e.g. `0 14 * * 6`)
* karpenter.sh/disruption-schedule-duration - duration for which the schedule is active (e.g. `3h`)

jukie:
Fixes #1719 
Related to #2744 (RFC)

**Description**
This is a renewal of https://github.com/kubernetes-sigs/karpenter/pull/1720 and adds the ability to define a pod disruption schedule at the pod-level which behaves somewhat akin to nodepool disruption budgets.

Note: This was mostly AI generated but I have reviewed and understand the implementation. I've built a custom controller that runs alongside Karpenter with similar logic and have previously added this functionality in #1720.

Add tests and will run inside a real cluster and update this comment with validation

### DIFF
--- a/pkg/apis/v1/labels.go
+++ b/pkg/apis/v1/labels.go
@@ -49,6 +49,8 @@ const (
 // Karpenter specific annotations
 const (
 	DoNotDisruptAnnotationKey                  = apis.Group + "/do-not-disrupt"
+	DisruptionScheduleAnnotationKey            = apis.Group + "/disruption-schedule"
+	DisruptionScheduleDurationAnnotationKey    = apis.Group + "/disruption-schedule-duration"
 	ProviderCompatibilityAnnotationKey         = apis.CompatibilityGroup + "/provider"
 	NodePoolHashAnnotationKey                  = apis.Group + "/nodepool-hash"
 	NodePoolHashVersionAnnotationKey           = apis.Group + "/nodepool-hash-version"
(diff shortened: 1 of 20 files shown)


ellistarn:
I think we want to think holistically about this space of pod oriented disruption controls, including whether or not these features should land in upstream's PDB. 

Happy to discuss more in working group.

jukie:
Having something in upstream PDBs would be great and I can start a KEP or design spec for that as well but in the best case scenario I suspect it'd be over a year before that becomes available. Would you be open to solving this within Karpenter first to figure out the best model which can then be added upstream? Perhaps you'd prefer the CRD approach described in my RFC vs this example?

This issues has been open since 2024 along with similar issues describing the desire for more pod-level disruption controls. I agree that we should think holistically about a solution but would also like to get some sense of progress in the area. My proposal would solve a real problem that exists in Karpenter where it's currently unsafe to use do-not-disrupt for anything besides ephemeral workloads that eventually exit on their own and doing so risks violating PDBs upon Nodes reaching max expiration + grace period. As a result of hitting that in the real world I currently maintain a custom controller as a workaround.

ellistarn:
I wonder if we might want to consider an alpha API in Karpenters group that is our proposal for an upstream KEP and we target eventually landing it upstream. It might not be able to solve everything (I.e. surge eviction), but could be a good middle ground to get things moving.

PR #2749, opened 2025-12-30
chore: throw errors out of nodeclaim informer loop

DerekFrank:
Fixes #N/A 

**Description**

This causes the error to get backoff retried instead of just logged

make presubmit

### DIFF
--- a/pkg/controllers/state/informer/nodeclaim.go
+++ b/pkg/controllers/state/informer/nodeclaim.go
@@ -24,7 +24,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
@@ -68,7 +67,7 @@ func (c *NodeClaimController) Reconcile(ctx context.Context, req reconcile.Reque
 			// notify cluster state of the node deletion
 			c.cluster.DeleteNodeClaim(req.Name)
 			if deleteErr := c.clusterCost.DeleteNodeClaim(ctx, nodeClaim); deleteErr != nil {
-				log.FromContext(ctx).Error(deleteErr, "failed to remove nodeclaim from cost tracking")
+				return reconcile.Result{}, deleteErr
 			}
 		}
 		return reconcile.Result{}, client.IgnoreNotFound(err)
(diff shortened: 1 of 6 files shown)


pkg/controllers/state/informer/nodeclaim.go
```diff
@@ -68,7 +67,7 @@ func (c *NodeClaimController) Reconcile(ctx context.Context, req reconcile.Reque
 			// notify cluster state of the node deletion
 			c.cluster.DeleteNodeClaim(req.Name)
 			if deleteErr := c.clusterCost.DeleteNodeClaim(ctx, nodeClaim); deleteErr != nil {
-				log.FromContext(ctx).Error(deleteErr, "failed to remove nodeclaim from cost tracking")
+				return reconcile.Result{}, deleteErr
```
moko-poi, line 70: @DerekFrank 
Thanks for the PR! I have a concern about this change. Returning errors here will cause the reconcile loop to fail. Since cost calculation is primarily for observability purposes, I think it would be better to just log the error rather than failing the main reconcile loop. What do you think?

DerekFrank: Its pretty standard practice to throw errors in a reconcile loop if they are retryable. If you throw an error in a reconcile loop, it'll get retried with backoff. 

The tradeoff if you don't throw the error is that it never gets retried and the metric becomes inaccurate. Thats not that much of a concern for adding nodeclaims, but if you don't ever remove nodeclaims you'll leak memory and the metric will slowly diverge from the truth.

moko-poi: You're right! I see now that the cluster state operations happen before the cost tracking calls, so the core state updates are not affected. Thanks for the explanation!

moko-poi (approved):

pkg/state/cost/cost.go
```diff
@@ -223,8 +223,8 @@ func (cc *ClusterCost) UpdateNodeClaim(ctx context.Context, nodeClaim *v1.NodeCl
 	if _, found := lo.Find(nodeClaim.GetOwnerReferences(), func(o metav1.OwnerReference) bool {
 		return o.Kind == object.GVK(np).Kind && o.UID == np.UID
 	}); !found {
-		failed = true
-		return serrors.Wrap(fmt.Errorf("nodepool not found for nodeclaim"), "nodepool", nodeClaim.Labels[v1.NodePoolLabelKey], "nodeclaim", klog.KObj(nodeClaim))
+		// Technically not an error, as users can create NodeClaims manually without a NodePool.
```
ryan-mist: Is this partially captured a few lines above this in `cc.client.Get(ctx, types.NamespacedName{Name: nodeClaim.Labels[v1.NodePoolLabelKey]}, np)`?

DerekFrank: Good point!

PR #2752, opened 2025-12-31
chore: remove cluster metric controller, standardize labels into metrics constants file

DerekFrank:
Fixes #N/A 

**Description**

Since the cluster cost metrics are nodepool based, they should be in the nodepool metrics controller. There are also several standard metric labels that can be used more liberally throughout the project.

`make presubmit`


pkg/controllers/metrics/cluster/suite_test.go
```diff
@@ -1,67 +0,0 @@
...
-	clusterCost = cost.NewClusterCost(ctx, cp, env.Client)
-	clusterController = cluster.NewController(env.Client, clusterCost)
-})
-
-var _ = AfterSuite(func() {
-	Expect(env.Stop()).To(Succeed(), "Failed to stop environment")
-})
-
-var _ = Describe("Controller", func() {
-	It("should not panic when reconciling with nodepools", func() {
```
engedaam, line 59: nit. Do we want to move this test else where?

PR #2758, opened 2026-01-03
chore: fix flaking static test, refactor to use lo

DerekFrank:
Fixes #N/A 

**Description**

This test flakes occasionally: https://github.com/kubernetes-sigs/karpenter/actions/runs/20669098988/job/59346783741?pr=2757

Unfortunately I ran it 115 times with `ginkgo -until-it-fails` before the fix and could not get it to flake, but this should close some race conditions in the test

### DIFF
--- a/pkg/controllers/static/provisioning/suite_test.go
+++ b/pkg/controllers/static/provisioning/suite_test.go
@@ -19,7 +19,6 @@ package static_test
 import (
 	"context"
 	"fmt"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/static/provisioning/suite_test.go
```diff
@@ -546,20 +547,18 @@ var _ = Describe("Static Provisioning Controller", func() {
 
 			// Run many reconciles in parallel
 			n := 50
-			errs := make(chan error, n)
+			wg := sync.WaitGroup{}
```
ellistarn: https://github.com/samber/lo/blob/master/parallel/slice.go#L8

pkg/controllers/static/provisioning/suite_test.go
```diff
@@ -546,33 +546,27 @@ var _ = Describe("Static Provisioning Controller", func() {
...
 			cluster.SetSynced(true)
-			ExpectObjectReconciled(ctx, env.Client, controller, nodePool)
-			// at the end we should have right counts in StateNodePool
-			ExpectStateNodePoolCount(cluster, nodePool.Name, 10, 0, 0)
+
+			Eventually(func() {
+				ExpectObjectReconciled(ctx, env.Client, controller, nodePool)
+				// at the end we should have right counts in StateNodePool
+				ExpectStateNodePoolCount(cluster, nodePool.Name, 10, 0, 0)
+			}, 10*time.Second)
```
ellistarn, line 569: Can you use WithTimeout and WithInterval? Want to make sure this interval is fast so we're not waiting.

ellistarn: nvm checked and interval is 10ms. nonblocking

PR #2771, opened 2026-01-10
docs: add details for e2e test coverage

Kausheel:
**Description**
Adds more granular details about each e2e test in the `test/README.md` to give contributors a better/faster overview of the test coverage without needing to dive into the test code to discover this information.

The change also updates the deprecated naming convention which was made in #2234.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -172,37 +172,56 @@ jobs:
 
 The Karpenter core test framework includes the following test suites that will be executed against your provider implementation:
 
-### Integration Tests (`test/suites/integration`)
+### Regression (`test/suites/regression`)
 
-1. **Basic Integration Tests**
-   - DaemonSet compatibility
-   - Pod scheduling with various constraints
-   - Resource allocation and limits
-   - Node startup and registration
+1. **Integration**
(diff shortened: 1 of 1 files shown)


ellistarn:
I think this kind of "code structure documentation" is better derived by agents at this point. I'd accept a PR that simply removes these suites.

Kausheel:
Hi @ellistarn !

I agree that many devs will be relying on agents for parsing the repo, but I think there's still value on keeping the test explanation in the README. 

There are still many devs who aren't exclusively relying on agents to read repo context who will prefer a human-readable explanation, and even when an agent is used, they can parse a README faster than parsing all the code.

What do you think? If you still prefer me to remove this section from the README, let me know your preference and I'll update the PR. 

Cheers!

ellistarn:
Did you get this PR with AI? If the answer is yes, let's move onto more interesting things 😉.

Kausheel:
Hi Ellis, 
The details in the PR were assisted with AI, but it took a while for an agent to parse the whole repo to determine accurate descriptions, hence I thought updating the doc with more details would help others who are trying to run e2e tests like me. No worries, I'll close the PR so we can move on to other things 😊

PR #2780, opened 2026-01-13
fix: separate volume requirements from TSC calculations

### LINKED ISSUE #1239: Zonal Volume Requirements Break Topology Spread Constraints

### LINKED ISSUE #2777: Volume zone injection pollutes TSC calculations causing minDomains scheduling failures
### Description

**Observed Behavior**:

When a StatefulSet pod with a zone-bound PVC is rescheduled (e.g., after node deletion), the pod remains `Pending` indefinitely even though:
- The PVC is bound to a valid zone
- Other pods are already spread across 3 zones satisfying minDomains
- The resulting distribution (2-2-1) would satisfy maxSkew=1

Example (based on the reproduction steps below):

When scheduling `poc-sts-4`:

1. **Volume injection**: Karpenter sees PVC is bound to `us-west-2a`, injects `topology.kubernetes.io/zone: us-west-2a` into pod's NodeAffinity

2. **TSC pod counting (BUG)**: Because of the injected affinity, Karpenter's `TopologyNodeFilter` only counts pods in `us-west-2a`:
   - us-west-2a: 1 pod (poc-sts-2)
   - us-west-2b: **0** (filtered out!)
   - us-west-2c: **0** (filtered out!)

3. **minDomains check (BUG)**: Karpenter thinks `podDomains = {us-west-2a}` (only 1 zone), so:
   - `numPodSupportedDomains = 1`
   - `minDomains = 3`
   - Since `1 < 3`, Karpenter sets `minCount = 0` (penalty mode)

4. **Skew calculation fails**: With incorrect counts, Karpenter calculates an invalid skew and rejects all candidate nodes. `poc-sts-4` stays pending.
  

**Expected Behavior**:

| Zone | Pods | After poc-sts-4 |
|------|------|-----------------|
| `us-west-2a` | poc-sts-2 | poc-sts-2, poc-sts-4 |
| `us-west-2b` | poc-sts-1, poc-sts-3 | poc-sts-1, poc-sts-3 |
| `us-west-2c` | poc-sts-0 | poc-sts-0 |

**Final distribution: 2-2-1 (skew = 2-1 = 1) ✓ Valid!**, the pod should schedule on a new node claim in `us-west-2a`.

**Reproduction Steps**:

1. Apply the following statefulset + nodepool YAML in a k8s cluster:
    <details><summary>YAML </summary>
    <p>
    
    ```
    # Karpenter POC - Single file for reproducing Karpenter PVC issue
    ---
    apiVersion: v1
    kind: Namespace
    metadata:
      name: karpenter-poc
    ---
    apiVersion: storage.k8s.io/v1
    kind: StorageClass
    metadata:
      name: poc-gp3
    provisioner: ebs.csi.aws.com
    parameters:
      type: gp3
      encrypted: "true"
    volumeBindingMode: WaitForFirstConsumer
    allowVolumeExpansion: true
    reclaimPolicy: Delete
    ---
    apiVersion: karpenter.k8s.aws/v1
    kind: EC2NodeClass
    metadata:
      name: poc-node
      annotations:
        kubernetes.io/description: "EC2NodeClass for Karpenter POC"
    spec:
      amiFamily: AL2023
      amiSelectorTerms:
        - alias: al2023@latest
      kubelet:
        imageGCLowThresholdPercent: 50
        imageGCHighThresholdPercent: 60
      metadataOptions:
        httpEndpoint: enabled
        httpProtocolIPv6: disabled
        httpPutResponseHopLimit: 1
        httpTokens: required
      role: example-role
      subnetSelectorTerms:
        - tags:
            karpenter.sh/discovery: example
      securityGroupSelectorTerms:
        - tags:
            karpenter.sh/discovery: example
      tags:
        nodeTemplate: poc-node
        purpose: karpenter-poc
      blockDeviceMappings:
        - deviceName: /dev/xvda
          ebs:
            encrypted: true
            volumeSize: 40Gi
            volumeType: gp3
    ---
    apiVersion: karpenter.sh/v1
    kind: NodePool
    metadata:
      name: poc-nodepool
      annotations:
        kubernetes.io/description: "NodePool for Karpenter POC"
    spec:
      disruption:
        consolidateAfter: 30m
        budgets:
          - nodes: "1"
      template:
        metadata:
          labels:
            app: karpenter-poc
            node-type: karpenter-poc
        spec:
          expireAfter: Never
          requirements:
              operator: In
              values:
                - on-demand
            - key: kubernetes.io/arch
              operator: In
              values:
                - arm64
            - key: kubernetes.io/os
              operator: In
              values:
                - linux
              operator: Gt
              values:
                - "6"
              operator: In
              values:
                - m
                - c
            - key: karpenter.k8s.aws/instance-hypervisor
              operator: In
              values:
                - nitro
          nodeClassRef:
            group: karpenter.k8s.aws
            kind: EC2NodeClass
            name: poc-node
          taints:
            - key: node-type
              value: karpenter-poc
              effect: NoSchedule
    ---
    apiVersion: apps/v1
    kind: StatefulSet
    metadata:
      name: poc-sts
      namespace: karpenter-poc
      labels:
        app: poc-sts
        app.kubernetes.io/component: poc
    spec:
      serviceName: poc-sts
      replicas: 5
      podManagementPolicy: Parallel
      updateStrategy:
        type: RollingUpdate
      selector:
        matchLabels:
          app: poc-sts
      template:
        metadata:
          labels:
            app: poc-sts
            app.kubernetes.io/component: poc
        spec:
          tolerations:
            - effect: NoSchedule
              key: node-type
              operator: Equal
              value: karpenter-poc
          topologySpreadConstraints:
            - maxSkew: 1
              minDomains: 3
              labelSelector:
                matchLabels:
                  app: poc-sts
          nodeSelector:
            node-type: karpenter-poc
          affinity:
            podAntiAffinity:
                - labelSelector:
                    matchExpressions:
                      - key: app.kubernetes.io/component
                        operator: In
                        values:
                          - poc
                  topologyKey: "kubernetes.io/hostname"
          containers:
            - name: busybox
              command:
                - sleep
                - "3600"
              image: docker.io/busybox:1.36.1
              resources:
                requests:
                  memory: 64Mi
                  cpu: 10m
              volumeMounts:
                - name: poc-data
                  mountPath: /data
      volumeClaimTemplates:
        - metadata:
            name: poc-data
          spec:
            accessModes: ["ReadWriteOnce"]
            storageClassName: poc-gp3
            resources:
              requests:
                storage: 10Gi
    ---
    apiVersion: v1
    kind: Service
    metadata:
      name: poc-sts
      namespace: karpenter-poc
      labels:
        app: poc-sts
    spec:
      clusterIP: None
      selector:
        app: poc-sts
      ports:
        - name: placeholder
          port: 80
          targetPort: 80
    ```
    
    </p>
    </details> 

2. Observe that pods `poc-sts-0`, `poc-sts-1`, `poc-sts-2`, `poc-sts-3`, and `poc-sts-4` successfully schedule.
3. Delete the node that `poc-sts-4` belongs to.
4. Observe that `poc-sts-4` is stuck pending as Karpenter won't create a new node claim.

This is most likely related to: https://github.com/kubernetes-sigs/karpenter/issues/1239.

**Versions**:
- Chart Version: v1.7.1 (but also exists on `main`)
- Kubernetes Version (`kubectl version`): v1.32

k24dizzle:
Fixes #2777, fixes #1239

### The Problem

When scheduling a pod with a PVC bound to a specific zone, Karpenter's `VolumeTopology.Inject()` added the volume's zone requirement directly into the pod's `NodeAffinity`. This "polluted" affinity then incorrectly affected:

1. **TSC pod counting** - The `TopologyNodeFilter` only counted pods in the volume's zone instead of all zones
2. **minDomains validation** - `podDomains` was restricted to 1 zone, incorrectly failing the `minDomains` check if that was set in the TSC.

This caused pods to remain pending because Karpenter believed the TSC constraints were unsatisfiable, even though provisioning a new node in the correct zone would have satisfied them.

### The Fix

Refactor to keep volume requirements **separate** from the pod's `NodeAffinity`:

- `VolumeTopology.Inject()` → `VolumeTopology.GetRequirements()` (no longer modifies pod)
- Volume requirements stored in `PodData.VolumeRequirements`
- Volume requirements added only to `nodeRequirements` (for NodeClaim zone selection), **not** to pod affinities (for TSC counting)
- Added compatibility check before adding volume requirements to catch conflicts (e.g., PV with hostname affinity to non-existent node)

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -533,9 +533,9 @@ func validateKarpenterManagedLabelCanExist(p *corev1.Pod) error {
 // getVolumeTopologyRequirements collects volume topology requirements for each pod
 // WITHOUT modifying the pods. These requirements will be added to nodeRequirements
 // (for NodeClaim zone selection) but NOT to pod affinities (for correct TSC counting).
-func (p *Provisioner) getVolumeTopologyRequirements(ctx context.Context, pods []*corev1.Pod) ([]*corev1.Pod, map[types.UID][]corev1.NodeSelectorRequirement, error) {
+func (p *Provisioner) getVolumeTopologyRequirements(ctx context.Context, pods []*corev1.Pod) ([]*corev1.Pod, map[types.UID]scheduling.Requirements, error) {
 	var schedulablePods []*corev1.Pod
-	volumeReqs := make(map[types.UID][]corev1.NodeSelectorRequirement)
+	volumeReqs := make(map[types.UID]scheduling.Requirements)
 	for _, pod := range pods {
 		reqs, err := p.volumeTopology.GetRequirements(ctx, pod)
 		if err != nil {
--- a/pkg/controllers/provisioning/scheduling/nodeclaim.go
+++ b/pkg/controllers/provisioning/scheduling/nodeclaim.go
@@ -460,14 +460,13 @@ func fits(instanceType *cloudprovider.InstanceType, requests corev1.ResourceList
 
 // addVolumeRequirements adds volume topology requirements to nodeRequirements after checking compatibility.
 // This catches cases like a PV with hostname affinity to a specific node that conflicts with the NodeClaim.
-func addVolumeRequirements(nodeRequirements scheduling.Requirements, volumeRequirements []corev1.NodeSelectorRequirement, opts ...option.Function[scheduling.CompatibilityOptions]) error {
+func addVolumeRequirements(nodeRequirements scheduling.Requirements, volumeRequirements scheduling.Requirements, opts ...option.Function[scheduling.CompatibilityOptions]) error {
 	if len(volumeRequirements) == 0 {
 		return nil
 	}
-	volumeReqs := scheduling.NewNodeSelectorRequirements(volumeRequirements...)
-	if err := nodeRequirements.Compatible(volumeReqs, opts...); err != nil {
+	if err := nodeRequirements.Compatible(volumeRequirements, opts...); err != nil {
 		return fmt.Errorf("incompatible volume requirements, %w", err)
 	}
-	nodeRequirements.Add(volumeReqs.Values()...)
+	nodeRequirements.Add(volumeRequirements.Values()...)
 	return nil
 }
--- a/pkg/controllers/provisioning/scheduling/scheduler.go
+++ b/pkg/controllers/provisioning/scheduling/scheduler.go
@@ -124,7 +124,7 @@ func NewScheduler(
 	daemonSetPods []*corev1.Pod,
 	recorder events.Recorder,
 	clock clock.Clock,
-	volumeReqsByPod map[types.UID][]corev1.NodeSelectorRequirement,
+	volumeReqsByPod map[types.UID]scheduling.Requirements,
 	opts ...Options,
 ) *Scheduler {
 	minValuesPolicy := option.Resolve(opts...).minValuesPolicy
@@ -188,7 +188,7 @@ type PodData struct {
 	Requirements             scheduling.Requirements
 	StrictRequirements       scheduling.Requirements
 	HasResourceClaimRequests bool
-	VolumeRequirements       []corev1.NodeSelectorRequirement // Volume topology requirements
+	VolumeRequirements       scheduling.Requirements // Volume topology requirements
 }
 
 type Scheduler struct {
@@ -199,8 +199,8 @@ type Scheduler struct {
 	remainingResources      map[string]corev1.ResourceList // (NodePool name) -> remaining resources for that NodePool
 	daemonOverhead          map[*NodeClaimTemplate]corev1.ResourceList
 	daemonHostPortUsage     map[*NodeClaimTemplate]*scheduling.HostPortUsage
-	cachedPodData           map[types.UID]*PodData                         // (Pod Namespace/Name) -> pre-computed data for pods to avoid re-computation and memory usage
-	volumeReqsByPod         map[types.UID][]corev1.NodeSelectorRequirement // Volume topology requirements per pod
+	cachedPodData           map[types.UID]*PodData                // (Pod Namespace/Name) -> pre-computed data for pods to avoid re-computation and memory usage
+	volumeReqsByPod         map[types.UID]scheduling.Requirements // Volume topology requirements per pod
 	preferences             *Preferences
 	topology                *Topology
 	cluster                 *state.Cluster
--- a/pkg/controllers/provisioning/scheduling/topology_test.go
+++ b/pkg/controllers/provisioning/scheduling/topology_test.go
@@ -3123,6 +3123,7 @@ var _ = Describe("Taints", func() {
 
 // Test for TSC + Volume bug: Pod with PVC bound to a specific zone and TSC with minDomains
 // should schedule correctly. Previously, volume zone requirements polluted TSC calculations.
+// See issues 1239 and 2777 for more context.
 var _ = Describe("Topology with Volume Requirements", func() {
 	var nodePool *v1.NodePool
 	labels := map[string]string{"app": "test"}
--- a/pkg/controllers/provisioning/scheduling/volumetopology.go
+++ b/pkg/controllers/provisioning/scheduling/volumetopology.go
@@ -53,7 +53,7 @@ type VolumeTopology struct {
 // These requirements should be:
 //   - Added to nodeRequirements (for NodeClaim zone selection)
 //   - NOT added to pod's NodeAffinity (to preserve correct TSC counting)
-func (v *VolumeTopology) GetRequirements(ctx context.Context, pod *v1.Pod) ([]v1.NodeSelectorRequirement, error) {
+func (v *VolumeTopology) GetRequirements(ctx context.Context, pod *v1.Pod) (scheduling.Requirements, error) {
 	var requirements []v1.NodeSelectorRequirement
 	for _, volume := range pod.Spec.Volumes {
 		req, err := v.getRequirements(ctx, pod, volume)
@@ -66,9 +66,9 @@ func (v *VolumeTopology) GetRequirements(ctx context.Context, pod *v1.Pod) ([]v1
 		return nil, nil
 	}
 	log.FromContext(ctx).
-		WithValues("Pod", klog.KObj(pod)).
-		V(1).Info(fmt.Sprintf("getting requirements derived from pod volumes, %s", requirements))
-	return requirements, nil
+		WithValues("Pod", klog.KObj(pod), "requirements", requirements).
+		V(1).Info("getting requirements from pod volumes")
+	return scheduling.NewNodeSelectorRequirements(requirements...), nil
 }
 
 func (v *VolumeTopology) getRequirements(ctx context.Context, pod *v1.Pod, volume v1.Volume) ([]v1.NodeSelectorRequirement, error) {


pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -48,42 +48,27 @@ type VolumeTopology struct {
...
-		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[i].MatchExpressions = append(
-			pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[i].MatchExpressions, requirements...)
+		return nil, nil
 	}
-
 	log.FromContext(ctx).
 		WithValues("Pod", klog.KObj(pod)).
-		V(1).Info(fmt.Sprintf("adding requirements derived from pod volumes, %s", requirements))
-	return nil
+		V(1).Info(fmt.Sprintf("volume topology requirements: %s", requirements))
```
jigisha620:
```suggestion
		V(1).Info(fmt.Sprintf("getting requirements derived from pod volumes, %s", requirements))
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -196,7 +199,8 @@ type Scheduler struct {
 	remainingResources      map[string]corev1.ResourceList // (NodePool name) -> remaining resources for that NodePool
 	daemonOverhead          map[*NodeClaimTemplate]corev1.ResourceList
 	daemonHostPortUsage     map[*NodeClaimTemplate]*scheduling.HostPortUsage
-	cachedPodData           map[types.UID]*PodData // (Pod Namespace/Name) -> pre-computed data for pods to avoid re-computation and memory usage
```
jigisha620, line 199: Is there a reason for dropping this comment?

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -3120,3 +3120,91 @@ var _ = Describe("Taints", func() {
...
+				},
+			},
+		})
+	})
+
+	It("should schedule pod with PVC and TSC minDomains=3 when pods are spread across 3 zones", func() {
+		// This test reproduces the bug where a pod with:
+		// 1. A PVC bound to zone-1
+		// 2. TSC with minDomains=3, maxSkew=1
+		// Would fail to schedule because volume zone requirement polluted TSC calculations
```
jigisha620: Maybe we can say previously it would have failed to schedule? Or just avoid explaining the previous bug here and just tell what should happen in the test?

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -3120,3 +3120,91 @@ var _ = Describe("Taints", func() {
...
+			ExpectScheduled(ctx, env.Client, pod)
+		}
+		// Verify spread: should be 1 pod per zone
+		ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(1, 1, 1))
+
+		// Now create a pod with PVC + same TSC
+		// BUG: This fails because volume zone (test-zone-1) pollutes podDomains,
+		// causing minDomains check to think only 1 zone is valid.
+		// FIX: Volume requirements should NOT affect TSC counting.
+		podWithPVC := test.UnschedulablePod(test.PodOptions{
```
jigisha620, line 3191: In my opinion, we shouldn't call out the bug here because once this code gets merged, the bug should be fixed. So we can talk about what should happen instead?

pkg/controllers/provisioning/scheduling/scheduler.go
k24dizzle: oops no there isn't, adding it back

pkg/controllers/provisioning/provisioner.go
```diff
@@ -286,13 +286,14 @@ func (p *Provisioner) NewScheduler(
 		instanceTypes[np.Name] = its
 	}
 
-	// inject topology constraints
-	pods, err = p.injectVolumeTopologyRequirements(ctx, pods)
+	// Get volume topology requirements WITHOUT modifying pods.
+	// Volume requirements are passed separately and added to nodeRequirements only.
+	pods, volumeReqs, err := p.getVolumeTopologyRequirements(ctx, pods)
```
k24dizzle, line 292: Similar to the previous injection [logic](https://github.com/kubernetes-sigs/karpenter/blob/ea4d81b89ee720241396e35a1b6d26f1d1238be6/pkg/controllers/provisioning/provisioner.go#L530-L543), any "volume injection" (or getting the volume requirements) failures would've excluded the pod from being included in the remaining scheduling logic. But I imagine this happens very rarely, so I went ahead and did your suggestion.

jigisha620: Actually you are right, we should keep it the way you had it before. I was thinking about it again and realized that we need it because if there is any failure in retrieving the volumetopology for the pod, we should exclude it from the list of schedulable pods. Sorry for the confusion.

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -48,42 +48,27 @@ type VolumeTopology struct {
...
-
-	// We add our volume topology zonal requirement to every node selector term.  This causes it to be AND'd with every existing
-	// requirement so that relaxation won't remove our volume requirement.
-	for i := 0; i < len(pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms); i++ {
-		pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[i].MatchExpressions = append(
-			pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms[i].MatchExpressions, requirements...)
+		return nil, nil
 	}
-
 	log.FromContext(ctx).
```
jmdeal, line 68: I'd drop the log here and add it where we're actually performing the injection into node requirements. Also, this would be a good opportunity to change this to a structured log (i.e. using `WithValues` instead of formatting the requirements into the string).

pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -48,42 +48,27 @@ type VolumeTopology struct {
 	kubeClient client.Client
 }
 
-func (v *VolumeTopology) Inject(ctx context.Context, pod *v1.Pod) error {
+// GetRequirements returns the volume topology requirements for the pod.
+//
+// These requirements should be:
+//   - Added to nodeRequirements (for NodeClaim zone selection)
+//   - NOT added to pod's NodeAffinity (to preserve correct TSC counting)
+func (v *VolumeTopology) GetRequirements(ctx context.Context, pod *v1.Pod) ([]v1.NodeSelectorRequirement, error) {
```
jmdeal: Thoughts on updating this to return `scheduling.Requirements` rather than `NodeSelectorRequirements`? Since we're directly injecting this into NodeClaim requirements that's the format they need to be in at the end, and `scheduling.Requirements` is our standard internal representation of requirements.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -130,7 +131,16 @@ func (n *NodeClaim) CanAdd(ctx context.Context, pod *corev1.Pod, podData *PodDat
 	}
 	nodeClaimRequirements.Add(podData.Requirements.Values()...)
 
+	// Add volume requirements to nodeClaimRequirements ONLY (not to pod's affinity).
+	// This ensures NodeClaim is created in the correct zone for volumes,
+	// while TSC counting uses pod's original affinity.
+	if err := addVolumeRequirements(nodeClaimRequirements, podData.VolumeRequirements, scheduling.AllowUndefinedWellKnownLabels); err != nil {
```
jmdeal, line 137: Shouldn't we also be enforcing WellKnownLabels for both existing NodeClaims and new NodeClaims? It should be safe to ignore them in practice since currently all well known label requirements are guaranteed to be populated for a new NodeClaim (at least in the Karpenter provider). I don't think this needs to hold as an invariant though, and if there are missing well known labels this could result in scheduling failures.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -447,3 +457,17 @@ func compatible(instanceType *cloudprovider.InstanceType, requirements schedulin
 func fits(instanceType *cloudprovider.InstanceType, requests corev1.ResourceList) bool {
 	return resources.Fits(requests, instanceType.Allocatable())
 }
+
+// addVolumeRequirements adds volume topology requirements to nodeRequirements after checking compatibility.
+// This catches cases like a PV with hostname affinity to a specific node that conflicts with the NodeClaim.
+func addVolumeRequirements(nodeRequirements scheduling.Requirements, volumeRequirements []corev1.NodeSelectorRequirement, opts ...option.Function[scheduling.CompatibilityOptions]) error {
```
jmdeal: If we do update both occurrences to enforce well known labels, let's drop the options from the function signature.

pkg/controllers/provisioning/scheduling/topology_test.go
```diff
@@ -3031,3 +3120,87 @@ var _ = Describe("Taints", func() {
 		Expect(node.Spec.Taints).To(HaveLen(1)) // Expect no taints generated beyond the default
 	})
 })
+
+// Test for TSC + Volume bug: Pod with PVC bound to a specific zone and TSC with minDomains
+// should schedule correctly. Previously, volume zone requirements polluted TSC calculations.
```
jmdeal, line 3125: Do you want to add a link to the original issue here? We could leave one to both your issue and the other issue you linked (https://github.com/kubernetes-sigs/karpenter/issues/1239) since I believe they are the same issue. I am pulling down your changes and testing with the reproduction steps there to verify.

pkg/controllers/provisioning/scheduling/topology_test.go
jmdeal: +1 to this, though I think referencing it at the top like you are is fine. It's good historical context.

pkg/controllers/provisioning/scheduling/nodeclaim.go
k24dizzle: Just to confirm my understanding: so for a new node claim, we can also assume that on this line, the node claim requirements already include all of the potential volume requirement labels (like `topology.kubernetes.io/zone`)? So we don't need the `scheduling.AllowUndefinedWellKnownLabels` option here and in general when we execute the `addVolumeRequirements` function?

pkg/controllers/provisioning/scheduling/volumetopology.go
k24dizzle: I think we attempt to "perform the injection into node requirements" for each existing node (and a potential new node claim) in `CanAdd`, which means if we move the log there, it might be a bit noisy vs. now. I went ahead and put the requirements in `WithValues` though.

pkg/controllers/provisioning/scheduling/nodeclaim.go
k24dizzle: When I remove the `scheduling.AllowUndefinedWellKnownLabels` here, I run into the following unit test failures:
```
Summarizing 3 Failures:
  [FAIL] Provisioning Volume Topology Requirements [It] should schedule to volume zones if volume already bound
  /Users/kzhao/go/src/github.com/k24dizzle/karpenter/pkg/controllers/provisioning/suite_test.go:1978
  [FAIL] Provisioning Volume Topology Requirements [It] should not relax an added volume topology zone node-selector away
  /Users/kzhao/go/src/github.com/k24dizzle/karpenter/pkg/controllers/provisioning/suite_test.go:2182
  [FAIL] Provisioning Volume Topology Requirements [It] should schedule to volume zones if volume already bound (ephemeral volume)
  /Users/kzhao/go/src/github.com/k24dizzle/karpenter/pkg/controllers/provisioning/suite_test.go:1999
```

jmdeal: It would be good to understand which requirements were missing that caused this to be unjustifiable, it could be a test issue.

pkg/controllers/provisioning/scheduling/volumetopology.go
jmdeal: Good point, sgtm

pkg/controllers/provisioning/scheduling/nodeclaim.go
jmdeal: Sorry, my original comment was incorrect. We don't currently inject the universe of possible domains into the NodeClaim, we rely on the instance types in `filterInstanceTypesByRequirements` which do have the universe of possible domains. Your current logic looks good - apologies for the churn.

(fwiw it looks like we don't do the stricter comparison in `filterInstanceTypesByRequirements` either, we probably should but it's a non-issue in our current state since all well-known labels must be defined)

pkg/controllers/provisioning/scheduling/nodeclaim.go
jmdeal: Determined this isn't necessary

jmdeal:
I've been able to do some additional testing on my side, these changes fix the original reproduction steps from #1239 so I marked this PR as fixing that as well. These changes LGTM though, thanks for the contribution!

/lgtm
/approve

clarifai-fmarceau:
@jmdeal @k24dizzle Thank you !!!

PR #2781, opened 2026-01-13
fix: deflake nomination test by using injected clock

ellistarn:
## Description

The nomination test was flaky because `Nominate()` and `Nominated()` used `time.Now()` directly while the test used `time.Sleep()` for real-time delays. In CI environments with timing jitter, the 21-second sleep could sometimes not exceed the 20-second nomination window.

This fix passes the clock as a parameter to `Nominate()` and `Nominated()` so nomination timing uses the test's fake clock instead of real time.

Note: We cannot store `clock.Clock` in `StateNode` because it has `+k8s:deepcopy-gen=true` and controller-gen cannot generate DeepCopy for interface types.

## Changes
- Update `Nominate(ctx, clk)` and `Nominated(clk)` to accept clock parameter
- Update `ValidateNodeDisruptable(clk)` to pass clock to `Nominated()`
- Update callers in `Cluster` to pass `c.clock`
- Replace `time.Sleep()` with `fakeClock.Step()` in test

## Testing

```
go test ./pkg/controllers/state/... -v
go test ./pkg/controllers/disruption/... -v
```

All tests pass. The nomination test now runs instantly instead of sleeping for 21 seconds.

### DIFF
--- a/pkg/controllers/disruption/types.go
+++ b/pkg/controllers/disruption/types.go
@@ -93,7 +93,7 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 	if queue.HasAny(node.ProviderID()) {
 		return nil, fmt.Errorf("candidate is already being disrupted")
 	}
-	if err = node.ValidateNodeDisruptable(); err != nil {
+	if err = node.ValidateNodeDisruptable(clk); err != nil {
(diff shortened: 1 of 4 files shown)


tzneal (approved):
lgtm

PR #2783, opened 2026-01-14
fix: deflake termination tests by using injected clock

ellistarn:
## Description

The termination tests were flaky because they used `time.Now()` to set the `NodeClaimTerminationTimestampAnnotationKey` while the controller uses the injected `fakeClock`. In CI environments with timing jitter, this mismatch could cause test failures.

This fix uses `fakeClock.Now()` for the annotation timestamps so the test timing is consistent with the controller's clock.

First observed: https://github.com/kubernetes-sigs/karpenter/actions/runs/20975167016/job/60288077617?pr=2781

## Changes

- Use `fakeClock.Now()` instead of `time.Now()` in two termination tests

## Testing

```
go test ./pkg/controllers/node/termination/... -v -count=100
```


tzneal (commented):
/lgtm

PR #2790, opened 2026-01-14
fix: Improve Karpenter memory when using Node Overlay with copy-on-write

### LINKED ISSUE #2655: Karpenter node overlay causing increased in memory usage

GnatorX:
**Fixes #2655**
Original PR: https://github.com/kubernetes-sigs/karpenter/pull/2675 (but i messed up the revert and the history of changes got way to messy)

## Problem

Full deep copying of instance types caused excessive memory usage in deployments with many instance types and node pools (~1GB+ for 200 types × 5 pools).

## Solution

Replace full deep copy with selective copying:
- **Share** immutable fields (Requirements, Overhead)
- **Copy** only modified fields (Capacity, affected Offerings)
- **New** `applyPriceOverlays()` for granular offering copying

Additional Fixes
- Move UnevaluatedNodePoolError out of nodeoverlay/store to cloudprovider/types to break circular dependency due to nodeoverlay/suite_test.go  importing test/expectations -> test/expectations  importing controllers/provisioning  -> controllers/provisioning  importing controllers/nodeoverlay. This allows nodeoverlay/suite_test.go to be under nodeoverlay package instead of nodeoverlay_test.

### Before
```go
overriddenInstanceType := it.DeepCopy()  // Always copies everything
```

### After
```go
overriddenInstanceType := &cloudprovider.InstanceType{
    Name:         it.Name,
    Requirements: it.Requirements, // Shared
    Overhead:     it.Overhead,     // Shared
}
// Only copy fields that need modification
```

## Results
Before: red line, After: Orange
Screenshot 2025-12-02 at 2 13 51 PM

Before:
```
      === BEFORE (Deep Copy) - Memory Tests ===

      TestMemoryUsage_OverlayScenarios:
        no_overlays:             Total Allocs: 71,514     | Bytes: 12.75 MB    | Heap: 3.97 MB
        price_overlays_only:     Total Allocs: 2,184,576  | Bytes: 164.56 MB   | Heap: 3.99 MB
        capacity_overlays_only:  Total Allocs: 1,140,409  | Bytes: 297.21 MB   | Heap: 3.99 MB
        mixed_overlays:          Total Allocs: 11,137,342 | Bytes: 1,108.82 MB | Heap: 2.76 MB

      TestMemoryUsage_ScaleWithNodePools:
        nodepools_1:   Total Allocs: 2,213,176   | Bytes: 1,271.41 MB  | Heap: 4.00 MB
        nodepools_5:   Total Allocs: 11,137,216  | Bytes: 2,083.00 MB  | Heap: 4.00 MB
        nodepools_10:  Total Allocs: 22,274,324  | Bytes: 3,716.96 MB  | Heap: 4.00 MB
        nodepools_20:  Total Allocs: 44,547,761  | Bytes: 7,028.17 MB  | Heap: 4.01 MB
        nodepools_50:  Total Allocs: 111,365,788 | Bytes: 15,697.76 MB | Heap: 4.01 MB

      TestMemoryUsage_ScaleWithInstanceTypes:
        instances_50:  Total Allocs: 3,893,618  | Bytes: 15,981.94 MB | Heap: 2.77 MB
        instances_100: Total Allocs: 7,788,094  | Bytes: 16,550.38 MB | Heap: 2.77 MB
        instances_200: Total Allocs: 11,137,399 | Bytes: 17,363.31 MB | Heap: 2.78 MB
        instances_500: Total Allocs: 11,137,410 | Bytes: 18,176.25 MB | Heap: 2.78 MB

      === BEFORE (Deep Copy) - Benchmarks ===

      BenchmarkStoreApply/selective-copy-apply    568,892 ns/op    943,506 B/op    12,475 allocs/op
      BenchmarkStoreApplyNoOverlay                  5,525 ns/op          0 B/op         0 allocs/op
      BenchmarkStoreApplyPriceOnly              1,297,943 ns/op  2,191,611 B/op    30,403 allocs/op
      BenchmarkStoreApplyCapacityOnly             818,722 ns/op  1,912,016 B/op    15,800 allocs/op
      BenchmarkStoreApplyBothOverlays           1,376,852 ns/op  2,329,316 B/op    30,803 allocs/op
      BenchmarkStoreApplyAllScenario            7,406,884 ns/op 12,321,060 B/op   162,776 allocs/op
      BenchmarkNodeOverlayControllerScenario       75,832 ns/op     86,400 B/op     1,080 allocs/op
```
After:

```
=== AFTER (Selective Copy-on-Write) - Memory Tests ===
      TestMemoryUsage_OverlayScenarios:
      No overlays: Total Allocs: 71509 | Bytes: 12.75 MB | Heap: 3.97 MB | Objects: 27144
      price_overlays_only: Total Allocs: 1129755 | Bytes: 44.64 MB | Heap: 3.98 MB | Objects: 27160
      capacity_overlays_only: Total Allocs: 71501 | Bytes: 64.87 MB | Heap: 3.98 MB | Objects: 27157
      mixed_overlays:  Total Allocs: 6006405 | Bytes: 323.89 MB | Heap: 2.75 MB | Objects: 15904
     
      TestMemoryUsage_ScaleWithNodePools:
      nodepools_1: Total Allocs: 1186975 | Bytes: 375.95 MB | Heap: 3.99 MB | Objects: 27192
      nodepools_5: Total Allocs: 6006364 | Bytes: 634.95 MB | Heap: 3.99 MB | Objects: 27216
      nodepools_10:  Total Allocs: 12012637 | Bytes: 1163.79 MB | Heap: 3.99 MB | Objects: 27245
      nodepools_20:  Total Allocs: 24025139 | Bytes: 2265.00 MB | Heap: 4.00 MB | Objects: 27289
      nodepools_50:  Total Allocs: 60061914 | Bytes: 5410.40 MB | Heap: 4.00 MB | Objects: 27292
     
      TestMemoryUsage_ScaleWithInstanceTypes:
      instances_50: Total Allocs: 2100180 | Bytes: 5501.46 MB | Heap: 2.76 MB | Objects: 16011
      instances_100: Total Allocs: 4200300 | Bytes: 5683.51 MB | Heap: 2.77 MB | Objects: 16035
      instances_200: Total Allocs: 6006410 | Bytes: 5943.84 MB | Heap: 2.76 MB | Objects: 15985
      instances_500: Total Allocs: 6006397 | Bytes: 6204.17 MB | Heap: 2.76 MB | Objects: 16007
    
    === AFTER (Selective Copy-on-Write) - Benchmarks ===
 
     BenchmarkStoreApply/selective-copy-apply 	   	273716 ns/op	  299578 B/op	    6723 allocs/op
     BenchmarkStoreApplyNoOverlay                5619 ns/op	       0 B/op	       0 allocs/op
     BenchmarkStoreApplyPriceOnly	       593105 ns/op	  493055 B/op	   16400 allocs/op
     BenchmarkStoreApplyCapacityOnly	  	102446 ns/op	  294401 B/op	    1000 allocs/op
     BenchmarkStoreApplyBothOverlays            	    693806 ns/op	  768427 B/op	   17201 allocs/op
     BenchmarkStoreApplyAllScenario            	       3802529 ns/op	 4127444 B/op	   91365 allocs/op
     BenchmarkNodeOverlayControllerScenario  	 	77965 ns/op	   86400 B/op	    1080 allocs/op

```

## Testing

```bash
make test-memory    # Run memory tests
make benchmark      # Run benchmarks
```

All tests pass. Comprehensive memory and benchmark tests added.

## Files

- `pkg/controllers/nodeoverlay/store.go` - Implementation
- `pkg/controllers/nodeoverlay/store_memory_test.go` - Memory tests
- `pkg/controllers/nodeoverlay/store_bench_test.go` - Benchmarks
- `Makefile` - New test targets

## Reviewers

Key areas:
1. Field sharing logic in `apply()` method
2. Selective copying in `applyPriceOverlays()`
3. Memory test validation

### DIFF
--- a/.github/actions/install-deps/action.yaml
+++ b/.github/actions/install-deps/action.yaml
@@ -7,7 +7,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
+    - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
       id: setup-go
       with:
         go-version-file: go.mod
@@ -16,7 +16,7 @@ runs:
     # Root path permission workaround for caching https://github.com/actions/cache/issues/845#issuecomment-1252594999
     - run: sudo chown "$USER" /usr/local
       shell: bash
-    - uses: actions/cache@9255dc7a253b0ccc959486e2bca901246202afeb # v5.0.1
+    - uses: actions/cache@8b402f58fbc84540c8b491a91e594a4576fec3d7 # v5.0.2
       id: cache-envtest
       with:
         path: /usr/local/kubebuilder/bin
--- a/.github/workflows/approval-comment.yaml
+++ b/.github/workflows/approval-comment.yaml
@@ -7,7 +7,7 @@ jobs:
     if: startsWith(github.event.review.body, '/karpenter perf')
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           fetch-depth: 0
       - name: Save info about the review comment as an artifact for other workflows that run on workflow_run to download them
@@ -19,7 +19,7 @@ jobs:
           mkdir -p /tmp/artifacts
           { echo "$REVIEW_BODY"; echo "$PULL_REQUEST_NUMBER"; echo "$COMMIT_ID"; } >> /tmp/artifacts/metadata.txt
           cat /tmp/artifacts/metadata.txt
-      - uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+      - uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
         with:
           name: artifacts
           path: /tmp/artifacts
--- a/.github/workflows/codeql-analysis.yaml
+++ b/.github/workflows/codeql-analysis.yaml
@@ -17,14 +17,14 @@ jobs:
       actions: read # github/codeql-action/init@v2
       security-events: write # github/codeql-action/init@v2
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
       - uses: ./.github/actions/install-deps
       - run: make vulncheck
-      - uses: github/codeql-action/init@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v4.31.4
+      - uses: github/codeql-action/init@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
         with:
           languages: go
-      - uses: github/codeql-action/autobuild@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v4.31.4
-      - uses: github/codeql-action/analyze@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v4.31.4
+      - uses: github/codeql-action/autobuild@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
+      - uses: github/codeql-action/analyze@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
   analyze-github-actions:
     name: Analyze Github Actions
     runs-on: ubuntu-latest
@@ -32,12 +32,12 @@ jobs:
       actions: read # github/codeql-action/init@v2
       security-events: write # github/codeql-action/init@v2
     steps:
-      - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
-      - uses: github/codeql-action/init@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v4.31.4
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+      - uses: github/codeql-action/init@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
         with:
           languages: actions
           config: |
             paths:
               - '.github/workflows'
               - '.github/actions'
-      - uses: github/codeql-action/analyze@e12f0178983d466f2f6028f5cc7a6d786fd97f4b # v4.31.4
+      - uses: github/codeql-action/analyze@cdefb33c0f6224e58673d9004f47f7cb3e328b89 # v4.31.10
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -28,9 +28,9 @@ jobs:
         echo "Event name: ${{ github.event_name }}"
         echo "Workflow run conclusion: ${{ github.event.workflow_run.conclusion }}"
         echo "Workflow run name: ${{ github.event.workflow_run.name }}"
-    - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
+    - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
     - name: Set up Python 3.10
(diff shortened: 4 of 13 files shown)


GnatorX:
@jmdeal sorry this is the new PR fix gingko changes..

GnatorX:
Note that I didn't move bench over because it needs the test framework and isn't really a test.

GnatorX:
PR should be good for review. I thought I could use AI to refactor but it made a mess of the old PR..

pkg/controllers/nodeoverlay/export_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+import (
+	corev1 "k8s.io/api/core/v1"
+
+	"sigs.k8s.io/karpenter/pkg/apis/v1alpha1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+)
+
+// Export internal types and functions for testing
+// This file is only compiled during tests (due to _test.go suffix)
```
jmdeal: Today I learned, that's good to know. That being said, is there any reason to not to just use the nodeoverlay package for your store test? I don't have a fundamental problem with this approach if it's needed, but it seems like we can avoid it. I thought that's how it was in your original PR, but with the history how it is I can't tell for sure.

GnatorX: I think with gingko (i could totally be wrong about this) you had to name the test as (bah_test) package which means you can't import the private structs/ functions from nodeoverlay.

GnatorX: let me try moving the test back to just `nodeoverlay` package

GnatorX: Ugh ok this is more complicated because nodeoverlay's suite_test has a circular dep hence why it is in package nodeoverlay_test.

nodeoverlay/suite_test.go (line 48)                           
 imports → test/expectations                                 
                     ▼                                           
  test/expectations/expectations.go (line 60)                   
     imports → controllers/provisioning      
                    ▼                                           
  provisioning/provisioner.go (was line ~50)                    
     imports → controllers/nodeoverlay
     
  Going to try to fix it..

GnatorX:
OK, that took a bit but fixed the tests (had to refactor something out the notes are in the PR description) and removed export_test.

pkg/controllers/nodeoverlay/export_test.go
GnatorX: Fixed. lmk if this works

jmdeal: Ah, I was thinking you could just change the store_test and work around the issue that way. Sorry you had to go down this refactoring path. To solve this long term we should probably ensure all types used in our expectations are used in top-level packages. No need to scope that in here though.

pkg/cloudprovider/types.go
```diff
@@ -583,3 +583,27 @@ func (e *CreateError) Error() string {
 func (e *CreateError) Unwrap() error {
 	return e.error
 }
+
+// UnevaluatedNodePoolError is an error when the node overlay controller has not updated the instance
```
jmdeal: nit: If we're going to make this a common error, we should remove any reference to node overlay. I'd update the error to be something like this:
```
applying nodeoverlays, nodepool %q is awaiting evaluation
```
The second part of this error would be in the common code, and we'd wrap it with "applying nodeoverlays" here: https://github.com/kubernetes-sigs/karpenter/blob/bc64ee6341b20eed08979c10c32736c90f92a0c1/pkg/cloudprovider/overlay/cloudprovider.go#L50.

GnatorX: Got it

GnatorX: Updated

PR #2794, opened 2026-01-15
fix: add check for deleted nodepools in cluster cost

### LINKED ISSUE #2793: Memory Leak In Cluster Cost
### Description

**Observed Behavior**:
As part of ClusterCost state, we [maintain a map of NodePools → NodePoolCost](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/state/cost/cost.go#L71). Currently, [we only remove NodePools from this map](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/state/cost/cost.go#L357) when a NodeClaim is deleted. 

In the case where NodePools are deleted before NodeClaims, we leak the following for the NodePool:

```
type NodePoolCost struct {
    cost            float64
    instanceTypeMap map[string]*cloudprovider.InstanceType // instance name -> instance
    offeringCounts map[OfferingKey]OfferingCount
}
```

This leaked memory can accumulate significantly over time depending on the providers caching layer.

**Expected Behavior**:
* Remove NodePools from this mapping when NodePool has been deleted

**Versions**:
- Chart Version:

ryan-mist:
Fixes #2793

**Description** 
* Fixes a memory leak in cluster cost mapping where deleted NodePools are never removed from the mapping (if they are deleted before NodeClaims for that NodePool)

* make presubmit

### DIFF
--- a/pkg/controllers/state/informer/pricing.go
+++ b/pkg/controllers/state/informer/pricing.go
@@ -63,6 +63,9 @@ func (c *PricingController) Reconcile(ctx context.Context) (reconciler.Result, e
 		return reconciler.Result{}, err
 	}
 
+	// Remove Deleted NodePools from Cluster Cost Mapping
+	c.clusterCost.CheckForDeletedNodePools(npl)
+
 	// Add a fake wide open nodepool to capture manually created nodeclaims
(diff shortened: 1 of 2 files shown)


jmdeal:
Do we still want to maintain the mapping until all NodeClaims for a NodePool have been deleted? I think we probably do since otherwise we'll see failures on updates until all NodeClaims are deleted.

PR #2807, opened 2026-01-20
fix: terminationGracePeriod not bypassing PDB when annotation is missing

### LINKED ISSUE #2802: Misconfigured PDB blocks draining node indefinitely
### Description

**Observed Behavior**: Karpenter cannot disrupt a node if there's a pod with PDB (minAvailable: 100%) despite having a NodeClaim with both expireAfter (720h) and terminationGracePeriod (1h).

**Expected Behavior**: Once the terminationGracePeriod elapses, remaining pods should be forcibly deleted and the underlying instance will be terminated. Source: https://karpenter.sh/docs/concepts/disruption/#terminationgraceperiod

1. Deploy a Pod with a PDB like the following:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-test-pdb-pod
  labels:
    app: nginx-test-pdb-pod
spec:
  nodeSelector:
    karpenter.sh/nodepool: test-huge-crowd-general
  tolerations:
  - key: taint.nr-ops.net/general-pool
    operator: Exists
    effect: NoSchedule
  containers:
  - name: nginx
    image: nginx:latest
    ports:
    - containerPort: 80
---
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: nginx-test-pdb
spec:
  minAvailable: 1
  selector:
    matchLabels:
      app: nginx-test-pdb-pod
```

2. Manually drain the node with `kubectl drain node <node> ...`

3. Observe that the node gets in unschedulable state and never gets deleted (even after the grace period of 1h has expired), causing a costs increase. (FWIW if the pod is deleted manually then Karpenter removes the node eventually).

```
│ Events:                                                                                                                                                                                                                       │
│   Type    Reason              Age                 From       Message                                                                                                                                                          │
│   ----    ------              ----                ----       -------                                                                                                                                                          │
│   Normal  Unconsolidatable    13m (x86 over 23h)  karpenter  Can't replace with a cheaper node                                                                                                                                │
│   Normal  DisruptionBlocked   26s                 karpenter  Pdb "default/nginx-test-pdb" prevents pod evictions
```

**Versions**:
- Chart Version: `1.3.2`
- Kubernetes Version (`kubectl version`): `v1.32.9-0`

moko-poi:
## What type of PR is this?
/kind bug

## What this PR does / why we need it:
Fixes #2802 

This PR fixes an issue where nodes remained stuck indefinitely when a misconfigured PodDisruptionBudget (PDB) blocked pod eviction, even after the `terminationGracePeriod` expired.

### Root Cause
The `nodeTerminationTime()` function returned `nil` when the `NodeClaimTerminationTimestampAnnotationKey` annotation was missing. This caused the forced eviction logic to never activate, leaving nodes blocked by PDB indefinitely.

### Solution
Modified `nodeTerminationTime()` to calculate the termination deadline from `NodeClaim.DeletionTimestamp + TerminationGracePeriod` when the annotation is not present, while still prioritizing the annotation value when available.

This ensures that `terminationGracePeriod` enforcement works consistently, allowing forced eviction to bypass PDB after the grace period expires.

## Which issue(s) this PR fixes:
Fixes #2802

## Does this PR introduce a user-facing change?
```release-note
Fixed an issue where terminationGracePeriod did not bypass PodDisruptionBudget when the termination timestamp annotation was not set, causing nodes to remain stuck indefinitely.
```

## Testing
- Added new test case: "should preemptively delete pods even without termination timestamp annotation (fallback calculation)"
- All existing tests pass (31/31)
- Verified forced eviction activates after grace period expires, even without annotation

### DIFF
--- a/pkg/controllers/node/termination/controller.go
+++ b/pkg/controllers/node/termination/controller.go
@@ -401,7 +401,7 @@ func (c *Controller) nodeTerminationTime(node *corev1.Node, nodeClaim *v1.NodeCl
 
 	// Fallback: compute the termination time from the NodeClaim's DeletionTimestamp and TerminationGracePeriod
 	// This ensures the terminationGracePeriod enforcement works even if the annotation is not yet set
-	terminationTime := nodeClaim.DeletionTimestamp.Time.Add(nodeClaim.Spec.TerminationGracePeriod.Duration)
+	terminationTime := nodeClaim.DeletionTimestamp.Add(nodeClaim.Spec.TerminationGracePeriod.Duration)
 	terminationTimeString := terminationTime.Format(time.RFC3339)
 	c.recorder.Publish(terminatorevents.NodeTerminationGracePeriodExpiring(node, terminationTimeString))
 	return &terminationTime, nil


jmdeal:
See my comment in the original issue: https://github.com/kubernetes-sigs/karpenter/issues/2802#issuecomment-3775505865. As far as I can tell this was a misunderstanding of Karpenter's drain logic and what initiates it, but let me know if you were encountering cases where this wasn't being added when a NodeClaim is deleted. We should dive into that if there are such cases.

jmdeal:
Are we good to close this PR, or was there a case you found where the annotation isn't persisted?

moko-poi:
pr closed

PR #72, opened 2022-11-16
feat: multi-node consolidation

tzneal:
Fixes https://github.com/aws/karpenter/issues/2510

**Description**

Consider consolidating multiple nodes simultaneously.  This has the benefit of reducing disruption in some cases and will sometimes replace multiple smaller nodes with a larger node, if the larger node is cheaper.  

**How was this change tested?**

Unit testing and deployed to EKS.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2812, opened 2026-01-21
refactor: simplify getCandidatePrices to not return error

ellistarn:
## Summary
- Remove error return from `getCandidatePrices` function
- When offerings are not found for a candidate, return 0 (free) instead of an error
- This is consistent with how reserved capacity was already handled

## Rationale
Returning 0 effectively blocks consolidation for that candidate (can't save money replacing something "free"), while drift handles nodes with mismatched offerings. This simplifies the API for callers.

## Test plan
- [x] Existing consolidation tests pass
- [x] Build succeeds

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/controllers/disruption/consolidation.go
+++ b/pkg/controllers/disruption/consolidation.go
@@ -23,7 +23,6 @@ import (
 	"sort"
 	"time"
 
-	"github.com/awslabs/operatorpkg/serrors"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/utils/clock"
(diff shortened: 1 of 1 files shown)


ellistarn:
Here's what I found:

  Original PR #72 (Nov 2022) - Multi-node consolidation

  The error was introduced without any recorded discussion. It was simply implemented as a defensive check - "if we can't find the offering, return an error."

  PR #1911 (Feb 2025) - Reserved capacity support

  This is the interesting one. When adding reserved capacity, they encountered the same issue and explicitly documented the rationale for returning 0:

  // It's expected that offerings may no longer exist for capacity reservations once a NodeClass stops selecting on
  // them (or they are no longer considered for some other reason on by the cloudprovider). By definition though,
  // reserved capacity is free. By modeling it as free, consolidation won't be able to succeed, but the node should be
  // disrupted via drift regardless.
  if reqs.Get(v1.CapacityTypeLabelKey).Has(v1.CapacityTypeReserved) {
      return 0.0, nil
  }

  The insight

  The author of PR #1911 recognized that:
  1. Offerings can disappear for legitimate reasons
  2. Returning 0 safely blocks consolidation
  3. Drift handles these nodes anyway

  But they only applied this logic to reserved capacity, leaving the error for other cases. The reasoning in their comment doesn't actually depend on the capacity type being reserved - the phrase "or they are no longer considered for some other reason by the cloudprovider" is generic.

  Bottom line: The error was a defensive default from 2022 that was never revisited. When reserved capacity was added, someone thought through the problem and found returning 0 was safe - but only applied it narrowly. Our change just generalizes their own documented reasoning.

Issue #2814, opened 2026-01-22
ConsolidationGroup for NodePool-aware Multi-node Consolidation


kimsehwan96:
### Description

**What problem are you trying to solve?**

In large-scale Kubernetes clusters with multiple NodePools separated by workload (using taints, nodeSelectors, or affinities), multi-node consolidation frequently fails with the message:

"Can't remove without creating N candidates"

Root Causes
1. Cross-NodePool Pod Scheduling Constraints: When multi-node consolidation candidates include nodes from different NodePools, and pods on those nodes can only be scheduled to their respective NodePools (due to taints/tolerations or nodeSelectors), the scheduling simulation requires multiple NewNodeClaims (one per NodePool). Since Karpenter only allows NewNodeClaims <= 1 for consolidation, this always fails.

2. Binary Search Algorithm Limitation: The current binary search in firstNConsolidationOption always evaluates candidates[0:mid+1], which are sorted by DisruptionCost across ALL NodePools. If candidates from different NodePools are interleaved in this range, consolidation will fail regardless of batch size reduction.

3. Pending Pods Interference: During consolidation evaluation and validation, pending pods from other NodePools can interfere with scheduling simulation, causing unnecessary failures.

**Proposed Solution: consolidationGroup Field**

Add a consolidationGroup field to the NodePool spec that allows cluster administrators to explicitly define which NodePools can be consolidated together.

API Change
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: ml-workers
spec:
  disruption:
    consolidationPolicy: WhenEmptyOrUnderutilized
    consolidateAfter: 5m
    consolidationGroup: "ml-workloads"  # New field
    budgets:
      - nodes: "25%"
```

| Field | Type | Default | Description |
|-------|------|---------|-------------|
| `consolidationGroup` | `string` | `"default"` | Identifier for grouping NodePools that can be consolidated together. Only nodes within the same consolidationGroup will be considered as candidates for multi-node consolidation. |

Expected Behavior Changes

1. **GetCandidates**: Group candidates by `consolidationGroup` instead of collecting all candidates globally.

2. **Multi-node Consolidation**: Execute consolidation evaluation **per consolidationGroup** rather than across all NodePools:
   ```go
   // Pseudocode
   for groupName, groupCandidates := range candidatesByGroup {
       cmd, err := m.firstNConsolidationOption(ctx, groupCandidates, maxParallel)
[...]

PR #2818, opened 2026-01-25
feat: add consolidation skip metrics with reason labels

kimsehwan96:
Fixes #N/A 

**Description**

In large-scale clusters, when consolidation doesn't occur as expected, operators typically need to piece together the root cause by correlating info/error logs. 

This PR adds metric-based observability to help identify consolidation skip reasons more efficiently. 

Added a new `karpenter_voluntary_disruption_consolidation_skipped_total` counter metric that tracks when consolidation is skipped in the computeConsolidation function. 

Each skip condition increments the metric with a specific reason label, and the metric distinguishes between single-node and multi-node consolidation attempts via a consolidation_type label. 

- Added 3 new test cases under ConsolidationSkippedTotal context that verify metric increments for representative skip reasons


jmdeal:
Could you give an example of a workflow where you would use this metric data to diagnose an issue? I have a hard time visualizing how this would be used. The main issue is the lack of context in these metrics. You don't know the node it applied to and you don't know the other nodes that were involved with the decision in the case of multi-node consolidation. This context is available with existing events, and I think that's the better solution since you don't have the cardinality concerns that you have with metrics.

kimsehwan96:
Thanks for the clarification. Here’s how I’d use the metric in practice

스크린샷 2026-02-09 오후 8 46 07
(It's my dashboard to track why consolidation(disruption) was filed)



Imagine a period where both multi‑node and single‑node consolidation are consistently not happening. One plausible root cause is that a pod needs to be scheduled, but no nodepool can provision a node that satisfies its constraints (e.g., taints/labels/requirements). 

Another example: in large clusters with high churn (pods constantly created/removed), consolidation decisions are affected by pending pods that can’t schedule onto the candidate node set. This frequently manifests as “multiple nodes needed.” A metric lets you correlate those time windows with consolidation stalls, even if you still need events/logs for root‑cause detail.

You can find this in logs, but if the metric tells you the skip reason (e.g.,PodNotSchedulable), you immediately see when consolidation was blocked and why, across time. In practice, alerting on metrics is also more common than building alerts on logs.

I agree the metric lacks context (no node/nodepool labels), and I also agree events provide better per‑decision detail. The issue is that events only exist for single‑node consolidation, and only when multi‑node consolidation fails and we fall back. That means we don’t have any visibility into multi‑node consolidation failures on their own.

In real clusters, “multiple nodes needed” is not rare — it’s common when candidate nodes host workloads that are strictly isolated via taints/constraints, which was the scenario in my issue: https://github.com/kubernetes-sigs/karpenter/issues/2814

Also, we already accept some coarse metrics without per‑node context (e.g., karpenter_voluntary_disruption_eligible_nodes has only a reason label). This proposed metric is similar in spirit: aggregate signal first, deeper context via events/logs

So while the metric is coarse, I still think it’s a useful aggregate signal to spot time windows where consolidation is blocked and to guide deeper investigation. If you still feel the signal is too weak without context, I’m OK with closing the PR!

Issue #2756, opened 2026-01-02
Mega Issue: Pod Disruption Controls


ellistarn:
### Description

There have been a number of discussions over the last year about how to more effectively make the cost/availability tradeoff in Karpenter. One of the recurring themes has been about providing application developers with more control over the cost of disruption. 

In the past, we've worked on [providing controls at the node level](https://github.com/kubernetes-sigs/karpenter/blob/main/designs/disruption-controls.md), but have relied more on upstream (i.e. PDBs) for pod level controls. This duality is deeply baked into the Kubernetes (and thus, Karpenter) operational model, where different user personas are able to take operational responsibility at their respective levels, and the system is capable of layering these constraints.

The goal of this issue is to centralize this discussion and identify a coherent technical approach to making progress on this problem space.

---

### kubernetes-sigs/karpenter

Issues:
- [#1719](https://github.com/kubernetes-sigs/karpenter/issues/1719) - Support more expressive Pod Disruption Controls
- [#1691](https://github.com/kubernetes-sigs/karpenter/issues/1691) - Karpenter does not respect reasons in budgets
- [#672](https://github.com/kubernetes-sigs/karpenter/issues/672) - Pod do-not-consolidate annotation

PRs/RFCs:
- [#2744](https://github.com/kubernetes-sigs/karpenter/pull/2744) - docs: Add Disruption Schedule RFC
- [#2745](https://github.com/kubernetes-sigs/karpenter/pull/2745) - feat: Add Pod Disruption Schedule

### aws/karpenter-provider-aws

Issues:
- [#3939](https://github.com/aws/karpenter-provider-aws/issues/3939) - Maintenance Window
- [#4422](https://github.com/aws/karpenter-provider-aws/issues/4422) - Add the ability to define how frequent Karpenter should consolidate

---

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

ellistarn:
[...]

Issue #2466, opened 2025-08-29
Karpenter ignores PDBs during drift resulting from a drifted AMI, preventing full node replacement


617m4rc:
### Description

**Observed Behavior**:

When Karpenter performs drift-based node disruption resulting from a drifted AMI, it does not consider PodDisruptionBudgets (PDBs). As a result, pods constrained by PDBs cannot be evicted, leaving drifted nodes in the cluster partially drained. Only a subset of workloads is moved to new nodes, preventing full replacement of drifted nodes.

**Expected Behavior**:

Karpenter should either provide a strategy to safely handle drift when PDBs prevent pod eviction (e.g., waiting for safe eviction, or alerting), or avoid starting consolidation of the drifted nodes until it can do so without violating PDB constraints.

**Reproduction Steps** (Please include YAML):

1. Create a Deployment or StatefulSet with multiple replicas.
2. Configure a PodDisruptionBudget (PDB) for the workload.
3. Launch nodes managed by Karpenter without pinning AMIs.
4. Introduce a new AMI that triggers Karpenter to attempt node replacement.
5. Observe that Karpenter is unable to fully drain the drifted nodes due to PDB constraints.

The observed behavior is specifically for drift resulting from a drifted AMI. We have not pinned AMIs and have not tested behavior for other kinds of drift.

**Versions**:
- Chart Version: v1.3.3
- Kubernetes Version (`kubectl version`): 1.32

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

engedaam:
> When Karpenter performs drift-based node disruption resulting from a drifted AMI, it does not consider PodDisruptionBudgets (PDBs). As a result, pods constrained by PDBs cannot be evicted, leaving drifted nodes in the cluster partially drained. Only a subset of workloads is moved to new nodes, preventing full replacement of drifted nodes.

[...]

PR #2827, opened 2026-01-29
feat(pdb): add pdb ignore annotation for edge cases

cenk1cenk2:
Hello dear maintainers,

I am opening up the following PR to partially remedy and add a new opt-in feature related to the following issues.

- re https://github.com/kubernetes-sigs/karpenter/issues/2756
- re https://github.com/kubernetes-sigs/karpenter/issues/1776
- re https://github.com/kubernetes-sigs/karpenter/issues/2466
- re https://github.com/kubernetes-sigs/karpenter/pull/2033
- re https://github.com/kubernetes-sigs/karpenter/issues/1928

**Description**

This PR aims to partially remedy the problem of unevictable PDBs preventing Karpenter from disrupting nodes.

The issue is as follows.
- Given a deployment with 1 replica and a PDB attached to it with effective `disruptionsAllowed` of 0.
- Karpenter never goes into the drift/eviction mode to further trigger the `terminationGracePeriod` and underlying mechanisms of `kubernetes` that actually prevent the rollover.
- Karpenter never considers the given node for consolidation since it does not meet the criteria.

The proposal for the feature is as follows.
- Add an annotation specifically for PDBs called `karpenter.sh/disruption-ignore`.
- Whenever the PDB discovery methodology sees this annotation, it will ignore it from the cases of `pdb.disruptionsAllowed == 0` when in zero disruptions mode and `pdb.isFullyBlocking` in fully blocking PDB case.

What this actually solves is as follows.
- Nodes successfully get marked as `Drifted`.
- The `terminationGracePeriod` starts successfully.
- Further operations, depending on the `v1/Node` `deletionTimestamp` can be processed.
- The `kubernetes` mechanisms for evaluating the disruptions on `PDB` will still block as intended.
- The node is also considered for consolidation to downsize, where it will be a worthy candidate.

This feature does not break any behavior and is completely opt-in.

- Through the automated tests.
- Real cluster in AWS with a patched version of `karpenter-provider-aws` patched with this version of `karpenter`.


DerekFrank:
Why do you want karpenter to try draining and disrupting the node when the PDB is blocking?

jmdeal:
> Why do you want karpenter to try draining and disrupting the node when the PDB is blocking?

There is a use-case for drift where the needs of application owners (availability) and the needs of cluster administrators (patching) are naturally at odds. TerminationGracePeriod should already address this though - when TGP is enabled drift will skip the blocking PDBs check and start the drain process for nodes regardless. The PDBs will then be respected until the node's TGP has elapsed, at which point the blocked pods will be forcibly removed.

I'm going to close out this PR since this use-case should already be addressed. If the use-case was slightly different, feel free to re-open and clarify. If there's an issue with the existing mechanism that you've identified, please open an issue.

cenk1cenk2:
Thank you @jmdeal, for your time to reply and go through this PR.

This case also covers a node as a consolidation target without waiting for the termination grace period, so I thought it could be complementry. In cases that you want no interruption, a custom operator listening for the node events can move the workloads gracefully to another node by stamping a label to roll the pods over when the node is marked as drifted. The node will be stuck until the pods can be removed, and the eviction will fail because of the PDB so that is where the termination grace period will actually kick in. Since the drift process has started with marked PDBs being non-blocking, this could be the mark for that given operator. Currently, if there is a PDB on the actual node, it is not considered as a target to drain at all, as far as I have tested.

But I would agree that this is a novel use case (or abuse) of `karpenter`.

jmdeal:
I don't think I really understand your use-case, could you give an example of how you would want to use this with consolidation?

> Currently, if there is a PDB on the actual node, it is not considered as a target to drain at all, as far as I have tested.

Just to clarify, you mean a PDB against pods on the node? The node can't have a PDB itself. As long as `terminationGracePeriod` is configured on the node, we'll drain the node. We just won't evict the pods until `terminationGracePeriod` elapses or the PDBs become non-blocking.

PR #2840, opened 2026-02-05
fix: prevent static capacity controllers from modifying NodeClaims during NodePool deletion

tallaxes:
**Description**

This PR fixes a race condition where static capacity controllers could create/delete NodeClaims while a NodePool is being deleted. 

It also makes static capacity E2E test preserve all settings from provider-specific NodePool (such as labels and startupTaints, if any), and wraps a small test section in Eventually to account for variable timing.

#### Controller Changes (the only production change; all other changes are in tests only)
- **Provisioning & Deprovisioning controllers**: Added early return when NodePool has a deletion timestamp set, preventing the controllers from creating new NodeClaims or deleting existing ones during NodePool deletion. Note that deletion should already be handled elsewhere, "should handle NodePool deletion gracefully" covers this scenario, and passes.

#### Unit Test Changes (`pkg/controllers/static/*/suite_test.go`)
- Added "should return early if nodepool is being deleted" test to both provisioning and deprovisioning controllers
- Verifies controllers return early with zero requeue and don't modify NodeClaims when NodePool is being deleted

#### Integration Test Changes (`test/suites/regression/staticcapacity_test.go`)
- **NodePool deletion test**: Added `ConsistentlyExpectNodeClaimCountNotExceed(30*time.Second, 3)` to verify no new NodeClaims are created during deletion

#### Integration Test Fixes (`test/suites/regression/staticcapacity_test.go`)
- **Dynamic NodeClaim Interaction**: Restructured `BeforeEach` to copy `nodePool` before applying static-specific modifications, ensuring `dynamicNodePool` inherits base properties - including Labels and StartupTaints, which were lost in the original. 
- **Edge Cases**: Wrapped annotation verification in `Eventually` with 10-minute timeout to handle slower VM provisioning on Azure

- [x] Unit tests pass for static provisioning controller
- [x] Unit tests pass for static deprovisioning controller
- [x] Integration tests pass (KWOK and Azure)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/karpenter
 
-go 1.25.6
+go 1.25.7
 
 require (
 	github.com/Pallinder/go-randomdata v1.2.0


test/suites/regression/staticcapacity_test.go
```diff
@@ -344,26 +342,21 @@ var _ = Describe("StaticCapacity", func() {
...
-			// Create a dynamic NodePool
-			dynamicNodePool = test.NodePool(v1.NodePool{
-				ObjectMeta: metav1.ObjectMeta{
-					Name: "dynamic-nodepool",
-				},
-				Spec: v1.NodePoolSpec{
-					Template: v1.NodeClaimTemplate{
-						Spec: v1.NodeClaimTemplateSpec{
-							Requirements: nodePool.Spec.Template.Spec.Requirements,
-							NodeClassRef: nodePool.Spec.Template.Spec.NodeClassRef,
```
tallaxes, line 362: This was copying requirements from the provider-specific NodePool, but was losing everything else. (For example, Azure provider has some labels and startupTaints in the test NodePool.)

tallaxes:
Vulnerabilities flagged here and in other PRs addressed by https://github.com/kubernetes-sigs/karpenter/pull/2844

pkg/controllers/static/deprovisioning/controller.go
```diff
@@ -81,7 +81,7 @@ func (c *Controller) Name() string {
 func (c *Controller) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.Result, error) {
 	ctx = injection.WithControllerName(ctx, c.Name())
 
-	if !nodepoolutils.IsManaged(np, c.cloudProvider) || np.Spec.Replicas == nil {
+	if !nodepoolutils.IsManaged(np, c.cloudProvider) || np.Spec.Replicas == nil || !np.DeletionTimestamp.IsZero() {
```
DerekFrank, line 84: I buy this for provisioning, but do we want this for deprovisioning as well? I guess we're saying that the deprovisioning will be handled by nodepool deletion instead?

tallaxes: Yes, I asked myself the same question, and arrived at the same conclusion - deletion should already be handled. Note that "should handle NodePool deletion gracefully" E2E (which has a minor unrelated tweak in this PR) already coveres this, and passes. (Added a note to PR description)

PR #2842, opened 2026-02-05
feat: Add annotation-based priority control for static NodePool deprovisioning

### LINKED ISSUE #2841: Allow specific nodes to be removed while decreasing nodepool replicas in static mode
### Description

We have a custom scheduler we use in an AI/ML platform that is mapping a set of pods to a node. Based on some internal quota allocation per project we want to scale the nodepool replicas based on quota. When the pods are done processing their job and the node is freed up, we want to remove it from the cluster. For that we need to be able to remove this particular node while decreasing the spec.replicas in the nodepool. 

We tried draining, cordoning and tainting the node, but karpenter doesn’t seem to take these into consideration while picking the node when replicas is decreased and might pick another node from the nodepool to delete.

For example, I drained, cordoned and tainted 2 out of 4 nodes from a static nodepool and then decreased the replicas from 4 to 2. The behaviour was unpredictable since karpenter removed 1 cordoned node and 1 “healthy” node I didn’t touch. I redid the same exercise few times and everytime the nodes picked for removal were not consistent, sometime it picked 2 not cordoned/tainted, other time 1 cordoned and 1 not cordoned, etc

Currently the scale down prioritizes these NodeClaims ([ref](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/static/deprovisioning/controller.go#L178)):
 1. Unresolved NodeClaims (no ProviderID yet - haven't launched)
 2. Empty nodes (nodes with no pods or only DaemonSet pods without do-not-disrupt annotation)
 3. If more nodes needed, nodes with lowest disruption cost (nodes with pods that have do-not-disrupt will have highest cost)

There isn't currently a way to specifically remove certain nodes during scale down operation.

It is pretty important for us since it is blocking us adopting karpenter.

adriananeci:
Fixes #2841

**Description**

This PR adds support for the `karpenter.sh/deprovisioning-priority` annotation on NodeClaims, allowing users to explicitly control which nodes are removed first when scaling down static NodePools.

When scaling down static NodePools, users need control over which specific nodes are removed. Current behavior selects nodes based on internal heuristics (empty nodes, disruption cost), but doesn't allow explicit targeting. This is problematic when:

  1. Specific nodes are manually cordoned/drained and should be removed
  2. Certain nodes have known issues or degraded performance
  3. Users want to implement custom deprovisioning strategies
  4. Nodes need to be removed in a specific order during maintenance

Example scenarios from issue #2841:
1. No manual nodeclaims deletion
  - Initial state: replicas=4, NodeClaims=[A, B, C, D], nodes don't have do-not-disrupt pods
  - User actions:
    - Drain, cordon and taint A, B manually
    - Scale down replicas=2
 
  - Expected:
    - Remove A, B nodes, keep C, D.
  - Reality:
    - Karpenter: Decides to remove C, D (based on disruption selection)
  
2. Manual nodeclaims deletion
  - Initial state: replicas=4, NodeClaims=[A, B, C, D], nodes don't have do-not-disrupt pods
  - User actions:
    - Drain, cordon and taint A, B manually
    - Delete NodeClaims A, B manually
    - Scale down replicas=2
 
  - Expected:
    - Remove A, B nodes, keep C, D.
  - Reality:
    - Karpenter: Creates replacements E, F (sees replicas=4, actual=2) between nodeclaims deletion and replicas scale down to 2
    - Karpenter: Remove C, D (based on disruption cost heuristics)
    - nodepool ends up with new nodes(e, F), disrupting more than expected nodes.
 

This PR introduces `karpenter.sh/deprovisioning-priority` annotation with integer values (higher = deprovisioned first, default = 0). Priority is applied across all candidate selection tiers while maintaining safety guarantees.

Deprovisioning priority hierarchy:
  1. Do-not-disrupt status (always first - protects critical workloads)
  2. Priority annotation (higher values removed first)
  3. Disruption cost (existing heuristic as tiebreaker)

This hierarchy ensures safety-critical workloads remain protected while giving users fine-grained control over deprovisioning order.

Updated the docs from `designs/static-capacity.md` docs too

Usage Example
```
  # Annotate NodeClaims to prioritize for removal (higher = removed first)
  kubectl annotate nodeclaim unwanted-node-1 karpenter.sh/deprovisioning-priority=100
  kubectl annotate nodeclaim unwanted-node-2 karpenter.sh/deprovisioning-priority=90

  # Scale down - annotated nodes removed first
  kubectl scale nodepool my-pool --replicas=5
```
 
Use Cases
  1. Targeted removal: Mark specific nodes for removal when scaling down
  2. Manual intervention: Remove nodes that were cordoned/drained
  3. Rolling updates: Control replacement order during updates
  4. Zone rebalancing: Influence which zones lose capacity first

These changes should be fully backward compatibile since nodes without annotation default to priority 0
Also, existing deprovisioning behavior unchanged when annotation not used and invalid annotation values gracefully default to 0

Annotation based node priority deprovisioning inspiration came from https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/#pod-deletion-cost

  New test coverage includes:
  - Priority sorting for unresolved NodeClaims
  - Priority sorting for empty nodes
  - Do-not-disrupt takes precedence over priority
  - Priority used as tiebreaker for non-empty nodes


adriananeci:
Note for the reviewers: initial changes were minimal, but because they introduced [a cyclomatic complexity](https://github.com/kubernetes-sigs/karpenter/actions/runs/21710734906/job/62613357684) issue I had to refactor the code a bit.

ellistarn:
From a naming perspective, do we think that https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/#pod-deletion-cost is an appropriate analog? cc: @jamesmt-aws. I think we may want to support something like this for dynamic provisioning as well, to provide customers with a way to give Karpenter hints about their preferred consolidation choices.

jmdeal:
We had an offline discussion regarding this PR yesterday, and I wanted to summarize the outcomes of that discussion. We think that we can achieve the desired behavior without introducing a new annotation, and instead leveraging the node-level `karpenter.sh/do-not-disrupt` annotation. In this model, Karpenter would prioritize disrupting nodes without the annotation, but will still disrupt nodes with the annotation if that's the only option. This is consistent with the annotation only serving as a blocker for graceful disruption mechanisms (like consolidation and drift) as opposed to forceful disruption mechanisms (like expiration or static deprovisioning). The worflow would look like this:
- Your NodePool would include the `karpenter.sh/do-not-disrupt: true` annotation on the NodeClaim template (`spec.template.metadata.annotations`), ensuring all created Nodes have this annotation by default.
- If no prioritization is needed, reduce the replica count and Karpenter will scale down with existing heuristics
- If an external orchestrator needs to influence prioritization, it can remove the do-not-disrupt annotation from specific nodes before reducing the replica count.

Based on our previous conversation, it sounds like this will meet your use-case but let me know if you've realized otherwise.

There are still some open questions:
- Should the annotation block deprovisioning outright, or only serve as a deprioritization mechanism?
  - I believe it's consistent to make it only a deprioritization mechanism.
- Which node would have priority? A node with the annotation but no pods, or a node without but with pods with the annotation?
  - I believe prioritizing the latter is the correct answer for your use-case, but I want to make sure it's the right choice of behavior for this annotation. I don't think either is a clearly correct choice, so we just need to choose one, but I welcome other opinions.

jmdeal:
I'm going to close this out as stale since there haven't been any updates since our last discussion, if you'd like to continue the PR feel free to re-open.

/close

PR #2847, opened 2026-02-10
feat: add a nodepool cost metric 2

DerekFrank:
Fixes #N/A 

**Description**

Second attempt at: https://github.com/kubernetes-sigs/karpenter/pull/2584

Updates:

* Now instead of storing instance types or offerings, we store just the relevant information. Notably this avoids storing any scheduling.Requirements, which is a significant memory sink. 
* Includes specific logic to clean up the cluster cost data when a nodepool has been deleted.

Ran some fuzz testing, scaling up deployments and scaling them down again, created nodepools, scaled them up, then deleted nodepools. Confirmed that the controller does not panic nor does the memory leak

### DIFF
--- a/pkg/controllers/metrics/nodepool/controller.go
+++ b/pkg/controllers/metrics/nodepool/controller.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	"sigs.k8s.io/karpenter/pkg/state/cost"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
 )
 
@@ -67,20 +68,32 @@ var (
 			metrics.NodePoolLabel,
 		},
 	)
+	ClusterCost = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodePoolSubsystem,
+			Name:      "cost_total",
+			Help:      "Total cost of the nodepool from Karpenter's perspective. Units are determined by the cloud provider. Not an authoritative source for billing. Includes modifications due to NodeOverlays",
+		},
+		[]string{metrics.NodePoolLabel},
+	)
 )
 
 type Controller struct {
 	kubeClient    client.Client
 	cloudProvider cloudprovider.CloudProvider
+	clusterCost   *cost.ClusterCost
 	metricStore   *metrics.Store
 }
 
 // NewController constructs a controller instance
-func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider) *Controller {
+func NewController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, clusterCost *cost.ClusterCost) *Controller {
 	return &Controller{
 		kubeClient:    kubeClient,
 		cloudProvider: cloudProvider,
 		metricStore:   metrics.NewStore(),
+		clusterCost:   clusterCost,
 	}
 }
 
@@ -98,12 +111,18 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if !nodepoolutils.IsManaged(nodePool, c.cloudProvider) {
 		return reconcile.Result{}, nil
 	}
-	c.metricStore.Update(req.String(), buildMetrics(nodePool))
+	c.metricStore.Update(req.String(), c.buildMetrics(nodePool))
 	// periodically update our metrics per nodepool even if nothing has changed
-	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+	return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
 }
 
-func buildMetrics(nodePool *v1.NodePool) (res []*metrics.StoreMetric) {
+func (c *Controller) buildMetrics(nodePool *v1.NodePool) (res []*metrics.StoreMetric) {
+	res = append(res, &metrics.StoreMetric{
+		GaugeMetric: ClusterCost,
+		Labels:      map[string]string{metrics.NodePoolLabel: nodePool.Name},
+		Value:       c.clusterCost.GetNodepoolCost(nodePool),
+	})
+
 	for gaugeVec, resourceList := range map[opmetrics.GaugeMetric]corev1.ResourceList{
 		Usage: nodePool.Status.Resources,
 		Limit: getLimits(nodePool),
--- a/pkg/controllers/state/informer/nodepool.go
+++ b/pkg/controllers/state/informer/nodepool.go
@@ -19,11 +19,11 @@ package informer
 import (
 	"context"
 
+	"k8s.io/apimachinery/pkg/api/errors"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -32,6 +32,7 @@ import (
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/controllers/state"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	"sigs.k8s.io/karpenter/pkg/state/cost"
 	utilscontroller "sigs.k8s.io/karpenter/pkg/utils/controller"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
 )
@@ -41,13 +42,15 @@ type NodePoolController struct {
 	kubeClient    client.Client
 	cloudProvider cloudprovider.CloudProvider
 	cluster       *state.Cluster
+	clusterCost   *cost.ClusterCost
 }
 
-func NewNodePoolController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, cluster *state.Cluster) *NodePoolController {
+func NewNodePoolController(kubeClient client.Client, cloudProvider cloudprovider.CloudProvider, cluster *state.Cluster, clusterCost *cost.ClusterCost) *NodePoolController {
 	return &NodePoolController{
 		kubeClient:    kubeClient,
 		cloudProvider: cloudProvider,
 		cluster:       cluster,
+		clusterCost:   clusterCost,
 	}
 }
 
@@ -55,8 +58,18 @@ func (c *NodePoolController) Name() string {
 	return "state.nodepool"
 }
 
-func (c *NodePoolController) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.Result, error) {
+func (c *NodePoolController) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	ctx = injection.WithControllerName(ctx, c.Name()) //nolint:ineffassign,staticcheck
+
+	np := &v1.NodePool{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, np); err != nil {
+		if errors.IsNotFound(err) {
+			// notify cluster state of the nodepool deletion
+			c.clusterCost.DeleteNodePool(ctx, req.Name)
+			return reconcile.Result{}, nil
+		}
+		return reconcile.Result{}, client.IgnoreNotFound(err)
+	}
 	if !nodepoolutils.IsManaged(np, c.cloudProvider) {
 		return reconcile.Result{}, nil
 	}
@@ -72,6 +85,5 @@ func (c *NodePoolController) Register(ctx context.Context, m manager.Manager) er
 		For(&v1.NodePool{}, builder.WithPredicates(nodepoolutils.IsManagedPredicateFuncs(c.cloudProvider))).
 		WithOptions(controller.Options{MaxConcurrentReconciles: utilscontroller.LinearScaleReconciles(utilscontroller.CPUCount(ctx), minReconciles, maxReconciles)}).
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
-		WithEventFilter(predicate.Funcs{DeleteFunc: func(event event.DeleteEvent) bool { return false }}).
-		Complete(reconcile.AsReconciler(m.GetClient(), c))
+		Complete(c)
 }
--- a/pkg/controllers/state/informer/pricing.go
+++ b/pkg/controllers/state/informer/pricing.go
@@ -1,0 +1,122 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package informer
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/awslabs/operatorpkg/reconciler"
+	"github.com/awslabs/operatorpkg/singleton"
+	"github.com/samber/lo"
+	"go.uber.org/multierr"
+	"k8s.io/apimachinery/pkg/types"
+
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/state/cost"
+)
+
+// This whole pricing controller only exists because CP's surface InstanceType information
+// via a poll-method. Long term, this controller should likely be removed in favor of refactoring the
+// cloudprovider interface to a push model for instance type information:
+// https://github.com/kubernetes-sigs/karpenter/issues/2605
+type PricingController struct {
+	client        client.Client
+	cloudProvider cloudprovider.CloudProvider
+	clusterCost   *cost.ClusterCost
+	npOfMap       map[types.NamespacedName]map[cost.OfferingKey]float64
+}
+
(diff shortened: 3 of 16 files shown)


pkg/state/cost/cost.go
```diff
@@ -0,0 +1,392 @@
...
+			} else {
+				return fmt.Errorf("offering not found for instance type %q in nodepool %q after retry attempt (zone, %q, capacity, %q)", offeringKey.InstanceName, np.Name, offeringKey.Zone, offeringKey.CapacityType)
+			}
+		}
+		oc = OfferingCount{
+			Count: 1,
+			Cost:  price,
+		}
+	} else {
+		oc.Count += 1
```
ryan-mist, line 319: Should we call `internalNodepoolUpdate` even if the offering does already exist in `offeringCounts` to ensure we have the updated pricing?

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,392 @@
...
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/metrics"
+)
+
+// NecessaryLabels defines the set of required Kubernetes labels that must be present
+// on NodeClaim objects for cost tracking to function properly.
+var NecessaryLabels = []string{corev1.LabelInstanceTypeStable, v1.CapacityTypeLabelKey, corev1.LabelTopologyZone, v1.NodePoolLabelKey}
+
+var (
+	CostTrackingErrorsTotal = opmetrics.NewPrometheusCounter(
```
ryan-mist, line 45: Does controller runtime metrics for reconcile errors cover the use case for having this metric?

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,392 @@
...
+
+	cloudProvider cloudprovider.CloudProvider
+	client        client.Client
+}
+
+// NodePoolCost represents the cost tracking information for a single NodePool.
+// It maintains the current cost, available instance types, and count of active offerings.
+type NodePoolCost struct {
+	cost         float64
+	offeringsMap map[OfferingKey]float64 // OfferingKey -> Price
```
ryan-mist: Is there any difference between the price in `offeringsMap` and the cost field in `offeringCounts`?

ryan-mist: Looks like `internalUpdateOfferings(..)` only updates the price in `offeringsMap` - does `offeringCounts` contain the historical cost while `offeringsMap` contains the updated one?

pkg/controllers/state/informer/nodepool.go
```diff
@@ -41,22 +42,34 @@ type NodePoolController struct {
...
 }
 
-func (c *NodePoolController) Reconcile(ctx context.Context, np *v1.NodePool) (reconcile.Result, error) {
+func (c *NodePoolController) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	ctx = injection.WithControllerName(ctx, c.Name()) //nolint:ineffassign,staticcheck
+
+	np := &v1.NodePool{}
+	if err := c.kubeClient.Get(ctx, req.NamespacedName, np); err != nil {
+		if errors.IsNotFound(err) {
+			// notify cluster state of the node deletion
```
ryan-mist: nit: should be nodepool deletion

pkg/state/cost/cost.go
DerekFrank: The errors would be split across like 3 different reconcilers. It might be nice to have something that just tracks clusterCost errors

pkg/state/cost/cost.go
DerekFrank: > Is there any difference between the price in offeringsMap and the cost field in offeringCounts?

Nope, its a good catch, I just missed it in my refactor :)

pkg/state/cost/cost.go
ryan-mist: Nevermind - I'm guessing the idea is to have the pricing updated every hour by the pricing controller and we're fine if the current pricing isn't 100% accurate?

DerekFrank: Yep thats the idea!

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+func (cc *ClusterCost) internalUpdateOfferings(np *v1.NodePool, instanceTypes []*cloudprovider.InstanceType) {
+	npCost, exists := cc.npCostMap[np.Name]
+
+	if !exists {
+		cc.createNewNodePoolCost(np.Name, instanceTypes)
+	} else {
+		instanceTypes = lo.Filter(instanceTypes, func(it *cloudprovider.InstanceType, _ int) bool {
+			return it != nil
```
ryan-mist, line 144: Do we expect instance types to ever be nil?

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,376 @@
...
+
+	oc, exists := cc.npCostMap[npName].offeringCounts[offeringKey]
+	if !exists {
+		// our offerings must be out of date, we should update and retry
+		if firstTry {
+			err := cc.internalNodepoolUpdate(ctx, np)
+			if err != nil {
+				return fmt.Errorf("failed to update nodepool %q during retry while searching for offering for instance %q in zone %q with capacity %q, %w", np.Name, offeringKey.InstanceName, offeringKey.Zone, offeringKey.CapacityType, err)
+			}
+			return cc.internalAddOffering(ctx, npName, offeringKey, false)
```
ryan-mist: nit: Instead of having a retry variable and making this function recursive, it might be a bit cleaner to just update `oc` and `exists` variables here, then have another check for if the offering exists

pkg/controllers/metrics/nodepool/controller.go
```diff
@@ -98,12 +111,18 @@ func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	if !nodepoolutils.IsManaged(nodePool, c.cloudProvider) {
 		return reconcile.Result{}, nil
 	}
-	c.metricStore.Update(req.String(), buildMetrics(nodePool))
+	c.metricStore.Update(req.String(), c.buildMetrics(nodePool))
 	// periodically update our metrics per nodepool even if nothing has changed
-	return reconcile.Result{RequeueAfter: 5 * time.Minute}, nil
+	return reconcile.Result{RequeueAfter: 10 * time.Second}, nil
```
ryan-mist, line 116: Are we increasing this because there's more metrics now?

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,126 @@
...
+
+// This whole pricing controller only exists because CP's surface InstanceType information
+// via a poll-method. Long term, this controller should likely be removed in favor of refactoring the
+// cloudprovider interface to a push model for instance type information:
+// https://github.com/kubernetes-sigs/karpenter/issues/2605
+type PricingController struct {
+	client        client.Client
+	cloudProvider cloudprovider.CloudProvider
+	clusterCost   *cost.ClusterCost
+	npItMap       map[string]map[string]*cloudprovider.InstanceType
```
ryan-mist: Should we just store the necessary instance type info needed to compare the old and new offerings?
`cloudprovider.InstanceType` internally contains the `scheduling.Requirements` which was where we saw the memory growth last time - although it probably isn't as much of an issue this time because deleted NodePools aren't sticking around in this map

pkg/controllers/metrics/nodepool/controller.go
DerekFrank: We're increasing this because now we need to catch nodeclaim additions and deletions. This only triggers when a nodepool is updated.

DerekFrank: Otherwise the cost metrics don't change for 5 minutes or a nodepool update

pkg/controllers/state/informer/pricing.go
DerekFrank: We store them in the node overlay CP as well, but you're right that this does increase the memory more than necessary. I could use the OfferingKey type here again, or make a new Scheduling-less Offering object.

Problem is we have copied all of the scheduling requirements already when we called GetInstanceTypes() so the memory has been allocated.

pkg/state/cost/cost.go
DerekFrank: They have been before in unit testing. Can't hurt to filter them out just in case

pkg/controllers/metrics/nodepool/controller.go
ryan-mist: Oops, I read it originally as 5 seconds -> 10 seconds not 5 minutes -> 10 seconds

pkg/controllers/state/informer/pricing.go
DerekFrank: I still did the refactor just in case its less memory intensive

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func (c *PricingController) Reconcile(ctx context.Context) (reconciler.Result, error) {
+	npl := &v1.NodePoolList{}
+	err := c.client.List(ctx, npl)
+	if err != nil {
+		return reconciler.Result{}, err
+	}
+
+	newNpOfMap := make(map[types.NamespacedName]map[cost.OfferingKey]float64)
```
jmdeal, line 65: We should probably use UID as opposed to NamespacedName since you could create a new NodePool with the same namespaced name and have a false collision.

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,122 @@
...
+		newIts, err := c.cloudProvider.GetInstanceTypes(ctx, &np)
+		if err != nil {
+			errs = multierr.Append(errs, err)
+			continue
+		}
+
+		newNpOfMap[client.ObjectKeyFromObject(&np)] = make(map[cost.OfferingKey]float64)
+
+		lo.ForEach(newIts, func(it *cloudprovider.InstanceType, _ int) {
+			lo.ForEach(it.Offerings, func(o *cloudprovider.Offering, _ int) {
```
jmdeal: nit: I'd just use standard for loops rather than `lo.ForEach`.

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,122 @@
...
+				offeringKey := cost.OfferingKey{InstanceName: it.Name, Zone: o.Zone(), CapacityType: o.CapacityType()}
```
jmdeal, line 79: We might want to leave a note that this isn't a hard invariant. I think it does hold true for all current cloudproviders, but you could have multiple offerings that resolve to this same offering key but have a different price. I think it's fine to do it this way for now, it might just be a good idea to leave a note for the future.

pkg/controllers/metrics/nodepool/controller.go
jmdeal: Do we want to add a watch for NodeClaim creations and deletions for this controller then rather than increasing the requeue frequency?

pkg/controllers/state/informer/pricing.go
```diff
@@ -0,0 +1,122 @@
...
+func equal(oldOfs map[cost.OfferingKey]float64, newOfs map[cost.OfferingKey]float64) bool {
+	if len(lo.Values(oldOfs)) != len(newOfs) {
+		return false
+	}
+	for newOf, newPrice := range newOfs {
+		oldPrice, exists := oldOfs[newOf]
+		if !exists {
+			return false
+		}
+		if oldPrice != newPrice {
```
jmdeal, line 106: ~~Just flagging this since floating point equality checks are typically not a good idea, but I think it's fine here since they're always sourced from the same computation which should always result in the same binary representation.~~

Actually, since we're just trying to detect cache invalidation in practice this seems fine. I'll just leave this comment here for posterity and for a sanity check.

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,375 @@
...
+// - Tracking NodeClaim additions and removals
+// - Managing instance type offerings and their prices
+// - Calculating aggregate costs per NodePool and cluster-wide
+//
+// All operations are thread-safe through internal locking mechanisms.
+type ClusterCost struct {
+	sync.RWMutex
+	npCostMap map[string]*NodePoolCost // nodepool.Name -> NodePoolCost
+	// nodeClaimSet tracks which NodeClaims are currently being monitored for cost
+	nodeClaimMap map[types.NamespacedName]NodeClaimMetaData // nodeClaim object key -> NodeClaimMetaData
```
jmdeal, line 72: Same comment here about using the UID instead of the NamespacedName

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+// OfferingCount tracks the number and cost of instances for a specific offering.
+type OfferingCount struct {
+	Count int
+	Price float64 // Price of the offering, not Price * count
+}
+
+type NodeClaimMetaData struct {
+	NodePoolName string
```
jmdeal, line 100: Same comment here about using the UID instead of the Name

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,375 @@
...
+
+	if !exists {
+		cc.createNewNodePoolCost(np.Name, instanceTypes)
+	} else {
+		instanceTypes = lo.Filter(instanceTypes, func(it *cloudprovider.InstanceType, _ int) bool {
+			return it != nil
+		})
+		newMap := map[OfferingKey]OfferingCount{}
+		lo.ForEach(instanceTypes, func(it *cloudprovider.InstanceType, _ int) {
+			lo.ForEach(it.Offerings, func(o *cloudprovider.Offering, _ int) {
```
jmdeal: Same comment about `lo.ForEach` - I'd favor a standard for loop.

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,375 @@
...
+				offeringKey := OfferingKey{InstanceName: it.Name, Zone: o.Zone(), CapacityType: o.CapacityType()}
+				oldCount, exists := npCost.offeringCounts[offeringKey]
+				newMap[offeringKey] = OfferingCount{
+					Count: lo.Ternary(exists, oldCount.Count, 0),
+					Price: o.Price,
```
jmdeal, line 153: I'd consider special casing this to 0 for `Reserved` since the cost just exists to provide relative ordering. The cost is small enough in practice that I doubt it would have any real world impact to the metric, but it doesn't hurt to be correct.

jmdeal: Either that or we could just ignore all reserved offerings for cluster cost, that's not a bad idea either.

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,375 @@
...
+				}
+			})
+		})
+		npCost.offeringCounts = newMap
```
jmdeal, line 167: Do we want to ensure we're retaining offering from the previous map that don't exist in the new map that have existing NodeClaims? In practice this doesn't capture much since this can mainly occur with Reserved offerings, but it can happen with non-reserved as well (e.g. instance retirement). 

FWIW if we're excluding reserved I don't think this matters in practice, at least for the AWS provider. I can't speak for other providers which is why it may be a good idea to retain the entries.

jmdeal:
I left the UID comments in for posterity, but @DerekFrank and I discussed it offline and have decided to leave it as is for now. We generally want to use UIDs because the same resource can be deleted and recreated, leading us to map a previous entry to the new resource. In theory this could be problematic if the new NodePool has substantially different costs per offering, but this is unlikely to happen in this case. Using UIDs is also problematic since we don't get the UID for delete events - just the namespaced name. Given the low criticality of this component (it's a rough billing estimate) and the general alignment of cost across NodePools, we've decided to defer for a more general solution.

pkg/controllers/state/informer/pricing.go
DerekFrank: Discussed this, we don't have the UID available when parsing deletes, so we would need to keep a separate mapping open. I re-opened an old controller runtime issue to see if UID makes sense to be included in reconcile requests: https://github.com/kubernetes-sigs/controller-runtime/issues/876

pkg/controllers/state/informer/pricing.go
DerekFrank: Fair point, I can add a comment

pkg/state/cost/cost.go
DerekFrank: Same response as above :D

pkg/state/cost/cost.go
DerekFrank: It depends on what you're using the metric for. If you actually want to know how much your cluster costs then I agree. If you're trying to track nodeoverlay impact on cluster cost calculations or see how well Karpenter bin packs then maybe not

pkg/controllers/state/informer/pricing.go
DerekFrank: I'd rather have a pricing generation check, we can refer to https://github.com/kubernetes-sigs/karpenter/pull/2866/changes for that

pkg/state/cost/cost.go
DerekFrank: I don't know if thats a blocker for release, we can always add it later

pkg/controllers/state/informer/pricing.go
jmdeal: Yeah, long term that works. I moreso wanted to sanity check that this logic worked today.

pkg/state/cost/cost.go
jmdeal: I don't think this should be difficult to address, and I'd rather address it now if we plan on leaving reserved offerings in the computation. I can see this being something we never come back to.

pkg/state/cost/cost.go
jmdeal: Fair enough, I actually misremembered a detail of the NodeOverlay implementation - at one point we blocked cost overrides for Reserved offerings but it looks like we did agree to remove that constraint before release. Since it's possible to apply reserved NodeOverlays I agree that it makes sense to keep them.

pkg/state/cost/cost.go
jmdeal: Actually, this would only be best effort since we don't have a way to handle controller restarts since the offering pricining information isn't persisted to the NodeClaim. Maybe we should annotate each NodeClaim with the cost estimate? This would need to be continuously updated, but that would solve the restart story.

DerekFrank: Discussed offline, we agreed that the error logging is sufficient for an initial release

pkg/state/cost/cost.go
```diff
@@ -0,0 +1,388 @@
...
+	if !exists {
+		// our offerings must be out of date, we should update and retry
+		err := cc.internalNodepoolUpdate(ctx, np)
+		if err != nil {
+			return fmt.Errorf("failed to update nodepool %q during retry while searching for offering for instance %q in zone %q with capacity %q, %w", np.Name, offeringKey.InstanceName, offeringKey.Zone, offeringKey.CapacityType, err)
+		}
+		oc, exists = cc.npCostMap[npName].offeringCounts[offeringKey]
+		if !exists {
+			oc = OfferingCount{Count: 1, Price: 0.0}
+			log.FromContext(ctx).Error(fmt.Errorf("failed to find offering %q during retry while searching for instance %q in zone %q with capacity %q in nodepool %q", offeringKey, offeringKey.InstanceName, offeringKey.Zone, offeringKey.CapacityType, npName), "Offering won't be counted towards total cluster cost")
```
jmdeal: The tiniest of nits, we don't capitalize the start of logs
```suggestion
			log.FromContext(ctx).Error(fmt.Errorf("failed to find offering %q during retry while searching for instance %q in zone %q with capacity %q in nodepool %q", offeringKey, offeringKey.InstanceName, offeringKey.Zone, offeringKey.CapacityType, npName), "offering won't be counted towards total cluster cost")
```

jmdeal:
Manually squashing and merging, CI failures are due to an ongoing coveralls outage: https://status.coveralls.io/

PR #2848, opened 2026-02-10
log consolidation outcome

grosser:
**Description**
it is really hard to tell atm if karpenter is doing anything or if it is struggeling to consolidate,
this might help, but not sure if it is the right log level or right place

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -166,16 +166,20 @@ func (c *Controller) Reconcile(ctx context.Context) (reconciler.Result, error) {
 		success, err := c.disrupt(ctx, m)
 		if err != nil {
 			if errors.IsConflict(err) {
+				log.FromContext(ctx).V(1).Info("consolidation outcome: conflict", "method", m)
 				return reconciler.Result{Requeue: true}, nil
 			}
+			log.FromContext(ctx).V(0).Error(err, "consolidation outcome: error", "method", m)
 			return reconciler.Result{}, serrors.Wrap(fmt.Errorf("disrupting, %w", err), strings.ToLower(string(m.Reason())), "reason")
 		}
 		if success {
+			log.FromContext(ctx).V(1).Info("consolidation outcome: success", "method", m)
 			return reconciler.Result{RequeueAfter: singleton.RequeueImmediately}, nil
 		}
 	}
 
 	// All methods did nothing, so return nothing to do
+	log.FromContext(ctx).V(1).Info("consolidation outcome: noop")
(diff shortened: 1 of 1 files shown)


pkg/controllers/disruption/controller.go
```diff
@@ -166,16 +166,20 @@ func (c *Controller) Reconcile(ctx context.Context) (reconciler.Result, error) {
...
 			return reconciler.Result{}, serrors.Wrap(fmt.Errorf("disrupting, %w", err), strings.ToLower(string(m.Reason())), "reason")
 		}
 		if success {
+			log.FromContext(ctx).V(1).Info("consolidation outcome: success", "method", m)
 			return reconciler.Result{RequeueAfter: singleton.RequeueImmediately}, nil
 		}
 	}
 
 	// All methods did nothing, so return nothing to do
+	log.FromContext(ctx).V(1).Info("consolidation outcome: noop")
```
DerekFrank, line 182: I'm not sure we want to be logging every consolidation loop that Karpenter performs. noops probably shouldn't be logged at least

pkg/controllers/disruption/controller.go
```diff
@@ -166,16 +166,20 @@ func (c *Controller) Reconcile(ctx context.Context) (reconciler.Result, error) {
...
 		if err != nil {
 			if errors.IsConflict(err) {
+				log.FromContext(ctx).V(1).Info("consolidation outcome: conflict", "method", m)
 				return reconciler.Result{Requeue: true}, nil
 			}
+			log.FromContext(ctx).V(0).Error(err, "consolidation outcome: error", "method", m)
 			return reconciler.Result{}, serrors.Wrap(fmt.Errorf("disrupting, %w", err), strings.ToLower(string(m.Reason())), "reason")
 		}
 		if success {
+			log.FromContext(ctx).V(1).Info("consolidation outcome: success", "method", m)
```
DerekFrank, line 176: If karpenter does consolidate, won't you see it log other lines like provisioning a new nodeclaim or deleting old ones?

pkg/controllers/disruption/controller.go
grosser: yeah good point, this one would be the spammiest and is the least useful ...

pkg/controllers/disruption/controller.go
grosser: yeah, there will be some other logs, but nothing really clear and obvious afaik "hey I'm consolidating X" or "consolidation failed because of err", most of the errors get silently turned into `return false` or go into the controller result, but not into logs

grosser:
let me close this for now, worked well enough to show that we did not have any errors, and will need some more time that I don't have atm

PR #2851, opened 2026-02-12
chore: ignore pods that reference a pv that is marked for deletion

jigisha620:
Fixes #N/A 

**Description**
Added validation to reject pods with PersistentVolumes marked for deletion. We are doing this because even if we create a node to which the pod can schedule, it will remain stuck because the volumeattachment will never succeed because the PV is marked for deletion.

- Updated validateVolume() function to check if a PersistentVolume has a DeletionTimestamp set
- Returns an error when attempting to schedule pods that reference PVs being deleted
- Aligns with existing validation pattern that rejects PVCs with deletion timestamps

- Added test case: "should not launch nodes for pod with bound persistentVolume that is marked for deletion"
- Validates that no nodes are created when PV has deletion timestamp set


jmdeal (commented):
/approve

PR #2853, opened 2026-02-12
feat: make disruption polling interval configurable

alperencelik:
**Description**

This PR aims to make the pollingInterval be configurable by operator options. Currently the pollingInterval is const and set to 10 seconds and IMO that's overkill for calculating the disruptions.

```go
func (c *Controller) Reconcile(ctx context.Context) (reconciler.Result, error) {
...
    success, err := c.disrupt(ctx, m)
    // If c.disrupt returns false, nil it fallbacks to return statement below
...
	// All methods did nothing, so return nothing to do
	return reconciler.Result{RequeueAfter: pollingPeriod}, nil
``` 

In a case Karpenter can't delete or replace the node (i.e: can't replace with cheaper node) it tries to do the calculation 10 seconds later. 

```shell
go test ./pkg/operator/options/... 

make apply-with-kind 
``` 

### DIFF
--- a/kwok/charts/README.md
+++ b/kwok/charts/README.md
@@ -51,9 +51,10 @@ For full Karpenter documentation please checkout [https://karpenter.sh](https://
 | serviceMonitor.additionalLabels | object | `{}` | Additional labels for the ServiceMonitor. |
 | serviceMonitor.enabled | bool | `false` | Specifies whether a ServiceMonitor should be created. |
 | serviceMonitor.endpointConfig | object | `{}` | Endpoint configuration for the ServiceMonitor. |
-| settings | object | `{"batchIdleDuration":"1s","batchMaxDuration":"10s","featureGates":{"nodeRepair":false,"spotToSpotConsolidation":false},"minValuesPolicy":"Strict","preferencePolicy":"Respect"}` | Global Settings to configure Karpenter |
+| settings | object | `{"batchIdleDuration":"1s","batchMaxDuration":"10s","disruptionPollingPeriod":"10s","featureGates":{"nodeRepair":false,"spotToSpotConsolidation":false},"minValuesPolicy":"Strict","preferencePolicy":"Respect"}` | Global Settings to configure Karpenter |
 | settings.batchIdleDuration | string | `"1s"` | The maximum amount of time with no new pending pods that if exceeded ends the current batching window. If pods arrive faster than this time, the batching window will be extended up to the maxDuration. If they arrive slower, the pods will be batched separately. |
 | settings.batchMaxDuration | string | `"10s"` | The maximum length of a batch window. The longer this is, the more pods we can consider for provisioning at one time which usually results in fewer but larger nodes. |
+| settings.disruptionPollingPeriod | string | `"10s"` | The interval between polling rounds for disruption calculations.  |
 | settings.featureGates | object | `{"nodeRepair":false,"spotToSpotConsolidation":false}` | Feature Gate configuration values. Feature Gates will follow the same graduation process and requirements as feature gates in Kubernetes. More information here https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features |
 | settings.featureGates.nodeRepair | bool | `false` | nodeRepair is ALPHA and is disabled by default. Setting this to true will enable node repair. |
 | settings.featureGates.spotToSpotConsolidation | bool | `false` | spotToSpotConsolidation is ALPHA and is disabled by default. Setting this to true will enable spot replacement consolidation for both single and multi-node consolidation. |
--- a/kwok/charts/templates/deployment.yaml
+++ b/kwok/charts/templates/deployment.yaml
@@ -110,6 +110,10 @@ spec:
             - name: BATCH_IDLE_DURATION
               value: "{{ . }}"
           {{- end }}
+          {{- with .Values.settings.disruptionPollingPeriod }}
+            - name: DISRUPTION_POLLING_PERIOD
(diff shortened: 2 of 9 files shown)


jigisha620:
@alperencelik,
Can you help me understand your use-case for making this configurable?

alperencelik:
> @alperencelik,
> 
> Can you help me understand your use-case for making this configurable?

Currently karpenter tries to consolidate node every 10 seconds and that duration is very small for the clusters having thousands of events (such as scale up & downs) so I'd like to tell Karpenter to wait for some time where the nodes are stabilizing. Also having a requeue of 10 seconds is generating a queue for controller which is having an outcome of calling cloud APIs etc. Shortly I'd say I'd like to hav sth as cooldown period in a case of high number of events in cluster.

jigisha620:
@alperencelik,
Just to clarify — is the concern here that the cluster is experiencing instability? Specifically, are you seeing excessive node churn due to the shorter reconciliation period?

alperencelik:
> @alperencelik, Just to clarify — is the concern here that the cluster is experiencing instability? Specifically, are you seeing excessive node churn due to the shorter reconciliation period?

Yes, that's completely true. What happens is once the consolidateAfter window passes, Karpenter kicks off consolidation but if it can't actually find a cheaper replacement, it just keeps retrying every pollingInterval (default 10 sec). So it's basically spinning in a loop doing nothing useful.

IMO, 10 seconds already feels like overkill to me, but the bigger ask here is just making it configurable so that I can override that and make my nodes less vulnerable to disruption.

DerekFrank:
I'm not sure this changes anything in terms of vulnerability. We generally want to avoid letting users tweak these kinds of reconciler specific configurations unless theres a very good reason. In your case since Karpenter isn't finding any valid candidates its not doing anything disruptive to your nodes. 

Is there a reason why Karpenter cycling on the problem ineffectively is an issue for you I'm not understanding?

alperencelik:
> I'm not sure this changes anything in terms of vulnerability. We generally want to avoid letting users tweak these kinds of reconciler specific configurations unless theres a very good reason. In your case since Karpenter isn't finding any valid candidates its not doing anything disruptive to your nodes.
> 
> Is there a reason why Karpenter cycling on the problem ineffectively is an issue for you I'm not understanding?

There are a couple of main issues with respect to your comment:

1) Evaluating disruption every 10 seconds doesn't give the cluster enough time to reach a steady state. When pods are frequently scaling up and down  Karpenter can't make a reasonable decision because the state keeps changing. A longer interval would allow things to settle before evaluating, leading to better disruption decisions.

2) Unnecessary reconciliation loops are using resources. They put load on the API server and increase Karpenter's own resource consumption to run that logic. On top of that, each reconciliation that doesn't result in any comman decision still fires a Kubernetes event. Reconcile each 10s means 6 events per minute per node and that makes in >100nodes it has more than 1000 events per minute.

3) I'm not clear on what value is gained by locking the requeue interval to a fixed value. Unless there's a clock-based implementation there is no reason that I could think of for making it not configurable.  The reason for making it configurable while keeping the default value is to give people more control over the Karpenter. That said, I'm happy to hear if there's a specific reason it needs to stay fixed.

alperencelik:
Screenshot 2026-02-25 at 10 58 27

Also I just realized that sometimes that disruption reconciliation takes ~30 seconds to complete. IMO having a requeue of 10s for something that took 30 seconds is a good example of overkill

I'm pretty sure I can mitigate by giving more resources to Karpenter but I don't think it's the right solution.

DerekFrank:
I buy the argument that Karpenter might want to wait for the cluster to reach a steadier state before acting, but tweaking the disruption polling interval is not the right way to get there.

> I'm pretty sure I can mitigate by giving more resources to Karpenter but I don't think it's the right solution.

Say more about this, are you seeing latency elsewhere or the disruption controller hogging resources?

alperencelik:
> I buy the argument that Karpenter might want to wait for the cluster to reach a steadier state before acting, but tweaking the disruption polling interval is not the right way to get there.

Well, I don't know if there is an existing mechanism I'm missing to tell Karpenter hold off on disruption evaluation until things settle but could you elaborate on that? and maybe you can tell why you think the pollingInterval for the disruption controller isn't solving the problem here. 

> Say more about this, are you seeing latency elsewhere or the disruption controller hogging resources?

At that time I do see karpenter is using more resources than usual but to me the issue here is logical. The disruption loop requeues every 10s, but each reconciliation takes ~30s to complete, and since new iterations don't wait for the previous one to finish, work just keeps piling up. What I meant is throwing more resources at Karpenter might speed up individual reconciliations, but it doesn't change the fact that the loop frequency doesn't account for how long the work actually takes.

Overall, I still don't quite understand why you're standing against making it non-configurable. Again if you have any reason for keeping it fixed value I'm happy to hear that.

DerekFrank:
The project has a strong bias towards avoiding knobs if at all possible. If there is a smarter value, either dynamic or static, we should find a way to get to that. We also don't like knobs that aren't user facing behavior and instead are knobs about the deep internals of the system. If we rewrote this project in another language or another framework, we still want the knobs to make sense. 

I'm still not understanding what the user facing behavior you are seeing that you would like to avoid. 

If this PR stales out, we can continue the discussion in an issue

DerekFrank:
Going to close as not planned, as the implementation suggested is against the tenets of the project: 

>  Exposing Internal Implementation Details

https://github.com/kubernetes-sigs/karpenter/blob/main/SCOPE.md

I remain curious and interested in diving deeper into discovering the root of the desire for a configurable disruption polling interval. If you or someone else coming across this issue would like to discuss more, please open an issue

/close
/not-planned

PR #2599, opened 2025-10-29
chore: use structured logging for errors

grosser:
**Description**
[structured logging](https://stackify.com/what-is-structured-logging-and-why-developers-need-it/) helps filter or exclude the same type of error, but it's harder when every error is different

**Todo**
if this is acceptable then find all the other .Error calls and convert them

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2861, opened 2026-02-17
chore: do not create errors just to log them

grosser:
tiny efficiency, just stumbled over that when digging though the code

cc @DerekFrank found while doing https://github.com/kubernetes-sigs/karpenter/pull/2599


DerekFrank:
When you say efficiency, do you mean CPU cycles or allocs or something? I don't know if we use the .Error(nil, "...") pattern anywhere

grosser (commented):
reverted that part
... it is more efficient for cpu/alloc, but looking through the code some more there are loggers that render err different than just `err.String() + msg`, so let's just leave these alone, removing the unnecessary WithValues calls could still be nice

PR #2864, opened 2026-02-18
fix: [buildDomainGroups] filter out empty string domains

### LINKED ISSUE #1384: Topology spread constraints fail with empty zone domain in counts
### Version

Karpenter version: 1.6.6, and 1.7.1
Mode: Self hosted (see below comment, someone reproduced with NAP as well).
Kubernetes version: 1.33
Region: eastus2

### Expected Behavior

With topology spread constraint `maxSkew: 1` and `whenUnsatisfiable: DoNotSchedule`, Karpenter should provision nodes in zones that satisfy the constraint.

**Current distribution:** `zone1=2, zone2=1, zone3=1`
**Expected:** Provision in zone2 or zone3 (maxSkew would be 1)

### Actual Behavior

Karpenter reports topology constraint as unsatisfiable and refuses to provision nodes. Logs show an unexpected empty string domain in topology counts:

```
counts = : 0, eastus2-1: 2, eastus2-2: 1, eastus2-3: 1
```

This appears to cause incorrect maxSkew calculation. My guess here is that the first part of that error string is missing a zone, hence ` : 0`, when others have format `$zone: $count`.

### Steps to Reproduce the Problem

1. Deploy with topology spread constraints in a zonal region:

```yaml
topologySpreadConstraints:
  - maxSkew: 1
    labelSelector:
      matchLabels:
        app.kubernetes.io/name: my-name
```

2. Scale to create pending pods requiring new nodes

3. Observe Karpenter refusing to provision despite valid zones available

### Resource Specs and Logs

Note, we have two node pools, `default` and `ephemeral` (spot instances for cronjob job pods).

In both cases for us, one Deployment, and one Statefulset, both using the `default` nodepool, caused this error.

```json
{"level":"ERROR","time":"2026-01-27T18:23:45.982Z","logger":"controller","message":"could not schedule pod","commit":"28204ca","controller":"provisioner","namespace":"","name":"","reconcileID":"7b152da1-16b0-41a1-a483-025bfe3f72ca","Pod":{"name":"<REDACTED_POD_NAME>","namespace":"<REDACTED_POD_NAMESPACE>"},"taint":"dedicated=ephemeral:NoSchedule","error":"did not tolerate taint (taint=dedicated=ephemeral:NoSchedule); unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = : 0, eastus2-1: 1, eastus2-2: 1, eastus2-3: 1, podDomains = topology.kubernetes.io/zone Exists, nodeDomains = topology.kubernetes.io/zone Exists","errorCauses":[{"error":"did not tolerate taint (taint=dedicated=ephemeral:NoSchedule)"},{"error":"unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = : 0, eastus2-1: 1, eastus2-2: 1, eastus2-3: 1, podDomains = topology.kubernetes.io/zone Exists, nodeDomains = topology.kubernetes.io/zone Exists"}]}
```

### Short-term mitigation

Switching from `DoNotSchedule` to `ScheduleAnyway` allowed Karpenter to create a node, and immediately following, saw the event `Pod should schedule on: nodeclaim/$name`.

I include this because it seems to me that this setting is only really for the kube scheduler, and since karpenter _still_ needed to create a new node to fit, it probably should have been able to do so in the earlier case with `DoNotSchedule`. In fact, we rely on `DoNotSchedule` to require karpenter to schedule a new node in another zone, so that we maintain zone spread.

This is what leads me to believe this is a bug in the implementation somewhere.

### Community Note

k24dizzle:
Fixes https://github.com/Azure/karpenter-provider-azure/issues/1384

**Description**

Filter out empty string (`""`) domain values when seeding topology domain groups from instance type requirements in `buildDomainGroups()`.

- Cloud providers (e.g. Azure) may represent non-zonal instance types by setting `topology.kubernetes.io/zone` to `In [""]`, [a set containing an empty string](https://github.com/Azure/karpenter-provider-azure/blob/287d5796c94e2e7051e407ad7900c4c2e294be78/pkg/providers/instancetype/instancetypes.go#L194-L208).
- When `buildDomainGroups()` iterates over all NodePools and their instance types, it inserts these values into a shared `domainGroups` map. If **any** NodePool lacks an explicit zone constraint (meaning there's no existing zone requirement to intersect against), the empty string flows through unopposed and becomes a topology domain.
- This causes topology spread constraints with `whenUnsatisfiable: DoNotSchedule` to become permanently unsatisfiable. The scheduler sees a phantom 4th domain that can never be satisfied:

`Failed to schedule pod, unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = : 0, westus2-1: 1, westus2-2: 1, westus2-3: 1, podDomains = topology.kubernetes.io/zone Exists, nodeDomains = topology.kubernetes.io/zone In [westus2-1 westus2-2 westus2-3];`

- Added a unit test
- Still need test this out in an AKS cluster via the reproduction yamls in this [comment](https://github.com/Azure/karpenter-provider-azure/issues/1384#issuecomment-3922716496).

### DIFF
--- a/pkg/controllers/provisioning/scheduling/topology.go
+++ b/pkg/controllers/provisioning/scheduling/topology.go
@@ -121,6 +121,9 @@ func buildDomainGroups(nodePools []*v1.NodePool, instanceTypes map[string][]*clo
 					domainGroups[topologyKey] = NewTopologyDomainGroup()
 				}
 				for _, domain := range requirement.Values() {
+					if domain == "" {
+						continue
+					}
 					domainGroups[topologyKey].Insert(domain, np.Spec.Template.Spec.Taints...)
 				}
 			}
--- a/pkg/controllers/provisioning/scheduling/topology_test.go
+++ b/pkg/controllers/provisioning/scheduling/topology_test.go
@@ -30,7 +30,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider/fake"
+	pscheduling "sigs.k8s.io/karpenter/pkg/scheduling"
 	"sigs.k8s.io/karpenter/pkg/test"
 	. "sigs.k8s.io/karpenter/pkg/test/expectations"
 )
@@ -3114,5 +3116,70 @@ var _ = Describe("Topology with Volume Requirements", func() {
 
 		// Verify final skew: should be 2,1,1 (zone-1 has 2 pods now)
 		ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(1, 1, 2))
+	})
+	It("should not include empty string domains when building topology domain groups", func() {
+		// Some cloud providers (e.g. Azure) return offerings with an empty zone string for
+		// non-zonal instance types. Empty domain strings should be excluded from topology
+		// domain groups so they don't interfere with topology spread decisions.
+		cloudProvider.InstanceTypes = []*cloudprovider.InstanceType{
+			fake.NewInstanceType(fake.InstanceTypeOptions{
+				Name: "instance-with-empty-zone",
+				Offerings: []*cloudprovider.Offering{
+					{
+						Available: true,
+						Requirements: pscheduling.NewLabelRequirements(map[string]string{
+							v1.CapacityTypeLabelKey:  v1.CapacityTypeOnDemand,
+							corev1.LabelTopologyZone: "",
+						}),
+						Price: 1.0,
+					},
+				},
+			}),
+			fake.NewInstanceType(fake.InstanceTypeOptions{
+				Name: "instance-with-real-zones",
+				Offerings: []*cloudprovider.Offering{
+					{
+						Available: true,
+						Requirements: pscheduling.NewLabelRequirements(map[string]string{
+							v1.CapacityTypeLabelKey:  v1.CapacityTypeOnDemand,
+							corev1.LabelTopologyZone: "test-zone-1",
+						}),
+						Price: 1.0,
+					},
+					{
+						Available: true,
+						Requirements: pscheduling.NewLabelRequirements(map[string]string{
+							v1.CapacityTypeLabelKey:  v1.CapacityTypeOnDemand,
+							corev1.LabelTopologyZone: "test-zone-2",
+						}),
+						Price: 1.0,
+					},
+					{
+						Available: true,
+						Requirements: pscheduling.NewLabelRequirements(map[string]string{
+							v1.CapacityTypeLabelKey:  v1.CapacityTypeOnDemand,
+							corev1.LabelTopologyZone: "test-zone-3",
+						}),
+						Price: 1.0,
+					},
+				},
+			}),
+		}
+		ExpectApplied(ctx, env.Client, nodePool)
+		topology := []corev1.TopologySpreadConstraint{{
+			TopologyKey:       corev1.LabelTopologyZone,
+			WhenUnsatisfiable: corev1.DoNotSchedule,
+			LabelSelector:     &metav1.LabelSelector{MatchLabels: labels},
+			MaxSkew:           1,
+		}}
+		pods := test.UnschedulablePods(test.PodOptions{
+			ObjectMeta:                metav1.ObjectMeta{Labels: labels},
+			TopologySpreadConstraints: topology,
+		}, 4)
+		ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pods...)
+		for _, pod := range pods {
+			ExpectScheduled(ctx, env.Client, pod)
+		}
+		ExpectSkew(ctx, env.Client, "default", &topology[0]).To(ConsistOf(2, 1, 1))
 	})
 })


justinwalz:
Hey, just wanted to bump this PR for a review if possible, as it would be very helpful to resolve an issue I'm seeing in the Azure provider here https://github.com/Azure/karpenter-provider-azure/issues/1384

Thank you!

jmdeal:
Copying over our conversation from the k8s slack about this issue ([ref](https://kubernetes.slack.com/archives/C04JW2J5J5P/p1772499673990729))

---

**Jason Deal**

I'm wondering if we should be filtering this out, or if cloud providers just shouldn't provide this type of requirement at all. When an instance with a non-zonal offering is launched and joins the cluster, what value for zone is populated by CCM? Does it populate the label at all?

I need to take a closer look at the PR, but based on the description I could see us taking a solution like this in the short-term since we make a bunch of assumptions about zonal requirements throughout the codebase. Long-term though I think we want a better way to express non-zonal instance type offerings. Are you able to point me to some Azure documentation for what types of instance type offerings this applies to? I just want to make sure I understand what we're trying to model — I don't think AWS has a comparable concept.

**Kevin**

I think in Azure/AKS, some instances have all their zones restricted (but tbh I'm not sure if these are considered non-zonal instances) and as a result when Azure reports back to Karpenter, it says that these instances have `topology.kubernetes.io/zone` set to an empty string.

Here's an example of a SKU that I think would return no zones, because of the restrictions:

```bash
% az vm list-skus --location westus2 --resource-type virtualMachines --size Standard_D15_v2 --output json | jq '.[0] | {locationInfo, restrictions}'
```
```json
{
  "locationInfo": [
    {
      "location": "westus2",
      "zoneDetails": [],
      "zones": [
        "2",
        "3",
        "1"
      ]
    }
  ],
  "restrictions": [
    {
      "reasonCode": "NotAvailableForSubscription",
      "restrictionInfo": {
        "locations": [
          "westus2"
        ],
        "zones": [
          "1",
          "2",
          "3"
        ]
      },
      "type": "Zone",
      "values": [
        "westus2"
      ]
    }
  ]
}
```

**Jason Deal**

Is it possible to launch these instances then? It seems like it's not, in which case I would filter it out from the `cloudprovider.GetInstanceTypes` result altogether rather than set the zonal requirement to `""`. Either that, or you could mark the offerings for each restricted zone with `available: false`.

I might misunderstand what it means for a zone to be restricted though — I'm guessing based on this JSON.

**Kevin**

Hmm, I feel like I've seen nodes provisioned before in my cluster like these docs:

```bash
% kubectl get nodes -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.labels.topology\.kubernetes\.io/zone}{"\n"}{end}'
...
aks-system-24644131-vmss000000	0
aks-system-24644131-vmss000001	0
aks-system-24644131-vmss000002	0
```

Where the zone is `0` instead of something like `eastus-1`. Unfortunately I forgot which size these instance types were.

**Jason Deal**

> the node pool creates regional (not zone-pinned) instances and implicitly places instances throughout the region

Does this mean that the instance doesn't have a zone then, or it's just left to whatever mechanism schedules VMs in Azure to assign the zone?

It's not clear to me if `0` is a placeholder to mean "no zone / regional" or if it's an actual zone ID.

**Kevin**

It doesn't have a zone label in k8s, but Azure does put it in one of the zones behind the scenes I believe.

Tbh I'm not an Azure expert, but I can try to see if I can provision a non-zonal instance type in my AKS cluster again and investigate what size it is.

cc: @tallaxes if you have any thoughts here

Here's an example of a non-zonal instance type that could hypothetically get provisioned in an AKS cluster — in `northcentralus`, zones aren't supported:

```bash
% az vm list-skus --location northcentralus --resource-type virtualMachines --size Standard_A1_v2 --output json | jq '.[0] | {locationInfo,restrictions}'
```
```json
{
  "locationInfo": [
    {
      "location": "northcentralus",
      "zoneDetails": [],
      "zones": []
    }
  ],
  "restrictions": []
}
```

**Jason Deal**

Interesting — I find it kind of odd that Azure would frame that as no AZs rather than a single AZ, but maybe there's a good reason. I'm surprised that nothing is populated for `topology.kubernetes.io/zone` in that case; that seems like it would break applications which are being deployed across regions which have zonal TSCs.

Are we sure there's no fallback value? I took a brief look at Azure's cloud provider manager implementation and it looks like it has a fallback.

**Kevin**

> that seems like it would break applications which are being deployed across regions which have zonal TSCs

Yes, from my experience, Karpenter + zonal TSCs don't seem to work in AKS: https://github.com/Azure/karpenter-provider-azure/issues/1384

To summarize my current understanding: when Karpenter gets instance types from the Azure provider, it can return two kinds of instances with empty zones:

1. They are non-zonal (which seems valid?)
2. They have been restricted in zones for that region (which probably should be filtered out)

When Karpenter gets those `topology.kubernetes.io/zone = ""` instances, that's when the provisioning logic around TSCs gets confused by this "phantom domain" (which my PR aims to fix).

> Are we sure there's no fallback value?

I think the fallback value for the zone is the `faultDomain` here, which might represent the `0` value — but I think this logic kicks in only once the nodes are actually provisioned.

**Jason Deal**

> Karpenter + zonal TSCs don't seem to work in AKS

My thought is that the issue is broader — if the `topology.kubernetes.io/zone` label isn't being populated for these regions without AZs, applications with zonal TSCs won't work in general, not just with Karpenter. Granted this is more of a quality of life issue than anything else in the general case, since you just need to update application manifests to remove the TSC in those regions.

> I think the fallback value for the zone is the `faultDomain` here, which might represent the `0` value, but I think this logic kicks in only once the nodes are actually provisioned?

Yeah, my expectation is that the Azure CCM would still add this label with a value of `0` when the node registers, not omit the label altogether. This addresses the general compatibility concerns. If that is the behavior, I think the Azure Karpenter provider should be updated to match. This is the behavior I think is right for the two scenarios we discussed:

1. **Non-zonal regions:** InstanceTypes should have the following requirement: `topology.kubernetes.io/zone In ['0']`. There shouldn't be offerings for any other zones.
2. **Instances with restricted AZs:** The offerings for those AZs should be omitted or marked as unavailable. Marking them as unavailable is probably the better option of the two, but it really just affects the error message a user sees when there are no instance types available that can satisfy constraints.

**Kevin**

Thank you for taking the time to talk through this with me.

I do think my fix would be harmless and would help, but I think you are right that the Azure provider needs to be smart enough to not return these `topology.kubernetes.io/zone = ""` instance types to Karpenter in the first place.

In my case, I don't want Karpenter to even consider provisioning a non-zonal instance type (`topology.kubernetes.io/hostname In ['0']`) if my workloads use TSCs that want to spread across zones. I don't want to spread 4 pods across 3 AZs and 1 non-zonal domain.

It would be nice if there was a way to tell the `karpenter-provider-azure` to not offer these instance types to the Karpenter controller.

I went ahead and added a comment to the Azure provider issue: https://github.com/Azure/karpenter-provider-azure/issues/1384#issuecomment-4092330807 to hopefully get the ball rolling there.

**Jason Deal**

> I do think my fix would be harmless and would help

I agree — my only hesitation with going forward with it would be that it could mask subtle bugs in the future. That's why I'd lean towards addressing the root cause in the provider, but I've made a lot of assumptions about Azure here.

**Kevin**

Makes sense, I'll try to get some momentum going in the Azure provider issue. 🙏

tallaxes:
The original issue in Azure provider is fixed in https://github.com/Azure/karpenter-provider-azure/pull/1615; with this fix core won't observe "" zone values anymore. A more comprehensive design for support of regional placement (including some background) is described in https://github.com/Azure/karpenter-provider-azure/pull/1628 (we plan to introduce `karpenter.azure.com/placement-scope: [zonal, regional]` selector/label).

jmdeal:
I'll go ahead and close this out since there's a path forward in Azure. Based on our discussion in slack I don't think there's a valid use-case for `""` valued domains, so I'll close this PR out. @k24dizzle feel free to re-open if you don't think your use-cases are addressed.

/close

PR #2869, opened 2026-02-24
POC prefer r types if new pods have high memory requests

grosser:
**Description**
our nodepools ended up always spinning up m or c types (cheapest) since the pendings pods were never big enough to require r types (we run 8xl as the smallest ~30 cpus) so we ended up with lots of c/m-s but the actual pod requests were very memory heavy so we had to wait for consolidation which lead to more pod interruptions
(also tried changing batching settings, but even 1m/5m did not help)

updated the nodepool and saw that we get a good mix of m and r on the first try now


jmdeal:
Going to close this out since we wouldn't accept an upstream change that's hardcoded for AWS instance types. If you have proposals for changes to the binpacking algorithm to better account for high-memory nodes in general though that would be interesting to review.

grosser:
I think we could do something where we look at the instance description and pick types that match the cpu/memory ratio pods to nodes, that should work for all clouds
... would need to be opt-in though

PR #2871, opened 2026-02-26
POC consolidate within each node-group

grosser:
POC for https://github.com/kubernetes-sigs/karpenter/issues/2814

**Description**
to avoid doing any math across node-pools that are incompatible,
group them by labels first (here node-type + az, but should be configureable later)

pros:
- improves consolidation outcomes
- circumvents 100 node limit

cons: 
- more validation errors are getting reporter since we run way more of them now
- validation is run twice for the chosen command, that could be optimized away (wanted to keep the patch self-container so I can rebase easily)

also fixes issues where to be consolidated pods have pods with different az requirements, leading to them always generating 2+ replacement nodes and therefore failing multi consolidation

running this in our clusters improved binpacking, sadly consolidation time has not improved much, which I'd have expected too

### DIFF
--- a/pkg/controllers/disruption/multinodeconsolidation.go
+++ b/pkg/controllers/disruption/multinodeconsolidation.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/awslabs/operatorpkg/option"
 	"github.com/samber/lo"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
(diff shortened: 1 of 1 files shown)


cnmcavoy:
It might be worth considering Operating Systems or CPU architectures as boundaries as well

grosser:
would generally say users should label their pools with what is important to them, then configure these labels to group by

for example we have nodepools that can take "arm or intel" so a strict arch separation would not work

kimsehwan96:
I think computed/automated boundary setting would be hard to cover all kinds of use cases, because the policies for logically dividing nodepools can vary significantly depending on how users organize their workloads.

So giving a field to control boundary definition or node label based boundary control would be more useful. 

Anyway, improved binpacking sounds nice!

jklaw90 (approved):

PR #2874, opened 2026-02-27
feat: Add grace period for do-not-disrupt annotation

### LINKED ISSUE #752: Add a gracePeriod for the `do-not-evict` pod annotation

AndrewMitchell25:
Fixes #752 

**Description**
RFC: https://github.com/kubernetes-sigs/karpenter/pull/2656

This PR allows the `karpenter.sh/do-not-disrupt` annotation on a pod to be set to a time.Duration, preventing disruption from the start time of the pod until the duration has elapsed instead of permanently being set to `true`.

Added unit tests and integration tests for functionality, manually tested, `make presubmit`
New integration test passes [here](https://github.com/aws/karpenter-provider-aws/actions/runs/22685151484/job/65765993182) and all other integration tests pass [here](https://github.com/aws/karpenter-provider-aws/actions/runs/22639761430/job/65639412450)

<details>
<summary>Manual testing details</summary>

 * Built and deployed Karpenter to a kind cluster
 * Created one large pod with no do-not-disrupt annotation, and one small pod with a 5 minute do-not-disrupt annotation, both on the same node
 *  
    ```
    === Pod Details ===
    Name: small-pod-5min-grace
    Annotation: 5m
    Start Time: 2026-02-27T19:45:25Z
    Node: affectionate-meninsky-1600157136
    Name: large-pod-no-protection
    Annotation: 
    Start Time: 2026-02-27T19:45:25Z
    Node: affectionate-meninsky-1600157136
    ```
 * Deleted the large pod, which should cause Karpenter to consolidate to a smaller node, but should be blocked by do-not-disrupt
 *  
    ```
    {"level":"DEBUG","time":"2026-02-27T19:45:55.001Z","logger":"controller","caller":"disruption/consolidation.go:75","message":"marking consolidatable","commit":"5e12af1-dirty","controller":"nodeclaim.disruption","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"default-7khgd"},"namespace":"","name":"default-7khgd","reconcileID":"79ceb98e-afe0-4b67-82d8-0d974d9c05c3","Node":{"name":"affectionate-meninsky-1600157136"}}
    {"level":"DEBUG","time":"2026-02-27T19:46:30.989Z","logger":"controller","caller":"disruption/consolidation.go:66","message":"removing consolidatable status condition","commit":"5e12af1-dirty","controller":"nodeclaim.disruption","controllerGroup":"karpenter.sh","controllerKind":"NodeClaim","NodeClaim":{"name":"default-7khgd"},"namespace":"","name":"default-7khgd","reconcileID":"6f0384bf-a49e-4ddf-ae1e-517735221213","Node":{"name":"affectionate-meninsky-1600157136"}}
    ```
 * Testing script output after waiting a while:
    ```
    Current time: 19:50:07
    Elapsed: 282s / 300s
    Status: PROTECTED (18s remaining)
    Karpenter should NOT disrupt the node yet
    
    Nodes:
    affectionate-meninsky-1600157136   Ready   <none>          4m46s   kwok-v0.5.2
    
    Pods:
    small-pod-5min-grace   1/1   Running   0     4m47s   10.244.3.0   affectionate-meninsky-1600157136   <none> <none>
    ```
 * After a little longer, logs showing the node was tainted and no nodes or pods:
    ```
    Current time: 19:50:38
    Elapsed: 313s / 300s
    Status: GRACE PERIOD EXPIRED
    Karpenter should now be able to disrupt the node
    
    Nodes:
    
    Pods:
    No resources found in default namespace.
    
    {"level":"INFO","time":"2026-02-27T19:50:22.026Z","logger":"controller","caller":"terminator/terminator.go:89","message":"tainted node","commit":"5e12af1-dirty","controller":"node.termination","controllerGroup":"","controllerKind":"Node","Node":{"name":"affectionate-meninsky-1600157136"},"namespace":"","name":"affectionate-meninsky-1600157136","reconcileID":"c29269a0-882a-4697-9259-25277f16f954","NodeClaim":{"name":"default-7khgd"},"taint.Key":"karpenter.sh/disrupted","taint.Value":"","taint.Effect":"NoSchedule"}
    ```

 * Ran the same test again after eventing was added, events from the cluster:
    ```
    default       5m23s                 Normal    Initialized                      NodeClaim/default-xpxxj               Status condition transitioned, Type: Initialized, Status: Unknown -> True, Reason: Initialized
    default       5m23s                 Normal    Ready                            NodeClaim/default-xpxxj               Status condition transitioned, Type: Ready, Status: Unknown -> True, Reason: Ready
    default       5m23s                 Normal    Registered                       NodeClaim/default-xpxxj               Status condition transitioned, Type: Registered, Status: Unknown -> True, Reason: Registered
    default       5m22s                 Normal    DisruptionBlocked                NodeClaim/default-xpxxj               Node is nominated for a pending pod
    default       5m22s                 Normal    DisruptionBlocked                Node/interesting-keldysh-2639058601   Node is nominated for a pending pod
    default       5m21s                 Normal    Scheduled                        Pod/small-pod-5min-grace              Successfully assigned default/small-pod-5min-grace to interesting-keldysh-2639058601
    default       5m21s                 Normal    Scheduled                        Pod/large-pod-no-protection           Successfully assigned default/large-pod-no-protection to interesting-keldysh-2639058601
    default       5m19s                 Normal    RegisteredNode                   Node/interesting-keldysh-2639058601   Node interesting-keldysh-2639058601 event: Registered Node interesting-keldysh-2639058601 in Controller
    default       82s (x2 over 3m22s)   Normal    DisruptionBlocked                NodeClaim/default-xpxxj               Pod has "karpenter.sh/do-not-disrupt" annotation (Pod=default/small-pod-5min-grace)
    default       82s (x2 over 3m22s)   Normal    DisruptionBlocked                Node/interesting-keldysh-2639058601   Pod has "karpenter.sh/do-not-disrupt" annotation (Pod=default/small-pod-5min-grace)
    default       62s (x3 over 5m2s)    Normal    DoNotDisruptUntil                Pod/small-pod-5min-grace              Pod will be disruptable at 2026-03-02T22:16:18Z
    default       24s                   Warning   FailedDraining                   Node/interesting-keldysh-2639058601   Failed to drain node, 1 pods are waiting to be evicted
    default       21s                   Normal    Evicted                          Pod/small-pod-5min-grace              Evicted pod: Forceful Termination
    default       21s                   Normal    DoNotDisruptGracePeriodElapsed   Pod/small-pod-5min-grace              The karpenter.sh/do-not-disrupt grace period has elapsed, pod is now disruptable
    default       20s                   Normal    Drained                          NodeClaim/default-xpxxj               Status condition transitioned, Type: Drained, Status: Unknown -> True, Reason: Drained
    default       20s                   Normal    Finalized                        NodeClaim/                            Finalized karpenter.sh/termination
    default       20s                   Normal    Finalized                        Node/                                 Finalized karpenter.sh/termination
    default       19s                   Normal    RemovingNode                     Node/interesting-keldysh-2639058601   Node interesting-keldysh-2639058601 event: Removing Node interesting-keldysh-2639058601 from Controller
    ```
</details>


pkg/utils/pod/events.go
```diff
@@ -0,0 +1,55 @@
...
+		DedupeValues:   []string{string(pod.UID), value},
+	}
+}
+
+func DoNotDisruptUntilEvent(pod *corev1.Pod, disruptableAt string) events.Event {
+	return events.Event{
+		InvolvedObject: pod,
+		Type:           corev1.EventTypeNormal,
+		Reason:         "DoNotDisruptUntil",
+		Message:        fmt.Sprintf("Pod will be disruptable at %s", disruptableAt),
```
jmdeal: nit: I'd mirror the wording you're using for the elapsed event

```suggestion
		Message:        fmt.Sprintf("The karpenter.sh/do-not-disrupt grace period will elapse at %s", disruptableAt),
```

pkg/utils/pod/events.go
```diff
@@ -0,0 +1,55 @@
...
+	return events.Event{
+		InvolvedObject: pod,
+		Type:           corev1.EventTypeWarning,
+		Reason:         "InvalidDoNotDisruptAnnotation",
+		Message:        fmt.Sprintf("Invalid karpenter.sh/do-not-disrupt annotation value %q, ignoring annotation. Use 'true' or a duration like '5m', '1h'", value),
+		DedupeValues:   []string{string(pod.UID), value},
+	}
+}
+
+func DoNotDisruptUntilEvent(pod *corev1.Pod, disruptableAt string) events.Event {
```
jmdeal: Rather than passing the formatted string, I'd take a `time.Time` directly. This ensures there won't be formatting differences between different call sites.

pkg/utils/pod/scheduling.go
```diff
@@ -105,12 +106,19 @@ func IsProvisionable(pod *corev1.Pod) bool {
...
 }
 
-// IsDisruptable checks if a pod can be disrupted based on validating the `karpenter.sh/do-not-disrupt` annotation on the pod.
+// IsDisruptable checks if a pod can be disrupted using clock-aware logic for time-based do-not-disrupt annotations
+// It considers both boolean ("true") and duration-based values (e.g., "5m", "1h")
+// For duration-based values, it checks if the pod has been running longer than the specified duration
+// Invalid annotation formats are treated as if the annotation doesn't exist and an event is emitted
 // It checks whether the following is true for the pod:
 // - Has the `karpenter.sh/do-not-disrupt` annotation
 // - Is an actively running pod
```
jmdeal:
```suggestion
// IsDisruptable checks if a pod can be disrupted using clock-aware logic for time-based do-not-disrupt annotations.
// It considers both boolean ("true") and duration-based values (e.g., "5m", "1h").
// For duration-based values, it checks if the pod has been running longer than the specified duration.
// Invalid annotation formats are treated as if the annotation doesn't exist and an event is emitted.
// Non-active pods are always considered disruptable.
```

pkg/utils/pod/scheduling.go
```diff
@@ -182,11 +190,79 @@ func IsOwnedBy(pod *corev1.Pod, gvks []schema.GroupVersionKind) bool {
...
 }
 
-func HasDoNotDisrupt(pod *corev1.Pod) bool {
+// parseDoNotDisrupt parses the do-not-disrupt annotation value
+// Returns (valid, indefinite, duration) where:
+// - valid=false means the annotation format is invalid (not "true" and not a valid duration)
+// - indefinite=true means permanent protection (value is "true")
+// - duration is the parsed time duration for valid duration strings
+// Invalid formats, zero, or negative durations return valid=false
+func parseDoNotDisrupt(value string) (valid bool, indefinite bool, duration time.Duration) {
```
jmdeal: nit: I'd consider simplifying code flow like this, lmk what you think:
- Do the `true`check directly in `IsDoNotDisruptActive`. Only call `parseDoNotDisrupt` if the value is not `true`.
  - This eliminates the need for the three value return since you don't need to consider infinite
- Change the signature to `parseDoNotDisrupt(string) (time.Duration, error)`
  - This is more stylistically consistent with the rest of the codebase, and the error message can be propagated to the event to give more information about why the duration is invalid.

pkg/utils/pod/scheduling.go
```diff
@@ -182,11 +190,79 @@ func IsOwnedBy(pod *corev1.Pod, gvks []schema.GroupVersionKind) bool {
...
+// For duration-based values, it checks if the pod has been running longer than the specified duration
+// Invalid annotation formats are treated as if the annotation doesn't exist (returns false)
+// and an event is emitted if a recorder is provided
+func IsDoNotDisruptActive(pod *corev1.Pod, clk clock.Clock, recorder events.Recorder) bool {
 	if pod.Annotations == nil {
 		return false
 	}
-	return pod.Annotations[v1.DoNotDisruptAnnotationKey] == "true"
+
+	value, exists := pod.Annotations[v1.DoNotDisruptAnnotationKey]
```
jmdeal: nit: `ok` is the standard value for existence checks, like `i` is for loop iteration

test/suites/regression/integration_test.go
```diff
@@ -423,4 +423,323 @@ var _ = Describe("Integration", func() {
 			})
 		})
 	})
+	Describe("DoNotDisrupt", func() {
+		Context("Grace Period", func() {
+			var dep *appsv1.Deployment
+			var selector labels.Selector
```
jmdeal: Is there any reason these are both scoped to the whole context when their being assigned on a per-test basis?

test/suites/regression/integration_test.go
```diff
@@ -423,4 +423,323 @@ var _ = Describe("Integration", func() {
...
+				smallSelector := labels.SelectorFromSet(smallDep.Spec.Selector.MatchLabels)
+				env.ExpectCreated(nodeClass, nodePool, largeDep, smallDep)
+
+				// Both pods should be scheduled to the same node
+				env.EventuallyExpectCreatedNodeCount("==", 1)
+				_ = env.EventuallyExpectHealthyPodCount(largeSelector, 1)
+				smallPods := env.EventuallyExpectHealthyPodCount(smallSelector, 1)
+
+				// Delete the large deployment to trigger consolidation
+				env.ExpectDeleted(largeDep)
```
jmdeal: Rather than relying on mechanisms like consolidation to handle detection, I'd be more direct and delete the NodeClaim directly. This is simpler to orchestrate, improving test readability. Rather than measuring how long it takes for the node to be disrupted, we'd instead be measuring how long it takes for the pod to be drained.

I agree it's a good idea to test this with the other disruption methods as well, but I'd consider handling that in the functional tests for consolidation and drift respectively, rather than through an integration test. I think as far as integration tests go, one is sufficient. Something similar to what you have further down with multiple grace periods. We could test three pods: two with different durations, and one with `true`. Test that the two with different durations only get drained once those durations elapse and the one with `true` only gets drained once the annotation is removed.

pkg/utils/pod/scheduling.go
```diff
@@ -182,11 +185,66 @@ func IsOwnedBy(pod *corev1.Pod, gvks []schema.GroupVersionKind) bool {
...
+
+	value, ok := pod.Annotations[v1.DoNotDisruptAnnotationKey]
+	if !ok {
+		return false
+	}
+
+	if value == "true" {
+		return true
+	}
+
```
moko-poi, line 219: The RFC explicitly states that invalid values should be treated as **indefinite protection (fail-safe)**:

> Invalid Values: If the value cannot be parsed as either "true" or a valid duration, it is treated as indefinite protection (backward compatible, fail-safe behavior)

And the pseudocode in the RFC shows:
```go
if err != nil {
    // Invalid format - treat as indefinite (fail-safe)
    return true, 0, nil
}
if d <= 0 {
    // Zero or negative - treat as indefinite (fail-safe)
    return true, 0, nil
}
```

The current implementation does the opposite — it treats invalid/zero/negative values as "annotation doesn't exist" (pod is disruptable). This means if a user typos `"4hours"` instead of `"4h"`, their pod loses all protection silently (aside from the event).

Was this an intentional deviation from the RFC? If so, I think the RFC should be updated to reflect this decision. If not, the fail-safe behavior from the RFC seems safer to protect users from misconfiguration.

pkg/utils/pod/scheduling.go
```diff
@@ -182,11 +185,66 @@ func IsOwnedBy(pod *corev1.Pod, gvks []schema.GroupVersionKind) bool {
...
+	duration, err := parseDoNotDisrupt(value)
+	if err != nil {
+		// Invalid format - emit event and treat as if annotation doesn't exist
+		if recorder != nil {
+			recorder.Publish(InvalidDoNotDisruptAnnotationEvent(pod, err.Error()))
+		}
+		return false
+	}
```
moko-poi, line 227: The RFC specifies that the grace period should be calculated from **pod binding time** (`pod.BoundTimestamp`), not `pod.Status.StartTime`:

> Duration string: Protection for the specified duration from pod binding time (e.g., "4h", "30m", "1h30m")

> expiration_time = pod.BoundTimestamp + parsed_duration

`StartTime` is set by the kubelet after it starts running the pod, which can be later than when the pod was bound to a node. While the difference is usually small, using a pod condition with type `PodScheduled` and status `True` would more accurately reflect the binding time as described in the RFC.

That said, if the team prefers `StartTime` for simplicity, it might be worth updating the RFC to match.

youwalther65: I feel, for end users choosing `pod.Status.StartTime` is more relevant in terms of application lifecycle.
Considering EC2 API throttling, large number of pending pods, time to provision large number of nodes during pod churn etc. using `pod.BoundTimestamp`  could be the wrong metric.

AndrewMitchell25: Yeah, this is why I chose `pod.Status.StartTime`, because the pod can be bound a while before actually running, which is probably when the end user wants to start the grace period. I'll update the RFC accordingly.

pkg/utils/pod/scheduling.go
AndrewMitchell25: Yes, this was an intentional deviation, and I can update the RFC accordingly. We decided on this implementation because of the current behavior of karpenter, which only considers if the annotation is "true". Setting the annotation to anything else would be an invalid value and karpenter would treat the pod as if it didn't have the annotation. We thought it better to continue this behavior with the new grace period. If this becomes an issue for a lot of users in the future we can discuss changing it, but for now we've decided to stick with the current behavior.

PR #2875, opened 2026-02-27
chore: improve naming of consolidationTTL

grosser:
this should make it clearer what is going on ... from initial reading I had the impression that a command would be invalid if it is older than 15s, but it is just a `sleep`

**Description**


grosser:
@jmdeal can you retry the failed CI, it's coveralls 502 again 😭

jmdeal:
I'll bypass and merge, not much we can do about a service error from coveralls

PR #2881, opened 2026-02-28
docs: Add guide for creating a new Karpenter provider

stuart23:
Fixes #N/A 

**Description**

Adds a dev guide for creating a karpenter provider for new kubernetes providers. Right now, the FAQs [mention](https://karpenter.sh/v1.7/faq/#can-i-write-my-own-cloud-provider-for-karpenter) that this is not a documented process - hopefully adding a guide will help people like me who want to run Karpenter on other platforms.

Docs only

### DIFF
--- a/NEW_PROVIDER.md
+++ b/NEW_PROVIDER.md
@@ -1,0 +1,90 @@
+# Creating a New Karpenter Provider
+
+A step-by-step guide to getting started with a new Karpenter cloud provider implementation.
+
+---
+
+## 1. Understand the Karpenter Architecture
+
+- Review the [Karpenter core repository](https://github.com/kubernetes-sigs/karpenter) to understand the separation between **core logic** (scheduling, deprovisioning, drift) and **cloud provider–specific logic**.
+- Study an existing provider (e.g., [karpenter-provider-aws](https://github.com/aws/karpenter-provider-aws)) as a reference implementation.
+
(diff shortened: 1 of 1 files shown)


jmdeal:
I don't know if we want to document this process before more work is done to stabilize / clean-up the cloudprovider interface. Right now some of the "contracts" between the cloudprovider and the upstream library aren't well enforced through the type-system, and there are a number of global overrides that are difficult to discover. These are the next best next steps for improving this documentation:
- Identifying all of the API surface that cloudproviders actually need to interact with (e.g. globals, the cloudprovider interface, etc) and introducing a proposal to consolidate this API surface
- Documenting contracts between the library and the API that can't properly be expressed via go's type system (or ideally updating the types to enforce it).
- Identifying which packages we can mark internal to reduce potential cloudprovider dependencies, and identifying anr required refactors

Only after we take these steps does it make sense to start putting together documentation since right now you need to be quite familiar with Karpenter's upstream implementation to write a provider successfully. I'm going to close this PR out for the time being for that reason.

stuart23:
OK @jmdeal . That's a bit too much work for me to take on. The doc was very useful for myself and I thought it could be useful for the community as well. Hopefully cementing the contracts and APIs will happen in the future so implementing a new provider can be documented.

Thanks for considering my PR.

PR #2892, opened 2026-03-05
feat(disruption): Add pod disruption tracking to perf test reporting

nathangeology:
## Pod Disruption Tracking for Performance Test Reporting
 Updates the performance test reporting framework to use this metric for tracking pods disrupted during benchmark runs.

### Changes

- Update performance test reporting to capture per-phase pod disruption counts using `karpenter_pods_drained_total`
- Add `PodsDisrupted` field to `PerformanceReport` struct with baseline-delta tracking

Performance benchmarking + `make verify`


jmdeal:
I think the drained metric actually could meet your use-case. The "drain reason" is a superset of the disruption reason, since it also covers forceful termination methods defined by cloudproviders (e.g. interruption). You can see where it's set for voluntary disruption here:

https://github.com/kubernetes-sigs/karpenter/blob/4e80a7e534d0c10674c0be835c1013c80e8b6dfe/pkg/controllers/node/termination/terminator/eviction.go#L213-L215

https://github.com/kubernetes-sigs/karpenter/blob/4e80a7e534d0c10674c0be835c1013c80e8b6dfe/pkg/controllers/node/termination/terminator/eviction.go#L223-L238

https://github.com/kubernetes-sigs/karpenter/blob/4e80a7e534d0c10674c0be835c1013c80e8b6dfe/pkg/controllers/disruption/queue.go#L268

I think it would be reasonable to add a NodePool label to this existing metric. There could be additional value to adding a "disrupted" metric in addition to a drained metric, but unless it's necessary I'd lean away from it to avoid confusion. Disrupted vs drained isn't clear to an end-user who isn't familiar with Karpenter's internals.

ryan-mist (approved):
/lgtm

DerekFrank (approved):
/lgtm 🚀

AndrewSirenko (approved):

DerekFrank (approved):
/lgtm

PR #2894, opened 2026-03-05
feat: Add feature-gated pod deletion cost management controller

nathangeology:
## Pod Deletion Cost Management Controller

**RFC:** #2935
**Status:** Implementation aligned with RFC.

### What this PR does

Adds a feature-gated controller (`PodDeletionCostManagement`) that manages `controller.kubernetes.io/pod-deletion-cost` annotations on pods running on Karpenter-managed nodes. The controller ranks nodes by consolidation priority and annotates pods so that ReplicaSet scale-down concentrates deletions on nodes Karpenter wants to consolidate.

This PR also introduces `karpenter.sh/disruption-cost` as the user-facing annotation for steering Karpenter consolidation. Customers who today set `controller.kubernetes.io/pod-deletion-cost` to influence consolidation should migrate to the new annotation.

### How it works

1. Reconciles every 60 seconds. ConsolidationState timestamp comparison skips cycles when nothing changed (zero API writes in steady state).
2. Partitions nodes into four groups:
   - **Group A (disrupted + PDB-blocked):** Gets MinInt32. Not counted against budget.
   - **Group B (drifted):** Bounded by NodePool's drift disruption budget.
   - **Group C (normal):** Bounded by NodePool's consolidation disruption budget.
   - **Group D (do-not-disrupt / no-budget):** Annotations cleared.
3. Annotates pods on top 50 ranked nodes (hard cap per cycle).
4. Writes `controller.kubernetes.io/pod-deletion-cost` directly. There is no overwrite-protection for customer-set PDC values: enabling the gate is the operator's opt-in to Karpenter managing that annotation. Customers steering consolidation are expected to be on `karpenter.sh/disruption-cost` by then.

### Feature gate

```
--feature-gates=PodDeletionCostManagement=true
```

**Gate ON:** controller writes `controller.kubernetes.io/pod-deletion-cost`; Karpenter's consolidation scoring reads only `karpenter.sh/disruption-cost`.

**Gate OFF (default):** controller does not run; Karpenter's consolidation scoring reads `karpenter.sh/disruption-cost` first and falls back to `controller.kubernetes.io/pod-deletion-cost` if the new annotation is absent. This preserves current behavior during the migration window.

### Key design decisions

- **Per-NodePool budget limiting:** Group B bounded by drift budget, Group C by consolidation budget. Overflow goes to Group D.
- **MinInt32 for Group A:** Disrupted+blocked nodes always get absolute lowest cost, not counted against any budget.
- **Group D clearing:** Annotations removed (not set high). RS controller applies default behavior.
- **`karpenter.sh/disruption-cost`:** User-facing annotation for consolidation steering. The auto-managed `pod-deletion-cost` is not interpreted as consolidation cost when the gate is on.
- **Forced cutover migration:** RFC v4 anchors migration to the next Kubernetes minor release. No multi-annotation coexistence protocol.

### What's included

- `ranking.go` - Four-tier ranking with budget limiting
- `annotation.go` - Pod annotation writes (no overwrite-protection per v4)
- `controller.go` - Reconciler with ConsolidationState change detection, budget enforcement, 50-node cap
- `queue.go` - Controller-runtime-backed fire-and-forget queue for per-pod annotation writes (modeled after `terminator.Queue`)
- `metrics.go` - Prometheus metrics for ranking duration, pods updated, cycles skipped
- `events.go` - Kubernetes events for ranking completion and failures
- `pkg/utils/disruption/disruption.go` - `karpenter.sh/disruption-cost` annotation with feature-gate-aware EvictionCost precedence
- Full unit test suite

### Follow-up work (not in this PR)

- Topology-aware ranking factors (Beta graduation criteria)
- Cleanup-on-disable (Possible Beta deliverable)
- Sparse sorting to save on annotations during re-ordering

### AI Disclosure

I used an LLM to parallel program with me on this. The original controller (Pod Deletion Cost Management) and its earlier revisions were drafted collaboratively over multiple sessions.


pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	if !c.cluster.Synced(ctx) {
+		return reconciler.Result{RequeueAfter: time.Second}, nil
+	}
+
+	// DeepCopyNodes matches the disruption controller convention: it takes
+	// a snapshot under the cluster mutex so downstream calls that outlive
+	// the iterator can't observe torn state. The alternative is iterating
+	// c.cluster.Nodes() and appending pointers, which relies on an implicit
+	// invariant that state.Cluster never mutates a StateNode in place.
+	nodes := c.cluster.DeepCopyNodes()
```
ryan-mist: Right now we'll deep copy the nodes twice (once here and once in `NodePoolStats`), can we try to avoid that since deep copies are pretty expensive?

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,195 @@
...
+		reconcileSkippedTotal.Add(1, noLabels)
+		return true
+	}
+	c.lastConsolidationState = currentState
+	return false
+}
+
+// buildNodePoolMap lists all managed NodePools and returns a map keyed by name.
+func (c *Controller) buildNodePoolMap(ctx context.Context) (map[string]*v1.NodePool, error) {
+	// TODO(maintainers): both this controller and disruption/controller.go list NodePools
```
ryan-mist: kube clients lists already hit the informer cache :)

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	nodePoolMap, err := c.buildNodePoolMap(ctx)
+	if err != nil {
+		return reconciler.Result{}, fmt.Errorf("building node pool map, %w", err)
+	}
+
+	nodeRanks, err := RankNodes(ctx, c.kubeClient, c.cluster, c.clock, nodes, nodePoolMap)
+	if err != nil {
+		return reconciler.Result{}, fmt.Errorf("ranking nodes, %w", err)
+	}
+	nodeRanks = capNodeRanks(nodeRanks, maxNodesPerCycle)
```
ryan-mist, line 137: Does moving this into `RankNodes` allow us to remove any unnecessary work?

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	// a snapshot under the cluster mutex so downstream calls that outlive
+	// the iterator can't observe torn state. The alternative is iterating
+	// c.cluster.Nodes() and appending pointers, which relies on an implicit
+	// invariant that state.Cluster never mutates a StateNode in place.
+	nodes := c.cluster.DeepCopyNodes()
+	if len(nodes) == 0 {
+		return reconciler.Result{RequeueAfter: reconcileInterval}, nil
+	}
+
+	if c.shouldSkipUnchanged(ctx) {
```
ryan-mist: We can move this above the deep copy to avoid unnecessary copies.

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,195 @@
...
+	"sigs.k8s.io/karpenter/pkg/controllers/state"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
+)
+
+const (
+	reconcileInterval = time.Minute
+	// maxNodesPerCycle bounds how many Group B/C/D nodes are annotated per
+	// reconcile. Group A nodes are exempt; see capNodeRanks.
+	maxNodesPerCycle = 50
```
ryan-mist, line 48: From RFC:
> API server write load: Each reconcile annotates pods on at most 50 nodes (hard cap). With an average of 30 pods/node, worst case is ~1,500 pod annotation writes per cycle. ConsolidationState-based change detection skips the cycle entirely when cluster state is unchanged (zero API writes in steady state). The change detection itself is O(1) with zero API calls.

Should we work backwards from pods instead? We have the `NodeRank.pods` already and pod density can vary a lot

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,195 @@
...
+		return nil, fmt.Errorf("listing node pools, %w", err)
+	}
+	return lo.SliceToMap(nodePools, func(np *v1.NodePool) (string, *v1.NodePool) { return np.Name, np }), nil
+}
+
+// capNodeRanks admits every Group A node (Rank == math.MinInt32) and caps the
+// remaining groups (B/C/D) at limit. Group A nodes are already tainted for
+// disruption and expected to be stable once labeled, so labeling churn stays
+// bounded even when Group A exceeds limit.
+func capNodeRanks(nodeRanks []NodeRank, limit int) []NodeRank {
```
ryan-mist, line 214: We could pre-filter out nodes where the pod rankings don't need to change - that way we don't hold a slot for a node where no pod annotations will actually change

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,330 @@
...
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"sigs.k8s.io/karpenter/pkg/metrics"
+)
+
+// podPatchWorkers caps the parallel-dispatch fan-out for per-pod patches. At
+// 50 nodes per cycle and up to 30 pods per node this bounds concurrent
```
ryan-mist: nit: Up to makes it seem like max pod density, maybe average or typical?
```suggestion
// 50 nodes per cycle and an average of 30 pods per node this bounds concurrent
```

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+// applyRankToPod issues the patch for a single pod. Skips the API call
+// entirely when the pod's annotation is already at the desired value, and
+// retries the retryable-API-error subset (429, server timeout, unavailable)
+// with podPatchRetryBackoff. 429s intentionally do not fall through to the
+// controller-runtime rate limiter here; per-pod retry keeps the burden on
+// this individual pod instead of failing the whole cycle.
+func applyRankToPod(ctx context.Context, kubeClient client.Client, pod *corev1.Pod, value string) podOutcome {
+	if !needsUpdate(pod, value) {
```
ryan-mist: nit: we can inline this


```suggestion
	if pod.Annotations[corev1.PodDeletionCost] != value
```

pkg/utils/disruption/disruption.go
```diff
@@ -76,6 +119,19 @@ func ReschedulingCost(ctx context.Context, pods []*corev1.Pod) float64 {
...
 }
 
+// podDeletionCostManagementEnabled reports whether the PodDeletionCostManagement feature
+// gate is enabled on the ctx's options. When options are absent (e.g. tests that construct
+// raw contexts without going through the operator injector), the gate is treated as its
+// default OFF value so the legacy pod-deletion-cost fallback path remains callable without
+// ceremony. Prod call sites always inject options via operator.Runtime, so this default
+// only affects call paths that never had a chance to reach the gate-ON branch.
+func podDeletionCostManagementEnabled(ctx context.Context) bool {
+	if !options.HasContext(ctx) {
```
ryan-mist: Is this needed? Ideally we want to avoid doing this to ensure our unit tests catch any bugs/regressions

kwok/charts/templates/clusterrole.yaml
```diff
@@ -68,7 +68,7 @@ rules:
     verbs: ["create"]
   - apiGroups: [""]
     resources: ["pods"]
-    verbs: ["delete"]
+    verbs: ["delete", "update", "patch"]
```
ryan-mist: Do we only need `patch` here?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,405 @@
...
+	drifted, driftOverflow = applyPerNodePoolBudget(drifted, driftBudget)
+	normal, normalOverflow = applyPerNodePoolBudget(normal, consolidationBudget)
+	doNotDisrupt = append(doNotDisrupt, driftOverflow...)
+	doNotDisrupt = append(doNotDisrupt, normalOverflow...)
+
+	// Group A nodes get math.MinInt32 and do not consume the contiguous rank
+	// space below zero. The remaining groups receive sequential ranks starting
+	// at -(B+C+D) so that drift and normal sort first under
+	// PodDeletionCost-ascending semantics.
+	remaining := len(drifted) + len(normal) + len(doNotDisrupt)
```
ryan-mist: Should we rank nodes with do not disrupt? It might seem a little off to have gaps in the annotations

e.g. if we have 10 Nodes with `DoNotDisrupt` and 10 without it we'll see pod annotations from -20 to -11 and nothing from -10 to 0

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,405 @@
...
+	if anyDisrupted(nodes) {
+		pdbs, err = fetchPDBs(ctx, kubeClient)
+		if err != nil {
+			return nil, fmt.Errorf("listing pod disruption budgets, %w", err)
+		}
+	}
+
+	disruptedBlocked, drifted, normal, doNotDisrupt := partitionNodes(nodes, nodePoolMap, nodePods, pdbs)
+
+	sortByPodCount(disruptedBlocked, nodePods)
```
ryan-mist: nit: since partitionNodes preserves ordering can we sort beforehand? It'll make it a bit easier to read IMO

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,405 @@
...
+		out[node.Name()] = pods
+	}
+	return out, nil
+}
+
+// fetchPDBs lists every PDB and pre-parses its selector. Malformed selectors
+// fail closed (the PDB is treated as matching every pod in its namespace) so
+// the controller errs on the side of classifying nodes as PDB-blocked rather
+// than ignoring potentially-blocking PDBs.
+func fetchPDBs(ctx context.Context, kubeClient client.Client) ([]parsedPDB, error) {
```
ryan-mist: Any reason not to use `pdb.NewLimits`?

pkg/utils/disruption/disruption.go
```diff
@@ -27,7 +27,23 @@ import (
...
+	// the cost of evicting a pod during consolidation. Customers set this on workloads
+	// to influence which pods Karpenter prefers to evict during consolidation.
+	//
+	// This is a new public annotation introduced alongside the PodDeletionCostManagement
+	// feature. No prior Karpenter release defined an annotation at this key in the
+	// karpenter.sh API group, so workloads upgrading from earlier versions are
+	// unaffected by default. Operators who choose this key for other purposes on a
+	// workload will have their value parsed as a consolidation cost; see the
+	// deletioncost README "Migration" section for the audit step.
+	DisruptionCostAnnotation = apis.Group + "/disruption-cost"
```
ryan-mist: Usually we have annotations and labels in https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/apis/v1/labels.go

pkg/controllers/pod/deletioncost/annotation.go
nathangeology: Applied in 52ec5b81 — reworded to "average of 30 pods per node".

pkg/controllers/pod/deletioncost/annotation.go
nathangeology: Applied in 52ec5b81 — dropped the needsUpdate helper and inlined the equality check.

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Good point. Kept the 50-node cap but added a pre-filter (see reply on L179 / 4073ac61) so the cap now bounds nodes-that-mutate rather than nodes-considered. With ~30 pods/node this keeps worst-case pod writes near the RFC's 1,500-writes target while the steady state (annotations already in place) reconciles for essentially free.

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Fixed in 3c337cc7 — added disruption.NodePoolStatsFromNodes that operates on a caller-supplied snapshot, and RankNodes uses it, so the reconcile now pays for exactly one DeepCopyNodes per cycle.

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Applied in 52ec5b81 — shouldSkipUnchanged now runs immediately after cluster.Synced and before DeepCopyNodes.

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Don't think that helps us avoid the work here. Both callers of nodePoolMap (partitionNodes for Group D routing, buildBudgetForReason for the budget calc) always need the full map when there are any candidates, so moving buildNodePoolMap into RankNodes wouldn't drop any work. The map is unconditionally required whenever RankNodes gets past its empty-node early return. Happy to reconsider if you have a specific work-elimination in mind.

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Cool, still figuring out the informer cache nuances here. I deleted the TODO in 52ec5b81. controller-runtime's client.Client.List already goes through the shared informer cache.

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Great suggestion, implemented in 4073ac61. New filterNoOpNodes helper walks nodeRanks before the cap and drops non-Group-A entries whose pods already carry the desired annotation state (matching rank for B/C or no annotation for D). Group A stays exempt. Unit tests in cap_internal_test.go cover the four cases: rank match drop, Group A always admitted, Group D still-annotated pod kept, empty pod list is no-op.

pkg/controllers/pod/deletioncost/ranking.go
nathangeology: Applied in f778f006 — single sortByPodCount(nodes, nodePods) at the top; partitionNodes preserves the order for each partition.

pkg/controllers/pod/deletioncost/ranking.go
nathangeology: Fixed in f778f006 — Group D nodes no longer consume rank space, since UpdatePodDeletionCosts clears their annotations rather than reading Rank. The annotated pods on Groups B and C now span a contiguous range with no gap.

pkg/controllers/pod/deletioncost/ranking.go
nathangeology: Switched to pdb.NewLimits + pdb.Limits.CanEvictPods in 3c337cc7. Two behavior differences to flag: (a) PDBs with unparseable selectors now bubble up an error instead of failing-closed as match-everything (apiserver validation rejects such selectors so it can't happen in prod), and (b) pods not eligible for the eviction API (terminal, DaemonSet) no longer count as PDB-blocked, matching Karpenter's actual disruption semantics.

pkg/utils/disruption/disruption.go
nathangeology: Moved to pkg/apis/v1/labels.go in 1b0f20e2 as DisruptionCostAnnotationKey to match the naming shape of the other annotation keys already there.

pkg/utils/disruption/disruption.go
nathangeology: Reverted in 798d47ec — dropped the HasContext guard and fixed the underlying cause (a balanced-scoring test's makeCandidate was passing context.Background() to computeRescheduleDisruptionCost; now it injects a default options context like the other tests in the file). EvictionCost trusts the caller to inject options via operator.Runtime, so a future missing-injection panics loudly instead of silently taking the gate-OFF fallback.

kwok/charts/templates/clusterrole.yaml
nathangeology: Applied in 52ec5b81 — dropped the update verb; the controller only issues merge patches (patchAnnotation / clearAnnotation), never plain updates.

REBASE_NOTES.md
```diff
@@ -0,0 +1,40 @@
+# Rebase Notes — cleanup/pr-2894-llm-hits
```
ryan-mist: nit: we can drop this

pkg/controllers/pod/deletioncost/README.md
```diff
@@ -0,0 +1,222 @@
+# Pod Deletion Cost Controller
```
ryan-mist: Can we move this to `/designs`? Or should we drop this in favor of the RFC?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,380 @@
...
+		currentRank++
+	}
+	for _, node := range doNotDisrupt {
+		// Rank is unused for Group D — UpdatePodDeletionCosts sees
+		// HasDoNotDisrupt=true and clears the annotation rather than reading
+		// Rank. Leave Rank at its zero value.
+		result = append(result, NodeRank{Node: node, HasDoNotDisrupt: true, Pods: nodePods[node.Name()]})
+	}
+
+	nodesRanked.Set(float64(len(result)), noLabels)
```
ryan-mist: Should this track the number of nodes we actually end up ranking - i.e. after `filterNoOpNodes` and  `capNodeRanks`?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,380 @@
...
+
+// hasNonRSOwnedPods returns true if any non-system pod on the node is owned by
+// a controller other than ReplicaSet/Job — e.g. StatefulSet, DaemonSet, raw
+// Pod. The RFC's Group A definition includes these because their replacement
+// path is more disruptive than evicting a ReplicaSet pod (StatefulSet pods
+// have ordinal identity and persistent volumes; bare pods cannot be recreated
+// at all). DaemonSet pods are excluded because they're tied to the node and
+// will be replaced anyway when the node is replaced — they don't make the
+// node "expensive to delete".
+func hasNonRSOwnedPods(pods []*corev1.Pod) bool {
```
ryan-mist, line 327: Can we add a test for non-replicaset-owned pods in `ranking_test.go`?

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,317 @@
...
+		totals.add(stats)
+		if perr != nil {
+			aggErr = multierr.Append(aggErr, perr)
+		}
+	}
+
+	podsUpdatedTotal.Add(float64(totals.updated), map[string]string{resultLabel: "updated"})
+	podsUpdatedTotal.Add(float64(totals.skipped), map[string]string{resultLabel: "skipped_unchanged"})
+	podsUpdatedTotal.Add(float64(totals.podErrors), map[string]string{resultLabel: "error"})
+	nodesErroredTotal.Add(float64(totals.nodeErrors), noLabels)
```
ryan-mist: Is `totals.nodeErrors` always 0?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,379 @@
...
+//   - Group D: node-level do-not-disrupt annotation, do-not-disrupt pods, or
+//     NodePool with consolidation disabled (ConsolidateAfter=Never).
+func partitionNodes(clk clock.Clock, nodes []*state.StateNode, nodePoolMap map[string]*v1.NodePool, nodePods map[string][]*corev1.Pod, pdbs pdb.Limits) (disruptedBlocked, drifted, normal, doNotDisrupt []*state.StateNode) {
+	for _, node := range nodes {
+		pods := nodePods[node.Name()]
+		// Group A first — any of the three Group A signals routes here
+		// regardless of do-not-disrupt signals on the node itself or its
+		// pods. RFC §"Group A" calls for OR semantics across all three
+		// predicates: A || B || C, not (A && B) || C.
+		if isDisrupted(node) || hasPDBBlockedPods(clk, pods, pdbs) || hasNonRSOwnedPods(pods) {
```
jigisha620: Here the Group A check  runs before the `do-not-disrupt` check. This means a node with `karpenter.sh/do-not-disrupt: true` that also has a StatefulSet pod gets routed to Group A and its deployment pods receive MinInt32 deletion cost. The customer explicitly marked the node as "don't touch," but we're telling the RS controller to preferentially delete pods from it first. The rationale for non-RS pods in Group A is "help Karpenter empty a node it will eventually consolidate," but Karpenter will never consolidate a do-not-disrupt node, so does steering RS toward it provides any benefit here?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,379 @@
...
+// tie-break. Nodes whose pod-list lookup is missing from the map sort to the
+// end (math.MaxInt sentinel) so a transient lookup failure doesn't skew the
+// top of the ranking — the missing-lookup case should be impossible on the
+// happy path because RankNodes pre-populates the map with one entry per
+// candidate node, but the sentinel is correct-by-construction defense.
+func sortByPodCount(nodes []*state.StateNode, nodePods map[string][]*corev1.Pod) {
+	if len(nodes) <= 1 {
+		return
+	}
+	sort.SliceStable(nodes, func(i, j int) bool {
```
jigisha620: nit: Is SliceStable needed here? Since node names are unique, the name tie-break means no two elements ever compare as equal, so stability doesn't come into play.

nathangeology: Looks like we don't need slice stable when I looked into it more. Good callout.

pkg/controllers/pod/deletioncost/ranking.go
nathangeology: Good catch, I made some changes to make this a little more straighforward to think through. Let me know the new if block makes sense to you or if we need to talk through it some more.

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,251 @@
...
+// Uses the same ConsolidationState timestamp that gates the disruption controller's consolidation
+// methods, ensuring this controller reacts to the same state changes that trigger consolidation.
+func (c *Controller) shouldSkipUnchanged(ctx context.Context) bool {
+	currentState := c.cluster.ConsolidationState()
+	if currentState.Equal(c.lastConsolidationState) {
+		log.FromContext(ctx).V(1).Info("no changes detected, skipping pod deletion cost update")
+		reconcileSkippedTotal.Add(1, noLabels)
+		return true
+	}
+	c.lastConsolidationState = currentState
```
jigisha620: Shouldn't `c.lastConsolidationState = currentState` be done after successful ranking and annotating? If there is some failure after the `shouldSkipUnchanged` is called then this may cause the reconcile to skip right?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,345 @@
...
+	return false
+}
+
+// sortByPodCount sorts nodes by pod count ascending with a deterministic name
+// tie-break. Nodes whose pod-list lookup is missing from the map sort to the
+// end (math.MaxInt sentinel) so a transient lookup failure doesn't skew the
+// top of the ranking — the missing-lookup case should be impossible on the
+// happy path because RankNodes pre-populates the map with one entry per
+// candidate node, but the sentinel is correct-by-construction defense.
+func sortByPodCount(nodes []*state.StateNode, nodePods map[string][]*corev1.Pod) {
```
jigisha620: `fetchNodePods` would get all the pods, even those that are deleting. Wondering if deleting pods on a node should be considered for ranking that node higher?

pkg/controllers/pod/deletioncost/ranking_test.go
```diff
@@ -0,0 +1,1033 @@
...
+		// rather than the disabled/budget-overflow paths.
+		nodePool.Spec.Disruption.ConsolidateAfter = v1.MustParseNillableDuration("0s")
+		nodePool.Spec.Disruption.Budgets = []v1.Budget{{Nodes: "100%"}}
+	})
+
+	// Migrated happy-path tests drive through Controller.Reconcile and assert
+	// on the observable pod-deletion-cost annotation. Direct-helper tests for
+	// partition edge cases remain in this file for the cases where the
+	// observable annotation does not distinguish the classification.
+	Context("Two-tier partitioning", func() {
```
jigisha620, line 100: Can these tests be single table driven tests? If there are more opportunities, elsewhere in this PR where tests could be combined into a table then can you update those as well?

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,301 @@
...
+	return foldOutcomes(outcomes)
+}
+
+// clearRankFromPod issues the clear patch for a single pod. Returns
+// outcomeSkipped for pods without the annotation and for NotFound/Conflict
+// responses; outcomeUpdated when a patch fired successfully; outcomeErrored
+// otherwise. Symmetric with applyRankToPod: NotFound/Conflict classification
+// happens here in the caller so we can count skips and log at V(1) alongside
+// the write path.
+func clearRankFromPod(ctx context.Context, kubeClient client.Client, pod *corev1.Pod) podOutcome {
```
jigisha620: nit: The code here is pretty much similar to what happens in `applyRankToPod`. Could share a generic helper function?

jigisha620 (commented):
Sorry for the delayed review. Very excited to see this PR get merged. One high level nit - the comments in the PR are pretty thorough and very helpful in understanding decisions that can't be easily inferred. But some of the per-line comments or function descriptions restate what's readable from the code. Do you think it makes sense to drop some of those?

pkg/controllers/pod/deletioncost/cap_internal_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+func ranks(pattern ...int) []NodeRank {
+	out := make([]NodeRank, 0, len(pattern))
+	for _, r := range pattern {
+		out = append(out, NodeRank{Rank: r})
+	}
+	return out
+}
+
+func TestCapNodeRanks_ReturnsAllGroupA_CapsTail(t *testing.T) {
```
jigisha620: nit: I don't feel super strongly about this but it feels like majority of the codebase uses Ginkgo for writing tests (including unit tests). Would it make sense to just align with that? I could go either way on this one as long as other maintainers are aligned.

nathangeology: Should be resolved now, should use ginkgo

pkg/controllers/pod/deletioncost/annotation.go
nathangeology: Made one, let me know if it looks better now

pkg/controllers/pod/deletioncost/ranking.go
nathangeology: Yep, good catch. We probably don't want to consider those in most cases. Went ahead and added a function to distinguish when we want all pods including deleting ones vs not

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Another good catch, should be fixed now!

pkg/apis/v1/labels.go
```diff
@@ -57,6 +57,18 @@ const (
 	// scheduled to this NodeClaim. The initialization controller can gate on these drivers having published their
 	// ResourceSlices before marking the node initialized.
 	DRADriversAnnotationKey = apis.Group + "/requested-dra-drivers"
+	// DisruptionCostAnnotationKey is the user-facing Karpenter annotation for
```
DerekFrank, line 60: https://github.com/kubernetes-sigs/karpenter/issues/3261 :D

pkg/controllers/disruption/helpers.go
```diff
@@ -255,15 +255,22 @@ func BuildNodePoolMap(ctx context.Context, kubeClient client.Client, cloudProvid
...
+// controller so the two stay in lockstep on which nodes count toward each budget.
+func NodePoolStats(cluster *state.Cluster) (numNodes, disrupting map[string]int) {
+	return NodePoolStatsFromNodes(cluster.DeepCopyNodes())
+}
+
+// NodePoolStatsFromNodes computes the same per-NodePool counts as NodePoolStats
+// against a caller-supplied snapshot. Callers that already hold a DeepCopy
+// (e.g. the deletion-cost controller) reuse it here instead of paying for a
+// second cluster-wide deep copy.
+func NodePoolStatsFromNodes(nodes []*state.StateNode) (numNodes, disrupting map[string]int) {
```
DerekFrank, line 270: hot take, shouldn't cluster.go just have the `disrupting map[string]int` field and keep it up to date? Probably outside the scope of this PR...

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,309 @@
...
+	"k8s.io/client-go/util/retry"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	"sigs.k8s.io/karpenter/pkg/metrics"
+)
+
+// podPatchWorkers caps the parallel-dispatch fan-out for per-pod patches. At
```
DerekFrank: We can probably cut a bunch of these comments and keep just the relevant out of code information like:

// deliberately low to prevent api server contention
// TODO: look into tuning on CPUs

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,257 @@
...
+	if c.consolidationStateUnchanged(ctx, currentState) {
+		return reconciler.Result{RequeueAfter: reconcileInterval}, nil
+	}
+
+	// DeepCopyNodes matches the disruption controller convention: it takes
+	// a snapshot under the cluster mutex so downstream calls that outlive
+	// the iterator can't observe torn state. The alternative is iterating
+	// c.cluster.Nodes() and appending pointers, which relies on an implicit
+	// invariant that state.Cluster never mutates a StateNode in place.
+	nodes := c.cluster.DeepCopyNodes()
```
DerekFrank: I don't think we need a deepcopy here, this controller seems best effort anyway

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,257 @@
...
+	if len(nodes) == 0 {
+		return reconciler.Result{RequeueAfter: reconcileInterval}, nil
+	}
+
+	nodePoolMap, err := c.buildNodePoolMap(ctx)
```
DerekFrank: If we do this and balanced consolidation, they should share a cache probably, but that can be a P1, lets cut an issue for that to keep track

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+}
+
+// dispatchPerPod fans op out across pods via runPodOp under a bounded worker
+// pool so a large per-node pod list does not blast the apiserver with a burst
+// of goroutines. Per-pod outcome classification (retry/skip/error) lives in
+// runPodOp; foldOutcomes aggregates. Bound is podPatchWorkers.
+func dispatchPerPod(ctx context.Context, pods []*corev1.Pod, op podOp) (podPatchStats, error) {
+	outcomes := make([]podOutcome, len(pods))
+	workqueue.ParallelizeUntil(ctx, podPatchWorkers, len(pods), func(i int) {
```
DerekFrank: Should we model this after the eviction queue? Its fire and forget

jigisha620 (commented):
High level comment that can be addressed in a follow-up is adding e2e tests for the functionality. I also think that we should analyze if the code increases memory footprint for karpenter.

pkg/controllers/disruption/helpers.go
jigisha620: Lesson learnt from one of the reviews on my older PR was that we shouldn't overload cluster state unless this is one of the things that absolutely needs to go in there.

pkg/apis/v1/labels.go
nathangeology: :)

pkg/utils/disruption/disruption.go
```diff
@@ -28,8 +28,56 @@ import (
...
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 )
 
+// PerNodeBaseDisruptionCost is the inherent cost of draining a node (cordon,
+// drain, API calls, replacement latency). Sets the minimum reschedule cost
+// so an "empty" node still costs something to disrupt. See
+// designs/balanced-consolidation.md.
+const PerNodeBaseDisruptionCost = 1.0
+
+// ResolveOfferingPrice returns the price of the offering that matches the
```
nathangeology: This comment can probably be cut

pkg/utils/disruption/disruption.go
```diff
@@ -45,26 +93,53 @@ func LifetimeRemaining(clock clock.Clock, nodePool *v1.NodePool, nodeClaim *v1.N
...
 		if err != nil {
-			log.FromContext(ctx).Error(err, "failed parsing pod deletion cost",
-				"annotation", corev1.PodDeletionCost, "value", podDeletionCostStr, "pod", client.ObjectKeyFromObject(p))
+			log.FromContext(ctx).Error(err, "failed parsing disruption cost",
+				"annotation", v1.DisruptionCostAnnotationKey, "value", costStr, "pod", client.ObjectKeyFromObject(p))
 		} else {
-			// the pod deletion disruptionCost is in [-2147483647, 2147483647]
-			// the min pod disruptionCost makes one pod ~ -15 pods, and the max pod disruptionCost to ~ 17 pods.
-			cost += podDeletionCost / math.Pow(2, 27.0)
+			// 2^27 = max representable pod-deletion-cost (int32 ceiling).
```
nathangeology: Not sure we want to remove the min disruption and max distruption cost comment in the old code's line 59

nathangeology (commented):
Leaving a few "comment clean up" comments for my bot helper

pkg/controllers/pod/deletioncost/controller.go
nathangeology: Took a stab at this, we can cut it and shift to a new PR if you want though

pkg/controllers/pod/deletioncost/annotation.go
nathangeology: done

pkg/utils/disruption/disruption.go
```diff
@@ -28,8 +28,51 @@ import (
...
+// so an "empty" node still costs something to disrupt. See
+// designs/balanced-consolidation.md.
+const PerNodeBaseDisruptionCost = 1.0
+
+func ResolveOfferingPrice(labels map[string]string, instanceType *cloudprovider.InstanceType) float64 {
+	if instanceType == nil {
+		return 0
+	}
+	price, ok := instanceType.OfferingPrice(labels[corev1.LabelTopologyZone], labels[v1.CapacityTypeLabelKey])
+	if !ok {
```
DerekFrank, line 46:
```suggestion
	if !ok || math.IsNaN(price) {
```

pkg/utils/disruption/disruption.go
```diff
@@ -28,8 +28,51 @@ import (
...
+	return cost
+}
+
+// SavingsRatio returns Price / RescheduleDisruptionCost (higher = prefer to
+// disrupt). Callers must guarantee a positive rescheduleDisruptionCost — use
+// ComputeRescheduleDisruptionCost which is floored at PerNodeBaseDisruptionCost.
+// Both the balanced-consolidation candidate sort and the pod-deletion-cost
+// ranking read this ratio so their orderings agree.
+func SavingsRatio(price, rescheduleDisruptionCost float64) float64 {
+	return price / rescheduleDisruptionCost
```
DerekFrank, line 73: maybe return an error if rescheduleDisruptionCost is 0? That or if we think it shouldn't hit that we panic and that is our way to catch it at testing?

pkg/utils/disruption/disruption.go
```diff
@@ -45,26 +88,52 @@ func LifetimeRemaining(clock clock.Clock, nodePool *v1.NodePool, nodeClaim *v1.N
...
+//     karpenter.sh/disruption-cost first; if absent, it falls back to
+//     controller.kubernetes.io/pod-deletion-cost. This preserves current behavior for
+//     customers who have not migrated to the new annotation.
 func EvictionCost(ctx context.Context, p *corev1.Pod) float64 {
 	cost := 1.0
-	podDeletionCostStr, ok := p.Annotations[corev1.PodDeletionCost]
-	if ok {
-		podDeletionCost, err := strconv.ParseFloat(podDeletionCostStr, 64)
+	if costStr, ok := p.Annotations[v1.DisruptionCostAnnotationKey]; ok {
+		parsedCost, err := strconv.ParseFloat(costStr, 64)
```
DerekFrank, line 108: Leaving a note here, this should probably be an int. I don't want people to get used to being able to set floats because we can always add float later if we need but I'd rather not remove specificity.

This is the reason we don't like annotations, we can't do strongly typed validation at resource creation

pkg/utils/disruption/disruption.go
```diff
@@ -76,6 +145,15 @@ func ReschedulingCost(ctx context.Context, pods []*corev1.Pod) float64 {
 	return cost
 }
 
+// podDeletionCostManagementEnabled reports whether the PodDeletionCostManagement
+// feature gate is enabled on the ctx's options. Prod call sites inject options
+// via operator.Runtime; tests that reach EvictionCost must do the same so that
+// a missing-options bug surfaces as a panic instead of a silently-wrong gate
+// reading.
+func podDeletionCostManagementEnabled(ctx context.Context) bool {
```
DerekFrank, line 153: lets flatten this, we don't need a helper func

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+		} else {
+			overflow = append(overflow, node)
+		}
+	}
+	return bounded, overflow
+}
+
+// fetchNodePods gathers the pod list per node so partitionNodes and
+// sortBySavingsRatio don't repeat the API call.
+func fetchNodePods(ctx context.Context, kubeClient client.Client, nodes []*state.StateNode) (map[string][]*corev1.Pod, error) {
```
DerekFrank, line 152: rip this out, we can just call stateNode.Pods() to get each node's pods

pkg/controllers/pod/deletioncost/controller.go
```diff
@@ -0,0 +1,227 @@
...
+	return reconciler.Result{RequeueAfter: reconcileInterval}, nil
+}
+
+// enqueueAnnotationWrites hands each pod's desired annotation state off to
+// the fire-and-forget Queue. The Queue's Reconcile method decides
+// per-pod whether to write or skip and handles retry via controller-runtime.
+func (c *Controller) enqueueAnnotationWrites(nodeRanks []NodeRank) {
+	for i := range nodeRanks {
+		nr := &nodeRanks[i]
+		for _, pod := range nr.Pods {
```
DerekFrank, line 159:
```suggestion
		for _, pod := range nr.node.Pods() {
```

pkg/controllers/pod/deletioncost/types.go
```diff
@@ -0,0 +1,33 @@
...
+	"sigs.k8s.io/karpenter/pkg/controllers/state"
+)
+
+// NodeRank is produced by RankNodes and consumed by the annotation Queue to
+// derive per-pod PDC values. HasDoNotDisrupt=true marks Group D entries whose
+// annotations get cleared rather than written; Rank is unused in that case.
+type NodeRank struct {
+	Node            *state.StateNode
+	Rank            int
+	HasDoNotDisrupt bool
```
DerekFrank, line 31: I think you can replace this with ValidatePodsDisruptable() && ValidateNodeDisruptable()

pkg/controllers/pod/deletioncost/types.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"sigs.k8s.io/karpenter/pkg/controllers/state"
+)
+
+// NodeRank is produced by RankNodes and consumed by the annotation Queue to
+// derive per-pod PDC values. HasDoNotDisrupt=true marks Group D entries whose
+// annotations get cleared rather than written; Rank is unused in that case.
+type NodeRank struct {
+	Node            *state.StateNode
+	Rank            int
```
DerekFrank, line 30: If we have the order of the nodes be the implicit rank, we cut this type out altogether!

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+// NodePoolStatsFromNodes reuse it so we don't pay for two cluster iterations.
+func RankNodes(ctx context.Context, kubeClient client.Client, clk clock.Clock, nodes []*state.StateNode, nodePoolMap map[string]*v1.NodePool, nodePoolToInstanceTypesMap map[string]map[string]*cloudprovider.InstanceType) ([]NodeRank, error) {
+	if len(nodes) == 0 {
+		return nil, nil
+	}
+	defer metrics.Measure(rankingDurationSeconds, noLabels)()
+
+	// Pre-fetch pods per node once so partitionNodes and sortBySavingsRatio
+	// don't repeat the API call.
+	nodePods, err := fetchNodePods(ctx, kubeClient, nodes)
```
DerekFrank, line 68: Probably don't need this if we use nodes.Pods() or nodes[0].Pods()

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("listing pods on candidate nodes, %w", err)
+	}
+	// PDB-blocked pods route to Group D on any candidate, tainted or not,
+	// so the cluster-wide PDB list is needed on every reconcile.
+	pdbs, err := pdb.NewLimits(ctx, kubeClient)
```
DerekFrank, line 74: can probably be ValidatePodsDisruptable() instead

pkg/utils/disruption/disruption.go
```diff
@@ -28,8 +28,51 @@ import (
...
+	"sigs.k8s.io/karpenter/pkg/operator/options"
 )
 
+// PerNodeBaseDisruptionCost is the inherent cost of draining a node (cordon,
+// drain, API calls, replacement latency). Sets the minimum reschedule cost
+// so an "empty" node still costs something to disrupt. See
+// designs/balanced-consolidation.md.
+const PerNodeBaseDisruptionCost = 1.0
+
+func ResolveOfferingPrice(labels map[string]string, instanceType *cloudprovider.InstanceType) float64 {
```
DerekFrank, line 41:
```suggestion
func ResolveOfferingPrice(instanceType *cloudprovider.InstanceType, zone, capacityType string) float64 {
```

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+// pods from is the same node the disruption controller would consolidate
+// first.
+//
+// Reschedulable pods (per pod.IsReschedulable) drive the disruption cost:
+// DaemonSet pods and RS-owned terminating pods contribute 0, so a
+// mostly-drained node scores higher than a same-priced node still carrying
+// its full pod set. The full pod list is retained on NodeRank so the queue
+// still writes/clears annotations on deleting pods until they finish
+// terminating.
+func sortBySavingsRatio(ctx context.Context, nodes []*state.StateNode, nodePods map[string][]*corev1.Pod, nodePoolToInstanceTypesMap map[string]map[string]*cloudprovider.InstanceType) {
```
DerekFrank, line 360:
```suggestion
func sortBySavingsRatio(ctx context.Context, nodes []*state.StateNode, nodePoolToInstanceTypesMap map[string]map[string]*cloudprovider.InstanceType) {
```

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+	ratio := make(map[string]float64, len(nodes))
+	for _, n := range nodes {
+		labels := n.Labels()
+		var it *cloudprovider.InstanceType
+		if m := nodePoolToInstanceTypesMap[labels[v1.NodePoolLabelKey]]; m != nil {
+			it = m[labels[corev1.LabelInstanceTypeStable]]
+		}
+		price := disruptionutils.ResolveOfferingPrice(labels, it)
+		reschedulable := lo.Filter(nodePods[n.Name()], func(p *corev1.Pod, _ int) bool { return podutils.IsReschedulable(p) })
+		cost := disruptionutils.ComputeRescheduleDisruptionCost(ctx, reschedulable)
```
DerekFrank, line 373:
```suggestion
		disruptioncost := disruptionutils.ComputeRescheduleDisruptionCost(ctx, reschedulable)
```

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+		return
+	}
+	ratio := make(map[string]float64, len(nodes))
+	for _, n := range nodes {
+		labels := n.Labels()
+		var it *cloudprovider.InstanceType
+		if m := nodePoolToInstanceTypesMap[labels[v1.NodePoolLabelKey]]; m != nil {
+			it = m[labels[corev1.LabelInstanceTypeStable]]
+		}
+		price := disruptionutils.ResolveOfferingPrice(labels, it)
```
DerekFrank, line 371:
```suggestion
		offeringprice := disruptionutils.ResolveOfferingPrice(labels, it)
```

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+	// PDB-blocked pods route to Group D on any candidate, tainted or not,
+	// so the cluster-wide PDB list is needed on every reconcile.
+	pdbs, err := pdb.NewLimits(ctx, kubeClient)
+	if err != nil {
+		return nil, fmt.Errorf("listing pod disruption budgets, %w", err)
+	}
+
+	sortBySavingsRatio(ctx, nodes, nodePods, nodePoolToInstanceTypesMap)
+
+	disruptedBlocked, drifted, normal, doNotDisrupt := partitionNodes(clk, nodes, nodePoolMap, nodePoolToInstanceTypesMap, nodePods, pdbs)
```
DerekFrank, line 81:
```suggestion
    nodePartitions := lo.GroupBy(nodes, func())
```

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+		return true
+	}
+	return false
+}
+
+// buildBudgetForReason returns allowed - already-disrupting per NodePool.
+// Negative results (more in-flight than allowed — should not occur in
+// steady-state) are logged and clamped to 0 so we don't move every candidate
+// to Group D.
+func buildBudgetForReason(ctx context.Context, nodePoolMap map[string]*v1.NodePool, numNodes, disrupting map[string]int, clk clock.Clock, reason v1.DisruptionReason) map[string]int {
```
DerekFrank, line 243: no!

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+	sortBySavingsRatio(ctx, nodes, nodePods, nodePoolToInstanceTypesMap)
+
+	disruptedBlocked, drifted, normal, doNotDisrupt := partitionNodes(clk, nodes, nodePoolMap, nodePoolToInstanceTypesMap, nodePods, pdbs)
+
+	// Per-NodePool budget limits move overflow from B/C into D.
+	// NodePoolStatsFromNodes is the shared helper disruption also uses (via
+	// NodePoolStats), so the two controllers count the same nodes against
+	// the same budget.
+	numNodes, disrupting := disruption.NodePoolStatsFromNodes(nodes)
+	driftBudget := buildBudgetForReason(ctx, nodePoolMap, numNodes, disrupting, clk, v1.DisruptionReasonDrifted)
```
DerekFrank, line 88: if we're just building budgets, re-use existing machinery :)

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+	consolidationBudget := buildBudgetForReason(ctx, nodePoolMap, numNodes, disrupting, clk, v1.DisruptionReasonUnderutilized)
+	var driftOverflow, normalOverflow []*state.StateNode
+	drifted, driftOverflow = applyPerNodePoolBudget(drifted, driftBudget)
+	normal, normalOverflow = applyPerNodePoolBudget(normal, consolidationBudget)
+	doNotDisrupt = append(doNotDisrupt, driftOverflow...)
```
DerekFrank, line 93: lets cut the overflow stuff, we can just leave the annotation on those

DerekFrank: Or at least, lets clean those annotations but at a different priority than the doNotDisrupt?

pkg/controllers/pod/deletioncost/ranking.go
```diff
@@ -0,0 +1,383 @@
...
+// regardless of any other signal.
+func partitionNodes(clk clock.Clock, nodes []*state.StateNode, nodePoolMap map[string]*v1.NodePool, nodePoolToInstanceTypesMap map[string]map[string]*cloudprovider.InstanceType, nodePods map[string][]*corev1.Pod, pdbs pdb.Limits) (disruptedBlocked, drifted, normal, doNotDisrupt []*state.StateNode) {
+	for _, node := range nodes {
+		pods := nodePods[node.Name()]
+		if isDisrupted(node) {
+			disruptedBlocked = append(disruptedBlocked, node)
+			continue
+		}
+		if hasNodeDoNotDisrupt(node) || hasPDBBlockedPods(clk, pods, pdbs) || hasNonRSOwnedPods(pods) {
+			doNotDisrupt = append(doNotDisrupt, node)
```
DerekFrank, line 177: Lets rename doNotDisrupt to something more like do not annotate. disruptionBlocked is the actual 'do-not-disrupt' annotation and we should keep those concept seperate

DerekFrank: Or maybe since we're also putting hasNonRSOwnedPods and isConsolidationDisabled in there, it should be 'cleanupOnly'?

pkg/controllers/pod/deletioncost/annotation.go
```diff
@@ -0,0 +1,51 @@
...
+
+// clearAnnotation removes pod-deletion-cost via merge-patch with optimistic
+// lock. MergeFromWithOptimisticLock is used so a concurrent writer of the
+// same annotation (customer kubectl, third-party HPAs, admission webhooks)
+// surfaces as a 409 Conflict — the queue's Reconcile then treats it as
+// Skipped and lets the next cycle converge. See
+// pkg/controllers/nodeclaim/lifecycle/controller.go:295-309 for the same
+// annotation-race precedent.
+func clearAnnotation(ctx context.Context, kubeClient client.Client, pod *corev1.Pod) error {
+	updated := pod.DeepCopy()
```
DerekFrank, line 34: check if we need this

pkg/controllers/pod/deletioncost/metrics.go
```diff
@@ -0,0 +1,94 @@
...
+	)
+	podsUpdatedTotal = opmetrics.NewPrometheusCounter(
+		crmetrics.Registry,
+		prometheus.CounterOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: podDeletionCostSubsystem,
+			Name:      "pods_updated_total",
+			Help:      "Number of pod deletion cost annotations updated in total. Labeled by result (updated, skipped_unchanged, error). The error label counts per-pod patch failures.",
+		},
+		[]opmetrics.Label{{Name: resultLabel, Help: "Outcome of the annotation write (updated, skipped_unchanged, error)."}},
```
DerekFrank, line 57: we can add helptext to the values

pkg/controllers/pod/deletioncost/metrics.go
```diff
@@ -0,0 +1,94 @@
...
+			Name:      "ranking_duration_seconds",
+			Help:      "Duration of node ranking computation in seconds.",
+			Buckets:   metrics.DurationBuckets(),
+		},
+		[]opmetrics.Label{},
+	)
+	// Per-pod queue-reconcile duration. Previously per-cycle when
+	// UpdatePodDeletionCosts ran synchronously; after the queue swap this
+	// measures each Queue.Reconcile call (single-pod write, retry, or skip).
+	annotationDurationSeconds = opmetrics.NewPrometheusHistogram(
```
DerekFrank, line 73: do we need this if we're using a queue? shouldn't controller runtime have this workqueue metric?

pkg/controllers/pod/deletioncost/metrics.go
```diff
@@ -0,0 +1,94 @@
...
+		prometheus.HistogramOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: podDeletionCostSubsystem,
+			Name:      "annotation_duration_seconds",
+			Help:      "Duration of a single pod annotation update operation in seconds.",
+			Buckets:   metrics.DurationBuckets(),
+		},
+		[]opmetrics.Label{},
+	)
+	reconcileSkippedTotal = opmetrics.NewPrometheusCounter(
```
DerekFrank, line 84: ditto

pkg/controllers/pod/deletioncost/metrics.go
```diff
@@ -0,0 +1,94 @@
...
+	// monotonic total.
+	nodesRanked = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: podDeletionCostSubsystem,
+			Name:      "nodes_ranked",
+			Help:      "Number of nodes ranked in the most recent reconcile cycle by the pod deletion cost controller.",
+		},
+		[]opmetrics.Label{},
```
DerekFrank, line 47: Label this with nodepool?

pkg/controllers/pod/deletioncost/queue.go
```diff
@@ -0,0 +1,204 @@
...
+		err = patchAnnotation(ctx, q.kubeClient, pod, strconv.Itoa(item.rank))
+	}
+	if err == nil {
+		podsUpdatedTotal.Inc(map[string]string{resultLabel: "updated"})
+		q.complete(qk)
+		return reconcile.Result{}, nil
+	}
+	// NotFound: pod is already gone. Conflict: another writer raced us and
+	// won; next reconcile will re-observe. Both are Skipped, not errors.
+	if apierrors.IsNotFound(err) || apierrors.IsConflict(err) {
```
DerekFrank, line 188: maybe split into two different labels. not_found, conflict

pkg/controllers/pod/deletioncost/annotation.go
nathangeology: Checked and yes, mergeFrom requires this.

pkg/utils/disruption/disruption.go
nathangeology: Currently logging the type error. Debated doing panic here, but that seems to be overkill maybe?

pkg/controllers/pod/deletioncost/ranking.go
nathangeology: On the helper itself: keeping isDrifted local rather than calling drift.ShouldDisrupt directly, because the disruption predicate takes a *Candidate and PDC operates on state.StateNode in the pre-Candidate partition step. The helper name is used such that PDC's drifted group must match what the drift controller will actually act on. The doc-comment now names disruption/drift.go:58 as the mirror site so future readers see the tie.

There's a lot of de-dup opportunities for follow up PRs. We can maybe cut an issue for this one and a few others and they should be smaller, easier to review PRs than rolling them into this one. In this particular one, the drift-condition read plus static gate is duplicated across three sites today (disruption/drift.go:58, disruption/staticdrift.go:49, and PDC's isDrifted). Consolidating requires either an adapter that builds a *Candidate from state.StateNode or a shared predicate package. Seems out of scope for this PR.

PR #2898, opened 2026-03-09
docs: Support Capacity Buffers in Karpenter

### LINKED ISSUE #749: Mega Issue: Manual node provisioning

sumukha-radhakrishna:
Fixes #749 #2571 

**Description**

Support sig-autoscaling CapacityBuffer API 
- https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/buffers.md


designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+Karpenter provisions nodes just-in-time based on pending pod demand. This creates scheduling latency as pods wait for node provisioning. Users need two capabilities that don't exist today:
+
+1. **Additional Capacity (Active Strategy)**: Pre-provision capacity so pods schedule immediately, maintaining buffer dynamically as workloads scale
+2. **Capacity Request (Static Strategy)**: Request capacity once and wait for confirmation before admitting workloads
+
+**Existing Workarounds:**
+
+Users maintain spare capacity through balloon pods, pause containers, or Static NodePools. These approaches have limitations:
+
+- **Balloon pods**: Hard to maintain, create scheduler overhead through preemption
```
jmdeal, line 11: Are these hard to maintain? The main issue I hear about balloon pods is the preemption overhead, which is non-negligible in large clusters.

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+- **Static NodePools**: Operate at node-level, require manual capacity planning, no workload awareness
+- **No standard API**: Not compatible with other autoscalers
```
jmdeal: By their nature, balloon pods should always be compatible with any autoscaler.

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+- [#3384](https://github.com/aws/karpenter-provider-aws/issues/3384)
+- [#4409](https://github.com/aws/karpenter-provider-aws/issues/4409)
+
+# Goals
+
+- Support the Kubernetes SIG Autoscaling CapacityBuffer API (`autoscaling.x-k8s.io/v1alpha1`)
+- Provide pod-level capacity abstraction where users specify pod requirements and Buffer capacity
+- Implement two provisioning strategies:
+  - **Active strategy**: Continuously maintain spare capacity that scales with workload changes
+  - **Static strategy**: One-time capacity request with completion semantics for external systems
```
jmdeal: Are these strategies specific to Karpenter or are they shared across all autoscalers which implement this API? If they are Karpenter specific, do we want to drive standardization?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+- Maintain buffer capacity through virtual pods that participate in scheduling simulation every provisioning cycle
+- Preserve buffer capacity during disruption operations (consolidation, drift, expiration)
+
+# Non-Goals
+
+- Optimizing scheduling simulation performance for Virtual pods
```
jmdeal: This is pretty broad, are there any specific optimizations that we want to implement that we're choosing not to implement for the initial integration?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+- Guaranteed capacity reservation: Buffers do not provide hard guarantees that capacity is reserved exclusively for specific workloads. Any pod that can schedule on buffer capacity may use it.
```
jmdeal: I'm not sure I'd even consider this a non-goal, it's something that should already be supported through existing taint and label semantics. When we release this feature, we should make sure we include documentation demonstrating how to achieve this.

designs/capacity-buffers.md
DerekFrank: More specifically, requires custom schedulers to have to construct bespoke Karpenter support

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+2. Resolves pod template from `scalableRef` or `podTemplateRef`
+3. Calculates replica count from `replicas`, `percentage`, or `limits`
+4. Writes to buffer status: `replicas` + `podTemplateRef`
+5. Updates status conditions based on strategy
+
+**Make changes to existing Provisioner:**
+1. Reads buffer status
+2. Constructs virtual pods in-memory (not actual pod objects)
+3. Runs scheduling simulation: Can these virtual pods fit on existing nodes?
+   - Yes → Virtual pods can be placed on existing capacity, set `Provisioning: True`
```
DerekFrank, line 62: Do we have these semantics flipped?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+![](./images/static_buffer.png)
+
+
+**Behavior:**
+- Maintains fixed capacity as one-time request
+- Ignores workload scaling and template changes after initial provisioning
+- Buffer size remains constant and no more changes are allowed to the spec
+- After capacity is provisioned, status moves to `Completed: True`
+- Once completed, buffer is no longer acted upon by provisioner
+- Capacity is preserved during disruption for a configurable grace period (default 1 hour)
```
DerekFrank: Configurable where?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+
+# Proposal
+
+Support the Kubernetes SIG Autoscaling CapacityBuffer API (`autoscaling.x-k8s.io/v1alpha1`). This provides:
+
+- Standard API compatible with Cluster Autoscaler
+- Pod-level abstraction (specify pod requirements, not node requirements)
+- Two provisioning strategies: active (spare capacity) and static (capacity request)
+- Integration with external admission controllers
```
jmdeal: I think it's a stretch to say this, unless any integrations already exist today that I'm unaware of. Long-term that's the goal, but I imagine it's going to require additional investment on our end outside of the initial implementation of this API. That might just be guidance, but I don't want to sell the narrative that as soon as we ship support for CapacityBuffers, Karpenter will work with these admission controllers out of the box.

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+**Lifecycle:**
+1. User creates buffer with active strategy
+2. Buffer controller resolves template and calculates replicas
+3. Every provisioning cycle, provisioner attempts to provision capacity
+4. Status updated to `Provisioning: True` only when virtual pods can be placed on existing nodes (without new NodeClaims)
+5. Buffer continuously maintains capacity, reacting to workload changes
+6. User deletes buffer when no longer needed
+
+### Static Capacity (`buffer.x-k8s.io/static-capacity`)
```
DerekFrank: I don't love that we've overloaded this term. Maybe ephemeral-request? or just request?

designs/capacity-buffers.md
jmdeal: We should include a section on graduation criteria for this feature, I think an integration with a third-party admission controller would be a solid milestone for beta.

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+ReadyForProvisioning (Applies to: Active & Static)
+- True: Pod template is successfully resolved and target replicas are calculated.
+- False: Missing references (ScalableRefNotFound, PodTemplateNotFound), validation errors, or calculation failures.
+
+Provisioning (Applies to: Active & Static)
+- True: Capacity is actually available. Virtual pods fit onto existing nodes without requiring new NodeClaims.
+- False: Virtual pods don't fit; new NodeClaims are required, limits prevent scaling (InsufficientCapacity), or provisioning failed.
+
+Completed (Applies to: Static only)
```
DerekFrank: Maybe fulfilled instead?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+- After grace period expires, buffer is no longer considered and capacity can be consolidated
+- Grace period prevents consolidation from immediately removing capacity before workloads use it
+	- The grace period is tracked from the timestamp when `Completed: True` condition is set in the buffer status. This ensures:
+		- Time starts only after capacity is actually available
+		- Buffers that take longer to provision don't have reduced protection time
+
+
+**Future: expireAfter Field**
+
+We propose adding an `expireAfter` field to the upstream CapacityBuffer API (sig-autoscaling) that would apply to both active and static strategies:
```
DerekFrank: We should probably actually write this proposal down :D

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+1. Watches CapacityBuffer CRDs filtered by provisioning strategy
+2. Resolves pod template from `scalableRef` or `podTemplateRef`
+3. Calculates replica count from `replicas`, `percentage`, or `limits`
+4. Writes to buffer status: `replicas` + `podTemplateRef`
+5. Updates status conditions based on strategy
+
+**Make changes to existing Provisioner:**
+1. Reads buffer status
+2. Constructs virtual pods in-memory (not actual pod objects)
+3. Runs scheduling simulation: Can these virtual pods fit on existing nodes?
```
jmdeal, line 61: Is the interaction with existing nodes different depending on the active or static strategy?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+7. Capacity preserved during disruption for grace period (default 1 hour, configurable)
+8. After grace period expires, capacity can be consolidated automatically
+9. User can delete buffer manually at any time
+
+## Supported References
+
+**scalableRef (Active Strategy Only):**
+- `apps/v1/Deployment`
+- `apps/v1/StatefulSet`
+- `apps/v1/ReplicaSet`
```
sumukha-radhakrishna, line 101: Need to check if we can directly support this. Right now we will support homogeneous types

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+**Q: Can users switch strategies on existing buffers?**
+A: No. Static strategy buffers are immutable (enforced via CEL validation) to prevent confusion and ensure predictable behavior. Once a buffer is created with a specific strategy, the strategy cannot be changed. Users must delete and recreate the buffer if they need a different strategy. Active strategy buffers allow spec updates (e.g., changing replicas or percentage) but the strategy itself cannot be changed.
+
+**Q: What happens if buffer can't be satisfied due to NodePool limits?**
+A: Buffer status reflects actual provisioned replicas may be less than requested. It is retried until replicas are satisfied. For static case, it will be retried until replicas are satisfied once.
```
DerekFrank: Do we want a more descriptive status than this?

Provisionable: False?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+- After capacity is provisioned, status moves to `Completed: True`
+- Once completed, buffer is no longer acted upon by provisioner
+- Capacity is preserved during disruption for a configurable grace period (default 1 hour)
+- After grace period, capacity can be automatically consolidated
+- Customer can delete buffer manually at any time
+
+**Restrictions:**
+- MUST use `podTemplateRef` (NOT `scalableRef`)
+- MUST use `replicas` or `limits` (NOT `percentage`)
+- Spec is immutable after creation (enforced via CEL validation)
```
DerekFrank: We're going to be writing this I assume? How do we guarantee it if the user installs the CRD right from sig-autoscaling, or also has CAS installed?

designs/capacity-buffers.md
sumukha-radhakrishna: No. We want to set it `Provisioning:True` when we can place virtual pods on the cluster and not when we provision capacity for these pods.

designs/capacity-buffers.md
sumukha-radhakrishna: `ephemeral-request` sounds good to me

designs/capacity-buffers.md
sumukha-radhakrishna: We want to introduce `expireAfter` in the upstream CRD

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+
+**Behavior:**
+- Maintains fixed capacity as one-time request
+- Ignores workload scaling and template changes after initial provisioning
+- Buffer size remains constant and no more changes are allowed to the spec
+- After capacity is provisioned, status moves to `Completed: True`
+- Once completed, buffer is no longer acted upon by provisioner
+- Capacity is preserved during disruption for a configurable grace period (default 1 hour)
+- After grace period, capacity can be automatically consolidated
```
jmdeal: Should we be consolidating capacity provisioned for a static buffer at all? I could see us performing consolidation if a cheaper set of instance types would enable the same pods to schedule, but I wouldn't expect us to scale all the way down until it's deleted. Even for that first case, I'm not sure we need additional disruption controls - is there a gap in the existing controls that we would need to close.

jmdeal: > I wouldn't expect us to scale all the way down until it's deleted

Based on further reading, I think this is the expected behavior. I don't love this idea because I think it violates expectations around declarative state in Kubernetes. I'm guessing the idea is that we don't want these third-party admission controllers to need to perform explicit cleanup, but I'm not sure that's a route we want to go down.

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+## CapacityBuffer Status Conditions
+The CapacityBuffer uses standard Kubernetes conditions to report its state:
+
+ReadyForProvisioning (Applies to: Active & Static)
+- True: Pod template is successfully resolved and target replicas are calculated.
+- False: Missing references (ScalableRefNotFound, PodTemplateNotFound), validation errors, or calculation failures.
+
+Provisioning (Applies to: Active & Static)
+- True: Capacity is actually available. Virtual pods fit onto existing nodes without requiring new NodeClaims.
```
jmdeal: Provisioning seems like an odd name for this given that when it's `True` we're not actually actively provisioning nodes. Should this just be the ready condition?

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+### Karpenter's Single-Loop Architecture
+
+Karpenter uses a single provisioning loop for all scheduling decisions. This single-loop design has important implications for buffer pods:
+- All pods (real + buffer) are scheduled together in one coherent decision, ensuring optimal resource utilization.
+- Cluster state remains consistent during scheduling, preventing race conditions in capacity tracking.
+- The singleton pattern helps with buffer pod tracking since there's no risk of concurrent provisioning loops interfering with each other.
+
+### Performance Trade-offs
+
+A large number of virtual pods might increase the latency of scheduling simulation. The time to simulate and schedule grows with the number of pods being considered. We will look into improving this after running benchmarking to understand real-world performance characteristics.
```
jmdeal: Do we expect there to be any difference between the affect of virtual pods and real pods? There's going to be some additional latency when generating the virtual pods, but I would expect that to be a pretty cheap operation.

designs/capacity-buffers.md
sumukha-radhakrishna: Yeah we need some kind of ownership logic for the CR/CRD

designs/capacity-buffers.md
```diff
@@ -0,0 +1,549 @@
...
+
+**Current Approach: Stateless Virtual Pod Construction**
+
+We reconstruct virtual pods from buffer status every provisioning cycle rather than maintaining persistent state. This approach prioritizes correctness and simplicity over performance:
+
+**Pros:**
```
jmdeal: What are we comparing against?

designs/capacity-buffers.md
sumukha-radhakrishna: It is not. Let me clarify

designs/capacity-buffers.md
```diff
@@ -0,0 +1,514 @@
...
+- Optimizing scheduling simulation performance for Virtual pods (will measure and optimize if needed)
+- Initial implementation will only support same-namespace references for `scalableRef` and `podTemplateRef`
+- Ephemeral capacity strategy: One-time capacity request pattern for batch systems like Kueue (deferred to future work)
+- Adding `expireAfter` field to the API in initial implementation (requires upstream sig-autoscaling consensus)
+
+# Future Work
+
+**Ephemeral Capacity Strategy:**
+
+After the initial active buffer implementation, we will implement ephemeral capacity strategy to support batch systems like Kueue. Ephemeral strategy will provide:
```
DerekFrank: Nit: To harden support for batch systems. Kueue can work with active buffers, its just a bit racey

designs/capacity-buffers.md
sumukha-radhakrishna: I can add the graduation criteria and add admission controller as the milestone.

designs/capacity-buffers.md
sumukha-radhakrishna: Removed the static case altogether

designs/capacity-buffers.md
```diff
@@ -0,0 +1,514 @@
...
+- Preserve buffer capacity during disruption operations (consolidation, drift, expiration)
+
+# Non-Goals
+
+- Optimizing scheduling simulation performance for Virtual pods (will measure and optimize if needed)
+- Initial implementation will only support same-namespace references for `scalableRef` and `podTemplateRef`
+- Ephemeral capacity strategy: One-time capacity request pattern for batch systems like Kueue (deferred to future work)
+- Adding `expireAfter` field to the API in initial implementation (requires upstream sig-autoscaling consensus)
+
+# Future Work
```
DerekFrank: Nit: this should probably all be towards the end, we want to talk about what we are working on first instead of what we will work on

designs/capacity-buffers.md
DerekFrank: I'm convinced this is not ideal from a readability perspective. We should go discuss with the CAS folks about changing these semantics to be more readable. For now lets mimic them, but call out fixing the semantics as graduation criteria

designs/capacity-buffers.md
sumukha-radhakrishna: Will keep the same semantics as cluster-autoscaler for now, will change once I get the alignment from CA

designs/capacity-buffers.md
```diff
@@ -0,0 +1,514 @@
...
+
+**Make changes to existing Provisioner:**
+1. Reads buffer status
+2. Constructs virtual pods in-memory (not actual pod objects)
+3. Runs scheduling simulation: Can these virtual pods fit on existing nodes?
+   - Yes → Virtual pods can be placed on existing capacity, set `Provisioning: True`
+   - No → Create NodeClaims, keep `Provisioning: False` until nodes are available
+4. Only sets buffer status to `Provisioning: True` when virtual pods can be successfully placed on existing cluster capacity without creating new NodeClaims
+
+**Key Point:** Virtual pods are reconstructed every provisioning loop from buffer status. No pod objects are created. The `Provisioning: True` status reflects actual available capacity in the cluster, ensuring the status accurately represents whether buffer capacity is ready for use even if NodeClaims fail to provision.
```
DerekFrank, line 66: I assume we'll cache the recreated pods?

sumukha-radhakrishna: We will add cache before GA

DerekFrank (approved):
/lgtm

PR #2400, opened 2025-07-24
chore: ignore pods that use unsupported provisioner in the storageClass

jigisha620:
Fixes #N/A 

**Description**
This change allows Karpenter to ignore pods that use unsupported provisioner within their storageClass. Cloudproviders can use this to ignore pods with certain provisioners.

**How was this change tested?**
Tested on my dev cluster with aws-provider by tweaking it to ignore a fakeProvisioner and ensuring that we don't create a node for it. Also added unit test for it.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2903, opened 2026-03-10
chore: ignore pods that use an unsupported topology key in the storageClass

AndrewMitchell25:
Fixes #N/A 

**Description**
This change allows Karpenter to ignore trying to schedule pods that use a StorageClass with an unsupported topology key, which can be set by cloud providers to ignore pods with certain topology keys. This is similar to #2400, which introduced similar functionality for provisioners.

`make presubmit`, added a unit test, manually tested

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/karpenter
 
-go 1.25.7
+go 1.26.1
 
 require (
 	github.com/Pallinder/go-randomdata v1.2.0


pkg/controllers/provisioning/scheduling/volumetopology.go
```diff
@@ -40,6 +40,11 @@ import (
 // that uses a StorageClass with any of these unsupported provisioners, Karpenter will skip scheduling that pod.
 var UnsupportedProvisioners = sets.New[string]()
 
+// UnsupportedTopologyKeys is a set of topology keys that are not supported. When a StorageClass has AllowedTopologies
+// containing any of these keys, Karpenter will skip scheduling pods that reference that StorageClass, since nodes
```
jigisha620:
```suggestion
// containing any of these keys, Karpenter will skip scheduling pods that reference a PVC with that StorageClass, since nodes
```

PR #2906, opened 2026-03-11
docs: RFC to introduce node replacement strategies during drift, starting with optionally not requiring replacements

### LINKED ISSUE #2905: Support disruption policies that allow terminate-before-create for fixed capacity use cases
### Description

_Moving over from https://github.com/aws/karpenter-provider-aws/issues/8935 based on feedback._ 

### Description

*Summary: We'd like to have an option on our NodePools to terminate instances before creating new ones as a drift resolution mechanism while respecting PDBs and node disruption budgets as normal.*  

Scenario: We leverage a set of GPU instances which are quite expensive to acquire and so get them as part of Capacity reservations (ODCRs) associated with Karpenter NodePools. We also maintain little to no buffer on these instances due to cost. As a result, when we try to do drift resolutions via Karpenter, Karpenter is never able to automate the resolutions because it can't scale up a new instance (capacity is just not available) before shutting down the old one. This means a human or other system now has to come in and handle node replacements for just this type of instance. 

Currently, for our rarely available GPU instances, it requires humans to be in the loop to manage AMI updates which is a significant operational pain point. For our other capacity, Karpenter significantly simplifies upgrades thanks to its excellent drift detection and resolution capabilities

vaietc:
Fixes #2905 

**Description**

This change adds a small design for supporting different node replacement strategies during drift resolution, starting with not requiring replacements at all. The design attempts to extend the API such that other replacement strategies can be introduced over time. 

Docs update only so far, no code yet.


vaietc:
@DerekFrank as discussed at the working group meeting, I moved the issue over from the AWS provider to the main Karpenter project. I've attached an early RFC for review and would appreciate any early feedback :)

DerekFrank:
Summarizing notes from working group:

* How do we handle budgets for static and dynamic capacity?
* Are there significant blockers to support dynamic capacity out of the gate?
* Do we want to be able to set different replacement policies for different types of disruption? (ex: Drift vs Consolidation), and how does that API look?
* Naming of the policy? Is it more accurately TerminateThenReplace or does DoNotCreateReplacement more accurately capture the setting?

vaietc:
@DerekFrank Updated the proposal to help answer the questions raised, summarizing here: 

> How do we handle budgets for static and dynamic capacity?

I've proposed a small tweak to the current disruption budget calculation to include Terminating and Initializing nodes in the list of total nodes in the NodePool AND in the 'disrupting' Nodes section. I believe this change more accurately reflects the spirit of the disruption budget to track all disrupted nodes, not necessarily just ones disrupted by Karpenter. 

> Are there significant blockers to support dynamic capacity out of the gate?

Did a quick read of the code and I think the answer is no. Since the disruption budget calculation is the same for both static and dynamic capacity, that change can be made in one common place. 

> Do we want to be able to set different replacement policies for different types of disruption? (ex: Drift vs Consolidation), and how does that API look?

I think its a bit tricky to reason about for Consolidation since sometimes the decision to consolidate is based on one of the following:

1. Node is empty (no replacement needed)
2. Node's pods can be housed in existing nodes (no replacement needed)
3. A different, lower cost node can house pods on that node (or more than one node in case of multi-node consolidation). In this case, its probably always wise to spin up the replacement node first to avoid longer term disruption for pods. 

As a result, I propose we don't touch consolidation for now but we can control it later if we want with changes to the API. For now, I'd like to have `spec.disruption.driftResolutionPolicy `as described in the pull request and have it only apply to drift. 

> Naming of the policy? Is it more accurately TerminateThenReplace or does DoNotCreateReplacement more accurately capture the setting?

I renamed to 'Terminate' since that is the only action the disruption controller takes and accurately captures the setting

vaietc:
@DerekFrank started a draft pull request based on the RFC so far here: https://github.com/kubernetes-sigs/karpenter/pull/2955 in case code is easier to read than my words :)

designs/node-replacement-drift-strategies.md
```diff
@@ -0,0 +1,61 @@
...
+3. As a Karpenter operator, I want design decisions on API changes to be extensible to future use cases of node replacement strategies.
+
+## Problem statement
+
+Karpenter is used to manage automated drift resolution  safely by cluster operators when rolling out new node software (ex. new OS upgrades, Kubernetes versions etc.). Karpenter provides multiple safety controls such as Node disruption budgets and do-not-disrupt annotations on Nodes and pods. In addition, Karpenter also spins up replacement nodes before starting eviction and termination of the drifted node. However, in some scenarios, users have very limited pools of capacity to choose from - two common cases are expensive instance types (ex. GPUs) and capacity reservations (ex. Amazon ODCRs) where it is not desirable to spin up additional high-cost / limited capacity before scaling down the drifted node. 
+
+## Proposed design
+
+The proposed design is a fairly simple extension over the current NodePool API and disruption controller.
```
DerekFrank, line 21: You probably don't need this sentence

designs/node-replacement-drift-strategies.md
```diff
@@ -0,0 +1,61 @@
+# RFC: Optionally allow disrupting nodes without creating replacements
+
+## Overview
+
+This design proposes an optional setting on NodePools that allows for applying disruption actions without needing to spin up a replacement Node. Currently, Karpenter's drift resolution workflow requires a replacement Node to be spun up and available before the drifted node is terminated, which is not viable in certain static capacity scenarios.
```
DerekFrank, line 5: The first sentence makes me think we're tackling consolidation here too, not just drift. The rest of the discussion is about drift explicitly

designs/node-replacement-drift-strategies.md
```diff
@@ -0,0 +1,61 @@
...
+
+Today, this setup works because virtually all disruption actions (aside from Empty consolidations), will spin up a Replacement node and blocks termination of the node until its up - this in turn prevents runaway scale down. However, with the introduction of the new proposed action to Terminate without creating a replacement, we run the risk of runaway deletions. To prevent this, we propose the following tweak to the algorithm 
+
+1. Include both terminating nodes and initializing nodes in the total count of Nodes in the NodePool (i.e N, now includes T and I as part of its totals)
+2. Include the terminating and initializing nodes in the NodePool in the count of 'disrupting' nodes. While these nodes are not actively 'disrupting', they should be considered as contributing to the disruption budget, in the same way 'NotReady' nodes are. 
+
+With this change, even with strategies where a replacement node is not spun up, the NodePool is aware of unhealthy nodes (either initializing or terminating) and will count them towards it overall disruption budget. This should work for all disruption strategies without significant changes to the current algorithm.
```
DerekFrank, line 57: Do we also want to block drift when that nodepool is failing to launch new nodes? i.e. NodePool Registration Unhealthy? That preserves more of the existing behavior IMO, and seems like a natural safety mechanism, but it could lock your cluster for static nodepools. Maybe it only applies to dynamic nodepools? If we do that, does the disruption budget calculation make sense to tweak still?

I'm not sure I have a good answer, but we should explore how it supplements the design

designs/node-replacement-drift-strategies.md
```diff
@@ -0,0 +1,61 @@
...
+
+** Release recommendation **
+
+We recommend releasing the updates to the disruption calculations behind a feature flag along with the API updates to avoid unexpected regressions
```
DerekFrank, line 61: We should flesh this out a bit more. Are the calculation changes the only thing thats behind a flag? Does that mean you can set Terminate without changing the disruption calculation? What are the graduation criteria for that flag to become default, or do we think it will always stay a setting? Can we just implicitly calculate the budget differently if you have Terminate set?

designs/node-replacement-drift-strategies.md
vaietc: Indeed we should.. I was hoping that fell under the 'Initializing' nodes case but I assume you're talking about being unable to fulfill open NodeClaims as well? Would it help to watch NodeClaims instead of Nodes in that case?

designs/node-replacement-drift-strategies.md
vaietc: Updated a little with my reocmmendations.. suggesting we put the new calculations also behind the same flag and once the flag graduates to beta, it becomes the default calculation for all scenarios.

designs/node-replacement-drift-strategies.md
vaietc: Ack, made it explicit that its about drift

designs/node-replacement-drift-strategies.md
vaietc: I haven't understood the comment on static capacity problems yet, hopefully we can discuss this more at the WG meeting tomorrow :)

designs/node-replacement-drift-strategies.md
```diff
@@ -0,0 +1,63 @@
...
+    # 2. Terminate: Karpenter terminates the node without spinning up a replacement. If evicted pods are unable to be placed, Karpenter's auto-scaling will instead be used to spin up a replacement node.
+    driftResolutionPolicy: CreateReplacement | Terminate # Default: CreateReplacement
+      
+```
+
+### Node disruption budget changes
+
+Today, the logic used to determine the number of disruptable nodes is the same for all disruption strategies (standard drift, static drift, consolidation etc.). It works as follows. For each NodePool, determine:
+
+1. Number of Nodes in the NodePool (N). *Exclude* nodes that are terminating or not fully initialized. For our design, assume that Terminating nodes for the NodePool are T and initializing (but not ready) nodes in the NodePool are I.
```
DerekFrank, line 46: We should be careful about a race condition here. We can count in-flight nodeclaims to know how many nodes we "should" have. However we need to prevent the drift controller from 'leaking' deletes while provisioning is failing.

If the provisioner fails to scale up, before it tries again, the drift controller might see there are no in-flight nodeclaims and scale down even further.

DerekFrank, line 46: One open question we should answer is how to count failed nodeclaims, i.e. nodeclaims that got ICE'd

DerekFrank:
I have been thinking about this, and I'm starting to wonder if this is a config the end user even wants to set. If you look at the usecases, it starts to feel like something that Karpenter should be able to derive ahead of time:

* Reserved Capacity: Karpenter knows that there isn't more capacity, it should terminate then replace
* Static Capacity: What else would a limit on static capacity even semantically mean?

If we get a little clever about how we calculate budgets, I think we can do this behavior without even surfacing a field. There is risk if users have not properly configured their PDBs, but that risk exists in all of Karpenter

DerekFrank:
@vaietc  do you want to close this in favor of https://github.com/kubernetes-sigs/karpenter/pull/3203

vaietc:
Closing in favour of https://github.com/kubernetes-sigs/karpenter/pull/3203

Issue #2803, opened 2026-01-19
Deployment with a zone TSC and ephemeral volumes block consolidation


IgorIgnatevBolt:
### Description

**Observed Behavior**:
We observe many ```Unconsolidatable``` events on nodes where deployments with ephemeral volumes and zone TSC are running. It's not a scheduling problem, only a consolidation problem.

The event contains information that zone TSC is calculated wrongly, even though the volume is ephemeral, and we shouldn't care about this. All pods are evenly distributed across 3 zones. However, the counts are zeroed for the other zones, different by pod.
```
....Unconsolidatable...
Not all pods would schedule, sre/deploy-test-pvc-6cfb4f4dd-x7dtx
....
unsatisfiable topology constraint for topology spread, key=topology.kubernetes.io/zone (counts = eu-central-1a: 3, eu-central-1b: 0, eu-central-1c: 0, podDomains = topology.kubernetes.io/zone In [eu-central-1a], nodeDomains = topology.kubernetes.io/zone In [eu-central-1a];
```
**Expected Behavior**:
TSC is calculated correctly, or ephemeral volumes are ignored for deployment, and the consolidation decision allows this pod to be rescheduled.

**Reproduction Steps** (Please include YAML):
Nodepool
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: test-arm64
spec:
  disruption:
    budgets:
    - nodes: 10%
      reasons:
      - Empty
      - Underutilized
    - nodes: "0"
      reasons:
      - Drifted
    consolidateAfter: 0s
    consolidationPolicy: WhenEmptyOrUnderutilized
  limits:
    cpu: "1500"
  template:
    metadata:
      labels:
        company/Workload: TEST
    spec:
      expireAfter: 120h
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: test
      requirements:
      - key: kubernetes.io/arch
        operator: In
        values:
        - arm64
      - key: karpenter.k8s.aws/instance-family
        operator: In
        values:
        - m7g
        - c7g
        - r7g
      - key: karpenter.k8s.aws/instance-size
        operator: In
        values:
        - large
        - xlarge
        - 2xlarge
        - 4xlarge
        - 8xlarge
      - key: topology.kubernetes.io/zone
        operator: In
        values:
        - eu-central-1a
        - eu-central-1b
        - eu-central-1c
      - key: kubernetes.io/os
        operator: In
[...]

PR #2907, opened 2026-03-11
fix: skip ephemeral volume topology requirements for bound pods during consolidation

moko-poi:
## Summary

Fix for #2803: Ephemeral volumes with zone TopologySpreadConstraints block consolidation with Unconsolidatable events.

Skip ephemeral volume topology requirements when evaluating bound pods for consolidation, since ephemeral PVCs are deleted with the pod and a new PVC will be created on rescheduling. This primarily targets WaitForFirstConsumer StorageClasses, which is the common case for ephemeral volumes. See inline comment for discussion on
Immediate binding mode.

## Test plan

- Unit tests: bound pod + ephemeral volume returns no topology requirements
- Unit tests: unbound pod + ephemeral volume returns topology requirements (unchanged behavior)
- Unit tests: bound pod + regular PVC returns topology requirements (unchanged behavior)
- Integration test: consolidation replaces node with ephemeral volumes and zone TSC
- Full scheduling test suite passes (339 tests)
- Full disruption test suite passes (233 tests)

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -31,6 +31,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	policyv1 "k8s.io/api/policy/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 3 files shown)


moko-poi:
@jmdeal Thanks for the analysis on this issue! I wanted to get your thoughts on one thing.

In your comment on #2803, you noted

> One nuance that I'll note is that if the VolumeBindingMode is Immediate, we would still want to honor the ephemeral volume constraints.

The current implementation skips topology requirements for all ephemeral volumes on bound pods, regardless of VolumeBindingMode:

```go
if volume.Ephemeral != nil && pod.Spec.NodeName != "" {
    return nil, nil
}
```

For WaitForFirstConsumer, this is correct — the new PVC will bind to whatever zone the pod is scheduled to, so the old PVC's zone constraint is irrelevant.

For Immediate, the situation is more nuanced. If we skip the constraint, Karpenter may create a replacement node in a zone that differs from where the provisioner binds the new PVC, which could leave the pod temporarily unschedulable. However, if we don't skip, the old PVC's zone constraint will continue to cause incorrect TSC
counts, blocking consolidation entirely (which is the current bug). Since the old PVC is deleted with the pod and the new PVC's zone is determined independently by the provisioner, honoring the old constraint doesn't guarantee correctness either.

Given that Immediate + ephemeral volumes is an uncommon combination and skipping is still an improvement over the current behavior (permanently blocked consolidation), I went with the simpler approach for now. That said, if you think Immediate mode warrants additional handling, I'd like to discuss what the right approach would
be. Let me know your thoughts.

PR #2910, opened 2026-03-12
POC: speedup single node consolidation by grouping nodes and pools to a single node-type

grosser:
**Description**
this took us from ~5m -> 30s because we have lots of nodes and node-pools
the label for grouping should be made user-configurable, see https://github.com/kubernetes-sigs/karpenter/pull/2871 and https://github.com/kubernetes-sigs/karpenter/issues/2814

we sadly had to also add a delay after each consolidation since otherwise our node-churn almost doubled
that would be another useful feature to have as a config option
see https://github.com/kubernetes-sigs/karpenter/pull/2961

### DIFF
--- a/pkg/controllers/disruption/helpers.go
+++ b/pkg/controllers/disruption/helpers.go
@@ -94,13 +94,16 @@ func SimulateScheduling(ctx context.Context, kubeClient client.Client, cluster *
 	}
 	pods = append(pods, deletingNodePods...)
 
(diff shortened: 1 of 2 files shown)


talayalon:
what's the status of this POC? we are hitting the same issues

grosser:
need some maintainer to agree this is a worthwhile feature / mergeable
I can do the work, but tired of making PRs that get auto-closed by the bot

PR #2913, opened 2026-03-17
fix(disruption): correctly accumulate pod-count in LogValues

J926L:
The lo.Reduce function in Command.LogValues was incorrectly discarding the accumulator, leading to inaccurate pod counts in logs. This PR fixes the reduction logic and adds a comprehensive test suite to verify correct accumulation across multiple candidates.


jmdeal:
Closing in favor of #2912

PR #2923, opened 2026-03-19
feat: add support for cloud provider node registration hooks

joshuakguo:
Fixes #N/A 

**Description**

- added `NodeLifecycleHook` for cloud providers to provide hooks for blocking node lifecycle steps
- added `WithRegistrationHook` for node registration hooks for cloud providers

- manual testing
- unit tests

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/registration.go
+++ b/pkg/controllers/nodeclaim/lifecycle/registration.go
@@ -88,7 +88,6 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
 	// If any hook is not ready, registration is deferred and the unregistered taint remains.
 	hooksResult, hookErrors := r.checkRegistrationHooks(ctx, nodeClaim)
 	if lo.IsEmpty(hooksResult) && hookErrors == nil {
-		// Remove karpenter.sh/unregistered taint
 		node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t corev1.Taint, _ int) bool {
 			return t.MatchTaint(&v1.UnregisteredNoExecuteTaint)
 		})
@@ -143,29 +142,25 @@ func (r *Registration) checkRegistrationHooks(ctx context.Context, nodeClaim *v1
 		}
 	})
 
-	// Collect pending hook names and compute the shortest requeue interval
+	// Collect pending and errored hook names and compute the shortest requeue interval
 	var pendingHooks []string
 	mergedResult := reconcile.Result{}
 	for i, result := range results {
-		if errs[i] != nil {
-			continue
-		}
-		if !lo.IsEmpty(result) {
+		if errs[i] != nil || !lo.IsEmpty(result) {
 			pendingHooks = append(pendingHooks, r.registrationHooks[i].Name())
-			if mergedResult.RequeueAfter == 0 || (result.RequeueAfter > 0 && result.RequeueAfter < mergedResult.RequeueAfter) {
-				mergedResult.RequeueAfter = result.RequeueAfter
+			if errs[i] == nil {
+				if mergedResult.RequeueAfter == 0 || (result.RequeueAfter > 0 && result.RequeueAfter < mergedResult.RequeueAfter) {
+					mergedResult.RequeueAfter = result.RequeueAfter
+				}
+				mergedResult.Requeue = mergedResult.Requeue || result.Requeue //nolint:staticcheck
 			}
-			mergedResult.Requeue = mergedResult.Requeue || result.Requeue //nolint:staticcheck
 		}
 	}
 
-	if err := multierr.Combine(errs...); err != nil {
-		return mergedResult, err
-	}
 	if len(pendingHooks) > 0 {
-		log.FromContext(ctx).V(1).Info("registration hooks not satisfied", "hooks", pendingHooks)
+		log.FromContext(ctx).V(1).Info("awaiting registration hooks", "hooks", pendingHooks)
 		nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeRegistered, "RegistrationHookPending", strings.Join(pendingHooks, ", "))
-		return mergedResult, nil
+		return mergedResult, multierr.Combine(errs...)
 	}
 	return reconcile.Result{}, nil
 }


pkg/cloudprovider/types.go
```diff
@@ -99,6 +99,17 @@ type CloudProvider interface {
 	GetSupportedNodeClasses() []status.Object
 }
 
+// RegistrationHook is implemented by cloud providers to gate the removal of the
+// karpenter.sh/unregistered taint. All registered hooks must return true before
+// the taint is removed and node registration completes.
+type RegistrationHook interface {
```
ryan-mist: I wonder if we should extend this to Initialization as well?

jmdeal: I think adding hooks for initialization is a natural extension. That being said, we don't have an immediate need - I don't have a strong preference on adding them now vs adding them in the future.

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -97,6 +108,24 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
 	return reconcile.Result{}, nil
 }
 
+// checkRegistrationHooks evaluates all registration hooks in order. If a hook returns an error,
+// it is returned. If a hook returns false (not ready), blocked is true and a requeue result is
+// returned with the status condition updated. Returns blocked=false if all hooks pass.
+func (r *Registration) checkRegistrationHooks(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, bool, error) {
```
jmdeal: I'd consider just returning an empty result for success rather than adding an extra bool. This is consistent with how we've designed other sub-reconcilers (e.g. [termination controller](https://github.com/kubernetes-sigs/karpenter/blob/2e04cf25f97f878db06f025c200be9158c5cf0e5/pkg/controllers/node/termination/controller.go#L150-L153))

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -97,6 +108,24 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+	for _, hook := range r.registrationHooks {
```
jmdeal: I'd invoke all subreconcilers per execution, probably in parallel since they may perform some synchronous work. Then our logging / eventing can just list the subreconcilers we're still waiting on. 

I'd also consider removing the "reason" messaging from the lifecycle controller. When introducing hooks, we should make the integration point as minimal as possible while giving individual hook implementations the flexibility to implement additional telementry. Since we're passing the hook the context, if they need to provide additional logging the can. I think in most cases though the name of the hook is sufficient in non-error cases.

pkg/cloudprovider/types.go
```diff
@@ -99,6 +99,17 @@ type CloudProvider interface {
...
+	// Name for the hook.
+	Name() string
+	// RegistrationCheck returns true if this hook's preconditions are satisfied and
+	// registration can proceed. If false, reason describes what's pending.
+	RegistrationCheck(ctx context.Context, nodeClaim *v1.NodeClaim) (ok bool, reason string, err error)
```
jmdeal: nit: when you name a method a noun, it suggests it returns that thing (e.g. `Name` returns the name). Also, parameter names are not typically included in interface definitions. I'd also consider using `reconcile.Result` directly or a similar struct.

```suggestion
	Registered(context.Context, *v1.NodeClaim) (NodeLifecycleHookResult, error)
```

where `NodeLifecycleHookResult` is defined something like this:
```go
type NodeLifecycleHookResult struct {
    // Indicates the lifecycle controller should requeue with exponential backoff
    Requeue bool
    // Indicates the lifecycle controller should requeue after the provided time. If multiple hooks are provided, the shortest interval is respected.
    RequeueAfter time.Time
}
```

I'm suggesting to implement a new struct rather than reuse `reconcile.Result` directly since the hook itself is not a reconciler and I think this improves clarity, but I don't have a strong opinion on that.

pkg/controllers/controllers.go
```diff
@@ -64,6 +65,18 @@ import (
 	"sigs.k8s.io/karpenter/pkg/state/nodepoolhealth"
 )
 
+type ControllerOptions struct {
+	registrationHooks []cloudprovider.RegistrationHook
+}
+
+// WithRegistrationHook registers a hook from the cloud provider that gates the removal of the
+// karpenter.sh/unregistered taint during node registration.
```
jmdeal: Rather than specifying that it blocks the removal of the taint, I'd generalize and say it blocks Karpenter from marking the node as registered. I'd also provide an example of when a cloudprovider should implement a registration hook. The canonical example would be if a well known label needs to be applied asynchronously.

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -75,6 +81,11 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
 		r.recorder.Publish(UnregisteredTaintMissingEvent(nodeClaim))
 	}
 	ctx = log.IntoContext(ctx, log.FromContext(ctx).WithValues("Node", klog.KObj(node)))
+	// Check all registration hooks before proceeding with node sync.
+	// If any hook returns false, registration is deferred and the unregistered taint remains.
+	if result, blocked, err := r.checkRegistrationHooks(ctx, nodeClaim); blocked || err != nil {
+		return result, err
+	}
 	if err = r.syncNode(ctx, nodeClaim, node); err != nil {
```
jmdeal: I think we should still perform most of the work in `syncNode` before invoking the registration hooks. It's not necessary for correctness, but the presence of the labels on the node helps with observability / debugging. The only things we want to do after are the addition of the `karpenter.sh/registered` label and the removal of the `unregistered` taint.

pkg/cloudprovider/types.go
joshuakguo: ill leave it to be extensible for initialization hooks in the future but i won't add them in this pr

pkg/controllers/nodeclaim/lifecycle/registration.go
joshuakguo: Split out the logic for the patch to still do most of the work in `syncNode` but still keep just one patch call

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -75,11 +81,32 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
-		if errors.IsConflict(err) {
-			return reconcile.Result{Requeue: true}, nil
+	stored := node.DeepCopy()
+	// Sync labels, annotations, taints, finalizer, and owner references onto the node.
+	r.syncNode(nodeClaim, node)
+	// Check all registration hooks before completing registration.
+	// If any hook is not ready, registration is deferred and the unregistered taint remains.
+	hooksResult, hookErrors := r.checkRegistrationHooks(ctx, nodeClaim)
+	if lo.IsEmpty(hooksResult) && hookErrors == nil {
+		// Remove karpenter.sh/unregistered taint
```
jmdeal: It's kind of obvious
```suggestion
```

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -97,6 +124,52 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+	})
+
+	// Collect pending hook names and compute the shortest requeue interval
+	var pendingHooks []string
+	mergedResult := reconcile.Result{}
+	for i, result := range results {
+		if errs[i] != nil {
+			continue
+		}
+		if !lo.IsEmpty(result) {
```
jmdeal: It's not just if there's a non-empty result, right? We also want to consider a hook pending if we got an error, at least for the purpose of the status conditions.

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -97,6 +124,52 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+			mergedResult.Requeue = mergedResult.Requeue || result.Requeue //nolint:staticcheck
+		}
+	}
+
+	if err := multierr.Combine(errs...); err != nil {
+		return mergedResult, err
+	}
+	if len(pendingHooks) > 0 {
+		log.FromContext(ctx).V(1).Info("registration hooks not satisfied", "hooks", pendingHooks)
+		nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeRegistered, "RegistrationHookPending", strings.Join(pendingHooks, ", "))
```
jmdeal, line 162: Let's make sure we update the status condition even if one of the hooks returned an error

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -97,6 +124,52 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
...
+			}
+			mergedResult.Requeue = mergedResult.Requeue || result.Requeue //nolint:staticcheck
+		}
+	}
+
+	if err := multierr.Combine(errs...); err != nil {
+		return mergedResult, err
+	}
+	if len(pendingHooks) > 0 {
+		log.FromContext(ctx).V(1).Info("registration hooks not satisfied", "hooks", pendingHooks)
```
jmdeal: nit: I think this word makes it clearer that they may eventually succeed
```suggestion
		log.FromContext(ctx).V(1).Info("awaiting registration hooks", "hooks", pendingHooks)
```

pkg/controllers/nodeclaim/lifecycle/registration.go
joshuakguo: changed it to include errored hooks too

pkg/controllers/nodeclaim/lifecycle/registration.go
joshuakguo: changed it to include errored hooks too

PR #2927, opened 2026-03-21
docs: RFC for time slicing disruption

GnatorX:
Fixes #N/A 

**Description**
This RFC proposes a new method for running disruption controller to ensure guarantee each disruption method gets time slice while preserving existing priority.

Screenshot 2026-06-15 at 11 40 47 AM

```
(
    sum by (host_cluster, msp_shard, consolidation_type, reason) (
      increase(
        karpenter_voluntary_disruption_decision_evaluation_duration_seconds_sum{}[10m]
      )
    )
  / on (host_cluster, msp_shard) group_left ()
    (
      sum by (host_cluster, msp_shard) (
        increase(
          karpenter_voluntary_disruption_decision_evaluation_duration_seconds_sum{}[10m]
        )
      )
    )
)
```

### DIFF
--- a/designs/disruption-time-slicing.md
+++ b/designs/disruption-time-slicing.md
@@ -15,6 +15,12 @@ This means consolidation can be indefinitely starved when higher-priority method
 
 Users running continuous drift (e.g., frequent AMI updates, config changes) see consolidation never run. Drift processes one node per `ComputeCommands` call, the loop restarts, and consolidation never gets a turn.
 
+### Use cases
+1. We have AMI changes every few days and a large cluster with large node pools. This means we have pretty inconsistent consolidations when large number of nodes are drifted causing wastages when consolidation are effectively paused for 20 minutes to an hour.
+2. Day to day we rely on Drift to cycle AMI, Node Class and Node Pool changes across our fleet and we don't need this to propagate quickly but we still need consolidation to run.
+3. Some times we want fast Drift to rollout, such as during emergencies, in which case we could turn off consolidations and give Drift full budget. (We have enough control already for this)
+4. This solution allows a way to reason about about changes to consolidation timeout. https://github.com/kubernetes-sigs/karpenter/issues/1733 
+
 ## Key Constraints
 
 ### Existing
--- a/designs/disruption-time-slicing.md
+++ b/designs/disruption-time-slicing.md
@@ -1,4 +1,4 @@
-# Disruption Controller Time-Slicing for Consolidation Throughput
+# CFS-Based Disruption Method Scheduling
 
 ## Motivation
 
--- a/designs/disruption-time-slicing.md
+++ b/designs/disruption-time-slicing.md
@@ -1,0 +1,306 @@
+# Disruption Controller Time-Slicing for Consolidation Throughput
+
+## Motivation
+
+The disruption controller runs as a singleton that sequentially evaluates five methods in priority order: Emptiness, StaticDrift, Drift, MultiNodeConsolidation, SingleNodeConsolidation. When any method succeeds, the loop restarts from the top:
+
+```go
+for _, m := range c.methods {
+    success, err := c.disrupt(ctx, m)
+    if success { return RequeueImmediately }  // restarts entire loop
+}
+```
+
+This means consolidation can be indefinitely starved when higher-priority methods continuously find work.
+
+Users running continuous drift (e.g., frequent AMI updates, config changes) see consolidation never run. Drift processes one node per `ComputeCommands` call, the loop restarts, and consolidation never gets a turn.
+
+## Key Constraints
+
+### Existing
+
+1. **Each disruption action mutates cluster state.** Any disruption removing nodes changes utilization, pod placement, and scheduling feasibility. Consolidation evaluating against stale state produces suboptimal or invalid decisions. Methods should evaluate settled state, not mid-mutation state.
+2. **Drift must repeat.** Unlike Emptiness and StaticDrift which batch all their work into a single call, Drift returns exactly one disruption per call. A loop that runs Drift once per cycle reduces drift throughput to ~1 node per cycle.
+3. **Consolidation is slow.** MultiNode consolidation has a 1-minute timeout. SingleNode consolidation has a 3-minute timeout.
+4. **Budgets are globally coupled.** `BuildDisruptionBudgetMapping` counts ALL `MarkedForDeletion()` nodes as "disrupting" regardless of reason. A drift budget of 10% means 10% of ALL nodes can be disrupting, and those disrupting nodes count against consolidation's budget too.
+5. **Singleton controller, no parallelism.** The disruption controller runs as a single reconciler. Only one method runs at a time.
+
+### New
+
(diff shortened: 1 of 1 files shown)


GnatorX:
Lay out use cases, what behaviors we need

GnatorX:
Added some use cases

designs/disruption-time-slicing.md
```diff
@@ -0,0 +1,312 @@
...
+The disruption controller runs as a singleton that sequentially evaluates five methods in priority order: Emptiness, StaticDrift, Drift, MultiNodeConsolidation, SingleNodeConsolidation. When any method succeeds, the loop restarts from the top:
+
+```go
+for _, m := range c.methods {
+    success, err := c.disrupt(ctx, m)
+    if success { return RequeueImmediately }  // restarts entire loop
+}
+```
+
+This means consolidation can be indefinitely starved when higher-priority methods continuously find work.
```
jmdeal, line 14: In terms of the end-result, there isn't a difference between single-node consolidation and drift. When we perform drift, we don't just launch an identical instance, we re-run the scheduling simulation to determine which NodeClaims we should create to replace the existing capacity. This is the same process as consolidation. There are two meaningful differences that I can think of:
- Drift and consolidation evaluate nodes in a different order
- Multi-node consolidation opportunities are not evaluated.

I do think both of these two are meaningful, but I was wondering if there were any other reasons you believed starvation of the single and multi-node consolidation methods were actually leading to worse cluster utilization?

jmdeal: I guess another scenario is that the nodes which need consolidation have already been processed by drift and drift is consuming the entire global budget. You could of course set the global budget to be always greater than the drift budget, but I don't think artificially lowering the floor for some methods to allow others to proceed is a great solution. Based on the rest of the RFC I think this is the main scenario your trying to address, but it wasn't immediately clear from the motivation section (if I'm correct that this is the intent).

designs/disruption-time-slicing.md
```diff
@@ -0,0 +1,312 @@
...
+2. Day to day we rely on Drift to cycle AMI, Node Class and Node Pool changes across our fleet and we don't need this to propagate quickly but we still need consolidation to run.
+3. Some times we want fast Drift to rollout, such as during emergencies, in which case we could turn off consolidations and give Drift full budget. (We have enough control already for this)
+4. This solution allows a way to reason about about changes to consolidation timeout. https://github.com/kubernetes-sigs/karpenter/issues/1733 
+
+## Key Constraints
+
+### Existing
+
+1. **Each disruption action mutates cluster state.** Any disruption removing nodes changes utilization, pod placement, and scheduling feasibility. Consolidation evaluating against stale state produces suboptimal or invalid decisions. Methods should evaluate settled state, not mid-mutation state.
+2. **Drift must repeat.** Unlike Emptiness and StaticDrift which batch all their work into a single call, Drift returns exactly one disruption per call. A loop that runs Drift once per cycle reduces drift throughput to ~1 node per cycle.
```
jmdeal, line 29: I don't think this needs to be a constraint, it's moreso an implementation detail. Granted it's not a super straightforward implementation detail to change while retaining correctness.

designs/disruption-time-slicing.md
```diff
@@ -0,0 +1,312 @@
...
+4. **Budgets are globally coupled.** `BuildDisruptionBudgetMapping` counts ALL `MarkedForDeletion()` nodes as "disrupting" regardless of reason. A drift budget of 10% means 10% of ALL nodes can be disrupting, and those disrupting nodes count against consolidation's budget too.
+5. **Singleton controller, no parallelism.** The disruption controller runs as a single reconciler. Only one method runs at a time.
+
+### New
+
+1. **Node disruption budget and time must be balanced.** Node disruption budgets limit how many nodes can be disrupting simultaneously. Consolidation methods consume minutes per evaluation while drift takes seconds. A scheduling policy must account for both.
+
+## Current Behavior
+
+The disruption controller reconciles every **10 seconds** (`pollingPeriod = 10s`). Each reconciliation runs the method list sequentially and restarts the loop on the first success. This means the maximum throughput for any method is bounded by the 10s cycle time plus the method's own execution cost.
```
jmdeal, line 40: This isn't quite accurate. The disruption controller only polls every 10 seconds if it didn't find any work to do. If, for example, it found a node to drift it will immediately requeue after enqueuing that drift command to the disruption queue.

jmdeal (commented):
checkpoint

jamesmthompson:
There are a lot of good ideas here. I'm hoping we can get there incrementally and pick up fairness as part of a larger refactor. What do we think about these two as separate but related PRs?

1. Have the disruption loop finish a full round before restarting, instead of jumping back to the top on the first action. Today any success restarts at the top, so while there's drift to work through we never reach consolidation. Running every method once per round gives consolidation a turn every round. Order still sets priority, since methods run highest to lowest. Sharing a tight budget is the open question, and per-reason budgets would answer it in the follow-up.
2. When it's drift's turn, let it update a batch of nodes instead of one. Drift does one per cycle today, below what the budget already allows, so it falls behind when a wave lands. Batching lets it keep pace. This is the dynamic Drift method, matching what StaticDrift already does.

#1 is small, and #2 is a little bigger. Would those be two PRs that we could make that would address this, and let us pick up the ideas about fairness as part of a bigger refactor? @GnatorX @jmdeal let me know what you think, happy to set up some time to chat.

designs/disruption-time-slicing.md
```diff
@@ -0,0 +1,312 @@
...
+| Emptiness   | 1 (batches all empty nodes)    | None    | All empty nodes             |
+| StaticDrift | Multiple (1 per candidate)     | None    | All static-drifted nodes    |
+| Drift       | 1 (first successful candidate) | None    | 1 node                      |
+| MultiNode   | 1 (covers multiple nodes)      | 1min    | 1 multi-node consolidation  |
+| SingleNode  | 1 (single node)                | 3min    | 1 single-node consolidation |
+
+
+### Drift throughput
+
+In the current loop, drift gets at most **1 node per 10s cycle** because a successful disruption restarts the loop. With continuous drift work, this means ~6 nodes/minute. This is adequate for small clusters but slow for large clusters with hundreds of drifted nodes. This is the motivation for time-capped drift in the proposed design, which allows drift to process multiple candidates within a single time-capped window.
```
cnmcavoy, line 56: I would also suggest that if Drifted nodeclaims are processed in batches, that they draw candidates from as diverse a selection as possible, rather than a global ordering by timestamp.

Drift today is ordered by which nodeclaims were drifted first. If multiple NodePools are drifted, the first NodePool to be seen by the NodePool controller and be reconciled will have it's nodeclaims processed first. This can greatly slow down drift processing when the nodeclaims being replaced are slow to come up (e.g GPU nodes). 

What we see in our large clusters with Karpenter is that, a nodepool with 50-100 GPU nodes, if processed first, can block Drift reconciliation for a 1000 nodeclaim NodePool with fast to replace nodes. If the Drift candidates were processed in a batch with a fair amount of eligible nodeclaims from each NodePool, Drift would be processed faster for all nodes in the cluster.

For more context: https://github.com/kubernetes-sigs/karpenter/issues/3080

cnmcavoy (approved):

PR #2928, opened 2026-03-21
feat: Do nit emit DisruptionBlocked events for do-not-disrupt annotation blocking since this is an intentional user action

grosser:
**Description**
karpenter spams our logs/events/api with these DisruptionBlocked events since our `karpenter.sh/do-not-disrupt` nodes are evaluated on each disruption cycle, the messages/events/api-calls are just noise though
image

Testing for the string is not very elegant, so open for ideas there, or if this makes sense to be an option (env var ?)

### DIFF
--- a/pkg/controllers/disruption/events/events.go
+++ b/pkg/controllers/disruption/events/events.go
@@ -18,6 +18,7 @@ package events
 
 import (
 	"fmt"
+	"strings"
 	"time"
 
 	"golang.org/x/text/cases"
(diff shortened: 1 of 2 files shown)


jmdeal:
What's the actual impact of these additional event calls? Is there a reason you want to filter this out on the server side (or producer side I guess), rather than the client side? These events, particularly those emitted against nodes, still have value in my opinion. Removing them outright would be a breaking change, albeit one that should have limited impact since listing events via operators is an anti-pattern in large clusters.

grosser:
initial impact for us was just the log spam (over quota),
but I figured removing them from the api server/etcd would be even better
also reduces noise when debugging node issues

Issue #994, opened 2024-02-07
Node Disruption Budgets should(?) be a standalone CRD


garvinp-stripe:
### Description

**What problem are you trying to solve?**
Similar to pod disruption budget, I believe we may get some benefits here by separating out NDB. 
1. We can iterate on the API without causing backward incompatibility within NodePool version
2. We can associate NDB to NodePools (or NodeClaims) allowing us to be more flexible in how Budgets are handled. This imo makes it cleaner because disruption probably should be its own separate controller without relying on NodePool or NodeClaims
- IMO disruption is orthogonal to grouping of nodes and node configurations 

3. Reusability of NDB across node pools so its easier for operator to understand disruption behaviors across many node pools/ nodes and doesn't have to repeat the same config on many node pools

Tagging @Bryce-Soghigian since we chatted a bit about this and some of the ideas here are from him
**How important is this feature to you?**

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

njtran:
We had discussed this internally a bit before. The biggest proponent of why we didn't want to do this was that having a separate NDP (which would likely have a labelSelector to reference a set of nodepools or a set of nodes) could mean that a user that has access to an NDP would have access to block disruption for NodePools it may or may not have access for.

For instance, if a user is able to create an NDP, let's say it can set the label selector to be `karpenter.sh/nodepool=a,b,c`, now that user can control the behavior of any NodePool in the cluster, where there's no RBAC that can be set to control what this NDP can reference. It's pretty reasonable to say that anyone who can set a NodePool should be able to set the limits/controls on that specific node pool, but it's almost impossible to enforce a NodePool-level of separation of responsibilities once you include it this way.

Bryce-Soghigian:
Similar to the Nodepool and provider ref concept where i can choose to link one budget to many nodepools not having to constantly redefine the configuration would be nice. You can either define one per pool or choose to define one you share across the whole cluster. Why selectors rather than ProviderRef style linking between the resources? 

[...]

PR #2930, opened 2026-03-23
docs: RFC for changing how Node Disruption Budget are defined and functions

GnatorX:
Fixes #N/A 

**Description**
  * Add design doc for HTB (Hierarchical Token Bucket) based disruption budget model
  * The current budget model is confusing: per reason budgets share a global disrupting
    counter, so they do not behave independently despite the configuration suggesting
    otherwise. "10% for drift" and "5% for consolidation" looks like a partition but
    behaves like a shared pool with soft hints.
  * HTB makes budgets mean what users write: each reason owns its configured budget
    independently, unused capacity flows to siblings via an excess pool, and the catch
    all budget acts as the parent cap.
  * Proposes a standalone DisruptionBudget CRD that NodePools reference, enabling
    cluster wide vs NodePool specific scoping and multi level budget hierarchies
    (cluster > NodePool > reason).
  * Includes walkthroughs for pure isolation (no excess pool) and borrowing (with
    excess pool) scenarios, mapping to existing API, implementation requirements,
    and backward compatibility analysis.

Design doc only, no code changes.

### DIFF
--- a/designs/htb-budget-guarantees.md
+++ b/designs/htb-budget-guarantees.md
@@ -1,0 +1,235 @@
+# HTB Based Disruption Budget Model
+
+## Context
+
+Disruption budgets today do not mean what users think they mean. The configuration syntax suggests independent, per reason control, but the runtime behavior is a single shared counter that makes per reason budgets decorative.
+
+### Budgets do not read the way they behave
+
+Disruption budgets are configured **per NodePool**, but within each NodePool, the disrupting counter is **shared across all reasons**. `BuildDisruptionBudgetMapping` counts ALL `MarkedForDeletion()` nodes as disrupting regardless of why they were marked. If Drift consumes 10 of 15 allowed disruptions, Consolidation sees only 5 remaining even if its configured budget is 5%. Drift's consumption directly reduces Consolidation's effective budget.
+
+This creates two problems for users trying to reason about their budgets:
+
+1. **Per reason budgets are not independent.** A user writing `{nodes: "10%", reasons: [Drifted]}` and `{nodes: "5%", reasons: [Underutilized]}` reads this as two separate pools. In reality, they share a single disrupting counter. The per reason budget only filters which `allowedDisruptions` value applies. The deduction is always global. The 5% consolidation budget is meaningless when drift is consuming 10%.
+
+2. **Budgets do not add up to a whole.** "10% for drift" and "5% for consolidation" looks like it partitions 15% of disruption capacity across two reasons. Instead, both budgets draw from the same counter. The catch all budget (no reasons specified) acts as the true cap, but its relationship to the per reason budgets is not obvious. The YAML looks like a partition but behaves like a shared pool with soft hints.
+
+### What HTB gives us
+
+The Hierarchical Token Bucket model makes the YAML mean what it says:
+
+- **The catch all budget becomes the parent**: it defines the total disruption capacity for the NodePool.
+- **Per reason budgets become children**: each owns its slice of disruption capacity independently.
+- **Budgets add up the way users expect**: "10% for drift" plus "5% for consolidation" totals 15%. Any difference between that sum and the parent is a shared excess pool available to any reason.
+- **The hierarchy can extend to multiple scopes**: per reason within a NodePool, and per NodePool within a cluster-wide cap.
+
+The result is a budget model that is more powerful but also easier to understand: what you write is what you get.
+
+## HTB Background
+
+HTB is a Linux traffic control algorithm (`tc-htb`) used for bandwidth management. Each class in the hierarchy has:
+
+- **`rate`**: guaranteed bandwidth. This class always gets at least this much, regardless of what siblings are doing.
+- **`ceil`**: maximum bandwidth including borrowed capacity. A class can burst above its rate by borrowing unused bandwidth from siblings, up to this ceiling.
+- **parent**: enforces the global cap across all children.
+
+When a class is not using its full `rate`, the unused portion flows up to the parent and becomes available to sibling classes. A class consuming above its `rate` is borrowing from the shared surplus. The parent's `rate` is the hard ceiling for all children combined.
+
+HTB is **work conserving up to each class's `ceil`**. No capacity sits idle when there is demand for it. If a class has no traffic, its entire `rate` is available to siblings that want to borrow. However, a borrowing class will never exceed its own `ceil`, even if more surplus is available. This means capacity is fully utilized across classes that have work to do, while each class's maximum consumption is still bounded.
+
+## Mapping HTB to Disruption Budgets
+
+| HTB concept | Disruption budget equivalent |
+|------------|------------------------------|
+| Class | Disruption reason (Drifted, Underutilized) |
+| `rate` | `reserved[reason]`: guaranteed minimum disruptions for this reason |
+| `ceil` | `global_cap`: maximum disruptions a reason can reach by borrowing |
+| Parent `rate` | `global_cap`: total allowed disruptions across all reasons |
+| Surplus from idle sibling | Excess pool: unused reservations available to other reasons |
+
+## Proposed Model
+
+Each reason becomes an HTB class with its own `rate` (the budget the user configured) and a `ceil` equal to the global cap. When a reason is not using its full budget, the unused portion is available to other reasons. This means budgets read as independent slices but unused capacity is not wasted.
+
+### Definitions
+
+```
+global_cap (parent rate) = total allowed disruptions (from catch all budget)
+rate[reason]             = guaranteed minimum for this reason (from reason specific budget)
+ceil[reason]             = global_cap (all reasons can burst up to global cap)
+used[reason]             = nodes currently disrupting for this reason
+excess_pool              = global_cap - sum(rate[all_reasons])
+```
+
+### Budget Computation
+
+For a given reason R:
+
+```
+own_remaining  = max(rate[R] - used[R], 0)
+used_from_pool = max(total_used - sum(min(used[reason], rate[reason]) for all reasons), 0)
+free_pool      = max(excess_pool - used_from_pool, 0)
+
+available[R]   = own_remaining + free_pool
+```
+
+In HTB terms: a class first consumes from its guaranteed `rate`. Once the `rate` is exhausted, it borrows from the parent's excess pool (surplus above all children's combined `rate`). It can borrow up to its `ceil` (which equals the global cap). The parent ensures total consumption never exceeds `global_cap`.
+
+### Requirement: global_cap >= sum(rate)
+
+If `global_cap == sum(rate)`: no excess pool, each reason is strictly limited to its `rate`. Pure isolation.
+
+If `global_cap > sum(rate)`: the difference is the excess pool. Reasons can burst above their `rate` by borrowing from it.
+
+If `global_cap < sum(rate)`: invalid configuration. The sum of guarantees exceeds the global cap. This should be rejected at validation time.
+
+### Walkthrough: No Excess Pool
+
+**Setup**: 100-node pool. Global cap 15 nodes. Drift rate 10. Consolidation rate 5. Excess pool = 0.
+
+Each reason is hard limited to its rate:
+
+| Scenario | Drift used | Consol used | Drift available | Consol available |
+|----------|-----------|-------------|-----------------|------------------|
+| Idle | 0 | 0 | 10 | 5 |
+| Drift active | 8 | 0 | 2 | 5 |
+| Drift at rate | 10 | 0 | 0 | 5 |
+| Both active | 10 | 3 | 0 | 2 |
+| Both at rate | 10 | 5 | 0 | 0 |
+
+The budget behaves the way the YAML reads: Drift gets 10, Consolidation gets 5, and neither interferes with the other.
+
+### Walkthrough: With Excess Pool
+
+**Setup**: 100-node pool. Global cap 20 nodes. Drift rate 10. Consolidation rate 5. Excess pool = 5.
+
+| Scenario | Drift used | Consol used | Pool used | Drift available | Consol available |
+|----------|-----------|-------------|-----------|-----------------|------------------|
+| Idle | 0 | 0 | 0 | 10 + 5 = 15 | 5 + 5 = 10 |
+| Drift active | 8 | 0 | 0 | 2 + 5 = 7 | 5 + 5 = 10 |
+| Drift at rate | 10 | 0 | 0 | 0 + 5 = 5 | 5 + 5 = 10 |
+| Drift borrowing | 13 | 0 | 3 | 0 + 2 = 2 | 5 + 2 = 7 |
+| Drift at global cap | 15 | 0 | 5 | 0 + 0 = 0 | 5 + 0 = 5 |
+| Both borrowing | 13 | 5 | 3 | 0 + 2 = 2 | 0 + 2 = 2 |
+
+Key observations:
+- Drift can exceed its 10-node budget by borrowing from the excess pool when other reasons are not using it.
+- Even when Drift borrows all 5 excess nodes, Consolidation still has its configured 5. The budget reads correctly.
+- Unused capacity is not wasted. If one reason is idle, others can use the slack. But each reason's configured budget is always respected.
+
+## Mapping to the Existing Budget API
+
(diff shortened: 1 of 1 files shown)


GnatorX:
https://github.com/kubernetes-sigs/karpenter/issues/994

nathangeology:
Hey Garvin, It was good chatting yesterday. This comment isn't on the specifics of the proposal, but could maybe reframe and simplify this proposal a bit. I was thinking, what if we add an annotation to the nodepool like 'expected-time-to-drift' or 'drift-time-sla' or something like that. You could set 7 days or 30 minutes or whatever there. Then, Karpenter does two things differently from today. One, it rate limits the drift budget such that it will complete a each new drift 'action' in the given SLA. There's some thought that needs to go into how define and track that. We also need to think about what happens to drift if the label is raised or lower mid-drift (I think we want it top speed up or slow down drift post annotation update). Lastly, we want to think about the overall rate limit and how we handle cases where the disruption limits prevent achieving the drift SLA in the given time (probably a warning emitted + respect the limit and complete the drift as fast as possible?). This sort of controls and separates out the drift disruption from the other efficiency based disruptions and gives a (hopefully) clear interface for setting the amount of disruption due to drift that should be happening. The second big change to how Karpenter handles this is that we should consider adjusting the candidate sorting for single/multi-node consolidation to put nodes that are pending a drift action first in the sorting. This makes it so we are more likely to 'knock two birds out with one stone' and boost both efficiency and on-going drifts at the same time. What do you think? Does this make sense as an interface + actions relative the issues you deal with your clusters? I like the idea of aligning the simple-as-possible-interface with the end goal (getting drifts done in a timely fashion without blocking cost saving disruptions too much) instead of making budgets have a lot more nuance and settings to manage. There are still some things we could do from your other proposal on fairness in conjunction with this to ensure that multi/single node disruption isn't totally starved by an aggressive drift too.

GnatorX:
Need some time to digest the full comment since I am currently out for the next 2 weeks.

My initial gut feeling on 'expected-time-to-drift' annotation is I am not a huge fan. I don't like the idea of using annotations to define behaviors and I would rather have a cleaner solution.

I am not sure the current state of cross disruption tracking but given the current single threaded disruption I feel like separating out drift might not do as much good until we have stronger concurrent disruption within Karpenter even though I do like the idea of having separation categories of disruptions that might be able to action independently. 

I still think https://github.com/kubernetes-sigs/karpenter/pull/2927 has a lot of value that solves a fundamental issue with how Karpenter works today without huge rework. I am very open to looking into how we want disruptions to look like. 

Will comment once I read through and think through the comment more deeply.

nathangeology:
> Need some time to digest the full comment since I am currently out for the next 2 weeks.
> 
> My initial gut feeling on 'expected-time-to-drift' annotation is I am not a huge fan. I don't like the idea of using annotations to define behaviors and I would rather have a cleaner solution.
> 
> I am not sure the current state of cross disruption tracking but given the current single threaded disruption I feel like separating out drift might not do as much good until we have stronger concurrent disruption within Karpenter even though I do like the idea of having separation categories of disruptions that might be able to action independently.
> 
> I still think #2927 has a lot of value that solves a fundamental issue with how Karpenter works today without huge rework. I am very open to looking into how we want disruptions to look like.
> 
> Will comment once I read through and think through the comment more deeply.

Thanks, as I've been thinking about this more deeply one question that keeps coming up for me is why we aren't doing more to integrate drift with the efficiency consolidation. Consider the difference if we pushed non-static drift the back of the consolidation stack, but at the same time we changed the candidate sorting to put all the drifted nodes first in line to be considered. This would mean that the drifted nodes with the lowest pod counts would be first in line, followed by drifted nodes with higher pods counts, followed by everything else sorted as it is today. This would give you the chance to find 'two-birds-with-one-stone' consolidations first. Then, after you would still do the current non-static drift disruption logic to keep the ball rolling there and avoid starving out drifts. I still think that saying, "Hey I want my drifts to complete in 7 days or 1 hour (ASAP basically)" ought to be an easy way to communicate the urgency of executing drifts and let you sort of auto-magically parcel out the right amount of disruption budget to execute drifts on time based on time remaining, remaining work, and total available disruptions during the remaining time. I think if you combined that with some changes to help consolidation find consolidations among drift candidates first would go a long way towards avoiding one or the other getting starved. Would we need this and the other PR on fairness in that world? Maybe! We'd have to start actually trying stuff and see how it goes I think. I'll talk with Derek and Jason soon to see what they think about all this too and maybe we can do a call soon. I'll be out all next week so it'll have to wait until the week of the 13th one way or another.

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
+# HTB Based Disruption Budget Model
+
+## Context
+
+Disruption budgets today do not mean what users think they mean. The configuration syntax suggests independent, per reason control, but the runtime behavior is a single shared counter that makes per reason budgets decorative.
+
+### Budgets do not read the way they behave
+
+Disruption budgets are configured **per NodePool**, but within each NodePool, the disrupting counter is **shared across all reasons**. `BuildDisruptionBudgetMapping` counts ALL `MarkedForDeletion()` nodes as disrupting regardless of why they were marked. If Drift consumes 10 of 15 allowed disruptions, Consolidation sees only 5 remaining even if its configured budget is 5%. Drift's consumption directly reduces Consolidation's effective budget.
```
nathangeology, line 9: I agree we should do something here to fix how confusing this budget definition stuff is!

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+2. **Budgets do not add up to a whole.** "10% for drift" and "5% for consolidation" looks like it partitions 15% of disruption capacity across two reasons. Instead, both budgets draw from the same counter. The catch all budget (no reasons specified) acts as the true cap, but its relationship to the per reason budgets is not obvious. The YAML looks like a partition but behaves like a shared pool with soft hints.
+
+### What HTB gives us
+
+The Hierarchical Token Bucket model makes the YAML mean what it says:
+
+- **The catch all budget becomes the parent**: it defines the total disruption capacity for the NodePool.
+- **Per reason budgets become children**: each owns its slice of disruption capacity independently.
+- **Budgets add up the way users expect**: "10% for drift" plus "5% for consolidation" totals 15%. Any difference between that sum and the parent is a shared excess pool available to any reason.
```
nathangeology, line 23: This is where communicating drift 'urgency' may be helpful because you can say "hey, I have 100 drifted nodes, but 6 days to manage them, maybe I shouldn't burn all of my disruption budget on these drifts right now, but do just enough to be on track to complete them before 6 days is up". Maybe this token budget model is how we do this under-the-hood, but I think as long as the drifts are on track to be completed in a timely fashion we don't need to take up all the budget.

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+## HTB Background
+
+HTB is a Linux traffic control algorithm (`tc-htb`) used for bandwidth management. Each class in the hierarchy has:
+
+- **`rate`**: guaranteed bandwidth. This class always gets at least this much, regardless of what siblings are doing.
+- **`ceil`**: maximum bandwidth including borrowed capacity. A class can burst above its rate by borrowing unused bandwidth from siblings, up to this ceiling.
+- **parent**: enforces the global cap across all children.
+
+When a class is not using its full `rate`, the unused portion flows up to the parent and becomes available to sibling classes. A class consuming above its `rate` is borrowing from the shared surplus. The parent's `rate` is the hard ceiling for all children combined.
```
nathangeology, line 36: The fact that unused rate bubbles up here worries me because the consolidation functions are behind the drift ones in the order and could end up getting starved still if the 'rate' looks like it is there when drift runs (when really it just hasn't looked at consolidation opportunities yet).

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+HTB is **work conserving up to each class's `ceil`**. No capacity sits idle when there is demand for it. If a class has no traffic, its entire `rate` is available to siblings that want to borrow. However, a borrowing class will never exceed its own `ceil`, even if more surplus is available. This means capacity is fully utilized across classes that have work to do, while each class's maximum consumption is still bounded.
```
nathangeology, line 38: This addresses my worry above. This would solve that, but its adding a little bit of complexity for operators to reason about here. How do we communicate to operators when their settings are problematic and how to fix it when that happens? Also, drift and consolidation for efficiency aren't mutually exclusive and we could be doing more to find 'two-birds-with-one-stone' disruptions here.

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+|------------|------------------------------|
+| Class | Disruption reason (Drifted, Underutilized) |
+| `rate` | `reserved[reason]`: guaranteed minimum disruptions for this reason |
+| `ceil` | `global_cap`: maximum disruptions a reason can reach by borrowing |
+| Parent `rate` | `global_cap`: total allowed disruptions across all reasons |
+| Surplus from idle sibling | Excess pool: unused reservations available to other reasons |
+
+## Proposed Model
+
+Each reason becomes an HTB class with its own `rate` (the budget the user configured) and a `ceil` equal to the global cap. When a reason is not using its full budget, the unused portion is available to other reasons. This means budgets read as independent slices but unused capacity is not wasted.
```
nathangeology, line 52: What do you imagine would be reasonable defaults here? When and why should people change those?

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+In HTB terms: a class first consumes from its guaranteed `rate`. Once the `rate` is exhausted, it borrows from the parent's excess pool (surplus above all children's combined `rate`). It can borrow up to its `ceil` (which equals the global cap). The parent ensures total consumption never exceeds `global_cap`.
+
+### Requirement: global_cap >= sum(rate)
+
+If `global_cap == sum(rate)`: no excess pool, each reason is strictly limited to its `rate`. Pure isolation.
+
+If `global_cap > sum(rate)`: the difference is the excess pool. Reasons can burst above their `rate` by borrowing from it.
+
+If `global_cap < sum(rate)`: invalid configuration. The sum of guarantees exceeds the global cap. This should be rejected at validation time.
```
nathangeology, line 84: How should we reject this? Do we log an warning and scale the sum(rate) components down to the global cap? Do we not allow operators to even deploy/apply this config?

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Mapping to the Existing Budget API
+
+Today's NodePool budget spec already expresses the HTB hierarchy:
+
+```yaml
+spec:
+  disruption:
+    budgets:
+    - nodes: "15%"                           # no reasons = parent rate (global cap)
```
nathangeology, line 128: How do we express the ceil values? (I'll read down further, maybe its explained further on)

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+  ├── NodePool "critical"
+  │     rate: 10 nodes         # guaranteed minimum for this pool
+  │     ceil: 20 nodes         # can borrow up to 20 from cluster surplus
+  │     │
+  │     ├── Drifted:       rate 6
+  │     └── Underutilized: rate 4
+  │
+  ├── NodePool "batch"
+  │     rate: 30 nodes         # batch workloads get more disruption capacity
+  │     ceil: 50 nodes         # can use full cluster budget if others are idle
```
nathangeology, line 161: Ah. So the ceil is computed on the allocated rate + the 'excess' auto-magically here?

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+// Current: global only
+available = allowedDisruptions - disrupting_all
+```
+
+With HTB computation:
+
+```
+// HTB: per reason with borrowing from excess pool
+own_remaining = max(rate[reason] - used[reason], 0)
+free_pool     = max(excess_pool - pool_used, 0)
```
nathangeology, line 208: Definitely that agree there is value is having some different budget pools instead of just the one global one in the code.

nathangeology: Also helps that the proposed pools are tied to different functions atm so there is some natural scope for each budget to work in.

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+### 4. Validation
+
+Add validation to reject configurations where `sum(rate) > global_cap`. The sum of per reason budgets cannot exceed the global cap because the parent cannot allocate more capacity than it has.
+
+## Backward Compatibility
+
+With a standalone DisruptionBudget CRD, backward compatibility is straightforward. The existing inline NodePool budgets keep their current behavior. HTB semantics only apply when a NodePool references a DisruptionBudget CR. Users opt in explicitly by creating the CR and wiring the reference. No silent semantic changes, no feature flags needed.
+
+## Open Questions
```
nathangeology, line 229: It would be good to walk through some alternative ways to that this could be approached or solved, discuss the pros/cons relative to this proposal and give some reasons why this is the better choice. I'd be particularly interested in you describing the simplest possible solution you could imagine to solving the problems you are seeing with starvation and why going beyond that for this proposal is worthwhile. I do think the budget definitions need work such that the interface presented lines up with the system behavior better, its genuinely confusing as things stand and its not easy to diagnose issues like you mentioned you were having with drift starving out consolidation. I would also encourage thinking through ways were we can get better at finding candidates that make the cluster more efficient and keep the drift progressing at the same time. Even fairly efficient nodes slated for drift might make a lot of sense if we can get the disrupted pods onto existing post-drift nodes. Also, we should think through how to organically push new pod placements away from nodes that need to be drifted and replicasets towards terminating pods on nodes slated for drift before nodes that are up-to-date. There's a lot of ways to approach making less work for disruption to do here and to steer the disruptions towards nodes that need to go anyway.

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+
+## Context
+
+Disruption budgets today do not mean what users think they mean. The configuration syntax suggests independent, per reason control, but the runtime behavior is a single shared counter that makes per reason budgets decorative.
+
+### Budgets do not read the way they behave
+
+Disruption budgets are configured **per NodePool**, but within each NodePool, the disrupting counter is **shared across all reasons**. `BuildDisruptionBudgetMapping` counts ALL `MarkedForDeletion()` nodes as disrupting regardless of why they were marked. If Drift consumes 10 of 15 allowed disruptions, Consolidation sees only 5 remaining even if its configured budget is 5%. Drift's consumption directly reduces Consolidation's effective budget.
+
+This creates two problems for users trying to reason about their budgets:
```
nathangeology, line 11: I like coming at this from the perspective of configuration and understanding. I do think it would be helpful to walk through some examples or cases where bad stuff happens because of the current state of things. In the meeting you talked about how drift can 'starve' out consolidation. I think that would be really good to walk folks through and explain how and why that happens here.

designs/htb-budget-guarantees.md
```diff
@@ -0,0 +1,235 @@
...
+| Drift at rate | 10 | 0 | 0 | 0 + 5 = 5 | 5 + 5 = 10 |
+| Drift borrowing | 13 | 0 | 3 | 0 + 2 = 2 | 5 + 2 = 7 |
+| Drift at global cap | 15 | 0 | 5 | 0 + 0 = 0 | 5 + 0 = 5 |
+| Both borrowing | 13 | 5 | 3 | 0 + 2 = 2 | 0 + 2 = 2 |
+
+Key observations:
+- Drift can exceed its 10-node budget by borrowing from the excess pool when other reasons are not using it.
+- Even when Drift borrows all 5 excess nodes, Consolidation still has its configured 5. The budget reads correctly.
+- Unused capacity is not wasted. If one reason is idle, others can use the slack. But each reason's configured budget is always respected.
+
```
nathangeology, line 119: Having a walk through of a toy/example drift starving consolidation case without these changes would help here too!

nathangeology (commented):
Thanks for putting this together. I hope the feedback is useful to you. I'll be out next week so I might slow to respond, but happy to collaborate a bit on this when I get back the week of the 13th.

PR #2931, opened 2026-03-25
fix: use accumulator in lo.Reduce for disruption pod count

### LINKED ISSUE #2888: lo.Reduce in Command.LogValues discards accumulator, reports last candidate's pod count instead of total
## Bug

In `pkg/controllers/disruption/types.go`, `LogValues()` computes `podCount` using `lo.Reduce`:

```go
podCount := lo.Reduce(c.candidates, func(_ int, cd *Candidate, _ int) int {
    return len(cd.reschedulablePods)
}, 0)
```

The first `_ int` parameter is the accumulator. Because it's discarded, the function returns `len(cd.reschedulablePods)` for the **last** candidate only, not the sum across all candidates.

## Expected

```go
podCount := lo.Reduce(c.candidates, func(acc int, cd *Candidate, _ int) int {
    return acc + len(cd.reschedulablePods)
}, 0)
```

## Impact

The `"pod-count"` field in structured disruption logs under-reports when multiple nodes are disrupted simultaneously (e.g., multi-node consolidation). Single-candidate disruptions are unaffected since last == only.

## History

The line dates back to the original `String()` method (before commit c0e7299 refactored it to `LogValues()`). The refactor carried it over unchanged.

nicknikolakakis:
## Summary

The `lo.Reduce` callback in `Command.LogValues()` discards the accumulator (`_ int`), so the `pod-count` field in disruption logs only reports the last candidate's pod count instead of the sum across all candidates.

**Before:** `func(_ int, cd *Candidate, _ int) int { return len(cd.reschedulablePods) }`
**After:** `func(acc int, cd *Candidate, _ int) int { return acc + len(cd.reschedulablePods) }`

Single-candidate disruptions were unaffected (last == only), but multi-node consolidation under-reported.

Fixes #2888

## Test plan

- [x] `go build ./pkg/controllers/disruption/...` — builds clean
- [ ] Existing disruption tests (CI)

### DIFF
--- a/pkg/controllers/disruption/types.go
+++ b/pkg/controllers/disruption/types.go
@@ -276,7 +276,7 @@ func (c Command) EmitRejectedEvents(recorder events.Recorder, reason string) {
 }
 
 func (c Command) LogValues() []any {
-	podCount := lo.Reduce(c.Candidates, func(_ int, cd *Candidate, _ int) int { return len(cd.reschedulablePods) }, 0)
+	podCount := lo.Reduce(c.Candidates, func(acc int, cd *Candidate, _ int) int { return acc + len(cd.reschedulablePods) }, 0)
 
 	candidateNodes := lo.Map(c.Candidates, func(candidate *Candidate, _ int) interface{} {
 		return map[string]interface{}{


jmdeal:
Closing in favor of https://github.com/kubernetes-sigs/karpenter/pull/2912

PR #2933, opened 2026-03-25
test: Allow for overhead in karpenter resource utilization for perf tests

AndrewMitchell25:
Fixes #N/A 

**Description**
Adds support for overhead in karpenter resource utilization for perf tests, so cloud providers running this suite can set the thresholds based on specific resource utilization. Also makes the profiler more robust to panics and potential failures.

Manually tested overhead logic and manually ran performance tests

### DIFF
--- a/test/pkg/environment/common/karpenter_profiler.go
+++ b/test/pkg/environment/common/karpenter_profiler.go
@@ -21,15 +21,10 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strings"
 	"time"
 
 	"github.com/google/pprof/profile"
 	. "github.com/onsi/ginkgo/v2"
-	"github.com/samber/lo"
-	coordinationv1 "k8s.io/api/coordination/v1"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
 )
 
 const CPUProfileSeconds = 20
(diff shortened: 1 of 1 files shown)


test/pkg/environment/common/karpenter_profiler.go
```diff
@@ -84,8 +91,21 @@ func (kp *KarpenterProfiler) captureProfiles(ctx context.Context) {
 		}
 	}()
 
-	pod := kp.env.ExpectActiveKarpenterPod()
```
DerekFrank, line 87: Was was this expanded?

AndrewMitchell25: Because this function uses an Expect it can cause a panic if the Expect fails, for example if the karpenter pods are in CrashLoopBackOff and no pod is holding the lease then you get a panic from trying to find a pod named "". Rewriting the function without the expect here makes it more robust by handling the errors normally instead of panicking

DerekFrank: I think we can use recover here and still use the Expect() function: https://go.dev/blog/defer-panic-and-recover

AndrewMitchell25: Updated

DerekFrank (approved):
/lgtm

PR #2935, opened 2026-03-26
docs: RFC for building a pod deletion cost controller to enable Karpenter to work better with the ReplicaSet Controller

nathangeology:
## Pod Deletion Cost Controller RFC

**Status:** Iterating on reviewer feedback (Derek, Garvin). v4 simplification pushed: anchored to next-K8s-minor forced cutover, dropped the three-annotation coexistence protocol.

### What this RFC proposes

A feature-gated controller that ranks Karpenter-managed nodes by their consolidation priority and propagates that ranking to pods via `controller.kubernetes.io/pod-deletion-cost` annotations. This aligns ReplicaSet scale-down decisions with Karpenter's consolidation targets, reducing voluntary pod disruption during scale-in events.

This also introduces `karpenter.sh/disruption-cost` as the user-facing annotation for steering Karpenter consolidation, replacing direct use of `controller.kubernetes.io/pod-deletion-cost` for that purpose.

### Key design decisions (v4)

**Two annotations, two purposes, one writer per annotation:**
- `controller.kubernetes.io/pod-deletion-cost` — Karpenter writes this on pods on Karpenter-managed nodes (gate-ON only) so the ReplicaSet controller's scale-down ordering aligns with Karpenter's consolidation ranking. ReplicaSet controller is the consumer.
- `karpenter.sh/disruption-cost` — new user-facing annotation. Customers set it; Karpenter reads it for consolidation scoring; Karpenter never writes it.

**Node categories (not ranked equally):**
- Draining: fixed MinInt32 (highest priority, not sequentially ranked)
- Drifted: ranked by consolidation priority
- Disruptable: ranked by consolidation priority
- Not Disruptable: annotations cleared (no Karpenter influence)

**Same heuristic as consolidation:** The ranking intentionally uses the same disruption cost calculation as consolidation candidate selection. This RFC does not prescribe how that cost is computed, only that both systems use the same function.

**Budget-limited scope:** Annotation writes are bounded by per-NodePool disruption budgets (drift budget for Drifted, consolidation budget for Disruptable) with a hard cap of 50 nodes per cycle.

**Migration anchored to next K8s minor:** Pre-cutover (alpha) is read-only with fallback to the legacy annotation; from beta onward, only `karpenter.sh/disruption-cost` is honored by Karpenter consolidation. The legacy annotation continues to work for ReplicaSet's own scale-down ordering (upstream Kubernetes behavior, unchanged).

**Observability:** Prometheus metrics only (no events). Metrics cover nodes ranked, pods updated, ranking duration, and cycles skipped.

**Graduation:** Alpha (gate off) -> Beta (gate on, fallback removed, cleanup-on-disable) -> GA (evaluate gate removal).

### Changes since last review (v3 -> v4)

- Removed the three-annotation ownership/conflict-detection protocol (`managed-deletion-cost`, `last-assigned-deletion-cost`). Using the feature flag to gate is sufficient in alpha, can revisit for beta if needed.
- "Skip pods with customer-set deletion costs" branch in the reconcile loop removed (Karpenter no longer needs to recognize and protect customer-set values, because gate-OFF is write-disabled and gate-ON is post-cutover).
- "Annotation conflicts with customer-set pod deletion costs" risk replaced by "Customers who miss the migration window".
- Migration section rewritten to describe the forced-cutover boundary plainly.

### Related

- Implementation: #2894
- Disruption metrics: #2892
- Balanced Consolidation: #2962
- RS Controller KEP: kubernetes/enhancements#5982

---

Does this change impact any existing behavior?

No. This PR adds a design document only. The proposed controller is feature-gated off by default and requires explicit opt-in.


designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,340 @@
...
+
+### TODO: Changing the replicaset controller behavior directly
+
+TODO: Came up on 3/20 in a convo with James. This note is to remind me to address this possibility here.
+
+### Do nothing
+
+The ReplicaSet controller's scale-down decisions and Karpenter's consolidation decisions remain completely uncoordinated. Scale-down scatters deletions uniformly across nodes. No single node moves closer to empty. For ConsolidateWhenEmpty NodePools, the cluster carries more nodes than necessary. For clusters with do-not-disrupt workloads, the ReplicaSet controller may delete pods from nodes Karpenter can't act on anyway. Over time, this manifests as higher infrastructure cost and slower convergence to optimal cluster size.
+
+## Risks and Mitigations
```
nathangeology, line 181: Note to self to address topology spreading and topology here.

nathangeology (commented):
Left a note to myself to address spreading constraints in the RFC and how to think about those. This proposal focuses on binpacking and doesn't address that topic enough yet.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Ranking strategy
+
+The controller ranks nodes the same way Karpenter ranks consolidation candidates: by pod count (disruption cost). Within this ranking, drifted nodes (`ConditionTypeDrifted=True`) are placed in the lowest-cost tier, ahead of non-drifted nodes. Since drifted nodes need replacement anyway, directing RS deletions toward them first means scale-down events naturally assist drift progress without additional disruptions. This maximizes the mutual information between the hint we send and the action Karpenter will take next, because the ranking signal and the consolidation signal are literally the same function. As Karpenter's consolidation candidate sorting evolves, we plan to follow it here to stay aligned.
```
GnatorX: I think we have to include AZ spread and other topology constraints into deletion cost or else Karpenter would cause skew during autoscaling events.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
...
+
+1. Karpenter ranks nodes by consolidation priority
+2. Pods on those nodes get the lowest deletion costs
+3. ReplicaSet scale-down removes pods from those nodes first
+4. Those nodes become empty or closer to empty
+5. Karpenter can consolidate them with less disruption (or they qualify for WhenEmpty consolidation)
+
+This logic extends naturally to nodes that Karpenter cannot consolidate and to nodes that are already marked for replacement. The ranking engine partitions nodes into three groups before ranking:
+
+- **Group A (Drifted):** Nodes with `ConditionTypeDrifted=True` and no do-not-disrupt pods. These nodes need replacement anyway for compliance or security reasons. Draining them first via RS scale-down means scale-down events naturally assist drift progress. Sorted by pod count ascending (fewest pods = lowest cost = drained first).
```
GnatorX: I think there should be a 4th group which are pods running on nodes that are has started disruption but can't be drained due to PDB. This group should have the lowest deletion cost. We want those pods that might be holding PDBs to go and be prioritized by replicaset controller

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
...
+### How it works
+
+All new code lives in `pkg/controllers/pod/deletioncost/`. A singleton reconciler runs every 60 seconds. The 60-second interval balances annotation freshness against API server write load — fast enough to react to scale events within one HPA evaluation period, slow enough to avoid amplifying watch events during steady state. On each tick it:
+
+1. Checks the `PodDeletionCostManagement` feature gate
+2. Collects all Karpenter-managed nodes from `state.Cluster`
+3. Runs change detection (compares `ConsolidationState` timestamp from `state.Cluster`). If nothing changed, skips with zero API writes
+4. Partitions nodes into Group A (drifted, no do-not-disrupt pods), Group B (not drifted, no do-not-disrupt pods), and Group C (has do-not-disrupt pods)
+5. Ranks each group independently by pod count ascending
+6. Assigns sequential integer ranks starting at -n (where n is the total number of Karpenter-managed nodes) for Group A, continuing for Group B, then Group C
```
GnatorX: I think you need to go backwards.  The problem is unprocessed pods will have no deletion cost meaning it would be cheaper to delete. If we process going from those we want to kill first and downwards then the opposite effect would happen until the whole cluster has been processed. We should be going from lowest priority first setting higher weights such that those that are unprocessed will be prioritized naturally

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
+# Pod Deletion Cost Controller
+
+## Summary
+
+This RFC proposes a new feature-gated controller for Karpenter that ranks nodes by consolidation preference and propagates that ranking to pods via the `controller.kubernetes.io/pod-deletion-cost` annotation. By aligning the ReplicaSet controller's scale-down decisions with Karpenter's consolidation targets, we measurably reduce voluntary pod disruption rate. This is tracked and validated via the pod disruption metrics being added in [kubernetes-sigs/karpenter#2892](https://github.com/kubernetes-sigs/karpenter/pull/2892). The feature is off by default and configured entirely through feature gates and CLI flags, with no CRD changes required.
```
DerekFrank: We need an answer for what we do for our existing EvictionCost calculation, this is a breaking change: 

https://github.com/kubernetes-sigs/karpenter/blob/eda8625eee05358a5eb4beb8a49eb35d1e469fa3/pkg/utils/disruption/disruption.go#L48

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
...
+
+### Related Proposals that can reduce the pod disruption rate
+
+- Most important: Consolidation cost thresholding -- see draft RFC
+
+## Proposal
+
+We introduce a new feature-gated controller (`pod.deletioncost`) that automatically manages the `controller.kubernetes.io/pod-deletion-cost` annotation on pods running on Karpenter-managed nodes. The controller ranks nodes by pod count (mirroring Karpenter's own consolidation candidate sorting), assigns sequential deletion cost values to pods so that Kubernetes' ReplicaSet scale-down logic preferentially removes pods from the "least valuable" nodes first, and partitions nodes hosting do-not-disrupt pods separately to protect them from early eviction.
```
DerekFrank: We might need to discuss the race conditions around other controllers setting `controller.kubernetes.io/pod-deletion-cost`

designs/pod-deletion-cost-controller.md
nathangeology: It's a good callout, but keep in mind that new pods default to a deletion cost of 1 so when we set the annotations negative that would mean that new pods should always rank above already annotated pods. So for the number 1 node we want to consolidate, we'd rank it the highest and then give the hosted pods the lowest pod deletion cost (say -1000), then the # ranked node's pods would be ranked -999, and so on.

designs/pod-deletion-cost-controller.md
nathangeology: So this would be pods on nodes with the disrupted annotation applied? It does make sense to me, that we'd want to encourage the RS controller to go for those first. I hadn't thought too much about them because you hope that the draining process proceeds quickly, but when a PDB blocks it, I can see the need there. Side note, it would be good not to disrupt nodes that cannot be drained due to PDBs in the first place, but that might be tricky to check in a timely fashion during consolidation.

designs/pod-deletion-cost-controller.md
nathangeology: I think we can evolve this mechanism over time to handle competing concerns like this. Right now, the system defaults towards a natural spreading heuristic and this change would encourage emptying of nodes which could cause the replica set to have imbalanced distribution of pods, but not necessarily. A lot of that can be mitigated in how the pod placement/scheduling config is done. If the pod replicas are already spread out well, it won't matter overly much which ones are picked and this would help you coordinate the scale-in across replicasets/deployments better. It's worth noting that this sort of imbalance can happen today anyway with the current RS behavior because it doesn't account for placement constraints or preferences at all in its decision and if you don't get good spreading at placement time, you won't have it post replica scale-in either. So its not an automatic regression from today's behavior and I think Karpenter is the right place to work through these trade-offs.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Risks and Mitigations
+
+- **Annotation conflicts with customer-set pod deletion costs (Medium):** An operator or another controller may have already set `controller.kubernetes.io/pod-deletion-cost` on pods. If Karpenter overwrites those values, it silently breaks the operator's intended behavior. *Mitigation:* Two-annotation protocol. Karpenter also sets `karpenter.sh/managed-deletion-cost: "true"` when it manages a pod. If a pod has a deletion cost but lacks the management annotation, it's treated as customer-managed and skipped entirely. Existing customer annotations are never overwritten. Additionally, Karpenter-managed ranks are always negative (starting at -n), so customer-set positive values naturally take precedence in the RS controller's sort order without any special handling.
```
nathangeology: Talked with Derek and leaving myself a note here. We need to post the last ranking/value that Karpenter set as an annotation here too such that we can see if another controller or the cluster operator is changing the values (even if the controller restarts). This will protect us from fighting/racing a third party for the right to control the pod deletion cost.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
...
+
+1. **Ship the pod-deletion-cost controller now** (this RFC) — immediate value, no upstream dependency
+2. **Pursue the RS controller KEP in parallel** — broader impact, zero-overhead solution
+3. **Deprecate the annotation controller** once the RS controller change reaches GA and the minimum supported Kubernetes version includes it
+
+### Do nothing
+
+The ReplicaSet controller's scale-down decisions and Karpenter's consolidation decisions remain completely uncoordinated. Scale-down scatters deletions uniformly across nodes. No single node moves closer to empty. For ConsolidateWhenEmpty NodePools, the cluster carries more nodes than necessary. For clusters with do-not-disrupt workloads, the ReplicaSet controller may delete pods from nodes Karpenter can't act on anyway. Over time, this manifests as higher infrastructure cost and slower convergence to optimal cluster size.
+
+## Risks and Mitigations
```
nathangeology, line 181: Need to add a migration story for folks that are using pod deletion cost as a way to steer the consolidation behavior (generally to steer consolidation away from higher value or difficult to disrupt pods)

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,405 @@
+# Pod Deletion Cost Controller
+
+## Summary
```
nathangeology, line 3: High level feedback I received is to talk about why Karpenter is the right place for this to live and how the calculation can evolve to cover more concerns (like factoring spreading or other placement constraints in). Also, should look for places to make this shorter without losing too much details for people to think through what the design is doing and why.

designs/pod-deletion-cost-controller.md
nathangeology: added a bit to the RFC addressing this, I do think we should integrate more with the consolidation logic over time, but I think focusing on candidate sorting is much simpler and gets us started.

designs/pod-deletion-cost-controller.md
nathangeology: Added a section on migration to address this one

designs/pod-deletion-cost-controller.md
nathangeology: Added the last pod deletion cost assigned annotation to deal with this one.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+This logic extends naturally to nodes that Karpenter cannot consolidate and to nodes that are already marked for replacement. The ranking engine partitions nodes into four groups before ranking:
+
+- **Group A (Disrupted + PDB-blocked):** Nodes that have the `karpenter.sh/disrupted` taint (disruption already committed) AND at least one pod blocked by a PDB with `DisruptionsAllowed=0`. These are the highest-value targets for RS scale-down assistance because: (1) the node is already committed to disruption, (2) PDB is blocking the drain, (3) getting replica pods off these nodes unblocks the PDB faster, (4) every pod removed directly accelerates an in-progress operation. Sorted by pod count ascending.
+- **Group B (Drifted):** Nodes with `ConditionTypeDrifted=True` and no do-not-disrupt pods. These nodes need replacement anyway for compliance or security reasons. Draining them first via RS scale-down means scale-down events naturally assist drift progress. Sorted by pod count ascending (fewest pods = lowest cost = drained first).
+- **Group C (Normal):** Nodes that are not drifted and have no do-not-disrupt pods. Standard consolidation targets. Sorted by pod count ascending.
```
DerekFrank: This should probably say:


```suggestion
- **Group C (Normal):** Nodes that are not drifted and have no do-not-disrupt pods. Standard consolidation targets. Sorted by candidate ranking.
```

This leaves it open to be however we rank candidates

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+- **Group D (Do-not-disrupt):** Nodes with at least one `karpenter.sh/do-not-disrupt` pod. Karpenter cannot act on these nodes regardless, and they are also protected from drift. Deleting pods from them has zero consolidation or drift value. Sorted by pod count ascending.
```
DerekFrank: That or the node is marked do-not-disrupt. Maybe also nodes that are in nodepools that don't have consolidation budget set to non-zero?

DerekFrank: There's gotta be some code that decomposes this into a 'can this node be consolidated ever'?

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Related Proposals that can reduce the pod disruption rate
+
+- Most important: Consolidation cost thresholding -- see draft RFC
+
+## Proposal
+
+We introduce a new feature-gated controller (`pod.deletioncost`) that automatically manages the `controller.kubernetes.io/pod-deletion-cost` annotation on pods running on Karpenter-managed nodes. The controller ranks nodes by pod count (mirroring Karpenter's own consolidation candidate sorting), assigns sequential deletion cost values to pods so that Kubernetes' ReplicaSet scale-down logic preferentially removes pods from the "least valuable" nodes first, and partitions nodes hosting do-not-disrupt pods separately to protect them from early eviction.
+
+### Ranking strategy
```
DerekFrank: Do you need this if you have 'why node level is the right signal?'

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+**With the controller:** The ReplicaSet controller sees all 3 pods on Node A have the lowest deletion cost (-3). It removes all 3 from Node A. Node A is now empty. Karpenter immediately consolidates it with zero disruption. The cluster goes from 3 nodes to 2.
+
+Additional examples (partial drain convergence, drifted node draining, disrupted + PDB-blocked priority) are in Appendix E.
+
+### Migration: separating consolidation steering from RS coordination
+
+Some Karpenter users currently set `controller.kubernetes.io/pod-deletion-cost` on their pods to influence which pods Karpenter prefers to evict during consolidation. With this controller also writing pod-deletion-cost for RS coordination, the same annotation now serves two purposes. To resolve this, we introduce `karpenter.sh/consolidation-priority` as the user-facing annotation for steering consolidation behavior.
+
+The consolidation scoring precedence becomes:
+1. `karpenter.sh/consolidation-priority` (if set by user)
```
DerekFrank: I think we want this to be


```suggestion
1. `karpenter.sh/disruption-cost` (if set by user)
```

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Changing the ReplicaSet Controller Behavior Directly
+
+A complementary approach is to modify the ReplicaSet controller's pod deletion algorithm itself via [KEP: ReplicaSet Consolidation-Aware Scale-In Strategy (kubernetes/enhancements#5982)](https://github.com/kubernetes/enhancements/issues/5982). This would add a `ConsolidatingScaleDown` feature gate to kube-controller-manager with zero API server overhead. The two approaches are complementary (pod-deletion-cost annotations take precedence in the RS sort order, so both can coexist). The recommended path is: ship this RFC now for immediate value, pursue the KEP in parallel, deprecate the annotation controller once the KEP reaches GA. See Appendix F for the full comparison.
+
+### Do nothing
```
DerekFrank: Remove this

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+- **Ranking oscillation causing excessive API server writes (Medium):** Unstable rankings could update annotations on many pods every 60-second cycle, creating sustained write load and watch event amplification. *Mitigation:* Three layers: (1) `ConsolidationState` timestamp comparison skips ranking when cluster state is unchanged (zero API writes and zero data collection cost in steady state), (2) `sort.SliceStable` with deterministic tie-breaking by node name prevents rank flipping, (3) 60-second reconcile interval bounds write frequency.
```
DerekFrank: We should take the 'how much we want to consolidate this node' functionality and share it between this and consolidation, make that a stable sort function. Then consolidation can build fairness on top but we have a single source of truth

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Placement and Spreading Constraints
+
+This proposal does not reason about topology spread constraints, pod affinity, or zone distribution. Neither does the current RS spreading heuristic it replaces. Both are simple heuristics optimizing for different goals (even distribution vs. enabling consolidation). The RS controller's current logic ignores all scheduling constraints; this proposal substitutes a different heuristic with the same level of placement awareness (none). Placement correctness is maintained by the scheduler on the create path. See Appendix G for the full analysis and comparison table.
+
+### Backward compatibility
+
+The feature gate defaults to `false`. When disabled, the controller is not registered, no annotations are written, no metrics are emitted, and all existing behavior is unchanged. When enabled, the controller only adds annotations. It never removes customer annotations, modifies pod specs, or changes node state.
+
+## Open Questions
```
nathangeology, line 189: Note to self Can probably re-work this section

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+- **Stale annotations after controller is disabled (Low):** Pods retain annotations indefinitely when the feature gate is turned off. *Mitigation:* Annotations live on pods, which are ephemeral. New pods start clean. Staleness is bounded by pod lifetime. Cleanup on disable is a known gap deferred to beta.
+
+- **Pod update RBAC escalation (Medium):** Karpenter's ClusterRole gains `update` and `patch` on pods, broader than the actual access pattern. *Mitigation:* The controller only processes pods on Karpenter-managed nodes via `node.Pods()` / `cluster.Nodes()`. The feature gate ensures the code path is dormant unless explicitly enabled. A future refinement could use server-side apply with a dedicated field manager.
```
DerekFrank: Should just cover this when we say that we're updating rbac

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+- **Consolidation-optimized deletions may conflict with topology spread constraints (Low):** When the controller concentrates deletions on specific nodes, the resulting pod distribution may temporarily violate `topologySpreadConstraints` until the scheduler places replacement pods. *Mitigation:* This is the same behavior as the current spreading heuristic. neither approach guarantees constraint satisfaction during scale-down. The Kubernetes scheduler enforces topology spread when placing new pods, so any temporary imbalance is corrected on the next scheduling cycle. Operators with strict spreading requirements can leave the feature disabled.
```
DerekFrank: I think we should reframe this to be a follow up. None of this design precludes supplementing the ranking to take topology into account later. 

We should decide if figuring that out is in graduation criteria

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+- **Active-scaling fan-out (Medium):** During active scaling events (e.g., HPA-driven burst), many nodes may change pod count simultaneously, causing the controller to re-rank and update annotations on a large fraction of pods in a single reconcile cycle. For a 500-node cluster averaging 30 pods/node, a full re-rank writes ~15,000 pod annotations in one 60-second window (~250 writes/sec). *Mitigation:* ConsolidationState-based change detection prevents re-ranking when state is stable. During active scaling, the writes are bounded by the reconcile interval (at most once per 60s). The API server write load is comparable to a large Deployment rollout. Clusters with >1,000 nodes should monitor `karpenter_pod_deletion_cost_annotation_duration_seconds` and consider disabling the feature if annotation latency exceeds acceptable thresholds.
```
nathangeology: Update based on caps/annotation limits discussed above

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+### How it works
+
+All new code lives in `pkg/controllers/pod/deletioncost/`. A singleton reconciler runs every 60 seconds. The 60-second interval balances annotation freshness against API server write load. fast enough to react to scale events within one HPA evaluation period, slow enough to avoid amplifying watch events during steady state. On each tick it:
+
+1. Checks the `PodDeletionCostManagement` feature gate
+2. Collects all Karpenter-managed nodes from `state.Cluster`
+3. Runs change detection (compares `ConsolidationState` timestamp from `state.Cluster`). If nothing changed, skips with zero API writes
+4. Partitions nodes into Group A (disrupted + PDB-blocked), Group B (drifted, no do-not-disrupt pods), Group C (not drifted, no do-not-disrupt pods), and Group D (has do-not-disrupt pods)
```
DerekFrank: Do we just want to set Group A to -MAX_INT and be done with it? I don't think Group A should count against our budget.

Maybe we do the inverse for group D which would be setting it to 0, so maybe just always making sure group D is cleared of annotations is the equivalent

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+5. Ranks each group independently by pod count ascending
+6. Assigns sequential integer ranks starting at -n (where n is the total number of Karpenter-managed nodes) for Group A, continuing for Group B, then Group C, then Group D
+7. Sets three annotations on each pod: `controller.kubernetes.io/pod-deletion-cost`, `karpenter.sh/managed-deletion-cost: "true"`, and `karpenter.sh/last-assigned-deletion-cost: "<value>"`. Skips pods with customer-set deletion costs. Detects third-party modifications by comparing current pod-deletion-cost against last-assigned (three-annotation protocol)
+
+```
+┌─────────────────────────────────────────────────────────────────┐
```
DerekFrank: I think you can remove the ascii art here :D

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+Some Karpenter users currently set `controller.kubernetes.io/pod-deletion-cost` on their pods to influence which pods Karpenter prefers to evict during consolidation. With this controller also writing pod-deletion-cost for RS coordination, the same annotation now serves two purposes. To resolve this, we introduce `karpenter.sh/consolidation-priority` as the user-facing annotation for steering consolidation behavior.
+
+The consolidation scoring precedence becomes:
+1. `karpenter.sh/consolidation-priority` (if set by user)
+2. `controller.kubernetes.io/pod-deletion-cost` (only if NOT auto-managed by this controller)
+3. Default
+
+When the deletion cost controller is enabled and manages a pod's deletion cost (sentinel annotation present), consolidation scoring ignores that pod's `pod-deletion-cost` since it reflects RS ranking, not user intent. Users who want to protect specific pods from consolidation should migrate to `karpenter.sh/consolidation-priority`.
+
+## Alternatives Considered
```
DerekFrank, line 161: Before alternatives, maybe a dedicated observability section?

designs/pod-deletion-cost-controller.md
DerekFrank: How do we feel about limiting the number of nodes we do this to for a given nodepool at its budget?

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+- **Annotation injection:** A malicious actor with pod write access could set `karpenter.sh/managed-deletion-cost: "true"` to trick Karpenter into managing their deletion cost. Low severity since the attacker already needs pod write access.
+
+### Performance
+
+- **API server write load:** Each reconcile can update up to N pods. With ConsolidationState-based change detection, writes only occur when cluster state changes. The change detection itself is O(1) with zero API calls. Worst case (1,000 nodes, 50 pods/node, constant changes): ~833 pod updates/sec. Mitigated by change detection and potential future annotation value diffing.
+- **Memory:** Negligible. References existing `state.StateNode` objects. Ranking data structures are O(n) and transient.
+- **CPU:** O(n log n) for sorting by pod count. Change detection is O(1) timestamp comparison.
+- **Watch event amplification:** Annotation updates trigger watch events for other controllers. Bounded by the 60-second reconcile interval. Annotation changes don't affect fields Karpenter's consolidation controller uses for decisions.
+
+## Appendix D: Metrics, Testing, and Rollout
```
nathangeology: Should be out of the appendix

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+### How it works
+
+All new code lives in `pkg/controllers/pod/deletioncost/`. A singleton reconciler runs every 60 seconds. The 60-second interval balances annotation freshness against API server write load. fast enough to react to scale events within one HPA evaluation period, slow enough to avoid amplifying watch events during steady state. On each tick it:
+
+1. Checks the `PodDeletionCostManagement` feature gate
+2. Collects all Karpenter-managed nodes from `state.Cluster`
+3. Runs change detection (compares `ConsolidationState` timestamp from `state.Cluster`). If nothing changed, skips with zero API writes
+4. Partitions nodes into Group A (disrupted + PDB-blocked), Group B (drifted, no do-not-disrupt pods), Group C (not drifted, no do-not-disrupt pods), and Group D (has do-not-disrupt pods)
+5. Ranks each group independently by pod count ascending
+6. Assigns sequential integer ranks starting at -n (where n is the total number of Karpenter-managed nodes) for Group A, continuing for Group B, then Group C, then Group D
```
DerekFrank: I think these groups should be nodepool scoped

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+7. Sets three annotations on each pod: `controller.kubernetes.io/pod-deletion-cost`, `karpenter.sh/managed-deletion-cost: "true"`, and `karpenter.sh/last-assigned-deletion-cost: "<value>"`. Skips pods with customer-set deletion costs. Detects third-party modifications by comparing current pod-deletion-cost against last-assigned (three-annotation protocol)
```
DerekFrank: This is maybe where we should put budgets on this, its an implementation detail

designs/pod-deletion-cost-controller.md
DerekFrank: There is the outstanding question of if you do it at the budget for drift, consolidation, or both.

DerekFrank: We can maybe put that as a beta criteria

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+- **Active-scaling fan-out (Medium):** During active scaling events (e.g., HPA-driven burst), many nodes may change pod count simultaneously, causing the controller to re-rank and update annotations on a large fraction of pods in a single reconcile cycle. For a 500-node cluster averaging 30 pods/node, a full re-rank writes ~15,000 pod annotations in one 60-second window (~250 writes/sec). *Mitigation:* ConsolidationState-based change detection prevents re-ranking when state is stable. During active scaling, the writes are bounded by the reconcile interval (at most once per 60s). The API server write load is comparable to a large Deployment rollout. Clusters with >1,000 nodes should monitor `karpenter_pod_deletion_cost_annotation_duration_seconds` and consider disabling the feature if annotation latency exceeds acceptable thresholds.
+
+### Placement and Spreading Constraints
```
DerekFrank: Again, lets move this to a graduation criteria follow up

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Ranking strategy
+
+The controller ranks nodes the same way Karpenter ranks consolidation candidates: by pod count (disruption cost). The ranking partitions nodes into four priority tiers before sorting. At the top priority: nodes already committed to disruption but blocked by PDBs, since every pod removed directly unblocks an in-progress operation. Next: drifted nodes that need replacement anyway. Then normal consolidation candidates. Finally, protected nodes. This maximizes the mutual information between the hint we send and the action Karpenter will take next, because the ranking signal and the consolidation signal are the same function. As Karpenter's consolidation candidate sorting evolves, we plan to follow it here to stay aligned.
+
+### How it works
+
+All new code lives in `pkg/controllers/pod/deletioncost/`. A singleton reconciler runs every 60 seconds. The 60-second interval balances annotation freshness against API server write load. fast enough to react to scale events within one HPA evaluation period, slow enough to avoid amplifying watch events during steady state. On each tick it:
```
DerekFrank: Is the HPA evaluation period configurable?

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+### Backward compatibility
+
+The feature gate defaults to `false`. When disabled, the controller is not registered, no annotations are written, no metrics are emitted, and all existing behavior is unchanged. When enabled, the controller only adds annotations. It never removes customer annotations, modifies pod specs, or changes node state.
+
+## Open Questions
+
+1. **Is Karpenter the right place for this?** [For SIG-Autoscaling / Karpenter maintainers] Would we be better off making this a standalone controller or trying to make changes directly to the ReplicaSet controller? The Alternatives section argues for co-location based on shared state and tight coupling to consolidation intent, but this deserves explicit discussion.
```
DerekFrank: You're arguing for yes in this doc, I'd cut this or keep it in the section where you argue for yes

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+2. **What do we think about the expanded RBAC permissions?** [For Karpenter maintainers / security reviewers] The PR adds `update` and `patch` on pods to Karpenter's ClusterRole. This is a meaningful privilege escalation. Should we explore server-side apply with a dedicated field manager to narrow the effective scope?
```
DerekFrank: We either need it or we don't. If Karpenter is the right place for it, then we need it. If we think Karpenter isn't the right place, then it isn't. FWIW, I'm convinced it is the right place

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+3. **How can we further minimize API server load from annotation updates?** [For Karpenter maintainers / SIG-Scalability] The current design relies on `ConsolidationState` timestamp comparison (O(1), zero API calls) and a 60-second reconcile interval. Additional ideas: diffing current vs. desired annotation values before writing, batching updates, or using server-side apply to reduce conflict retries.
```
DerekFrank: The question is really do we need to let users control how much load we send or is there a number that satisfies all usecases

designs/pod-deletion-cost-controller.md
nathangeology: (Most of these are covered now)

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+
+## Future Work
+
+Annotations are the v1 mechanism, not the end state. They have known limitations (last-writer-wins, no multi-contributor support, timing constraints). Future directions include: a dedicated `PodDisruptionPreference` API for multi-system input, scheduler library integration (merging scale-in signals inside the scheduler with full topology awareness), and deprecation of the annotation controller once a proper API exists. The controller is designed to be replaceable: it writes standard annotations that any future mechanism would also produce. See Appendix H for details.
+
+## Appendix A: Detailed Component Descriptions
```
DerekFrank: Lets save this for implementation, I'd rather this talk about how the system is supposed to work instead of how we structure the code

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,417 @@
...
+  Node D (4 pods) → rank -1 (always above all Group B and C nodes)
+```
+
+The ReplicaSet controller removes pods from drifted Node G first (rank -4), then normal Node E (rank -3). Node D's ML training job is never touched. Drifted Node G drains before non-drifted Node E despite Node E having fewer pods, because drift priority takes precedence over pod count across groups.
+
+**Without the controller:** The ReplicaSet controller might delete pods from Node D, but Karpenter can never consolidate it. Those deletions are wasted.
+
+**With the controller:** Pods on drifted Node G have rank -4 (lowest). The ReplicaSet controller removes pods from Node G first, assisting drift progress. Karpenter consolidates Node G while the ML training job on Node D is never touched.
+
+## Appendix C: Security and Performance Implications
```
DerekFrank: Probably keep these in the relevant sections

designs/pod-deletion-cost-controller.md
nathangeology: Per our discussion will key off of disruption budgets + have a hard cap on nodes for Alpha. Can revisit this for beta based on feedback.

designs/pod-deletion-cost-controller.md
nathangeology: I reworked this open questions sections to better reflect the current questions.

designs/pod-deletion-cost-controller.md
nathangeology: done

designs/pod-deletion-cost-controller.md
nathangeology: done

designs/pod-deletion-cost-controller.md
nathangeology: updated

designs/pod-deletion-cost-controller.md
nathangeology: Tried to address this, there is more detail in the 'How this works section'

designs/pod-deletion-cost-controller.md
nathangeology: done

designs/pod-deletion-cost-controller.md
nathangeology: Yes

nathangeology: 5 min between actions, 15 sec polling interval is default

designs/pod-deletion-cost-controller.md
nathangeology: Removed, will let the big section above cover this area

designs/pod-deletion-cost-controller.md
nathangeology: Gonna remove this section and let the migration section cover this

designs/pod-deletion-cost-controller.md
nathangeology: Updated

designs/pod-deletion-cost-controller.md
nathangeology: Just deleted this for now, we can revisit this when we get further along with the Scheduler-Aware autoscaling work upstream

GnatorX:
Overall looks good to me! I think we might learn a few more things(?) when you start testing this and running in it in cluster. I feel like when it comes to pod deletion cost, it is difficult to figure out what happens during design vs when running.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
...nd remove. These two controllers share no information about each other's intent. Without coordination, the ReplicaSet controller uses its default pod selection heuristic during scale-in: prefer pending over ready, respect pod-deletion-cost, spread terminations across nodes, prefer newer pods, then break ties randomly. With no pod-deletion-cost set, the spreading heuristic dominates, and terminations distribute roughly evenly across nodes. This increases the entropy of the cluster: most nodes end up packed at roughly the same density, and often no single node moves meaningfully closer to empty than other nodes. The result is that Karpenter's consolidation controller finds the same unfavorable state after a pod replica scale-in that it found before, with all nodes still occupied, none empty, and utilization distribution roughly unchanged but slightly lower everywhere. This matters especially for “ConsolidateWhenEmpty” NodePools, where the consolidation policy requires a node to be completely free of pods before it can be removed. If the ReplicaSet controller never concentrates deletions on a single node, that condition is rarely met, and the cluster carries more nodes than necessary.
```
DerekFrank: A link to the replicaset controller documentation would be nice here

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+1. Karpenter ranks nodes by consolidation priority
+2. Pods on those nodes get the lowest deletion costs
+3. ReplicaSet scale-down removes pods from those nodes first
+4. Those nodes become empty or closer to empty
+5. Karpenter can consolidate them with less disruption (or they qualify for WhenEmpty consolidation)
+
+Today the ranking uses pod count as a proxy for disruption cost (matching Karpenter's current candidate sorting). This will evolve. The Balanced Consolidation proposal ([#2962](https://github.com/kubernetes-sigs/karpenter/pull/2962)) introduces cost/benefit scoring for consolidation candidates that weighs savings against disruption. As that lands, the ranking here will follow: nodes with the best consolidation score (highest savings relative to disruption) will get the lowest deletion costs. The key property is that the ranking always mirrors how Karpenter itself prioritizes consolidation, regardless of how that prioritization evolves.
```
DerekFrank: We should mention that this ranking is just the calculated 'disruption cost'. If no user overrides are specified, that is pod count, but for this RFC we don't actually care about how disruption cost is calculated, merely that we're making the intentional choice to use the same heuristic for this signal

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Goals
+
+- Measurably reduce voluntary pod disruption rate for clusters using Deployments/ReplicaSets, as tracked by the pod disruption metrics in [#2892](https://github.com/kubernetes-sigs/karpenter/pull/2892)
+- Maintain or improve total compute cost and utilization efficiency across all benchmarks
+- Minimal impact on API server load from annotation updates
+
+### Value by consolidation policy
+
+The benefit of this feature varies by consolidation policy. For `ConsolidateWhenEmpty` NodePools, the value is direct: concentrating pod deletions on a single node is the only way to create the fully-empty condition that triggers consolidation. Without this controller, scale-down events rarely produce empty nodes. For `ConsolidateWhenUnderutilized` NodePools, the value is indirect but meaningful: concentrating deletions reduces the number of pods Karpenter must evict when it consolidates, lowering the total disruption cost of each consolidation move even though the consolidation would eventually happen regardless.
```
DerekFrank: Its more accurate to say that the value for empty is improving your bin-packing. the value for underutilized is that it takes less disruption. Its not really about directness

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Proposal: How it works
+
+We introduce a new feature-gated controller (`pod.deletioncost`) that automatically manages the `controller.kubernetes.io/pod-deletion-cost` annotation on pods running on Karpenter-managed nodes. All new code lives in `pkg/controllers/pod/deletioncost/`. A singleton reconciler runs every 60 seconds. The 60-second interval balances annotation freshness against API server write load. fast enough to react to scale events within one HPA evaluation period, slow enough to avoid amplifying watch events during steady state. On each tick it:
```
DerekFrank: We can leave the 60s reconcile loop as an implementation detail and just cut it. The important thing is that it reconciles on an interval, not what interval

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+1. Checks the `PodDeletionCostManagement` feature gate
```
DerekFrank: nit: we can just say the reconciler will only run if the feature gate is enabled

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+2. Collects all Karpenter-managed nodes from `state.Cluster`
+3. Runs change detection (compares `ConsolidationState` timestamp from `state.Cluster`). If nothing changed, skips with zero API writes
```
DerekFrank: I do wonder how much of this calculation can be put into state.Cluster. I think that can come out of implementation, we can leave the design ambiguous. The important thing is the dependency exists

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+4. Partitions nodes into Group A (disrupted + PDB-blocked), Group B (drifted, no do-not-disrupt pods), Group C (not drifted, no do-not-disrupt pods), and Group D (do-not-disrupt pods, do-not-disrupt nodes, or NodePools with no consolidation budget)
```
DerekFrank: Draining, Drifted, Disruptable, and Not Disruptable?

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+5. Ranks each group independently by candidate ranking
```
DerekFrank: Not quite true, we don't rank Draining or Not Disruptable.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+7. For each NodePool, the number of nodes eligible for annotation is limited by that NodePool's disruption budget. Since drift and consolidation can have separate budgets per NodePool, the controller respects each independently: Group B (drifted) nodes are bounded by the drift disruption budget, while Group C (normal) nodes are bounded by the consolidation disruption budget. Only nodes Karpenter could actually act on get annotated. Across all NodePools, a hard cap of 50 nodes per cycle prevents excessive labeling when budgets are permissive. For each eligible node's pods, sets three annotations: `controller.kubernetes.io/pod-deletion-cost`, `karpenter.sh/managed-deletion-cost: "true"`, and `karpenter.sh/last-assigned-deletion-cost: "<value>"`. Skips pods with customer-set deletion costs. Detects third-party modifications by comparing current pod-deletion-cost against last-assigned (three-annotation protocol). Nodes that drop out of scope have their managed annotations cleaned up.
```
DerekFrank: I wonder if during implementation we can't refactor consolidation to use the same GetDisruptableNodesStored() function as this and have it precomputed in the cluster? Not really input for this design 🙃

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+1. Checks the `PodDeletionCostManagement` feature gate
+2. Collects all Karpenter-managed nodes from `state.Cluster`
+3. Runs change detection (compares `ConsolidationState` timestamp from `state.Cluster`). If nothing changed, skips with zero API writes
+4. Partitions nodes into Group A (disrupted + PDB-blocked), Group B (drifted, no do-not-disrupt pods), Group C (not drifted, no do-not-disrupt pods), and Group D (do-not-disrupt pods, do-not-disrupt nodes, or NodePools with no consolidation budget)
+5. Ranks each group independently by candidate ranking
+6. Assigns Group A nodes the minimum int32 value (-2147483648) rather than sequential ranking. Group A nodes do not count against any annotation budget. Group D nodes have their Karpenter-managed annotations cleared (removed), so the RS controller treats them with default priority. Groups B and C receive sequential integer ranks starting at -n (where n is the total number of rankable nodes), with Group B first, then Group C
```
DerekFrank: There's another option here thats a bit rough to think about, we could keep the numbers sparse so that we only need to change a couple nodes when new ones are added instead of most? Whoever implements this could play around with sparse numbering to see if there is a solution that makes sense

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+**With the controller:** The ReplicaSet controller sees all 3 pods on Node A have the lowest deletion cost (-3). It removes all 3 from Node A. Node A is now empty. Karpenter immediately consolidates it with zero disruption. The cluster goes from 3 nodes to 2.
+
+Additional examples (partial drain convergence, drifted node draining, disrupted + PDB-blocked priority) are in Appendix B.
+
+### Migration: separating consolidation steering from RS coordination
+
+Some Karpenter users currently set `controller.kubernetes.io/pod-deletion-cost` on their pods to influence which pods Karpenter prefers to evict during consolidation. With this controller also writing pod-deletion-cost for RS coordination, the same annotation now serves two purposes. To resolve this, we introduce `karpenter.sh/disruption-cost` as the user-facing annotation for steering consolidation behavior.
+
+The consolidation scoring precedence becomes:
+1. `karpenter.sh/disruption-cost` (if set by user)
```
DerekFrank: Introducing this annotation needs to be in the API changes, and also in the intro, we can't bury the lede on that

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+No CRD changes. The feature is purely controller-side, gated behind `PodDeletionCostManagement` and configured via CLI flags / environment variables. RBAC is extended to add `update` and `patch` verbs on pods. The controller only processes pods on Karpenter-managed nodes via `node.Pods()` / `cluster.Nodes()`, and the feature gate ensures the code path is dormant unless enabled. A future refinement could use server-side apply with a dedicated field manager to narrow the effective scope.
+
+### Configuration
+
+| Option | CLI Flag | Env Var | Default | Description |
+|--------|----------|---------|---------|-------------|
+| Feature gate | `--feature-gates=PodDeletionCostManagement=true` | `FEATURE_GATES` | `false` | Enables the controller |
+
+### Example: Scale-down concentrates deletions on the consolidation target
```
DerekFrank, line 112: I'd move this below the others

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+2. `controller.kubernetes.io/pod-deletion-cost` (only if NOT auto-managed by this controller)
+3. Default
+
+When the deletion cost controller is enabled and manages a pod's deletion cost (sentinel annotation present), consolidation scoring ignores that pod's `pod-deletion-cost` since it reflects RS ranking, not user intent. Users who want to protect specific pods from consolidation should migrate to `karpenter.sh/disruption-cost`.
+
+## Observability
+
+The controller emits Prometheus metrics for nodes ranked per cycle, pods updated (broken down by result: updated, skipped-customer, skipped-unchanged, error), ranking duration, annotation write duration, and skipped-no-changes count. Kubernetes events fire on ranking completion, annotation update failures, and when the feature gate is disabled. Operators can monitor controller health via the `karpenter_pod_deletion_cost_annotation_duration_seconds` histogram and the `karpenter_pod_deletion_cost_reconcile_skipped_total` counter (high skip rate indicates stable state; zero skips with high annotation write latency indicates pressure). Structured logs at `V(1)` cover reconcile decisions; `V(2)` covers per-node ranking detail.
```
DerekFrank: I think this just needs metrics, probably no logging? We can list the intended metrics out

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Rollout and Graduation
+
+- Alpha (current): gate off by default, no stability guarantees. Hard cap of 50 nodes.
+- Beta: gate off, API stable, add cleanup-on-disable, evaluate topology-aware ranking factors, adjust node cap based on feedback.
```
DerekFrank: Gate on by default, typically

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
...
+- GA: gate on by default.
+- Rollback is safe: disable the feature gate and existing annotations become stale, bounded by pod lifetime.
```
DerekFrank: I'm not sure that rollback is entirely safe here? We don't actually remove the annotations.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,271 @@
+# Pod Deletion Cost Controller
+
+## Summary
+
+This RFC proposes a new feature-gated controller for Karpenter that ranks nodes by consolidation preference and propagates that ranking to pods via the `controller.kubernetes.io/pod-deletion-cost` annotation which is used by the ReplicaSet controller to determine which pods to scale in first. By aligning the ReplicaSet controller's scale-down decisions with Karpenter's consolidation targets, we measurably reduce voluntary pod disruption rate. This is tracked and validated via the pod disruption metrics being added in [kubernetes-sigs/karpenter#2892](https://github.com/kubernetes-sigs/karpenter/pull/2892). The feature is off by default and configured entirely through feature gates and CLI flags, with no CRD changes required.
```
nathangeology: Need to mention the API change/label migration up front (don't bury the lede) for pod-deletion-cost.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+- Maintain or improve total compute cost and utilization efficiency across all benchmarks
+- Minimal impact on API server load from annotation updates
+
+### Value by consolidation policy
+
+For ConsolidateWhenEmpty NodePools, concentrating pod deletions on specific nodes improves bin-packing by creating fully-empty nodes that qualify for removal. Without this controller, scale-down events rarely produce empty nodes. For ConsolidateWhenUnderutilized NodePools, concentrating deletions means fewer pods need to be evicted when Karpenter consolidates, reducing the total disruption per consolidation move.
+
+## Proposal
+
+We introduce a new feature-gated controller that automatically manages the `controller.kubernetes.io/pod-deletion-cost` annotation on pods running on Karpenter-managed nodes. The controller ranks nodes using Karpenter's disruption cost heuristic, assigns deletion cost values to pods so that Kubernetes' ReplicaSet scale-down logic preferentially removes pods from the best consolidation targets first, and partitions nodes Karpenter cannot act on separately to protect them from early eviction.
```
DerekFrank:
```suggestion
We introduce a new feature-gated controller that automatically manages the `controller.kubernetes.io/pod-deletion-cost` annotation on pods running on Karpenter-managed nodes that are replicaset controlled. The controller ranks nodes using Karpenter's disruption cost heuristic, assigns deletion cost values to pods so that Kubernetes' ReplicaSet scale-down logic preferentially removes pods from the best consolidation targets first, and partitions nodes Karpenter cannot act on separately to protect them from early eviction.
```

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
...nd remove. These two controllers share no information about each other's intent. Without coordination, the ReplicaSet controller uses its default pod selection heuristic during scale-in: prefer pending over ready, respect pod-deletion-cost, spread terminations across nodes, prefer newer pods, then break ties randomly. With no pod-deletion-cost set, the spreading heuristic dominates, and terminations distribute roughly evenly across nodes. This increases the entropy of the cluster: most nodes end up packed at roughly the same density, and often no single node moves meaningfully closer to empty than other nodes. The result is that Karpenter's consolidation controller finds the same unfavorable state after a pod replica scale-in that it found before, with all nodes still occupied, none empty, and utilization distribution roughly unchanged but slightly lower everywhere. This matters especially for "ConsolidateWhenEmpty" NodePools, where the consolidation policy requires a node to be completely free of pods before it can be removed. If the ReplicaSet controller never concentrates deletions on a single node, that condition is rarely met, and the cluster carries more nodes than necessary.
```
DerekFrank: * A root cause, not the root cause

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Goals
+
+- Measurably reduce voluntary pod disruption rate for clusters using Deployments/ReplicaSets, as tracked by the pod disruption metrics in [#2892](https://github.com/kubernetes-sigs/karpenter/pull/2892)
+- Maintain or improve total compute cost and utilization efficiency across all benchmarks
+- Minimal impact on API server load from annotation updates
+
+### Value by consolidation policy
```
DerekFrank, line 63: Honestly we can probably move this to an appendix or cut it?

DerekFrank: Maybe its fine

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+
+## Proposal
+
+We introduce a new feature-gated controller that automatically manages the `controller.kubernetes.io/pod-deletion-cost` annotation on pods running on Karpenter-managed nodes. The controller ranks nodes using Karpenter's disruption cost heuristic, assigns deletion cost values to pods so that Kubernetes' ReplicaSet scale-down logic preferentially removes pods from the best consolidation targets first, and partitions nodes Karpenter cannot act on separately to protect them from early eviction.
+
+### How it works
+
+The controller reconciles on a periodic interval, gated behind the `PodDeletionCostManagement` feature gate (the reconciler does not run when the gate is disabled). Where the data for this controller lives is an implementation detail, but the controller could reuse the existing `state.Cluster` data structures and functions to perform its ranking. On each reconcile it:
```
DerekFrank:
```suggestion
The controller reconciles on a periodic interval, gated behind the `PodDeletionCostManagement` feature gate. On each reconcile it:
```

designs/pod-deletion-cost-controller.md
nathangeology: I think it adds a bit of value/context.

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+
+1. Collects all Karpenter-managed nodes
+2. Runs change detection (compares `ConsolidationState` timestamp). If nothing changed, skips with zero API writes
```
DerekFrank:
```suggestion
1. If ConsolidationState has changed, for nodes that are Disruptable, Drifted, or Draining:
```

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+3. Partitions nodes into Draining, Drifted, Disruptable, and Not Disruptable
+4. Ranks Drifted and Disruptable nodes by the current Karpenter consolidation candidate ranking function. Draining nodes get a fixed minimum value. Not Disruptable nodes are excluded from ranking.
```
DerekFrank:
```suggestion
4. Ranks Drifted and Disruptable nodes by the current Karpenter consolidation candidate ranking function. Draining nodes are marked as the best to disrupt. Not Disruptable nodes are excluded from ranking.
```

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+5. For each NodePool, the number of nodes eligible for annotation is limited by that NodePool's disruption budget. Since drift and consolidation can have separate budgets per NodePool, the controller respects each independently: Drifted nodes are bounded by the drift disruption budget, Disruptable nodes are bounded by the consolidation disruption budget. Only nodes Karpenter could actually act on get annotated. Across all NodePools, a hard cap of 50 nodes per cycle prevents excessive labeling when budgets are permissive.
```
DerekFrank:
```suggestion
5. For each NodePool, the number of nodes eligible for annotation is limited by that NodePool's disruption budget. Since drift and consolidation can have separate budgets per NodePool, the controller respects each independently: Drifted nodes are bounded by the drift disruption budget, Disruptable nodes are bounded by the consolidation disruption budget. Across all NodePools, a hard cap of 50 nodes per cycle prevents excessive labeling when budgets are permissive.
```

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Standalone controller outside Karpenter
+
+The ranking strategies that matter most require the same cluster state Karpenter already maintains in its state.Cluster informer cache. A standalone controller would duplicate this state and API server watch load. If the ranking logic evolves to directly consume Karpenter's consolidation scoring, that integration is trivial when they share a process. The operational cost is also higher: separate RBAC, Helm chart, release lifecycle, etc. For a feature tightly coupled to consolidation behavior, co-location is the right call, but the controller could certainly be kept independent and still work the same.
+
+The ranking function should be factored into a shared location that the consolidation controller can also consume, establishing a single source of truth for "how much do we want to consolidate this node." This avoids drift between the deletion cost controller's ranking and consolidation candidate selection.
+
+## Risks and Mitigations
```
DerekFrank, line 181: We should discuss offline, I'm not sure I understand whats going on here

designs/pod-deletion-cost-controller.md
nathangeology: Addressed in 3e4372e8. The deletion cost controller now explicitly depends on the consolidation controller's exported candidate sort rather than defining its own. New "Shared ranking function" subsection in the proposal points at `sortCandidates` in `pkg/controllers/disruption/consolidation.go` and cites #2962, which reshapes that sort to savings-ratio descending. The implementation will lift the comparator into an exported helper so the consolidation candidate-selection function and the deletion cost ranking are literally one piece of code. Future changes to candidate selection automatically flow through. Leaving this thread open for you to resolve.

designs/pod-deletion-cost-controller.md
nathangeology: Added a section on candidate sorting to clarify this

designs/pod-deletion-cost-controller.md
```diff
@@ -0,0 +1,305 @@
...
+
+The ranking uses Karpenter's calculated disruption cost for each node, the same heuristic used for consolidation candidate ranking. This RFC does not prescribe how disruption cost is calculated (today it is based on pod count/pod-deletion-cost, but the Balanced Consolidation proposal [#2962](https://github.com/kubernetes-sigs/karpenter/pull/2962) will introduce cost/benefit scoring). What matters is that we intentionally use the same node ranking heuristic for this signal as for consolidation decisions.
+
+This logic extends naturally to nodes that Karpenter cannot consolidate and to nodes that are already marked for replacement. The ranking engine partitions nodes into four categories:
+
+- **Draining:** Nodes that have the `karpenter.sh/disrupted` taint (disruption already committed) AND at least one pod blocked by a PDB with `DisruptionsAllowed=0`. These are the highest-value targets for RS scale-down assistance because the node is already committed to disruption and getting replica pods off unblocks the PDB faster. Draining nodes receive a fixed minimum int32 deletion cost. They are not sequentially ranked.
```
DerekFrank:
```suggestion
- **Draining:** Nodes that have the `karpenter.sh/disrupted` taint (disruption already committed) OR at least one pod blocked by a PDB with `DisruptionsAllowed=0`. These are the highest-value targets for RS scale-down assistance because the node is already committed to disruption and getting replica pods off unblocks the PDB faster. Draining nodes receive a fixed minimum int32 deletion cost. They are not sequentially ranked.
```

YurDuiachenko:
Hi @nathangeology, I’m drafting an AEP for [kubernetes/autoscaler#9399](https://github.com/kubernetes/autoscaler/issues/9399), which also proposes using `disruption-cost` annotation.

@towca suggested aligning it with the related Karpenter work instead of introducing a CA-only annotation.

Do you think a shared user-facing annotation like `node-autoscaling.kubernetes.io/disruption-cost` makes sense, with similar semantics to `karpenter.sh/disruption-cost`?

nathangeology:
> Hi @nathangeology, I’m drafting an AEP for [kubernetes/autoscaler#9399](https://github.com/kubernetes/autoscaler/issues/9399), which also proposes using `disruption-cost` annotation.
> 
> @towca suggested aligning it with the related Karpenter work instead of introducing a CA-only annotation.
> 
> Do you think a shared user-facing annotation like `node-autoscaling.kubernetes.io/disruption-cost` makes sense, with similar semantics to `karpenter.sh/disruption-cost`?

Yes, I think we definitely need to build a way to combine autoscaling preferences and signals in Karpenter (and perhaps later the autoscaling aware scheduler) from other sources. Your proposal plus Garvin's mention of factoring in scheduling concerns like TSCs all could fall under this sort of API. I'll check with Derek and Jason on what they think here, but I think this could probably use its own RFC and probably should be addressed before the beta stage of this more simple proposal.

YurDuiachenko:
> Your proposal plus Garvin's mention of factoring in scheduling concerns like TSCs all could fall under this sort of API. I'll check with Derek and Jason on what they think here, but I think this could probably use its own RFC

Hi @nathangeology, do you have any news on this?

I also finished the draft AEP [#9850](https://github.com/kubernetes/autoscaler/pull/9850), and now it is open. I’d really appreciate your thoughts if you have a chance to take a look.

Thank you!

DerekFrank:
Left my vote of support for the AEP, if we're going to introduce that annotation, lets not introduce a karpenter one just to deprecate it

szuecs:
I tried to read into all issues related to the problem of HPA scale down and so I also found this RFC PR.
I see that it does not want to target the HPA kind of issues I showed in https://github.com/kubernetes/kubernetes/issues/124149. 
Do I understand correctly the motivation and summary?
I built an internal controller to use pod-deletion-cost to fix the hpa scale down to make sure topologySpreadConstraints in our case by zone stays similar.
One of the more interesting issues is that if you get node termination from something else (Cloud or rolling updates or karpenter). 
I think what is missing here is that there is no topologySpreadConstraints awareness. Do I get this right?

PR #2942, opened 2026-04-02
docs: Balanced consolidation policy RFC

jamesmthompson:
## Summary

A new `consolidationPolicy: Balanced` that scores each consolidation move by comparing savings and disruption as fractions of NodePool totals. Moves where disruption outweighs savings are rejected.

- `consolidationPolicy` is an IntOrString field. `Balanced` maps to k=2. Integer values (1-3) pass k directly as an escape hatch.
- All policies expressed through the disruption cost model: WhenEmpty approves only when no pod contributes positive disruption cost, WhenEmptyOrUnderutilized is k=+inf
- Per-node disruption cost of 1.0 eliminates division-by-zero edge cases
- Score-based ranking replaces disruption-only ranking when budget limits move count
- Exhaustive verification across c7i/m7i/r7i confirms k=2 is the smallest value that makes within-family REPLACEs viable

## Related issues

aws#8868, aws#8536, aws#6642, aws#7146, kubernetes-sigs#2319, kubernetes-sigs#1019, kubernetes-sigs#735, kubernetes-sigs#1851, kubernetes-sigs#2705, kubernetes-sigs#2883, kubernetes-sigs#1440, kubernetes-sigs#1686, kubernetes-sigs#1430, aws#5218, aws#3577


designs/balanced-consolidation.md
```diff
@@ -0,0 +1,440 @@
...
+
+This filter applies to single-node consolidation. A group of individually-failing nodes could produce a passing batch if their combined savings outweigh combined disruption. The filter misses these opportunities. Evaluating all multi-node combinations is exponential, so the implementation takes single-node savings first and attempts multi-node moves only when single-node opportunities are exhausted.
+
+The NodePool totals only need to be sensible relative to each other. The implementation may cache totals or estimate them from a subset of nodes, as long as cost and disruption are estimated from the same sample.
+
+### Interaction with Existing Features
+
+Existing feasibility checks (disruption budgets, PDBs, `consolidateAfter`, `do-not-disrupt`) are unchanged. Scoring applies only to consolidation, not to spot interruptions, expiration, or drift.
+
+`consolidateAfter` determines candidacy. Young nodes are not candidates but contribute to the denominators, preventing the post-replacement cycle from being artificially aggressive.
```
jukie: The end result doesn't seem changed but to clarify, `consolidateAfter` measures the last pod event on a node so node age might be the wrong framing here.

As "moves" are triggered that will reset consolidateAfter on the node that pods have been consolidated into and the effective node pool size will then be reduced. This might not matter, and I actually think it might preferred for giving a "cooldown" effect, but could you confirm the cycling loop called out in a few of the issues will be addressed?

designs/balanced-consolidation.md
```diff
@@ -0,0 +1,440 @@
...
+
+Every REPLACE scores low (e.g., 0.2 for a one-tier downsize) and is rejected. DELETEs still work: a DELETE of a node whose pods fit elsewhere scores 1.0 in a uniform pool. The system consolidates by deleting nodes, not replacing them. Operators who want all feasible moves can use `WhenEmptyOrUnderutilized`.
+
+### Does the score account for kube-scheduler pod placement?
+
+No. The score assumes pods land on the intended destination. In practice, kube-scheduler may scatter pods across existing nodes instead of packing onto the replacement. The replacement ends up nearly empty and becomes a consolidation candidate itself.
+
+This exists in all consolidation modes. The cost threshold concentrates the remaining moves onto higher-impact candidates. The system self-corrects: a nearly-empty replacement scores as a trivial DELETE next cycle. Cascades terminate because each round has strictly fewer displaced nodes.
+
+Configuring kube-scheduler with `MostAllocated` scoring reduces divergence. The [Workload-Aware Scheduling proposal](https://docs.google.com/document/d/1mPYqS4cFmsHPaVQDKyCz7-TKyWNJGjTaZQD3Umkvmgk) (Kepka, Feb 2026) addresses this more directly.
```
jukie: Btw this doc isn't accessible. Presumably it's private or possibly a bad link?

jamesmthompson: let me ask around for an updated link that I can send you on Kubernetes Slack. I think there was a lot of chatter at KubeCon about the path forward, and I don't really need the link. That will come up in other ways.

designs/balanced-consolidation.md
jamesmthompson: You're right, consolidateAfter tracks last pod event, not node age. Will fix. 

On cycling, the scoring filter is what breaks the loop. A same-type REPLACE saving $0 scores 0, and a move with near-zero savings with any meaningful disruption scores below 1/k, and both get rejected. The consolidateAfter reset on destination nodes adds a natural cooldown between rounds, which as you note is a good property, but it's not the primary mechanism.

designs/balanced-consolidation.md
```diff
@@ -0,0 +1,442 @@
+# Balanced Consolidation: Scoring Moves by Savings and Disruption
```
ellistarn, line 1: I am super excited about this approach!

designs/balanced-consolidation.md
```diff
@@ -0,0 +1,442 @@
...
+
+```yaml
+apiVersion: karpenter.sh/v1
+kind: NodePool
+metadata:
+  name: default
+spec:
+  disruption:
+    consolidationPolicy: Balanced
+    consolidationThreshold: 2.0
```
ellistarn: Thoughts on exposing different Enum values that codify these numbers, rather than the number itself? Also -- note that json cannot support floats in a stable way.

designs/balanced-consolidation.md
```diff
@@ -0,0 +1,442 @@
...
+The score compares a move's savings and disruption as fractions of NodePool totals.
+
+#### Per-Pod Disruption Cost
+
+[`EvictionCost`](../pkg/utils/disruption/disruption.go) starts at 1.0 per pod and adds two terms:
+
+1. **Pod deletion cost** ([`controller.kubernetes.io/pod-deletion-cost`](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/#pod-deletion-cost)), divided by 2^27, range -16 to +16. Default 0. The ReplicaSet controller uses this for scale-down ordering; Karpenter reuses it as a disruption signal.
+2. **Pod priority**, divided by 2^25, range -64 to +30. Default 0. Higher-priority pods increase their node's disruption cost.
+
+With neither set, per-pod disruption cost is 1.0. `EvictionCost` clamps to [-10, 10]. The scoring path clamps negative values to 0 via `max(0, EvictionCost(pod))` in the per-node formula (see [NodePool Totals](#nodepool-totals)). Other consumers of `EvictionCost` (eviction ordering) still see negatives. Scoring range per pod: [0, 10].
```
ellistarn, line 94: How does a user communicate that disrupting a pod is free?

designs/balanced-consolidation.md
jamesmthompson: good call on the float. the formally motivated values are all integers. k=1 is break-even (deletes only, no replaces). k=2 is where within-family replaces become viable, 4-step max steps to stasis. k=3 adds 8 cross-family pairs with the same 4-step number of steps to stasis. at k=4 churn chains, jump to 9 steps and the formal analysis starts arguing against k=4 (or any higher value). so the natural set is {1, 2, 3} and I'll restrict the input type.

I thought about named presets but Karpenter doesn't have an ordinal enum pattern today, and picking names that age well is hard. "Conservative/Balanced/Aggressive" reuses "Balanced" which is already the policy name. I think the integer is simpler, but we can do whatever you and the rest of the community want to do here.

designs/balanced-consolidation.md
jamesmthompson: if users want their pods to have zero disruption cost, they can set pod-deletion-cost to a large negative value. that drives EvictionCost negative, and the disruption cost in this RFC clamps the result to 0, so the pod contributes nothing to the node's total disruption cost.

the node still has a disruption cost of 1. nodes have a disruption cost independent of their pods (cordoning, draining, API calls, replacement latency). we haven't modeled this precisely and cost=1 is a placeholder. I'll make that clearer in the design. for today it eliminates a divide-by-zero that comes up if node disruption cost is zero. it could be larger if we wanted, but we don't need that yet.

so a node where every pod has negative deletion cost scores the same as an empty node. cheap to disrupt, not free. if a user wants truly zero-friction disruption at a NodePool level, they want WhenEmptyOrUnderutilized (or k=+inf)

jamesmthompson:
squashed to fix EasyCLA (removed Co-Authored-By trailer that the bot couldn't resolve). no content changes beyond what was already pushed.

/easycla

designs/balanced-consolidation.md
ellistarn: From an API design perspective, I am not sure that we need both knobs.

Right now, WhenEmpty: k=INF and WhenEmptyOrUnderutilized, k=0. 

I see two approaches:
1. Expand the enum that aliases other K values
2. Expose a new `consolidationThreshold` that works when consolidationPolicy: `WhenEmptyOrUnderutilized` and simply changes the threshold. 


cc: @jmdeal, @DerekFrank  curious to your thoughts.

designs/balanced-consolidation.md
ellistarn: Got it -- so in our docs we would recommend setting it to -1 if you want to treat it as free.

designs/balanced-consolidation.md
jamesmthompson: Yeah, I think your idea is better. The doc as-written assumes (maybe defensively) that can't justify k=2 uniquely for customers. If that's right, then we need k=3 and k=4, and then we might as well just make this parameter adjust the behavior of WhenEmptyOrUnderutilized. If we can uniquely justify k=2, then we can have a new enum. 

I'm leaning towards making k a parameter of WhenEmptyOrUnderutilized based purely on this RFC. @jmdeal and @DerekFrank I'm happy to do whatever you two think is sensible, I'll try to catch up with you you today.

jamesmt-aws:
Updated the RFC based on a prototype implementation against the karpenter codebase (12 files, 15 unit tests, all passing). The implementation branch is in https://github.com/kubernetes-sigs/karpenter/pull/2962, would love feedback on that. 

Four substantive changes:

1. **Examples corrected.** All worked numbers now include the 1.0 per-node disruption cost base defined in the NodePool Totals section. The original examples showed `total disruption cost = 80` for 80 pods across 10 nodes when it should be `90` (80 pods + 10 node bases). Scores shift (3.33 becomes 2.81, 1.67 becomes 1.50, etc.) but all approve/reject decisions hold.

2. **`evicted_pods` clarified.** Reschedulable pods only. DaemonSet, mirror, and node-owned pods are excluded since they are not evicted during consolidation and do not represent rescheduling disruption.

3. **Event cardinality made explicit.** Single-node moves emit on Node + NodeClaim. Multi-node moves emit a single event on the NodePool. "Must not fan out per-candidate" is now stated directly because the prototype got this wrong on the first pass.

4. **Totals caching guidance.** Computing NodePool totals requires knowing each node's pods. At 10k nodes that's O(N) API calls per cycle. The RFC now says: use cached pod mappings, or approximate from candidates only. Candidates-only underestimates the denominator, making scores slightly more conservative, which is the safe direction.

Also: added a brief definition of cycling loops, reframed the churn chain analysis as a formula safety property (on-demand picks cheapest so chains are length 1 in practice; spot has the 15-instance guard), and did an editing pass.

designs/balanced-consolidation.md
```diff
@@ -0,0 +1,457 @@
...
+
+#### Near-Zero-Cost Nodes (ODCRs, Reserved Capacity)
+
+Karpenter prices ODCRs and reserved instances at on-demand price divided by a large constant, keeping the most expensive ODCR cheaper than the cheapest spot node (see the provider's pricing implementation for the exact divisor). Within an ODCR-only pool, the divisor cancels in the score (it divides both savings and total cost). Scores equal the on-demand price ratios, so ODCR pools consolidate normally. This is correct: replacing an expensive reservation with a cheaper one frees capacity. The divided values are small but well within float64 precision.
+
+When a positive-cost source node is consolidated and its pods land on an ODCR destination node, this is a DELETE from the source pool's perspective. The score reflects the source pool's cost structure. The destination node's near-zero cost does not affect the score.
+
+### Candidate Filtering
+
+Move generation is expensive (find a destination, compute replacement costs, verify scheduling). A node's best possible score is its delete ratio: a DELETE saving the full node cost with no replacement.
```
sumukha-radhakrishna, line 172: The replacement cost can diverge from actual reality, so we might end up getting a node without the expected savings?

jamesmt-aws: That's true. Is that a problem with the balanced consolidation proposal, or is it a problem that is fundamental to consolidation and cost savings in Karpenter generally? 

I'm happy to fix it if you think that's necessary, but that feels separable to me. What do you think? Are there any specific cases that we can talk through? This feels like a real problem, but the proposal here doesn't make that problem worse.

grosser:
FYI since I'm not sure that is addressed, but the main problems that we are trying to solve is:
- consolidate high waste nodes first [POC](https://github.com/kubernetes-sigs/karpenter/pull/2873)
- do not consolidate when not saving at least X% [POC](https://github.com/kubernetes-sigs/karpenter/pull/2977)
- some kind of "nodes killed per hour" limit

jamesmthompson:
> FYI since I'm not sure that is addressed, but the main problems that we are trying to solve is:
> 
> * consolidate high waste nodes first [POC](https://github.com/kubernetes-sigs/karpenter/pull/2873)
> * do not consolidate when not saving at least X% [POC](https://github.com/kubernetes-sigs/karpenter/pull/2977)
> * some kind of "nodes killed per hour" limit

My goal here is to solve your problems with fewer user-facing controls. Basic idea is that we prioritize consolidation actions that target consolidation "moves" that have two properties. First, we should get a cost savings that looks pretty good. Second, we shouldn't disrupt a ton of pods. There's a simple estimation technique that we use to normalize these two factors to NodePool totals (more details in the RFC), so that a node with a lot of waste and small pod disruption is top priority, high waste/high disruption and low waste/low disruption are a rung down in priority, and low waste/high disruption moves are avoided entirely. 

There's more detail in the RFC, and I'm happy to talk with you about particular cases on Kubernetes Slack, over a Zoom call, or over email. I think the one thing we're missing is the "node killed per-hour" limit, but I'm hoping that can also be an implicit effect of setting the disruption budget on your nodepool.

ellistarn:
x-posting from #2962 -- I should've posted on the RFC here: 

---

Discussed offline.

Let's model this with consolidationPolicy: WhenEmpty | Balanced | WhenEmptyOrUnderutilized, with the option to expand that policy to support additional values in the future, with an escape hatch to pass k directly into that field.

Ready to approve once the RFC reflects this. cc: @jonathan-innis @DerekFrank @jmdeal

grosser:
prioritizing the nodes with most waste sounds about right (need some formula for gpu waste > cpu waste > mem waste, I usually use factors of 1gpu=8cpu 1cpu=4GB mem)

the killed per hour is less important if all disruptions are "worth it" (saving at least x% for example)

jamesmthompson:
> Discussed offline.
> 
> Let's model this with consolidationPolicy: WhenEmpty | Balanced | WhenEmptyOrUnderutilized, with the option to expand that policy to support additional values in the future, with an escape hatch to pass k directly into that field.
> 
> Ready to approve once the RFC reflects this. cc: @jonathan-innis @DerekFrank @jmdeal

While working through some of the cases here, I thought about what it really means for this RFC to take into account the whole spectrum between WhenEmpty and WhenEmptyOrUnderutilized. There's a node disruption cost in the current RFC, defaulting to just 1 (the same as disrupting a single pod). That means WhenEmpty can be expressed as: approve only when move disruption cost equals the per-node disruption cost. No pod contributes positive disruption cost. WhenEmptyOrUnderutilized is k=+inf (any positive savings passes). The whole spectrum lives in one model.

This also surfaces a small behavioral improvement. Today's WhenEmpty checks for literally zero pods. Under the disruption cost framing, a node whose pods all have large negative pod-deletion-cost (disruption cost clamped to 0) also qualifies. Pods that declared themselves free to disrupt shouldn't block consolidation.

RFC is updated. Also making consolidationPolicy an IntOrString so integer values (1-3) pass k directly as the escape hatch.

jamesmthompson:
@grosser Yeah, I completely agree. That's a resource-weighted cost concern. The RFC currently uses Karpenter's pricing model (dollar cost per node) for the savings side, so we're comparing node costs before/after proposed consolidation moes. This already implicitly weights GPUs higher because GPU instances cost more. It doesn't have an explicit resource-type weighting for disruption cost though, but if you see a pressing need for that I'm happy to add it.

ellistarn (approved):
/lgtm

designs/balanced-consolidation.md
```diff
@@ -0,0 +1,445 @@
...
+```
+
+`consolidationPolicy` is an `IntOrString` field. All values are expressed through the disruption cost model:
+
+| Value | Behavior |
+|---|---|
+| `WhenEmpty` | Approve only when move disruption cost equals the per-node disruption cost (no pod contributes positive disruption cost) |
+| `1` | Scoring with break-even threshold (deletes only, no replaces in uniform pools) |
+| `Balanced` | Scoring with k=2 (within-family replaces viable) |
+| `3` | Scoring with k=3 (adds cross-family replace pairs) |
```
ellistarn, line 72: One note. We could call this Conservative, Balanced, and Aggressive if we are only supporting 3.

PR #2944, opened 2026-04-02
fix: correct locking in state cost

### LINKED ISSUE #2938: karpenter_nodepools_cost_total reports negative values for NodePools
### Description

**Observed Behavior**:
`karpenter_nodepools_cost_total` intermittently reports negative values for the gitlab-runners NodePool.

This is observed on a NodePool that regularly provisions nodes for jobs, then becomes empty and consolidates nodes away. After repeated start/stop cycles, the metric may briefly or persistently drop below 0.

Example query:

`karpenter_nodepools_cost_total{nodepool="gitlab-runners"}
`

This behavior is observed on the raw metric itself.

**Expected Behavior**:
karpenter_nodepools_cost_total should never be negative, unless explicitly configured via negative pricing adjustments.

Even though this metric is a gauge and may decrease when nodes are removed, the absolute value should remain greater than or equal to 0.

Create a NodePool that provisions ephemeral runner nodes and consolidates when empty:
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: gitlab-runners
spec:
  disruption:
    budgets:
      - nodes: 10%
    consolidateAfter: 1m
  limits:
    cpu: "1000"
    memory: 1000Gi
  template:
    metadata:
      labels:
        type: gitlab-runner
    spec:
      expireAfter: 24h
      nodeClassRef:
        group: karpenter.k8s.aws
        kind: EC2NodeClass
        name: runners
      requirements:
        - key: kubernetes.io/os
          operator: In
          values: ["linux"]
        - key: kubernetes.io/arch
          operator: In
          values: ["amd64"]
          operator: In
          values: ["c"]
          operator: Gt
          values: ["5"]
          operator: Gt
          values: ["7"]
        - key: karpenter.k8s.aws/instance-memory
          operator: Gt
          values: ["15000"]
          operator: In
          values: ["on-demand"]
      taints:
        - key: gitlab-runner
          effect: NoSchedule
```
Run workloads that cause this NodePool to scale up and then return to empty.
Let Karpenter consolidate nodes when empty.
Repeat multiple scale-up / scale-down cycles.
Observe:
`karpenter_nodepools_cost_total{nodepool="gitlab-runners"}`

After repeated churn, the metric may become negative.

**Versions**:
- Chart Version: karpenter-1.10.0
- Kubernetes Version (`kubectl version`): 1.34

ryan-mist:
Fixes #2938

**Description**
Originally, the path is 
```
1. Acquire Read Lock
2. Check nodeClaimMap -> if exists then exit
3. Drop Read Lock
4. Acquire Write lock and make modifications
```
* instead of using read locks, we now acquire a write lock from the state
* this fixes a race condition in cost controller where between steps 3 and 4 another thread can modify costs, resulting in incorrect costs

* make presubmit

### DIFF
--- a/pkg/state/cost/cost.go
+++ b/pkg/state/cost/cost.go
@@ -199,11 +199,9 @@ func (cc *ClusterCost) createNewNodePoolCost(npName string, instanceTypes []*clo
 // UpdateNodeClaim adds a NodeClaim to cost tracking. The NodeClaim must have
 // all required labels or it will be ignored and logged as an error.
 func (cc *ClusterCost) UpdateNodeClaim(ctx context.Context, nodeClaim *v1.NodeClaim) error {
-	cc.RLock()
-	_, exists := cc.nodeClaimMap[client.ObjectKeyFromObject(nodeClaim)]
-	cc.RUnlock()
-
-	if exists {
+	cc.Lock()
+	defer cc.Unlock()
+	if _, exists := cc.nodeClaimMap[client.ObjectKeyFromObject(nodeClaim)]; exists {
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/state/cost/cost.go
DerekFrank, line 244: Maybe we should just .Lock() at the top and call it a day?

ryan-mist: Probably the best idea just so we don't make any mistakes later on, I'll change it. From a performance standpoint the impact is minimal (or more likely nothing) but still pains me a little to write

PR #2946, opened 2026-04-03
fix: correct offering count tracking in cost controller re-add path

jamesmthompson:
## Summary

`internalAddOffering`'s retry path (when an offering key was deleted after count hit 0) never incremented `oc.Count`. The offering was re-added with `Count: 0`, so `UpdateOfferings` recomputed `0 * price = $0` instead of `1 * price`. Cost tracking diverged from reality after any offering type fully scaled down and then scaled back up in a surviving nodepool.

## Fix

Move `oc.Count += 1` out of the `else` branch so it runs unconditionally. Change fallback `Count` from 1 to 0 since the unconditional increment handles it. The remove path was audited — decrement is already unconditional, no symmetric bug.

## Before / after

**Before** (new test on unfixed code):
```
Expected: 4.5
Actual:   3.0
cost should still be 4.50 after UpdateOfferings recomputes from counts,
got 3 — indicates count was not incremented on re-add
```

The re-added spot offering has `Count: 0`, so `UpdateOfferings` computes `0 * 1.50 + 1 * 3.00 = 3.00` instead of `1 * 1.50 + 1 * 3.00 = 4.50`.

**After**: all tests pass. The unconditional increment ensures the count is correct regardless of code path.

## Test

New case in the `UpdateOfferings` context: cycles an offering through add → remove → re-add with a surviving nodepool, then triggers `UpdateOfferings` to expose the count corruption.

### DIFF
--- a/pkg/state/cost/cost.go
+++ b/pkg/state/cost/cost.go
@@ -305,12 +305,12 @@ func (cc *ClusterCost) internalAddOffering(ctx context.Context, npName string, o
 		}
 		oc, exists = cc.npCostMap[npName].offeringCounts[offeringKey]
 		if !exists {
-			oc = OfferingCount{Count: 1, Price: 0.0}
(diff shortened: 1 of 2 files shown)


DerekFrank (approved):
/lgtm

PR #2947, opened 2026-04-04
fix: Track pending nodeclaims in provisioning state

### LINKED ISSUE #2854: Nodeclaims are not counted against nodepool limits
### Description

**Observed Behaviour**:

If nodes are not in ready state, an infinite number of nodeclaims can be created.

**Expected Behaviour**:

Nodeclaims cannot be created if the sum of nodeclaim resources are greater than the nodepool limits.\*

\*within reason, as there are multiple async threads provisioning an extra "surge" of a few nodeclaims is expected.

1. Edit nodeconfig in such a manner that nodes cannot join the cluster. eg for AWS you can set the osFamily to AL2 and the AMI to AL2023 If I recall correctly.
2. Create a pod
3. Nodeclaim is created
4. Pod is deleted
5. Nodeclaim remains
6. GOTO 2.

In my case, we had defined CPU Limits on the nodepools. 

**Versions**:
- Chart Version: Karpenter 1.3

w1t:
## Summary

  Track newly created pending `NodeClaim`s in cluster state immediately after `Create`, before lifecycle/status updates populate the real provider ID and resolved capacity fields.

  This addresses `kubernetes-sigs/karpenter#2854`, where repeated scheduling rounds could create additional `NodeClaim`s while an earlier claim remained pending and outside the state used by provisioning.

  The change also adds a regression test covering repeated scheduling after an unlaunched `NodeClaim` remains.

  ## Problem

  The reported failure mode is:

  1. provisioning creates a `NodeClaim`
  2. the original pod goes away, but the `NodeClaim` remains pending
  3. the node never progresses far enough to be fully represented in cluster state
  4. a later scheduling round still behaves as if that pending claim does not exist
  5. Karpenter creates another `NodeClaim`

  That can repeat when nodes fail to register, allowing over-creation relative to the intended `NodePool` budget.

  ## Fix

  After `Create` persists the `NodeClaim`, update cluster state with a tracked copy when the claim does not yet have a `ProviderID`.

  That tracked copy:
  - uses the nodeclaim name as a temporary provider ID
  - sets `Status.Capacity` to the max capacity across `InstanceTypeOptions`
  - sets `Status.Allocatable` to the max allocatable across `InstanceTypeOptions`

  This makes the pending claim visible to the same state consumed by provisioning before lifecycle/status reconciliation catches up.

  Depending on the pod's constraints, duplicate creation is prevented in one of two ways:
  - if the later pod is compatible with the pending tracked node, the scheduler can place it there and no new `NodeClaim` is produced
  - if the later pod cannot use that tracked node, the pending claim is still represented in nodepool accounting, so create-time limit enforcement remains a backstop

  ## Regression test

  The new regression test covers the reported loop:

  1. create a CPU-limited `NodePool`
  2. schedule a pod that creates one `NodeClaim`
  3. delete the original pod while leaving the pending `NodeClaim`
  4. schedule a second pod
  5. assert that no second `NodeClaim` is created

  The important end-to-end behavior is that the first pending claim is already represented in cluster state, so the second scheduling round does not create another claim.

  ## Tradeoff

  This uses a pessimistic reservation model for pending claims by taking the max capacity/allocatable across the candidate instance types.

  That can temporarily over-reserve in tight-limit pools when the eventual launched instance would have been smaller, but it is the intended tradeoff for this fix: prefer bounding provisioning during the async gap over allowing
  repeated over-creation while pending claims are invisible.

  ## Issue

  Fixes kubernetes-sigs/karpenter#2854
  
  PR #2526 addressed node-count limit enforcement once a pending claim is represented in StateNode capacity (by adding Node as a static resource). This PR addresses the earlier gap for resource limits (CPU, memory) before a newly
   created pending NodeClaim has a real ProviderID and enters state in a form provisioning can reuse.

### DIFF
--- a/pkg/controllers/provisioning/nodeclaim_limits_regression_test.go
+++ b/pkg/controllers/provisioning/nodeclaim_limits_regression_test.go
@@ -1,0 +1,98 @@
+package provisioning_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/awslabs/operatorpkg/status"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	clock "k8s.io/utils/clock/testing"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider/fake"
+	"sigs.k8s.io/karpenter/pkg/controllers/provisioning"
+	"sigs.k8s.io/karpenter/pkg/controllers/state"
+	"sigs.k8s.io/karpenter/pkg/events"
+	"sigs.k8s.io/karpenter/pkg/operator/options"
+	"sigs.k8s.io/karpenter/pkg/test"
+)
+
+func TestPendingNodeClaimPreventsOverProvisioning(t *testing.T) {
+	t.Parallel()
+
+	ctx := options.ToContext(context.Background(), test.Options())
+	kubeClient := fakeclient.NewClientBuilder().
+		WithScheme(scheme.Scheme).
+		WithIndex(&corev1.Pod{}, "spec.nodeName", func(obj client.Object) []string {
+			return []string{obj.(*corev1.Pod).Spec.NodeName}
+		}).
+		Build()
+
+	cloudProvider := fake.NewCloudProvider()
+	fakeClock := clock.NewFakeClock(time.Now())
+	cluster := state.NewCluster(fakeClock, kubeClient, cloudProvider)
+	prov := provisioning.NewProvisioner(kubeClient, events.NewRecorder(&record.FakeRecorder{}), cloudProvider, cluster, fakeClock)
+
+	nodePool := test.NodePool(v1.NodePool{
+		Spec: v1.NodePoolSpec{
+			Limits: v1.Limits(corev1.ResourceList{
+				corev1.ResourceCPU: resource.MustParse("2"),
+			}),
+		},
+	})
+	nodePool.StatusConditions().SetTrue(status.ConditionReady)
+	if err := kubeClient.Create(ctx, nodePool); err != nil {
+		t.Fatalf("creating nodepool, %v", err)
+	}
+
+	pod1 := test.UnschedulablePod(test.PodOptions{
+		ResourceRequirements: corev1.ResourceRequirements{
+			Requests: corev1.ResourceList{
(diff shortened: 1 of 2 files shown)


jmdeal:
> Track newly created pending NodeClaims in cluster state immediately after Create, before lifecycle/status updates populate the real provider ID and resolved capacity fields.

Karpenter's provisioner already ensures that cluster state is synced before proceeding with provisioning ([ref](https://github.com/kubernetes-sigs/karpenter/blob/5e512024a6e93e3304f325f30a2f91d2accfe004/pkg/controllers/provisioning/provisioner.go#L129-L134)). This should ensure that this state can never occur. Can you provide an example that demonstrates how this is insufficient? If there is an issue that can result in this check being insufficient, we should address it at the root.

w1t:
@jmdeal 

Synced() only proves the informer/cache has caught up to the current API object stream. It does not prove that a newly created NodeClaim has completed the lifecycle path that populates status.providerID and resolved capacity.

Concrete case:

1. Provisioner calls p.cluster.Synced(ctx) and proceeds.
2. Provisioner creates a NodeClaim.
3. The NodeClaim now exists in the API server, but initially has empty status:

status.providerID == ""
resolved capacity/allocatable not populated yet

4. Before the launch/lifecycle controller updates status, another provisioning starts.
5. That can be “synced” in the informer sense: it has observed the new NodeClaim object.
6. But the observed NodeClaim is still status-empty, so if cluster accounting ignores providerID-less / capacity-less NodeClaims, the in-flight capacity contributes zero.

So the insufficiency is not that the cache is stale. The cache may be perfectly accurate. The problem is that the accurate object state is still pre-launch / pre-status.

That is why tracking the pending NodeClaim immediately after Create is addressing the root accounting gap: the provisioner knows it just created intended capacity, even though the lifecycle controller has not yet resolved the real provider ID and capacity.


I'm also working on a more robust fix for the same root cause.

jmdeal:
> That can be “synced” in the informer sense: it has observed the new NodeClaim object.

Synced captures more than just if every NodeClaim has been added to the informer cache. It has explicit checks to ensure the NodeClaim has had its provider ID populated.

https://github.com/kubernetes-sigs/karpenter/blob/eb62f771941a6768733b562b60a51594e7f78920/pkg/controllers/state/cluster.go#L171-L177

By injecting a fake provider ID in provisioning, you actually bypass this protection measure. Long-term we may want to consider removing cluster sync and simulating in-flight NodeClaims as a superposition of instance types until we populate the launch details. However, this has limited benefit (launch detail population is fast) and we've taken a number of subtle dependencies on the current behavior. When we've looked into it before, we've decided the effort isn't worth the limited benefit.

I'm going to close out this PR since we don't have a reproduction for the race it's attempting to fix, and the proposed fix essentially disables Karpenter's cluster sync mechanism (breaking anything that has dependencies on it). If you want to continue working on this issue, I'd recommend to start by putting together reproduction steps that can be executed against a real Karpenter installation. I'd only move on to cutting a PR once we've reviewed the reproduction steps and positively identified the race condition.

PR #2949, opened 2026-04-07
test: prevent logging panics from failing tests

theunrepentantgeek:
**Description**

Depending on environmental timing issues, it's sometimes possible for goroutines launched as a part of a test to still be running (at least momentarily) after the test completes. 

If one of those goroutines tries to log, the resulting panic can cause the entire test suite to fail.

This PR adds a safety wrapper around the logger to capture and suppress those panics, eliminating this cause of test flakes.

Full unit test coverage of `safeTestingT`, plus `make test` locally (but noting that test flakes were normally observed during CI).


DerekFrank:
Looks like this broke all of the tests. Does this belong in Karpenter or should this be something we cut an issue for upstream in go or the test frameworks?

I haven't seen this happen personally, I'd like to see some evidence if possible

DerekFrank (changes requested):
See above comment

theunrepentantgeek:
> Looks like this broke all of the tests. 

Fixed - the `test` target in `Makefile` passes a number of `ginkgo` specific command line flags which weren't being interpreted because my tests are pure Go style. I added `suite_test.go` to consume those - other packages with Go style tests do the same.

> Does this belong in Karpenter or should this be something we cut an issue for upstream in go or the test frameworks?

I created uber-go/zap [#1526](https://github.com/uber-go/zap/issues/1526) about two months ago, but it hasn't gained much traction (if any). I have the changes in a WIP branch and will create a PR, but I'm not confident it will be merged.

> I haven't seen this happen personally, I'd like to see some evidence if possible

Three examples from CI runs for `karpenter-provider-azure`: [[Link](https://github.com/Azure/karpenter-provider-azure/actions/runs/21382388476/job/61551704122#step:6:5946) | [Link](https://github.com/Azure/karpenter-provider-azure/actions/runs/21382388476/job/61551704120#step:6:5946) | [Link](https://github.com/Azure/karpenter-provider-azure/actions/runs/22084237393/job/63815373157#step:6:8634)]

We've suppressed a number of these test flakes by disabling certain log calls, effectively we're following this model:

``` go
if !testing {
    // log goes here
}
```

But this has issues of it's own.

theunrepentantgeek:
I've looked into these

image

and they appear to be simple flakes, both failing with this error:

>  [FAIL] Termination Reconciliation [It] should not finish draining until minDrainTime has passed

I expect they'll pass if retried.

jmdeal:
> it's sometimes possible for goroutines launched as a part of a test to still be running (at least momentarily) after the test completes

Do we have examples of where this occurs, and why? This strikes me as the fundamental issue. Generally, we're not spawning goroutines that aren't scoped to the unit under test, and if we are it seems like something our scaffolding for that integration test should handle like [this comment](https://github.com/uber-go/zap/issues/1526#issuecomment-4033221310) suggests on your upstream zap issue. It would be good to have some concrete examples of where these goroutine leaks are occurring. Apologies if it was apparent from the test runs you've linked, they've since expired.

theunrepentantgeek:
> Do we have examples of where this occurs, and why? 

I can offer some indirect pointers.

Some time ago I wrote **karpenter-provider-azure** [PR#1366](https://github.com/Azure/karpenter-provider-azure/pull/1366) to reduce test flakes, adding guard clauses like this to the provider:

``` go
// Only log if context is still active to avoid logging after test completes
if ctx.Err() == nil {
    // ... elided ...
}
```

There are (currently) five of these in the karpenter-provider-azure source, each representing a case where someone tracked down a flaky test and found this was the best way to resolve the flakiness.

Two race condition exist in the above. One from the runtime: IIRC there's a narrow window of time _after_ the test has finished (during which logging will panic) but _before_ the context is cancelled. There's also one between the `ctx.Err() == nil` test and when the logging framework tries to log against the original `*testing.T`.

> Generally, we're not spawning goroutines that aren't scoped to the unit under test

As I wrote on the zaptest PR you linked above:

> Part of our issue stems from the size of the "system under test" - when we're testing the whole thing, everything behaves. 
> But some of our tests are exercising system fragments (entire subsystems) and that's where things become complicated.

I know that significant effort was put into finding and fixing these issues before I joined the AKS Autoscale team. What we're seeing is a small number of remaining cases where flakiness remains despite prior effort.

> It would be good to have some concrete examples of where these goroutine leaks are occurring.

I believe most (if not all) of the remaining "leaks" are short-lived. That is, we do not have goroutines that are abandoned to run forever, rather we have goroutines that _do_ terminate, but not always _quickly enough_.

PR #2877, opened 2026-02-27
test: set node name in claim status during simulation

matthchr:
**Description**
This matches the behavior during the Registration controller.

**How was this change tested?**
`make presubmit`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2951, opened 2026-04-07
fix: use injected clock for minDrainTime check

jamesmthompson:
## Summary

The `minDrainTime` check in `awaitDrain` compared `c.clock.Since()` against the condition's `LastTransitionTime`, which is set by `operatorpkg` using `metav1.Now()` (real clock). When tests use a fake clock, the two clocks diverge, causing the `minDrainTime` check to produce incorrect results.

This is the same class of bug fixed in #2783 ("deflake termination tests by using injected clock"), reintroduced by #2709 ("add minDrainTime").

- Track drain start time using the injected clock via `sync.Map` on the controller
- Clean up tracked time in `removeFinalizer`

## Before (main, fails deterministically)

```
$ go test ./pkg/controllers/node/termination/ -ginkgo.focus="should not finish draining until minDrainTime has passed"

Ran 1 of 31 Specs in 8.980 seconds
FAIL! -- 0 Passed | 1 Failed | 0 Pending | 30 Skipped
--- FAIL: TestAPIs (8.98s)
FAIL
FAIL	sigs.k8s.io/karpenter/pkg/controllers/node/termination	9.980s
```

## After (this PR, 10/10 passes)

```
$ for i in $(seq 1 10); do \
    result=$(go test ./pkg/controllers/node/termination/ \
      -ginkgo.focus="should not finish draining until minDrainTime has passed" 2>&1 \
      | tail -2 | head -1); \
    echo "Run $i: $result"; \
  done

Run 1: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	8.956s
Run 2: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	9.913s
Run 3: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	9.156s
Run 4: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	9.025s
Run 5: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	8.841s
Run 6: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	10.111s
Run 7: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	8.967s
Run 8: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	9.503s
Run 9: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	9.271s
Run 10: ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	8.583s
```

## Test plan

- [x] "should not finish draining until minDrainTime has passed" passes 10/10 (was 0/10)
- [x] Full termination suite passes (31/31 specs)


jamesmthompson:
This also affects #2906, which hit the same `should not finish draining until minDrainTime has passed` failure on 1.29.x and 1.30.x. Recently merged PRs (#2874, #2877, #2946) passed clean, so the timing race doesn't trigger on every run, but it's deterministic locally.

DerekFrank:
Should operatorpkg use an injected clock?

jamesmthompson:
> Should operatorpkg use an injected clock?

I'm happy with that. Who do I bother about https://github.com/awslabs/operatorpkg/pull/204? If we can't get that upstreamed in a reasonable amount of time, I think my original fix is better than nothing? You can let me know.

DerekFrank:
Alright the change merged into operator pkg, we can bump the version and consume the functionality:

https://github.com/awslabs/operatorpkg/pull/204

pkg/controllers/node/termination/controller.go
```diff
@@ -203,6 +205,9 @@ func (c *Controller) awaitDrain(
 	if nodeClaim != nil && nodeClaim.StatusConditions().Get(v1.ConditionTypeDrained) == nil {
 		nodeClaim.StatusConditions().SetUnknownWithReason(v1.ConditionTypeDrained, "Draining", "Draining")
 	}
+	// Track drain start time using the injected clock. The condition's LastTransitionTime
```
DerekFrank: Can't we use LastTransitionTime if we use the operatorpkg functionality elsewhere?

pkg/apis/v1/nodeclaim_status.go
```diff
@@ -70,11 +71,24 @@ type NodeClaimStatus struct {
...
 func (in *NodeClaim) StatusConditions() status.ConditionSet {
+	return in.statusConditions()
+}
+
+// StatusConditionsWithClock returns the ConditionSet with an injected clock.
+// Use this in controllers that hold a clock (e.g. for testability) so that
+// LastTransitionTime is stamped with the injected clock instead of the
+// real wall clock. The variadic StatusConditions() above is required to
+// satisfy operatorpkg's status.Object interface, which is non-variadic.
+func (in *NodeClaim) StatusConditionsWithClock(clk clock.Clock) status.ConditionSet {
```
DerekFrank, line 82: Short term we should probably not have a new function, we should just change the StatusConditions() to require a clock. Long term we should investigate this new package: https://go.dev/blog/testing-time

DerekFrank (changes requested):

jamesmt-aws:
Makes sense. I have awslabs/operatorpkg#204 open which adds `WithClock` to `ConditionSet.For()`. Once that merges, the fix here becomes: bump operatorpkg, thread the clock through `StatusConditions()`, and drop the `sync.Map`. I'll update this PR after #204 lands.

On the `testing/synctest` package, agreed that's the right long-term direction. It's stdlib, no dependency, and eliminates the fake clock pattern entirely.

DerekFrank:
I opened a PR for the bump here: https://github.com/kubernetes-sigs/karpenter/pull/3009

DerekFrank:
Injected clocks merged here: https://github.com/kubernetes-sigs/karpenter/pull/3009

PR #2952, opened 2026-04-07
chore: Upgrade go-licenses to fix builds

### LINKED ISSUE #2950: Silent fails of the 'presubmit' build workflow
### Description

**Observed Behavior**:

The `presubmit` build workflow (defined in [`presubmit.yaml`](https://github.com/kubernetes-sigs/karpenter/blob/main/.github/workflows/presubmit.yaml) ) is failing without triggering a failed build status.

For example, look at [this set of workflow runs](https://github.com/kubernetes-sigs/karpenter/actions/runs/24030021591/job/70076633350) where it appears that 5/7 runs worked:

Image

However, all five of the "green" workflow runs are themselves failures, all having failed with this error:

```
# github.com/google/go-licenses
Error: ../../../go/pkg/mod/github.com/google/go-licenses@v1.6.0/save.go:125:9: cannot use func(src string) (bool, error) {…} (value of type func(src string) (bool, error)) as func(srcinfo os.FileInfo, src string, dest string) (bool, error) value in struct literal
```

**Expected Behavior**:

Compilation failures during CI workflows should result in `Failed` workflows.

Go to the list of [presubmit workflow runs](https://github.com/kubernetes-sigs/karpenter/actions/workflows/presubmit.yaml) and inspect any recent run, even those showing "Passed".

theunrepentantgeek:
Fixes #2950 

**Description**

The existing version of `go-licenses` doesn't build with Go >= 1.24 (see go-licenses [#312](https://github.com/google/go-licenses/issues/312)), resulting in the license check not executing.

This wasn't detected in the Makefile because the `!` prefix on the line ended up ignoring the build failure, when the intention was to fail if unexpected licenses were found.

Two fixes required

* Update `go-licenses` to v2 
* Split the Makefile line into two parts so that any future compilation issue fails the build

Local testing with `make licenses`, followed by `make test` to catch any integration issues.


go.tools.mod
```diff
@@ -3,11 +3,11 @@ module tools
 // Tool dependencies isolated from main module to avoid conflicts.
 // Usage: go tool -modfile=go.tools.mod <toolname>
 
-go 1.25.5
+go 1.26.1
```
theunrepentantgeek, line 6: This now matches the version of Go in the main `go.mod` file.

go.tools.mod
theunrepentantgeek, line 1: Almost all of the changes to this file came from running `go mod tidy` across it _before_ upgrading `go-licenses`.

tallaxes: `go mod tidy` on `go.tools.mod`, in addition to tools dependencies, pulls in all of the current module packages and their dependencies - not pretty, as it makes it appear tools depend on things they don't. Not sure if this is a big deal, or if there is a way to avoid this - short of isolating tools into its own module.

theunrepentantgeek: Good point. Running `go mod tidy` appeared to be a necessary prerequisite for the new version of `go-license` - I got errors trying to pull it in until I reset `go-tools.*` and started with a tidy. I'll look into paring these down so the delta in this PR is smaller.

theunrepentantgeek: I temporarily isolated `go.tools.*` in a separate directory, pruned out everything bar the tool references, and reran `go mod tidy`. A few dependency versions were updated, but the list of dependencies was unchanged.

theunrepentantgeek:
This indirect dependency update seems to have resulted in some generated files receiving a modified annotation. The only change is the version mentioned in the annotation, so I've included them in this PR.

```
sigs.k8s.io/controller-tools v0.20.1 // indirect
```

Makefile
```diff
@@ -152,7 +152,8 @@ vulncheck: ## Verify code vulnerabilities
 	@go tool -modfile=go.tools.mod govulncheck ./pkg/...
 
 licenses: download ## Verifies dependency licenses
-	! go tool -modfile=go.tools.mod go-licenses csv ./... | grep -v -e 'MIT' -e 'Apache-2.0' -e 'BSD-3-Clause' -e 'BSD-2-Clause' -e 'ISC' -e 'MPL-2.0'
+	go tool -modfile=go.tools.mod go-licenses csv ./... > /tmp/licenses.csv
+	! grep /tmp/licenses.csv -v -e 'MIT' -e 'Apache-2.0' -e 'BSD-3-Clause' -e 'BSD-2-Clause' -e 'ISC' -e 'MPL-2.0'
```
jmdeal: Is there any reason we can't continue to pipe directly to grep? The previous approach would still work if this target was called multiple times simultaneously, this modification will break down since all will access the same file. If the file based approach is necessary, we should use `mktemp`

theunrepentantgeek: Direct piping is what lead to the original bug where `go-licenses` didn't compile and nobody noticed.

As I read it, the leading `!` was _intended_ to only invert the result of the `grep`, failing the build if an unexpected license was found. But, it also inverted the `go tool` call, treating any build failure as a pass. 

I split the two steps to avoid this happening again.

Using `mktemp` is a good idea, I'll do that.

theunrepentantgeek: @jmdeal I've updated the PR as requested, is there anything else you want/need from me?

PR #2953, opened 2026-04-08
fix: multi node consolidation: do not block all consolidation because some incompatible nodes are deleting or some incompatible pods are pending

grosser:
**Description**
given 2 nodepools with incompatible taints
- when a pod for 1 pool is pending consolidation for the other is blocked since the pod can never fit on any replacement nodes
- when a pod for 1 pool is deleting consolidation for the other is blocked since the pod can never fit on any replacement nodes

before this I can see that replacement candidate evaluation always fail in addToInflightNode because the picked type (A) does not match the pending pod (B) and then karpenter proposes to add a new replacement node, which leads to 2->2 replacement which is invalid

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -3333,6 +3333,172 @@ var _ = Describe("Consolidation", func() {
 			// and delete the old one
 			ExpectNotFound(ctx, env.Client, nodeClaims[1], nodes[1])
 		})
+		It("can delete nodes when a pending pod from another NodePool with incompatible taints exists", func() {
+			// NodePool A has taint nodepool=a:NoSchedule, NodePool B has taint nodepool=b:NoSchedule.
+			// A pending pod tolerates only nodepool=a — it should not block consolidation of NodePool B's node.
+			nodepoolA := test.NodePool(v1.NodePool{
+				Spec: v1.NodePoolSpec{
+					Template: v1.NodeClaimTemplate{
+						Spec: v1.NodeClaimTemplateSpec{
+							Taints: []corev1.Taint{{Key: "nodepool", Value: "a", Effect: corev1.TaintEffectNoSchedule}},
+						},
+					},
+					Disruption: v1.Disruption{
+						ConsolidationPolicy: v1.ConsolidationPolicyWhenEmptyOrUnderutilized,
+						Budgets:             []v1.Budget{{Nodes: "100%"}},
+						ConsolidateAfter:    v1.MustParseNillableDuration("0s"),
+					},
+				},
+			})
+			nodepoolB := test.NodePool(v1.NodePool{
+				Spec: v1.NodePoolSpec{
+					Template: v1.NodeClaimTemplate{
+						Spec: v1.NodeClaimTemplateSpec{
+							Taints: []corev1.Taint{{Key: "nodepool", Value: "b", Effect: corev1.TaintEffectNoSchedule}},
+						},
+					},
+					Disruption: v1.Disruption{
+						ConsolidationPolicy: v1.ConsolidationPolicyWhenEmptyOrUnderutilized,
+						Budgets:             []v1.Budget{{Nodes: "100%"}},
+						ConsolidateAfter:    v1.MustParseNillableDuration("0s"),
+					},
+				},
+			})
+			nodeclaimB, nodeB := test.NodeClaimAndNode(v1.NodeClaim{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{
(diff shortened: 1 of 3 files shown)


praveen9354 (commented):
@grosser thanks for tracking this down — the diagnosis matches the mental model of how multi-NodePool consolidation simulates placement, and the filter approach is the right shape. A few observations from a careful read:

**1. The renamed `PodCompatibleWithNode` deserves a clearer doc comment.** In its current form it checks taint toleration plus nodeSelector/affinity but not resource fit, volume topology, or PDB. That's fine for the consolidation-prefilter use case (the scheduler will reject incompatible placements downstream regardless), but the function's new public name suggests it's a general "can this pod fit on this node" oracle. Future readers may reach for it as such. Two concrete suggestions:
- Add a doc comment explicitly listing what the check covers (taints + selector/affinity) and what it doesn't (resources, volumes, PDB), so the prefilter semantics are visible at the call site.
- Consider keeping `isDaemonPodCompatibleWithNode` as a thin wrapper that delegates to `PodCompatibleWithNode`. The daemon-pod call site reads more clearly when the function name carries the intent, and you avoid changing the package's external surface for downstream consumers depending on `karpenter` as a library.

**2. Missing regression test for the cross-NodePool taint interaction.** Karpenter's disruption code is a frequent flake/regression magnet, and this is exactly the kind of cross-NodePool effect that's easy to silently reintroduce in a future refactor. A test in `pkg/controllers/disruption/suite_test.go` along the lines of:

- 2 NodePools with mutually-incompatible taints
- 1 pending pod targeting NodePool A (won't tolerate NodePool B's taint)
- 1 idle (consolidatable) candidate node from NodePool B
- Trigger consolidation reconcile
- Assert: NodePool B's candidate gets consolidated, and the pending pod stays pending without affecting the disruption decision

...would lock the behavior in place. Worth mirroring for the deleting-pod path (pod from a draining NodePool B node) since that's the second half of the description.

**3. Edge case worth surfacing in the doc comment.** With `lo.SomeBy(candidates, ...)`, a pod is included if it's compatible with ANY candidate. For multi-node consolidation where candidates span NodePools, the simulator picks a replacement based on its own logic — if the chosen replacement is incompatible with the pod (e.g., the replacement comes from a NodePool whose taints the pod doesn't tolerate), the simulation will still fail for that pod and block consolidation. That's pre-existing simulator behavior, not introduced here, but the helper's doc comment should note that the filter is "necessary but not sufficient" so the next person tracing a consolidation regression doesn't conclude the filter has them covered.

**4. The PR likely needs a rebase before CI can be re-evaluated.** It's 50 days old and the recent `golang.org/x/net@v0.55.0` bump (#3041) was needed to clear the `idna.ToASCII` `govulncheck` failure that's currently breaking all PRs against older main. Should be a clean rebase since `pkg/controllers/disruption/helpers.go` and `pkg/controllers/provisioning/scheduling/scheduler.go` haven't churned much in that window.

Solid fix — happy to take another look once the doc comment and test are added.

grosser:
thx, think I addressed all

praveen9354 (commented):
Thanks @grosser — all three points addressed cleanly:

1. **`PodCompatibleWithNode` doc comment** (`scheduler.go`): exactly the right framing — *"intentionally a prefilter ... does not check resource fit, volume topology, or PDB constraints — those are enforced by the scheduler downstream."* That tells the next reader to reach for the scheduler when they need a full admission check, which is what I was hoping for.

2. **`isDaemonPodCompatibleWithNode` wrapper preserved**: nicer than what I suggested — promoting it to a package-level function rather than keeping it as a `Scheduler` method drops the receiver and clarifies that it's a pure helper, while the daemon-pod call site still reads with its intent-carrying name. External-surface change is minimized to just exporting `PodCompatibleWithNode`.

3. **`filterPodsByCompatibleCandidates` doc comment** (`helpers.go`): the "necessary but not sufficient" framing is on point — the next person tracing a consolidation regression won't conclude the filter has them covered when the simulator picks an incompatible replacement.

4. **Regression test in `consolidation_test.go`** (`It("can delete nodes when a pending pod from another NodePool with incompatible taints exists")`): matches the shape I had in mind almost exactly — two NodePools with mutually-incompatible taints, a pending pod tolerating only A, asserts B's node consolidates. Solid.

One small thing — CI is failing on `go vet`:

```
vet: pkg/controllers/disruption/consolidation_test.go:3392:173: not enough arguments in call to ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated
```

Looks like that test helper's signature may have changed in upstream `main` since you branched. Should be a quick adjustment (probably one extra argument).

+1 from a non-maintainer reader — this looks ready for a maintainer's eyes once the `go vet` is cleared.

PR #2961, opened 2026-04-10
feat: disruption: make delay between disruptions configurable

grosser:
**Description**

we had too much node churn after making consolidation faster,
here is some simple way to push back against something like that

TODO: either make it an option or add env docs, lmk what you want

### DIFF
--- a/pkg/controllers/disruption/controller.go
+++ b/pkg/controllers/disruption/controller.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	stderrors "errors"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


jmdeal:
We don't expose this type of configuration due to the likelihood of the behavior of the field drifting as internal changes are made to Karpenter. Addressing node churn issues is something that we're invested in though. I'd recommend checking out the balanced consolidation RFC as a solution: https://github.com/kubernetes-sigs/karpenter/pull/2942. 

I'm going to close this PR since we wouldn't accept a change to expose this knob, but please leave feedback on that RFC. I think the implementation PR may be in good enough of a state to try out as well, but I'd check with @jamesmt-aws.

/close

PR #15, opened 2022-10-20
chore: migrate controllers, operator, etc

ellistarn:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2962, opened 2026-04-10
feat: implement Balanced consolidation policy

jamesmthompson:
## Summary

Implements the Balanced consolidation policy from #2942. A NodePool with `consolidationPolicy: Balanced` approves a consolidation move only when the savings it produces are at least half the disruption it costs, both measured as fractions of pool totals (k=2, score >= 0.5).

Scoring is a filter on top of the existing pipeline. Scheduling feasibility, price comparison, and budgets still gate moves as they always have; Balanced narrows the surviving set by rejecting marginal trades whose savings don't justify their disruption.

Default behavior is unchanged. Pools that don't opt in keep the prior semantics, expressed in the same k-value model: `WhenEmptyOrUnderutilized` is k=+inf (threshold 0, any positive savings), `Balanced` is k=2.

## Why this shape

- **API config, no feature gate.** Operators opt in per pool by setting `consolidationPolicy: Balanced`, the same surface as `WhenEmpty` and `WhenEmptyOrUnderutilized`. A cluster-wide gate would have forced two ways of saying the same thing.
- **k=2 fixed for now.** Tunable k ships separately once we have field data on whether the right discriminator is k or a different formulation. The internal model already carries k as a parameter, so the change is additive.
- **Emptiness stays a separate method.** Budgets filter by `DisruptionReason`. Routing empty nodes through consolidation would silently break `reasons: [Empty]` budgets. Emptiness owns empty candidates; consolidation rejects them.

## Test plan

- Unit tests over the RFC worked examples (Oversized Node 2.81, Marginal Move 0.42, scale invariance, heterogeneous disruption).
- Full-loop integration tests through `ComputeCommands -> EvaluateBalancedMove -> Validate -> execute` for single-node approve, single-node reject, multi-node approve, and Balanced under a 0% disruption budget.
- Adversarial tests for NaN offering prices, cross-pool conjunction (a low-score approving pool no longer masks a high-score rejecting one), and nil instance types.
- CEL validation tests for the `consolidationPolicy` enum.

```
ok  sigs.k8s.io/karpenter/pkg/controllers/disruption
ok  sigs.k8s.io/karpenter/pkg/apis/v1
```

## Related

- RFC: #2942

### DIFF
--- a/pkg/apis/v1/nodepool.go
+++ b/pkg/apis/v1/nodepool.go
@@ -80,6 +80,7 @@ type NodePoolSpec struct {
 	Replicas *int64 `json:"replicas,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!has(self.consolidationThreshold) || self.consolidationPolicy == 'Balanced'",message="consolidationThreshold is only valid when consolidationPolicy is Balanced"
 type Disruption struct {
 	//nolint:kubeapilinter
 	// ConsolidateAfter is the duration the controller will wait
@@ -96,9 +97,18 @@ type Disruption struct {
 	// algorithm. This policy defaults to "WhenEmptyOrUnderutilized" if not specified
 	// When replicas is set, ConsolidationPolicy is simply ignored
 	// +kubebuilder:default:="WhenEmptyOrUnderutilized"
-	// +kubebuilder:validation:Enum:={WhenEmpty,WhenEmptyOrUnderutilized}
+	// +kubebuilder:validation:Enum:={WhenEmpty,WhenEmptyOrUnderutilized,Balanced}
 	// +optional
 	ConsolidationPolicy ConsolidationPolicy `json:"consolidationPolicy,omitempty"`
+	//nolint:kubeapilinter
+	// ConsolidationThreshold controls how much disruption one unit of savings can buy
+	// when using the Balanced consolidation policy. A move is approved when
+	// score >= 1/consolidationThreshold. Higher values approve more aggressively.
+	// Only valid when consolidationPolicy is "Balanced". Defaults to 2.
+	// +kubebuilder:validation:Minimum:=1
+	// +kubebuilder:validation:Maximum:=3
+	// +optional
+	ConsolidationThreshold *int32 `json:"consolidationThreshold,omitempty"`
 	//nolint:kubeapilinter
 	// Budgets is a list of Budgets.
 	// If there are multiple active budgets, Karpenter uses
@@ -158,7 +168,11 @@ type ConsolidationPolicy string
 const (
 	ConsolidationPolicyWhenEmpty                ConsolidationPolicy = "WhenEmpty"
 	ConsolidationPolicyWhenEmptyOrUnderutilized ConsolidationPolicy = "WhenEmptyOrUnderutilized"
+	ConsolidationPolicyBalanced                 ConsolidationPolicy = "Balanced"
 )
+
+// DefaultConsolidationThreshold is the default value for ConsolidationThreshold when using the Balanced policy.
+const DefaultConsolidationThreshold int32 = 2
 
 // DisruptionReason defines valid reasons for disruption budgets.
 // +kubebuilder:validation:Enum={Underutilized,Empty,Drifted}
--- a/pkg/apis/v1/nodepool.go
+++ b/pkg/apis/v1/nodepool.go
@@ -93,12 +93,14 @@ type Disruption struct {
 	ConsolidateAfter NillableDuration `json:"consolidateAfter"`
 	//nolint:kubeapilinter
 	// ConsolidationPolicy describes which nodes Karpenter can disrupt through its consolidation
-	// algorithm. This policy defaults to "WhenEmptyOrUnderutilized" if not specified
-	// When replicas is set, ConsolidationPolicy is simply ignored
+	// algorithm. This policy defaults to "WhenEmptyOrUnderutilized" if not specified.
+	// Valid values: "WhenEmpty", "WhenEmptyOrUnderutilized", "Balanced".
+	// When replicas is set, ConsolidationPolicy is simply ignored.
 	// +kubebuilder:default:="WhenEmptyOrUnderutilized"
-	// +kubebuilder:validation:Enum:={WhenEmpty,WhenEmptyOrUnderutilized}
+	// +kubebuilder:validation:Enum:=WhenEmpty;WhenEmptyOrUnderutilized;Balanced
 	// +optional
 	ConsolidationPolicy ConsolidationPolicy `json:"consolidationPolicy,omitempty"`
+	//nolint:kubeapilinter
 	// Budgets is a list of Budgets.
 	// If there are multiple active budgets, Karpenter uses
 	// the most restrictive value. If left undefined,
@@ -159,7 +161,19 @@ type ConsolidationPolicy string
 const (
 	ConsolidationPolicyWhenEmpty                ConsolidationPolicy = "WhenEmpty"
 	ConsolidationPolicyWhenEmptyOrUnderutilized ConsolidationPolicy = "WhenEmptyOrUnderutilized"
+	ConsolidationPolicyBalanced                 ConsolidationPolicy = "Balanced"
 )
+
+// BalancedK is the scoring parameter for the Balanced policy. A move is
+// approved when score >= 1/k = 0.5. k=2 is the smallest value where
+// within-family replaces pass, with 4-step max churn. See
+// designs/balanced-consolidation.md "Why k=2".
+const BalancedK int32 = 2
+
+// IsBalanced returns true for the Balanced consolidation policy.
+func (p ConsolidationPolicy) IsBalanced() bool {
+	return p == ConsolidationPolicyBalanced
+}
 
 // DisruptionReason defines valid reasons for disruption budgets.
 // +kubebuilder:validation:Enum={Underutilized,Empty,Drifted}
--- a/pkg/apis/v1/nodepool_balanced_test.go
+++ b/pkg/apis/v1/nodepool_balanced_test.go
@@ -1,0 +1,191 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1
+
+import (
+	"context"
+	"strings"
+	"testing"
+)
+
+func int32Ptr(v int32) *int32 { return &v }
+
+// --- Test 2: Validation ---
+
+// TestValidateConsolidationThreshold_RejectsWithoutBalanced verifies that
+// RuntimeValidate returns an error when consolidationThreshold is set but
+// consolidationPolicy is not Balanced.
+func TestValidateConsolidationThreshold_RejectsWithoutBalanced(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyWhenEmptyOrUnderutilized
+	np.Spec.Disruption.ConsolidationThreshold = int32Ptr(2)
+
+	err := np.RuntimeValidate(context.Background())
+	if err == nil {
+		t.Fatal("expected error when consolidationThreshold is set with non-Balanced policy")
+	}
+	if !strings.Contains(err.Error(), "consolidationThreshold is only valid when consolidationPolicy is Balanced") {
+		t.Errorf("unexpected error message: %s", err.Error())
+	}
+}
+
+// TestValidateConsolidationThreshold_RejectsWithWhenEmpty verifies rejection
+// when policy is WhenEmpty.
+func TestValidateConsolidationThreshold_RejectsWithWhenEmpty(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyWhenEmpty
+	np.Spec.Disruption.ConsolidationThreshold = int32Ptr(2)
+
+	err := np.RuntimeValidate(context.Background())
+	if err == nil {
+		t.Fatal("expected error when consolidationThreshold is set with WhenEmpty policy")
+	}
+}
+
+// TestValidateConsolidationThreshold_PassesWithBalanced verifies that validation
+// passes when consolidationThreshold is set with consolidationPolicy: Balanced.
+func TestValidateConsolidationThreshold_PassesWithBalanced(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyBalanced
+	np.Spec.Disruption.ConsolidationThreshold = int32Ptr(2)
+
+	err := np.Spec.Disruption.validateConsolidationThreshold()
+	if err != nil {
+		t.Errorf("expected no error for Balanced policy with threshold, got: %s", err)
+	}
+}
+
+// TestValidateConsolidationThreshold_PassesWithNilThreshold verifies that
+// validation passes when consolidationThreshold is nil regardless of policy.
+func TestValidateConsolidationThreshold_PassesWithNilThreshold(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyWhenEmptyOrUnderutilized
+	np.Spec.Disruption.ConsolidationThreshold = nil
+
+	err := np.Spec.Disruption.validateConsolidationThreshold()
+	if err != nil {
+		t.Errorf("expected no error when threshold is nil, got: %s", err)
+	}
+}
+
+// TestValidateConsolidationThreshold_BoundaryValues tests the valid range [1, 3].
+// Note: The range [1, 3] is enforced by kubebuilder CEL validation at the API level.
+// The RuntimeValidate only checks the policy association. These tests verify the
+// policy check works for each valid value.
+func TestValidateConsolidationThreshold_BoundaryValues(t *testing.T) {
+	tests := []struct {
+		name      string
+		threshold int32
+		policy    ConsolidationPolicy
+		wantErr   bool
+	}{
+		{"threshold 1 with Balanced", 1, ConsolidationPolicyBalanced, false},
+		{"threshold 2 with Balanced", 2, ConsolidationPolicyBalanced, false},
+		{"threshold 3 with Balanced", 3, ConsolidationPolicyBalanced, false},
+		{"threshold 1 with WhenEmptyOrUnderutilized", 1, ConsolidationPolicyWhenEmptyOrUnderutilized, true},
+		{"threshold 3 with WhenEmptyOrUnderutilized", 3, ConsolidationPolicyWhenEmptyOrUnderutilized, true},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			d := &Disruption{
+				ConsolidationPolicy:    tc.policy,
+				ConsolidationThreshold: int32Ptr(tc.threshold),
+			}
+			err := d.validateConsolidationThreshold()
+			if tc.wantErr && err == nil {
+				t.Errorf("expected error, got nil")
+			}
+			if !tc.wantErr && err != nil {
+				t.Errorf("expected no error, got: %s", err)
+			}
+		})
+	}
+}
+
+// --- Test 3: Defaulting ---
+
+// TestSetDefaults_BalancedPolicyDefaultsThresholdTo2 verifies that SetDefaults
+// sets consolidationThreshold to 2 when consolidationPolicy is Balanced and
+// threshold is nil.
+func TestSetDefaults_BalancedPolicyDefaultsThresholdTo2(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyBalanced
+	np.Spec.Disruption.ConsolidationThreshold = nil
+
+	np.SetDefaults(context.Background())
+
+	if np.Spec.Disruption.ConsolidationThreshold == nil {
+		t.Fatal("expected consolidationThreshold to be set, got nil")
+	}
+	if *np.Spec.Disruption.ConsolidationThreshold != DefaultConsolidationThreshold {
+		t.Errorf("expected threshold %d, got %d", DefaultConsolidationThreshold, *np.Spec.Disruption.ConsolidationThreshold)
+	}
+	if *np.Spec.Disruption.ConsolidationThreshold != 2 {
+		t.Errorf("expected threshold 2, got %d", *np.Spec.Disruption.ConsolidationThreshold)
+	}
+}
+
+// TestSetDefaults_NonBalancedPolicyDoesNotSetThreshold verifies that SetDefaults
+// does NOT set consolidationThreshold when policy is not Balanced.
+func TestSetDefaults_NonBalancedPolicyDoesNotSetThreshold(t *testing.T) {
+	policies := []ConsolidationPolicy{
+		ConsolidationPolicyWhenEmpty,
+		ConsolidationPolicyWhenEmptyOrUnderutilized,
+	}
+
+	for _, policy := range policies {
+		t.Run(string(policy), func(t *testing.T) {
+			np := &NodePool{}
+			np.Spec.Disruption.ConsolidationPolicy = policy
+			np.Spec.Disruption.ConsolidationThreshold = nil
+
+			np.SetDefaults(context.Background())
+
+			if np.Spec.Disruption.ConsolidationThreshold != nil {
+				t.Errorf("expected threshold to remain nil for %s policy, got %d", policy, *np.Spec.Disruption.ConsolidationThreshold)
+			}
+		})
+	}
+}
+
+// TestSetDefaults_DoesNotOverrideExplicitThreshold verifies that SetDefaults
+// does NOT override an explicitly set threshold.
+func TestSetDefaults_DoesNotOverrideExplicitThreshold(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyBalanced
+	np.Spec.Disruption.ConsolidationThreshold = int32Ptr(3)
+
+	np.SetDefaults(context.Background())
+
+	if *np.Spec.Disruption.ConsolidationThreshold != 3 {
+		t.Errorf("expected threshold to remain 3, got %d", *np.Spec.Disruption.ConsolidationThreshold)
+	}
+}
+
+// TestSetDefaults_ExplicitThreshold1Preserved verifies threshold=1 is preserved.
+func TestSetDefaults_ExplicitThreshold1Preserved(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyBalanced
+	np.Spec.Disruption.ConsolidationThreshold = int32Ptr(1)
+
+	np.SetDefaults(context.Background())
+
+	if *np.Spec.Disruption.ConsolidationThreshold != 1 {
+		t.Errorf("expected threshold to remain 1, got %d", *np.Spec.Disruption.ConsolidationThreshold)
+	}
+}
--- a/pkg/apis/v1/nodepool_status.go
+++ b/pkg/apis/v1/nodepool_status.go
@@ -29,6 +29,9 @@ const (
 	ConditionTypeNodeClassReady = "NodeClassReady"
 	// ConditionTypeNodeRegistrationHealthy = "NodeRegistrationHealthy" condition indicates if a misconfiguration exists that is preventing successful node launch/registrations that requires manual investigation
 	ConditionTypeNodeRegistrationHealthy = "NodeRegistrationHealthy"
+	// ConditionTypeConsolidationPolicyUnsupported indicates that the NodePool's consolidationPolicy
+	// is set to Balanced but the BalancedConsolidation feature gate is disabled.
+	ConditionTypeConsolidationPolicyUnsupported = "ConsolidationPolicyUnsupported"
 )
 
 // NodePoolStatus defines the observed state of NodePool
--- a/pkg/controllers/disruption/balanced.go
+++ b/pkg/controllers/disruption/balanced.go
@@ -24,6 +24,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
+	"sigs.k8s.io/karpenter/pkg/events"
 	"sigs.k8s.io/karpenter/pkg/scheduling"
 	disruptionutils "sigs.k8s.io/karpenter/pkg/utils/disruption"
 )
@@ -240,6 +242,55 @@ func EvaluateBalancedMove(ctx context.Context, cmd Command, nodePoolTotals map[s
 	}
 
 	return worstResult
+}
+
+// EmitBalancedMultiNodeEvents emits events and metrics for a final multi-node
+// command, one per Balanced NodePool in the batch. This ensures cross-NodePool
+// batches get events on each participating pool, not just the first.
+func EmitBalancedMultiNodeEvents(ctx context.Context, cmd Command, nodePoolTotals map[string]NodePoolTotals, recorder events.Recorder) {
+	byPool := lo.GroupBy(cmd.Candidates, func(c *Candidate) string { return c.NodePool.Name })
+	savings := cmd.EstimatedSavings()
+
+	for poolName, poolCandidates := range byPool {
+		nodePool := poolCandidates[0].NodePool
+		if nodePool.Spec.Disruption.ConsolidationPolicy != v1.ConsolidationPolicyBalanced {
+			continue
+		}
+
+		consolidationThreshold := GetConsolidationThreshold(nodePool)
+		disruptionCost := ComputeMoveDisruptionCost(ctx, poolCandidates)
+		totals := nodePoolTotals[poolName]
+
+		poolCost := candidatesCost(poolCandidates)
+		totalCost := candidatesCost(cmd.Candidates)
+		poolSavings := savings
+		if totalCost > 0 && len(byPool) > 1 {
+			poolSavings = savings * (poolCost / totalCost)
+		}
+
+		result := ScoreMove(poolSavings, disruptionCost, totals, consolidationThreshold)
+
+		decisionLabel := "approved"
+		if !result.Approved {
+			decisionLabel = "rejected"
+		}
+		ConsolidationScoreHistogram.Observe(result.Score, map[string]string{"decision": decisionLabel, "nodepool": poolName})
+		ConsolidationMovesTotal.Inc(map[string]string{"decision": decisionLabel, "nodepool": poolName})
+
+		if result.Approved {
+			recorder.Publish(disruptionevents.BalancedConsolidationApprovedMultiNode(
+				nodePool,
+				result.Score, result.Threshold, result.ConsolidationThreshold,
+				result.SavingsFraction*100, result.DisruptionFraction*100,
+			))
+		} else {
+			recorder.Publish(disruptionevents.BalancedConsolidationRejectedMultiNode(
+				nodePool,
+				result.Score, result.Threshold, result.ConsolidationThreshold,
+				result.SavingsFraction*100, result.DisruptionFraction*100,
+			))
+		}
+	}
 }
 
 // candidatesCost returns the total price of a set of candidates.
--- a/pkg/controllers/disruption/balanced.go
+++ b/pkg/controllers/disruption/balanced.go
@@ -1,0 +1,252 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package disruption
+
+import (
+	"context"
+	"math"
+
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/scheduling"
+	disruptionutils "sigs.k8s.io/karpenter/pkg/utils/disruption"
+)
+
+// BalancedScoreResult holds the result of scoring a consolidation move.
+type BalancedScoreResult struct {
+	Score                  float64
+	SavingsFraction        float64
+	DisruptionFraction     float64
+	Approved               bool
+	Threshold              float64
+	ConsolidationThreshold int32
+}
+
+// NodePoolTotals holds the precomputed totals for a NodePool needed by the scoring function.
+type NodePoolTotals struct {
+	TotalCost           float64
+	TotalDisruptionCost float64
+}
+
+// computeNodePoolTotals computes NodePool totals from the full set of candidates
+// before any ShouldDisrupt filtering, so that balanced scoring normalizes against
+// the entire NodePool. Candidates with no resolvable price (nil instanceType or
+// no compatible offerings) are excluded from both cost and disruption totals to
+// avoid skewing the ratio.
+func computeNodePoolTotals(ctx context.Context, allCandidates []*Candidate) map[string]NodePoolTotals {
+	totalsMap := map[string]NodePoolTotals{}
+	for _, c := range allCandidates {
+		price := candidatePrice(c)
+		if price == 0 {
+			continue
+		}
+		name := c.NodePool.Name
+		totals := totalsMap[name]
+		totals.TotalCost += price
+		totals.TotalDisruptionCost += 1.0 // per-node base
+		for _, p := range c.reschedulablePods {
+			evictionCost := disruptionutils.EvictionCost(ctx, p)
+			totals.TotalDisruptionCost += math.Max(0, evictionCost)
+		}
+		totalsMap[name] = totals
+	}
+	return totalsMap
+}
+
+// candidatePrice returns the cheapest compatible offering price for a single
+// candidate. Returns 0 if the candidate has no instance type or no compatible
+// offerings (e.g., deprecated instance types -- drift should handle those).
+func candidatePrice(c *Candidate) float64 {
+	if c == nil || c.instanceType == nil {
+		return 0
+	}
+	reqs := scheduling.NewLabelRequirements(c.Labels())
+	offerings := c.instanceType.Offerings.Compatible(reqs)
+	if len(offerings) == 0 {
+		return 0
+	}
+	return offerings.Cheapest().Price
+}
+
+// ScoreMove scores a consolidation move using the balanced scoring formula.
+//
+//	savings_fraction = savings / nodepool_total_cost
+//	disruption_fraction = disruption_cost / nodepool_total_disruption_cost
+//	score = savings_fraction / disruption_fraction
+//
+// A move is approved when score >= 1/consolidationThreshold.
+func ScoreMove(savings float64, disruptionCost float64, totals NodePoolTotals, consolidationThreshold int32) BalancedScoreResult {
+	// Zero nodepool cost: no consolidation possible
+	if totals.TotalCost <= 0 {
+		return BalancedScoreResult{Score: 0, Approved: false}
+	}
+
+	savingsFraction := savings / totals.TotalCost
+
+	// Zero savings: never approved
+	if savings <= 0 {
+		return BalancedScoreResult{
+			Score:           0,
+			SavingsFraction: savingsFraction,
+			Approved:        false,
+		}
+	}
+
+	// Zero total disruption cost: any move with positive savings is approved
+	if totals.TotalDisruptionCost <= 0 {
+		return BalancedScoreResult{
+			Score:           math.Inf(1),
+			SavingsFraction: savingsFraction,
+			Approved:        savings > 0,
+		}
+	}
+
+	disruptionFraction := disruptionCost / totals.TotalDisruptionCost
+	score := savingsFraction / disruptionFraction
+	threshold := 1.0 / float64(consolidationThreshold)
+
+	return BalancedScoreResult{
+		Score:                  score,
+		SavingsFraction:        savingsFraction,
+		DisruptionFraction:     disruptionFraction,
+		Approved:               score >= threshold,
+		Threshold:              threshold,
+		ConsolidationThreshold: consolidationThreshold,
+	}
+}
+
+// ComputeMoveDisruptionCost computes the disruption cost for a consolidation move.
+// It adds a per-node base of 1.0 plus sum(max(0, EvictionCost(pod))) for all
+// reschedulable pods on the candidate nodes.
+//
+// This does not include LifetimeRemaining adjustment. Candidate ordering already
+// uses lifetime-adjusted DisruptionCost, so nodes near expiration sort first.
+// Scoring evaluates the static cost structure. The two compose: lifetime affects
+// which node is tried, scoring affects whether the move is worth it.
+//
+// To justify adding lifetime adjustment here, we would need evidence that Balanced
+// rejects moves on near-expiration nodes that should be approved. Analysis shows
+// the crossover is at lifetime_remaining=0.0045 for the marginal replace case
+// (29.9 days into a 30-day expireAfter), where expiration handles the node within
+// hours. At k=2, zero moves flip when lifetime >= 25%. See designs/balanced-consolidation.md
+// "Resolved Questions" section.
+func ComputeMoveDisruptionCost(ctx context.Context, candidates []*Candidate) float64 {
+	cost := float64(len(candidates)) // per-node base of 1.0
+	for _, c := range candidates {
+		for _, p := range c.reschedulablePods {
+			evictionCost := disruptionutils.EvictionCost(ctx, p)
+			cost += math.Max(0, evictionCost)
+		}
+	}
+	return cost
+}
+
+// GetConsolidationThreshold returns the consolidation threshold for a NodePool,
+// defaulting to DefaultConsolidationThreshold if not set.
+func GetConsolidationThreshold(nodePool *v1.NodePool) int32 {
+	if nodePool.Spec.Disruption.ConsolidationThreshold != nil {
+		return *nodePool.Spec.Disruption.ConsolidationThreshold
+	}
+	return v1.DefaultConsolidationThreshold
+}
+
+// AnyBalancedCandidate returns true if any candidate in the list uses the
+// Balanced consolidation policy. This handles cross-NodePool batches where
+// only some candidates may use Balanced.
+func AnyBalancedCandidate(candidates []*Candidate) bool {
+	return lo.SomeBy(candidates, func(c *Candidate) bool {
+		return c.NodePool.Spec.Disruption.ConsolidationPolicy == v1.ConsolidationPolicyBalanced
+	})
+}
+
+// EvaluateBalancedMove evaluates whether a consolidation command should be
+// approved under the Balanced policy. For cross-NodePool moves, each source
+// NodePool with Balanced policy is scored independently using its own totals
+// and tolerance. The move is approved only if ALL Balanced pools approve.
+func EvaluateBalancedMove(ctx context.Context, cmd Command, nodePoolTotals map[string]NodePoolTotals) BalancedScoreResult {
+	if len(cmd.Candidates) == 0 {
+		return BalancedScoreResult{Score: 0, Approved: false}
+	}
+
+	// Group candidates by NodePool
+	byPool := lo.GroupBy(cmd.Candidates, func(c *Candidate) string { return c.NodePool.Name })
+
+	// Total savings for the whole command
+	savings := cmd.EstimatedSavings()
+
+	// Score per Balanced NodePool. Each pool must independently approve.
+	var worstResult BalancedScoreResult
+	worstResult.Approved = true
+	worstResult.Score = math.Inf(1)
+
+	for poolName, poolCandidates := range byPool {
+		nodePool := poolCandidates[0].NodePool
+		if nodePool.Spec.Disruption.ConsolidationPolicy != v1.ConsolidationPolicyBalanced {
+			continue
+		}
+
+		consolidationThreshold := GetConsolidationThreshold(nodePool)
+		disruptionCost := ComputeMoveDisruptionCost(ctx, poolCandidates)
+		totals := nodePoolTotals[poolName]
+
+		// For cross-NodePool moves, attribute savings proportionally to each
+		// pool's share of the total source cost
+		poolCost := candidatesCost(poolCandidates)
+		totalCost := candidatesCost(cmd.Candidates)
+		poolSavings := savings
+		if totalCost > 0 && len(byPool) > 1 {
+			poolSavings = savings * (poolCost / totalCost)
+		}
+
+		result := ScoreMove(poolSavings, disruptionCost, totals, consolidationThreshold)
+
+		log.FromContext(ctx).V(1).Info("balanced consolidation score",
+			"nodepool", poolName,
+			"score", result.Score,
+			"savings_fraction", result.SavingsFraction,
+			"disruption_fraction", result.DisruptionFraction,
+			"savings", poolSavings,
+			"disruption_cost", disruptionCost,
+			"nodepool_total_cost", totals.TotalCost,
+			"nodepool_total_disruption_cost", totals.TotalDisruptionCost,
+			"threshold", 1.0/float64(consolidationThreshold),
+			"approved", result.Approved,
+			"decision", cmd.Decision(),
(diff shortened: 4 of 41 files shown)


jamesmt-aws:
Note: the RFC discussion (#2942) has an open question about whether this should be a new \`Balanced\` policy or a \`consolidationThreshold\` parameter on the existing \`WhenEmptyOrUnderutilized\`. The implementation supports either direction with minimal changes. If the decision is to go with a threshold on \`WhenEmptyOrUnderutilized\`, the scoring logic and tests are unchanged. The API surface changes are: remove the \`Balanced\` enum value, allow \`consolidationThreshold\` on \`WhenEmptyOrUnderutilized\`, and update \`ShouldDisrupt\` to check the threshold instead of the policy.

ellistarn:
Discussed offline.

Let's model this with `consolidationPolicy: WhenEmpty | Balanced | WhenEmptyOrUnderutilized`, with the option to expand that policy to support additional values in the future, with an escape hatch to pass k directly into that field. 

Ready to approve once the RFC reflects this. cc: @jonathan-innis @DerekFrank @jmdeal

jamesmthompson:
> Discussed offline.
> 
> Let's model this with `consolidationPolicy: WhenEmpty | Balanced | WhenEmptyOrUnderutilized`, with the option to expand that policy to support additional values in the future, with an escape hatch to pass k directly into that field.
> 
> Ready to approve once the RFC reflects this. cc: @jonathan-innis @DerekFrank @jmdeal

While working through some of the cases here, I thought about what it really means for this RFC to take into account the whole spectrum between WhenEmpty and WhenEmptyOrUnderutilized. There's a node disruption cost in the current RFC, defaulting to just 1 (the same as disrupting a single pod). That means WhenEmpty can be expressed as: approve only when move disruption cost equals the per-node disruption cost. No pod contributes positive disruption cost. WhenEmptyOrUnderutilized is k=+inf (any positive savings passes). The whole spectrum lives in one model.

This also surfaces a small behavioral improvement. Today's WhenEmpty checks for literally zero pods. Under the disruption cost framing, a node whose pods all have large negative pod-deletion-cost (disruption cost clamped to 0) also qualifies. Pods that declared themselves free to disrupt shouldn't block consolidation.

Updated RFC incoming. Also making consolidationPolicy an IntOrString so integer values (1-3) pass k directly as the escape hatch.

pkg/controllers/disruption/consolidation.go
```diff
@@ -111,25 +126,64 @@ func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
...
-	if cn.NodePool.Spec.Disruption.ConsolidationPolicy != v1.ConsolidationPolicyWhenEmptyOrUnderutilized {
+	// Check if the policy enables non-empty consolidation (WhenEmptyOrUnderutilized or Balanced)
+	policy := cn.NodePool.Spec.Disruption.ConsolidationPolicy
+	if policy != v1.ConsolidationPolicyWhenEmptyOrUnderutilized && policy != v1.ConsolidationPolicyBalanced {
 		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("NodePool %q has non-empty consolidation disabled", cn.NodePool.Name))...)
 		return false
 	}
+	// If Balanced is set but the feature gate is disabled, set a status condition and reject
+	if policy == v1.ConsolidationPolicyBalanced && !options.FromContext(ctx).FeatureGates.BalancedConsolidation {
+		cn.NodePool.StatusConditions().SetTrueWithReason(v1.ConditionTypeConsolidationPolicyUnsupported,
```
nathangeology: My robo-Karpenter maintainers gave three bits of feedback to this particular code: 
"This sets the condition in-memory but there's no `Status().Patch()` call to persist it to the API server. Users won't see this via kubectl. Either add a status patch with optimistic locking, or remove the condition code and rely on the events (which are correctly emitted)."

"Is cn.NodePool deep-copied before this? If it's a reference from the informer cache via nodePoolMap, these StatusConditions mutations are visible to other goroutines reading the same NodePool."

"When the Balanced gate is disabled, this rejects the candidate entirely (returns false). The design doc says it should fall back to WhenEmptyOrUnderutilized behavior. Which is intended?" |

pkg/controllers/disruption/consolidation.go
```diff
@@ -111,25 +126,64 @@ func (c *consolidation) ShouldDisrupt(_ context.Context, cn *Candidate) bool {
...
+			"BalancedConsolidationDisabled",
+			"consolidationPolicy is Balanced but the BalancedConsolidation feature gate is disabled; change the policy or re-enable the gate")
+		c.recorder.Publish(disruptionevents.Unconsolidatable(cn.Node, cn.NodeClaim, fmt.Sprintf("NodePool %q uses Balanced consolidation but the BalancedConsolidation feature gate is disabled", cn.NodePool.Name))...)
+		return false
+	}
+	// Clear the condition if Balanced is enabled and working
+	if policy == v1.ConsolidationPolicyBalanced {
+		_ = cn.NodePool.StatusConditions().Clear(v1.ConditionTypeConsolidationPolicyUnsupported)
```
nathangeology: Again, setting status conditions without patching them to the API Server here is problematic according to my quality checker.

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,303 @@
...
+			"savings_fraction", result.SavingsFraction,
+			"disruption_fraction", result.DisruptionFraction,
+			"savings", poolSavings,
+			"disruption_cost", disruptionCost,
+			"nodepool_total_cost", totals.TotalCost,
+			"nodepool_total_disruption_cost", totals.TotalDisruptionCost,
+			"threshold", 1.0/float64(consolidationThreshold),
+			"approved", result.Approved,
+			"decision", cmd.Decision(),
+			"candidates", lo.Map(poolCandidates, func(c *Candidate, _ int) string { return c.Name() }),
```
nathangeology, line 175: "Minor: this `lo.Map` allocates even when V(1) logging is disabled. Could guard with `if log.V(1).Enabled()`." |

pkg/apis/v1/nodepool_status.go
```diff
@@ -29,6 +29,9 @@ const (
 	ConditionTypeNodeClassReady = "NodeClassReady"
 	// ConditionTypeNodeRegistrationHealthy = "NodeRegistrationHealthy" condition indicates if a misconfiguration exists that is preventing successful node launch/registrations that requires manual investigation
 	ConditionTypeNodeRegistrationHealthy = "NodeRegistrationHealthy"
+	// ConditionTypeConsolidationPolicyUnsupported indicates that the NodePool's consolidationPolicy
+	// is set to Balanced but the BalancedConsolidation feature gate is disabled.
+	ConditionTypeConsolidationPolicyUnsupported = "ConsolidationPolicyUnsupported"
```
nathangeology: This condition type is defined but never persisted (see consolidation.go:137). Either wire up a status patch or remove this constant if events are the intended mechanism.

pkg/controllers/disruption/balanced_test.go
```diff
@@ -0,0 +1,272 @@
...
+// savings = 7.26, disruption = 1.0 (node) + 3.0 (pods) = 4.0
+func TestScoreMove_OversizedNode_Approved(t *testing.T) {
+	savings := 7.26       // m7i.2xlarge - m7i.large = 9.68 - 2.42
+	disruptionCost := 4.0 // 1.0 per-node + 3 pods with default cost
+	result := ScoreMove(savings, disruptionCost, defaultTotals, defaultTolerance)
+
+	if !result.Approved {
+		t.Errorf("expected approved, got rejected (score=%.2f)", result.Score)
+	}
+	if !approxEqual(result.Score, 2.81, 0.02) {
```
nathangeology: Where does 2.81 come from? Input is `savings=4.84, disruption=5.0, totals={58.08, 90}`. A reviewer must manually compute `(4.84/58.08) / (5.0/90) / (1/2)` to verify. Might be good to have these expected values computed dynamically so that you don't have either erroneous passing tests after a change or erroneous failing tests after a change.

pkg/controllers/disruption/balanced_test.go
```diff
@@ -0,0 +1,272 @@
...
+// Test: Spare capacity delete -> score 1.50, approved
+func TestScoreMove_SpareCapacityDelete_Approved(t *testing.T) {
+	savings := 4.84       // full m7i.xlarge cost
+	disruptionCost := 5.0 // 1.0 per-node + 4 pods
+	result := ScoreMove(savings, disruptionCost, defaultTotals, defaultTolerance)
+
+	if !result.Approved {
+		t.Errorf("expected approved, got rejected (score=%.2f)", result.Score)
+	}
+	if !approxEqual(result.Score, 1.50, 0.02) {
```
nathangeology: See earlier comment on hard coded test conditions.

pkg/controllers/disruption/balanced_test.go
nathangeology: //instead of
if !approxEqual(result.Score, 2.81, 0.02) { ... }

// Prefer:
savingsFrac := savings / defaultTotals.TotalCost           // 4.84/58.08
disruptionFrac := disruptionCost / defaultTotals.TotalDisruptionCost  // 5.0/90
expectedScore := savingsFrac / disruptionFrac / (1.0 / float64(defaultTolerance))
if !approxEqual(result.Score, expectedScore, 0.02) { ... }

nathangeology: Also, you could potentially make test helper type functions to compute these.

pkg/controllers/disruption/balanced_scoring_test.go
```diff
@@ -0,0 +1,800 @@
...
+	return &cloudprovider.InstanceType{
+		Name:      name,
+		Offerings: cloudprovider.Offerings{makeOffering(price)},
+	}
+}
+
+// makeCandidate creates a minimal Candidate for testing balanced scoring.
+// nodeName is used for the node, poolName for the NodePool, instanceType may be nil,
+// and pods are the reschedulable pods assigned to the candidate.
+func makeCandidate(nodeName string, np *v1.NodePool, it *cloudprovider.InstanceType, pods []*corev1.Pod) *Candidate {
```
nathangeology, line 65: The integration tests in `balanced_integration_test.go` construct candidates manually rather than going through the full controller reconcile loop. This means:
- A bug in how `GetCandidates` builds candidates for Balanced scoring won't be caught
- The `ComputeCommands` -> `EvaluateBalancedMove` -> `Validate` path is never exercised end-to-end with real cluster state
- The interaction between balanced scoring and disruption budgets is untested

pkg/controllers/disruption/balanced_test.go
```diff
@@ -0,0 +1,272 @@
...
+import (
+	"math"
+	"testing"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+)
+
+// RFC example pool: 10 nodes, eight m7i.xlarge ($4.84/day) and two m7i.2xlarge ($9.68/day).
+// Total NodePool cost = $58.08/day. 80 pods, total disruption cost = 90 (10 nodes * 1.0 + 80 pods * 1.0).
+var defaultTotals = NodePoolTotals{
```
nathangeology: Where do 58.08 and 90 come from? These appear to be 12 nodes * $4.84/node and some disruption sum, but nothing documents this. If a test changes the number of nodes or their prices, the totals must be manually recalculated. A helper like `computeTotals(candidates)` would make tests self-maintaining.

nathangeology (commented):
The scoring formula, per-NodePool normalization, cross-NodePool handling, feature gate, observability (events + metrics), and candidate filtering seem good. The tests cover the key scenarios from the RFC.

The main blocker my PR checker sees is the nodepool status condition that is set in-memory but never persisted. This might need to be either wired to a status patch or removed entirely (just let events drive it?). Jason or Derek can maybe give some advice here, I don't fully understand this one apart from the idea that modifying the cached/copied nodepool doesn't auto-magically update the corresponding object living in the API server, but maybe we are ok in this case? 

There might be a feature gate fallback mismatch (balanced looks like it falls back to emptiness only instead of empty or underutilized). If this is indeed happening, it might be worth adding a unit test to catch this particular problem of falling back to the wrong consolidation mode. 

The tests have a lot of hard coded numbers and the integration tests don't seemt o cover some end-to-end paths. Throwing in some helpers to generate the correct numbers to test against would help make the test conditions up-to-date over time.

jamesmthompson:
@nathangeology — thanks for the detailed review. All points addressed:

**Status condition persistence (your main blocker):** Fixed in `f87f0f82`. DeepCopy each NodePool before candidate filtering, compare after, patch if status changed. Same pattern as the nodepool validation controller. The deep-copy also addresses your concern about mutating the informer cache.

**Feature gate fallback:** Fixed in `98c67e2d` (merged via PR #15). Balanced now falls back to WhenEmptyOrUnderutilized when the gate is disabled. `effectiveBalanced()` returns false when the gate is off, so the scoring path is skipped and the candidate proceeds under the old consolidation semantics.

**lo.Map allocation:** Wrapped in `log.V(1).Enabled()` guard in `57dc9f63`.

**Hard-coded test values:** Replaced with computed expected values derived from test inputs. Tests are self-maintaining when prices change. See `57dc9f63`.

**Integration test coverage:** The file you commented on (`balanced_scoring_test.go`) tests scoring in isolation. Full-loop Ginkgo integration tests exercising ComputeCommands through scoring through validation through execution with real cluster state are in `28cf7640` (`balanced_integration_test.go`). File-level comment clarifies the test layering.

pkg/apis/v1/nodepool_balanced_test.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// TestSetDefaults_NoOp verifies SetDefaults is a no-op (no separate threshold field to default).
+func TestSetDefaults_NoOp(t *testing.T) {
+	np := &NodePool{}
+	np.Spec.Disruption.ConsolidationPolicy = ConsolidationPolicyBalanced
+
+	np.SetDefaults(context.Background())
+
+	// With IntOrString, "Balanced" implies k=2 at the controller level.
```
Ignoramuss: nit: PR and design mention `IntOrString` but technically this field is a string. So users  likely cant specify  `consolidationPolicy: 1` ( As opposed to `consolidationPolicy: "1"`) which can be slightly confusing.

jamesmthompson:
@Ignoramuss good catch. Fixed in `5e0d18c9`. The field uses `x-kubernetes-int-or-string: true` in the CRD schema with CEL validation, plus a custom `UnmarshalJSON` that normalizes integers to strings. Users can write `consolidationPolicy: 2` without quotes. Added a TODO on Budget.Nodes which has the same limitation.

jamesmthompson:
@nathangeology replying inline to each thread since I can't resolve them from my side:

**consolidation.go (robo-maintainer feedback):**
- Status condition persistence: fixed in `f87f0f82`. DeepCopy before filtering, Status().Patch() after with optimistic locking.
- Deep-copy concern: same commit. `GetCandidatesWithTotals` DeepCopies each NodePool before `ShouldDisrupt` mutates it.
- Gate fallback: fixed in `98c67e2d` (merged via PR #15). Returns true and falls back to WhenEmptyOrUnderutilized semantics. Scoring gated separately via `effectiveBalanced()`.

**consolidation.go (second status condition comment):** Same fix, `f87f0f82`.

**balanced.go (lo.Map allocation):** Wrapped in `log.V(1).Enabled()` guard in `57dc9f63`.

**nodepool_status.go (condition never persisted):** Now persisted via `f87f0f82`.

**balanced_test.go (hard-coded values, 3 threads):** Replaced with computed expected values in `57dc9f63`.

**balanced_scoring_test.go (integration test coverage):** Full-loop Ginkgo tests added in `28cf7640`. Budget interaction test added in `17854ac3`. File-level comment added explaining the test layering.

**nodepool_balanced_test.go (IntOrString quoting):** Fixed in `5e0d18c9`. Field uses `x-kubernetes-int-or-string: true` with CEL validation.

All threads addressed. Please resolve when you get a chance.

pkg/controllers/disruption/consolidation.go
jamesmthompson: All three fixed:
1. Status persistence: `f87f0f82`. DeepCopy + Status().Patch() with optimistic locking.
2. Deep-copy: same commit. NodePools are DeepCopied in `GetCandidatesWithTotals` before `ShouldDisrupt` mutates them.
3. Gate fallback: `98c67e2d`. Returns true and falls back to WhenEmptyOrUnderutilized. Scoring gated via `effectiveBalanced()`.

pkg/controllers/disruption/consolidation.go
jamesmthompson: Fixed in `f87f0f82`. DeepCopy + Status().Patch() with optimistic locking after candidate filtering.

pkg/controllers/disruption/balanced.go
jamesmthompson: Wrapped in `log.V(1).Enabled()` guard in `57dc9f63`.

pkg/apis/v1/nodepool_status.go
jamesmthompson: Now persisted. Fixed in `f87f0f82`. DeepCopy before filtering, Status().Patch() after.

pkg/controllers/disruption/balanced_test.go
jamesmthompson: Replaced with computed expected values in `57dc9f63`. Tests now derive expectations from inputs using the same formula.

pkg/controllers/disruption/balanced_test.go
jamesmthompson: Fixed in `57dc9f63`. All hard-coded expected scores replaced with computed values.

pkg/controllers/disruption/balanced_scoring_test.go
jamesmthompson: Full-loop Ginkgo tests added in `28cf7640` (`balanced_integration_test.go`). Exercises ComputeCommands through scoring through validation through execution with real cluster state. Budget interaction test in `17854ac3`. File-level comment explains the test layering.

pkg/controllers/disruption/balanced_test.go
jamesmthompson: Fixed in `57dc9f63`. Expected values are now computed from test inputs, not magic numbers.

pkg/apis/v1/nodepool_balanced_test.go
jamesmthompson: Fixed in `5e0d18c9`. Field now uses `x-kubernetes-int-or-string: true` with CEL validation + custom `UnmarshalJSON`. Users can write `consolidationPolicy: 2` without quotes.

pkg/operator/options/options.go
```diff
@@ -61,6 +61,7 @@ type FeatureGates struct {
 	SpotToSpotConsolidation bool
 	NodeOverlay             bool
 	StaticCapacity          bool
+	BalancedConsolidation   bool
```
jmdeal: Users need to opt-in to balanced consolidation via an API configuration, do we also need a feature gate? The main reason to introduce a feature gate is if we believed we may roll this feature back entirely - while we may tweak the implementation, I don't see us rolling this back in a way that would break user contracts.

pkg/apis/v1/nodepool.go
```diff
@@ -130,6 +134,9 @@ type Budget struct {
 	// This cannot be of type intstr.IntOrString since kubebuilder doesn't support pattern
 	// checking for int nodes for IntOrString nodes.
 	// Ref: https://github.com/kubernetes-sigs/controller-tools/blob/55efe4be40394a288216dab63156b0a64fb82929/pkg/crd/markers/validation.go#L379-L388
+	//
+	// TODO(jamesmt): switch to XIntOrString + CEL (same as consolidationPolicy) so users
+	// can write nodes: 5 without quotes.
```
jmdeal: We should move this TODO or keep it on one line, it looks like kubebuilder drops the TODO line from the generated CRD but "can write nodes: 5 without quotes" is there.

pkg/apis/v1/nodepool.go
```diff
@@ -159,8 +166,28 @@ type ConsolidationPolicy string
 const (
 	ConsolidationPolicyWhenEmpty                ConsolidationPolicy = "WhenEmpty"
 	ConsolidationPolicyWhenEmptyOrUnderutilized ConsolidationPolicy = "WhenEmptyOrUnderutilized"
+	ConsolidationPolicyBalanced                 ConsolidationPolicy = "Balanced"
 )
 
+// UnmarshalJSON implements json.Unmarshaler, allowing ConsolidationPolicy to accept
+// both unquoted integers (1, 2, 3) and strings in YAML/JSON input.
+func (c *ConsolidationPolicy) UnmarshalJSON(data []byte) error {
```
jmdeal: What's the round trip behavior on this? If the user specifies a number as an integer rather than a string and we apply and we apply an update, will it be coerced to an integer (or vice-versa)? This type coercion behavior can be problematic when combined with tools like Flux and Argo since, even though the value is semantically equivalent, they will detect the CRD as being out of sync. API machinery's `intstr.IntOrString`  type correctly accounts for this by storing a type flag - if we're going to take this approach I'd use that. 

I do wonder if it's worth supporting this though - we should be able to ensure that the round trip behavior in Karpenter is correct, but it's a sharp edge for users writing custom controllers or mutating webhooks.

pkg/apis/v1/nodepool_balanced_test.go
jmdeal, line 1: We use Ginkgo for testing almost everywhere, I think the only exception is the scheduling benchmark tests and we should probably address that from a consistency standpoint at some point. Admittedly, I'm not the biggest fan of Ginkgo, but there's value in consistency across tests.

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -159,6 +166,24 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
 		}
+		// Apply balanced scoring if any candidate uses the Balanced policy
+		if validDecision {
+			if approved, perPool := m.balancedEvaluator.ApproveCommand(ctx, cmd); !approved {
+				validDecision = false
+				for _, poolCandidates := range lo.GroupBy(candidatesToConsolidate, func(c *Candidate) string { return c.NodePool.Name }) {
+					np := poolCandidates[0].NodePool
+					if IsBalancedPolicy(np.Spec.Disruption.ConsolidationPolicy) {
+						poolResult := perPool[np.Name]
+						m.recorder.Publish(disruptionevents.BalancedConsolidationRejectedMultiNode(
```
jmdeal: I'm a bit concerned about the volume of events here since, but I absolutely understand why we want to have events for both acceptance and rejection. I'm wondering what our best options to reduce the cardinality are right here. Right now we emit an event each time we restrict the size of the search window, which would be log2(100) ~= 6 times multiplied by the number of nodepools we evaluate in each window. These events likely won't be deduped because of different pool scores based on the different candidates being evaluated. A simple step that would reduce this by up to 6x would be to hold the failed events until we've evaluated all window sizes (or found a solution) and emit the failed events then. We could also consider removing some of the debug metadata from the message so it dedupes via the recorders native method or add an additional layer of deduping on top of the recorder.

pkg/apis/v1/nodepool_balanced_test.go
ellistarn: I've recently had a ton of success with go's native testing framework: https://pkg.go.dev/testing. Very open minded towards shifting the project towards that over time.

pkg/apis/v1/nodepool.go
```diff
@@ -93,10 +95,12 @@ type Disruption struct {
...
 	// ConsolidationPolicy describes which nodes Karpenter can disrupt through its consolidation
-	// algorithm. This policy defaults to "WhenEmptyOrUnderutilized" if not specified
-	// When replicas is set, ConsolidationPolicy is simply ignored
+	// algorithm. This policy defaults to "WhenEmptyOrUnderutilized" if not specified.
+	// Named values: "WhenEmpty", "WhenEmptyOrUnderutilized", "Balanced" (equivalent to k=2).
+	// Integer values 1-3 set the scoring threshold k directly.
+	// When replicas is set, ConsolidationPolicy is simply ignored.
 	// +kubebuilder:default:="WhenEmptyOrUnderutilized"
-	// +kubebuilder:validation:Enum:={WhenEmpty,WhenEmptyOrUnderutilized}
+	// +kubebuilder:validation:XIntOrString
```
ellistarn: I'm curious if we might want to cutt scope on the IntOrString. I know it's already implemented, but I'd love to pressure test why a default value isn't good enough to start with. Are we not confident in the choice?

pkg/apis/v1/nodepool_balanced_test.go
```diff
@@ -0,0 +1,52 @@
+/*
```
ellistarn: Feels odd to me to have a whole separate test file just to test this new feature. Imbalanced conceptually with the rest of this package.

pkg/apis/v1/nodepool_status.go
ellistarn: Do we need a whole condition for this? Can't we just draw an invariant at apply time? If we had to expose every invariant violation in the API, our API surface would grow quickly.. IIRC, this is what the consistency controller is for.

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,362 @@
...
+// NodePoolTotals holds precomputed cost and disruption sums for a NodePool.
+type NodePoolTotals struct {
+	TotalCost           float64
+	TotalDisruptionCost float64
+}
+
+// computeNodePoolTotals sums cost and disruption for each NodePool across all
+// candidates (before ShouldDisrupt filtering). Candidates with no resolvable
+// price are excluded to avoid skewing the ratio.
+func computeNodePoolTotals(allCandidates []*Candidate) map[string]NodePoolTotals {
```
ellistarn: How confident are we that this should be per node pool, rather than cluster global?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,362 @@
...
+		totals := totalsMap[name]
+		totals.TotalCost += price
+		totals.TotalDisruptionCost += candidateDisruptionCost(c)
+		totalsMap[name] = totals
+	}
+	return totalsMap
+}
+
+// candidatePrice returns the precomputed offering price for a candidate.
+func candidatePrice(c *Candidate) float64 {
```
ellistarn: nit: Can candidates actually be nil? I'd hope we have an invariant that prevents this. Reads a bit like AI defensive dead weight.

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,362 @@
...
+// computeNodePoolTotals sums cost and disruption for each NodePool across all
+// candidates (before ShouldDisrupt filtering). Candidates with no resolvable
+// price are excluded to avoid skewing the ratio.
+func computeNodePoolTotals(allCandidates []*Candidate) map[string]NodePoolTotals {
+	totalsMap := map[string]NodePoolTotals{}
+	for _, c := range allCandidates {
+		price := candidatePrice(c)
+		if price == 0 {
+			continue
+		}
```
ellistarn: Can the price really be 0?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,362 @@
...
+	for _, c := range candidates {
+		cost += candidateDisruptionCost(c)
+	}
+	return cost
+}
+
+// candidateDisruptionCost returns the precomputed reschedule disruption cost.
+func candidateDisruptionCost(c *Candidate) float64 {
+	return c.RescheduleDisruptionCost
+}
```
ellistarn: dead weight

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+}
+
+// ComputeMoveDisruptionCost sums disruption cost across candidates: 1.0 per node
+// plus sum(max(0, EvictionCost(pod))) for reschedulable pods.
+//
+// Does not include LifetimeRemaining. Candidate ordering already uses lifetime-
+// adjusted DisruptionCost (near-expiration nodes sort first); scoring evaluates
+// static cost structure. The two compose without overlap.
+func ComputeMoveDisruptionCost(ctx context.Context, candidates []*Candidate) float64 {
```
ellistarn: dead weight accumulator

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// candidateDisruptionCost returns the precomputed reschedule disruption cost.
+func candidateDisruptionCost(c *Candidate) float64 {
+	return c.RescheduleDisruptionCost
+}
+
+// GetConsolidationThreshold returns k for a NodePool. Falls back to
+// DefaultConsolidationThreshold for non-scoring policies.
+func GetConsolidationThreshold(nodePool *v1.NodePool) int32 {
```
ellistarn: why not encapsulate this inside of consolidationpolicy.Threshold()?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,355 @@
...
+	return k
+}
+
+// effectiveBalanced returns true when the NodePool uses a scoring policy.
+func effectiveBalanced(_ context.Context, np *v1.NodePool) bool {
+	return IsBalancedPolicy(np.Spec.Disruption.ConsolidationPolicy)
+}
+
+// IsBalancedPolicy reports whether the policy uses scoring (Balanced or integer 1-3).
+func IsBalancedPolicy(policy v1.ConsolidationPolicy) bool {
```
ellistarn: I thought it all collapsed into a single model, where whenEmpty is k=inf and whenUnderutilized is k=0

pkg/controllers/nodepool/consolidationpolicy/controller.go
```diff
@@ -0,0 +1,84 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/operator/injection"
+	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
+)
+
+// Controller manages the ConsolidationPolicyUnsupported condition on NodePools.
```
ellistarn: Is this worth its weight?

.gitignore
```diff
@@ -14,7 +14,6 @@ bin/
 .idea
 .vscode
 .DS_Store
-.kiro
```
ellistarn: intentional?

pkg/controllers/disruption/balanced.go
jamesmthompson: Deleted.

pkg/controllers/disruption/balanced.go
jamesmthompson: Simplified to lo.SumBy one-liner.

pkg/apis/v1/nodepool_balanced_test.go
jamesmthompson: Agree. Will fold into the existing nodepool test files.

pkg/apis/v1/nodepool_balanced_test.go
jamesmthompson: I ended up converting all of the balanced consolidation tests to Ginkgo. If the project moves to native testing later, or if people want it the other way, I'm happy to update.

pkg/controllers/disruption/balanced.go
jamesmthompson: The RFC defines WhenEmpty in terms of disruption scoring, so we approve when no pod contributes positive disruption cost (the eviction costs can't be negative, the pod disruption costs are in a very wide range limited only by signed integer limits). A node full of pods with negative pod-deletion-cost gets the same behavior as a node with zero pods. Pods that declared themselves free to disrupt don't block consolidation, but they used to with WhenEmpty. This is a behavioral change, but I think it's the right one. I've had a hard time getting comfortable with it, but I've convinced myself that this is right. Happy to convince anyone else willing to listen as well!

I've updated the code to match my newfound conviction. WhenEmpty checks RescheduleDisruptionCost <= 1.0 (the per-node base) instead of literal pod count. WhenEmptyOrUnderutilized keeps its existing path (any positive savings, equivalent to k=+inf). Balanced runs the scoring formula.

.gitignore
jamesmthompson: No. Restored .kiro to .gitignore after an accidental removal.

pkg/controllers/disruption/balanced.go
jamesmthompson: Done. `GetConsolidationThreshold` is removed. Callers use `policy.Threshold()` directly.

pkg/controllers/nodepool/consolidationpolicy/controller.go
jamesmthompson: Definitely not now. The feature gate and condition are both gone. The code now trusts that users can opt-in via the API field, CEL validates at apply time, and we don't need any of this any more.

pkg/apis/v1/nodepool_status.go
jamesmthompson: Agree. Removed the condition and the dedicated controller. CEL validates at apply time.

pkg/controllers/disruption/balanced.go
jamesmthompson: Removed the nil candidate check. Candidates can't be nil after `NewCandidate` (it returns an error instead). Kept the nil instanceType path since `NewCandidate` allows it for deprecated instance types that aren't in the instance type map.

pkg/controllers/disruption/balanced.go
jamesmthompson: Yes. Nodes with deprecated instance types (not in the cloud provider's current list) have no matching offering, so price resolves to 0. Drift should handle those nodes, but they can appear as candidates briefly.

pkg/controllers/disruption/balanced.go
jamesmt-aws: Removed.

pkg/operator/options/options.go
jamesmthompson: Agree. Removed the feature gate. Users opt-in via the API field. If we need to roll back, we remove the policy value in a future release.

pkg/apis/v1/nodepool.go
jamesmthompson: Fixed. Moved the TODO after the kubebuilder markers.

pkg/apis/v1/nodepool.go
jamesmthompson: Valid concern. The UnmarshalJSON normalizes integers to strings, so the stored value is always a string. On re-read via the API, the field comes back as a string even if submitted as an integer. Flux/Argo would detect drift on the first cycle, then stabilize.

If this is unacceptable, we can revert to requiring quotes (same as Budget.Nodes) or switch to `intstr.IntOrString` with its type flag. Waiting on Ellis's take on whether to cut this scope.

jamesmthompson: I ended up deciding to cut this. The updated PR reverts to quoted strings with pattern validation, same as Budget.Nodes. The round-trip concern you raised and Ellis's scope feedback both argue against carrying custom type coercion when the scoring logic works identically either way. Users write `"2"` or `"Balanced"` and we avoid the Flux/Argo drift problem entirely.

pkg/apis/v1/nodepool.go
jamesmthompson: I'm confident in k=2 as the default. The design doc's "Why k=2" section exhaustively enumerates the state space: k=2 is the smallest integer where within-family REPLACEs are viable, with max 4-step churn chains. k=1 approves only free replacements (useless). k=3 adds 8 cross-family pairs but the same churn bound. k=4 doubles max churn to 9 steps.

Cutting the IntOrString is fine. jmdeal raised a round-trip concern (Argo/Flux drift from type coercion), so I've reverted to quoted strings with pattern validation. Balanced ships with k=2 hardcoded as the default. If operators need k=1 or k=3 later, they quote the integer. The scoring formula and the API both support it today without any additional machinery.

pkg/controllers/disruption/balanced.go
jamesmthompson: (Correcting my earlier reply.)

The RFC covers this under "NodePool Totals" (lines 100-111) and "Scale Invariance" (lines 252-262). The scoring formula normalizes against NodePool totals so it's scale-invariant: a move on a 10-node pool produces the same score as the same move on a 100-node pool.

Imagine a 3-node pool where each node costs $5/day, total pool cost $15. Deleting one node saves $5, which is 33% of the pool. Looks like a great move. If the rest of the cluster is running expensive instances and we used a cluster-global denominator, this smaller general-purpose pool might never have plausible moves. Per-pool means "is this move proportionally good for this pool?" regardless of what else exists in the cluster.

ellistarn: Customers use node pools in weird ways, though. Often you have multiple pools with different weights that are logically a single pool of capacity. It's essentially an OR preference fall through for instance types. Depending on capacity availability, these pools could be arbitrarily sized.

pkg/controllers/disruption/balanced.go
ellistarn: Doesn't this let you collapse the code into a single algorithm, then, and whenEmpty just treats k=intmax

pkg/controllers/disruption/balanced.go
jamesmthompson: What's the actual case we're worried about here? I think that per-pool scoring still answers the right question, which is "is the disruption cost to the pods hosted by this NodePool justified by the cost savings we get from turning off one or more nodes and doing appropriate replacements?" Quick thoughts:

1. Replacement nodes are in different NodePools, which is weird. But I think this is ok as long as the costs are comparable. If the costs aren't comparable, cross-NodePool moves get weird. But I think that the balanced consolidation behavior is pretty sensible for cross-NodePool moves. 
2. The size of the pool doesn't matter for average scores, because we're normalizing to NodePool totals. It does matter for variance though. Scores of consolidation moves for nodes in a smaller NodePool have higher variance. This should only matter if that smaller NodePool has constantly good consolidation opportunities though.
3. The per-NodePool normalization matches how we think about budgets and policies. 

The biggest risk I could imagine is that if you have a small NodePool that just constantly has good moves. But is that a problem with budgets being NodePool-specific? It doesn't worry me. But if you have a good example case to think through, I'm happy to do that!

pkg/controllers/disruption/balanced.go
jamesmthompson: yeah, that's much better. updated!

pkg/controllers/disruption/balanced.go
jamesmthompson: Updating from our offline conversation on Slack. I totally misunderstood the question from Ellis. We talked about "why do normalization of costs per-nodepool over cluster-global? Is the added complexity worth it?"

My point is that we have a lot of existing things in consolidation that are already per-nodepool, like disruption budgets and feasible instances. Being in-line with the existing design is good. Nodepools are this nice little sub-slice of a cluster, with its own instance prices, and its own pods. If more than one nodepool exists, there's probably a reason, and that reason could plausibly result in different instance costs and pod disruption costs.

Consolidation scoring should follow the nodepool boundary. The important thing about consolidation is the pod disruption cost from the source nodes, the dollar cost of the source nodes, and the dollar cost of any replacement nodes (we only generate moves with one replacement today, the balanced consolidation code handles more than one if that ever changes). But then we have to normalize the dollar cost and disruption costs to something. The RFC argues that this should be done relative to the source nodepool, because that's the node we're turning off. I can't find any examples where this would make me concerned.

pkg/controllers/disruption/multinodeconsolidation.go
jamesmthompson: Ok, thought about this more over the weekend and I can't make a good case for rejection events, so I removed them. Only approval events are published as Kubernetes events. Rejections are logged at DEBUG level (V(1)) for debugging. Event volume is O(approved moves per cycle), not O(candidates).

pkg/controllers/disruption/balanced.go
ellistarn: Thanks for the summary. I was convinced by the fact that this score is configured on the node pool, and there isn't really a natural place elsewhere to configure this elsewhere. Node pool is a little bit overloaded in Karpenter, its both a capacity sharding boundary, but also a preference mechanism (weight). This means that sometimes workloads are schedulable to multiple node pools, but sometimes they are bound to a single one. For the purposes of this feature, I think it makes sense to lean into the sharding use case, even if it's not perfect.

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
+	"sigs.k8s.io/karpenter/pkg/events"
+)
+
+// ScoreResult holds the three values needed to decide whether a move passes.
+type ScoreResult struct {
+	SavingsFraction    float64
+	DisruptionFraction float64
+	K                  int32
```
DerekFrank: Is there a chance that K changes between results? Why include it in the struct? To make Score() pure?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+	if len(cmd.Candidates) == 1 {
+		candidate := cmd.Candidates[0]
+		poolName := candidate.NodePool.Name
+		result, scored := perPool[poolName]
+
+		decisionLabel := "approved"
+		if scored && !result.Approved() {
+			decisionLabel = "rejected"
+		}
+		ConsolidationMovesTotal.Inc(map[string]string{"decision": decisionLabel, "nodepool": poolName})
```
DerekFrank: So we only count metrics for single?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+			}
+		}
+	}
+
+	return allApproved, perPool
+}
+
+// EmitMultiNodeEvents emits events and metrics for a multi-node command.
+// approved is the overall command outcome (every Balanced pool approved).
+func (e *BalancedEvaluator) EmitMultiNodeEvents(ctx context.Context, cmd Command, perPoolResults map[string]ScoreResult, approved bool) {
```
DerekFrank: Do we need both this and EmitBalancedMultiNodeEvents?

pkg/controllers/disruption/balanced.go
DerekFrank: Maybe we want policy as a label? It might be nice to compare balanced scores against Underutilized, but I'm not convinced

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+	savingsFraction := c.Price / totals.TotalCost
+	disruptionFraction := c.RescheduleDisruptionCost / totals.TotalDisruptionCost
+	deleteRatio := savingsFraction / disruptionFraction
+	threshold := 1.0 / float64(v1.BalancedK)
+	return deleteRatio >= threshold
+}
+
+// candidateSavingsRatio returns cost per unit disruption (higher = prefer to disrupt).
+func candidateSavingsRatio(c *Candidate) float64 {
```
DerekFrank: Is this worth putting into candidate?

i.e.


```
candidate.SavingsRatio()
```

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+// candidatesCost sums Price across a set of candidates.
+func candidatesCost(candidates []*Candidate) float64 {
+	return lo.SumBy(candidates, func(c *Candidate) float64 { return c.Price })
+}
+
+// BalancedEvaluator wraps precomputed NodePool totals and provides a single
+// ApproveCommand entry point for balanced scoring in both consolidation paths.
+type BalancedEvaluator struct {
```
DerekFrank: So right now we have the Validator pattern that lets different methods validate a command. This strikes me as an Evaluator pattern where a command is evaluated based on the policy. Does introducing a shared abstraction for both Underutilized and Balanced simplify the internal code architecture?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/samber/lo"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	disruptionevents "sigs.k8s.io/karpenter/pkg/controllers/disruption/events"
+	"sigs.k8s.io/karpenter/pkg/events"
+)
+
+// ScoreResult holds the three values needed to decide whether a move passes.
+type ScoreResult struct {
```
DerekFrank: For nits, these typically belong in types.go and then the resulting functions for it would be in a file like scoreresult.go. I don't know how much simpler that makes things

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+		return math.Inf(1)
+	}
+	return r.SavingsFraction / r.DisruptionFraction
+}
+
+func (r ScoreResult) Threshold() float64 { return 1.0 / float64(r.K) }
+func (r ScoreResult) Approved() bool     { return r.Score() >= r.Threshold() }
+
+// NodePoolTotals holds precomputed cost and disruption sums for a NodePool.
+type NodePoolTotals struct {
```
DerekFrank, line 33: Does this belong in cluster.go? Strikes me as an efficient index we'd like to keep around

DerekFrank: For instance, total cost of a nodepool is already pre-computed in clustercost IIRC. Total disruption cost could be treated similarly

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+	disruptionFraction := disruptionCost / totals.TotalDisruptionCost
+	return ScoreResult{
+		SavingsFraction:    savingsFraction,
+		DisruptionFraction: disruptionFraction,
+		K:                  k,
+	}
+}
+
+// ComputeMoveDisruptionCost sums RescheduleDisruptionCost across candidates.
+func ComputeMoveDisruptionCost(candidates []*Candidate) float64 {
```
DerekFrank, line 124: Should this be in command? Like each command has its cmd.DisruptionCost

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+func (e *BalancedEvaluator) EmitMultiNodeEvents(ctx context.Context, cmd Command, perPoolResults map[string]ScoreResult, approved bool) {
+	EmitBalancedMultiNodeEvents(ctx, cmd, perPoolResults, approved, e.recorder)
+}
+
+// CanPassThreshold returns true if a candidate's best-case score (its delete
+// ratio) meets the 1/k threshold. A DELETE saves the full node cost with zero
+// replacement cost, so it is the upper bound on any move's score. If even the
+// DELETE cannot pass, no REPLACE will either. Non-Balanced pools impose no
+// scoring constraint, so this always returns true for them.
+func (e *BalancedEvaluator) CanPassThreshold(c *Candidate) bool {
```
DerekFrank: If we do switch to an evaluator pattern, we'll want one entry point, something like Evaluate(cmd)

pkg/controllers/disruption/controller.go
```diff
@@ -98,12 +98,12 @@ func NewController(clk clock.Clock, kubeClient client.Client, provisioner *provi
 func NewMethods(clk clock.Clock, cluster *state.Cluster, kubeClient client.Client, provisioner *provisioning.Provisioner, cp cloudprovider.CloudProvider, recorder events.Recorder, queue *Queue) []Method {
 	c := MakeConsolidation(clk, cluster, kubeClient, provisioner, cp, recorder, queue)
 	return []Method{
-		// Delete any empty NodeClaims as there is zero cost in terms of disruption.
-		NewEmptiness(c),
 		// Terminate and create replacement for drifted NodeClaims in Static NodePool
 		NewStaticDrift(cluster, provisioner, cp),
 		// Terminate any NodeClaims that have drifted from provisioning specifications, allowing the pods to reschedule.
 		NewDrift(kubeClient, cluster, provisioner, recorder, clk),
+		// Delete empty nodes across all consolidation policies (WhenEmpty, WhenEmptyOrUnderutilized, Balanced).
```
DerekFrank: Why move this down?

pkg/controllers/disruption/types.go
```diff
@@ -79,12 +81,58 @@ type Candidate struct {
...
+func (c *Candidate) computePrice() float64 {
+	if c.instanceType == nil {
+		return 0
+	}
+	reqs := scheduling.NewLabelRequirements(c.Labels())
+	offerings := c.instanceType.Offerings.Compatible(reqs)
+	if len(offerings) == 0 {
+		return 0
+	}
+	price := offerings.Cheapest().Price
```
DerekFrank: Why grab the cheapest? Why not the offering that actually matches the node type?

pkg/controllers/disruption/types.go
```diff
@@ -132,11 +180,14 @@ func NewCandidate(ctx context.Context, kubeClient client.Client, recorder events
 		reschedulablePods: lo.Filter(pods, func(p *corev1.Pod, _ int) bool { return pod.IsReschedulable(p) }),
 		// We get the disruption cost from all pods in the candidate, not just the reschedulable pods
 		DisruptionCost: disruptionutils.ReschedulingCost(ctx, pods) * disruptionutils.LifetimeRemaining(clk, nodePool, node.NodeClaim),
-	}, nil
+	}
+	c.Price = c.computePrice()
+	c.RescheduleDisruptionCost = c.computeRescheduleDisruptionCost(ctx)
```
DerekFrank: Can we inline these into the instantiation?

pkg/controllers/disruption/balanced.go
```diff
@@ -0,0 +1,267 @@
...
+				nodePool,
+				result.Score(), result.Threshold(), result.K,
+				result.SavingsFraction*100, result.DisruptionFraction*100,
+			))
+		}
+	}
+}
+
+// candidatesCost sums Price across a set of candidates.
+func candidatesCost(candidates []*Candidate) float64 {
```
DerekFrank: Probably belongs in helpers.go

DerekFrank: Are we mostly doing this on a per-command basis? we'd probably rather this just be a field in the command, no?

pkg/controllers/disruption/balanced.go
jamesmthompson: Done. Moved to `c.SavingsRatio()` on Candidate.

pkg/controllers/disruption/controller.go
jamesmthompson: Incidental. Reverted.

pkg/controllers/disruption/balanced.go
jamesmthompson: Added. Policy label on both ConsolidationMovesTotal and ConsolidationScoreHistogram.

pkg/controllers/disruption/balanced.go
jamesmthompson: Keeping ScoreResult in balanced.go. It's scoring-specific, not a general type. CanPassThreshold now uses ScoreMove + result.Approved() instead of duplicating the math inline, so ScoreResult has two independent consumers within the package (filtering and scoring). If it grows further, moving it makes sense.

pkg/controllers/disruption/types.go
jamesmthompson: Done. Pulled both into the struct literal. computePrice and computeRescheduleDisruptionCost are now standalone functions.

pkg/controllers/disruption/balanced.go
jamesmthompson: K is static per-policy. I wanted to make Threshold() and Approved() pure, so that callers don't need to pass K around. I can pull it out if you like but in every use case I could imagine, having different values of K feels more like a footgun than a feature.

pkg/controllers/disruption/balanced.go
jamesmthompson: Done. Introduced an Evaluator interface with a noopEvaluator (always approves) and balancedEvaluator (scores). Removes nil checks. The no-op handles the Underutilized path.

pkg/controllers/disruption/balanced.go
jamesmthompson: Done. Single entry point: evaluator.ApproveCommand(ctx, cmd). Callers don't know whether it's noop or balanced internally.

pkg/controllers/disruption/balanced.go
jamesmthompson: Inlined. EmitBalancedMultiNodeEvents is gone. The body lives directly in balancedEvaluator.EmitMultiNodeEvents(). One less level of indirection.

pkg/controllers/disruption/balanced.go
jamesmthompson: No. Multi-candidate gets metrics too, via evaluator.EmitMultiNodeEvents after the binary search completes. Single-candidate emits inline because the decision is immediate.

pkg/controllers/disruption/balanced.go
jamesmthompson: Looked at this. The only caller passes a per-pool candidate slice, not cmd.Candidates. A cmd.DisruptionCost() would compute the total across all pools, but scoring needs each pool's disruption independently. Same reason sumCandidatePrices stayed as a helper. If we add cross-pool metrics that need the aggregate, we'll add it then.

pkg/controllers/disruption/types.go
jamesmthompson: A node can run on multiple compatible offerings (different spot zones, reserved vs on-demand). Cheapest captures what Karpenter would actually pay to place this workload. Using the node's original offering would miss cheaper alternatives and overstate savings.

pkg/controllers/disruption/balanced.go
jamesmthompson: Done in `cd166276`. TotalDisruptionCost now lives in NodePoolCost alongside TotalCost. Both read from the same index via GetNodepoolCost / GetNodepoolDisruptionCost.

The asymmetry: TotalCost is maintained incrementally (offering add/remove). TotalDisruptionCost depends on per-reconcile pod state, so it gets written during candidate computation and read back during scoring. Same index, different update cadence.

pkg/controllers/disruption/balanced.go
jamesmthompson: Yes, it's a field on Command now (cd166276). One wrinkle: it's per-pool rather than a single aggregate, because multi-pool commands score each pool independently. The method falls back to computing from candidates if the map is nil, so existing test code that constructs bare Command literals still works.

pkg/controllers/disruption/balanced.go
DerekFrank: cluster.go already gets called on every UpdatePod RemovePod we should also maintain disruptionCost incrementally IMO

pkg/controllers/disruption/types.go
DerekFrank: But isn't this the instanceType of a resolved node? not a new nodeclaim, but [here](https://github.com/kubernetes-sigs/karpenter/pull/2962/changes#diff-2a7cf721cf6d408a7025db62063f6d774475dc1f9acf69ff7957c802f52b39f3R187) you're calling it on instanceType, which is resolved [here](https://github.com/kubernetes-sigs/karpenter/pull/2962/changes#diff-2a7cf721cf6d408a7025db62063f6d774475dc1f9acf69ff7957c802f52b39f3R166) from the candidate node? 

Is the returned price supposed to be the estimated price of the new node or the actual price of the old one?

pkg/controllers/disruption/balanced.go
jamesmthompson: Updating my earlier reply: the follow-up commit (cd166276) caches per-pool disruption costs on Command via `PoolDisruptionCosts`, which I think is what you were getting at. Scoring evaluates each pool independently, so an aggregate `cmd.DisruptionCost()` would need to be decomposed back into per-pool slices at the call site. Per-pool map avoids that round-trip.

pkg/controllers/disruption/balanced.go
jamesmthompson: Yeah, you were right. Fixed in `4177d463`. Per-node disruption cost is now maintained incrementally on StateNode via `updateForPod`/`cleanupForPod`. Same lifecycle as TotalCost: updated on every pod event, read at reconcile time. The approximation is gone, that was a premature optimization that we really didn't need.

pkg/controllers/disruption/types.go
jamesmthompson: My earlier reply was wrong. This IS the actual price of the old node. The offerings are filtered to the node's exact zone + capacity type via its labels, which resolves to a single offering on standard clouds. `.Cheapest()` is just a safe accessor for that one match. Renamed to `resolveNodePrice` in `4a557e8e` to make the intent obvious.

pkg/controllers/disruption/balanced.go
jamesmthompson: Done in `3bbb8959`. Moved ScoreResult and its methods to types.go.

pkg/controllers/disruption/balanced.go
jamesmthompson: (Updating commit ref after squash: `4c7c06e6`)

pkg/controllers/disruption/types.go
jamesmthompson: (Updating commit ref after squash: `4c7c06e6`, function renamed to `resolveNodePrice`)

DerekFrank: Ah gotcha, just wanted to make sure that I understand what this is for.

It might be outside the scope of this PR, but we do now have two ways to find an offering price for a nodeclaim (see [cost.go](https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/state/cost/cost.go#L161)), its probably worth deduping those two, maybe in a follow up PR if we can't  Maybe we can introduce a ClusterCost.GetNodeClaimCost(nodeclaim) and just reference that? The alternative is to plumb that through cluster.go or similar

DerekFrank:
Looking very close! Just want to close on the cost accessor and I think we're there

pkg/controllers/disruption/types.go
jamesmthompson: Done in `9e5e8349`. Added `InstanceType.OfferingPrice(zone, capacityType)` so both `resolveNodePrice` and `findOfferingPrice` in cost.go delegate to the same method. The logic lives where the data lives.

jamesmthompson:
> Looking very close! Just want to close on the cost accessor and I think we're there

Awesome! I think I addressed that last item, let me know if you'd like for me to do anything else.

DerekFrank (approved):
/lgtm

PR #2963, opened 2026-04-11
feat: topology-aware drift disruption ordering via DriftPolicy

BrunoChauvet:
**Description**

When many nodes are drifted at once, Karpenter fans out disruption across all availability zones simultaneously. This is fast, but it can reduce availability for workloads that are sensitive to how replacements roll across failure domains.

For instance stateful workloads replicating data across topology domains, may be impacted of nodes holding the same partition across distinct domains are concurrently disrupted.

This PR adds an opt-in `DriftPolicy` field to `NodePool.Spec.Disruption` that lets cluster operators control how drift disruption proceeds across topology domains. When `topologyKey` is set, Karpenter disrupts one domain at a time — finishing replacements in `us-east-1a` before moving to `us-east-1b`, for example — rather than spreading disruption across all zones at once. An optional `maxConcurrentPerDomain` field (absolute or percentage) controls how many nodes can be replaced simultaneously within the active domain.

The active domain is selected alphabetically, which gives stable and predictable behaviour across controller restarts and as replacements progress. Nodes that do not carry the topology label are unaffected and remain eligible for disruption regardless. Without `DriftPolicy` set, behaviour is unchanged from today.

```yaml
# Disrupt one zone at a time, one node at a time
disruption:
  driftPolicy:

# Disrupt one zone at a time, up to 3 nodes at once
disruption:
  driftPolicy:
    maxConcurrentPerDomain: "3"
```

`DriftPolicy` is intentionally separate from `Budget` — budgets control how many disruptions are allowed in total; `DriftPolicy` controls the order in which domains are disrupted. The two are orthogonal and can be combined.

Documentation update https://github.com/aws/karpenter-provider-aws/pull/9073

- Unit tests for `DriftPolicy` API validation (valid keys, invalid keys, percentage values)
- Integration tests for domain selection ordering, per-domain budget gating, configurable concurrency, unlabelled node fall-through, and the empty-first invariant within the active domain
- CRDs regenerated and verified
- `make presubmit`


GnatorX:
Have you tried using pod deletion cost to inform Karpenter (and kube controller manager) to evenly tear down nodes/ pods by AZ to keep AZ balance?

DerekFrank:
Discussed this in Karpenter Working Group

Given the usecase is actually application aware disruption (don't disrupt my kafka brokers until the new ones are in-sync), we're hoping that this can be a launching point of discussion with node-lifecycle working group and sig-apps to build a workload\application level abstraction instead of a node level abstraction

BrunoChauvet:
> Have you tried using pod deletion cost to inform Karpenter (and kube controller manager) to evenly tear down nodes/ pods by AZ to keep AZ balance?

This could be an option, Kafka broker pods could be annotated with `controller.kubernetes.io/pod-deletion-cost` with a value increasing in each AZ (a: 10, b: 20, c: 30)
My understanding is this feature will be supported as part of [Karpenter Balanced Consolidation RFC](https://github.com/kubernetes-sigs/karpenter/blob/da2a854455cb78705dbc501cb3415d83e95930c3/designs/balanced-consolidation.md)

DerekFrank:
The pod deletion cost is already supported, but its not a solid solution because it doesn't block drain, just informs which nodes cost more to disrupt.

DerekFrank:
Closing in favor of a sig-apps proposal

PR #2971, opened 2026-04-14
fix: Skip storing scheduling timestamp for already-bound pods

Sarthug99:
Fixes #N/A

**Description**
Pods from deleting nodes are included in the scheduling simulation for capacity planning but are already bound. Updating their scheduling time results in the emission of negative metrics.

`make presubmit`

### DIFF
--- a/pkg/controllers/state/cluster.go
+++ b/pkg/controllers/state/cluster.go
@@ -422,6 +422,7 @@ func (c *Cluster) PodAckTime(podKey types.NamespacedName) time.Time {
 // It updates podHealthyNodePoolScheduledTime for pods scheduled against nodePool that have
 // NodeRegistrationHealthy=true. This also marks when the pod is first seen as schedulable for pod metrics.
 // We'll only emit a metric for a pod if we haven't done it before.
+// nolint:gocyclo
 func (c *Cluster) MarkPodSchedulingDecisions(ctx context.Context, podErrors map[*corev1.Pod]error, npPods map[string][]*corev1.Pod, ncPods map[string][]*corev1.Pod) {
 	now := c.clock.Now()
 	for pod := range podErrors {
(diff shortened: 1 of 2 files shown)


pkg/controllers/state/cluster.go
```diff
@@ -458,8 +459,14 @@ func (c *Cluster) MarkPodSchedulingDecisions(ctx context.Context, podErrors map[
 			}
 			// If the pod is scheduled to a nodePool and if the nodePool has NodeRegistrationHealthy=true
 			// then mark the time when we thought it can schedule to now.
+			// Skip pods that are already bound to a node (e.g. pods from deleting nodes
+			// included in the scheduling simulation for capacity planning). Storing a new
+			// timestamp for already-bound pods would cause negative metric values since
+			// their PodScheduled LastTransitionTime is in the past.
 			if nodePoolName != "" && nodePool.StatusConditions().IsTrue(v1.ConditionTypeNodeRegistrationHealthy) {
-				c.podHealthyNodePoolScheduledTime.LoadOrStore(nn, c.clock.Now())
+				if !podutils.IsScheduled(p) {
```
jigisha620, line 467: Can we add a test for this? We should be able to get the `podHealthyNodePoolScheduledTime` stored in the memory using - https://github.com/kubernetes-sigs/karpenter/blob/4e80a7e534d0c10674c0be835c1013c80e8b6dfe/pkg/controllers/state/cluster.go#L509.

jigisha620:
Doesn't this affect the metrics defined here - https://github.com/kubernetes-sigs/karpenter/blob/4e80a7e534d0c10674c0be835c1013c80e8b6dfe/pkg/controllers/metrics/pod/controller.go? Should we also update https://github.com/kubernetes-sigs/karpenter/blob/4e80a7e534d0c10674c0be835c1013c80e8b6dfe/pkg/controllers/state/cluster.go#L450 to not store scheduledTime for already bound pods?

pkg/controllers/state/cluster.go
Sarthug99: yes, I've added the test in a new commit.

jigisha620 (approved):
/lgtm

DerekFrank (approved):
/lgtm

PR #2975, opened 2026-04-16
feat: account for instance type requirements for daemonset scheduling

### LINKED ISSUE #715: DaemonSets not being correctly calculated when choosing a node

GaneshBannur:
Fixes #715  

**Description**
Continuation of #2486 with feedback addressed

`make presubmit`


pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -779,49 +770,62 @@ func parallelizeUntil(workers, pieces int, doWorkPiece func(int) bool) {
...
-			hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+			occupiedPorts[it.Name] = hostPortUsage
 		}
-		nctToOccupiedPorts[nct] = hostPortUsage
+		nctToOccupiedPorts[nct] = occupiedPorts
 	}
 	return nctToOccupiedPorts
 }
 
+func getCompatibleDaemonPods(ctx context.Context, daemonSetPods []*corev1.Pod, nct *NodeClaimTemplate, it *cloudprovider.InstanceType) []*corev1.Pod {
```
ryan-mist: Can we compute `getCompatibleDaemonPods` once per DS x NCT x IT? I don't think both `getDaemonHostPortUsage` and `getDaemonOverhead` need to individually call this function

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -322,65 +340,98 @@ type InstanceTypeFilterError struct {
 	// We capture podRequests here since when a pod can't schedule due to requests, it's because the pod
 	// was on its own on the simulated Node and exceeded the available resources for any instance type for this NodePool
 	podRequests corev1.ResourceList
-	// We capture daemonRequests since this contributes to the resources that are required to schedule to this NodePool
-	daemonRequests corev1.ResourceList
+	// Min and max daemon overhead bounds for error reporting.
```
ryan-mist: nit: maybe something like
```
We capture the min and max daemonRequests since this contributes to the resources that are required to schedule to this NodePool. The min/max DaemonRequests are the min / max across all instance types.
```

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -179,16 +164,49 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements
...
+		n.hostPortUsage[it.Name].Add(pod, hostPorts)
+	}
 	n.reservationManager.Reserve(n.hostname, offeringsToReserve...)
 	n.releaseReservedOfferings(n.reservedOfferings, offeringsToReserve)
 	n.reservedOfferings = offeringsToReserve
 }
 
+// ToNodeClaim overrides NodeClaimTemplate.ToNodeClaim to add the minimum daemon overhead
+// across remaining candidates. Daemon overhead is excluded during bin packing to avoid double-counting.
+func (n *NodeClaim) ToNodeClaim() *v1.NodeClaim {
```
ryan-mist: ~~Following up on https://github.com/kubernetes-sigs/karpenter/pull/2486/changes#r2539237525 from previous PR, do we want to add daemon resources to NodeClaim requests?~~

~~I think it makes sense to add it the minimum set of resources from the daemon sets but I don't feel strongly either way. If we do choose to do it this way,~~

Should we move this elsewhere - IMO the override of `ToNodeClaim` is a little indirect and might be tricky while debugging, should it instead be in `FinalizeScheduling`?

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -322,65 +340,98 @@ type InstanceTypeFilterError struct {
...
 	podRequests corev1.ResourceList
-	// We capture daemonRequests since this contributes to the resources that are required to schedule to this NodePool
-	daemonRequests corev1.ResourceList
+	// Min and max daemon overhead bounds for error reporting.
+	minDaemonRequests corev1.ResourceList
+	maxDaemonRequests corev1.ResourceList
+}
+
+// Updates the min/max daemon overhead bounds for error reporting.
+func (e *InstanceTypeFilterError) trackDaemonOverhead(dr corev1.ResourceList) {
```
ryan-mist, line 462: Can we push this to `utils/resources/resources.go`? We have `MaxResources(resources ...v1.ResourceList)` and I think a `MinResources(resources ...v1.ResourceList)` would also make sense here

pkg/controllers/provisioning/scheduling/scheduler.go
ryan-mist: This is the main performance dip we see from this change - `scheduling.NewScheduler` a little more than doubles in CPU time, although the relative percent is small. The main cause of this is from this shared `getCompatibleDaemonPods` function

Before:
Image

After:
Image

ryan-mist (commented):
I've done some performance tests here and there doesn't seem to be any large performance degradations. Though, there is a simple optimization I've recommended in one of the threads. 

Additionally, we don't have an existing performance tests around daemonsets (in `test/suites/performance/`). Since it looks like this isn't a large performance hit, I don't feel too strongly about adding one but may be good to look into if you have the bandwidth for it.

Testing Scenario:
> Roughly the scale change is from `Count(Daemonsets) x Count(NodePools)` to `Count(Daemonsets) x Count(NodePools) x Count(Instance Types per NodePool)` - so I've tested with 8 Daemonset and 4 NodePools with varying instance type selection (but selects among most instance types). I used 100 replicas with host name spread and a max skew of 5.

I've only added one of the runs here, but I've run this a few times and the results are relatively similar:
Image

Flame Graph After Change:
Image

pkg/controllers/provisioning/scheduling/nodeclaim.go
ryan-mist: Another idea is to split up daemonResources into 2 parts - one is the set of dameonsets that apply to all instance types as the "baseline" and another is the set that are instance type specific (these are disjoint sets). It makes it easier to track minimum daemonset requirements as its pretty similar to what we have now (although technically it could be under the actually minimum). This would also have very minor performance improvements, it is a micro-optimization though

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -322,65 +340,98 @@ type InstanceTypeFilterError struct {
...
+	if len(e.maxDaemonRequests) == 0 {
+		e.minDaemonRequests = dr.DeepCopy()
+		e.maxDaemonRequests = dr.DeepCopy()
+		return
+	}
+	for rn, q := range dr {
+		if cur, ok := e.minDaemonRequests[rn]; !ok || q.Cmp(cur) < 0 {
+			e.minDaemonRequests[rn] = q.DeepCopy()
```
ryan-mist: We should align the min Daemon Requests in `trackDaemonOverhead` and `ToNodeClaim`. The result in trackDaemonOverhead could include resources that don't apply to each instance type which doesn't seem to be the case in `ToNodeClaim`

pkg/controllers/provisioning/scheduling/nodeclaim.go
GaneshBannur: I've moved it to `FinalizeScheduling` as you suggested. One worry I had is that the list of instance types is not final in `FinalizeScheduling`. `TruncateInstanceTypes` and `RemoveInstanceTypeOptionsByPriceAndMinValues` are both called on the list of instance types after `FinalizeScheduling`. But both of these remove instance types on the expensive end of the list. So the instance types that are removed *should* generally be higher capacity instance types with larger DaemonSet requests. The minimum DaemonSet request that we are adding *should* still be the same. There is an edge case (when an expensive instance type has the min DaemonSet request) where we would be underestimating the DaemonSet request. But overall keeping it in `FinalizeScheduling` seems fine.

pkg/controllers/provisioning/scheduling/nodeclaim.go
GaneshBannur: Both of them now use the `MinResources` function in `pkg/utils/resources/resources.go` which follows the behaviour `ToNodeClaim` had.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -393,17 +440,33 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
-		daemonRequests: daemonRequests,
+		requirements: requirements,
+		podRequests:  podRequests,
 	}
 	remaining := cloudprovider.InstanceTypes{}
 
 	for _, it := range instanceTypes {
+		hostPortUsageForInstanceType := lo.ValueOr(hostPortUsage, it.Name, &scheduling.HostPortUsage{})
+		// exposed host ports on the pod
+		hostPorts := scheduling.GetHostPorts(pod)
```
ryan-mist: nit: we can move this outside of the instance type loop

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -393,17 +440,33 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
-		requirements:   requirements,
-		podRequests:    podRequests,
-		daemonRequests: daemonRequests,
+		requirements: requirements,
+		podRequests:  podRequests,
 	}
 	remaining := cloudprovider.InstanceTypes{}
 
 	for _, it := range instanceTypes {
+		hostPortUsageForInstanceType := lo.ValueOr(hostPortUsage, it.Name, &scheduling.HostPortUsage{})
```
ryan-mist: nit: we don't need to allocate `scheduling.HostPortUsage{}` when the key doesn't exist
e.g.
```
hostPortUsageForInstanceType, ok := hostPortUsage[it.Name]
if ok {
    if err := hostPortUsageForInstanceType.Conflicts(pod, hostPorts); err != nil {
        continue
    }
}
```

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -393,17 +440,33 @@ func filterInstanceTypesByRequirements(instanceTypes []*cloudprovider.InstanceTy
...
+		hostPorts := scheduling.GetHostPorts(pod)
+		if err := hostPortUsageForInstanceType.Conflicts(pod, hostPorts); err != nil {
+			continue
+		}
+
+		var totalRequestsForInstanceType corev1.ResourceList
+		daemonRequestsForInstanceType := daemonRequests[it.Name]
+		if len(daemonRequestsForInstanceType) != 0 {
+			err.trackDaemonOverhead(daemonRequestsForInstanceType)
+			totalRequestsForInstanceType = resources.MergeInto(totalRequestsForInstanceType, totalRequests)
```
ryan-mist, line 569: I'll have to do another perf test - I didn't see before that this will also scale allocations with the number of pods (I only used 100 before so the difference there might have been negligible)

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -179,11 +174,15 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements
...
 	n.InstanceTypeOptions = instanceTypes
+	// Daemon overhead is excluded here to avoid double-counting
 	n.Spec.Resources.Requests = resources.Merge(n.Spec.Resources.Requests, podData.Requests)
 	n.Requirements = nodeClaimRequirements
 	n.topology.Register(corev1.LabelHostname, n.hostname)
 	n.topology.Record(pod, n.Spec.Taints, nodeClaimRequirements, scheduling.AllowUndefinedWellKnownLabels)
-	n.hostPortUsage.Add(pod, scheduling.GetHostPorts(pod))
+	hostPorts := scheduling.GetHostPorts(pod)
+	for _, it := range instanceTypes {
+		n.hostPortUsage[it.Name].Add(pod, hostPorts)
```
ryan-mist: It doesn't look like this is safe to mutate - multiple NodeClaims created from the same NCT share the same underlying HostPortUsage struct. 

e.g. When a pod is added to nodeclaim A, the mutation is visible to nodeclaim B. This will incorrectly reject pods due to host port conflicts from pods that we don't intended to be scheduled on that node.

---

It looks like its the same behavior we have today, but I think it makes sense to fix it in this PR as it also can cause over-provisioning.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -322,65 +342,92 @@ type InstanceTypeFilterError struct {
...
-	daemonRequests corev1.ResourceList
+	// We capture the min and max daemonRequests since this contributes to the resources that are required to schedule to this NodePool. The min/max DaemonRequests are the min / max across all instance types.
+	minDaemonRequests corev1.ResourceList
+	maxDaemonRequests corev1.ResourceList
+}
+
+// Updates the min/max daemon overhead bounds for error reporting.
+func (e *InstanceTypeFilterError) trackDaemonOverhead(dr corev1.ResourceList) {
+	if len(e.maxDaemonRequests) == 0 {
+		e.minDaemonRequests = dr.DeepCopy()
```
ryan-mist: Do we need the `.DeepCopy()` here?

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -257,6 +256,23 @@ func (n *NodeClaim) offeringsToReserve(
 	return reservedOfferings, nil
 }
 
+// Add min daemon resource requests (min across instance types)
+func (n *NodeClaim) addDaemonRequests() {
+	var minDaemonOverhead corev1.ResourceList
+	for _, it := range n.InstanceTypeOptions {
+		if dr, ok := n.daemonResources[it.Name]; ok {
+			if len(minDaemonOverhead) == 0 {
+				minDaemonOverhead = dr.DeepCopy()
```
ryan-mist: Do we need the `.DeepCopy()` here?

pkg/controllers/provisioning/scheduling/nodeclaim.go
GaneshBannur: Will it make sense in the same PR? I get that it causes us to incorrectly track more ports than required (so in a way we overestimate the post usage similar to how we overestimate daemon overhead) but other than that it seems unrelated. Should I create a new PR and link it here?

pkg/utils/resources/resources.go
```diff
@@ -130,6 +130,27 @@ func MaxResources(resources ...v1.ResourceList) v1.ResourceList {
...
+func MinResources(resources ...v1.ResourceList) v1.ResourceList {
+	if len(resources) == 0 {
+		return v1.ResourceList{}
+	}
+	resourceList := resources[0].DeepCopy()
+	for _, rl := range resources[1:] {
+		for resourceName, quantity := range resourceList {
+			if rlQuantity, exists := rl[resourceName]; exists {
+				if rlQuantity.Cmp(quantity) < 0 {
+					resourceList[resourceName] = rlQuantity.DeepCopy()
```
GaneshBannur: Looking at this made me think about the behaviour in `MaxResources`. It doesn't deep copy the quantity
https://github.com/kubernetes-sigs/karpenter/blob/0ec1a0068bad191f2201cd7cd9bac34cbf857528/pkg/utils/resources/resources.go#L126
but technically it should right, since quantity contains `infDecAmount` which contains a pointer?
The implicit assumption seems to be that both `MaxResources` and `MinResources` return new lists without references to the inputs.
This doesn't really affect behaviour since `infDecAmount` is almost always `nil` but I thought I'd clarify.

pkg/controllers/provisioning/scheduling/nodeclaim.go
ryan-mist: Either way is fine with me. If we're going to do it in a separate PR maybe we should leave a TODO note in here somewhere

pkg/utils/resources/resources.go
ryan-mist: Maybe its worth adding a comment about how its not safe to mutate? I don't think we should add a DeepCopy unnecessarily though, since as you called out its not mutated today

pkg/controllers/provisioning/scheduling/nodeclaim.go
GaneshBannur: Created a new PR just so we can point to a separate PR where this specific behaviour was changed: https://github.com/kubernetes-sigs/karpenter/pull/3022
I'll update the code in this PR too.

pkg/controllers/provisioning/scheduling/nodeclaim.go
GaneshBannur: I profiled for 120s. The testing scenario is as follows:
- 6 NodePools (relatively high number of NodeClaim templates) with broad requirements (lots of instance types per NodePool ~500)
- 12 DaemonSets (relatively high number of DaemonSets) with 6 different affinity patterns (relatively high number of instance type groups) — each pattern matches a different subset of instance families, creating groups of ~50-150 instance types each
- Each DaemonSet requests cpu, memory, and ephemeral-storage
- Workload requests cpu, memory, and ephemeral-storage

The results below are only for `filterInstanceTypesByRequirements` since out of the functions changed, it is the one which takes most CPU time by far.

#### Scaling results

perf-comp

#### Flame graphs when tested with 10,000 replicas

##### Before PR

before-pr

##### After PR

after-pr

ryan-mist: How long are you profiling for? 1.8s seems pretty small for total CPU time

Can you also describe the testing setup - NodePools, NodeClasses, DaemonSets, etc?

GaneshBannur: I've updated the comment with new graphs and a description of the profiling scenario.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -869,50 +868,93 @@ func parallelizeUntil(workers, pieces int, doWorkPiece func(int) bool) {
...
+				groups[key] = &InstanceTypeGroup{
+					InstanceTypes: []*cloudprovider.InstanceType{it},
+					DaemonPods:    compatible,
+				}
 			}
-			return isDaemonPodCompatible(nct, p)
-		})...)
+		}
+		result := lo.Map(lo.Values(groups), func(g *InstanceTypeGroup, _ int) InstanceTypeGroup { return *g })
+		computeDaemonOverhead(result)
```
ryan-mist: nit: should we inline (or maybe combine) `computeDaemonOverhead` and `computeDaemonHostPortUsage`?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -869,50 +868,93 @@ func parallelizeUntil(workers, pieces int, doWorkPiece func(int) bool) {
...
 }
 
-// getDaemonOverhead determines the overhead for each NodeClaimTemplate required for daemons to schedule for any node provisioned by the NodeClaimTemplate
-func getDaemonOverhead(ctx context.Context, nodeClaimTemplates []*NodeClaimTemplate, daemonSetPods []*corev1.Pod) map[*NodeClaimTemplate]corev1.ResourceList {
-	return lo.SliceToMap(nodeClaimTemplates, func(nct *NodeClaimTemplate) (*NodeClaimTemplate, corev1.ResourceList) {
-		return nct, resources.RequestsForPods(lo.Filter(daemonSetPods, func(p *corev1.Pod, _ int) bool {
-			// Exclude daemon pods with DRA requirements when IgnoreDRARequests is enabled
-			if pod.HasDRARequirements(p) && karpopts.FromContext(ctx).IgnoreDRARequests {
-				return false
+type InstanceTypeGroup struct {
```
ryan-mist: nit: maybe `DaemonOverheadGroup`?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -869,50 +868,93 @@ func parallelizeUntil(workers, pieces int, doWorkPiece func(int) bool) {
...
+	InstanceTypes  []*cloudprovider.InstanceType
+	DaemonPods     []*corev1.Pod
```
ryan-mist: Is `DaemonPods` used outside of `buildInstanceTypeGroups`? We may be able to drop it

ryan-mist (commented):
A few small comments, besides those its LGTM

ryan-mist (approved):
/lgtm

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -208,11 +210,15 @@ func (n *NodeClaim) Add(pod *corev1.Pod, podData *PodData, nodeClaimRequirements
 	// Update node
 	n.Pods = append(n.Pods, pod)
 	n.InstanceTypeOptions = instanceTypes
+	// Daemon overhead is excluded here to avoid double-counting
```
jmdeal, line 251: Can we add an explicit test to ensure this doesn't regress?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -869,50 +868,77 @@ func parallelizeUntil(workers, pieces int, doWorkPiece func(int) bool) {
...
+	return lo.SliceToMap(nodeClaimTemplates, func(nct *NodeClaimTemplate) (*NodeClaimTemplate, []DaemonOverheadGroup) {
+		groups := map[string]*DaemonOverheadGroup{}
+		for _, it := range nct.InstanceTypeOptions {
+			compatible := lo.Filter(daemonSetPods, func(p *corev1.Pod, _ int) bool {
+				if pod.HasDRARequirements(p) && karpopts.FromContext(ctx).IgnoreDRARequests {
+					return false
+				}
+				return isDaemonPodCompatible(nct, it, p)
+			})
+			key := podSetKey(compatible)
```
jmdeal, line 982: We're using UIDs to compute this key, but not all daemonset pods consumed here necessarily have a UID set. We construct the set of pods here: https://github.com/kubernetes-sigs/karpenter/blob/22e837a28ab70e3e8480dd9d9adf14bfffd4f324/pkg/controllers/provisioning/provisioner.go#L490-L493

When we don't have a real pod cached, we fall back and generate a pod based on the DS definition. In this case we currently don't set a UID which could result in key collisions for separate sets of DaemonSets. We'll want to address this. I'd also ensure we have a test to exercise this case.

pkg/utils/resources/resources.go
```diff
@@ -130,6 +131,35 @@ func MaxResources(resources ...v1.ResourceList) v1.ResourceList {
 	return resourceList
 }
 
+// MinResources returns the minimum quantities for resources present in all lists (intersection of resources)
+// Ex: MinResources({cpu: 1}, {cpu: 2, gpu: 1}) = {cpu: 1}.
+// Quantity structs in the list returned are only safe to mutate if Quantity.d.Dec == nil
+func MinResources(resources ...v1.ResourceList) v1.ResourceList {
```
jmdeal, line 145: Let's add some unit tests for this.

pkg/controllers/provisioning/scheduling/scheduler.go
GaneshBannur: I changed `PodForDaemonSet` to assign a UUID to every pod it creates so we shouldn't run into this anymore. Also added a unit test for it.

pkg/utils/resources/resources.go
GaneshBannur: Also added for `MaxResources` since it was untested before

pkg/controllers/provisioning/scheduling/scheduler.go
GaneshBannur: Added an e2e test for it which verifies the fix (fails before fix, succeeds after).

GaneshBannur: @ryan-mist also wanted to make sure that no downstream code relied on uncached DaemonSets having an empty UID. I checked all the references to `UID` and as far as I can tell no downstream code relies on it.

GaneshBannur: Since this changes the instance type grouping which is what `filterInstanceTypesByRequirements` iterates over, I re-ran the performance tests. The testing scenario is similar to the one before except I used KWOK nodes rather than the real EC2 nodes I used for the previous test (I was having issues with EC2 limits). Note: we can't compare these results with the previous ones since we switched to KWOK.
Here is the updated testing scenario with KWOK (profiled for 120s):
- 6 NodePools (relatively high number of NodeClaim templates) with broad requirements (lots of instance types per NodePool ~36-72)
- 12 DaemonSets (relatively high number of DaemonSets) with 6 different affinity patterns (relatively high number of instance type groups). Each pattern matches a different subset of instance families, creating groups of ~6-24 instance types each
- Each DaemonSet requests cpu, memory, and ephemeral-storage
- Workload requests cpu, memory, and ephemeral-storage

Again, the results below are only for filterInstanceTypesByRequirements since out of the functions changed, it is the one which takes most CPU time by far.

**Scaling results**
daemonset-scaling-time-after-fix

I also have results for the percent of total CPU time taken up by `filterInstanceTypesByRequirements`. It increases after the PR when we scale to 10,000 pods but not sure if that is a noisy result.

daemonset-scaling-percent-after-fix

**Flame graphs when tested with 10,000 replicas**

**Before PR**

10k-kwok-before

**After PR**

10k-kwok-pr

GaneshBannur: @jmdeal identified another issue. Setting `UID` is not enough, `HostPortUsage.Add` uses the name and namespace of the pod as the key. These were not being set for uncached DaemonSet pods and the host port requests collided since the keys would be empty. Fixed this by setting name and namespace and using that as the key for constructing instance type groups (since this is the canonical K8s way to get keys for an object). Also verified the E2E test still works as expected.

PR #2977, opened 2026-04-17
feat: allow setting a min consolidation savings percent

grosser:
**Description**

our node churn is high because karpenter consolidates on any savings
but I want a lever to make the cluster more stable

adding `CONSOLIDATION_MIN_SAVINGS_PERCENT=10` so if we are not saving 10% of the cost then don't consolidate


jmdeal:
I'd take a look at this RFC, this is the way we plan on addressing this holistically moving forward: https://github.com/kubernetes-sigs/karpenter/pull/2942. I believe it's addressed in the RFC, but min savings can be a problematic lever particularly in highly heterogeneous clusters.

PR #2980, opened 2026-04-21
fix: re-sync node labels after registration hooks complete

AndrewMitchell25:
Fixes #N/A 

**Description**
In [this integration test run](https://github.com/aws/karpenter-provider-aws/actions/runs/24738491539/job/72371644543 ), the `[It] should launch an instance into a partition placement group with auto-assigned partition` test failed

I gave the logs to Kiro and this is what it came up with:

```
Confirmed. Here's the exact bug:

Line 88: r.syncNode(nodeClaim, node) — copies nodeClaim.Labels (with partition="") to node.Labels
Line 90: r.checkRegistrationHooks(ctx, nodeClaim) — hook updates nodeClaim.Labels[partition] = "3" but node.Labels still has partition=""
Line 91-94: If hooks pass, the unregistered taint is removed and NodeRegisteredLabelKey is set on the node
Line 96-103: Node is patched — but with the stale empty partition label
Line 108: ConditionTypeRegistered is set to True on the nodeClaim
Next reconcile: line 57 short-circuits because ConditionTypeRegistered is no longer Unknown
The node ends up with placement-group-partition: "" forever. The test then reads node.Labels[v1.LabelPlacementGroupPartition] and gets an empty string, failing the Expect(partitionStr).ToNot(BeEmpty()) assertion.

The fix: after hooks complete successfully, re-sync the node labels from the updated nodeClaim. The simplest approach is to call syncNode again after the hooks, or just re-assign the labels after hooks pass.
```

Putting up this PR so @joshuakguo can take a look

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/registration.go
+++ b/pkg/controllers/nodeclaim/lifecycle/registration.go
@@ -91,8 +91,8 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
 	// If any hook is not ready, registration is deferred and the unregistered taint remains.
 	hooksResult, hookErrors := r.checkRegistrationHooks(ctx, nodeClaim)
 	if lo.IsEmpty(hooksResult) && hookErrors == nil {
-		// Re-sync labels after hooks complete since hooks may have mutated nodeClaim labels.
-		node.Labels = lo.Assign(node.Labels, nodeClaim.Labels)
+		// Re-sync the node after hooks complete since hooks may have mutated the nodeClaim.
+		r.syncNode(nodeClaim, node)
 		node.Spec.Taints = lo.Reject(node.Spec.Taints, func(t corev1.Taint, _ int) bool {
 			return t.MatchTaint(&v1.UnregisteredNoExecuteTaint)
 		})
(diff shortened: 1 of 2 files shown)


joshuakguo:
Might be good to not include provider specific language in the comment but otherwise LGTM

jmdeal:
Can you add a test to ensure that labels added by a registration hook are added to both the node and the nodeclaim?

pkg/controllers/nodeclaim/lifecycle/registration.go
```diff
@@ -91,6 +91,8 @@ func (r *Registration) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (
 	// If any hook is not ready, registration is deferred and the unregistered taint remains.
 	hooksResult, hookErrors := r.checkRegistrationHooks(ctx, nodeClaim)
 	if lo.IsEmpty(hooksResult) && hookErrors == nil {
+		// Re-sync labels after hooks complete since hooks may have mutated nodeClaim labels.
+		node.Labels = lo.Assign(node.Labels, nodeClaim.Labels)
```
jmdeal: Let's use `syncNode` again here to support registration hooks that update annotations.

AndrewMitchell25: Will update

PR #2982, opened 2026-04-24
chore: Fix do-not-disrupt grace period integ test race

AndrewMitchell25:
Fixes #N/A 

**Description**
In [this integ test run](https://github.com/aws/karpenter-provider-aws/actions/runs/24872115086/job/72820655463), the hardcoded wait times for the pods caused the test to fail. This PR updates the wait times to dynamically calculate based on the test pods status.StartTime, which is how the do-not-disrupt grace period is actually calculated.

Ran the integ test muultiple times locally against a cluster, saw logs like these demonstrating new dynamically calculated wait times -
```
STEP: expecting 3 pods to be live for 51.195169782s
STEP: expecting 2 pods to be live for 53.9979467s
STEP: expecting 1 pods to be live for 30s
```

### DIFF
--- a/test/suites/regression/integration_test.go
+++ b/test/suites/regression/integration_test.go
@@ -426,6 +426,8 @@ var _ = Describe("Integration", func() {
 	Describe("DoNotDisrupt", func() {
 		Context("Grace Period", func() {
 			It("should drain pods according to their grace period durations and block indefinitely for 'true'", func() {
+				const shortGrace = time.Minute
+				const longGrace = 2 * time.Minute
 				// Pod with a 1-minute grace period
 				shortPod := test.Pod(test.PodOptions{
 					ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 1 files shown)


test/suites/regression/integration_test.go
```diff
@@ -481,18 +483,32 @@ var _ = Describe("Integration", func() {
...
+				shortExpiry := shortPod.Status.StartTime.Add(shortGrace)
+				longExpiry := longPod.Status.StartTime.Add(longGrace)
+
 				// Delete the nodeclaim to trigger draining
 				env.ExpectDeleted(nodeClaim)
 
-				// During the first minute, all three pods should remain active
-				env.ConsistentlyExpectActivePods(45*time.Second, shortPod, longPod, indefinitePod)
+				// All three pods should remain active until the short grace period expires.
+				env.ConsistentlyExpectActivePods(max(time.Until(shortExpiry)-5*time.Second, 0), shortPod, longPod, indefinitePod)
```
DerekFrank: Should we actually max(x, 0) this? I think we'd rather have it fail loudly

AndrewMitchell25: Updated to expect that value to be > 0

DerekFrank (approved):
/lgtm

PR #2990, opened 2026-04-28
perf: improve skip condition for `applyPriceOverlays`

jesseanttila-cai:
**Description**

Replaces a `m != nil` check with a `len(m) != 0` check to avoid running `applyPriceOverlays` with an empty `priceUpdates` map. This reduced disruption controller CPU usage by ~50% for an idle cluster with NodeOverlays that only affect capacity.

The actual root causes of poor `applyPriceOverlays` performance are not addressed.

Profiling and metrics. 

### DIFF
--- a/pkg/controllers/nodeoverlay/store.go
+++ b/pkg/controllers/nodeoverlay/store.go
@@ -146,7 +146,7 @@ func (s *internalInstanceTypeStore) apply(nodePoolName string, it *cloudprovider
 	}
 
 	// Handle offerings - copy-on-write only for offerings that need price overlay
-	if instanceTypeUpdate.Price != nil {
+	if len(instanceTypeUpdate.Price) != 0 {
 		overriddenInstanceType.Offerings = s.applyPriceOverlays(it.Offerings, instanceTypeUpdate.Price)
 	} else {
 		overriddenInstanceType.Offerings = it.Offerings // Shared - not modified
--- a/pkg/controllers/nodeoverlay/store.go
+++ b/pkg/controllers/nodeoverlay/store.go
@@ -73,9 +73,7 @@ func (s *InstanceTypeStore) ApplyAll(nodePoolName string, its []*cloudprovider.I
 	}
 
 	for _, it := range its {
-		if updatedIt, err := internalStore.apply(nodePoolName, it); err == nil {
-			result = append(result, updatedIt)
-		}
+		result = append(result, internalStore.apply(nodePoolName, it))
 	}
 	return result, nil
 }
@@ -87,11 +85,7 @@ func (s *InstanceTypeStore) Apply(nodePoolName string, it *cloudprovider.Instanc
 		return &cloudprovider.InstanceType{}, cloudprovider.NewUnevaluatedNodePoolError(nodePoolName)
 	}
 
-	updatedIt, err := internalStore.apply(nodePoolName, it)
-	if err != nil {
-		return &cloudprovider.InstanceType{}, err
-	}
-	return updatedIt, nil
+	return internalStore.apply(nodePoolName, it), nil
 }
 
 // InstanceTypeStore manages instance type updates for node pools.
@@ -120,14 +114,14 @@ func newInternalInstanceTypeStore() *internalInstanceTypeStore {
 // - Shared: Requirements and Overhead (never modified, safe to share)
 // - Selective copy: Offerings (only copied if price overlay applied)
 // - Selective copy: Capacity (only copied if capacity overlay applied)
-func (s *internalInstanceTypeStore) apply(nodePoolName string, it *cloudprovider.InstanceType) (*cloudprovider.InstanceType, error) {
+func (s *internalInstanceTypeStore) apply(nodePoolName string, it *cloudprovider.InstanceType) *cloudprovider.InstanceType {
 	instanceTypeList, ok := s.updates[nodePoolName]
 	if !ok {
-		return it, nil
+		return it
 	}
 	instanceTypeUpdate, ok := instanceTypeList[it.Name]
 	if !ok {
-		return it, nil
+		return it
 	}
 
 	// Create a shallow copy of the instance type, sharing immutable fields
@@ -151,7 +145,7 @@ func (s *internalInstanceTypeStore) apply(nodePoolName string, it *cloudprovider
 		overriddenInstanceType.Offerings = it.Offerings // Shared - not modified
 	}
 
-	return overriddenInstanceType, nil
+	return overriddenInstanceType
 }
 
 // applyPriceOverlays creates a new offerings slice with selective copying:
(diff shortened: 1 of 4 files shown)


jigisha620:
Could you share the profiling data or before/after metrics that show the ~50% CPU reduction? A flame graph or CPU usage graph would also help here. I am also curious about your test setup, how many instance types/offerings/capacity-only NodeOverlays were involved?

jesseanttila-cai:
@jigisha620 here's the flamegraph (from karpenter-provider-aws) where more than half of the disruption controller reconcile loop is spent in `applyPriceOverlays`:

image


The test setup involves two nodepools and eight mutually exclusive nodeoverlays. Each overlay applies a "bandwidth" capacity value based on the `instance-network-bandwidth` label provided by karpenter-provider-aws. The number of instance types is probably a bit over a thousand, with maybe slightly less than five thousand offerings in total.

Since none of the overlays modify the price, the price update map is empty. Even then `applyPriceOverlays` still checks every offering, which is very heavy due to the need to stringify the set of requirements for each offering.

</br>

I've also looked into a couple more optimization opportunities in another branch [here](https://github.com/jesseanttila-cai/karpenter/commits/overlay-optimization-2/), including caching the entire modified InstanceType for each overlay update. This works well for capacity updates, but there doesn't seem to be an easy way to invalidate the cache when offering prices/availabilities change, so it seems to be necessary to run `applyPriceOverlays` each time regardless. The rest of the optimizations are mainly removing unnecessary allocations, e.g. dropping the `lo.Keys` in  `len(lo.Keys(instanceTypeUpdate.Capacity.OverlayUpdate)) != 0`, which also helps reduce GC overhead.

jigisha620:
Thanks for the flame graph and detailed explanation! The graph clearly shows the impact. The fix makes sense, no point entering applyPriceOverlays when there's nothing to apply.

One ask, could you update` BenchmarkStoreApplyCapacityOnly` in `store_bench_test.go` to use `Price: map[string]*priceUpdate{}` instead of `Price: nil`? The current benchmark already skips `applyPriceOverlays` since nil passes both the old and new check, so it doesn't reflect the real code path (which always initializes an empty map for capacity-only overlays). With that fix, running make benchmark on main vs your branch would show the actual gain and guard against regression.

Could you share the before/after benchmark output?

jesseanttila-cai:
I added two more optimizations. The first one removes the deep copy when applying capacity overlays, so only the ResourceList container is copied. The second one lifts the evaluatedNodePools check out of `apply` to avoid it being checked in every iteration when using `ApplyAll`.

The benchmarks for the baseline and the three optimizations (cumulative) are below. Scenarios involving price overlays were largely unaffected, but the rest show improvements of 10x or better.

```
eca1dcb57316ae5742512447747467b4980db714 - "use empty price update map instead of nil in benchmark"

BenchmarkStoreApply/selective-copy-apply-20         2719            382829 ns/op          286642 B/op       6318 allocs/op
BenchmarkStoreApplyNoOverlay-20                   199999              6065 ns/op               0 B/op          0 allocs/op
BenchmarkStoreApplyPriceOnly-20                     1388            833130 ns/op          461089 B/op      15400 allocs/op
BenchmarkStoreApplyCapacityOnly-20                  1310            881344 ns/op          688430 B/op      15200 allocs/op
BenchmarkStoreApplyBothOverlays-20                  1197            972088 ns/op          736471 B/op      16200 allocs/op
BenchmarkStoreApplyAllScenario-20                    230           5167989 ns/op         3959610 B/op      86114 allocs/op
BenchmarkNodeOverlayControllerScenario-20          12048             99820 ns/op           86400 B/op       1080 allocs/op

bc0f34db1a1a5269bb6ae7b76204d50f1371a15d - "skip empty price update map"

BenchmarkStoreApply/selective-copy-apply-20         2852            385414 ns/op          286649 B/op       6318 allocs/op
BenchmarkStoreApplyNoOverlay-20                   197562              6015 ns/op               0 B/op          0 allocs/op
BenchmarkStoreApplyPriceOnly-20                     1290            839346 ns/op          461093 B/op      15400 allocs/op
BenchmarkStoreApplyCapacityOnly-20                  9076            131874 ns/op          294400 B/op       1000 allocs/op
BenchmarkStoreApplyBothOverlays-20                  1196            978316 ns/op          736479 B/op      16201 allocs/op
BenchmarkStoreApplyAllScenario-20                    231           5257659 ns/op         3959671 B/op      86114 allocs/op
BenchmarkNodeOverlayControllerScenario-20          12070             99595 ns/op           86400 B/op       1080 allocs/op

d3507485b75e74e1d93581ddaf59fa9e4adef531 - "remove unnecessary deep copy"

BenchmarkStoreApply/selective-copy-apply-20         2792            358842 ns/op          230870 B/op       6156 allocs/op
BenchmarkStoreApplyNoOverlay-20                   197121              5986 ns/op               0 B/op          0 allocs/op
BenchmarkStoreApplyPriceOnly-20                     1370            826343 ns/op          461089 B/op      15400 allocs/op
BenchmarkStoreApplyCapacityOnly-20                 16170             73274 ns/op          156800 B/op        600 allocs/op
BenchmarkStoreApplyBothOverlays-20                  1294            917320 ns/op          598773 B/op      15800 allocs/op
BenchmarkStoreApplyAllScenario-20                    248           4819944 ns/op         3236825 B/op      84013 allocs/op
BenchmarkNodeOverlayControllerScenario-20          12162             99297 ns/op           86400 B/op       1080 allocs/op

d32ccf6d472df07e99a8eae5b14f0800dc26fcc1 - "lift evaluatedNodePools check out of apply"

BenchmarkStoreApply/selective-copy-apply-20         3000            351934 ns/op          230873 B/op       6156 allocs/op
BenchmarkStoreApplyNoOverlay-20                  3156326               375.5 ns/op             0 B/op          0 allocs/op
BenchmarkStoreApplyPriceOnly-20                     1394            827498 ns/op          461090 B/op      15400 allocs/op
BenchmarkStoreApplyCapacityOnly-20                 17841             66339 ns/op          156800 B/op        600 allocs/op
BenchmarkStoreApplyBothOverlays-20                  1267            901605 ns/op          598780 B/op      15800 allocs/op
BenchmarkStoreApplyAllScenario-20                    253           4720681 ns/op         3152361 B/op      82958 allocs/op
BenchmarkNodeOverlayControllerScenario-20         157286              7453 ns/op               0 B/op          0 allocs/op
```

pkg/controllers/nodeoverlay/store.go
```diff
@@ -115,44 +113,39 @@ func newInternalInstanceTypeStore() *internalInstanceTypeStore {
...
 		Requirements: it.Requirements, // Shared - never modified
 		Overhead:     it.Overhead,     // Shared - never modified
+		Capacity:     it.Capacity,
 	}
 
 	// Handle capacity overlay - only deep copy if we're modifying it
-	if len(lo.Keys(instanceTypeUpdate.Capacity.OverlayUpdate)) != 0 {
-		overriddenInstanceType.Capacity = it.Capacity.DeepCopy()
+	if len(instanceTypeUpdate.Capacity.OverlayUpdate) != 0 {
+		// This method replaces overriddenInstanceType.Capacity with a shallow copy
```
jigisha620: nit: This comment is slightly confusing, it reads like ApplyCapacityOverlay does a shallow copy of the struct, but it actually replaces Capacity with a new map via lo.Assign. What do you think about something like `// ApplyCapacityOverlay replaces Capacity with a new merged map (original untouched)` ?

jigisha620 (commented):
Thanks for the benchmark results and the additional optimizations! The progression from each commit makes it easy to see the impact of each change. One minor nit otherwise LGTM.

jigisha620 (approved):
/lgtm

PR #2991, opened 2026-04-28
chore: switch from docker/docker to karpenter test package for kwok cloud provider

joshuakguo:
Fixes #N/A 

**Description**

the https://pkg.go.dev/github.com/docker/docker go package is now github.com/moby/moby and only the client and the api are public packages now, so when there are CVEs in the package, they will show up and we can't upgrade

so we move to use a different package instead

make presubmit

### DIFF
--- a/kwok/cloudprovider/cloudprovider.go
+++ b/kwok/cloudprovider/cloudprovider.go
@@ -27,7 +27,6 @@ import (
 
 	"github.com/awslabs/operatorpkg/serrors"
 	"github.com/awslabs/operatorpkg/status"
-	"github.com/docker/docker/pkg/namesgenerator"
 	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


kwok/cloudprovider/cloudprovider.go
```diff
@@ -183,7 +183,7 @@ func (c CloudProvider) getInstanceType(instanceTypeName string) (*cloudprovider.
 }
 
 func (c CloudProvider) toNode(nodeClaim *v1.NodeClaim) (*corev1.Node, error) {
-	newName := strings.ReplaceAll(namesgenerator.GetRandomName(0), "_", "-")
+	newName := names.SimpleNameGenerator.GenerateName("kwok-")
```
DerekFrank: Don't we have test.RandomName() somewhere?

joshuakguo: yeah we can use this one instead https://github.com/kubernetes-sigs/karpenter/blob/da2a854455cb78705dbc501cb3415d83e95930c3/pkg/test/metadata.go#L41

kwok/cloudprovider/cloudprovider.go
```diff
@@ -183,7 +183,7 @@ func (c CloudProvider) getInstanceType(instanceTypeName string) (*cloudprovider.
 }
 
 func (c CloudProvider) toNode(nodeClaim *v1.NodeClaim) (*corev1.Node, error) {
-	newName := strings.ReplaceAll(namesgenerator.GetRandomName(0), "_", "-")
+	newName := "kwok-" + test.RandomName()
```
DerekFrank: Sorry one more change, should we add something in from the nodeclaim to ensure uniqueness?

joshuakguo: just add the nodeclaim name? is that enough? or hash?

joshuakguo: something like `fmt.Sprintf("kwok-%s-%s", nodeClaim.Name, test.RandomName())`?

DerekFrank (approved):
/lgtm

PR #2994, opened 2026-05-01
perf: integrate multi-iteration benchmarks into PR review flow

### LINKED ISSUE #2997: Performance regressions can reach users undetected due to single-run benchmarks
## Problem

PR #2671 (topology domain filtering) introduced a 1.95x consolidation performance regression that reached customers as 10-20x CPU spikes (issue #2954) before being identified and reverted (#2957). The regression was not caught before merge because:

1. **Benchmarks only run after reviewer approval**, not before. By the time results are available, the PR is already approved or merged.
2. **No automated baseline comparison.** Results are uploaded as artifacts but never compared against a known-good baseline. A reviewer would need to manually download and compare.
3. **Each test runs once.** The consolidation benchmark has a coefficient of variation of 40-60% on kind clusters. A single run has only ~60% probability of detecting a 2x regression — 25% of known-bad runs in our positive control experiment fell *below* the control median.

## Evidence

A controlled experiment with 80+ benchmark runs across the PR #2671 regression confirmed:

| Sample size | Detection power (2x regression) |
|---|---|
| n=3 | ~45% (t-test underpowered) |
| n=10 | ~64% |
| n=20 | ~95% |

The high variance comes from kind cluster scheduling jitter, resource contention, and GC pauses in the controller. Individual runs of Hostname Spread Consolidation ranged from 194s to 1,828s on the *same* code.

## Impact

- Customer-facing CPU spikes (10-20x) in Karpenter v1.11.0
- Emergency revert required (#2957)
- Loss of the topology domain filtering feature until a regression-safe implementation is developed

## Desired Outcome

- High-risk PRs (large feat/fix changes) should trigger benchmarks automatically before approval
- Each benchmark should run multiple iterations to produce statistically meaningful results
- Results should be compared against a rolling baseline with automated pass/fail judgment
- Maintainers should be able to skip benchmarks for known-safe changes (refactors, docs)

## References

- #2954 — High CPU usage regression in Karpenter v1.11.0
- #2671 — Filter topology domains by NodePool compatibility (the regression)
- #2957 — Revert #2671

nathangeology:
Fixes #2997

# perf: Integrate multi-iteration performance benchmarks into PR review flow

## Motivation

PR#2671 (topology domain filtering) caused a 1.95x consolidation performance
regression that reached customers as 10-20x CPU spikes (issue #2954) before
being identified and reverted. The regression was not caught before merge
because:

1. Performance tests only ran after reviewer approval, not before.
2. Results were not compared against a baseline automatically.
3. Each test ran once making it hard to confirm regressions statistically.

### `.github/workflows/kind-perf-e2e.yaml`

**New triggers:**

- `pull_request` (labeled/synchronize): auto-triggers on `size/XL` or
  `size/XXL` PRs with a `feat:` or `fix:` title prefix. These are the PRs
  most likely to introduce performance regressions.
- `perf-check` label: manual trigger for reviewers who suspect performance
- `ok-to-test` label: required for all pull_request triggers. Ensures
  benchmarks only run on PRs vetted by an org member.
  impact on PRs that don't meet the auto-trigger criteria.
- `skip-perf` label: allows maintainers to suppress the pre-approval run
  (e.g. large refactors, renames). Does **not** suppress post-merge runs.
- Push to `main` always runs (no skip path) to maintain the rolling baseline.

**Iterations:** All triggered runs now pass `repeat: 10` to the e2e workflow,
running each test suite 10 times per trigger.

### `.github/workflows/e2e.yaml`

**New input:** `repeat` (default: 1) — number of iterations to run the test.

**Iteration loop:** The run step loops `repeat` times, copying each
iteration's `*_performance_report.json` into `iter_N/` subdirectories under
`OUTPUT_DIR`.

**Aggregation step:** After all iterations, a Go program
(`hack/perf-aggregate`) computes per-test statistics (n, median, mean,
stddev, min, max) and writes `aggregated_summary.json`. Results are printed
as a human-readable table in the job log. The median is the primary metric
(robust to outliers).

**Artifact layout:** Artifacts now include all per-iteration reports and the
aggregated summary, enabling downstream baseline comparison.

## Trigger behavior summary

| Trigger | When | Iterations | Can skip? |
|---|---|---|---|
| `size/XL` or `size/XXL` + feat/fix PR | Automatic on label (requires ok-to-test) | 10 per test | Yes, with `skip-perf` |
| `perf-check` label | Manual by reviewer (requires ok-to-test) | 10 per test | N/A (opt-in) |
| `/karpenter perf` comment (ApprovalComment) | Manual by reviewer | 10 per test | Yes, with `skip-perf` |
| Merge to `main` | Automatic | 10 per test | No (always runs) |

## Cost estimate

~23 hours of GitHub Actions compute per triggered PR (7 tests × 10 iterations
× ~20 min each, running in parallel via matrix). Actual wall-clock time
depends on the `kubernetes-sigs` org's concurrency limits. This cost is
incurred only on the triggers above — documentation, CI, and test-only PRs
are unaffected.

## Regression detection

Each test runs 10 iterations. The aggregation program computes the batch
median and emits it as a `customSmallerIsBetter` metric for
[benchmark-action/github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark).

The action compares the current batch median against the previous commit's
batch median. If the median is 50% worse (configurable via `alert-threshold`),
the job fails and a comment is posted on the commit. Benchmark history is
stored in the GitHub Actions cache (per test, per OS), so no additional
branches or repo configuration are needed.

This approach has two layers of noise reduction:
1. The median of 10 runs absorbs per-run variance (CV of 40-60% on kind).
2. The action compares medians across commits, not individual runs.

A 150% threshold on batch medians is much tighter than 150% on single runs.
With the observed CV of ~48% on Hostname Spread Consolidation, a single run
can legitimately vary by 2x. A batch median varies much less.

The `range` field in the benchmark output carries the stddev, and the `extra`
field carries the full stats (mean, cv, min, max, n). These appear in the
GitHub Pages chart tooltips and PR comments for context.

**Future improvement:** If the project enables a `gh-pages` branch,
`benchmark-action` can auto-push results there and generate a time-series
chart dashboard at `https://kubernetes-sigs.github.io/karpenter/dev/bench/`.
This gives visual trend tracking across releases and makes it easy to spot
gradual performance drift. The switch is a one-line change (`auto-push: true`
+ creating the branch). The cache-based approach works without it.

This replaces the per-test hard-coded thresholds (e.g., the 350MB memory
assertion that flakes on every Hydra run). Multi-run statistics are robust
to single-run outliers and sensitive to sustained shifts. But per-test hard coded thresholds are not removed in this PR. I might include those in this PR or do a separate one for those based on feedback.

## Testing

Workflow syntax validated with `actionlint` (if available). The iteration
loop is straightforward shell; the Go aggregation program has no external
dependencies beyond the stdlib. Full end-to-end validation requires a GitHub
Actions run on the fork.

## References

- Issue #2954: High CPU usage regression in Karpenter v1.11.0
- PR#2671: Filter topology domains by NodePool compatibility (the regression)
- PR#2957: Revert PR#2671

## AI Disclosure

I used an LLM to parallel program with me on this. The original iteration
(multi-iteration benchmark integration) was drafted collaboratively. T

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.


.github/workflows/kind-perf-e2e.yaml
```diff
@@ -10,7 +12,18 @@ on:
...
+      (github.event_name == 'workflow_dispatch') ||
+      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success') ||
+      (github.event_name == 'pull_request' &&
+       contains(github.event.pull_request.labels.*.name, 'ok-to-test') &&
+       !contains(github.event.pull_request.labels.*.name, 'skip-perf') &&
+       (contains(github.event.pull_request.labels.*.name, 'perf-check') ||
+        ((contains(github.event.pull_request.labels.*.name, 'size/XL') ||
+          contains(github.event.pull_request.labels.*.name, 'size/XXL')) &&
+         (startsWith(github.event.pull_request.title, 'feat:') ||
+          startsWith(github.event.pull_request.title, 'fix:')))))
```
ryan-mist: Should we also include perf PRs - `startsWith(github.event.pull_request.title, 'perf:'`?

.github/scripts/perf-aggregate.py
```diff
@@ -0,0 +1,121 @@
+"""Aggregate multi-iteration benchmark results and emit benchmark-action JSON.
```
ryan-mist: Can we migrate this to go?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +76,62 @@ jobs:
...
+    - name: Compare against performance baseline
+      if: always() && env.OUTPUT_DIR != '' && inputs.repeat > 1
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        name: Karpenter Performance (${{ inputs.test_name || inputs.suite }})
+        tool: customSmallerIsBetter
+        output-file-path: ${{ env.OUTPUT_DIR }}/benchmark-results.json
+        external-data-json-path: ./cache/benchmark-data.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        # Alert if batch median is 50% worse than previous batch median.
```
ryan-mist, line 131: Should this be based on variance between the runs as well?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +76,62 @@ jobs:
...
+        python3 .github/scripts/perf-aggregate.py
+    - name: Download previous benchmark data
+      if: always() && env.OUTPUT_DIR != '' && inputs.repeat > 1
+      uses: actions/cache@v4
+      with:
+        path: ./cache
+        key: ${{ runner.os }}-perf-benchmark-${{ inputs.test_name || inputs.suite }}
+    - name: Compare against performance baseline
+      if: always() && env.OUTPUT_DIR != '' && inputs.repeat > 1
+      uses: benchmark-action/github-action-benchmark@v1
```
ryan-mist: nit: can we pin this to a specific sha?

e.g. with upload performance results we do
```
uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```

.github/workflows/e2e.yaml
```diff
@@ -71,19 +76,62 @@ jobs:
...
+          iter_dir="$OUTPUT_DIR/iter_${i}"
+          mkdir -p "$iter_dir"
+          # Move reports from this iteration into a per-iteration subdirectory
+          find "$OUTPUT_DIR" -maxdepth 1 -name "*_performance_report.json" -exec mv {} "$iter_dir/" \;
+        done
+    - name: Aggregate and compare against baseline
+      if: always() && env.OUTPUT_DIR != '' && inputs.repeat > 1
+      env:
+        ITERATIONS: ${{ inputs.repeat }}
+        IS_MAIN_PUSH: ${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
ryan-mist: nit: I think we can drop `IS_MAIN_PUSH` - it doesn't look like its used?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +76,62 @@ jobs:
...
+        name: Karpenter Performance (${{ inputs.test_name || inputs.suite }})
+        tool: customSmallerIsBetter
+        output-file-path: ${{ env.OUTPUT_DIR }}/benchmark-results.json
+        external-data-json-path: ./cache/benchmark-data.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        # Alert if batch median is 50% worse than previous batch median.
+        # At n=10 with observed CV ~48%, batch median CV is ~15%. A 130%
+        # threshold would yield ~7% false positives per test (~40% across
+        # the 7-test matrix). 150% keeps false alarm rate under 5% total.
+        alert-threshold: '150%'
```
ryan-mist, line 135: It looks like this will alarm when some of these metrics get better - e.g. with `total_reserved_cpu_utilization`, `total_reserved_memory_utilization`, and `resource_efficiency_score` if it increases 50% then we're performing better

.github/workflows/e2e.yaml
ryan-mist: Similar comment for `actions/cache@v4`

.github/workflows/e2e.yaml
nathangeology: This should be addressed in the latest version, good callout!

.github/workflows/e2e.yaml
nathangeology: This should be addressed in the latest commit, its definitely a good call out as well.

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+	"encoding/json"
+	"fmt"
+	"math"
+	"os"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// direction distinguishes benchmark metrics whose regression alarm fires on
```
nathangeology: This comment could be deleted

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+// an increase (Smaller) from those whose alarm fires on a decrease (Bigger).
```
nathangeology: Comment could be deleted

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+// metricSpec describes how to extract, label, and classify one field from a
+// performance report.
+type metricSpec struct {
+	jsonField string
+	display   string
+	unit      string
+	dir       direction
+}
+
+// metrics is the ordered list of report fields we aggregate. Order controls
```
nathangeology: all of these comments aren't needed

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+	for i := 1; i <= iterations; i++ {
+		pattern := filepath.Join(outputDir, fmt.Sprintf("iter_%d", i), "*_performance_report.json")
+		matches, err := filepath.Glob(pattern)
+		if err != nil {
+			return nil, fmt.Errorf("glob %s: %w", pattern, err)
+		}
+		for _, path := range matches {
+			data, err := readReport(path)
+			if err != nil {
+				// Skip unreadable / malformed reports; match prior Python behavior.
```
nathangeology: this comment not needed

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+			}
+			testKey := filepath.Base(path)
+			reportsByTest[testKey] = append(reportsByTest[testKey], data)
+		}
+	}
+	return reportsByTest, nil
+}
+
+func readReport(path string) (map[string]any, error) {
+	// path is the output of filepath.Glob under OUTPUT_DIR, which the CI
```
nathangeology: can remove this comment

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+		return nil, err
+	}
+	var m map[string]any
+	if err := json.Unmarshal(b, &m); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// extractValues pulls the numeric value at jsonField from each report,
```
nathangeology: probably don't need this comment

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,387 @@
...
+		CVPct:  round1(cv),
+		Min:    round2(sorted[0]),
+		Max:    round2(sorted[n-1]),
+	}
+}
+
+func round2(f float64) float64 { return math.Round(f*100) / 100 }
+func round1(f float64) float64 { return math.Round(f*10) / 10 }
+
+// prettifyTestName reverses the report-file naming convention so table rows
```
nathangeology: don't need this comment

nathangeology (commented):
Posting some "comment clean-up" comments for my bot to work through real quick

.github/scripts/perf-aggregate/main_test.go
```diff
@@ -0,0 +1,262 @@
...
+	"encoding/json"
+	"math"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+func TestComputeStats(t *testing.T) {
```
ryan-mist: nit: Can we use ginkgo for this to align with the rest of the project?

ryan-mist: Also, I'll defer to you on this, but we may be able to drop this file

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,370 @@
...
+
+// buildResults computes stats for every (test, metric) pair and appends
+// the corresponding gating and CV benchmark entries. It also returns the
+// summary map keyed by test then metric display name.
+func buildResults(testKeys []string, reportsByTest map[string][]map[string]any) (map[string]map[string]stats, benchmarkResults) {
+	summary := map[string]map[string]stats{}
+	var results benchmarkResults
+	for _, testKey := range testKeys {
+		datas := reportsByTest[testKey]
+		if len(datas) == 0 {
```
ryan-mist: IIUC, this is a check for missing data - should this be fail-closed instead?

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,370 @@
...
+	if err := run(outputDir, iterations, os.Stdout); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+// run performs the aggregation and returns nil on success. It is separated
+// from main so tests can drive it with synthetic input.
+func run(outputDir string, iterations int, out *os.File) error {
+	reportsByTest, err := collectReports(outputDir, iterations)
```
ryan-mist, line 138: Can we add a check after this to fail closed on missing data?

Maybe something like:
```
if len(reportsByTest) == 0 {
    return fmt.Errorf("no performance reports found under %s across %d iterations", outputDir, iterations)
}
for test, datas := range reportsByTest {
    if len(datas) < iterations {
        return fmt.Errorf("%s: got %d/%d iteration reports; refusing to gate on partial data", test, len(datas), iterations)
    }
}
```

.github/workflows/e2e.yaml
```diff
@@ -71,19 +72,122 @@ jobs:
...
+        name: Karpenter Performance (${{ inputs.test_name || inputs.suite }}) - smaller-is-better
+        tool: customSmallerIsBetter
+        output-file-path: ${{ env.OUTPUT_DIR }}/benchmark-results-smaller.json
+        external-data-json-path: ./cache-smaller/benchmark-data.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        # Alert if batch median is 50% worse than previous batch median.
+        # At n=10 with observed CV ~48%, batch median CV is ~15%. A 130%
+        # threshold would yield ~7% false positives per test (~40% across
+        # the 7-test matrix). 150% keeps false alarm rate under 5% total.
+        alert-threshold: '150%'
```
ryan-mist, line 150: Should we move this into `perf-aggregate` that way we can also gate using stddev here instead of the flat 150%?

.github/scripts/perf-aggregate/main_test.go
```diff
@@ -0,0 +1,262 @@
+/*
```
ryan-mist, line 1: nit: can we use ginkgo for this instead

.github/workflows/e2e.yaml
nathangeology: Did a per-key approach here, let me know what you think, seems to work when tested against the regression PR.

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,417 @@
...
+		}
+		if jsonField == "total_time" && v > 1e9 {
+			v = v / 1e9
+		}
+		values = append(values, v)
+	}
+	return values
+}
+
+func toFloat(v any) (float64, bool) {
```
ryan-mist: Can this be simplified?

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,417 @@
...
+	for _, d := range datas {
+		raw, ok := d[jsonField]
+		if !ok || raw == nil {
+			continue
+		}
+		v, ok := toFloat(raw)
+		if !ok {
+			continue
+		}
+		if jsonField == "total_time" && v > 1e9 {
```
ryan-mist, line 309: OOC, where does the `1e9` come from?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +72,159 @@ jobs:
...
+        external-data-json-path: ./cache-smaller-loose/benchmark-data.json
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+        # Loose tier: Controller CPU (per-test P90 CV 38%, max 44%). A flat
+        # 150% here fired analytically on ~5-27% of identical-code batches;
+        # 250% pushes analytical per-key FP under 1%. Still catches the
+        # PR#2671-shape 3x controller-CPU regression the customer saw
+        # (issue #2954), because ratio 3.0 > 2.5. Consolidation Rounds is
+        # excluded from both smaller-tier gate files (see main.go
+        # tierInformational) — it lives only in the CV file below.
+        alert-threshold: '250%'
```
ryan-mist, line 171: This is pretty high. Based on the comment it seems like this was chosen based on the p90 CV / max per-test - now that the success gate is p50 across 10 runs instead of 1, hopefully we can lower this

ryan-mist: Why does it need to be that high?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +72,159 @@ jobs:
...
+        fail-on-alert: true
+        comment-on-alert: true
+        alert-comment-cc-users: '@nathangeology'
+        summary-always: true
+    - name: Compare batch CV (informational only)
+      if: always() && env.OUTPUT_DIR != '' && inputs.repeat > 1
+      uses: benchmark-action/github-action-benchmark@52576c92bccf6ac60c8223ec7eb2565637cae9ba # v1.22.1
+      with:
+        name: Karpenter Performance (${{ inputs.test_name || inputs.suite }}) - batch CV
+        tool: customSmallerIsBetter
```
ryan-mist: what is customSmallerIsBetter?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +72,159 @@ jobs:
...
+          find "$OUTPUT_DIR" -maxdepth 1 -name "*_performance_report.json" -exec mv {} "$iter_dir/" \;
+        done
+    - name: Aggregate and compare against baseline
+      if: always() && env.OUTPUT_DIR != '' && inputs.repeat > 1
+      env:
+        ITERATIONS: ${{ inputs.repeat }}
+      shell: bash
+      run: |
+        go run ./.github/scripts/perf-aggregate
+    - name: Download previous benchmark data (smaller-tight)
```
ryan-mist: what is smaller-tight vs ....?

.github/workflows/e2e.yaml
```diff
@@ -71,19 +72,159 @@ jobs:
...
+        # batch-median CV for these metrics stayed under 10% across the
+        # 15-run sample (see the CV comment below). 150% is ~3 batch-median
+        # sigmas above the noise floor for the noisiest tight-tier metric, so
+        # analytical FP < 1% per (test, metric).
+        alert-threshold: '150%'
+        fail-on-alert: true
+        comment-on-alert: true
+        alert-comment-cc-users: '@nathangeology'
+        summary-always: true
+    - name: Compare smaller-is-better metrics (loose)
```
ryan-mist, line 155: The `Compare smaller-is-better metrics (loose)`, `Compare bigger-is-better metrics`, etc pushes a lot of complexity into `e2e.yaml`. Is there any way we can move it to the `main.go` script instead?

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,417 @@
...
+	case int:
+		return float64(x), true
+	case int64:
+		return float64(x), true
+	case json.Number:
+		f, err := x.Float64()
+		return f, err == nil
+	}
+	return 0, false
+}
```
ryan-mist, line 332: I think this can be simplified to

```suggestion
func toFloat(v any) (float64, bool) {
    f, ok := v.(float64)
    return f, ok
}
```

edit: and maybe inlined as well

ryan-mist: and probably inlined

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,417 @@
...
+
+func main() {
+	outputDir := os.Getenv("OUTPUT_DIR")
+	if outputDir == "" {
+		fmt.Fprintln(os.Stderr, "OUTPUT_DIR is required")
+		os.Exit(2)
+	}
+	iterations := 1
+	if v := os.Getenv("ITERATIONS"); v != "" {
+		if _, err := fmt.Sscanf(v, "%d", &iterations); err != nil || iterations < 1 {
```
ryan-mist, line 124: nit: can we change `fmt.Sscanf` to `strconv.Atoi` to align with the rest of the codebase

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,417 @@
...
+	b, err := json.MarshalIndent(v, "", "  ")
+	if err != nil {
+		return err
+	}
+	// path is constructed from OUTPUT_DIR (CI-created via mktemp -d) plus a
+	// literal filename. Not attacker-controlled.
+	return os.WriteFile(path, b, 0o600) //nolint:gosec // G304: path is scoped to CI-created OUTPUT_DIR
+}
+
+func printTable(out *os.File, testKeys []string, summary map[string]map[string]stats) {
```
ryan-mist, line 395: nit: should use `io.Writer` interface instead of `*os.File`

Same comment on `func run(outputDir string, iterations int, out *os.File)`

.github/scripts/perf-aggregate/main.go
```diff
@@ -0,0 +1,417 @@
...
+	display   string
+	unit      string
+	dir       direction
+	tier      tier
+}
+
+var metrics = []metricSpec{
+	{"total_time", "Duration", "seconds", smallerIsBetter, tierTight},
+	{"karpenter_p95_memory_mb", "Controller Peak Memory", "MB", smallerIsBetter, tierTight},
+	{"karpenter_p95_cpu_cores", "Controller CPU", "cores", smallerIsBetter, tierLoose},
```
ryan-mist, line 87: I wonder if we can tweak these to get tighter bounds? (maybe as a followup once we get more data from the actions runs)

Using `karpenter_p50_cpu_cores` we can get a smaller variation:
```
11 runs across recent GH actions (p50 from logs so its only a subset, but avg is full set of results)

┌────────────────────────────────┬─────────┬─────────┬─────────┐
│            sub-test            │ P95 CV% │ avg CV% │ P50 CV% │
├────────────────────────────────┼─────────┼─────────┼─────────┤
│ wide_deployments_consolidation │ 44.1    │ 28.2    │ 38.4    │
├────────────────────────────────┼─────────┼─────────┼─────────┤
│ wide_deployments_scale_out     │ 31.8    │ 20.5    │ 15.0    │
├────────────────────────────────┼─────────┼─────────┼─────────┤
│ drift_execution                │ 21.6    │ 14.8    │ 11.8    │
├────────────────────────────────┼─────────┼─────────┼─────────┤
│ drift_initial_deployment       │ 23.6    │ 15.6    │ 15.4    │
└────────────────────────────────┴─────────┴─────────┴─────────┘
```

I think it also aligns with what we care about - memory at P95 since we do care about bursts (i.e. OOMKills) but CPU at p50 since what we care about is sustained usage and bursts of CPU are less meaningful

ryan-mist: We could also consider tightening bounds based on the test (also as a followup once we get better data on GH action runners)

.github/workflows/e2e.yaml
nathangeology: Yeah, I ran the tests for a bit and that's the defensible threshold that came out, but hopefully we can tighten that down a bit over time.

.github/scripts/perf-aggregate/main.go
nathangeology: It has to do with converting between nanoseconds and seconds

PR #2995, opened 2026-05-02
fix: multi-node consolidation reaches non-prefix subsets (#1962)

### LINKED ISSUE #1962: Consolidation does not happen even when there is cheaper combination of instances available
### Description

**Observed Behavior**:
For context, we wanted to leave the cost effectiveness decision to Karpenter, thus we gave a variety of instance types: c5a,c6a,m6a,m5a,c7a,r6a,r5a,r4, in large/xlarge/2xlarge, thinking the different combinations of cpu:memory options would allow Karpenter to make the best decisions for optimal usage to cost on our behalf. 

However, based on the nodes karpenter chose for us, our memory utilization is pretty good at 90%, however cpu usage is very low (around 50%). 
For instance, we have many c5a.xlarge instances (in the same AZ) that use less than 50% CPU. So two of these could be consolidated into a cheaper m6a.xlarge that has double the amount of memory and the same CPU. But the event on the node says
`Normal  Unconsolidatable  4m36s (x47 over 15h)  karpenter  Can't replace with a cheaper node`

Instance types with CPU usage looks like this:
Image

This ends up being more expensive than our original pre-provisioned nodepool of nodes (which was around 60-65% utilization for both cpu and memory).
To alleviate this issue, we have removed certain instance types from our list of instance types in our NodePool configuration. However we are curious to know if this is the expected behavior because if so then it seems users still have to understand which specific subset of instances fits the resource needs of our clusters, only then can we make use of Karpenter to minimise costs.

**Expected Behavior**:
We expect that we should see multi-node consolidation, defined as
> Multi Node Consolidation - Try to delete two or more nodes in parallel, possibly launching a single replacement whose price is lower than that of all nodes being removed

For instance, we would expect to see 2 c5a.xlarge instances get consolidated into 1 m6a.xlarge as the cpu and memory would fit into that instance and cost lost.
Image

nodepool config:
```{
  "object": {
      "apiVersion": "karpenter.sh/v1",
      "kind": "NodePool",
      "metadata": {
        "annotations": {
          "karpenter.sh/nodepool-hash": "10589712261218411145",
          "karpenter.sh/nodepool-hash-version": "v3"
        },
        "creationTimestamp": null,
        "deletionGracePeriodSeconds": null,
        "deletionTimestamp": null,
        "finalizers": null,
        "generateName": null,
        "generation": null,
        "labels": null,
        "managedFields": null,
        "name": "node-pool-1",
        "namespace": null,
        "ownerReferences": null,
        "resourceVersion": null,
        "selfLink": null,
        "uid": null
      },
      "spec": {
        "disruption": {
          "budgets": [
            {
              "duration": null,
              "nodes": "5%",
              "reasons": null,
              "schedule": null
            },
          ],
          "consolidateAfter": "30s",
          "consolidationPolicy": "WhenEmptyOrUnderutilized"
        },
        "limits": {
          "cpu": "140",
          "memory": "1000Gi"
        },
        "template": {
          "metadata": {
            "annotations": null,
            "labels": null
          },
          "spec": {
            "expireAfter": "Never",
            "nodeClassRef": {
              "group": "karpenter.k8s.aws",
              "kind": "EC2NodeClass",
              "name": "node-pool-1"
            },
            "requirements": [
              {
                "key": "node.kubernetes.io/instance-type",
                "minValues": null,
                "operator": "In",
                "values": [
                  "c5a.xlarge",
                  "c5a.2xlarge",
                  "c6a.xlarge",
                  "c6a.2xlarge",
                  "c7a.xlarge",
                  "c7a.2xlarge",
                  "m5a.xlarge",
                  "m5a.2xlarge",
                  "m6a.xlarge",
                  "m6a.2xlarge",
                  "r4.xlarge",
                  "r4.2xlarge",
                  "r5a.xlarge",
                  "r5a.2xlarge",
                  "r6a.xlarge",
                  "r6a.2xlarge"
                ]
              },
              {
                "key": "karpenter.sh/capacity-type",
                "minValues": null,
                "operator": "NotIn",
                "values": [
                  "spot"
                ]
              },
              {
                "key": "eks.amazonaws.com/capacityType",
                "minValues": null,
                "operator": "In",
                "values": [
                  "ON_DEMAND"
                ]
              },
              {
                "key": "topology.kubernetes.io/zone",
                "minValues": null,
                "operator": "In",
                "values": [
                  "ap-southeast-1a",
                  "ap-southeast-1b",
                  "ap-southeast-1c"
                ]
              }
            ],
            "startupTaints": null,
            "taints": null,
            "terminationGracePeriod": null
          }
        },
        "weight": null
      }
    },
    "timeouts": [],
    "wait": [],
    "wait_for": null
  }
}```

**Versions**:
- Chart Version: v1.1.1
- Kubernetes Version (`kubectl version`): v1.30

jamesmthompson:
Fixes #1962.

## Problem

Multi-node consolidation's binary search over sorted prefixes `[0:mid+1]` is prefix-blind in two related ways:

1. **Prefix-blindness (the originally reported #1962 case).** When a candidate that blocks joint deletion sorts early, every prefix containing it fails the simulator. The search exits empty even when a non-prefix subset of the same candidates would consolidate fine.

2. **Prefix non-maximality.** When the binary search finds a feasible prefix `[0:k]`, a strictly larger non-prefix superset can exist beyond `k` (joint feasibility is non-monotone in prefix length: removing one more candidate can free space that other pods chain into). The prior short-circuit returned the prefix immediately and never probed the superset.

## Fix

The same pairwise non-prefix walk handles both cases:

- **Binary search returns empty:** run the walk from an empty accepted set. Walks candidates in order; accepts each that composes feasibly with the running set; skips otherwise. Non-prefix subsets become reachable because skipping does not narrow the search.
- **Binary search returns a feasible prefix:** run the walk starting from the prefix as initial accepted set, over the candidates beyond the prefix's tail. Extends the prefix into any larger feasible superset.

Both paths are bounded by `MultiNodePairwiseSearchTimeoutDuration` (10s) so the latency added on the cases binary search misses stays capped regardless of N.

## Latency contract

| primary path | additional cost |
|---|---|
| Binary search succeeds, extension finds nothing further | one walk over `n - k` candidates, one simulator probe each |
| Binary search empty, fallback finds something | one walk over `n` candidates |
| Binary search empty, fallback also empty | one walk over `n` candidates, all skipped |

In every case the added work is bounded by the 10s pairwise budget. Big-cluster regret stays inside that budget.

## Verification

- Existing envtest suite (`pkg/controllers/disruption`) passes, ~170s.
- New envtest `multinode_1962_test.go` passes (it fails on the prior code).

A separate property-test corpus on `jamesmt-aws/karpenter:consolidation-corpus-framework` runs 100 seeded scenarios through three configurations (legacy binary-search-only, this PR's behavior, brute-force enumeration baseline). Legacy under-consolidates on 31 of 100 scenarios: 14 prefix-blind (case 1) and 17 prefix-non-maximal (case 2). With this PR, branch matches the brute-force baseline on disruption count, savings, and slack entropy on all 100 scenarios. Latency overhead vs legacy: +33% on average (89ms -> 119ms); 10s cap unchanged.

## Commits

- `test:` envtest reproducer for the prefix-blindness case.
- `fix:` pairwise non-prefix fallback for the empty-binary-search case.
- `chore:` `slices.Clone` for pairwise walk's accepted-set construction.
- `fix:` pairwise extension past the binary-search prefix for the non-maximal-prefix case.

### DIFF
--- a/pkg/controllers/disruption/multinode_1962_test.go
+++ b/pkg/controllers/disruption/multinode_1962_test.go
@@ -1,0 +1,183 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// End-to-end test for kubernetes-sigs/karpenter#1962: multi-node
+// consolidation's binary search over sorted prefixes is prefix-blind.
+// When a candidate that blocks joint deletion (e.g. a node hosting a
+// pod whose requirements no other node can satisfy) sorts between
+// good candidates, every prefix containing the bad candidate fails
+// the simulator. The binary search exits empty and multi-node
+// consolidation contributes nothing, even though a non-prefix subset
+// excluding the bad candidate would consolidate cleanly.
+//
+// This test runs against the real disruption stack (envtest, fake
+// CloudProvider, real provisioner). It constructs three candidates
+// sorted [good_0, bad, good_2] by disruption cost, where bad's pod
+// has a NodeSelector that no remaining node and no replacement can
+// satisfy. The good pods are unrestricted and fit anywhere.
+//
+// Without the fix in firstNConsolidationOption, multi-node returns
+// no command. With the fix (binary search first, pairwise non-prefix
+// fallback when it returns empty), multi-node returns one command
+// deleting good_0 and good_2 together.
+
+package disruption_test
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
+	"sigs.k8s.io/karpenter/pkg/controllers/disruption"
+	"sigs.k8s.io/karpenter/pkg/test"
+	. "sigs.k8s.io/karpenter/pkg/test/expectations"
+)
+
+var _ = Describe("MultiNode Consolidation Non-Prefix Subset (#1962)", func() {
+	It("finds a non-prefix consolidation when a bad candidate sorts between good ones", func() {
+		nodePool := test.NodePool(v1.NodePool{
+			Spec: v1.NodePoolSpec{
+				Disruption: v1.Disruption{
+					ConsolidationPolicy: v1.ConsolidationPolicyWhenEmptyOrUnderutilized,
+					ConsolidateAfter:    v1.MustParseNillableDuration("0s"),
+					Budgets:             []v1.Budget{{Nodes: "100%"}},
+				},
+			},
+		})
+
+		nodeClaims := make([]*v1.NodeClaim, 3)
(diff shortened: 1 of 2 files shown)


jamesmthompson:
Closing for now; keeping the branch on jamesmt-aws while we iterate. Will reopen when ready.

PR #2998, opened 2026-05-04
perf: add consolidation microbenchmarks for regression detection

nathangeology:
Refs #2997

## Related work

**PR #3299** ("test: add microbenchmarks for NewScheduler and NewTopology" by @ryan-mist) is a complementary, coordinated addition covering the provisioning-scheduling slice (`NewScheduler`, `NewTopology`, `ForEachDomain`) that this PR does not touch. Together the two PRs land:

- **PR 2998** — consolidation-loop wall-time + allocs, with a 20% p<0.05 CI gate (currently the only working gate prototype).
- **PR 3299** — provisioning-scheduling component-level allocs, measurement-only for now with a follow-up gate promised.

The two PRs share no benchmark names, no code conflict, and no test overlap. A cross-PR fixture consolidation into `pkg/test/bench` is a plausible follow-up once both land.

## Motivation

PR #2671 (topology domain filtering) introduced a ~2x consolidation perf regression that went undetected because:

1. Scheduling benchmarks only cover the provisioning path (`Scheduler.Solve()`), not consolidation rescheduling.
2. Kind-cluster e2e benchmarks have 40-60% CV and run too infrequently to catch per-PR regressions.

Go microbenchmarks fill this gap. They run in seconds with low variance and catch algorithmic complexity changes on every PR.

## Changes

**`pkg/controllers/disruption/consolidation_benchmark_test.go`**

Benchmarks for `SimulateScheduling()`, the hot path where consolidation checks if a candidate node's pods can land elsewhere. Parameterized by node count (10/50/100), topology spread fraction (0%/50%/100%), and NodePool count (1/3/9). The `100Nodes_HostnameSpread_9NP` case targets the #2671 regression pattern directly.

500-node variants are included for local profiling but gated behind `testing.Short()` to avoid CI timeouts on shared runners.

**`.github/workflows/scheduling-benchmarking.yaml`**

- Adds `pkg/controllers/disruption` to the benchmark matrix
- New `compare` job: downloads before/after artifacts, runs `benchstat`, posts a sticky PR comment, fails if any benchmark regresses >20% at p<0.05

**`.github/workflows/run-bench-test.yaml`**

- `-run='^$'` instead of `-run=1` (the old regex matched test functions, triggering Ginkgo suite setup)
- `-benchtime=1x` since these benchmarks are already ms-scale per op
- `-short` to skip 500-node cases in CI
- `-count=6` for benchstat confidence intervals
- Adds `setup-go` and envtest binary installation

## Not covered

Concurrent reconciliation, controller interactions, cloud provider latency. Those need the kind-cluster benchmarks (PR #2994) or canaries monitoring cloud infra performance.

## Validation

Ran the benchmarks with CI-equivalent flags (`-short -benchtime=1x -count=6`) against a cherry-pick of #2671:

```
                                             │ good (reverted) │        bad (PR #2671)         │
                                             │     sec/op      │    sec/op      vs base        │
Consolidation_100Nodes_HostnameSpread-14           12.92µ ± 18%   19.69µ ± 44%  +52.42% (p=0.002 n=6)
Consolidation_100Nodes_HostnameSpread_3NP-14       12.77µ ± 51%   21.23µ ± 30%  +66.23% (p=0.009 n=6)
Consolidation_100Nodes_HostnameSpread_9NP-14       11.85µ ± 27%   21.54µ ±  6%  +81.72% (p=0.002 n=6)
geomean                                            12.51µ          20.80µ        +51.10%
```

All topology cases trip the 20% gate at p<0.01.

**AI Disclosure**

I used an LLM to parallel program with me on this. The original consolidation microbenchmark set was drafted collaboratively. 


pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
+//go:build test_performance
```
ryan-mist: should we drop this?

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+	b.ReportMetric(float64(cfg.nodeCount), "nodes")
+	b.ReportMetric(float64(cfg.nodePoolCount), "nodepools")
+	b.ReportMetric(cfg.topologySpreadFraction*100, "topo%")
+}
+
+// --- Setup ---
+
+func setupConsolidationBench(b *testing.B, cfg benchConfig) (
+	context.Context, client.Client, *clock.FakeClock, *state.Cluster,
+	*fake.CloudProvider, *provisioning.Provisioner, []*Candidate,
```
ryan-mist: nit: it doesn't look like `*fake.CloudProvider` ever gets used - should we drop this from the return function definition?

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+				Labels: map[string]string{
+					v1.NodePoolLabelKey:            np.Name,
+					corev1.LabelInstanceTypeStable: it.Name,
+					corev1.LabelTopologyZone:       fmt.Sprintf("zone-%d", i%3),
+				},
+			},
+			Status: v1.NodeClaimStatus{
+				ProviderID: fmt.Sprintf("fake://node-%d", i),
+				Allocatable: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse("16"),
```
ryan-mist: Theres a few values like these that we can probably randomize to get (hopefully) better coverage on  edge cases. I think some things like NodePool and pod configurations (especially around TSCs) would be nice to randomize but might not be strictly necessary because it'll be more complexity to do it

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+				Requests: corev1.ResourceList{
+					corev1.ResourceCPU:    resource.MustParse(fmt.Sprintf("%dm", 100+benchRand.Intn(400))),
+					corev1.ResourceMemory: resource.MustParse(fmt.Sprintf("%dMi", 128+benchRand.Intn(512))),
+				},
+			},
+		}
+		if i < topologyCount {
+			opts.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{{
+				MaxSkew:           1,
+				TopologyKey:       corev1.LabelHostname,
```
ryan-mist: Should we randomize this as well?

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+			ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("pool-%d", i)},
+			Spec: v1.NodePoolSpec{
+				Limits: v1.Limits{
+					corev1.ResourceCPU:    resource.MustParse("100000"),
+					corev1.ResourceMemory: resource.MustParse("100000Gi"),
+				},
+			},
+		})
+		nodePools[i] = np
+		if err := kubeClient.Create(ctx, np); err != nil {
```
ryan-mist, line 290: Right now only the NodePool is created - I think we also need to create the Node, NodeClaim, and pod objects as well. Otherwise they never make it into `SimulateScheduling` and we solve against an empty cluster

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+					corev1.ResourceMemory: resource.MustParse("64Gi"),
+					corev1.ResourcePods:   resource.MustParse("110"),
+				},
+			},
+		})
+
+		pods := makeBenchPods(cfg.podsPerNode, cfg.topologySpreadFraction, node.Name)
+
+		candidates = append(candidates, &Candidate{
+			StateNode: &state.StateNode{
```
ryan-mist: We should update cluster state instead of creating local slices - IIUC, what happens now is we don't pass in any nodes into scheduler (as `stateNode` is an empty slice) so we end up solving on an empty cluster:

https://github.com/kubernetes-sigs/karpenter/blob/2266468104f36373ed002241afaee2d3bbc26078/pkg/controllers/provisioning/provisioner.go#L265-L270

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+	// This exercises the full rescheduling path: collect pods from the candidate,
+	// create a scheduler with topology constraints, and solve placement.
+	candidate := candidates[0]
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		_, _ = SimulateScheduling(ctx, kubeClient, clusterState, prov, clk, rec, nil, candidate)
+	}
+	b.ReportMetric(float64(len(candidate.reschedulablePods)), "pods")
+	b.ReportMetric(float64(cfg.nodeCount), "nodes")
```
ryan-mist, line 242: Can we also add reporting for any allocations we make - ` b.ReportAllocs()` along with adding the flag for `-benchmem`

.github/workflows/run-bench-test.yaml
```diff
@@ -22,14 +22,36 @@ jobs:
...
         run: |
+          echo "Go version: $(go version)"
+          echo "Working directory: $(pwd)"
+          cd ${{ inputs.path }}
+          echo "Test directory: $(pwd)"
+          echo "Listing test files with build tag:"
+          grep -rl "test_performance" . --include="*.go" || echo "No files with test_performance tag"
+          mkdir -p output
+          echo "Running benchmarks..."
+          if ! go test -tags=test_performance -run='^$' -bench=. -benchtime=1x -short -count=6 -timeout=30m -cpuprofile output/cpu.out -memprofile output/mem.out -v 2>&1 | tee output/results.txt; then
```
ryan-mist: Should we drop  `-benchtime=1x`? It looks like without that flag it'll run as many times as possible in one second and average the runs together so we'll get more accurate results

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,267 @@
...
+	benchmarkConsolidationSim(b, benchConfig{100, 10, 3, 1.0})
+}
+
+func BenchmarkConsolidation_100Nodes_HostnameSpread_9NP(b *testing.B) {
+	// This is the case that PR#2671 regressed: O(pods * domains * NodePools)
+	benchmarkConsolidationSim(b, benchConfig{100, 10, 9, 1.0})
+}
+
+func BenchmarkConsolidation_500Nodes_NoTopology(b *testing.B) {
+	if testing.Short() {
```
ryan-mist: How long to these tests end up being? Depending on the timing, it may still be useful to run them

ryan-mist: Does precomputing `setupConsolidationBench` before each run (currently 6 based on `-count=6`) help with speedup?

nathangeology: It's about 88% setup for the tests. Pre-computing would probably save some time though and wouldn't change the tests too much as they are pretty deterministic atm.

pkg/controllers/disruption/consolidation_benchmark_test.go
nathangeology: Maybe! A lot of the benefit of these tests is that they have low variance and will flag even very small performance regressions. If we randomize inputs then we get more variance and have a harder time picking up regression signals. That said, I/we should think about how to evaluate what additional configs to test in a more formalized way.

pkg/controllers/disruption/consolidation_benchmark_test.go
nathangeology: Going to go ahead and add some changes to do that and see how it looks.

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -22,15 +22,92 @@ jobs:
...
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        continue-on-error: true
+        with:
+          name: after-${{ matrix.index }}
+          path: after
+      - name: Check artifacts
+        id: check
+        run: |
+          if [ ! -f before/results.txt ] || [ ! -f after/results.txt ]; then
+            echo "::warning::Benchmark artifacts not found. Skipping comparison."
```
ryan-mist: Should we fail hard here?

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -22,15 +22,92 @@ jobs:
...
+        run: |
+          RESULT=$(benchstat before/results.txt after/results.txt 2>&1 || true)
+          {
+            echo "result<<EOF"
+            echo "$RESULT"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+          # Check for statistically significant regressions > 20%
+          if echo "$RESULT" | grep -E '\+[2-9][0-9]\.[0-9]+%|\+[1-9][0-9]{2,}\.[0-9]+%' | grep -v '~' | grep -qE 'p=0\.0[0-4]'; then
```
ryan-mist: I wonder if we should only gate on `allocs/op` and `B/op`? `sec/op` can vary a lot depending on the runner that we get

.github/workflows/scheduling-benchmarking.yaml
```diff
@@ -22,15 +22,92 @@ jobs:
...
+        shell: bash
+        run: |
+          RESULT=$(benchstat before/results.txt after/results.txt 2>&1 || true)
+          {
+            echo "result<<EOF"
+            echo "$RESULT"
+            echo "EOF"
+          } >> "$GITHUB_OUTPUT"
+
+          # Check for statistically significant regressions > 20%
```
ryan-mist: I think we should change this 20% depending on the key by parsing the output. `allocs/op` is pretty stable, so the 20% can be better tuned while `sec/op` varies a lot (I tested it and 2/5 runs triggered a regression for identical code)

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,416 @@
...
+// expensive-to-build objects.
+type cachedBench struct {
+	kubeClient   client.Client
+	clk          *clock.FakeClock
+	clusterState *state.Cluster
+	prov         *provisioning.Provisioner
+	candidates   []*Candidate
+}
+
+// benchCache memoizes setupConsolidationBench output keyed on benchConfig so
```
ryan-mist: I think some of these comments can be shortened a little or removed

.github/workflows/scheduling-benchmarking.yaml
nathangeology: For the 'after' change artifacts, yes. If the before artifacts are missing, probably not, just an info/warn there. This should be clearer in the current version.

pkg/controllers/disruption/consolidation_benchmark_test.go
nathangeology: Tried trimming them down a bit, feel free to flag any more that you think are a bit much haha

.github/workflows/scheduling-benchmarking.yaml
nathangeology: I checked and saw the same thing, switched to allocs/ops only, this seems to be pretty reliable in the test data I ran and still catches the regression PR.

nathangeology: Claude said this: "Switched the gate to allocs/op only at 20% + p<0.05. FP was 0/31 across two variance regimes (95% CI 0-11%). Still catches #2671: allocs +51% at n=10 and +35% at np=9, p=0.002 both. Allocs are deterministic from code paths, so the signal is portable across runners."

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,355 @@
...
+type benchConfig struct {
+	nodeCount              int
+	podsPerNode            int
+	nodePoolCount          int
+	topologySpreadFraction float64
+}
+
+// BenchmarkConsolidation sweeps three vectors (nodes, nodepools, topology)
+// over the SimulateScheduling hot path. 500-node sub-benchmarks are gated
+// by testing.Short() to avoid CI timeouts; they run under local profiling.
```
ryan-mist: How close are we to CI timeouts? I ran n=500 locally and it seems to run these 4 cases in ~3 mins

.github/workflows/run-bench-test.yaml
```diff
@@ -22,14 +22,36 @@ jobs:
...
         run: |
+          echo "Go version: $(go version)"
+          echo "Working directory: $(pwd)"
+          cd ${{ inputs.path }}
+          echo "Test directory: $(pwd)"
+          echo "Listing test files with build tag:"
+          grep -rl "test_performance" . --include="*.go" || echo "No files with test_performance tag"
+          mkdir -p output
+          echo "Running benchmarks..."
+          if ! go test -tags=test_performance -run='^$' -bench=. -benchmem -short -count=6 -timeout=30m -cpuprofile output/cpu.out -memprofile output/mem.out -v 2>&1 | tee output/results.txt; then
```
ryan-mist: Can we add `-benchtime=1x` (or maybe a `2x`, etc)

IIUC, this suite could timeout - `b.N` scales to the suite timeout based on the timer, but since the timer doesn't include `candidatesForBench` then it scales slightly inaccurately

```
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		budgets, cands := candidatesForBench(b, singleNode, numNodes)
		b.StartTimer()
		if _, err := singleNode.ComputeCommands(benchCtx, budgets, cands...); err != nil {
			b.Fatalf("compute commands: %v", err)
		}
	}
```

ryan-mist: I think we can also drop `-short` - it looks unused now

pkg/controllers/disruption/consolidation_benchmark_test.go
```diff
@@ -0,0 +1,350 @@
...
+	return budgets, cands[:numNodes]
+}
+
+func BenchmarkSingleNodeConsolidation_ComputeCommands_100(b *testing.B) {
+	benchmarkSingleNodeConsolidation(b, 100)
+}
+func BenchmarkSingleNodeConsolidation_ComputeCommands_400(b *testing.B) {
+	benchmarkSingleNodeConsolidation(b, 400)
+}
+func BenchmarkSingleNodeConsolidation_ComputeCommands_1000(b *testing.B) {
```
ryan-mist, line 320: Not sure how you feel about messing with the inputs too much (especially based on behavior of the underlying method) - but since single node consolidation exists early on first approved candidate. Should we order the list in such a way to ensure most of the candidates are evaluated?

.github/workflows/run-bench-test.yaml
```diff
@@ -22,14 +22,36 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           ref: ${{ inputs.githubSha }}
+      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+        with:
+          go-version-file: go.mod
+          cache-dependency-path: "**/go.sum"
+      - name: Install envtest binaries
```
ryan-mist: I think this `Install envtest binaries` step can be dropped

pkg/controllers/disruption/consolidation_benchmark_test.go
nathangeology: Looked at this and ran some tests and it wasn't really going deep into the list. The current version should be doing a better job of avoid early exits on the first candidate. I also rerun it on the original regression PR to check that we are still catching that as verification.

pkg/controllers/disruption/consolidation_benchmark_test.go
nathangeology: With -count=6 the runs that I measured sum to ~20 min for SingleNode plus MultiNode → ~25-27 min total. Workflow budget is -timeout=30m so headroom is ~3-5 min. 

If we want a bigger margin: option to reduce -count from 6 to 5 (saves ~4-5 min but noisier benchstat), or cap SingleNode max at 800 instead of 1000 (saves ~2 min per count, less coverage on the loop-scaling regime this bench is designed to detect), or bump -timeout=40m.

PR #3000, opened 2026-05-04
chore: add pending pods by zone metric

ryan-mist:
Fixes #N/A 

**Description**
* adds metric `karpenter_scheduler_pending_pods_by_effective_zone` which records if the zone is pending pod is constrained to a zone during scheduling. 
  * the value for this metric is either `flexible`, `none`, or the specific zone name

* unit tests

**Perf Tests**
With 1000 pods with TSC on hostname and a max skew of 10, `computeEffectiveZoneFromPod` uses a very small amount of CPU time:
Screenshot 2026-05-12 at 10 57 51 AM
Full flame graph:
Screenshot 2026-05-12 at 10 58 50 AM


pkg/controllers/provisioning/scheduling/metrics.go
```diff
@@ -92,4 +92,17 @@ var (
 			ControllerLabel,
 		},
 	)
+	PendingPodsByEffectiveZone = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: schedulerSubsystem,
+			Name:      "pending_pods_by_effective_zone",
```
joshuakguo: small nit but maybe `pending_pods_by_effective_zone_count`? just to follow the names of the other metrics?

ryan-mist: makes sense, changed

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -85,6 +85,7 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	for i := range nodePoolList.Items {
 		its, err := c.cloudProvider.GetInstanceTypes(ctx, &nodePoolList.Items[i])
 		if err != nil {
+			// instead of returning here we should still evalaute non-error NodePools
```
DerekFrank: Think you pulled something else in here

pkg/controllers/provisioning/scheduling/metrics.go
```diff
@@ -92,4 +92,17 @@ var (
...
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: schedulerSubsystem,
+			Name:      "pending_pods_by_effective_zone_count",
+			Help:      "Pending pods dimensioned by effective zone constraint, or the intersection of pod-level zone signals, volume topology (PVC zones), and topology constraints. Values: specific zone name (e.g., 'us-west-2a'), 'flexible' (multiple zones), or 'none' (no valid intersection).",
+		},
+		[]string{
+			ControllerLabel,
+			"zone",
```
DerekFrank, line 105: zone or zone-id?

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -241,9 +243,9 @@ func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, no
 		// Because Karpenter can always create a new domain for hostname, we assume the global miniumum is always zero
 		// This means we can just check whether our current count is less than or equal to the skew to check if the domain is valid
 		if count <= t.maxSkew {
-			return scheduling.NewRequirement(t.Key, corev1.NodeSelectorOpIn, hostName)
+			return scheduling.NewRequirement(t.Key, corev1.NodeSelectorOpIn, hostName), sets.New[string](hostName)
```
DerekFrank: nit: double initialization, you made sets.New() above and then never used it in this path

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -272,18 +277,21 @@ func (t *TopologyGroup) nextDomainTopologySpread(pod *corev1.Pod, podDomains, no
...
+						minCount = count
+					}
 				}
 			}
 		}
 	}
 	if minDomain == "" {
 		// avoids an error message about 'zone in [""]', preferring 'zone in []'
-		return scheduling.NewRequirement(t.Key, corev1.NodeSelectorOpDoesNotExist)
+		return scheduling.NewRequirement(t.Key, corev1.NodeSelectorOpDoesNotExist), nil
```
DerekFrank: Why nil here?

pkg/controllers/provisioning/scheduling/topologygroup.go
```diff
@@ -125,14 +125,14 @@ func NewTopologyGroup(
 	}
 }
 
-func (t *TopologyGroup) Get(pod *corev1.Pod, podDomains, nodeDomains *scheduling.Requirement) *scheduling.Requirement {
+func (t *TopologyGroup) Get(pod *corev1.Pod, podDomains, nodeDomains *scheduling.Requirement) (*scheduling.Requirement, sets.Set[string]) {
```
DerekFrank, line 128: Does this ever actually return a set?

pkg/controllers/provisioning/scheduling/metrics.go
ryan-mist: zone

pkg/controllers/provisioning/scheduling/topologygroup.go
ryan-mist: Good point, fixed

pkg/controllers/provisioning/scheduling/topologygroup.go
ryan-mist: Currently only for spread constraints, for pod affinity and anti-affinity the domains are in scheduling requirements already.

Its why in `GetTopologyZoneConstraints`, theres 2 cases:
```
		case TopologyTypeSpread:
			if validDomains != nil && validDomains.Len() > 0 {
				validZones = validDomains
			}
		case TopologyTypePodAffinity, TopologyTypePodAntiAffinity:
			if req != nil && req.Len() > 0 {
				validZones = sets.New[string](req.Values()...)
			}
		}
```
---

Admittedly, its a bit confusing - I'll change it to return `sets.New[string](req.Values()...)` here instead of doing the processing later on.

pkg/controllers/provisioning/scheduling/topologygroup.go
ryan-mist: No reason, its treated the same as an empty set - Ill unify the return to always do validDomains

---

While looking at this, I just realized that unsatisfiable topology groups still report as flexible. Fixed in the next rev and added a test case

DerekFrank (approved):
/lgtm

PR #3002, opened 2026-05-05
fix: isolate per-NodePool failures in nodeoverlay controller

### LINKED ISSUE #3003: NodeOverlay: one misconfigured NodePool blocks all provisioning after Karpenter restart
### Description

**Title:** NodeOverlay controller: single failing NodePool blocks all provisioning after restart

**Observed Behavior**:

When the NodeOverlay feature gate is enabled, a single NodePool with an invalid configuration (for example, referencing a NodeClass that doesn't exist) causes all node provisioning to stop cluster-wide after a Karpenter restart.

The nodeoverlay controller's `Reconcile()` function iterates over all NodePools and calls `GetInstanceTypes()` for each. If any single call fails, the entire function returns an error immediately without updating the `InstanceTypeStore`. Because the store never gets populated with `evaluatedNodePools`, every downstream caller (provisioner, disruption, pricing) receives an `UnevaluatedNodePoolError` for every pool, and no pods can be scheduled.

The cascade only triggers after a Karpenter restart because the store is initialized empty on startup. While Karpenter is running continuously, a previously successful reconcile keeps the old store in memory (the atomic pointer swap only happens on successful reconciles). But once Karpenter restarts, the fresh empty store combined with a broken NodePool means the first reconcile fails, and the store is never populated.

The bug is in `pkg/controllers/nodeoverlay/controller.go`, `Reconcile()`:

```go
    if err != nil {
        return reconcile.Result{}, fmt.Errorf("listing instance types, %w", err)
        // returns immediately, no remaining pools are evaluated
    }
    nodePoolToInstanceTypes[nodePoolList.Items[i].Name] = its
}
```

The `return` should be a `continue` so that the remaining pools are still evaluated and the store gets updated with partial results. The provisioner already uses this pattern for the same scenario.

**Expected Behavior**:

The broken NodePool should be skipped, and all other healthy NodePools should continue to work normally. This is already how the provisioner handles the same error type in `provisioner.go`: it logs a message, skips the pool, and continues with the rest.

1. Create a Karpenter cluster with `NodeOverlay=true` feature gate
2. Wait for Karpenter to be healthy and the default NodePool to be Ready
3. Create an AKSNodeClass and a NodePool that references a different, non-existent NodeClass name:

```yaml
apiVersion: karpenter.azure.com/v1beta1
kind: AKSNodeClass
metadata:
  name: repro-test
spec:
  imageFamily: AzureLinux
  osDiskSizeGB: 32
---
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: repro-test
spec:
  disruption:
    consolidateAfter: 10m
  limits:
    cpu: 4
  template:
    spec:
      expireAfter: 720h
      nodeClassRef:
        group: karpenter.azure.com
        kind: AKSNodeClass
        name: repro-missing  # does not exist
      requirements:
        - key: kubernetes.io/arch
          operator: In
          values: [amd64]
        - key: kubernetes.io/os
          operator: In
          values: [linux]
          operator: In
          values: [on-demand]
        - key: karpenter.azure.com/sku-family
          operator: In
          values: [D]
```

4. Restart Karpenter (e.g., upgrade the cluster, or trigger a control plane reconcile)
5. All healthy NodePools show `NoCompatibleInstanceTypes` and no pods can be scheduled

We reproduced this on AKS managed NAP clusters (Karpenter `1.10.2-aks`, commit `643351d`) across three configurations:

- 3 NodePools (1 broken + 2 healthy): cascade triggered but self-resolved through retries
- 13 NodePools (1 broken + 12 healthy): cascade triggered and all 12 healthy pools showed `NoCompatibleInstanceTypes` simultaneously

The original production incident involved a customer with 12 NodePools where one had a typo in the NodeClass reference. After a cluster upgrade, all provisioning stopped.

**Versions**:
- Karpenter version: `1.10.2-aks` (commit `643351d`), bug also present on `main` as of `eda8625e`
- Feature gate: `NodeOverlay=true`
- Cloud provider: Azure (AKS managed NAP), but the bug is in core Karpenter and is provider-agnostic

**Fix**: PR #3002

jovieir:
The nodeoverlay controller's Reconcile() loop calls GetInstanceTypes() for every NodePool. Previously, if any single NodePool failed (e.g., its NodeClass doesn't exist), the entire reconcile returned an error immediately. This prevented the InstanceTypeStore from being updated, causing ALL NodePools to be stuck in 'awaiting evaluation' state and blocking all node provisioning cluster-wide.

This change:
- Skips failing NodePools with a log message instead of aborting
- Only marks successfully evaluated NodePools in evaluatedNodePools
- Passes only evaluated NodePools to overlay validation
- Allows the store to be updated with partial results

This matches the error handling pattern already used by the provisioner in provisioner.go for the same error type.

Fixes a cascading failure where one misconfigured NodePool (referencing a non-existent NodeClass) blocks provisioning on all other healthy NodePools after any Karpenter restart.

Fixes #3003  

**Description**

One misconfigured NodePool (e.g., referencing a non-existent NodeClass) causes the nodeoverlay controller to abort its entire reconcile loop on the first error. After a Karpenter restart, this prevents the InstanceTypeStore from being populated, blocking provisioning on all NodePools cluster-wide, not just the broken one.

2 new tests added to `suite_test.go`, full suite passes. The new tests verify:

1. Healthy NodePools are still evaluated and usable when one pool fails `GetInstanceTypes`
2. The store is properly updated with partial results after a fresh start (simulating Karpenter restart)

One existing test (`should have an empty instance types set when cloudprovider return an error`) updated to match the new skip-instead-of-abort behavior: reconcile now succeeds and the failing pool is excluded from `evaluatedNodePools` rather than aborting the entire reconcile.

### DIFF
--- a/pkg/controllers/nodeoverlay/controller.go
+++ b/pkg/controllers/nodeoverlay/controller.go
@@ -121,6 +121,12 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	c.instanceTypeStore.UpdateStore(temporaryStore)
 	c.clusterState.MarkUnconsolidated()
 
+	// If some NodePools failed to resolve instance types, requeue sooner to
+	// recover from transient errors (e.g., temporary API failures, NodeClass
+	// not yet created) without waiting for the full 6h polling interval.
+	if len(evaluatedNodePoolItems) < len(nodePoolList.Items) {
+		return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
+	}
 	return reconcile.Result{RequeueAfter: 6 * time.Hour}, nil
 }
 
--- a/pkg/controllers/nodeoverlay/controller.go
+++ b/pkg/controllers/nodeoverlay/controller.go
@@ -27,10 +27,12 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -82,12 +84,15 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	temporaryStore := newInternalInstanceTypeStore()
 	nodePoolToInstanceTypes := map[string][]*cloudprovider.InstanceType{}
 
+	evaluatedNodePoolItems := make([]v1.NodePool, 0, len(nodePoolList.Items))
 	for i := range nodePoolList.Items {
 		its, err := c.cloudProvider.GetInstanceTypes(ctx, &nodePoolList.Items[i])
 		if err != nil {
-			return reconcile.Result{}, fmt.Errorf("listing instance types, %w", err)
+			log.FromContext(ctx).WithValues("NodePool", klog.KObj(&nodePoolList.Items[i])).Error(err, "skipping, unable to list instance types for nodepool")
+			continue
 		}
 		nodePoolToInstanceTypes[nodePoolList.Items[i].Name] = its
+		evaluatedNodePoolItems = append(evaluatedNodePoolItems, nodePoolList.Items[i])
 	}
 
 	overlayList.OrderByWeight()
@@ -97,11 +102,11 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 			continue
 		}
 
-		if !c.validateAndUpdateInstanceTypeOverrides(temporaryStore, nodePoolList.Items, nodePoolToInstanceTypes, overlayList.Items[i]) {
+		if !c.validateAndUpdateInstanceTypeOverrides(temporaryStore, evaluatedNodePoolItems, nodePoolToInstanceTypes, overlayList.Items[i]) {
 			overlaysWithConflict = append(overlaysWithConflict, overlayList.Items[i].Name)
 		}
 	}
-	temporaryStore.evaluatedNodePools.Insert(lo.Map(nodePoolList.Items, func(np v1.NodePool, _ int) string {
+	temporaryStore.evaluatedNodePools.Insert(lo.Map(evaluatedNodePoolItems, func(np v1.NodePool, _ int) string {
 		return np.Name
 	})...)
 
--- a/pkg/controllers/nodeoverlay/controller.go
+++ b/pkg/controllers/nodeoverlay/controller.go
@@ -27,12 +27,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/klog/v2"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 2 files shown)


nathangeology:
Thanks for submitting this. First pass through, this looks good to me. Was able to repro the issue with your and some similar tests that I tried. I also tested that the issue was fixed with this PR and it does.

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -82,12 +84,15 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	temporaryStore := newInternalInstanceTypeStore()
 	nodePoolToInstanceTypes := map[string][]*cloudprovider.InstanceType{}
 
+	evaluatedNodePoolItems := make([]v1.NodePool, 0, len(nodePoolList.Items))
 	for i := range nodePoolList.Items {
 		its, err := c.cloudProvider.GetInstanceTypes(ctx, &nodePoolList.Items[i])
 		if err != nil {
-			return reconcile.Result{}, fmt.Errorf("listing instance types, %w", err)
+			log.FromContext(ctx).WithValues("NodePool", klog.KObj(&nodePoolList.Items[i])).Error(err, "skipping, unable to list instance types for nodepool")
```
ryan-mist: Bringing a comment over from a previous PR - https://github.com/kubernetes-sigs/karpenter/pull/2550
> We should requeue more aggressively if we fail to get instance types for any NodePools, otherwise we might get resolution failures until we poll again.

jovieir: good callout. It won't fix persistent misconfigs but certainly helps with transient errors.

jovieir: Addressed in 92d5657

ryan-mist (approved):
LGTM 🚀

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -116,6 +121,12 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	c.instanceTypeStore.UpdateStore(temporaryStore)
 	c.clusterState.MarkUnconsolidated()
 
+	// If some NodePools failed to resolve instance types, requeue sooner to
+	// recover from transient errors (e.g., temporary API failures, NodeClass
+	// not yet created) without waiting for the full 6h polling interval.
+	if len(evaluatedNodePoolItems) < len(nodePoolList.Items) {
+		return reconcile.Result{RequeueAfter: 1 * time.Minute}, nil
```
DerekFrank: Instead can we save the errors from above and return them here as a multierr? That'll do logging by default, and I'm not sure if we want to requeue with backoff or not. I think we want to aggressively retry so RequeueAfter is the right behavior

jovieir: Updated in ebdeaa7, PTAL :-). Per-NodePool errors are now collected and returned as a multierr alongside the RequeueAfter, so controller-runtime's "Reconciler error" path does the logging for us.

One thing worth flagging though: in controller-runtime v0.22.4 (pkg/internal/controller/controller.go around line 477) the RequeueAfter is dropped whenever the reconciler returns a non-nil error, so on the error path the requeue delay is entirely governed by the workqueue's rate limiter. With karpenter's reasonable.RateLimiter(), repeated failures back off up to ~1m and stay there, which lines up with the cadence we had before. So we still get aggressive retries, just not through the explicit RequeueAfter.

If you'd rather RequeueAfter be the authoritative signal on this path (so the delay is fixed at exactly 1m regardless of failure count), we can log the multierr ourselves and return nil err. Happy to swap if you prefer that shape.

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -116,6 +124,15 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
...
 	c.clusterState.MarkUnconsolidated()
 
+	// If some NodePools failed to resolve instance types, return the
+	// aggregated errors so controller-runtime logs them via its standard
+	// "Reconciler error" path, and requeue sooner to recover from transient
+	// errors (for example temporary API failures or a NodeClass that has
+	// not been created yet) without waiting for the full 6h polling
+	// interval.
+	if len(nodePoolErrs) > 0 {
+		return reconcile.Result{RequeueAfter: 1 * time.Minute}, multierr.Combine(nodePoolErrs...)
```
ryan-mist: nit: we can drop the `RequeueAfter: 1 * time.Minute` since it won't be used

pkg/controllers/nodeoverlay/controller.go
DerekFrank: No you're right, we should just defer to the rate limiter if we're returning an error

pkg/controllers/nodeoverlay/controller.go
```diff
@@ -82,12 +82,20 @@ func (c *Controller) Reconcile(ctx context.Context, _ reconcile.Request) (reconc
 	temporaryStore := newInternalInstanceTypeStore()
 	nodePoolToInstanceTypes := map[string][]*cloudprovider.InstanceType{}
 
+	evaluatedNodePoolItems := make([]v1.NodePool, 0, len(nodePoolList.Items))
+	nodePoolErrs := make([]error, 0, len(nodePoolList.Items))
```
DerekFrank: nit: We typically do this by making a multierr here and then calling multierr.Append in the loop

ryan-mist (approved):
/lgtm

DerekFrank (approved):

PR #3004, opened 2026-05-06
docs: RFC for supporting price expression in Node Overlay

### LINKED ISSUE #2616: Node Overlays - multiple price adjustments
### Description

Creating this here as as a followup to [karpenter-provider-aws#8618](https://github.com/aws/karpenter-provider-aws/issues/8618), at the suggestion of @jmdeal.

I opened [karpenter-aws-provider#5033](https://github.com/aws/karpenter-provider-aws/issues/5033) 2 years ago and with node overlays in Karpenter 1.7 I think we now have a solution to most of that problem. This is a great improvement, thanks for adding node overlays.

However, it seems that node overlays limit the number of price adjustments applied to a node to just 1. It would be beneficial to allow multiple price adjustments applied in a predictable order.

In my case, I would like to apply a % decrease for an enterprise discount, and then a fixed $ license fee for a vendor license fee. I don't think I can do that currently. The docs are currently somewhat ambiguous (see #8545), but I've tried this and it does not seem to work.

Another hypothetical but realistic case is where different per-node license fees are applied in independent and possibly-intersecting cases. For example, you might have a fixed per-node support fee that applies to all nodes; but you might also have some GPU-specific software that is only supported on an operating system with a per-server license fee, so there's an additional price incentive to consolidate GPU nodes. In some cases, then, you want to apply 2 adjustments, and in some cases 1 adjustment. It's desirable to manage these price adjustments as 2 distinct layers of overlays with intersecting requirements.

It's much less important to me than [karpenter-aws-provider#5033](https://github.com/aws/karpenter-provider-aws/issues/5033) was, and just being able to apply one license fee is a big improvement. But the current limitation to a single price adjustment is surprising and doesn't seem necessary.

In my case if I could apply the % discount before the fixed $ license fee in Karpenter's node selections then I would get better cost efficiency. The license fee is proportionately more impactful on pricing when the % discount is preapplied.

I might be able to work around this by inverting the discount and applying that to the license fee in advance (so that the adjusted prices are proportionate with billed prices), but it's non-intuitive to other K8s cluster maintainers so I probably won't do that.

GnatorX:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/2616

**Description**
- RFC for supporting multiple price adjustments in node overlay


jmdeal:
I haven't read the full RFC yet, but one other option that's worth considering is making the price adjustment a CEL expression. For example, we could express your scenario like this:
```yaml
spec:
  price: "(self.price * 0.9 + 0.05) * 1.03"
```
This provides the user with full control over order of operations while also enabling all adjustments to be made from a single overlay. We're actively considering similar CEL based controls for kubelet settings like kubeReserved and systemReserved in the AWS provider. Thoughts?

GnatorX:
At the moment, the RFC is suggesting price adjustments (as a list of adjustments) on a single overlay. 

I don't mind cel but my worry is around is it too expressive and would users end up just creating incorrect expressions that are difficult to validate? Is there any performance concerns (Not sure myself)

designs/node-overlay-price-adjustments.md
```diff
@@ -0,0 +1,237 @@
+# NodeOverlay `priceAdjustments`
+
+## Summary
+
+This RFC proposes two approaches for enabling cumulative price adjustment stacking across matching NodeOverlays, solving the combinatorial problem that arises when multiple independent cost dimensions (enterprise discounts, licensing fees, regional surcharges) need to be modeled independently but applied together.
+
+**Option 1 (preferred)** introduces a `spec.priceAdjustments` field—an ordered list of adjustments applied sequentially within a single overlay. This is the recommended approach: it requires no cross-overlay coordination, has a simple and explicit ordering model, and avoids the policy-mixing error states inherent in Option 2.
```
jigisha620: How does Option 1 handle the case where cost dimensions have different selectors? For example, if a global discount applies to all on-demand instances but a licensing fee only applies to a specific instance family, these can't share a single overlay's requirements. With highest-weight-wins, only one overlay applies and the other is lost. Would users need to create separate overlays for every unique combination of selectors with their adjustments merged together?

jigisha620 (commented):
Thanks for the RFC!
Would it make sense to split the readiness-ttl RFC into its own PR? It seems like a separate feature that might benefit from independent review and discussion.

I think it is worth considering the CEL approach (from @jmdeal 's  comment) as a section in the RFC? It would be helpful to see it compared against the other two options explicitly

GnatorX:
cel may have a way to evaluate how expensive computationally an expression is. We should gate expressions if possible. Add more examples on pricing scenarios in document (outside of RFC) to help users setup their pricing expression

designs/node-overlay-price-adjustments.md
GnatorX: I added this section to talk about this https://github.com/kubernetes-sigs/karpenter/pull/3004/changes#diff-e155309e90cf5e289192e00c6fa8a845bd97c4bbf39403c43badb66d324e66b3R87

GnatorX:
Update doc to deprecate price + priceAdjustment to just price expression + fix the PR description expression (self.price)

designs/node-overlay-price-adjustments.md
```diff
@@ -0,0 +1,231 @@
...
+spec:
+  weight: 100
+  requirements:
+    - key: karpenter.sh/capacity-type
+      operator: In
+      values: ["on-demand"]
+  priceExpression: "(self.price * 0.9 + 0.05) * 1.03"
+```
+
+The three-factor example from the motivation section (−10% EDP, +$0.05 agent fee, +3% AZ surcharge) is expressed as a single formula. The operator controls order of operations directly via parenthesization.
```
AlexanderYastrebov: Would it be too bad to allow overlay merging via price expression chaining such that `self.price` input is the output of an overlay expression of a lower weight? 
Then this example could be expressed as three overlays (potentially with different requirements):
1. `priceExpression: "self.price * 0.9"  # −10% EDP, self.price is cloud provider's offering price`
2. `priceExpression: "self.price + 0.05" # +$0.05 agent fee, self.price is output of 1.`
3. `priceExpression: "self.price * 1.03" # +3% AZ surcharge, self.price is output of 2.`

This also enables usecase like global capacity type discount plus per-instance type cost penalty based on some preference model (https://github.com/kubernetes-sigs/karpenter/issues/2616#issuecomment-4845163532)

AlexanderYastrebov: Direct price value can be seen as `self.price * 0 + 0.50`

GnatorX: I think the worry is pricing would be the only field that resolves in this way (resolving via merging vs highest weight wins). 

I don't think its necessarily a bad idea but i think we want to make node overlay support this in a wider(?) manner so its not confusing for users where different fields have different behavior of resolution?

GnatorX:
Realigned the doc to RFC template, added graduation criterias

designs/node-overlay-price-adjustments.md
DerekFrank: The benefit of CEL is that you don't need to reason about merging. We either pick merging or we pick CEL I think. I'm curious to hear which you guys think is more ergonomic for your usecases. I don't know how much different expressions need to be sliced and diced across different node selectors.

If users need to be able to say '20% reduction on this instance family' and '- 0.03 on this instance size' and have karpenter know that means `.8x -0.03' for instances that overlap then I think merging is more ergonomic.

FWIW, I'm not opposed to both especially during alpha where people can try it out and find out what they like

GnatorX: I think, imo, this can come later? I think stacking is a bigger feature due to how it changes how overlay works in a more profound way? vs just changing a field within overlay

DerekFrank (commented):
Comments from working group:
* We can add layering later if we need it, we shouldn't block on that
* additive priceExpression field for now, drop it in beta or ga
* drop `self.`

GnatorX:
Updated the doc based on working group comments

PR #3005, opened 2026-05-08
chore: Create kwok/util randomname to fix e2e tests

AndrewMitchell25:
Fixes #N/A 

**Description**
The kind e2e and perf tests are currently failing ([Example](https://github.com/kubernetes-sigs/karpenter/actions/runs/25080335506/job/73483629189)) because the kwok karpenter controllers are never coming up. This is due to a recent change that updated to use `test.RandomName()`, which unknowingly imported `controller-runtime/pkg/envtest`, which imports `internal/testing/addr`, which has an `init()` that does `os.MkdirAll("/tmp/kubebuilder-envtest")`, which causes a panic on the startup of the controller since in the perf tests the pod has `readOnlyRootFilesystem: true`.

Also bumps go version for CVEs

CI run - https://github.com/AndrewMitchell25/karpenter/actions/runs/25569933358/job/75062806308


joshuakguo:
it looks good to me, we can just pull in this function into the aws provider and use it directly

DerekFrank (approved):
/lgtm

PR #3006, opened 2026-05-08
feat: consolidateAfter now applies to destination nodes during consolidation

### LINKED ISSUE #2705: Karpenter consolidateAfter is not working as expected
### Description
For consolidateAfter, can we make -> consolidateAfter is the amount of time Karpenter should wait to have a node as part of the consolidation process. The amount of time Karpenter has to wait until pods can move **into** the node and **out** of the node due to consolidation. What we want is that when a node has a pod movement (pod added or removed for any reason - k8s scale up/down, consolidation, drift, etc.) then this node is invisible for the Karpenter underutilized consolidation process

- E.g How can we avoid nodes are disturbed before the 30min(consolidateAfter setting) configuration?

- E.g We want to know what is the logic behind picking a node for consolidation, for those perfectly packed nodes, can we avoid picking them for consolidation before 30min(the consolidateAfter period)?It is ok to consolidate these nodes after 30min.

- E.g We also concerned the new nodes who joined in the cluster recently can get consolidated in a short period of time again(it's OK to consolidate them after 30min)

**Observed Behavior**:
We have the following settings:
```
disruption:
    consolidateAfter: 30m
    budgets:
    # Drifted nodes limited to terminate 10% in parallel
       - nodes: "10%"
       reasons:
         - Drifted
    # Underutilized nodes always compacting to avoid a low packing growth
      - nodes: "5%"
      reasons:
        - Underutilized
    # Empty nodes can be terminated any time

```

Image

- Some well-packed nodes get interrupted  
- Several pods are disrupted unnecessarily. 
- The interruption is too frequent even if we set consolidateAfter: 30m

**Expected Behavior**:

Image

- We expect to see only 3 nodes get consolidated at this moment since they are empty now, other nodes and pods get untouched

Nodepool yaml
```
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: aaa
spec:
    disruption:
      consolidateAfter: 30m
      budgets:
      # Drifted nodes limited to terminate 10% in parallel
         - nodes: "10%"
         reasons:
           - Drifted
      # Underutilized nodes always compacting to avoid a low packing growth
        - nodes: "5%"
        reasons:
          - Underutilized
      # Empty nodes can be terminated any time
    template:
      spec:
        expireAfter: Never
        nodeClassRef:
          group: karpenter.k8s.aws
          kind: EC2NodeClass
          name: nearcorepool2
        requirements:
        - key: kubernetes.io/arch
          operator: In
          values:
          - arm64
          - amd64
          operator: In
          values:
          - on-demand
          operator: In
          values:
          - m7i.8xlarge
        taints:
        - effect: NoSchedule
          key: dedicated
          value: aaa
```

In the workload, we just set the toleration to use this nodepool
```
  Node-Selectors:  nodepool=aaa
  Tolerations:  dedicated=aaa:NoSchedule
```

**Versions**:
- Chart Version: 1.3.3
- Kubernetes Version (`kubectl version`):  The bug is not related to k8s version.

DerekFrank:
Fixes #2705, https://github.com/aws/karpenter-provider-aws/issues/7146

**Description**

`consolidateAfter` now applies to destination nodes during consolidation scheduling. Previously it only controlled when a node became eligible as a disruption *source*. Now the scheduler also refuses to place non-pending pods onto nodes still within their `consolidateAfter` window, preventing consolidation churn where recently-created or recently-modified nodes immediately become targets for the next consolidation round.

New unit test in `consolidation_test.go` verifying that consolidation is blocked when the destination node is under its `consolidateAfter` window, and proceeds after the window expires.

### DIFF
--- a/pkg/controllers/disruption/consolidation_test.go
+++ b/pkg/controllers/disruption/consolidation_test.go
@@ -3033,6 +3033,201 @@ var _ = Describe("Consolidation", func() {
 			})
 			Expect(ok).To(BeTrue())
 		})
+		It("should not consolidate pods onto a destination node that is under its consolidateAfter window", func() {
+			// Override consolidateAfter to a nonzero duration.
+			// Both nodes are leastExpensiveInstance (from BeforeEach), so no cheaper replacement
+			// exists — the only consolidation path is to delete nodes[1] and move its pod
+			// to nodes[0]. With consolidateAfter still active on nodes[0], this should be blocked.
+			nodePool.Spec.Disruption.ConsolidateAfter = v1.MustParseNillableDuration("1m")
+
+			// create our RS so we can link a pod to it
+			rs := test.ReplicaSet()
+			ExpectApplied(ctx, env.Client, rs)
+			Expect(env.Client.Get(ctx, client.ObjectKeyFromObject(rs), rs)).To(Succeed())
+
+			pods := test.Pods(3, test.PodOptions{
+				ObjectMeta: metav1.ObjectMeta{Labels: labels,
+					OwnerReferences: []metav1.OwnerReference{
+						{
+							APIVersion:         "apps/v1",
+							Kind:               "ReplicaSet",
+							Name:               rs.Name,
+							UID:                rs.UID,
+							Controller:         lo.ToPtr(true),
+							BlockOwnerDeletion: lo.ToPtr(true),
+						},
+					}}})
+			ExpectApplied(ctx, env.Client, rs, pods[0], pods[1], pods[2], nodeClaims[0], nodes[0], nodeClaims[1], nodes[1], nodePool)
+
+			// bind pods to node — 2 on nodes[0], 1 on nodes[1]
+			ExpectManualBinding(ctx, env.Client, pods[0], nodes[0])
+			ExpectManualBinding(ctx, env.Client, pods[1], nodes[0])
+			ExpectManualBinding(ctx, env.Client, pods[2], nodes[1])
+
+			// Initialize both nodes — this sets the Initialized condition timestamp to ~env.Clock.Now()
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, env.Clock, nodeStateController, nodeClaimStateController, []*corev1.Node{nodes[0], nodes[1]}, []*v1.NodeClaim{nodeClaims[0], nodeClaims[1]})
+
+			// Do NOT advance clock — both nodes are within the 1m consolidateAfter window.
+			// The scheduler should refuse to place non-pending pods onto nodes[0] because it's under consolidateAfter,
+			// blocking the delete of nodes[1] (which would need to reschedule pods[2] onto nodes[0]).
+			ExpectSingletonReconciled(ctx, disruptionController)
+
+			// No consolidation should happen — destination node is under consolidateAfter
+			cmds := queue.GetCommands()
+			Expect(cmds).To(HaveLen(0))
+			Expect(ExpectNodes(ctx, env.Client)).To(HaveLen(2))
+
+			// Advance clock past the 1m consolidateAfter window
+			env.Clock.Step(2 * time.Minute)
+
+			// Re-sync cluster state and re-mark nodeClaims as consolidatable after state update
+			ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated(ctx, env.Client, env.Clock, nodeStateController, nodeClaimStateController, []*corev1.Node{nodes[0], nodes[1]}, []*v1.NodeClaim{nodeClaims[0], nodeClaims[1]})
+			for _, nc := range nodeClaims {
+				nc = ExpectExists(ctx, env.Client, nc)
+				nc.StatusConditions().SetTrue(v1.ConditionTypeConsolidatable)
+				ExpectApplied(ctx, env.Client, nc)
+				ExpectReconcileSucceeded(ctx, nodeClaimStateController, client.ObjectKeyFromObject(nc))
+			}
+			// Reset the disruption controller so consolidation methods are not cached as consolidated
+			*queue = lo.FromPtr(disruption.NewQueue(env.Client, recorder, cluster, env.Clock, prov))
+			disruptionController = disruption.NewController(env.Clock, env.Client, prov, cloudProvider, recorder, cluster, queue,
+				disruption.WithMethods(NewMethodsWithNopValidator()...))
+			ExpectSingletonReconciled(ctx, disruptionController)
+
+			// Now consolidation should proceed — consolidateAfter window has expired
+			cmds = queue.GetCommands()
+			Expect(cmds).To(HaveLen(1))
+			ExpectObjectReconciled(ctx, env.Client, queue, cmds[0].Candidates[0].NodeClaim)
+			ExpectNodeClaimsCascadeDeletion(ctx, env.Client, nodeClaims[1])
+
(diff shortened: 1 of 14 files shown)


pkg/controllers/disruption/consolidation_test.go
```diff
@@ -3033,6 +3033,77 @@ var _ = Describe("Consolidation", func() {
 			})
 			Expect(ok).To(BeTrue())
 		})
+		It("should not consolidate pods onto a destination node that is under its consolidateAfter window", func() {
```
nathangeology, line 3036: Can we add tests for consolidateAfter: Never, consolidateAfter: 0s, testing drift's exemption, and the replacement path (not just the delete path).

nathangeology: There's some other tests that might be worth having like:

- What happens when LastPodEventTime is updated mid-sim? (Shouldn't happen since StateNodes are deep copies, but would be good to have a test documenting this and checking things in the event this changes)
- What happens when consolidateAfter is changed on the NodePool between computation and evaluation? (Answer: both use the current value at their point in time, which is good, but still good to have a test for this edge case).

nathangeology (commented):
Minor testing comments. Might try some experiments later and will update if those turn anything up we should worry about.

pkg/controllers/disruption/consolidation_test.go
DerekFrank: For the `0s` case — the entire existing Delete test suite already uses `consolidateAfter: 0s` in BeforeEach and succeeds, so that's implicitly covered.

For drift's exemption — drift_test.go's BeforeEach sets `consolidateAfter: 1h` and drift tests pass without advancing the clock, implicitly proving drift is not gated by consolidateAfter on destinations.

For scenarios 5 and 6 (LastPodEventTime changing mid-sim, consolidateAfter changed on NodePool between computation and validation): I think these are out of scope for this PR. The mid-sim case is protected by the deep-copy-at-simulation-time pattern (StateNodes are snapshots), and the computation-vs-validation case is protected by validation re-running SimulateScheduling with a fresh snapshot. Neither of these behaviors changed with this PR — they're emergent properties of the existing architecture. Happy to add them as a follow-up if you feel strongly.

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -114,6 +117,10 @@ var MinValuesPolicy = func(policy karpopts.MinValuesPolicy) func(*options) {
 	}
 }
 
+var EnforceConsolidateAfter = func(opts *options) {
```
jigisha620: nit: EnforceConsolidateAfter is a bit ambiguous since consolidateAfter is already enforced on source nodes. Could we make the name more specific to convey this applies to destinations? Something like `EnforceConsolidateAfterOnDestination`?

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -533,20 +548,24 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
...
 
 	// determine the volumes that will be mounted if the pod schedules
-	volumes, err := scheduling.GetVolumes(ctx, s.kubeClient, pod)
+	volumes, err := scheduling.GetVolumes(ctx, s.kubeClient, p)
 	if err != nil {
 		return err
 	}
 	parallelizeUntil(s.numConcurrentReconciles, len(s.existingNodes), func(i int) bool {
-		r, err := s.existingNodes[i].CanAdd(pod, s.cachedPodData[pod.UID], volumes)
+		if s.existingNodes[i].isUnderConsolidateAfter && !pod.IsPending(p) {
```
jigisha620: SimulateScheduling should feed three types of pods into the scheduler - pending pods, candidate pods and pods from deleting nodes. The check here exempts pending pods, but deleting-node pods are also running, so they'd be blocked from nodes under consolidateAfter too. Is this intentional?

pkg/utils/disruption/disruption.go
```diff
@@ -75,3 +75,19 @@ func ReschedulingCost(ctx context.Context, pods []*corev1.Pod) float64 {
...
 }
+
+func IsUnderConsolidateAfter(nodePool *v1.NodePool, nodeClaim *v1.NodeClaim, c clock.Clock) bool {
+	if nodePool == nil || nodeClaim == nil || nodePool.Spec.Disruption.ConsolidateAfter.Duration == nil || lo.FromPtr(nodePool.Spec.Disruption.ConsolidateAfter.Duration) == 0 {
+		return false
+	}
+	initialized := nodeClaim.StatusConditions().Get(v1.ConditionTypeInitialized)
+	if initialized == nil || !initialized.IsTrue() {
+		return false
+	}
```
jigisha620:
```suggestion
	if !nodeClaim.StatusConditions().IsTrue(v1.ConditionTypeInitialized) {
        return false
    }
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -533,20 +548,24 @@ func (s *Scheduler) add(ctx context.Context, pod *corev1.Pod) error {
...
+			// We shouldn't try to schedule pods that aren't pending onto nodes that are under consolidate after.
```
jigisha620: nit:
```suggestion
			// We shouldn't try to schedule pods that aren't pending onto nodes that are under consolidateAfter.
```

pkg/utils/disruption/disruption.go
```diff
@@ -75,3 +75,19 @@ func ReschedulingCost(ctx context.Context, pods []*corev1.Pod) float64 {
 	}
 	return cost
 }
+
+func IsUnderConsolidateAfter(nodePool *v1.NodePool, nodeClaim *v1.NodeClaim, c clock.Clock) bool {
```
jigisha620, line 79: nit: Worth adding a short descriptive comment for this function?

pkg/controllers/provisioning/scheduling/scheduler.go
DerekFrank: Yes, because the destination node may not have been under consolidate after when the deleting pod's node was deleted

pkg/utils/disruption/disruption.go
DerekFrank: What information not contained in the function signature would you include?

pkg/controllers/provisioning/scheduling/scheduler.go
DerekFrank: Yea I am not sure I'm super pleased with the flag. We need to opt-out drift because there isn't another way to tell if the simulation is for drift or consolidation. Technically the flag means 'as a destination', but the simulation doesn't enforce consolidateAfter on targets, that happens during candidate selection. I thought about making the flag 'IsConsolidationSimulation', maybe thats better?

pkg/controllers/provisioning/scheduling/scheduler.go
jigisha620: If pending pods are allowed on nodes under consolidateAfter because that reflects what kube-scheduler will actually do, shouldn't deleting-node pods be allowed too? Once evicted, they become pending and kube-scheduler will place them on those nodes regardless.

pkg/utils/disruption/disruption.go
jigisha620: Fair, the inline comment covers it.

jigisha620 (approved):
/lgtm

PR #3008, opened 2026-05-08
feat(scheduling): add opt-in NodeClaim optimization pass

mmanders-amzn:
Fixes #N/A

**Description**

Adds an **opt-in post-scheduling optimization pass** that detects when a NodeClaim's greedy, one-pass instance-type growth left earlier pods on a machine larger than they need, and splits the claim at the transition point. Displaced pods are re-queued and re-scheduled onto a fresh, right-sized NodeClaim.

The full design, tradeoffs, and correctness argument are proposed separately as an RFC in #3133. The summary below is for reviewer orientation.

### Problem

Karpenter's scheduler can commit a batch of pods to a single NodeClaim whose instance type is much larger — and much more expensive — than most of those pods actually need. The cause is single-NodeClaim packing: `Solve` processes pods in sort order, grows each new NodeClaim's instance-type set as pods are added, and never revisits a placement within the same call. As pods accumulate, the cheapest fitting instance type steps up through tiers; once it steps up, the claim stays at that larger instance type for the rest of `Solve`, regardless of whether an earlier subset would have packed cleanly on the previous tier. A single NodeClaim often ends up holding a set of pods that would be cheaper split across two claims of different tiers.

For example, 16.5 vCPU of pods land on a single 32 vCPU instance at ~51% efficiency. The same pods split as 16 vCPU on a 16 vCPU instance and 0.5 vCPU on a 2 vCPU instance (assuming 2 vCPU is the smallest available tier) run at ~91% efficiency on cheaper total capacity. The scheduler had enough information to choose the second layout; it simply never reconsidered once the 32 vCPU tier was locked in.

### Opt-in

Off by default. Production callers enable via the `NodeClaimOptimization` feature gate (`FEATURE_GATES=NodeClaimOptimization=true`). Tests and direct integrators opt in via the `scheduling.EnableNodeClaimOptimization` functional option.

### Core algorithm (`nodeclaim_optimization.go`)

- **Instance-type transition history**: `recordSchedulingOption` captures a `SchedulingOption` snapshot each time the cheapest compatible instance type changes during `NodeClaim.Add`. Consecutive pods that don't change the cheapest type replace the last entry in place, so history is sparse — one entry per transition, not one per pod.
- **Split-point detection**: `findSplitPoint` walks the history backwards looking for an earlier state that satisfies four gates:
  1. **Price savings** ≥ 24% of current (`minPriceSavingsRatio`)
  2. **Weighted efficiency gain** ≥ 10% (`minEfficiencyGain`)
  3. **Displaced-pod eligibility**: no displaced pod carries a required hostname-scoped constraint the cost estimator can't price correctly. Rejects (a) self-selecting `requiredDuringScheduling` pod anti-affinity or `DoNotSchedule` topology spread on `kubernetes.io/hostname` — forces one-per-node fan-out the single-NodeClaim estimate can't represent; and (b) `requiredDuringScheduling` pod affinity on hostname whose target isn't present in the displaced cohort — target is anchored to a locked survivor, so the displaced pod has no compatible host in the re-queue. Cross-label anti-affinity and `Preferred*` variants are intentionally allowed.
  4. **Net savings** after estimating displaced-pod placement cost > `costEpsilon` AND > 3% of current price (`minNetSavingsRatio`)
- **DRA guard**: NodeClaims holding a pod with ResourceClaim requests skip splitting entirely. With DRA, device requests move out of `pod.spec.resources.requests` and into ResourceClaims, where the displaced-pod cost estimator can't see them, so gate 4 could approve a split that raises total cost. Revert is independently unsafe for these claims: a committed device allocation has no inverse (`Allocation.Commit` is one-way), and `RevertTo` re-widens `InstanceTypeOptions` back to the snapshot superset, which would re-advertise instance types the allocator already released for that hostname. The check keys off the pod rather than the `IgnoreDRARequests` flag so it stays correct after DRA GA, and it is per-claim, so a skipped DRA claim doesn't prevent its siblings from splitting.
- **Reserved-offering guard**: NodeClaims holding reserved offerings in `ReservedOfferingModeStrict` skip splitting entirely. Splitting would fragment the reservation's accounting and strict mode would promote a post-split "no compatible reservation" into a scheduling failure. Correctness over savings.
- **`RevertTo`**: Rolls a NodeClaim back to a snapshot index, unwinding local fields, topology counts (via stored per-pod deltas, not re-derived from potentially-tightened requirements), host-port reservations, and reserved offerings (intersection with the snapshot's instance types).
- **Locking**: After inspection, `CanAdd` rejects further pods on the claim. This keeps displaced pods from re-landing on their source claim when they flow back through the queue, and prevents a subsequent pass from re-evaluating a decision that was correct at the time.
- **Pass limit**: At most 2 optimization passes per `Solve` call (`maxOptimizationPasses`). Cloud instance sizing roughly doubles per tier, so recoverable savings decay exponentially across passes.

### Topology primitives (`topology.go`, `topologygroup.go`)

- `Topology.Record` now returns `[]TopologyDelta` — the exact per-group domain increments performed for a pod. **Source-compatible**: return type went from `()` to `[]TopologyDelta`, so existing callers that ignore the return compile unchanged. The only other in-tree caller (`ExistingNode.Add`) doesn't need the deltas because existing nodes are never reverted.
- `Topology.Unrecord(deltas)` replays decrements from stored deltas. Safe even when `NodeClaim.Requirements` has tightened since `Record` was called — re-deriving domains from current requirements would decrement the wrong counters for earlier-added pods.
- `TopologyGroup.Unrecord(domains...)` decrements counts and re-inserts into `emptyDomains` at zero. Guards against negative counts.

### Integration points (`nodeclaim.go`, `scheduler.go`, `provisioner.go`)

- `NodeClaim` embeds `optimizationState` (enable flag, scheduling history, topology deltas, lock flag).
- `NodeClaim.Add` captures topology deltas and calls `recordSchedulingOption`. When the feature is disabled, `recordSchedulingOption` self-gates and no history is accumulated.
- `NodeClaim.CanAdd` short-circuits with an error when the claim is locked.
- `Scheduler.Solve` calls `s.tryOptimize(q)` when the pod queue drains, gated on `s.OptimizationEnabled`. Captures pre/post `OptimizationSnapshot` for test invariants and logging.
- `Provisioner.NewScheduler` appends `scheduling.EnableNodeClaimOptimization` to the scheduler options when the feature gate is on.
- **Disruption's simulated scheduler is deliberately not wired to the gate.** `pkg/controllers/disruption/consolidation.go` discards any simulation whose `Results.NewNodeClaims` has more than one entry ("we're not going to turn a single node into multiple candidates"). A NodeClaim-splitting pass inside a consolidation simulation would turn valid single-replacement plans into discarded multi-NodeClaim plans. Consolidation also already performs its own cost minimization.

### Feature gate (`pkg/operator/options/options.go`)

- New `FeatureGates.NodeClaimOptimization` field, default `false`, parsed from the standard `FEATURE_GATES` env/flag string.
- Mirror `*bool` override in `pkg/test/options.go` for test fixtures.

### Observability

`Scheduler.OptimizationSnapshot` holds `Pre`/`Post` NodeClaim summaries and `PreCost`/`PostCost` totals for a single `Solve` call. Used by the rapid cost-invariant property test and also emitted as a single summary log line per `Solve` when the pass ran.

### Correctness tests

- **Unit tests** (`nodeclaim_optimization_test.go`, external package, Ginkgo): split detection on large instance jumps, no-op on already-efficient claims, no-op with <2 instance-type options, feature-toggle correctness, correctness invariants (every pod on exactly one NodeClaim; no split when displaced pods need the same tier; no split with a single instance type; never increases cost across multiple transitions).
- **Internal tests** (`nodeclaim_optimization_internal_test.go`, `package scheduling`): direct verification of `findSplitPoint`'s DRA guard (DRA pod in the retained prefix, DRA pod in the displaced set, and per-claim scope so siblings still split) and strict-mode guard (with/without held reservations, fallback mode), `RevertTo` field restoration, topology count reversal under tightening requirements, host-port cleanup, reservation intersection (hold-drop-reacquire retains both; release-current-not-in-snapshot frees the seat), double-revert no-op via lock.
- **TopologyGroup tests** (`topologygroup_test.go`): `Unrecord` decrement, reinsert at zero, unknown-domain no-op, negative guard, variadic, symmetric with `Record`.

### Property-based tests (`nodeclaim_optimization_rapid_test.go`, `nodeclaim_optimization_scaling_test.go`)

Behind `//go:build rapid`, so default `go test ./pkg/...` and CI don't pick them up. Invoked via `go test -tags rapid ./pkg/controllers/provisioning/scheduling/...`.

- **Cost property**: 1–200 random pods (0.25–8 CPU, variable memory), KWOK instance types ≤64 vCPU. Asserts `optimized cost ≤ baseline cost` (within `costEpsilon`).
- **Diverse-constraint property**: 2–20 workload groups × 1–50 replicas with randomized zone affinity, topology spread, pod anti-affinity, host ports, and pod affinity. Asserts `post cost ≤ pre cost × 1.02` per run and pod conservation (`scheduled + errors == total`).
- **Scaling benchmarks**: a deterministic, fixed-seed sweep over a ladder of pod counts that times baseline (gate off) against optimized (gate on) solves on identical pod sets, for local characterization of the pass's cost. Timings are reported, never asserted.

Both tests include an opt-in CSV/JSONL diagnostic exporter (activated via `TEST_OUTPUT_DIR`) that emits per-run cost summaries, NodeClaim details, pod details, and pod configs for offline analysis. The helpers live in the same file and are no-ops when the env var is unset.

### Empirical results

Measured on KWOK instance types ≤64 vCPU. Both tests report *pre → post* (cost at the first-optimization-pass entry vs. final, inside the same `Solve`), so the delta isolates the pass itself without cross-Solve nondeterminism.

**Cost property — 250 runs:**
```
=== COST SUMMARY: 250 runs ===
  pre → post:  96 cheaper,   0 costlier | cost 778.78 → 763.11 | saved +15.67 (+2.0%) | NCs 707 → 828
    Of runs that got cheaper (96): cost 357.71 → 342.04 | saved 15.67 (4.4%) | avg saved 0.16/run (12.2%) | NCs 307 → 428
  250 of 250 runs triggered the optimization pass (100%)
```

**Diverse-constraint property — 253 runs:**
```
=== DIVERSE SUMMARY: 253 runs ===
  pre → post: 108 cheaper,   0 costlier | cost 1260.96 → 1206.64 | saved +54.31 (+4.3%) | NCs 8086 → 8255
    Of runs that got cheaper (108): cost 672.85 → 618.53 | saved 54.31 (8.1%) | avg saved 0.50/run (10.4%) | NCs 3661 → 3830
  253 of 253 runs triggered the optimization pass (100%)
```

Zero costlier runs in either suite across 503 total property-test invocations — the `optimized cost ≤ baseline cost` invariant held on every run.

### Files changed

| File | Lines | Purpose |
|------|-------|---------|
| `pkg/controllers/provisioning/scheduling/nodeclaim_optimization.go` | +792 | Core optimization logic |
| `pkg/controllers/provisioning/scheduling/nodeclaim_optimization_internal_test.go` | +792 | Internal unit tests |
| `pkg/controllers/provisioning/scheduling/nodeclaim_optimization_rapid_test.go` | +988 | Property tests + CSV/JSONL diagnostic exporter (build-tag `rapid`) |
| `pkg/controllers/provisioning/scheduling/nodeclaim_optimization_test.go` | +480 | Ginkgo behavioral tests |
| `pkg/controllers/provisioning/scheduling/nodeclaim_optimization_scaling_test.go` | +953 | Scaling benchmarks (build-tag `rapid`) |
| `pkg/controllers/provisioning/scheduling/topologygroup_test.go` | +148 | `TopologyGroup.Unrecord` tests |
| `pkg/controllers/provisioning/scheduling/topology.go` | +60 | `TopologyDelta`, `Record` return, `Unrecord` |
| `pkg/controllers/provisioning/scheduling/topologygroup.go` | +20 | `TopologyGroup.Unrecord` |
| `pkg/controllers/provisioning/scheduling/scheduler.go` | +28 | Option, fields, `Solve` integration |
| `pkg/controllers/provisioning/scheduling/nodeclaim.go` | +11 | `optimizationState` embed, `Add`/`CanAdd` hooks |
| `pkg/controllers/provisioning/provisioner.go` | +3 | Feature-gate wire-up at scheduler construction |
| `pkg/operator/options/options.go` | +7 | `NodeClaimOptimization` feature gate + parsing |
| `pkg/operator/options/suite_test.go` | +12 | Feature-gate coverage in existing suite |
| `pkg/test/options.go` | +2 | Test `*bool` override for the gate |
| `go.mod` / `go.sum` | +3 | `pgregory.net/rapid` |
| `.gitignore` | +2 | `test_output/`, `testdata/` |


jigisha620:
Have you considered how this interacts with DRA (Dynamic Resource Allocation)? With DRA, GPU requests move from `pod.spec.resources.requests` to ResourceClaims, which would make them invisible to the displaced-pod cost estimator.

Specifically, a displaced pod with requests: {cpu: 1, memory: 8Gi} plus a ResourceClaim for a GPU would be estimated as fitting on a small general-purpose instance, but actual placement requires a GPU instance at significantly higher cost. This could cause gate 4 to approve splits where the actual total cost increases, violating the stated invariant.

Even without DRA, for the efficiency metric being CPU/memory-only means gate 2 will tend to block valid splits on GPU instances, since both states look "underutilized" regardless of GPU utilization.

Is DRA in scope for the initial implementation, or something to track as a follow-up?

mmanders-amzn:
> Have you considered how this interacts with DRA (Dynamic Resource Allocation)? With DRA, GPU requests move from `pod.spec.resources.requests` to ResourceClaims, which would make them invisible to the displaced-pod cost estimator.
> 
> Specifically, a displaced pod with requests: {cpu: 1, memory: 8Gi} plus a ResourceClaim for a GPU would be estimated as fitting on a small general-purpose instance, but actual placement requires a GPU instance at significantly higher cost. This could cause gate 4 to approve splits where the actual total cost increases, violating the stated invariant.
> 
> Even without DRA, for the efficiency metric being CPU/memory-only means gate 2 will tend to block valid splits on GPU instances, since both states look "underutilized" regardless of GPU utilization.
> 
> Is DRA in scope for the initial implementation, or something to track as a follow-up?

We might ignore nodes with pods that have DRA requirements at the beginning. The current heuristic does make the assumption that the future (split) NodeClaim will be of the same NodePool as the one it was split from. As long as the check filters for DRA requirements when checking whether the pods can land on an instance - it should pick an instance that includes these additional requirements. The instance pricing would include the cost of the additional resources, if the cost of the two expected NodeClaims was more expensive than the original single NodeClaim, splitting should be skipped.  However, that check (for additional requirements) is not included at the moment.

GnatorX:
(Garvin here). Chatted during working group wanted to add a few points from the discussion into the thread.

- Love the idea of doing a second pass to improve correct sizing of nodes however concerned about performance of pod getting scheduled onto node due to reprocessing.
- Want to see some way to confirm improvement in efficiency.
- Eventually would like to see preprocessing of grouping of pods, not relying on Nodeclaim, then creating nodeclaim from there
- Pricing of nodes where individual nodes may have fixed cost that may be cause efficiency to flip if we spread too many nodes (self call out on https://github.com/kubernetes-sigs/karpenter/pull/3004)

pkg/controllers/provisioning/scheduling/nodeclaim_optimization.go
```diff
@@ -0,0 +1,809 @@
...
+// n.InstanceTypeOptions (and its price) under n.Requirements. The options
+// have already been filtered for compatibility and fit at Add time, so this
+// is a straight price reduction, no re-check.
+//
+// Returns (nil, +Inf) for a claim with an empty InstanceTypeOptions list.
+//
+// Distinct from cheapestFit, which is for hypothetical placements against
+// an arbitrary instance-type list and re-runs the compatibility and fit
+// checks.
+func (n *NodeClaim) CheapestInstance() (*cloudprovider.InstanceType, float64) {
```
DerekFrank, line 280: Is there an existing .Cheapest() we can call?

pkg/controllers/provisioning/scheduling/nodeclaim_optimization.go
```diff
@@ -0,0 +1,809 @@
...
+// been reached. Returns true if pods were displaced. Caller is expected to
+// have gated on Scheduler.OptimizationEnabled; no internal feature-flag
+// check here.
+//
+// On first entry (OptimizationPasses == 0), captures the Pre half of
+// OptimizationSnapshot (total price and per-NodeClaim state) before any
+// revert mutates the claim set, so tests can compare the post-Solve cost
+// against this baseline to verify each split decision paid off against its
+// own estimate.
+func (s *Scheduler) tryOptimize(q *Queue) bool {
```
DerekFrank, line 771:
```suggestion
func (s *Scheduler) tryOptimize() []Pods{
```

then in solve you do:

```
pods := s.tryOptimize()
if len(pods) != 0 {
 ....
}
```

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -436,7 +454,7 @@ func (r Results) TruncateInstanceTypes(ctx context.Context, maxInstanceTypes int
 	return r
 }
 
-//nolint:gocyclo
+//nolint:gocyclo // Solve is the scheduler loop; its branches are the distinct phases (topology update, pod pop, trySchedule error handling, optimization pass, finalize). Splitting them hurts readability.
 func (s *Scheduler) Solve(ctx context.Context, pods []*corev1.Pod) (Results, error) {
 	defer metrics.Measure(DurationSeconds, map[string]string{ControllerLabel: injection.GetControllerName(ctx)})()
```
DerekFrank, line 459: Add a defer to emit the metrics you want out of the optimization passes

DerekFrank, line 459: You can steal this metrics.Measure() to measure the runtime seconds of your tryOptimize loop to know if you have perf there

pkg/controllers/provisioning/scheduling/nodeclaim_optimization.go
```diff
@@ -0,0 +1,809 @@
...
+// requests against an instance type's allocatable capacity.
+type ResourceEfficiencyResult struct {
+	CPU      float64
+	Memory   float64
+	Weighted float64
+}
+
+// ResourceEfficiency computes utilization ratios of requests against the given
+// instance type's allocatable resources.
+func ResourceEfficiency(requests corev1.ResourceList, it *cloudprovider.InstanceType) ResourceEfficiencyResult {
```
DerekFrank, line 96: Is there an existing function we can use?

PR #3011, opened 2026-05-09
chore: add reason to consolidatable status condition removal log

### LINKED ISSUE #2238: Add more context to "removing consolidatable status condition" message

praveen9354:
Fixes #2238

## Description

Brings the bare 'removing consolidatable status condition' log line into
parity with its two siblings (which already explain themselves with
', consolidation is disabled' and ', isn't initialized'). Adds structured
fields (lastPodEventTime, consolidateAfter, timeSincePodEvent) so
operators can correlate the removal against pod-level events visible
from kubectl get events.

Fixes #2238

Signed-off-by: praveen9354 <praveen9354@gmail.com>

Local verification of the changed file's adjacent test suite.
Upstream CI runs the full matrix on push.

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/consolidation.go
+++ b/pkg/controllers/nodeclaim/disruption/consolidation.go
@@ -42,7 +42,7 @@ func (c *Consolidation) Reconcile(ctx context.Context, nodePool *v1.NodePool, no
 	if nodePool.Spec.Disruption.ConsolidateAfter.Duration == nil {
 		if hasConsolidatableCondition {
 			_ = nodeClaim.StatusConditions().Clear(v1.ConditionTypeConsolidatable)
-			log.FromContext(ctx).V(1).Info("removing consolidatable status condition, consolidation is disabled")
+			log.FromContext(ctx).V(1).Info("removing consolidatable status condition", "reason", "consolidation is disabled")
 		}
 		return reconcile.Result{}, nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/nodeclaim/disruption/consolidation.go
```diff
@@ -63,7 +63,11 @@ func (c *Consolidation) Reconcile(ctx context.Context, nodePool *v1.NodePool, no
 	if c.clock.Since(timeToCheck) < lo.FromPtr(nodePool.Spec.Disruption.ConsolidateAfter.Duration) {
 		if hasConsolidatableCondition {
 			_ = nodeClaim.StatusConditions().Clear(v1.ConditionTypeConsolidatable)
-			log.FromContext(ctx).V(1).Info("removing consolidatable status condition")
+			log.FromContext(ctx).V(1).Info("removing consolidatable status condition, consolidateAfter window not yet elapsed",
```
DerekFrank: Can we move the reason into another log field for this and the others?


```suggestion
			log.FromContext(ctx).V(1).Info("removing consolidatable status condition",
			"reason",  "consolidateAfter window not yet elapsed",
			...
```

DerekFrank (changes requested):

pkg/controllers/nodeclaim/disruption/consolidation.go
praveen9354: Good call — done in ed98cac. Moved the reason into a dedicated `"reason"` key/value across all three removal sites in `consolidation.go`. Keeps the message constant for grep, and log shippers can index on the reason value cleanly.

DerekFrank (approved):
/lgtm 

🚀 Thanks for the contribution! 🚀

PR #3012, opened 2026-05-09
feat(metrics): add karpenter_pods_disrupted_total counter

### LINKED ISSUE #2020: Add `karpenter_pods_disrupted_total` when NodeClaims are disrupted

praveen9354:
Fixes #2020

## Description

Mirrors karpenter_nodeclaims_disrupted_total but counts the reschedulable
pods that were on each disrupted node. Operators can now correlate the
NodeClaim-level disruption volume with the pod-level user impact across
the same (reason, nodepool, capacity_type) labels.

Emit sites match the four NodeClaim disruption paths where pod records
exist on the node:

- pkg/controllers/disruption/queue.go uses Candidate.reschedulablePods
  directly (already cached on the Candidate)
- pkg/controllers/nodeclaim/expiration/controller.go,
  pkg/controllers/node/health/controller.go, and
  pkg/controllers/nodeclaim/garbagecollection/controller.go list pods
  via the new node.CountReschedulablePodsOnNode helper, which applies
  the same pod.IsReschedulable filter (active or stateful-set
  terminating; not DaemonSet- or Node-owned)

Skipped: nodeclaim/lifecycle/launch.go and liveness.go disrupt
NodeClaims that never bound pods (launch failure, liveness watchdog),
so a pods counter on those paths would always be 0.

A list error during pod counting is logged at V(1) and the metric
reports 0 for that disruption — never blocks the disruption itself.

Refs #2020

Signed-off-by: praveen9354 <praveen9354@gmail.com>

Local verification of the changed file's adjacent test suite.
Upstream CI runs the full matrix on push.

### DIFF
--- a/pkg/controllers/disruption/queue.go
+++ b/pkg/controllers/disruption/queue.go
@@ -238,11 +238,13 @@ func (q *Queue) waitOrTerminate(ctx context.Context, cmd *Command) (err error) {
 			return
 		}
 		q.recorder.Publish(disruptionevents.Terminating(cmd.Candidates[i].Node, cmd.Candidates[i].NodeClaim, string(cmd.Reason()))...)
-		metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
+		labels := map[string]string{
 			metrics.ReasonLabel:       pretty.ToSnakeCase(string(cmd.Reason())),
 			metrics.NodePoolLabel:     cmd.Candidates[i].NodeClaim.Labels[v1.NodePoolLabelKey],
 			metrics.CapacityTypeLabel: cmd.Candidates[i].NodeClaim.Labels[v1.CapacityTypeLabelKey],
-		})
+		}
+		metrics.NodeClaimsDisruptedTotal.Inc(labels)
+		metrics.PodsDisruptedTotal.Add(float64(len(cmd.Candidates[i].reschedulablePods)), labels)
 	})
 	// If there were any deletion failures, we should requeue.
 	// In the case where we requeue, but the timeout for the command is reached, we'll mark this as a failure.
--- a/pkg/controllers/node/health/controller.go
+++ b/pkg/controllers/node/health/controller.go
@@ -171,11 +171,19 @@ func (c *Controller) deleteNodeClaim(ctx context.Context, nodeClaim *v1.NodeClai
 	}
 	// The deletion timestamp has successfully been set for the Node, update relevant metrics.
 	log.FromContext(ctx).Info("deleting unhealthy node")
-	metrics.NodeClaimsDisruptedTotal.Inc(map[string]string{
+	labels := map[string]string{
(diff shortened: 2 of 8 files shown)


praveen9354:
`presubmit (1.30.x)` flake — the failed test is in `pkg/controllers/node/termination` (untouched by this PR), and 6/7 k8s versions are green. Same job passed on #3011 just before this run.

/retest

DerekFrank:
Does this metric not satisfy your requirements? 


https://karpenter.sh/docs/reference/metrics/#karpenter_pods_drained_total
https://github.com/kubernetes-sigs/karpenter/blob/660279ee54ddb403a35aad36a38bca20a6e9ba9d/pkg/controllers/node/termination/terminator/metrics.go#L45


Please re-open if it doesn't

praveen9354:
Thanks for the careful look @DerekFrank. After comparing both metric definitions, I think `pods_drained_total` and the proposed `pods_disrupted_total` are adjacent rather than overlapping. Three concrete cases the existing `pods_drained_total` doesn't cover:

1. **Pods that fail to evict** (PDB violations, repeated 429s, ungraceful terminations) — `pods_drained_total` only increments on the successful 200 path in `terminator/eviction.go:215`. The `pods_disrupted_total` proposal counts at NodeClaim-disruption decision time, so it captures the intended disruption regardless of eviction outcome. The ratio `pods_drained_total / pods_disrupted_total` is the eviction success rate per nodepool/reason, which neither metric alone can express.

2. **Ungraceful node loss** (`node/health/controller.go` GC path) — when an unhealthy node's NodeClaim is GC'd, the terminator never runs, so `pods_drained_total` is structurally blind to those pod disruptions. `pods_disrupted_total` catches them because it increments in the same 4 controllers that already increment `nodeclaims_disrupted_total`.

3. **Label cardinality parity with `nodeclaims_disrupted_total`** — that metric already exists with `(reason, nodepool, capacity_type)`. Operators can answer "how many NodeClaims were disrupted per nodepool" but not the corresponding pod-level question with matching labels — `pods_drained_total` is `(reason)` only.

Happy to drop the `capacity_type` label if cardinality is the concern, or to rename for clarity (e.g., `pods_disruption_initiated_total` to make the timing difference explicit). Re-opening with the case above — let me know if you'd prefer either of those adjustments or want me to close again.

praveen9354:
Opened #3015 with the same code and the justification above in the description, since the reopen path was structurally blocked (GitHub disallows reopen after the head branch was cleaned up on close). Continuing the conversation there.

PR #3013, opened 2026-05-12
feat: feature-gated custom drain strategy based on 'rollout restart' for single replica deploy/sts.

### LINKED ISSUE #740: Custom drain flow

EvgeniGordeev:
Fixes #740. Codependent with https://github.com/aws/karpenter-provider-aws/pull/9158

## Rollout Restart Drain Strategy

An opt-in drain mode for single-replica workloads (deploys and statefulsets). Instead of evicting the only pod and taking downtime, Karpenter asks the owner to roll itself - a fresh pod comes up first, then the old one goes away.

- **Singleton deploys**: zero downtime - maxSurge brings the new pod up before the old one is deleted.
- **Singleton statefulsets**: brief downtime (no surge), but the drain bypasses blocking PDBs that would otherwise stall it.
- **Multi-replica workloads**: unchanged - still drain via eviction.

Turn it on with the `RolloutRestartDrainStrategy` feature gate (off by default).

### How it decides when to patch

We only trigger a rollout when the owner's `Status` says the previous one is done - same predicate `kubectl rollout status` uses. If a rollout is still in flight, we skip and emit a `RolloutRestartInProgress` event so you can see why the drain is waiting.

For deploys the predicate is:

```go
dep.Generation > dep.Status.ObservedGeneration ||
    dep.Status.UpdatedReplicas < *dep.Spec.Replicas ||
    dep.Status.Replicas > dep.Status.UpdatedReplicas ||       // surge cleanup still in flight
    dep.Status.AvailableReplicas < dep.Status.UpdatedReplicas
```

The `Replicas > UpdatedReplicas` term is the load-bearing one for singletons: once the new surge pod becomes Ready, `UpdatedReplicas == Spec.Replicas == 1` already, but the old pod hasn't been deleted yet - without this term we'd re-patch and clobber the rollout.

The first cut used a 2-minute timestamp dedup, which matched Karpenter's consolidation cadence and basically never fired. Slow-booting pods (~2+ min) got killed every cycle and never reached Ready. The status-based gate fixes that loop.

### Testing

- 8 unit tests in `pkg/controllers/node/termination/terminator/suite_test.go` - feature gate off/on, singleton vs multi-replica, deploys and statefulsets, plus three in-flight-rollout regression cases. `make presubmit` passes.
- Live cluster test with `egordeev/karpenter-controller-aws:1.12.1-rollout-restart2`:
  - Singleton deploy with blocking PDB (`minAvailable: 1`) → drained with zero downtime, PDB bypassed.
  - Singleton statefulset → drained with brief downtime, as expected.
  - Multi-replica deploy and statefulset → still evicted, unchanged.

### DIFF
--- a/designs/rollout-restart-drain-strategy.md
+++ b/designs/rollout-restart-drain-strategy.md
@@ -1,0 +1,345 @@
+# Rollout-Restart Drain Strategy
+
+## Summary
+
+This feature adds an opt-in drain strategy that, for singleton (`spec.replicas: 1`) Deployments and StatefulSets, replaces the pod-eviction call during node drain with a `kubectl rollout restart`
+–equivalent patch on the workload's pod template. The behavior is gated behind a new `RolloutRestartDrainStrategy` feature gate (default `false`) and only applies to pods whose owner is an apps/v1
+Deployment or StatefulSet with `spec.replicas == 1`. All other pods continue to drain via the existing eviction path.
+
+## Motivation
+
+### Problem Statement
+
+Karpenter drains nodes by calling the [eviction subresource](https://kubernetes.io/docs/concepts/scheduling-eviction/api-eviction/) on each non-DaemonSet pod. Eviction is the right primitive for the
+multi-replica case — it respects `PodDisruptionBudget` (PDB), so Kubernetes can guarantee a minimum of available replicas during the drain. For singleton workloads, however, eviction has two
+well-known shortcomings:
+
+1. **PDB deadlock on singletons.** A common policy is to set `minAvailable: 1` (or `maxUnavailable: 0`) on every workload's PDB so that disruption tooling can never voluntarily take the workload below
+   one Ready replica. For a Deployment or StatefulSet with `spec.replicas: 1`, that policy is *self-conflicting under eviction*: the eviction API will refuse the request indefinitely (HTTP 429,
+   `Cannot evict pod as it would violate the pod's disruption budget`). Karpenter retries forever, the node never drains, and the only way out is either to forcibly delete the pod (via
+   `terminationGracePeriod` expiry) — losing the PDB guarantee anyway — or for an operator to manually `kubectl rollout restart` the workload.
+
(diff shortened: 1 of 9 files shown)


DerekFrank:
/assign

Thanks for the contribution! I think this is a big enough change to warrant an RFC

EvgeniGordeev:
> /assign
> 
> Thanks for the contribution! I think this is a big enough change to warrant an RFC

An RFC added.

designs/rollout-restart-drain-strategy.md
```diff
@@ -0,0 +1,345 @@
+# Rollout-Restart Drain Strategy
+
+## Summary
+
+This feature adds an opt-in drain strategy that, for singleton (`spec.replicas: 1`) Deployments and StatefulSets, replaces the pod-eviction call during node drain with a `kubectl rollout restart`
```
DerekFrank, line 5: Unfortunately despite really liking this feature, I don't think it belongs in Karpenter.

I think this feature would be a fantastic feature in upstream kubernetes, part of eviction in general. 

There is at least one upstream request for this: https://github.com/kubernetes/kubernetes/issues/140599

If upstream k8s is willing to build surge eviction properly, I will ensure that Karpenter supports it right out of the gate

PR #3014, opened 2026-05-12
feat: dra device allocation tracking

jmdeal:
Fixes #N/A 

**Description**

The first in a sequence of PRs to support DRA in Karpenter. This PR adds a controller (`dynamicresources.deviceallocation`) responsible for tracking allocated devices based on `ResourceClaim` statuses. The controller provides an iterator interface which will be consumed by the provisioner and disruption controllers when creating the DRA allocator (PR to follow). 

Some design considerations:
- Multiple pods can consume the same ResourceClaim. To avoid overprovisioning, we need to release a device when all pods with reservations for that device are in a deleting state. A subset of pods deleting is not sufficient since it won't result in the device being released.
- This controller ensures that initial data is hydrated before returning any results. While this isn't strictly necessary since under counting allocated devices would just result in under provisioning on startup, making this deterministic isn't particularly complicated and eases reproducibility while testing.
- While ConsumableCapacity is out of scope of the initial DRA release, this implementation should be forward compatible. It supports tracking multiple claims allocated to the same device and taking the union of pods.

Additionally, I have staged changes to add a debug endpoint for querying internal device allocation state. I'm debating if this should live in the upstream project, but we've had several discussions about exposing similar endpoints for cluster state as the main mechanism for exposing high-cardinality data for issue diagnosis. I think this would be a good first-step in that direction: https://github.com/jmdeal/karpenter/pull/4

`make test`

### DIFF
--- a/pkg/controllers/dynamicresources/deviceallocation/controller.go
+++ b/pkg/controllers/dynamicresources/deviceallocation/controller.go
@@ -1,0 +1,240 @@
+package deviceallocation
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"iter"
+	"sync"
+	"time"
+	"unique"
+
+	"github.com/awslabs/operatorpkg/serrors"
+	"github.com/samber/lo"
+	corev1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/api/resource/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	utilscontroller "sigs.k8s.io/karpenter/pkg/utils/controller"
+)
+
+const (
+	minReconciles = 10
+	maxReconciles = 3000
+)
+
+type Controller struct {
+	kubeClient client.Client
+
+	mu               sync.RWMutex
+	allocatedDevices map[cloudprovider.DeviceID]Metadata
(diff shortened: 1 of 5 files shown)


pkg/controllers/dynamicresources/deviceallocation/controller.go
```diff
@@ -0,0 +1,254 @@
...
+}
+
+// Metadata contains supplementary information about an allocated device, derived from the ReservedFor status of all
+// ResourceClaims that reference it.
+type Metadata struct {
+	// Releasable is true when every ResourceClaim referencing the device has a non-empty ReservedFor list composed
+	// entirely of pod consumers. A device that is not reserved, or that is reserved by any non-pod consumer, is not
+	// releasable.
+	Releasable bool
+	// PodUIDs is the aggregate set of pod UIDs from the ReservedFor entries of all ResourceClaims that reference the
```
jigisha620, line 70: The comment here says "the aggregate set of pod UIDs" but the type is []types.UID and `computeDeviceMetadata` just appends without deduplication. If two claims for the same device are both reserved by the same pod, the UID appears twice. Should this be a sets.Set[types.UID] to match the documented semantics, or are consumers expected to handle duplicates?

jmdeal: I had kept this as a slice since membership is assumed to be small and iteration across a sufficiently small slice is typically cheaper than a set. You're right that this could result in duplicates (though it likely wouldn't in practice). I can add a comment, I'm not sure the deduplication logic is worth the CPU and / or memory overhead.

pkg/controllers/dynamicresources/deviceallocation/suite_test.go
```diff
@@ -0,0 +1,617 @@
...
+				seq, err := controller.AllocatedDevices(ctx)
+				Expect(err).ToNot(HaveOccurred())
+				devices := collectDevices(seq)
+				meta := devices[deviceID("device-0")]
+				Expect(meta.Releasable).To(BeFalse())
+				Expect(meta.PodUIDs).To(ConsistOf(types.UID("uid-a")))
+			})
+		})
+
+		Describe("Mutation", func() {
```
ryan-mist, line 556: nit: maybe a test for when a claim's reservations is cleared? (i.e. transition from reserved for pod -> nil ReservedFor)

pkg/controllers/dynamicresources/deviceallocation/controller.go
jigisha620: Makes sense, the performance tradeoff is reasonable for small membership. Updating the comment to clarify it's intentionally a slice (and that duplicates are possible but benign) would be sufficient.

jigisha620 (approved):
/lgtm

PR #3019, opened 2026-05-13
feat: cel expression price adjustment

GnatorX:
Fixes [3004](https://github.com/kubernetes-sigs/karpenter/pull/3004/)

**Description**

## Summary

This PR extends the NodeOverlay feature with CEL-based price expressions and improves observability for NodeOverlays applied to launched NodeClaims.

### CEL price expressions

Adds `spec.priceExpression` to NodeOverlay, allowing operators to define dynamic price adjustments using CEL (Common Expression Language). The expression receives the current offering price as `price` and returns the adjusted price as a float.

```yaml
spec:
  priceExpression: "self.price * 0.9"  # 10% discount
```

- Expression is validated at admission time and pre-evaluated during the NodeOverlay reconcile loop before being stored; evaluation errors at apply time cannot occur
- `PriceApplied=False/ExpressionEvaluationError` is set when an expression fails to evaluate during reconcile, distinguished from `NoMatchingInstanceTypes` when the overlay simply didn't match any offerings
- `PriceNonNegative=False` is set as an informational (non-Ready-blocking) condition when an expression produces a negative price; the NodeOverlay remains `Ready=True`
- Negative-price logging is done once per overlay in the reconcile path, not per offering per call to `Evaluate()`

### NodeClaim overlay observability

At launch time, the lifecycle controller annotates each NodeClaim with the names of all NodeOverlays that influenced its scheduling:

```
karpenter.sh/node-overlays-applied: my-price-overlay,my-capacity-overlay
```

Multiple overlays are comma-separated. The annotation is set once at launch and not re-written on subsequent reconciles. When no overlay applies (or the NodeOverlay feature gate is off), the annotation is not set.

The annotation is populated from the `InstanceTypeStore` — no additional cloud provider calls are made at launch time. The store records which overlay name was applied to each `(nodePool, instanceType, zone, capacityType)` combination during the preceding reconcile cycle.

## Observability

| Signal | Where | Notes |
|---|---|---|
| `karpenter.sh/node-overlays-applied` annotation | NodeClaim | Comma-separated list of overlay names applied at launch |
| `PriceApplied=True/False` status condition | NodeOverlay | False with `ExpressionEvaluationError` reason when CEL fails; `NoMatchingInstanceTypes` when no offerings matched |
| `PriceNonNegative=False` status condition | NodeOverlay | Informational only — does not block `Ready=True` |
| Controller log: `price expression evaluated to negative value` | nodeoverlay controller | Logged once per overlay per reconcile, not per offering |

  - pkg/apis/v1alpha1/cel — unit tests for Compile and Evaluate: invalid syntax, wrong return type, dyn passthrough, negative values, bare literals
  - pkg/controllers/nodeoverlay suite — PriceExpression Evaluation Error describe block:
    - Expression that compiles but fails at eval (self.notprice * 0.9) → ValidationSucceeded=False, reason ExpressionEvaluationError
    - Bare literal "0.50" → ValidationSucceeded=True, offering price = 0.50 via store.ApplyAll
    - High-weight eval-failing overlay does not block lower-weight valid price overlay (store contamination regression test)
  - PriceApplied and PriceNonNegative condition tests cover expression paths
  - All existing price/capacity/conflict tests continue to pass


kwok/charts/crds/karpenter.sh_nodeoverlays.yaml
DerekFrank, line 74: Do we just collapse priceAdjustment into priceExpression? The later strikes me as a superset

jmdeal: I think this is what we want to do for beta, maybe in beta the field is just `cost`. You can provide a constant, or an arbitrarily complex adjustment expression. This is inline with how we're thinking about supporting CEL in the EC2NodeClass in the AWS provider.

hakman:
@GnatorX this is an interesting feature that would be helpful with some of our current issues.
Do you plan to continue?

kwok/charts/crds/karpenter.sh_nodeoverlays.yaml
DerekFrank: Yea I like that

GnatorX: Ya i think this should be part of the graduation criterias

GnatorX:
@hakman yes I am. I want to get the RFC approved before I continue in case we have other changes

hakman:
> @hakman yes I am. I want to get the RFC approved before I continue in case we have other changes

Awesome, thanks!

PR #3023, opened 2026-05-18
fix: report all-pruned volume topology pods as unschedulable

tallaxes:
Follow-up to #2743, as discussed in https://github.com/kubernetes-sigs/karpenter/pull/2743#discussion_r3238450425

**Description**

This PR handles pods whose volume topology alternatives are all incompatible across multiple PVC-backed volumes. Instead of falling back to the old merged requirement result, Karpenter now treats this as an evaluated-but-unschedulable scheduling result.

Note: This changes the surfaced scheduling error for pods whose volume topology alternatives are all incompatible. These pods now report a single `incompatible volume topology requirements` error before queue processing, instead of later NodePool/NodeClaim compatibility failures such as `incompatible requirements`.

Key implementation decisions:

- Add scheduler support for initial pod errors that are known before queue construction.
- Seed initial pod errors into `Results.PodErrors` and skip those pods during cached pod data setup, queue construction, preference relaxation, topology updates, and node placement.
- Keep `NewScheduler()` and `Solve()` signatures stable by passing this optional state through `InitialPodErrors(...)`.
- Preserve `Results.PodErrors map[*corev1.Pod]error` because downstream metrics, state, logging, and eventing consume pod objects.
- Keep the all-pruned detection in volume topology merging, where the full cross-product is already available.

This preserves existing unschedulable accounting while ensuring invalid volume topology pods do not enter the scheduler queue without valid volume requirements.

**Implementation Rationale**

The main alternative was to push this into validation, for example through `Provisioner.Validate()` or `VolumeTopology.ValidatePersistentVolumeClaims()`. That would reject pods earlier, but the all-pruned case is only known after computing the cross-product of volume topology alternatives. Doing that in validation would either duplicate the work already done by `GetRequirements()` or move requirement generation into validation.

Using the existing validation path would also make these pods look like ignored pods from pending-pod filtering. That would affect `IgnoredPodCount` and bypass the normal scheduling-result path. This PR instead treats the failure as a scheduling decision: the pod appears in `Results.PodErrors`, updates unschedulable accounting, records scheduling decision state, and emits the same failure event path as other scheduler pod errors.

The scheduler option approach keeps this plumbing narrow. Initial pod errors are optional provisioning-path state, so `InitialPodErrors(...)` avoids changing `NewScheduler()` or `Solve()` signatures for callers such as disruption and consolidation that cannot produce these setup-time pod errors.

* `make presubmit`


jigisha620 (approved):
/lgtm

pkg/controllers/provisioning/scheduling/scheduler.go
```diff
@@ -379,6 +389,7 @@ func (r Results) TruncateInstanceTypes(ctx context.Context, maxInstanceTypes int
 	return r
 }
 
+//nolint:gocyclo
```
tallaxes, line 392: The most recent update from main pushed Solve past the preferred complexity budget. Opting for nolint:gocyclo for now, to ease review of logic changes in the PR; can follow-up with minor refactoring (maybe extracting `prepareSchedulingQueue` and/or `recordPendingPodsByEffectiveZone`) to reduce complexity.

jigisha620: Sounds good, wouldn't block on that.

jigisha620 (approved):
/lgtm

jigisha620 (approved):
/lgtm

PR #3024, opened 2026-05-18
chore: go fix ./...

tallaxes:
**Description**

* Run `go fix ./...` (https://go.dev/blog/gofix) https://github.com/kubernetes-sigs/karpenter/pull/3024/commits/33b1d05326921d9a4131669bde90440db90be3d2
* Update `golangci-lint` to the latest version (one that does not flag anything used in `new` as unused) https://github.com/kubernetes-sigs/karpenter/pull/3024/commits/77dfdc3edac18903f711cb6c1bbb4ad98233c65d
* Fix gosec "G703: Path traversal via taint analysis" flagged by the latest `golangci-lint` https://github.com/kubernetes-sigs/karpenter/pull/3024/commits/f9307b9609390e6a3b6f6294670715375164f468

**How was this change tested?**

* `make presubmit`

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3027, opened 2026-05-19
chore: add local replace for sigs.k8s.io/karpenter in go.tools.mod

DerekFrank:
Fixes https://github.com/kubernetes-sigs/karpenter/pull/3024#discussion_r3256562241

**Description**

The tools module requires sigs.k8s.io/karpenter which contains kwok/cloudprovider that imports kwok/utils. Without a local replace, go mod tidy resolves this against the published module version which doesn't contain kwok/utils, forcing the use of -e flag.

Adding 'replace sigs.k8s.io/karpenter => ./' makes the tools module resolve against the local source tree, fixing the issue.

Compiles locally


go.tools.mod
tallaxes, line 1: Let's also capture the results of `go mod tidy -modfile=go.tools.mod` after this change? For me it makes a couple of minor updates to `go.tools.mod` (including go version bump) + `go.tools.sum`

PR #3037, opened 2026-05-21
docs: InPlacePodVerticalScaling Support RFC

### LINKED ISSUE #829: feat: InPlacePodVerticalScaling support
### Description

Currently, Karpenter make scheduling and disruption decisions with the assumption that pod resource requests are immutable. Since Kubernetes 1.27, [InPlacePodVerticalScaling](https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/) was introduced as a new alpha feature, It's targeting beta post 1.30. With InPlacePodVerticalScaling, pod resource requests and limits become mutable.

A common use case of InPlacePodVerticalScaling is to mitigate startup issues for heavy application like Java, that big resource requests is initially allocated at startup, when the Pod becomes ready, a controller then lowers the resource requests to free up resources on the node.

With current Karpenter implementation, it is possible to create a loop that
- pods run in idle state with low resource requests
- Karpenter determines node pool under utilized, evict and consolidation is triggered
- new pods start up with big resource requests, they become unscheduled
- new nodes are bought up to schedule new pods

Karpenter need to be updated to recognize the mutable resource requests, to prevent such loop. Due to the flexibility of current InPlacePodVerticalScaling, this might be a difficult task if Karpenter itself doesn't understand the resource request mutation strategy.

Many users have shown interest in InPlacePodVerticalScaling feature. https://github.com/aws/containers-roadmap/issues/512 might provide some datapoint. As a cluster autoscaler, Karpenter's awareness of InPlacePodVerticalScaling is critical so together the node usage efficiency can be further improved while keeping applications stable and performant.

jigisha620:
Fixes #829

**Description**
RFC proposing how Karpenter should handle InPlacePodVerticalScaling to prevent incorrect disruption decisions when 
pods have been resized in place.

N/A

### DIFF
--- a/designs/in-place-pod-vertical-scaling.md
+++ b/designs/in-place-pod-vertical-scaling.md
@@ -6,10 +6,9 @@ Since Kubernetes 1.27 (alpha) and 1.33 (beta), [InPlacePodVerticalScaling](https
 
 Karpenter's disruption logic simulates pod placement using `pod.spec.containers[].resources.requests`. With InPlacePodVerticalScaling, two correctness issues arise:
 
-1. **During active resize**, `spec.requests` reflects the desired value while the kubelet's actual allocation (`status.allocatedResources`) may differ — either higher (resize-down in progress) or lower (resize-up in progress). Karpenter's view of node capacity is inaccurate until the resize completes.
-2. **Resize activity creates churn** that triggers unnecessary consolidation evaluation on nodes where pods are still stabilizing.
+1. **During active resize**, `spec.requests` reflects the desired value while the kubelet's actual allocation (`status.allocatedResources`) may differ. Karpenter's scheduling simulation diverges from the kube-scheduler's view, causing consolidation decisions the scheduler can't fulfill.
+2. **After eviction**, a mutating webhook (e.g., VPA) may set different resource requests on the recreated pod than what the evicted pod had. Karpenter has no way to predict this, so it may evict pods expecting them to land on a replacement node that can't actually fit the recreated pod's larger requests.
 
-For VPA's InPlace mode, VPA's admission webhook should set the replacement pod's requests to the current recommendation, which generally aligns with the pod's current spec. This makes `spec.requests` a reasonable approximation of replacement cost, and the problems above are primarily transitional.
 
 
 ## Background: How InPlacePodVerticalScaling Works
@@ -57,11 +56,11 @@ Karpenter's affected decision paths are **all disruption paths** - consolidation
 
 ### InPlace Mode
 
-During active resize, `spec.requests` and `status.allocatedResources` diverge. `spec.requests` reflects the desired value while `status.allocatedResources` reflects what the kubelet actually holds. This divergence can go in either direction — spec is lower during resize-down, higher during resize-up. Karpenter's view of node capacity is inaccurate until the resize completes. After resize completes, spec aligns with VPA's recommendation, so replacement cost is correctly estimated via `spec.requests`.
+During active resize, `spec.requests` and `status.allocatedResources` diverge. `spec.requests` reflects the desired value while `status.allocatedResources` reflects what the kubelet actually holds. This divergence can go in either direction, spec is lower during resize-down, higher during resize-up. Karpenter's view of node capacity is inaccurate until the resize completes. After resize completes, spec aligns with VPA's recommendation, so replacement cost is correctly estimated via `spec.requests`.
 
 ### CPU Startup Boost
 
-After the boost period completes and VPA resizes down, the node may appear underutilized. If Karpenter consolidates, the replacement pod will be boosted again at creation, potentially requiring more capacity than the steady-state spec suggests. Currently, `consolidateAfter` does not reset on resize activity, so Karpenter may evaluate the node for consolidation during the boost/unboost cycle.
+After the boost period completes and VPA resizes down, the node may appear underutilized. If Karpenter consolidates, the replacement pod will be boosted again at creation, requiring more capacity than the steady-state spec suggests. Without prediction, Karpenter bases its consolidation decision on the current (post-boost) spec and may move the pod to a node that can't fit the boosted value on recreation.
 
 ### Initial and Recreate Modes
 
@@ -70,11 +69,7 @@ In Initial and Recreate modes, VPA's recommendation evolves over time. When Karp
 ## Goals
 
 - Ensure accurate node capacity accounting during active pod resize across all disruption paths (consolidation, drift, expiration)
-- Prevent unnecessary consolidation evaluation on nodes with active resize activity
-
-## Non-Goals
-
-- Predicting replacement pod requests for VPA Initial or Recreate modes, where the recommendation may differ from the current pod spec
+- Predict post-eviction pod resource requests so that consolidation decisions account for mutating webhooks (e.g., VPA) that may change a pod's resources on recreation
 
 ## Proposal
 
@@ -84,16 +79,17 @@ When a pod is actively being resized down, its `spec.requests` reflects the new 
 
 **Why this matters:**
 
-During a resize-down transition, Karpenter's view of node capacity is inaccurate, it operates on incorrect data until the resize completes.
+During a resize-down transition, Karpenter's scheduling simulation diverges from the kube-scheduler's view. The kube-scheduler uses `UseStatusResources` to account for what pods actually hold, while Karpenter (without this change) only reads `spec.requests`. This divergence causes Karpenter to make consolidation decisions the scheduler can't fulfill, Karpenter thinks a target node has room, moves pods there, but the scheduler sees the node is fuller than Karpenter thought and can't place the incoming pods. The result is unnecessary pod disruption and temporarily pending pods.
 
 ```
-Node-A: 8 vCPU
-  - Pod-X: spec.requests = 2 (resize down in progress, allocated = 6)
+Node-B (target): 8 vCPU
+  - Pod-Z: spec.requests = 2 (resize down in progress, allocated = 6)
   Karpenter sees: used=2, free=6
-  Actual (until resize completes): used=6, free=2
+  Scheduler sees: used=6, free=2
+  Karpenter moves pods expecting 6 free → scheduler rejects → pods pending
 ```
 
-Enabling `UseStatusResources: true` ensures Karpenter always has an accurate view of what is physically committed on each node, rather than what is desired. This is good practice for correctness, consolidation decisions, scheduling simulations, and capacity metrics all benefit from reflecting reality rather than intent.
+Enabling `UseStatusResources: true` aligns Karpenter's resource accounting with the kube-scheduler, ensuring consolidation decisions are compatible with actual pod placement.
 
 **Fix:** Enable `UseStatusResources: true` in Karpenter's resource calculation (`resources.Ceiling()`).
 
@@ -108,25 +104,133 @@ Today Karpenter computes pod resource usage with `UseStatusResources=false`, so 
 | Resize UP infeasible | 6 | 2 | 6 | Conservative (treats as needing 6) |
 | Field not set (resize never occurred) | 4 | nil | 4 | Falls back to spec, backward compatible |
 
-### Part 2: Resize Events Count for consolidateAfter
+### Part 2: Predict Post-Eviction Resource Requests via Dry-Run
 
-Part 1 fixes resource accounting during active resize. Part 2 prevents Karpenter from continuously evaluating a node for disruption while resize activity is ongoing.
+Part 1 fixes resource accounting during active resize. Part 2 ensures consolidation decisions account for how pods will look after recreation.
 
-Today, the `podevents` controller updates `nodeClaim.Status.LastPodEventTime` when pods are bound, go terminal, or go terminating. This timestamp is compared against `consolidateAfter` to determine when a node becomes consolidatable. Pod resize events do not currently trigger this.
+When Karpenter consolidates a node, it evicts pods that are then recreated by their controllers. If a mutating webhook (e.g., VPA's admission controller) changes the pod's resource requests on creation, the recreated pod may require more resources than the original. Without prediction, Karpenter might move pods to a node that can't fit them post-recreation, causing unnecessary pending pods.
 
-By adding resize (change in `spec.requests`) as a pod event trigger, each resize resets the `consolidateAfter` timer. The node must be quiet (no resize activity) for the full `consolidateAfter` duration before becoming a consolidation candidate.
+**Mechanism:**
+
+A prediction cache runs dry-run pod creates (`POST /api/v1/namespaces/{ns}/pods?dryRun=All`) to determine what a pod's resources would be after recreation. This goes through the full API server admission chain (all mutating webhooks) without persisting anything.
+
+**Opt-in via annotation:**
+
+Pods (or their templates) must be annotated with `autoscaling.k8s.io/volatile-requests: "true"` to signal that their requests may change on recreation. This annotation is set by the vertical pod autoscaler (or any autoscaler that uses a mutating webhook to adjust resources at pod creation). Karpenter only runs dry-runs for annotated pods, avoiding unnecessary API calls for workloads whose requests are static.
+
+**Flow:**
+
+1. A background controller lists pods with the `volatile-requests` annotation, groups them by owner (one representative per ReplicaSet), strips runtime fields, and submits dry-run creates. Pods without an owner are skipped, they won't be recreated after eviction.
+2. The predicted `Ceiling()` result is cached, keyed by `(namespace, owner-uid)`.
+3. During consolidation simulation, annotated pods use their cached predicted resources instead of current `Ceiling(pod)`. Non-annotated pods use current resources as today.
+4. Before eviction, Karpenter re-runs a fresh dry-run for annotated pods in the consolidation plan. If the result differs from what the simulation assumed (e.g., VPA recommendation changed), the plan is aborted and re-evaluated on the next cycle.
+5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
+
+**Pros:**
+
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
+
+**Cons:**
+
+- Requires `create pods` RBAC cluster-wide. RBAC does not distinguish dry-run from real creates.
+- During active rollouts (Deployment or StatefulSet), the recreated pod may use a newer template than what was dry-run'd. This can cause one incorrect consolidation where the pod doesn't fit on the target node. The pod goes pending until Karpenter provisions a new node. The cache self-corrects on the next refresh once the recreated pod is running with the new template.
 
 ## Alternatives Considered
 
-1. **Block consolidation during active resize (instead of `UseStatusResources`):** Simply skip nodes with resizing pods. Simpler but less precise, doesn't fix destination capacity calculation and blocks all consolidation on the node even when safe.
+### 1. Read VPA Objects Directly
 
-2. **Track original requests at pod creation:** Record initial requests when Karpenter first observes a pod. If requests later decrease, use the original for simulation. Doesn't require owner lookups but loses state on Karpenter restart and can't distinguish webhook-modified pods.
+Read `VerticalPodAutoscaler.status.recommendation` and use the recommendation as the predicted post-eviction size.
 
-3. **Use template requests for displaced pod cost (`max(spec, template)`):** For displaced pods, look up the owner's pod template and use `max(pod.spec.requests, template.requests)` as the effective cost in scheduling simulation. This doesn't work well with VPA because VPA's webhook sets requests to the current recommendation at creation time, which may differ from the template. The template is not a reliable indicator of what the replacement pod will start with when VPA is involved.
+**Mechanism:**
 
-## Future Work
+Karpenter watches `VerticalPodAutoscaler` objects and reads `status.recommendation.containerRecommendations[].target`, the per-container recommended requests.
 
-- Upstream coordination with VPA to expose recommendations on pods (e.g., via an annotation) so that autoscalers like Karpenter can predict replacement pod requests
+**Discovery:**
+
+The VPA object itself signals which pods are managed:
+1. Read `spec.targetRef` (e.g., kind: Deployment, name: my-app)
+2. Only consider VPAs where `spec.updatePolicy.updateMode` != Off (webhook won't mutate if Off)
+3. Resolve: VPA → Deployment → ReplicaSets → Pods
+
+**Flow:**
+
+1. A background controller watches VPA objects. For each VPA with updateMode != Off, it reads `status.recommendation.containerRecommendations[]` and picks a representative pod from the targeted workload.
+2. For each container in the pod: if the container appears in the recommendation and is a controlled resource (per `spec.resourcePolicy.containerPolicies[].controlledResources`), use the `target` value. Otherwise, use that container's current requests from the pod spec.
+3. Compute `Ceiling()` from the merged result. Store in cache keyed by `(namespace, owner-uid)`.
+4. During consolidation simulation, VPA-targeted pods use their cached predicted resources. Non-targeted pods use current resources as today.
+5. Before eviction, re-read VPA recommendation fresh. If it differs from what the simulation assumed, abort and re-evaluate on the next cycle.
+6. Cache is refreshed via watch on VPA status updates (no TTL polling needed). Pod spec changes (deployment updates) also trigger recomputation.
+
+**RBAC:**
+
+Requires `get`, `list`, `watch` on `verticalpodautoscalers.autoscaling.k8s.io`.
+
+**Startup boost:**
+
+VPA's `status.recommendation.target` is the steady-state recommendation. Startup boost is configured separately in `spec.startupBoost` and computed dynamically by the webhook at creation time (`target × factor`). The boosted value is not stored in VPA status. To predict it, Karpenter would need to read the boost configuration and replicate the computation.
+
+**Pros:**
+
+- No `create pods` RBAC
+- No annotation needed (VPA object is the signal)
+- Immediate reaction to recommendation changes via watch
+
+**Cons:**
+
+- Couples Karpenter to VPA's CRD schema
+- Only works for VPA, other resource-mutating webhooks are not covered
+- Replicates webhook logic: `controlledResources`, `updateMode`, startup boost computation, container name matching and merging
+
+### 2. Annotation Carrying the Prediction
+
+The vertical pod autoscaler annotates workloads with the total predicted pod-level resource requests, the Ceiling equivalent of what its webhook would produce on a new pod. Karpenter reads this annotation during consolidation simulation.
+
+**Annotation:**
+
+```
+autoscaling.k8s.io/predicted-pod-requests: '{"cpu": "8500m", "memory": "2304Mi"}'
+```
+
+The autoscaler computes this by taking its recommendation for managed containers (with startup boost applied if configured), current spec for unmanaged containers, and computing the total using standard Ceiling logic.
+
+**Where the annotation lives:**
+
+Two variants:
+
+1. **On the owner (Deployment/StatefulSet)** (recommended): VPA patches `metadata.annotations` on the Deployment. One patch per workload per recommendation change. Karpenter resolves pod → RS → Deployment → read annotation.
+2. **On live pods**: VPA patches the annotation on every managed pod whenever the recommendation changes. Simpler resolution (annotation is on the pod directly) but scales poorly, N patches per recommendation change.
+
+**Flow:**
+
+1. VPA computes the predicted post-eviction resources and patches the annotation on the owner (or pods).
+2. Karpenter reads the annotation during consolidation simulation. If present, uses it as the predicted resources for pods being moved. If absent, uses current `Ceiling(pod)`.
+3. Before eviction, Karpenter re-reads the annotation. If it changed since simulation, abort and re-evaluate.
+
+**RBAC:**
+
+No additional RBAC for Karpenter (already reads pods and watches Deployments). VPA needs patch on Deployments/StatefulSets (if annotating the owner).
+
+**Pros:**
+
+- No additional RBAC for Karpenter
+- Autoscaler-agnostic (any controller can adopt the convention)
+- Simple read path for Karpenter (just read annotation)
+- No coupling to specific CRDs
+
+**Cons:**
+
+- Schema must be agreed upon across projects
+- If annotating live pods (variant 2): API load at scale (N patches per recommendation change)
+
+
+### 3. Call Webhook Endpoints Directly
+
+Discover `MutatingWebhookConfiguration` resources, construct `AdmissionReview` requests, and call webhook services. Gets exact results without pod create RBAC but is extremely complex (auth, ordering, chaining, network access) and fragile. Essentially re-implements the API server's admission chain.
+
+### 4. Delay or Block Consolidation During Resize
+
+Delay consolidation after a resize event (e.g., reset `consolidateAfter`) or skip nodes with active resizes entirely. Simpler but doesn't solve the core problem (incorrect prediction of post-eviction size), doesn't fix destination capacity calculation, and blocks consolidation on a node even when safe. With accurate predictions, consolidation can safely proceed without artificial delays.
 
 ## References
 
--- a/designs/in-place-pod-vertical-scaling.md
+++ b/designs/in-place-pod-vertical-scaling.md
@@ -64,7 +64,7 @@ After the boost period completes and VPA resizes down, the node may appear under
 
 ### Initial and Recreate Modes
 
-In Initial and Recreate modes, VPA's recommendation evolves over time. When Karpenter evicts a pod, the replacement receives VPA's current recommendation, which may differ from the evicted pod's spec. Karpenter has no way to predict this value today.
+In Initial and Recreate modes, VPA's admission webhook sets pod requests to the current recommendation at creation time. Since recommendations evolve continuously, when Karpenter evicts a pod the replacement may receive different resource requests than the evicted pod had. Karpenter has no way to predict this value today.
 
 ## Goals
 
@@ -104,43 +104,11 @@ Today Karpenter computes pod resource usage with `UseStatusResources=false`, so 
 | Resize UP infeasible | 6 | 2 | 6 | Conservative (treats as needing 6) |
 | Field not set (resize never occurred) | 4 | nil | 4 | Falls back to spec, backward compatible |
 
-### Part 2: Predict Post-Eviction Resource Requests via Dry-Run
+### Part 2: Predict Post-Eviction Resource Requests via VPA Recommendation
 
 Part 1 fixes resource accounting during active resize. Part 2 ensures consolidation decisions account for how pods will look after recreation.
 
-When Karpenter consolidates a node, it evicts pods that are then recreated by their controllers. If a mutating webhook (e.g., VPA's admission controller) changes the pod's resource requests on creation, the recreated pod may require more resources than the original. Without prediction, Karpenter might move pods to a node that can't fit them post-recreation, causing unnecessary pending pods.
-
-**Mechanism:**
-
-A prediction cache runs dry-run pod creates (`POST /api/v1/namespaces/{ns}/pods?dryRun=All`) to determine what a pod's resources would be after recreation. This goes through the full API server admission chain (all mutating webhooks) without persisting anything.
-
-**Opt-in via annotation:**
-
-Pods (or their templates) must be annotated with `autoscaling.k8s.io/volatile-requests: "true"` to signal that their requests may change on recreation. This annotation is set by the vertical pod autoscaler (or any autoscaler that uses a mutating webhook to adjust resources at pod creation). Karpenter only runs dry-runs for annotated pods, avoiding unnecessary API calls for workloads whose requests are static.
-
-**Flow:**
-
-1. A background controller lists pods with the `volatile-requests` annotation, groups them by owner (one representative per ReplicaSet), strips runtime fields, and submits dry-run creates. Pods without an owner are skipped, they won't be recreated after eviction.
-2. The predicted `Ceiling()` result is cached, keyed by `(namespace, owner-uid)`.
-3. During consolidation simulation, annotated pods use their cached predicted resources instead of current `Ceiling(pod)`. Non-annotated pods use current resources as today.
-4. Before eviction, Karpenter re-runs a fresh dry-run for annotated pods in the consolidation plan. If the result differs from what the simulation assumed (e.g., VPA recommendation changed), the plan is aborted and re-evaluated on the next cycle.
-5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
-
-**Pros:**
-
-- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
-- No coupling to specific CRDs or autoscaler implementations
-
-**Cons:**
-
-- Requires `create pods` RBAC cluster-wide. RBAC does not distinguish dry-run from real creates.
-- During active rollouts (Deployment or StatefulSet), the recreated pod may use a newer template than what was dry-run'd. This can cause one incorrect consolidation where the pod doesn't fit on the target node. The pod goes pending until Karpenter provisions a new node. The cache self-corrects on the next refresh once the recreated pod is running with the new template.
-
-## Alternatives Considered
-
-### 1. Read VPA Objects Directly
-
-Read `VerticalPodAutoscaler.status.recommendation` and use the recommendation as the predicted post-eviction size.
+When Karpenter consolidates a node, it evicts pods that are then recreated by their controllers. If a mutating webhook (e.g., VPA's admission controller) changes the pod's resource requests on creation, the recreated pod may require more or less resources than the original. Without prediction, Karpenter might move pods to a node that can't fit them post-recreation, causing unnecessary pending pods.
 
 **Mechanism:**
 
@@ -181,6 +149,36 @@ VPA's `status.recommendation.target` is the steady-state recommendation. Startup
 - Couples Karpenter to VPA's CRD schema
 - Only works for VPA, other resource-mutating webhooks are not covered
 - Replicates webhook logic: `controlledResources`, `updateMode`, startup boost computation, container name matching and merging
+
+
+## Alternatives Considered
+
+### 1. Dry-Run Pod Creates
+
+A prediction cache runs dry-run pod creates dry-run pod creates (`POST /api/v1/namespaces/{ns}/pods?dryRun=All`) to determine what a pod's resources would be after recreation. This goes through the full API server admission chain (all mutating webhooks) without persisting anything.
+
+**Opt-in via annotation:**
+
+Pods (or their templates) must be annotated with `autoscaling.k8s.io/volatile-requests: "true"` to signal that their requests may change on recreation. This annotation is set by the vertical pod autoscaler (or any autoscaler that uses a mutating webhook to adjust resources at pod creation). Karpenter only runs dry-runs for annotated pods, avoiding unnecessary API calls for workloads whose requests are static.
+
+**Flow:**
+
+1. A background controller lists pods with the `volatile-requests` annotation, groups them by owner (one representative per ReplicaSet), strips runtime fields, and submits dry-run creates. Pods without an owner are skipped, they won't be recreated after eviction.
+2. The predicted `Ceiling()` result is cached, keyed by `(namespace, owner-uid)`.
+3. During consolidation simulation, annotated pods use their cached predicted resources instead of current `Ceiling(pod)`. Non-annotated pods use current resources as today.
+4. During consolidation validation, Karpenter re-runs a fresh dry-run for annotated pods in the plan. If the result differs from what the simulation assumed, the decision is rejected.
+5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
+
+**Pros:**
+
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
+
+**Cons:**
+
+- Requires `create pods` RBAC cluster-wide. RBAC does not distinguish dry-run from real creates.
+- During active rollouts (Deployment or StatefulSet), the recreated pod may use a newer template than what was dry-run'd. This can cause one incorrect consolidation where the pod doesn't fit on the target node. The pod goes pending until Karpenter provisions a new node. The cache self-corrects on the next refresh once the recreated pod is running with the new template.
+
 
 ### 2. Annotation Carrying the Prediction
 
--- a/designs/in-place-pod-vertical-scaling.md
+++ b/designs/in-place-pod-vertical-scaling.md
@@ -230,6 +230,42 @@ Discover `MutatingWebhookConfiguration` resources, construct `AdmissionReview` r
 
 Delay consolidation after a resize event (e.g., reset `consolidateAfter`) or skip nodes with active resizes entirely. Simpler but doesn't solve the core problem (incorrect prediction of post-eviction size), doesn't fix destination capacity calculation, and blocks consolidation on a node even when safe. With accurate predictions, consolidation can safely proceed without artificial delays.
 
+## Graduation Criteria
+
+The VPA prediction mechanism (Part 2) starts with reading VPA CRD objects directly. This is the correct starting point regardless of which long-term mechanism materializes, because it requires no upstream changes and works with all existing VPA installations. Karpenter only produces predictions when VPA objects exist in the cluster; if VPA is not installed or no VPAs are configured, there is no effect on Karpenter's behavior.
+
+The long-term graduation path depends on which upstream mechanism is adopted:
+
+### Path A: VPA-Independent Mechanism
+
+If a VPA-independent prediction mechanism becomes available (e.g., dry-run pod creates with dedicated RBAC), Karpenter can switch without requiring any changes to VPA. The existing admission webhook already handles dry-run requests, so VPA doesn't need to ship a new API.
+
+| | Mechanism | When | Customer Action |
+|--|-----------|------|-----------------|
+| Initial | CRD reading | Until the new mechanism is available in K8s | None |
+| Dual support | Both (runtime K8s version check) | While Karpenter supports K8s versions with and without the new mechanism | None |
+| CRD reading dropped | New mechanism only | Once Karpenter's minimum supported K8s version includes the new mechanism | Upgrade cluster to meet Karpenter's minimum K8s version |
+
+No VPA coordination required. The transition is driven entirely by K8s version support.
+
+### Path B: VPA-Dependent Mechanism
+
+If the prediction mechanism requires VPA to ship a new API, the transition requires coordination with the customer's VPA version.
+
+| | Mechanism | When | Customer Action |
+|--|-----------|------|-----------------|
+| Initial | CRD reading | Until VPA ships a new prediction API | None |
+| Dual support | Both (prefers new API, falls back to CRD reading) | Karpenter versions that support both old and new VPA | None (auto-detected) |
+| CRD reading dropped | New API only | Karpenter version that removes CRD reading | Upgrade VPA before upgrading Karpenter |
+
+Requires VPA to ship the new API, which is outside our control. CRD reading is dropped at a documented Karpenter version boundary: customers must upgrade VPA before upgrading Karpenter past that version.
+
+
+### Notes
+
+- CRD reading has near-zero maintenance cost (GA API, stable fields, read-only). We retain the option to keep it indefinitely if neither path materializes.
+
+
 ## References
 
 - [InPlacePodVerticalScaling KEP-1287](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources)
--- a/designs/in-place-pod-vertical-scaling.md
+++ b/designs/in-place-pod-vertical-scaling.md
@@ -1,0 +1,135 @@
+# InPlacePodVerticalScaling Support
+
+## Motivation
+
+Since Kubernetes 1.27 (alpha) and 1.33 (beta), [InPlacePodVerticalScaling](https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/) allows controllers to mutate a running pod's resource requests without restarting it. With Kubernetes 1.33 enabling this feature by default, controllers like VPA can now right-size pods without eviction. Karpenter must correctly account for resized pods in its disruption decisions to avoid incorrect consolidation during and after resize operations.
+
+Karpenter's disruption logic simulates pod placement using `pod.spec.containers[].resources.requests`. With InPlacePodVerticalScaling, two correctness issues arise:
+
+1. **During active resize**, `spec.requests` reflects the desired value while the kubelet's actual allocation (`status.allocatedResources`) may differ — either higher (resize-down in progress) or lower (resize-up in progress). Karpenter's view of node capacity is inaccurate until the resize completes.
+2. **Resize activity creates churn** that triggers unnecessary consolidation evaluation on nodes where pods are still stabilizing.
+
+For VPA's InPlace mode, VPA's admission webhook should set the replacement pod's requests to the current recommendation, which generally aligns with the pod's current spec. This makes `spec.requests` a reasonable approximation of replacement cost, and the problems above are primarily transitional.
+
+
+## Background: How InPlacePodVerticalScaling Works
+
+[KEP-1287](https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1287-in-place-update-pod-resources) introduced InPlacePodVerticalScaling, allowing pod resource requests to be mutated without restart. The feature graduated to alpha in Kubernetes 1.27 and beta in 1.33.
+
+### The Resize Subresource
+
+Before InPlacePodVerticalScaling, `pod.spec.containers[].resources` was immutable after creation. With the feature enabled, a controller (e.g., VPA) can patch the pod's resources via the resize subresource:
+
+```bash
+kubectl patch pod my-pod --subresource=resize --patch '{
+  "spec": {"containers": [{"name": "app", "resources": {"requests": {"cpu": "2"}}}]}
+}'
+```
+
+The kubelet applies the new resource values and updates `pod.status.containerStatuses[].resources` with the actual allocation. Resize status is communicated via Pod conditions:
+
+- `PodResizePending`: The kubelet cannot immediately grant the request.
+  - `reason: Infeasible` - the resize is impossible on the current node (e.g., requesting more resources than the node has).
+  - `reason: Deferred` - the resize is not currently possible but may become feasible later (e.g., when another pod is removed). The kubelet retries deferred resizes periodically, prioritized by PriorityClass, then QoS class, then wait time.
+- `PodResizeInProgress`: The kubelet has accepted the resize and allocated resources, but changes are still being applied. Errors during actuation are reported in the condition's message with `reason: Error`.
+
+### VPA Update Modes
+
+VPA supports multiple update modes that determine how recommendations are applied to pods:
+
+| Mode | Behavior |
+|------|----------|
+| **Off** | VPA computes recommendations but takes no action |
+| **Initial** | Admission webhook sets requests at pod creation only; no further updates |
+| **Recreate** | VPA evicts pods whose requests diverge from recommendation; webhook sets new value on replacement |
+| **InPlace** | VPA patches running pod's requests via resize subresource; webhook sets recommendation on replacement |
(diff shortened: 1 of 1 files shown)


designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,267 @@
...
+
+- `PodResizePending`: The kubelet cannot immediately grant the request.
+  - `reason: Infeasible` - the resize is impossible on the current node (e.g., requesting more resources than the node has).
+  - `reason: Deferred` - the resize is not currently possible but may become feasible later (e.g., when another pod is removed). The kubelet retries deferred resizes periodically, prioritized by PriorityClass, then QoS class, then wait time.
+- `PodResizeInProgress`: The kubelet has accepted the resize and allocated resources, but changes are still being applied. Errors during actuation are reported in the condition's message with `reason: Error`.
+
+## How This Affects Karpenter Today
+
+Karpenter's affected decision paths are **all disruption paths** - consolidation, drift, and expiration. All three use the same scheduling simulation that computes pod placement based on current `pod.spec.containers[].resources.requests`. Provisioning is unaffected because pending pods have their template requests (correct at creation time).
```
ryan-mist: Do we have to consider this in provisioning when pods are being resized up? When pod resources are scaled up then we might end up provisioning a node without enough capacity

With the current proposal, we could handle this as part of disruption as long as we do treat `reason: Infeasible` as a signal to disrupt the pod.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,267 @@
...
+| Scenario | spec | allocated | max(spec, allocated) | Effect |
+|----------|:---:|:---:|:---:|------|
+| Normal (no resize) | 4 | 4 | 4 | No change |
+| Resize DOWN in progress | 2 | 6 | **6** | Correctly reflects committed resources |
+| Resize DOWN completed | 1 | 1 | 1 | No change (Phase 2 problem remains) |
+| Resize UP in progress | 6 | 2 | 6 | Same as reading spec |
+| Resize UP infeasible | 6 | 2 | 6 | Conservative (treats as needing 6) |
+| Field not set (resize never occurred) | 4 | nil | 4 | Falls back to spec, backward compatible |
+
+### Part 2: Use Template Requests for Post-Resize Consolidation Simulation
```
DerekFrank: Theres an alternative we should consider: annotating all pods with their initial resource requests. That way we support all workload managers out of the box, at the cost of more writes

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Goals
+
+- Prevent consolidation thrash loops caused by pods that have been resized down from their template requests
+- Ensure replacement node sizing accounts for the "restart cost" of resized pods across all disruption paths (consolidation, drift, expiration)
+- Gate the feature behind a feature flag for incremental rollout
+
+## Proposal
+
+### Part 1: Use Allocated Resources During Active Resize
```
DerekFrank, line 76: Should these scaling events count for consolidateAfter?

jigisha620: Good callout, I have mentioned this in Alternatives Considered as "Consolidation cooldown after resize", rejected it as a standalone solution because it doesn't fix the incorrect placement cost, just delays it. But combined with the proposed solution in the RFC, having resize events count for `consolidateAfter` makes sense as a stability measure to prevent continuously evaluating nodes while resize activity is ongoing.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+During a resize-down transition, Karpenter's view of node capacity is inaccurate, it operates on incorrect data until the resize completes.
+
+```
+Node-A: 8 vCPU
+  - Pod-X: spec.requests = 2 (resize down in progress, allocated = 6)
+  Karpenter sees: used=2, free=6
+  Actual (until resize completes): used=6, free=2
+```
+
+Enabling `UseStatusResources: true` ensures Karpenter always has an accurate view of what is physically committed on each node, rather than what is desired. This is good practice for correctness, consolidation decisions, scheduling simulations, and capacity metrics all benefit from reflecting reality rather than intent.
```
jmdeal: Has kube-scheduler also been updated to use this (or has it always)?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+| Resize DOWN completed | 1 | 1 | 1 | No change (Phase 2 problem remains) |
+| Resize UP in progress | 6 | 2 | 6 | Same as reading spec |
+| Resize UP infeasible | 6 | 2 | 6 | Conservative (treats as needing 6) |
+| Field not set (resize never occurred) | 4 | nil | 4 | Falls back to spec, backward compatible |
+
+### Part 2: Use Template Requests for Post-Resize Consolidation Simulation
+
+After a resize completes, `spec` and `allocatedResources` both reflect the new value, Part 1 no longer helps. But if the pod is evicted, its replacement starts at the Deployment template's requests. Karpenter must account for this "restart cost" in all disruption simulations.
+
+**Approach:** For displaced pods (being evicted from candidate nodes), use `max(pod.spec.requests, ownerTemplate.requests)` as the effective resource cost in the scheduling simulation. This ensures the simulation accounts for the restart cost, what the replacement pod will actually need.
```
jmdeal: I'm wondering if this should be the max, or if it should just be the size from the owner ref. The problem with using max is that it's not going to be what's respected by the scheduler when the pod is recreated. We also have no way to measure intent - how do we know the pod is going to be scaled back up again? I think this could lead to its own consolidation loops where we overprovision, the scheduler makes different placement decisions where it doesn't allow for scale-down or the pod isn't scaled up, and we consolidate again.

designs/in-place-pod-vertical-scaling.md
nathangeology: Finding a new home for a pod that needs to get bigger seems like a scheduler job to me. It might be the autoscaler needs to help there, but the scheduler should probably be the one to initiate the process.

designs/in-place-pod-vertical-scaling.md
jigisha620: I gave it some thought initially, but dropped it considering things like who is responsible for writing the annotation? If we want to make a change in upstream replicaSet or deployment controller (a great option if we get alignment), I think that would be a long-term approach. Worth considering and maybe we can think of this as an optimization.

If we were to do it in Karpenter then we need to think about the mechanism, is it a webhook or a controller? How does it react with let's say VPA, especially VPA's admission webhook? I can think of webhook ordering issues if we introduce a Karpenter webhook for annotating. If we use a controller, there could a state where VPA's webhook updates the pod resources first before the controller even looks at the pod's initial resources.

designs/in-place-pod-vertical-scaling.md
jmdeal: It would be good to understand what the common use-cases for growing pods are as well. We have a pretty well understood use-case for shrinking pods, and shrinking should always work regardless of pod placement. However growing only works if your node has headroom, something Karpenter is designed to reduce via consolidation

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+
+Karpenter's disruption logic makes decisions by simulating pod placement using each pod's current `pod.spec.containers[].resources.requests`. This breaks a core assumption in Karpenter's model: that a pod's current requests represent its cost if evicted and rescheduled. When a pod is resized down (e.g., after JVM startup completes), Karpenter sees lower requests and may consolidate the node. But eviction creates a new pod from the workload's pod template with the original (higher) requests, triggering a new node provision and restarting the expensive startup cycle. This creates an infinite loop of consolidation and re-provisioning.
+
+
+## Background: How InPlacePodVerticalScaling Works
+
+### The Resize Subresource
+
+Before InPlacePodVerticalScaling, `pod.spec.containers[].resources` was immutable after creation. With the feature enabled, a controller (e.g., VPA) can patch the pod's resources via the resize subresource:
```
nathangeology, line 20: Question, In a replicaset would we be resizing in-place all the replicas at once or could the replicas have different sizes with this?

nathangeology: Related question: If replicas can have different sizes, how do we reason about disruption in this case? Do we just assume all replicas come back at their original size? Also, it just seems weird to me that replicas could... well... not be replicas of each other? hahaha

designs/in-place-pod-vertical-scaling.md
DerekFrank: I think we still need an answer one way or the other, once we say Karpenter supports InPlacePodVerticalScaling, we need to support it. Maybe it can be in graduation criteria, but it is an open design question

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+
+Parts 1 and 2 fix the correctness of disruption decisions. Part 3 prevents Karpenter from continuously evaluating a node for disruption while resize activity is ongoing.
+
+Today, the `podevents` controller updates `nodeClaim.Status.LastPodEventTime` when pods are bound, go terminal, or go terminating. This timestamp is compared against `consolidateAfter` to determine when a node becomes consolidatable. Pod resize events do not currently trigger this.
+
+By adding resize (change in `spec.requests`) as a pod event trigger, each resize resets the `consolidateAfter` timer. The node must be quiet (no resize activity) for the full `consolidateAfter` duration before becoming a consolidation candidate. See Open Question 3 for whether `consolidateAfter` is the right timer to reuse here.
+
+## Feature Gating
+
+All three parts should be gated behind a feature flag. This is a behavioral change to disruption logic that directly affects replacement node sizing. A feature gate allows safe incremental rollout and gives users a way to revert if unexpected issues arise.
```
jmdeal: I'm not sure if this is something I would feature gate. Karpenter's current behavior is wrong in a post-inplace vertical pod scaling world and I think we can make an objective improvement. To me this isn't a new feature, this is Karpenter gaining conformance with the current state of upstream. Are there any pain-points that we think customers would want to opt out due to?

designs/in-place-pod-vertical-scaling.md
DerekFrank: Can we explore the option with sig-apps and sig-autoscaling in parallel with the design? I'd like to define the interface once for everyone and then just have to support the interface instead of building informer caches for every workload manager we want to support

designs/in-place-pod-vertical-scaling.md
jigisha620: Yes, I have added it to the doc as well.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+
+1. **Block consolidation during active resize (instead of `UseStatusResources`):** Simply skip nodes with resizing pods. Simpler but less precise, doesn't fix destination capacity calculation and blocks all consolidation on the node even when safe.
+
+2. **Track original requests at pod creation:** Record initial requests when Karpenter first observes a pod. If requests later decrease, use the original for simulation. Doesn't require owner lookups but loses state on Karpenter restart and can't distinguish webhook-modified pods.
+
+3. **Consolidation cooldown after resize:** Extend the `Consolidatable` timer on nodes with resized pods. Prevents immediate thrash but doesn't solve the problem, just delays it. Arbitrary timeout, wasteful.
+
+## Open Questions
+
+1. **Should Karpenter proactively disrupt nodes with Infeasible resizes?** If a pod's resize is marked `PodResizePending` with `reason: Infeasible`, the pod can never get the resources it needs on the current node. Karpenter could treat this as a disruption signal and move the pod to a node with sufficient capacity.
```
jmdeal: I don't think so, but I'd be curious to see what user stories around this exist. My opinion is that it would be better for whatever is orchestrating the scale-up to evict the pod rather than disrupt all other pods on the node.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+
+2. **Which workload types should be supported for template lookup?** ReplicaSets and StatefulSets are the primary targets. Should we also add a Job informer for Job pods, or are short-lived Job pods unlikely to be resized in practice?
+
+3. **Should Part 3 reuse `consolidateAfter` or introduce a separate timer?** Reusing `consolidateAfter` is simple (no new configuration) and treats resize as just another pod event. But users might want different cooldowns for resize activity vs pod scheduling events.
```
Ignoramuss: Also worth considering how this plays out for bursty workloads with in-place VPA, can we introduce behaviors that make nodes practically not eligible for consolidation?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+T=6m    Startup completes, VPA resizes: {cpu: 8} → {cpu: 2}
+        Node utilization: 25% again
+
+T=7m    Consolidation triggers again → INFINITE LOOP
+```
+
+### Problem 2: Destination Node Free Capacity Can Be Overestimated
+
+When evaluating whether a destination node has room, Karpenter calculates existing pod usage from current `spec.requests`. If pods on that node have a resize-down in progress or a resize that has failed, `spec.requests` reflects the new lower value while the kubelet still holds the old (higher) allocation. Karpenter may incorrectly conclude that disruption is safe based on capacity that isn't actually available. This only applies to resize-down. For resize-up, current spec is higher than what the kubelet has allocated, so Karpenter underestimates free capacity, which is the conservative/safe direction.
+
```
nathangeology, line 68: Even apart from these two problems you can have the problem of groups of pods vertically scaling in place and triggering multiple waves of consolidation where the new pods back to original size pods have to squeeze in around the existing, running pods that have already vertically scaled (or if you don't follow the pod size changes in the Karpenter data model, you can end up with free capacity that Karpenter can't reclaim easily). So you sign up for a lot of disruption waves and wasted cycles to get to the end state.

designs/in-place-pod-vertical-scaling.md
DerekFrank: This also prevents this from being a breaking change

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+### Part 2: Use Template Requests for Post-Resize Consolidation Simulation
+
+After a resize completes, `spec` and `allocatedResources` both reflect the new value, Part 1 no longer helps. But if the pod is evicted, its replacement starts at the Deployment template's requests. Karpenter must account for this "restart cost" in all disruption simulations.
+
+**Approach:** For displaced pods (being evicted from candidate nodes), use `max(pod.spec.requests, ownerTemplate.requests)` as the effective resource cost in the scheduling simulation. This ensures the simulation accounts for the restart cost, what the replacement pod will actually need.
+
+#### Implementation: Informer-Based Caching
+
+We add informer watches for the owner types. Karpenter already uses this pattern for Pods, Nodes, NodeClaims, NodePools, and DaemonSets.
```
Ignoramuss: Any perf considerations with adding new Informers?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
+# InPlacePodVerticalScaling Support
+
+## Motivation
+
+Since Kubernetes 1.27 (alpha) and 1.33 (beta), [InPlacePodVerticalScaling](https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/) allows controllers to mutate a running pod's resource requests without restarting it. With Kubernetes 1.33 enabling this feature by default, pods can now be resized in place by VPA or custom controllers.
+
+Karpenter's disruption logic makes decisions by simulating pod placement using each pod's current `pod.spec.containers[].resources.requests`. This breaks a core assumption in Karpenter's model: that a pod's current requests represent its cost if evicted and rescheduled. When a pod is resized down (e.g., after JVM startup completes), Karpenter sees lower requests and may consolidate the node. But eviction creates a new pod from the workload's pod template with the original (higher) requests, triggering a new node provision and restarting the expensive startup cycle. This creates an infinite loop of consolidation and re-provisioning.
+
+
+## Background: How InPlacePodVerticalScaling Works
```
DerekFrank, line 14: Should link to the KEP here

designs/in-place-pod-vertical-scaling.md
DerekFrank: Should it be `max(pod.spec.requests, ownerTemplate.requests)` or `max(pod.status.containerStatuses[].resources, ownerTemplate.requests)`?

designs/in-place-pod-vertical-scaling.md
jigisha620: Yes, kube-scheduler uses max(spec, allocatedResources) whenever InPlacePodVerticalScaling is enabled.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+
+### Part 3: Resize Events Count for consolidateAfter
+
+Parts 1 and 2 fix the correctness of disruption decisions. Part 3 prevents Karpenter from continuously evaluating a node for disruption while resize activity is ongoing.
+
+Today, the `podevents` controller updates `nodeClaim.Status.LastPodEventTime` when pods are bound, go terminal, or go terminating. This timestamp is compared against `consolidateAfter` to determine when a node becomes consolidatable. Pod resize events do not currently trigger this.
+
+By adding resize (change in `spec.requests`) as a pod event trigger, each resize resets the `consolidateAfter` timer. The node must be quiet (no resize activity) for the full `consolidateAfter` duration before becoming a consolidation candidate. See Open Question 3 for whether `consolidateAfter` is the right timer to reuse here.
+
+## Feature Gating
```
DerekFrank: Since we're scoping some questions out of this design doc (consolidateAfter), can we get graduation criteria here?

designs/in-place-pod-vertical-scaling.md
nathangeology: Personally I think the scheduler should maybe be in charge of this one. It can see that the pod doesn't fit and try to reschedule a new copy someplace at the new size and mark it pending if it can't find a home.

DerekFrank: Does anything do this today? Will VPA do this?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+→ Blocked. Neither pod can restart on Node-C alongside Pod-Z. ✓
+```
+
+### Part 3: Resize Events Count for consolidateAfter
+
+Parts 1 and 2 fix the correctness of disruption decisions. Part 3 prevents Karpenter from continuously evaluating a node for disruption while resize activity is ongoing.
+
+Today, the `podevents` controller updates `nodeClaim.Status.LastPodEventTime` when pods are bound, go terminal, or go terminating. This timestamp is compared against `consolidateAfter` to determine when a node becomes consolidatable. Pod resize events do not currently trigger this.
+
+By adding resize (change in `spec.requests`) as a pod event trigger, each resize resets the `consolidateAfter` timer. The node must be quiet (no resize activity) for the full `consolidateAfter` duration before becoming a consolidation candidate. See Open Question 3 for whether `consolidateAfter` is the right timer to reuse here.
```
DerekFrank: Is this an open question or are you saying it should affect consolidateAfter?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+
+2. **Track original requests at pod creation:** Record initial requests when Karpenter first observes a pod. If requests later decrease, use the original for simulation. Doesn't require owner lookups but loses state on Karpenter restart and can't distinguish webhook-modified pods.
+
+3. **Consolidation cooldown after resize:** Extend the `Consolidatable` timer on nodes with resized pods. Prevents immediate thrash but doesn't solve the problem, just delays it. Arbitrary timeout, wasteful.
+
+## Open Questions
+
+1. **Should Karpenter proactively disrupt nodes with Infeasible resizes?** If a pod's resize is marked `PodResizePending` with `reason: Infeasible`, the pod can never get the resources it needs on the current node. Karpenter could treat this as a disruption signal and move the pod to a node with sufficient capacity.
+
+2. **Which workload types should be supported for template lookup?** ReplicaSets and StatefulSets are the primary targets. Should we also add a Job informer for Job pods, or are short-lived Job pods unlikely to be resized in practice?
```
nathangeology: I could imagine certain types of batch jobs needing to resize (e.g. processing data and starts running low on memory and requests a resize or something similar).

nathangeology (commented):
Thanks for taking this on!

designs/in-place-pod-vertical-scaling.md
jigisha620: I am saying that we should consider `consolidateAfter`, but should we reuse the same or consider a totally different timeout before we consider a node eligible for consolidation?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
+# InPlacePodVerticalScaling Support
+
+## Motivation
+
+Since Kubernetes 1.27 (alpha) and 1.33 (beta), [InPlacePodVerticalScaling](https://kubernetes.io/blog/2023/05/12/in-place-pod-resize-alpha/) allows controllers to mutate a running pod's resource requests without restarting it. With Kubernetes 1.33 enabling this feature by default, pods can now be resized in place by VPA or custom controllers.
+
+Karpenter's disruption logic makes decisions by simulating pod placement using each pod's current `pod.spec.containers[].resources.requests`. This breaks a core assumption in Karpenter's model: that a pod's current requests represent its cost if evicted and rescheduled. When a pod is resized down (e.g., after JVM startup completes), Karpenter sees lower requests and may consolidate the node. But eviction creates a new pod from the workload's pod template with the original (higher) requests, triggering a new node provision and restarting the expensive startup cycle. This creates an infinite loop of consolidation and re-provisioning.
```
ellistarn: >  But eviction creates a new pod from the workload's pod template with the original (higher) requests, triggering a new node provision and restarting the expensive startup cycle. This creates an infinite loop of consolidation and re-provisioning.

Have we raised this with the VPA folks?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,277 @@
...
+  NodePool: consolidationPolicy: WhenUnderutilized
+
+T=0m    Pod running, resized to {cpu: 2}
+        Node utilization: 2/8 vCPU = 25%
+
+T=1m    Consolidation simulates: "Pod (2 vCPU) fits on Node-B" → evicts
+        New pod created from template: requests {cpu: 8}
+        Needs a new node → Karpenter provisions one
+
+T=6m    Startup completes, VPA resizes: {cpu: 8} → {cpu: 2}
```
ellistarn: Have we thought through: https://kubernetes.io/docs/concepts/workloads/autoscaling/vertical-pod-autoscale/#update-modes

The webhook should resize this pod before it lands.

jigisha620: Good catch, that's right. In case of VPA, the webhook should take care of resizing the pod before it lands. Updated the RFC to make it generic for any controller that performs resizing. For VPA though, we need a way to tell for different update modes, what would the pod look like when it comes back.

designs/in-place-pod-vertical-scaling.md
jigisha620: Have started a discussion with them.

designs/in-place-pod-vertical-scaling.md
jigisha620: Kubelet can mark the request as Infeasible.

designs/in-place-pod-vertical-scaling.md
jigisha620: Discussed offline

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,135 @@
...
+
+## How This Affects Karpenter Today
+
+Karpenter's affected decision paths are **all disruption paths** - consolidation, drift, and expiration. All three use the same scheduling simulation that computes pod placement based on current `pod.spec.containers[].resources.requests`. Provisioning is unaffected because pending pods have their template requests (correct at creation time).
+
+### InPlace Mode
+
+During active resize, `spec.requests` and `status.allocatedResources` diverge. `spec.requests` reflects the desired value while `status.allocatedResources` reflects what the kubelet actually holds. This divergence can go in either direction — spec is lower during resize-down, higher during resize-up. Karpenter's view of node capacity is inaccurate until the resize completes. After resize completes, spec aligns with VPA's recommendation, so replacement cost is correctly estimated via `spec.requests`.
```
omerap12: I think that, as a best-effort solution, Karpenter can assume that for `InPlace`, `InPlaceOrRecreate`, and `Recreate` update modes, the current `status.allocatedResources` matches the recommendations. If the recommendations differed from the current `status.allocatedResources`, the VPA would have already resized the pod.

That said, the VPA also takes disruptions into consideration when performing in-place pod updates, since this operation is not always 100% safe. For example, if only a single pod replica is running, the VPA may avoid performing an in-place update (this behavior is configurable by the user).

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,135 @@
...
+
+### CPU Startup Boost
+
+After the boost period completes and VPA resizes down, the node may appear underutilized. If Karpenter consolidates, the replacement pod will be boosted again at creation, potentially requiring more capacity than the steady-state spec suggests. Currently, `consolidateAfter` does not reset on resize activity, so Karpenter may evaluate the node for consolidation during the boost/unboost cycle.
```
omerap12: VPA stored some info in annotations, maybe Karpenter can leverage that?
For example:
```bash
- apiVersion: v1
  kind: Pod
  metadata:
    annotations:
      vpaCpuStartupBoost/nginx: '{"requests":{"cpu":"100m","memory":"128Mi"},"limits":{"cpu":"500m","memory":"256Mi"}
```
Once the pod is unboosted it will deleted.

designs/in-place-pod-vertical-scaling.md
omerap12: With `Initial` mode, however, we cannot make the same assumption because the recommendations may differ significantly from the current allocation, especially for pods that have been running for a long time.

designs/in-place-pod-vertical-scaling.md
omerap12: > if we want to make a change in upstream replicaSet or deployment controller (a great option if we get alignment)

I’m not sure that’s a realistic option. Requiring other controllers - especially core Kubernetes controllers to implement this behavior doesn’t seem scalable to me.
Also, what about workloads that aren’t managed by Deployment/ReplicaSet, but still support the scale subresource?
IMO, having a unified CRD/interface that Karpenter can read, while each workload autoscaler (whether VPA or something else) is responsible for managing it, sounds like a cleaner and more extensible approach.
I know @adrianmoisey may have a different opinion though 🙂

designs/in-place-pod-vertical-scaling.md
jigisha620: So if it's already unboosted then Karpenter wouldn't know. That's why we are proposing adding these details to the pod itself so any node autoscaler can read it from there and make correct decisions.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,135 @@
...
+
+## How This Affects Karpenter Today
+
+Karpenter's affected decision paths are **all disruption paths** - consolidation, drift, and expiration. All three use the same scheduling simulation that computes pod placement based on current `pod.spec.containers[].resources.requests`. Provisioning is unaffected because pending pods have their template requests (correct at creation time).
+
+### InPlace Mode
+
+During active resize, `spec.requests` and `status.allocatedResources` diverge. `spec.requests` reflects the desired value while `status.allocatedResources` reflects what the kubelet actually holds. This divergence can go in either direction — spec is lower during resize-down, higher during resize-up. Karpenter's view of node capacity is inaccurate until the resize completes. After resize completes, spec aligns with VPA's recommendation, so replacement cost is correctly estimated via `spec.requests`.
+
+### CPU Startup Boost
```
jmdeal, line 61: What's the sequencing with startup boost? Is the pod created with the initial recommendation size and then boosted? Or is it created at the boosted size? I think our best effort behavior should be respecting the startup boost size during disruption, but I'm wondering what the worst case scenario is if it is not respected.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,135 @@
...
+In Initial and Recreate modes, VPA's recommendation evolves over time. When Karpenter evicts a pod, the replacement receives VPA's current recommendation, which may differ from the evicted pod's spec. Karpenter has no way to predict this value today.
+
+## Goals
+
+- Ensure accurate node capacity accounting during active pod resize across all disruption paths (consolidation, drift, expiration)
+- Prevent unnecessary consolidation evaluation on nodes with active resize activity
+
+## Non-Goals
+
+- Predicting replacement pod requests for VPA Initial or Recreate modes, where the recommendation may differ from the current pod spec
```
jmdeal: We assume in-place does not differ because VPA will apply the update to spec.resources even if the resize operation is infeasible?

designs/in-place-pod-vertical-scaling.md
jigisha620: From what I understand, the pod is created at the boosted size. VPA admission webhook sets the boosted value (recommendation × factor) directly at creation time. So if Karpenter evicts a boosted pod (currently at the post-unboost size), the replacement will come up at the boosted size again. Worst case if we don't respect it, Karpenter simulates placement based on the current spec (post-unboost), the replacement doesn't fit on the destination (needs the boost size), goes Pending, and Karpenter provisions a new node.

designs/in-place-pod-vertical-scaling.md
jigisha620: Yes, VPA patches spec.requests via the resize subresource regardless of whether the kubelet can fulfill it. So even in the infeasible case, spec already reflects the recommendation. The replacement pod should get the same value via the webhook.

designs/in-place-pod-vertical-scaling.md
omerap12: > So if it's already unboosted then Karpenter wouldn't know.

Correct.

>  That's why we are proposing adding these details to the pod itself so any node autoscaler can read it from there and make correct decisions.

We discussed a couple of additional details you'd like to include in the pod annotations. Regarding resizing, we can certainly add more information there (we're already annotating pods with some resizing-related details today).

As for recommendations, I don't think it's reasonable to expect VPA (or any workload autoscaler) to update pod annotations every time recommendations change. The frequency of recommendation updates can be quite high, and annotating every pod on each change would introduce significant overhead.

designs/in-place-pod-vertical-scaling.md
omerap12: You are correct.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,135 @@
...
+| **Initial** | Admission webhook sets requests at pod creation only; no further updates |
+| **Recreate** | VPA evicts pods whose requests diverge from recommendation; webhook sets new value on replacement |
+| **InPlace** | VPA patches running pod's requests via resize subresource; webhook sets recommendation on replacement |
+| **InPlaceOrRecreate** | InPlace with eviction fallback if resize is infeasible |
+
+In all modes where the webhook is deployed, the replacement pod's requests are set to VPA's current recommendation at creation time.
+
+### CPU Startup Boost
+
+VPA can temporarily boost CPU requests at pod creation to accelerate startup (e.g., JVM warm-up), then resize back down after the pod becomes Ready. This is configured via `spec.startupBoost` on the VPA object and controlled by a `CPUStartupBoost` feature gate. The boosted value is computed dynamically (recommendation × factor, or a fixed quantity).
```
omerap12, line 51: FYI: it can also works with Off `updatePolicy`: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/enhancements/7862-cpu-startup-boost/README.md#workflow

```
If updatePolicy is Off for the VPA object, or mode is Off in a container policy, the VPA Updater will revert the CPU resources to the values specified in the pod spec.
```

jigisha620:
For Initial/Recreate modes and Startup Boost, `spec.requests` alone doesn't capture what the replacement pod will 
actually get. Two options we're considering:

Option A: Conditional dry-run - Customers annotate their Deployments (e.g., karpenter.sh/volatile-requests: "true") to signal that pod requests may change on replacement. For only those pods, Karpenter does a dry-run create against the API server to see what webhooks would set on the replacement. This is autoscaler-agnostic, works with VPA, custom  controllers, or any mutating webhook.

Option B: VPA dependency - Read the VPA object's `status.recommendation` directly to know what the webhook would set on the replacement. Can multiple VPAs apply to same pods? Do we need to replicate VPA's way to resolve those conflicts?

Thoughts? @omerap12 @adrianmoisey

adrianmoisey:
> Option A: Conditional dry-run - Customers annotate their Deployments (e.g., karpenter.sh/volatile-requests: "true") to signal that pod requests may change on replacement. For only those pods, Karpenter does a dry-run create against the API server to see what webhooks would set on the replacement. This is autoscaler-agnostic, works with VPA, custom controllers, or any mutating webhook.

I like this idea!

However, what about an annotation on the pod that you can read, to know it may not be recreated with as the same size?

May be something like `autoscaling.kubernetes.io/volatile-requests: true` for pods that are know to resize on recreation (for VPA that's Initial and startup boost pods).

Alternatively, `autoscaling.kubernetes.io/resizable: true` for all pods that live under a VPA?

(obviously we need to go through the process of getting this annotation formalised, but it's just an idea)

My preference would be to annotation all pods under the VPA with a common annotation, but, that may cause unnecessary lookups for you if pods sizes match the size of their replacement.
Something else to consider, someone may use the VPA in Recreate mode, and point it at a Deployment with a single replica. This would end up causing it to be treated in the same way as Initial, since the VPA won't (by default) evict it.
I assume it's an edge case though

omerap12:
> For Initial/Recreate modes and Startup Boost, `spec.requests` alone doesn't capture what the replacement pod will actually get. Two options we're considering:
> 
> Option A: Conditional dry-run - Customers annotate their Deployments (e.g., karpenter.sh/volatile-requests: "true") to signal that pod requests may change on replacement. For only those pods, Karpenter does a dry-run create against the API server to see what webhooks would set on the replacement. This is autoscaler-agnostic, works with VPA, custom controllers, or any mutating webhook.
> 
> Option B: VPA dependency - Read the VPA object's `status.recommendation` directly to know what the webhook would set on the replacement. Can multiple VPAs apply to same pods? Do we need to replicate VPA's way to resolve those conflicts?
> 
> Thoughts? @omerap12 @adrianmoisey

Option 1 seems more autoscaler-agnostic and doesn't require any RBAC changes, so I think it's preferable. The downside is that for every VPA-managed pod that Karpenter evicts, the API server ends up doing the work twice (an additional dry-run loop). I think that's probably acceptable since not all pods are targeted by VPA.

What does concern me is that not all customers/users will be aware that they need to add this annotation. Maybe we (the VPA folks) could automatically add these autoscaler-agnostic annotations, as Adrian mentioned above, so customers don't have to manage them themselves? Otherwise, customers would also need to remember to remove the annotation if they delete the VPA object, etc.

Maybe the CA folks would be interested in this as well? (cc @towca @jackfrancis @adrianludwin)

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+
+When Karpenter consolidates a node, it evicts pods that are then recreated by their controllers. If a mutating webhook (e.g., VPA's admission controller) changes the pod's resource requests on creation, the recreated pod may require more resources than the original. Without prediction, Karpenter might move pods to a node that can't fit them post-recreation, causing unnecessary pending pods.
+
+**Mechanism:**
+
+A prediction cache runs dry-run pod creates (`POST /api/v1/namespaces/{ns}/pods?dryRun=All`) to determine what a pod's resources would be after recreation. This goes through the full API server admission chain (all mutating webhooks) without persisting anything.
+
+**Opt-in via annotation:**
+
+Pods (or their templates) must be annotated with `autoscaling.k8s.io/volatile-requests: "true"` to signal that their requests may change on recreation. This annotation is set by the vertical pod autoscaler (or any autoscaler that uses a mutating webhook to adjust resources at pod creation). Karpenter only runs dry-runs for annotated pods, avoiding unnecessary API calls for workloads whose requests are static.
```
adrianmoisey: Since this is a `k8s.io` annotation, I assume we need permission or something to make it formal and list it here: https://kubernetes.io/docs/reference/labels-annotations-taints/

omerap12: I added this topic for the sig meeting today so we can discuss it.
@jigisha620 , Is anyone from the Karpenter team will be able to join?

jigisha620: I should be able to join.

ellistarn: cc: @towca, who may find this interesting for CAS

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+autoscaling.k8s.io/predicted-pod-requests: '{"cpu": "8500m", "memory": "2304Mi"}'
+```
+
+The autoscaler computes this by taking its recommendation for managed containers (with startup boost applied if configured), current spec for unmanaged containers, and computing the total using standard Ceiling logic.
+
+**Where the annotation lives:**
+
+Two variants:
+
+1. **On the owner (Deployment/StatefulSet)** (recommended): VPA patches `metadata.annotations` on the Deployment. One patch per workload per recommendation change. Karpenter resolves pod → RS → Deployment → read annotation.
```
ellistarn, line 199: This is a really interesting option.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+
+- Schema must be agreed upon across projects
+- If annotating live pods (variant 2): API load at scale (N patches per recommendation change)
+
+
+### 3. Call Webhook Endpoints Directly
+
+Discover `MutatingWebhookConfiguration` resources, construct `AdmissionReview` requests, and call webhook services. Gets exact results without pod create RBAC but is extremely complex (auth, ordering, chaining, network access) and fragile. Essentially re-implements the API server's admission chain.
+
+### 4. Delay or Block Consolidation During Resize
```
ellistarn, line 229: This doesn't cover the concerns we discuss around sufficiently large clusters losing their best consolidation candidates.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
+# InPlacePodVerticalScaling Support
```
ellistarn, line 1: nice! Broadly agree with the recommendations here. I am a bit queasy about the RBAC implications of the dry run functionality. I see it as a pragmatic alternative if we cannot influence pod autoscalers to tell us through other means.

x13n:
One more alternative from me, copying what I proposed on Slack:

> Dry run is one option, but - besides RBAC concerns - requires API calls on the critical path when considering pods replacement. If that was just one hop to apiserver, it would be less of a problem, but since this is meant to test admission, there may be multiple webhooks involved. One alternative I thought about was to standardize on "replacement pod" template. The way it would work:
> 
> - Owning controller would create PodTemplate in addition to Pods. It would be possible to discover it from owned pods.
> - CA/Karpenter would simply maintain a watch on PodTemplates and do in-memory lookup by owner ref. If there's no match, fall back to the pod object like today.
> - VPA (or any other controller modifying created pods) would not only modify Pods but also PodTemplates on admission.
> - There would have to be a channel to communicate when admission logic changes (in VPA case - different recommendation was calculated). Simplest option would be VPA deleting PodTemplate object and waiting for owner controller to recreate, but this again raises RBAC concerns. So maybe a more indirect solution would be required (separate resource? subresource on PodTemplate?).
> 
> 
> Of course, the downside of this solution is that every admission webhook would have to support it. OTOH, by having VPA support, we'd likely solve most cases where this is a problem today.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+| Resize DOWN completed | 1 | 1 | 1 | No change |
+| Resize UP in progress | 6 | 2 | 6 | Same as reading spec |
+| Resize UP infeasible | 6 | 2 | 6 | Conservative (treats as needing 6) |
+| Field not set (resize never occurred) | 4 | nil | 4 | Falls back to spec, backward compatible |
+
+### Part 2: Predict Post-Eviction Resource Requests via Dry-Run
+
+Part 1 fixes resource accounting during active resize. Part 2 ensures consolidation decisions account for how pods will look after recreation.
+
+When Karpenter consolidates a node, it evicts pods that are then recreated by their controllers. If a mutating webhook (e.g., VPA's admission controller) changes the pod's resource requests on creation, the recreated pod may require more resources than the original. Without prediction, Karpenter might move pods to a node that can't fit them post-recreation, causing unnecessary pending pods.
```
jmdeal: nit: the more case is the main one we want to make sure we address, but this is also beneficial for the less case since it will enable us to take consolidation actions we wouldn't have been able to otherwise.
```suggestion
When Karpenter consolidates a node, it evicts pods that are then recreated by their controllers. If a mutating webhook (e.g., VPA's admission controller) changes the pod's resource requests on creation, the recreated pod may require more or less resources than the original. Without prediction, Karpenter might move pods to a node that can't fit them post-recreation, causing unnecessary pending pods.
```

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+
+A prediction cache runs dry-run pod creates (`POST /api/v1/namespaces/{ns}/pods?dryRun=All`) to determine what a pod's resources would be after recreation. This goes through the full API server admission chain (all mutating webhooks) without persisting anything.
+
+**Opt-in via annotation:**
+
+Pods (or their templates) must be annotated with `autoscaling.k8s.io/volatile-requests: "true"` to signal that their requests may change on recreation. This annotation is set by the vertical pod autoscaler (or any autoscaler that uses a mutating webhook to adjust resources at pod creation). Karpenter only runs dry-runs for annotated pods, avoiding unnecessary API calls for workloads whose requests are static.
+
+**Flow:**
+
+1. A background controller lists pods with the `volatile-requests` annotation, groups them by owner (one representative per ReplicaSet), strips runtime fields, and submits dry-run creates. Pods without an owner are skipped, they won't be recreated after eviction.
```
jmdeal: Are there any counter-examples where pods provisioned via the same owner at the same time get different requests? It's hypothetically possible, but I'm wondering if there are any examples in practice that we would want to consider.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+2. The predicted `Ceiling()` result is cached, keyed by `(namespace, owner-uid)`.
+3. During consolidation simulation, annotated pods use their cached predicted resources instead of current `Ceiling(pod)`. Non-annotated pods use current resources as today.
+4. Before eviction, Karpenter re-runs a fresh dry-run for annotated pods in the consolidation plan. If the result differs from what the simulation assumed (e.g., VPA recommendation changed), the plan is aborted and re-evaluated on the next cycle.
```
jmdeal: This is part of consolidation evaluation, right? While that is before eviction, this wording implies to me that it would be right before eviction, i.e., after we've already started the drain flow for the node which is non-reversible.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
+
+**Pros:**
+
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
+
+**Cons:**
+
+- Requires `create pods` RBAC cluster-wide. RBAC does not distinguish dry-run from real creates.
```
jmdeal: Have there been any proposals to enhance the granularity of RBAC permissions to scope dry-run calls? I wouldn't want to block on that type of enhancement for this feature, just curious if there's been any ongoing discussion. It would make this an easier con to ignore if there was a fast path to tightening the RBAC scope.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+- During active rollouts (Deployment or StatefulSet), the recreated pod may use a newer template than what was dry-run'd. This can cause one incorrect consolidation where the pod doesn't fit on the target node. The pod goes pending until Karpenter provisions a new node. The cache self-corrects on the next refresh once the recreated pod is running with the new template.
```
jmdeal: Can we use the template hash as part of the cache key?

designs/in-place-pod-vertical-scaling.md
jmdeal: Do we always want to abort? What if the command would still be valid (e.g. the size of the pod was reduced or wasn't increased enough to invalidate the command)? In both cases I can think of, we should be able to validate the command without rerunning the scheduling simulation.

designs/in-place-pod-vertical-scaling.md
jmdeal, line 199: I had a similar comment on the previous section, but is there any scenario where VPA makes heterogeneous scaling decisions within an owner-ref group?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+2. Karpenter reads the annotation during consolidation simulation. If present, uses it as the predicted resources for pods being moved. If absent, uses current `Ceiling(pod)`.
+3. Before eviction, Karpenter re-reads the annotation. If it changed since simulation, abort and re-evaluate.
+
+**RBAC:**
+
+No additional RBAC for Karpenter (already reads pods and watches Deployments). VPA needs patch on Deployments/StatefulSets (if annotating the owner).
+
+**Pros:**
+
+- No additional RBAC for Karpenter
```
jmdeal, line 214: Do we have RBAC permissions for all of the owner-refs we would need? I don't think adding read permissions for those top-level objects would be problematic regardless though.

designs/in-place-pod-vertical-scaling.md
jmdeal, line 229: Does this solve the problem when in-place autoscaling isn't enabled?

designs/in-place-pod-vertical-scaling.md
jigisha620: Yeah, I think that's something we could do.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+
+### CPU Startup Boost
+
+After the boost period completes and VPA resizes down, the node may appear underutilized. If Karpenter consolidates, the replacement pod will be boosted again at creation, requiring more capacity than the steady-state spec suggests. Without prediction, Karpenter bases its consolidation decision on the current (post-boost) spec and may move the pod to a node that can't fit the boosted value on recreation.
+
+### Initial and Recreate Modes
+
+In Initial and Recreate modes, VPA's recommendation evolves over time. When Karpenter evicts a pod, the replacement receives VPA's current recommendation, which may differ from the evicted pod's spec. Karpenter has no way to predict this value today.
```
maxcao13: Might be being pedantic here, but VPA recommendations evolve over time no matter what the update mode is. Update mode only controls what the vpa-updater decides to do with the recommendation.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+## Goals
+
+- Ensure accurate node capacity accounting during active pod resize across all disruption paths (consolidation, drift, expiration)
+- Predict post-eviction pod resource requests so that consolidation decisions account for mutating webhooks (e.g., VPA) that may change a pod's resources on recreation
+
+## Proposal
+
+### Part 1: Use Allocated Resources During Active Resize
+
+When a pod is actively being resized down, its `spec.requests` reflects the new (lower) desired value, but `status.containerStatuses[].allocatedResources` reflects what the kubelet still physically holds. Karpenter today only reads `spec.requests`, which causes it to underestimate node utilization during the transition.
```
GnatorX, line 78: This is a bit confusing to read unless you understand the in place scaling RFC. I would clearly lay out what happens during in place scaling then explain why Karpenter runs into issues. Something like

When a pod is resized down:
1. spec.request or limit is updated
2. Tries to scale down
3. Scale down successful, resource freed
4. The kubelet updates AllocatedResources to the lower value.

Karpenter reads spec.request so before full scale down happens it already thinks there is free space etc..

designs/in-place-pod-vertical-scaling.md
GnatorX: I think its possible if you use VPA recommender and in place scaling. Let say the first pod was under some load but when VPA got to the next the load have significantly increased and it decided to pick an higher resource.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+| Normal (no resize) | 4 | 4 | 4 | No change |
+| Resize DOWN in progress | 2 | 6 | **6** | Correctly reflects committed resources |
+| Resize DOWN completed | 1 | 1 | 1 | No change |
+| Resize UP in progress | 6 | 2 | 6 | Same as reading spec |
+| Resize UP infeasible | 6 | 2 | 6 | Conservative (treats as needing 6) |
+| Field not set (resize never occurred) | 4 | nil | 4 | Falls back to spec, backward compatible |
+
+### Part 2: Predict Post-Eviction Resource Requests via Dry-Run
+
+Part 1 fixes resource accounting during active resize. Part 2 ensures consolidation decisions account for how pods will look after recreation.
```
GnatorX, line 109: IMO part 2 (prediction) should be a follow up separate from this RFC. Part 2 solutions are all a bit "hacky" or have a lot of unknown unknowns and I wonder if we should figure out a universal solution between Karpenter, CAS and VPA that make sense for all 3. 

To me running a dry run for a large number of pods from Karpenter against apiserver is going to very problematic.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+3. During consolidation simulation, annotated pods use their cached predicted resources instead of current `Ceiling(pod)`. Non-annotated pods use current resources as today.
+4. Before eviction, Karpenter re-runs a fresh dry-run for annotated pods in the consolidation plan. If the result differs from what the simulation assumed (e.g., VPA recommendation changed), the plan is aborted and re-evaluated on the next cycle.
+5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
+
+**Pros:**
+
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
+
+**Cons:**
```
GnatorX: - making `create pod` is a huge con imo regardless of dry-run or not.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,239 @@
...
+2. The predicted `Ceiling()` result is cached, keyed by `(namespace, owner-uid)`.
+3. During consolidation simulation, annotated pods use their cached predicted resources instead of current `Ceiling(pod)`. Non-annotated pods use current resources as today.
+4. Before eviction, Karpenter re-runs a fresh dry-run for annotated pods in the consolidation plan. If the result differs from what the simulation assumed (e.g., VPA recommendation changed), the plan is aborted and re-evaluated on the next cycle.
+5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
+
+**Pros:**
+
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
```
GnatorX: I dont find this as a pro? VPA is an CRD that existed for a long time and feels as reasonable of a CRD as PDB to be consumed by Karpenter. Its also kinda the same point as https://github.com/kubernetes-sigs/karpenter/pull/3037/changes#diff-4c0dff63090da4730bbc149976f3d42283a69309d87938963ca5e03c519404e1R131

designs/in-place-pod-vertical-scaling.md
jigisha620: That's true for the running state, pods can end up with different requests if VPA resized them at different times. But for prediction (what would the pod get on recreation), the webhook applies VPA's current status.recommendation.target, which should be uniform for all pods in the group at a given point in time.

omerap12:
After the SIG meeting today, I would like to suggest another option for this integration. This is only my personal opinion, and I'm not sure we should go with it - others may disagree with the idea.

We (VPA) could expose a synchronous server-side API in the VPA that takes a PodSpec and the Pod metadata as input and returns the recommended resources.

This was actually intended to be part of the original VPA design proposal (a very long time ago), and I have no idea why it was never implemented. If this would help the node autoscaler folks, I think we (VPA) should implement it. 
Again, this is just my personal opinion, and I'm not speaking for the other VPA maintainers (@maxcao13, @adrianmoisey).

https://github.com/kubernetes/design-proposals-archive/blob/main/autoscaling/vertical-pod-autoscaler.md?plain=1#L474-L477

cc @kubernetes-sigs/sig-autoscaling-leads @elmiko

adrianmoisey:
> After the SIG meeting today, I would like to suggest another option for this integration. This is only my personal opinion, and I'm not sure we should go with it - others may disagree with the idea.
> 
> We (VPA) could expose a synchronous server-side API in the VPA that takes a PodSpec and the Pod metadata as input and returns the recommended resources.
> 
> This was actually intended to be part of the original VPA design proposal (a very long time ago), and I have no idea why it was never implemented. If this would help the node autoscaler folks, I think we (VPA) should implement it. Again, this is just my personal opinion, and I'm not speaking for the other VPA maintainers

I'm supportive of this idea.
In a slack thread I proposed a webhook idea where we register a API (similar to the metrics API) in k/k, and get the VPA to listen on that path, much like how metrics-server works today, but for "pod recommendations".

The part I'm unsure on how to handle is when a pod is known to resize (ie: CPU boost)

omerap12:
Would that solution work for you?
@DerekFrank @jmdeal

maxcao13:
> In a slack thread I proposed a webhook idea where we register a API (similar to the metrics API) in k/k, and get the VPA to listen on that path, much like how metrics-server works today, but for "pod recommendations".

I think this makes sense to me. Maybe VPA could be the first to implement this new API and be able to serve arbitrary recommendations based on the `VPA.status.recommendation` and CPU boost number. For CPU requests and CPU boost in particular, I'm assuming we would recommend the `max(boost, request)` number, so Karpenter doesn't underprovision. This allows any other recommendation system to register as well, though now we have to mandate/document that any other non-VPA recommender/mutating-webhook would have to go through this new API in order to work with Karpenter.

I agree with @GnatorX that the prediction mechanics probably should be separated out since it seems these concerns are going way outside of the VPA project now. I think this type of extension-api solution or any general solution would need probably need a KEP. Plus, I'm assuming sig-autoscaling would be the ones maintaining this API + server if we are not planning this to be some bespoke one-off VPA plugin?

I liked the dry-run idea because it required the least amount of work from both sides, but the rbac concern and the extra apiserver requests do make it hacky without further k8s architecture changes.

omerap12:
Yeah, I agree with both @maxcao13 and @adrianmoisey.
Before we start implementing anything, I'd like to hear whether this is something node autoscalers are interested in.

elmiko:
> elmiko

i think this sounds like a nice path forward. it reminds me a little of the grpc work that has been done in the CA for expanders and even an entire provider.

++ from me.

designs/in-place-pod-vertical-scaling.md
jigisha620: Thanks for the feedback! I think the scenario table a few lines below covers this, it shows the step-by-step divergence between spec and allocatedResources during resize-down, along with the Node-B example showing how Karpenter's view diverges from the scheduler's. Let me know if you think it still needs more context after seeing those together.

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,237 @@
...
+- Immediate reaction to recommendation changes via watch
+
+**Cons:**
+
+- Couples Karpenter to VPA's CRD schema
+- Only works for VPA, other resource-mutating webhooks are not covered
+- Replicates webhook logic: `controlledResources`, `updateMode`, startup boost computation, container name matching and merging
+
+
+## Alternatives Considered
```
omerap12, line 154: Can we also add https://github.com/kubernetes-sigs/karpenter/pull/3037#issuecomment-4743741643 to that list? I'm not sure this is the direction we'd ultimately want to pursue, but it would be useful to document it so we know this option was considered if we revisit the topic in the future.

jigisha620: @omerap12 I could add it here or do you think it's better tracked through a GH issue here - https://github.com/kubernetes/autoscaler/issues? I can create one.

omerap12: Yeah, it would be better to open an issue in the autoscaler repo. In today’s SIG meeting, @towca said that CA is also interested in this, so we should track it there. Once you create the issue, we can link this RFC to it.

GnatorX:
I know we looked at this from pod's perspective but we likely would do this for daemonsets too. Could we make sure this design works with that. https://github.com/kubernetes/kubernetes/issues/122753. Specifically using resizing to resize daemonset when larger or smaller nodes etc.

designs/in-place-pod-vertical-scaling.md
jigisha620: Done - https://github.com/kubernetes/autoscaler/issues/9898

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,237 @@
...
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
+
+**Cons:**
+
+- Requires `create pods` RBAC cluster-wide. RBAC does not distinguish dry-run from real creates.
+- During active rollouts (Deployment or StatefulSet), the recreated pod may use a newer template than what was dry-run'd. This can cause one incorrect consolidation where the pod doesn't fit on the target node. The pod goes pending until Karpenter provisions a new node. The cache self-corrects on the next refresh once the recreated pod is running with the new template.
+
+
+### 2. Annotation Carrying the Prediction
```
IdoKendo, line 183: I like the idea of using an annotation as an alternative to the proposed approach, allowing controllers other than VPA to help Karpenter make more informed consolidation decisions.

I even think that adding a Karpenter-specific annotation such as `karpenter.sh/predicted-pod-requests` could be a good middle-ground. I see it as a complementary mechanism that supports two ways of handling pods whose resource requests differ at creation time—whether they're managed by VPA or by another controller.

WDYT?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,237 @@
...
+
+- No additional RBAC for Karpenter
+- Autoscaler-agnostic (any controller can adopt the convention)
+- Simple read path for Karpenter (just read annotation)
+- No coupling to specific CRDs
+
+**Cons:**
+
+- Schema must be agreed upon across projects
+- If annotating live pods (variant 2): API load at scale (N patches per recommendation change)
```
IdoKendo, line 222: I would imagine that while changing the requests of a pod, the controller will also add the annotation, so this shouldn't cause scaling issues, or maybe I am misunderstanding some logic here?

designs/in-place-pod-vertical-scaling.md
```diff
@@ -0,0 +1,237 @@
...
+5. The cache refreshes periodically (e.g., every 60s). On startup, consolidation is gated until the first pass completes (consistent with how other caches like instance types and cluster state gate disruption decisions). If no annotated pods exist, this completes immediately.
+
+**Pros:**
+
+- Autoscaler-agnostic: works with any mutating webhook (VPA, custom controllers, sidecar injectors, etc.)
+- No coupling to specific CRDs or autoscaler implementations
+
+**Cons:**
+
+- Requires `create pods` RBAC cluster-wide. RBAC does not distinguish dry-run from real creates.
```
adrianmoisey, line 179: The discussion over at https://github.com/kubernetes/autoscaler/issues/9898 made me wonder if we should be looking at this `Dry-Run Pod Creates` again.

The list of pros here is very compelling.
What I'm wondering is if a new subresource could be created in k/k that only allows a dry-run.

That way any tooling that needs to make a prediction could do so by having access to that dry-run subresource without having the ability to create pods.

It should solve for any resource that could be mutated on admission.
The one downside I can think of is that if anyone injects secrets into a Pod directly, it could mean that whomever has access to that dry-run subresource could read that secret. (This feels like an edge case, and, also why the subresource will live behind RBAC)

DerekFrank (approved):
I love the discussion about solutions for the broader pod autoscaler -> node autoscaler communication pathway, but the karpenter project is not going to block on that conversation. We're going to move forward with directly consuming VPA's objects, then deprecating that later on a karpenter version boundary

Lets move the potential solution discussion into https://github.com/kubernetes/autoscaler/issues/9898

DerekFrank (approved):
/lgtm

omerap12:
> I love the discussion about solutions for the broader pod autoscaler -> node autoscaler communication pathway, but the karpenter project is not going to block on that conversation. We're going to move forward with directly consuming VPA's objects, then deprecating that later on a karpenter version boundary
> 
> Lets move the potential solution discussion into [kubernetes/autoscaler#9898](https://github.com/kubernetes/autoscaler/issues/9898)

ACK.
Feel free to reach out if anything is needed from VPA side.

PR #3063, opened 2026-06-01
chore: Consolidate pod force-deletion into the eviction queue

### LINKED ISSUE #3030: Refactor: Consolidate pod force-deletion into the eviction queue
### Description

Currently there are two pod removal paths during node termination:

1. **Eviction Queue** (`terminator/eviction.go`): async, uses Kubernetes Eviction API, respects PDBs, exponential backoff, emits metrics and events
2. **Terminator.DeleteExpiringPods()** (`terminator/terminator.go:140`): synchronous `kubeClient.Delete()` in the reconcile loop, bypasses PDBs when terminationGracePeriod is expiring

The force-deletion path should be consolidated into the eviction queue to:

- Unify observability (metrics, events) for all pod removals
- Avoid synchronous API calls in the reconcile loop
- Make the "bypasses PDBs" behavior explicit and configurable in one place
- Simplify the Terminator's `Drain()` method which currently mixes filtering, force-deletion, and queue submission

Medium — this is a code quality improvement that simplifies the architecture and improves observability, but the current behavior is functionally correct.

ketanjani21:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/3030

**Description**

### Description

Currently there are two pod removal paths during node termination:

1. **Eviction Queue** (`terminator/eviction.go`): async, uses Kubernetes Eviction API, respects PDBs, exponential backoff, emits metrics and events
2. **Terminator.DeleteExpiringPods()** (`terminator/terminator.go:140`): synchronous `kubeClient.Delete()` in the reconcile loop, bypasses PDBs when terminationGracePeriod is expiring

This PR consolidates the  force-deletion path  into the eviction queue to:

- Unify observability (metrics, events) for all pod removals
- Avoid synchronous API calls in the reconcile loop
- Make the "bypasses PDBs" behavior explicit and configurable in one place
- Simplify the Terminator's `Drain()` method which currently mixes filtering, force-deletion, and queue submission

 - New unit tests in `terminator/suite_test.go` covering force-delete via the queue: bypassing a blocking PDB, upgrading an already-enqueued eviction to force-delete, and cleaning up the queue entry when the pod is already gone.
- Updated `node/termination/suite_test.go` for the now-asynchronous force-delete
    (the pod is removed via a queue reconcile rather than inline in `Drain`).

```
 KUBEBUILDER_ASSETS="$(setup-envtest use 1.34.1 -p path)" go test -race ./pkg/controllers/node/termination/...

ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination	110.124s
ok  	sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator	19.539s
?   	sigs.k8s.io/karpenter/pkg/controllers/node/termination/terminator/events	[no test files]
```

### DIFF
--- a/pkg/controllers/node/termination/terminator/eviction.go
+++ b/pkg/controllers/node/termination/terminator/eviction.go
@@ -53,23 +53,11 @@ import (
 	podutils "sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
-// evictionMode determines how the Queue removes a pod from its node.
-type evictionMode int
-
-const (
-	// modeEvict uses the Kubernetes eviction subresource and respects PDBs.
-	modeEvict evictionMode = iota
-	// modeForceDelete uses kubeClient.Delete with a clamped grace period and
-	// bypasses PDBs. Used when a pod's terminationGracePeriodSeconds would
-	// otherwise extend past the node's terminationGracePeriod.
-	modeForceDelete
-)
-
-// queueItem holds the per-pod state needed by the Queue's reconciler.
+// queueItem holds the per-pod state needed by the Queue's reconciler. The
+// reconciler decides at reconcile time whether to evict or force-delete based
+// on nodeTerminationTime and the pod's grace period — there is no upfront
+// per-pod mode. A nil nodeTerminationTime means "no deadline; always evict".
 type queueItem struct {
-	mode evictionMode
-	// nodeTerminationTime is set when mode == modeForceDelete. It is used to
-	// clamp the pod's grace period at delete time.
 	nodeTerminationTime *time.Time
 }
 
@@ -158,39 +146,19 @@ func (q *Queue) Register(ctx context.Context, m manager.Manager) error {
 		Complete(reconcile.AsReconciler(m.GetClient(), q))
 }
 
-// Add adds pods to the Queue for eviction via the Kubernetes eviction API.
-// Pods already enqueued (in any mode) are left untouched so that an in-flight
-// force-delete is not downgraded to an eviction.
-func (q *Queue) Add(pods ...*corev1.Pod) {
+// Add enqueues pods for drain. The queue decides per reconcile whether to
+// evict (respecting PDB) or force-delete based on whether the pod's grace
+// period would extend past nodeTerminationTime; pass nil to disable the
+// force-delete path entirely. Re-adding a pod overwrites its deadline so the
+// next reconcile picks up the latest value.
+func (q *Queue) Add(nodeTerminationTime *time.Time, pods ...*corev1.Pod) {
 	q.Lock()
 	defer q.Unlock()
 
 	for _, pod := range pods {
 		qk := NewQueueKey(pod)
-		if _, ok := q.items[qk]; ok {
-			continue
-		}
-		q.items[qk] = queueItem{mode: modeEvict}
-		q.source <- event.TypedGenericEvent[*corev1.Pod]{Object: pod}
-	}
-}
-
-// AddForceDelete adds pods to the Queue to be force-deleted with a grace period
-// clamped to the node's remaining terminationGracePeriod. This bypasses PDBs
-// and the do-not-disrupt annotation. If a pod is already enqueued for eviction,
-// its entry is upgraded to force-delete so that the next reconcile honors the
-// new mode.
-func (q *Queue) AddForceDelete(nodeTerminationTime *time.Time, pods ...*corev1.Pod) {
-	q.Lock()
-	defer q.Unlock()
-
-	for _, pod := range pods {
-		qk := NewQueueKey(pod)
-		// Overwrite any existing entry so a pod already enqueued for eviction is
-		// upgraded to force-delete. The reconciler reads the mode at the start of
-		// each reconcile, so the upgrade takes effect on the next attempt.
 		_, enqueued := q.items[qk]
-		q.items[qk] = queueItem{mode: modeForceDelete, nodeTerminationTime: nodeTerminationTime}
+		q.items[qk] = queueItem{nodeTerminationTime: nodeTerminationTime}
 		if !enqueued {
 			q.source <- event.TypedGenericEvent[*corev1.Pod]{Object: pod}
 		}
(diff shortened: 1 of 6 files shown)


pkg/controllers/node/termination/terminator/suite_test.go
```diff
@@ -218,31 +216,50 @@ var _ = Describe("Eviction/Queue", func() {
...
+		It("should force-delete a pod via the queue, bypassing PDBs", func() {
+			pod.Spec.TerminationGracePeriodSeconds = lo.ToPtr[int64](120)
+			ExpectApplied(ctx, env.Client, pdb, pod, node)
+			ExpectManualBinding(ctx, env.Client, pod, node)
 
-			nodeTerminationTime := time.Now().Add(time.Minute * 5)
-			Expect(terminatorInstance.DeleteExpiringPods(ctx, []*corev1.Pod{pod}, &nodeTerminationTime)).To(Succeed())
-			ExpectExists(ctx, env.Client, pod)
-			Expect(recorder.Calls(events.Disrupted)).To(Equal(0))
+			nodeTerminationTime := time.Now().Add(time.Minute * 1)
```
jigisha620: Since the queue computes grace period with the injected fake clock `q.clock.Now()`, should these tests use `env.Clock.Now()` instead of `time.Now()` for computing nodeTerminationTime?

jigisha620 (commented):
Thanks for making the change! One comment.

pkg/controllers/node/termination/terminator/suite_test.go
ketanjani21: Thank you for catching this. Agree with you, I've updated the tests to use `env.Clock.Now()`. Please review again ?

jigisha620 (approved):
/lgtm

ketanjani21:
@jigisha620 - Thank you for reviewing. Do we need to assign it to [maciekpytel](https://github.com/maciekpytel) for further approval? can you please help with that?

ketanjani21:
@MaciekPytel - can you please help with reviewing/approving this PR? Thank you.

ketanjani21:
@jigisha620 - I've rebased the branch as the branch was outdated. Would you mind reviewing again please? Thank you.

jigisha620 (approved):
/lgtm

jigisha620 (approved):
/lgtm

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -53,6 +53,26 @@ import (
...
+	// modeEvict uses the Kubernetes eviction subresource and respects PDBs.
+	modeEvict evictionMode = iota
+	// modeForceDelete uses kubeClient.Delete with a clamped grace period and
+	// bypasses PDBs. Used when a pod's terminationGracePeriodSeconds would
+	// otherwise extend past the node's terminationGracePeriod.
+	modeForceDelete
+)
+
+// queueItem holds the per-pod state needed by the Queue's reconciler.
+type queueItem struct {
```
DerekFrank: I'd rather have the queue get all the information it needs to perform all of the pod related drain steps

The idea behind the issue was to allow the terminator to fire-and-forget the 'pod eviction\deletion' workflow to the eviction queue and let it perform the timestamp logic.

We need the PR to simplify things enough to justify touching an operationally sensitive part of the code.

DerekFrank (changes requested):

pkg/controllers/node/termination/terminator/eviction.go
jigisha620: The PR does consolidate the delete call, unify observability, and remove sync API calls, but your vision is for the queue to also own the timing decision so `Drain()` is purely filter-and-enqueue. Happy to re-review once restructured.

ketanjani21: @DerekFrank - Thank you for the feedback. I've refactored the changes accordingly. Does this match with your vision?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -136,15 +146,20 @@ func (q *Queue) Register(ctx context.Context, m manager.Manager) error {
 		Complete(reconcile.AsReconciler(m.GetClient(), q))
 }
 
-// Add adds pods to the Queue
-func (q *Queue) Add(pods ...*corev1.Pod) {
+// Add enqueues pods for drain. The queue decides per reconcile whether to
+// evict (respecting PDB) or force-delete based on whether the pod's grace
+// period would extend past nodeTerminationTime; pass nil to disable the
+// force-delete path entirely. Re-adding a pod overwrites its deadline so the
```
DerekFrank: Do we ever need to re-add a pod?

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -53,6 +53,14 @@ import (
 	podutils "sigs.k8s.io/karpenter/pkg/utils/pod"
 )
 
+// queueItem holds the per-pod state needed by the Queue's reconciler. The
+// reconciler decides at reconcile time whether to evict or force-delete based
+// on nodeTerminationTime and the pod's grace period — there is no upfront
+// per-pod mode. A nil nodeTerminationTime means "no deadline; always evict".
+type queueItem struct {
```
DerekFrank: I don't think we need a struct for a single type, we can simplify that

pkg/controllers/node/termination/terminator/eviction.go
DerekFrank: If so, maybe we should do a min(old time, new time)? If a less forceful delete comes in after, we should honor the more forceful one.

If we don't expect that to happen, maybe that should be an error instead

pkg/controllers/node/termination/terminator/terminator.go
DerekFrank: We should fix this link here too: https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/

pkg/controllers/node/termination/terminator/eviction.go
```diff
@@ -154,20 +169,62 @@ func (q *Queue) Has(pod *corev1.Pod) bool {
...
+		return reconcile.Result{Requeue: true}, nil
+	}
+	return q.evict(ctx, pod)
+}
+
+// needsForceDelete reports whether the pod should be force-deleted now: its
+// grace period would (or already does) extend past nodeTerminationTime if
+// allowed to run on its own. Re-evaluated on every reconcile, so a pod stuck
+// on a PDB upgrades to force-delete naturally once the deadline passes.
+func needsForceDelete(pod *corev1.Pod, nodeTerminationTime *time.Time, clk clock.Clock) bool {
```
DerekFrank, line 226: nit: 
```suggestion
func (q *Queue) needsForceDelete(pod *corev1.Pod, nodeTerminationTime *time.Time) bool {
```

pkg/controllers/node/termination/terminator/terminator.go
DerekFrank, line 138: I do wonder if we need to group pods by priority anymore. The queue doesn't provide any guarantees around the order in which they are drained without us making it a priority queue and I don't know if that refactor is worth the complexity

@jigisha620 do you have a strong opinion?

DerekFrank:
@ketanjani21 are you still planning to work on this?

ketanjani21:
> @ketanjani21 are you still planning to work on this?

yes, sorry for the delay.

pkg/controllers/node/termination/terminator/terminator.go
ketanjani21: Fair point. I think this is still doing work. The loop returns after adding the first non-empty tier, so critical/daemon pods stay off the queue entirely until the earlier tier fully drains. For eg. under load or PDB pressure this keeps system-node-critical / CNI / kube-proxy pods from competing with user-pod evictions and preserves the graceful-shutdown ordering. I agree that making the queue itself priority-aware would be a bigger refactor, so for now leaning towards keeping it as-is, unless you have strong opinions. Happy to update the PR as we prefer.

pkg/controllers/node/termination/terminator/eviction.go
ketanjani21: updated, Thank you

pkg/controllers/node/termination/terminator/terminator.go
ketanjani21: updated, Thank you

pkg/controllers/node/termination/terminator/eviction.go
ketanjani21: Thank you for the suggestion, updated the PR accordingly as pods might get re-added.

ketanjani21:
@DerekFrank  - I've addressed all your feedback, the PR is ready for another review when you get a chance. Thank you.

DerekFrank (commented):
Apologies, I lost track of this PR :(

One thing on Drain: delete is now gated behind priority-tier draining, but it wasn't before — a past-deadline pod in a later tier now waits for earlier tiers to drain. At a hard deadline we should delete it regardless of tier. Suggest splitting the two paths: enqueue delete-eligible pods immediately (all tiers), keep wave ordering only for graceful eviction. Extracting needsDelete to a shared func lets both Drain and the queue reuse one predicate. 

WYT?

DerekFrank:
Sorry to leave you hanging on this one btw, I must have lost track of it

ketanjani21:
> Apologies, I lost track of this PR :(
> 
> One thing on Drain: delete is now gated behind priority-tier draining, but it wasn't before — a past-deadline pod in a later tier now waits for earlier tiers to drain. At a hard deadline we should delete it regardless of tier. Suggest splitting the two paths: enqueue delete-eligible pods immediately (all tiers), keep wave ordering only for graceful eviction. Extracting needsDelete to a shared func lets both Drain and the queue reuse one predicate.
> 
> WYT?

Derek - Thank you for reviewing again, agree with your assessment. I've updated the PR accordingly, please let me know if that works.

DerekFrank (approved):
/lgtm

PR #3064, opened 2026-06-02
fix: pairwise non-prefix fallback for multi-node consolidation

### LINKED ISSUE #1962: Consolidation does not happen even when there is cheaper combination of instances available

jamesmthompson:
## Summary

Fixes #1962. When a candidate that blocks joint deletion (e.g. a node whose pod can't reschedule) sorts early in the disruption-cost order, every prefix containing it fails the simulator. The binary search exits empty even when a non-prefix subset would consolidate fine.

- Adds a pairwise non-prefix walk that runs only when binary search returns no plan
- Bounded by a 10s timeout (`MultiNodePairwiseSearchTimeoutDuration`), so worst-case end-to-end is 1m10s
- Includes an envtest characterization test that fails on mainline and passes with the fix
- Follow-up cleanup replacing append aliasing with `slices.Clone`

## Test plan

- [x] Existing disruption suite passes (`go test ./pkg/controllers/disruption/...`)
- [x] New `multinode_1962_test.go` exercises the exact shape from the issue: three candidates sorted [good_0, bad, good_2] where bad blocks prefixes but the non-prefix {good_0, good_2} consolidates
- [x] Verify no regression on binary-search-solvable inputs (equivalence preserved)


pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -112,14 +123,53 @@ func (m *MultiNodeConsolidation) ComputeCommands(ctx context.Context, disruption
...
+	if primary.Decision() != NoOpDecision {
+		// Existing behavior: binary search found a feasible prefix.
+		return primary, nil
+	}
+
+	// Binary search exhausted without finding a feasible prefix
+	// (the karpenter#1962 case). Fall back to the pairwise non-prefix
+	// walk, bounded by its own timeout so the latency added on big
+	// clusters stays capped.
+	return m.pairwiseSearchFallback(ctx, candidates, max)
```
Ignoramuss, line 163: Is there any improvement we can make by passing the information computed by binary search to the pairwise walk to further reduce the search space?

pkg/controllers/disruption/multinode_pairwise_test.go
```diff
@@ -0,0 +1,183 @@
+/*
+Copyright The Kubernetes Authors.
+
```
Ignoramuss, line 3: nit: file name should probably omit the 1962 reference

pkg/controllers/disruption/multinodeconsolidation.go
```diff
@@ -170,6 +220,91 @@ func (m *MultiNodeConsolidation) firstNConsolidationOption(ctx context.Context,
...
+		}
+
+		candidatesToConsolidate := append(slices.Clone(accepted), c)
+
+		// When accepted is empty this probes a single candidate, which
+		// single-node consolidation already evaluated. The probe is still
+		// needed: the walk uses the result to decide whether to accept or
+		// skip c. Without it, a bad first candidate would poison the
+		// entire accepted set with no way to eject it.
+		cmd, err := m.computeConsolidation(timeoutCtx, candidatesToConsolidate...)
```
Ignoramuss, line 275: Each call to SimulateScheduling re-runs `cluster.DeepCopyNodes()`, `GetPendingPods()`, and `pdb.NewLimits()` which are likely invariants across calls in a single reconcile. Not something that needs to go with this change, but worth considering snapshotting a simulation state by breaking down `SimulateScheduling` into a prepare + simulate?

pkg/controllers/disruption/multinode_pairwise_test.go
jamesmthompson: Done — renamed to `multinode_pairwise_test.go`.

pkg/controllers/disruption/multinodeconsolidation.go
jamesmthompson: Agreed — the repeated DeepCopyNodes/GetPendingPods/pdb.NewLimits work is wasteful when we're probing N candidates in the same reconcile cycle. Added a TODO in the code noting the prepare+simulate split as a future optimization. Out of scope here since the 10s timeout bounds the cost, but it would be a nice follow-up especially if we want to raise that budget later.

pkg/controllers/disruption/multinodeconsolidation.go
jamesmthompson: Not meaningfully, no. When binary search exits empty, the only information it produced is "no prefix [0:k] is feasible for any k." That tells us something is blocking prefixes, but it doesn't identify *which* candidate is the blocker — a failed prefix [0:mid] could fail because of one bad element, or because of a pair interaction between two elements that are individually fine.

To extract useful signal we'd need to do individual feasibility probes (O(N) calls), which is exactly what the pairwise walk already does as its first action on each candidate. So the binary search doesn't give us a shortcut here.

jamesmthompson: One additional thought: once Balanced consolidation lands (#2962), we could pre-filter candidates whose individual savings don't clear the score gate before entering the pairwise walk. That would shrink the candidate set and let us cover more of the search space within the 10s budget. Gated on that RFC merging first, but it's a natural follow-up.

PR #3065, opened 2026-06-02
chore: use status resources in pod resource calculation

jigisha620:
Fixes #N/A 

**Description**
Sets `UseStatusResources: true` so that pod resource calculations account for in-place resized pods.

With in-place pod vertical scaling, a pod's actual resource allocation can differ from `spec.requests` (e.g. during an in-progress resize-down, the pod still holds its old allocation). The kube-scheduler already uses UseStatusResources to account for this. Without this change, Karpenter's scheduling simulation diverges from the scheduler, causing consolidation decisions the scheduler may not be able to fulfill.

Added unit tests.

### DIFF
--- a/pkg/utils/resources/resources.go
+++ b/pkg/utils/resources/resources.go
@@ -109,10 +109,12 @@ func SubtractFrom(dest v1.ResourceList, src v1.ResourceList) {
 }
 
 // Ceiling computes the effective resource requirements for a given Pod,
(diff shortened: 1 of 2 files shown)


Sarthug99:
LGTM 🚀

DerekFrank (approved):
/lgtm 🔥

DerekFrank (approved):
/lgtm

PR #2866, opened 2026-02-18
docs: RFC to separately cache instance types and offerings

DerekFrank:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/2605

**Description**

**How was this change tested?**

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3068, opened 2026-06-03
perf: memory optimizations in cluster cost tracking

ryan-mist:
Fixes #N/A 

**Description**
Adds 2 memory optimizations for node pool cluster cost tracking:
1. Hash Based Change Detection
> Previously we stored `map[OfferingKey]float64` per NodePool and doing map equality comparison
  each reconcile. This PR changes it to a single hash which reduces memory from `O(offerings)`
  to `O(1)` per NodePool for the diff check.

2. Lazy Offering Tracking
> Previously we tracked the price of every offering for every NodePool, even if the instance was not in use. This PR changes it so we only track the price of offerings when there is an active NodeClaim for that offering.

**Testing Scenario**: 25 wide-open NodePools with 1 NodeClaim per NodePool launched

Screenshot 2026-06-03 at 11 13 56 AM

> Yellow - v1.9.0 (without cluster cost)
Green - mainline
Blue - this PR

`make presubmit`

**Testing Scenario (between commits during development)**: 25 wide-open NodePools with 1 NodeClaim per NodePool launched

Screenshot 2026-06-02 at 10 05 15 PM

> Yellow - v1.9.0 (without cluster cost)
Blue - v1.10.0 (with cluster cost)
Red - commit 193d066f0b6516b8c39019e43721f04116da668d (only store prices of used instances)
Orange - commit b941319b14fa3e4448ea67c26ce38e746ce9033c (hashing offerings)

### DIFF
--- a/pkg/controllers/state/informer/pricing.go
+++ b/pkg/controllers/state/informer/pricing.go
@@ -23,9 +23,7 @@ import (
 
 	"github.com/awslabs/operatorpkg/reconciler"
 	"github.com/awslabs/operatorpkg/singleton"
-	"github.com/mitchellh/hashstructure/v2"
 	"go.uber.org/multierr"
-	"k8s.io/apimachinery/pkg/types"
 
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -44,7 +42,6 @@ type PricingController struct {
 	client        client.Client
 	cloudProvider cloudprovider.CloudProvider
 	clusterCost   *cost.ClusterCost
-	npOfMap       map[types.NamespacedName]uint64
 }
 
 func NewPricingController(client client.Client, cloudProvider cloudprovider.CloudProvider, clusterCost *cost.ClusterCost) *PricingController {
@@ -52,7 +49,6 @@ func NewPricingController(client client.Client, cloudProvider cloudprovider.Clou
 		client:        client,
 		cloudProvider: cloudProvider,
 		clusterCost:   clusterCost,
-		npOfMap:       make(map[types.NamespacedName]uint64),
 	}
 }
 
@@ -63,7 +59,6 @@ func (c *PricingController) Reconcile(ctx context.Context) (reconciler.Result, e
 		return reconciler.Result{}, err
 	}
 
-	newNpOfMap := make(map[types.NamespacedName]uint64, len(npl.Items))
 	var errs error
 	for _, np := range npl.Items {
 		newIts, err := c.cloudProvider.GetInstanceTypes(ctx, &np)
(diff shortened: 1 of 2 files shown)


pkg/state/cost/cost.go
```diff
@@ -330,13 +282,13 @@ func (cc *ClusterCost) internalRemoveOffering(npName string, offeringKey Offerin
 		return fmt.Errorf("attempted to remove nonexistent offering from nodepool %q (instance, %q, zone, %q, capacity, %q)", npName, offeringKey.InstanceName, offeringKey.Zone, offeringKey.CapacityType)
 	}
 
-	oc.Count -= 1
+	oc.Count--
 	npc.offeringCounts[offeringKey] = oc
```
AndrewMitchell25: nit: maybe we should put the if/delete block before this and put this in an else? Seems like no need to set it if we're just gonna delete it

AndrewMitchell25 (approved):
LGTM 🚀

pkg/controllers/state/informer/pricing.go
```diff
@@ -94,20 +89,26 @@ func (c *PricingController) Reconcile(ctx context.Context) (reconciler.Result, e
 	return reconciler.Result{RequeueAfter: 1 * time.Hour}, nil
 }
 
-func equal(oldOfs map[cost.OfferingKey]float64, newOfs map[cost.OfferingKey]float64) bool {
-	if len(lo.Values(oldOfs)) != len(newOfs) {
-		return false
+func hashOfferings(instanceTypes []*cloudprovider.InstanceType) uint64 {
+	type offeringSnapshot struct {
```
DerekFrank: OfferingKey?

pkg/controllers/state/informer/pricing.go
```diff
@@ -94,20 +89,26 @@ func (c *PricingController) Reconcile(ctx context.Context) (reconciler.Result, e
 	return reconciler.Result{RequeueAfter: 1 * time.Hour}, nil
 }
 
-func equal(oldOfs map[cost.OfferingKey]float64, newOfs map[cost.OfferingKey]float64) bool {
-	if len(lo.Values(oldOfs)) != len(newOfs) {
-		return false
+func hashOfferings(instanceTypes []*cloudprovider.InstanceType) uint64 {
```
DerekFrank: This is really just recreating the instancetype generation logic I wanted to introduce with https://github.com/kubernetes-sigs/karpenter/pull/2866

We didn't take this approach initially because so many places want to recreate this cache invalidation logic , we should just do it in one place

ryan-mist: Dropped nodepool equality check between reconciles completely as `UpdateOfferings` is not too computationally expensive and only will happen once per hour. Odds are the hash will be changed anyway so we're already doing the work of `UpdateOfferings` most of the time.

DerekFrank (approved):
/lgtm

PR #3069, opened 2026-06-04
perf: controller-runtime enable unstructured cache

daimaxiaxie:
Fixes #N/A 

**Description**

During Karpenter's idle operation, tens of thousands of Get Node requests are sent to the Apiserver per minute (visible in audit logs), which imposes a huge burden on the Apiserver.

However, I believe this request doesn't necessarily have to access the Apiserver, as the data obtained via the informer's watch is fresh enough.

The following is the path where the issue occurs:
```
When registering status.NewGenericObjectController[*corev1.Node]
For(&corev1.Node{}) → watches all Nodes in the cluster

In Reconcile, uses UnstructuredAdapter to call kubeClient.Get()
↓
UnstructuredAdapter embeds unstructured.Unstructured → implements runtime.Unstructured interface
↓
controller-runtime cached client checks shouldBypassCache():
cacheUnstructured defaults to false + object is runtime.Unstructured → bypasses cache
↓
Direct HTTP GET /api/v1/nodes/ to apiserver → visible in audit log
↓
At the end of Reconcile, returns RequeueAfter: 10s → reconciles again after 10 seconds
```

Startup is normal, and the get operations in the audit logs decrease.

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -186,6 +186,11 @@ func NewOperator(o ...option.Function[Options]) (context.Context, *Operator) {
 			ctx = injection.WithOptionsOrDie(ctx, options.Injectables...)
 			return ctx
 		},
+		Client: client.Options{
+			Cache: &client.CacheOptions{
+				Unstructured: true,
+			},
+		},
 		Cache: cache.Options{
 			ByObject: map[client.Object]cache.ByObject{
 				&coordinationv1.Lease{}: {


GnatorX:
Although I agree with the issue I don't think this is the fix. I am not sure what the full side effect of this is.. We should fix the library Karpenter is using for NewGenericObjectController instead if possible https://github.com/awslabs/operatorpkg/blob/c49b4ca8d58d12f490c6c7a1e9edd1de5a8e052d/status/controller.go#L169

GnatorX:
testing https://github.com/awslabs/operatorpkg/pull/211

PR #3086, opened 2026-06-12
feat: support time zones in disruption budget schedules

### LINKED ISSUE #3074: Feature request: Support timezones in budgets
### Description

https://github.com/kubernetes-sigs/karpenter/issues/943 made the deliberate decision to enforce UTC on all budget schedules. This is fine (since you can calculate to your specific timezone), until daylight savings time comes into play (depending on your region).

In our use case, we want to avoid node consolidation during working hours, since not all our workload is completely stateless. Currently, we either have to write suboptimal schedules, or we have to edit the schedules twice a year.

There should be a way to explicitly set the timezone. The original issue suggests adding a `timezone` field to the budgets; for our use case we would be happy if we could set it on the controller directly or to explicitly opt into using the nodes timezone.

Due to the way budgets currently work _backwards_ to our use case (restricting disruptions instead of allowing them (ref: https://github.com/kubernetes-sigs/karpenter/issues/1939#issuecomment-2628533814)), this makes dealing with budgets even more complex, so we'd really like to see this.

gabrielvieirabra:
Fixes #3074

**Description**

Adds an optional `timeZone` field to NodePool disruption budgets, following the upstream CronJob `spec.timeZone` precedent:

```yaml
disruption:
  budgets:
    - nodes: "0"
      schedule: "0 8 * * mon-fri"
      timeZone: "America/New_York"
      duration: 10h
```

- The budget `schedule` is interpreted in the configured IANA time zone; when unset it defaults to UTC, so existing budgets are unaffected (`TZ=UTC` was already prepended before this change).
- DST transitions follow the upstream CronJob behavior: a schedule hit that falls in a skipped hour (spring forward) does not occur, and a hit that falls in a repeated hour (fall back) occurs at both offsets. This is the native behavior of `robfig/cron`'s `TZ=` support and is covered by tests at both transition boundaries.
- CEL validation rejects `timeZone` without `schedule` (mirroring the existing `schedule`/`duration` rule). Since IANA names can't be validated by CEL, invalid time zones fail closed at evaluation time (0 allowed disruptions) and surface through the NodePool `ValidationSucceeded` status condition via `RuntimeValidate`.
- The time zone is validated with `time.LoadLocation` *before* being interpolated into the cron spec, so a malformed value can't be misinterpreted as part of the schedule.
- The IANA database is embedded via `time/tzdata` so resolution is deterministic on images that don't ship a system tzdata (e.g. distroless).

- `make presubmit`
- New unit tests: UTC default unchanged, schedule interpreted in the configured zone, spring-forward gap (2024-03-10 America/New_York → no hit that day), fall-back overlap (2024-11-03 → active at both 05:30Z and 06:30Z, inactive between), invalid time zone fails closed
- New CEL tests: accepts `timeZone` with schedule+duration, rejects `timeZone` without schedule, rejects empty `timeZone`
- New nodepool validation controller tests: invalid `timeZone` → `ValidationSucceeded=False`


AlexanderYastrebov:
For cross reference - there is another attempt https://github.com/kubernetes-sigs/karpenter/pull/2032

PR #3091, opened 2026-06-15
chore: add managed dimension to node and pod metrics

AndrewMitchell25:
Fixes #N/A

**Description**

Karpenter's node and pod metrics are emitted for every node and pod, including those not managed by Karpenter. When these metrics are aggregated downstream (e.g. computing cluster resource utilization as `sum(pod_requests) / sum(allocatable)`), the non-Karpenter nodes and pods skew the numbers, and there's no way to scope the result to just the capacity Karpenter manages.

This change adds a `managed` label to the node and pod metrics so operators can filter them to Karpenter-managed capacity:

- The per-node metrics: `karpenter_nodes_allocatable`, `karpenter_nodes_total_pod_requests`, `karpenter_nodes_total_pod_limits`, `karpenter_nodes_total_daemon_requests`, `karpenter_nodes_total_daemon_limits`, `karpenter_nodes_system_overhead`, and `karpenter_nodes_current_lifetime_seconds`.
- `karpenter_pods_state`.

For node metrics, a node is considered managed when it has an associated NodeClaim (`StateNode.Managed()`). For pod metrics, a pod is considered managed when it is scheduled onto a node that carries the nodepool label, consistent with how cluster state determines node management.

These are all additive raw-value metrics, so adding the label is backwards compatible: existing queries that don't reference `managed` continue to work unchanged, and the un-labeled total is still recoverable by aggregating across the label (e.g. `sum without (managed) (...)`). The `managed` label is a boolean, so it adds minimal cardinality.

`make test` for the node and pod metrics suites. Added assertions covering both the `managed="true"` and `managed="false"` cases for the per-node and pod state metrics.

### DIFF
--- a/pkg/controllers/metrics/node/controller.go
+++ b/pkg/controllers/metrics/node/controller.go
@@ -140,7 +140,7 @@ func initializeMetrics() {
 			Name:      "utilization_percent",
 			Help:      "Utilization of allocatable resources by pod requests",
 		},
-		[]string{metrics.ResourceTypeLabel, managed},
+		[]string{metrics.ResourceTypeLabel},
 	)
 }
 
@@ -207,19 +207,7 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 }
 
 func buildClusterUtilizationMetric(nodes state.StateNodes) []*metrics.StoreMetric {
-	// Partition nodes by whether they are managed by this Karpenter instance (i.e. have a
-	// NodeClaim owned by the configured cloud provider). Emitting utilization with a "managed"
-	// dimension lets dashboards filter out nodes Karpenter doesn't manage, which would otherwise
-	// skew the cluster utilization numbers.
-	managedNodes, unmanagedNodes := lo.FilterReject(nodes, func(n *state.StateNode, _ int) bool {
-		return n.Managed()
-	})
 
-	res := buildUtilizationForNodes(managedNodes, true)
-	return append(res, buildUtilizationForNodes(unmanagedNodes, false)...)
-}
-
-func buildUtilizationForNodes(nodes state.StateNodes, isManaged bool) []*metrics.StoreMetric {
 	// Aggregate resources allocated/utilized for all the nodes and pods inside the nodes
 	allocatableAggregate, utilizedAggregate := corev1.ResourceList{}, corev1.ResourceList{}
 
@@ -248,10 +236,7 @@ func buildUtilizationForNodes(nodes state.StateNodes, isManaged bool) []*metrics
 		res = append(res, &metrics.StoreMetric{
 			GaugeMetric: ClusterUtilization,
 			Value:       utilizationPercentage,
-			Labels: map[string]string{
-				metrics.ResourceTypeLabel: resourceNameToString(resourceName),
-				managed:                   strconv.FormatBool(isManaged),
-			},
+			Labels:      map[string]string{metrics.ResourceTypeLabel: resourceNameToString(resourceName)},
 		})
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/controllers/metrics/node/controller.go
```diff
@@ -250,18 +269,22 @@ func buildMetrics(n *state.StateNode) (res []*metrics.StoreMetric) {
 		Allocatable:         n.Node.Status.Allocatable,
 	} {
 		for resourceName, quantity := range resourceList {
+			labels := getNodeLabelsWithResourceType(n.Node, resourceNameToString(resourceName))
+			labels[managed] = isManaged
```
DerekFrank: Why not put this into the getNodeLabels*?

pkg/controllers/metrics/node/controller.go
```diff
@@ -204,7 +207,19 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
 }
 
 func buildClusterUtilizationMetric(nodes state.StateNodes) []*metrics.StoreMetric {
+	// Partition nodes by whether they are managed by this Karpenter instance (i.e. have a
+	// NodeClaim owned by the configured cloud provider). Emitting utilization with a "managed"
+	// dimension lets dashboards filter out nodes Karpenter doesn't manage, which would otherwise
```
DerekFrank: We don't need this expansive of a comment, we probably don't even need one, its just context rot imo

pkg/controllers/metrics/node/controller.go
```diff
@@ -204,7 +207,19 @@ func (c *Controller) Register(_ context.Context, m manager.Manager) error {
...
+	// skew the cluster utilization numbers.
+	managedNodes, unmanagedNodes := lo.FilterReject(nodes, func(n *state.StateNode, _ int) bool {
+		return n.Managed()
+	})
+
+	res := buildUtilizationForNodes(managedNodes, true)
+	return append(res, buildUtilizationForNodes(unmanagedNodes, false)...)
```
DerekFrank: I don't know if I love that you've broken it out into a separate function, we don't treat any of the other labels that way.

A more generic version of this might be:

```
	func buildUtilizationForNodes(nodes state.StateNodes, additionalLabels map[string]string) []*metrics.StoreMetric {
```

qq. What does the dimensionless metric look like with this count? Like can we still see managed+unmanaged in the /metrics endpoint?

pkg/controllers/metrics/node/controller.go
AndrewMitchell25: It seemed cleaner than just passing the isManaged bool down to those functions, but I'll update it if you think that's better

pkg/controllers/metrics/node/controller.go
AndrewMitchell25: Will remove

pkg/controllers/metrics/node/controller.go
AndrewMitchell25: I don't think so, would it be better to just remove this managed label for the cluster utilization metric?

pkg/controllers/metrics/node/controller.go
AndrewMitchell25: Actually maybe we should just pass the state node and then call .Managed in getNodeLabels. Does that work?

DerekFrank: Yea, I like that better

pkg/controllers/metrics/node/controller.go
DerekFrank: Its an interesting metric, but we can't make the breaking change to the metrics, we need the combined metric to stick around if we're going to add a new dimension

DerekFrank: Maybe prometheus scrapers just handle that

PR #3092, opened 2026-06-16
fix: skip deviceallocation tests when resource.k8s.io/v1 unavailable

jamesmthompson:
## Summary

- The `deviceallocation` test suite fails in envtest environments where `resource.k8s.io/v1` isn't served. The existing version check (`Minor() < 34`) is insufficient because envtest binaries at 1.35+ don't necessarily serve DRA APIs without additional runtime components.
- Adds a discovery check that skips the suite when the API group is actually unavailable, matching the pattern used by the DRA e2e tests.

## Test plan

- [x] `go test ./pkg/controllers/dynamicresources/deviceallocation/` passes locally (25 skipped via discovery check)
- [ ] CI with full DRA support runs tests normally (version + discovery both pass)

cc @jmdeal @ryan-mist

### DIFF
--- a/pkg/controllers/dynamicresources/deviceallocation/suite_test.go
+++ b/pkg/controllers/dynamicresources/deviceallocation/suite_test.go
@@ -57,6 +57,12 @@ var _ = BeforeSuite(func() {
 	if env.Version.Minor() < 34 {
 		Skip("ResourceClaims are only available starting in K8s version >= 1.34.x")
 	}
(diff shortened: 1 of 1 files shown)


jamesmthompson:
Closing. The issue is a stale local envtest binary (1.29 at /usr/local/kubebuilder/bin), not a test bug. CI passes because it installs the correct version. The right fix is making the Makefile set up KUBEBUILDER_ASSETS the same way CI does, not silently skipping tests.

Issue #2352, opened 2025-07-07
Support pods owned by custom Daemonset-like controllers


matskiv:
### Description

**What problem are you trying to solve?**
When making scaling decisions Karpenter treats DaemonSet pods differently from other types of pods. I would like to add support for scenarios where a pod is not owned by a DaemonSet controller, but should be treated as if it was.
There are two scenarios where this would be useful:
- DaemonSet pods scheduled by [vCluster](https://github.com/loft-sh/vcluster/) (popular k8s distro that uses a k8s cluster for infrastructure) will not have DaemonSet in the owner reference
- Custom controller that behaves similar to DaemonSet controller

The kubernetes/autoscaler handles these scenarios with the `cluster-autoscaler.kubernetes.io/daemonset-pod` annotation ([source](https://github.com/kubernetes/autoscaler/blob/563f074dd1244de9cb8c3c7997a095775244c985/cluster-autoscaler/utils/pod/pod.go#L31-L42)). Perhaps karpenter could just support the same annotation.

**How important is this feature to you?**
High

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

matskiv:
I'm happy to contribute a PR for this once it is triaged.

jmdeal:
Could you include an example of one of these DaemonSet like pods? I'd be curious why this needed to be handled by a custom controller and couldn't be a DaemonSet in vCluster, but I'm not very familiar with that project. This doesn't seem too unreasonable, but there's also two parts of the problem. It's not enough for Karpenter to be able to identify existing pods as belonging to a daemonset, but Karpenter also needs to be able to identify what daemonsets will schedule to each NodePool. Failing to do so could result in scheduling failures since Karpenter could provision nodes with insufficient capacity for both the daemonset like pod and the workload.

daimaxiaxie:
> Could you include an example of one of these DaemonSet like pods?

OpenKruise has a custom controller that enhances DaemonSet. https://openkruise.io/docs/user-manuals/advanceddaemonset

rschalo:
@matskiv following up for other use cases here and the two parts of the problem pointed out by @jmdeal

[...]

PR #3100, opened 2026-06-18
docs: add system workloads RFC

dm3ch:
Fixes #N/A
Refs #2352

**Description**

Adds an RFC draft for supporting daemon-like / node-associated system workloads that are not owned by native DaemonSets. The proposal introduces a feature-gated `SystemWorkload` CRD for classification, covers provider-installed rules, and explicitly defers template-based pre-provisioning overhead modeling to future work.

Documentation-only change. Reviewed locally and compared against recent RFC structure in `designs/`.

### DIFF
--- a/designs/system-workloads.md
+++ b/designs/system-workloads.md
@@ -1,0 +1,404 @@
+# System Workload Classification for DaemonSet-like Pods
+
+> Draft RFC for [kubernetes-sigs/karpenter#2352](https://github.com/kubernetes-sigs/karpenter/issues/2352). The final name for this concept is intentionally unresolved; this draft uses `SystemWorkload` as a placeholder.
+
+## Summary
+
+Karpenter currently treats native DaemonSet pods as node-associated overhead rather than normal workload demand. This proposal introduces `SystemWorkload` as the common model for describing additional pods that should receive similar treatment, and starts with a feature-gated, cluster-scoped CRD as the access mechanism for users and providers to supply that information.
+
+The first iteration is intentionally classification-focused. It does not introduce synthetic pod templates or generic custom-controller template extraction. Karpenter will not infer pre-provisioning overhead for never-observed `SystemWorkload` pods in this version.
+
+## Motivation
+
+Karpenter treats DaemonSet pods differently from normal workload pods in several paths:
+
+1. Pending DaemonSet pods are not treated as provisionable workload demand.
+2. Bound DaemonSet pods are not treated as reschedulable workload pods during disruption simulations.
+3. Observed DaemonSet pod requests and limits are tracked separately as daemon overhead on `StateNode`s.
+4. During provisioning, Karpenter lists DaemonSets and passes representative DaemonSet pods to the scheduler so daemon overhead is considered when choosing new capacity.
+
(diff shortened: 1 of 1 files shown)


dm3ch:
@matskiv @daimaxiaxie @Code2Life Inviting to review and comment this proposal, as it proposing possible direction for supporting "Support pods owned by custom Daemonset-like controllers"

designs/system-workloads.md
```diff
@@ -0,0 +1,424 @@
...
+
+```yaml
+apiVersion: karpenter.sh/v1alpha1
+kind: SystemWorkload
+metadata:
+  name: cluster-autoscaler-daemonset-pod-annotation
+spec:
+  annotationSelector:
+    matchExpressions:
+    - key: cluster-autoscaler.kubernetes.io/daemonset-pod
```
matskiv, line 129: This covers vCluster case 👍

designs/system-workloads.md
```diff
@@ -0,0 +1,424 @@
...
+
+1. Full DaemonSet parity in the first iteration.
+2. Modeling resource overhead for a system workload before any matching pod has ever existed, unless Karpenter can derive that information from an existing built-in source such as a native DaemonSet. Optional phase-two models may cover proportional workloads such as node/core-linear or ladder-scaled provider agents.
+3. Generic extraction of pod templates from arbitrary custom controller schemas.
+4. Provider-specific Go hooks in the initial API. Provider extension mechanisms may be considered later if declarative CRs are insufficient.
+5. Replacing native DaemonSet handling.
+
+## Proposal
+
+Introduce a feature-gated, cluster-scoped `SystemWorkload` CRD that classifies pods as system/node-associated workloads rather than normal application workload demand.
```
DerekFrank, line 50: How much of this can be solved by a `karpenter.sh/is-daemon` annotation? Is the issue there that most users don't own these non-daemon daemons?

DerekFrank: CAS already has this, if we want we can get sig-autoscaling to introduce a shared annotation and use that. 

https://github.com/kubernetes/autoscaler/blob/4cae5199c77395df1e2cad273a03fd74d66156a6/cluster-autoscaler/utils/pod/pod.go#L36

cc: @towca

dm3ch: So this can be solved by karpenter annotation, but I see following downsides of such approach:
1) In some cases users are not owning this workloads, for example in case GKE managed workloads, so injecting this annotation can be problematic due to that.
2) Also, in the future we might want to have non-DS pods accounted during scheduling simulations to avoid consolidations (thinking Pod would disappear after the node deprovision, but in realiry this pod would become Pending and result in provision of a new node). For example in case of GKE owned ladder autoscaled workloads in might be needed/cruccial. And CRD would possibly allow to extend it to provide autoscaling details for a proper scheduling simulations

DerekFrank: I don't know if I buy that argument. Right now Karpenter assumes all pods will be recreated when we perform consolidation, and unless upstream creates an API that lets a pod say "I'll be recreated|I won't be recreated", I don't think we need to index on making a CRD just in case Karpenter changes that very fundamental assumption. 

My vote is still to standardize the annotation in sig-autoscaling, then GKE needs to inject the annotation to be compliant with k8s autoscalers generally. Open to hearing other perspectives!

DerekFrank (changes requested):

PR #3102, opened 2026-06-18
fix: use namespaced name as resourceclaim index

jmdeal:
Fixes #N/A 

**Description**

Updates Karpenter's DRA allocator to use namespaced names for ResourceClaim indexes. This is a miss in the original implementation—since ResourceClaims are namespaced resources this should have always been indexed this way.

New regression test + `make presubmit`


ryan-mist (approved):
/lgtm

PR #3106, opened 2026-06-23
fix: expiration respects NodePool disruption budgets

### LINKED ISSUE #1750: v1: Bring back a "soft expiration" mechanism
### Description

`expireAfter` should respect disruption budgets in v1, like it was in 0.37

6 out of 10.
See: https://github.com/aws/karpenter-provider-aws/issues/7122

derlopes-amzn:
Fixes #1750

### Description

In v1, the nodeclaim expiration controller (`pkg/controllers/nodeclaim/expiration`) forcefully deletes an expired NodeClaim as soon as its `expireAfter` elapses, without consulting the owning NodePool's disruption budgets. Disruption budgets are only enforced by the disruption controller (drift / emptiness / consolidation); expiration bypasses them entirely.

As a result, a set of nodes that expire around the same time can all be terminated simultaneously, ignoring the availability protection that disruption budgets are meant to provide. This is a regression from the v0.37 "soft expiration" behavior described in #1750.

This change gates expiration on the NodePool's available disruption budget before deleting an expired NodeClaim:

- The expiration controller now receives the cluster state and event recorder, and calls `disruption.BuildDisruptionBudgetMapping` to compute how many more nodes may be disrupted in the NodePool right now.
- If no budget is available, the NodeClaim is requeued (30s) instead of being deleted, so it expires as soon as budget frees up.
- If budget is available, behavior is unchanged — the NodeClaim is deleted and drains as before (PDBs are still honored during the drain; `terminationGracePeriod` remains the upper bound).

The budget is looked up under the `Drifted` reason, since there is no dedicated `Expired` `DisruptionReason`. A budget with no `reasons` set — the common case — applies to all reasons regardless, so it gates expiration as expected. Happy to add a dedicated `Expired` reason (with the corresponding CRD enum change) if maintainers prefer that approach.

`make test` for the expiration controller (`go test ./pkg/controllers/nodeclaim/expiration/...`), all green. Added cases:

- An expired NodeClaim is **not** deleted when the NodePool budget is `0`, and the reconcile requeues.
- An expired NodeClaim **is** deleted once the budget allows it (`nodes: "1"`).
- A budget scoped to a different reason (`Empty`) does **not** block expiration (verifies the `Drifted` scoping).

Existing expiration tests were updated to register nodes in cluster state (via `ExpectMakeNodesAndNodeClaimsInitializedAndStateUpdated`) so the budget is computed against a known node count; with the default (no-budget) NodePool, expiration proceeds exactly as before.

I also manually tested this change against an EKS cluster with expiry set to 15 minutes with a disruption budget of 1. I provisioned 5 nodes at the same time and validated that only 1 was terminated at a time. I increased expiry to 24 hours and let it bake through other development work and confirmed only 1 node ever terminated at the same time.

### DIFF
--- a/pkg/controllers/controllers.go
+++ b/pkg/controllers/controllers.go
@@ -105,7 +105,7 @@ func NewControllers(
 		provisioning.NewPodController(kubeClient, p, cluster),
 		provisioning.NewNodeController(kubeClient, p),
 		nodepoolhash.NewController(kubeClient, cloudProvider),
-		expiration.NewController(clock, kubeClient, cloudProvider),
+		expiration.NewController(clock, kubeClient, cloudProvider, cluster, recorder),
 		informer.NewDaemonSetController(kubeClient, cluster),
 		informer.NewNodeController(kubeClient, cluster),
 		informer.NewPodController(kubeClient, cluster),
--- a/pkg/controllers/nodeclaim/expiration/controller.go
+++ b/pkg/controllers/nodeclaim/expiration/controller.go
@@ -18,6 +18,7 @@ package expiration
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -34,6 +35,9 @@ import (
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/cloudprovider"
+	"sigs.k8s.io/karpenter/pkg/controllers/disruption"
+	"sigs.k8s.io/karpenter/pkg/controllers/state"
+	"sigs.k8s.io/karpenter/pkg/events"
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
 )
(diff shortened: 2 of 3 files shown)


jmdeal:
Expiration does not respect disruption budgets by design. This gives cluster admins a knob, when combined with TerminationGracePeriod, to support strict maximum node lifetimes for compliance purposes. We can't introduce disruption budgets directly to the existing expiration controller — this would be a breaking change.

If you're still interested in this feature, the next step would be to write an RFC that either introduces a new soft-expiration mechanism **or** meets your desired use-case without breaking the use-case above. If the goal is to re-implement soft-expiration from pre-v1 Karpenter, the mechanism is unambiguous but the RFC should highlight the use-cases and why they aren't supported by Karpenter's existing disruption mechanisms (e.g. drift).

> ignoring the availability protection that disruption budgets are meant to provide

A note on this: this sounds like a symptom of PDB misconfiguration more than anything else. Node disruption budgets are **not** an availability mechanism, pod disruption budgets are. Karpenter will still respect PDBs during the drain process for expired nodes. The only thing forceful expiration changes is when drain is initiated, not how it proceeds.

derlopes-amzn:
That's understandable, I will iterate on this based on your feedback.

PR #3108, opened 2026-06-23
feat: Add nodepool status for nodeclaim conditions

### LINKED ISSUE #3071: Surface NodeClaim drift/rollout progress in NodePool status
### Description

Today, drift is only observable per-NodeClaim, via the `Drifted` status condition. There is no aggregate, NodePool-level signal that tells you how far a NodePool is through reconciling drift across the nodes it owns.

This makes it hard for external systems to gate on "this NodePool has finished rolling out a change" without listing and aggregating NodeClaims themselves. Concretely, we run Karpenter under Argo CD (GitOps). When a NodePool/EC2NodeClass change triggers drift, we want Argo CD to report the Application as `Progressing` until the drift-driven node replacement is substantially complete, and `Healthy` once it is.

Argo CD's health is evaluated **per resource** with a sandboxed Lua check that has no access to other resources. So a NodePool health check can only read the NodePool's own `status`. A NodeClaim health check could in principle read each NodeClaim's `Drifted` condition, but:

- It would require NodeClaims to appear as children of the NodePool in Argo's resource tree (which, in our multi-cluster setup, they currently do not), and
- It forces all-or-nothing (100%) semantics, with no notion of a tolerance threshold or a settling window.

As a result we maintain a PostSync Job that polls NodeClaims, groups them by `karpenter.sh/nodepool`, computes the percentage no longer `Drifted`, and blocks until each NodePool crosses a threshold (e.g. 90%). This is exactly the kind of aggregation we'd expect Karpenter itself to be able to expose, since it already owns the NodeClaims and tracks their drift state.

**Proposal**

Expose drift/rollout progress at the NodePool level, in `NodePool.status`. Any of the following would be sufficient for our use case (in rough order of preference):

1. Counts in `NodePool.status`, e.g. `status.driftedNodeClaims` / `status.nodeClaims` (analogous to the existing `status.nodes` and `status.resources`), so consumers can compute completion percentage directly.
2. A NodePool-level condition such as `Drifted` (`status: "True"` while any owned NodeClaim is drifted, `"False"` once reconciliation is complete), mirroring how the per-NodeClaim `Drifted` condition works.
3. Both — a condition for a simple boolean gate plus counts for threshold-based logic.

This would let any GitOps/automation system gate on a NodePool's own `status` without re-implementing NodeClaim aggregation, and it would make rollout progress visible in `kubectl get nodepool` and dashboards.

Moderately important. We have a working PostSync Job that does the aggregation today, so we are not blocked, but it adds operational surface (a Job + ServiceAccount + ClusterRole + a maintained image per cluster) purely to compute information Karpenter already has internally. A NodePool-level status field/condition would let us replace that machinery with a standard Argo CD custom health check and would benefit anyone integrating Karpenter rollouts with external orchestration.

cnmcavoy:
Fixes #3071

**Description**
Adds a status field in the `NodePool` CRD for `nodeClaimConditions` - A list of conditions and the count of nodeclaims with that condition type. This field is kept in sync by the nodepool counter controller, which refreshes the conditions from the nodeclaim resources attached to the nodepool.

Tested in a live cluster, unit tests are also provided. Passed `make presubmit` as well.


jigisha620:
@cnmcavoy I asked this in the original issue as well but what do you think about exposing this information via a metric instead? Not asking to make that change in this PR right away but just curious.

cnmcavoy:
> @cnmcavoy I asked this in the original issue as well but what do you think about exposing this information via a metric instead? Not asking to make that change in this PR right away but just curious.

I think a metric could also be useful. However, Our use-case for this field in the status is ArgoCD, so that it can understand the state of a NodePool resource and accurately report it to the Application which owns the resource. ArgoCD supports custom health information for resource types, like NodePools.

We have an ArgoCD Application for each cluster we manage NodePools, and want to only sync Applications after prior ones have completed processing Drift. So with this status information, we can configure ArgoCD with a custom NodePool healthcheck such that it reports the application is "Progressing" rather than "Healthy" when processing Drift.

DerekFrank:
We definitely should integrate with ArgoCD well. Lets discuss this briefly with the ArgoCD project to see if they want to become metrics aware or if we should surface this condition. https://github.com/kubernetes-sigs/karpenter/issues/3071

pkg/controllers/state/cluster.go
```diff
@@ -623,6 +624,33 @@ func (c *Cluster) NodePoolResourcesFor(nodePoolName string) corev1.ResourceList
 	return maps.Clone(c.nodePoolResources[nodePoolName])
 }
 
+func (c *Cluster) NodePoolNodeClaimConditionsFor(nodePoolName string) []v1.NodeClaimConditions {
```
DerekFrank, line 627: Is it more efficient to pre-compute this every UpdateNode?

cnmcavoy: My suspicion is no, the values here are only recomputed at most every 5s. If the values were used for something other than the `status` I'd be more concerned about the performance cost.

My feelings aren't very strong about this though, so if you prefer it be precomputed, I can do that change.

cnmcavoy:
@jigisha620 I have a separate PR here for metrics: https://github.com/kubernetes-sigs/karpenter/pull/3177 - ideally both would get merged, I think having the nodeclaim conditions on the status is useful for argocd or other tooling that is parsing the k8s resources directly, while the prometheus metrics is useful for looking at the same data over a time window.

jigisha620:
> @jigisha620 I have a separate PR here for metrics: #3177 - ideally both would get merged, I think having the nodeclaim conditions on the status is useful for argocd or other tooling that is parsing the k8s resources directly, while the prometheus metrics is useful for looking at the same data over a time window.

Sure, will take a look at that one as well.

jmdeal:
Cross-posting my comment from the issue, would love your input there @cnmcavoy 

https://github.com/kubernetes-sigs/karpenter/issues/3071#issuecomment-5170006562

PR #3109, opened 2026-06-23
feat: add offering overrides and multi-allocatable scheduling support

jigisha620:
Fixes #N/A 

**Description**

Adds support for per-offering resource overrides, enabling cloud providers to model instance type variants that have different capacity and 
overhead characteristics from the same base instance type.

Changes made -
1. Offering struct - Add CapacityOverride (ResourceList) and OverheadOverride (*InstanceTypeOverhead) fields. Values are merged with the instance type's base capacity/overhead — new keys are added, existing keys are replaced.
2. InstanceType.Allocatables() - New method returning distinct allocatable resource sets, one per unique (CapacityOverride, OverheadOverride) combination across offerings. The first entry is always the base allocatable.
3. fits() in scheduling - Updated to iterate Allocatables() instead of checking a single Allocatable(). A pod fits if any allocatable satisfies its resource requests.

Added unit tests.


pkg/cloudprovider/types.go
```diff
@@ -138,6 +138,7 @@ type InstanceType struct {
 	Overhead               *InstanceTypeOverhead
 	once                   sync.Once
 	allocatable            corev1.ResourceList
```
jmdeal: Shouldn't allocatable be dropped in favor of allocatables?

pkg/cloudprovider/types.go
```diff
@@ -138,6 +138,7 @@ type InstanceType struct {
...
+	allocatables           []corev1.ResourceList
```
jmdeal: Each allocatable is associated with a fixed set of offerings. How, at the scheduling layer, do we ensure that we only check allocatables for offering that are still available? Additionally, how do we ensure if a workload is only compatible with a subset of allocatables it's also compatible with the offerings that produce that subset? I think we need to track this mapping.

pkg/cloudprovider/types.go
jigisha620: Good call, will drop the private allocatable field and have `Allocatable()` return `allocatables[0]`. Keeping `Allocatable()` itself since it's used throughout the codebase and most callers just need the base allocatable without caring about overrides. What do you think?

pkg/cloudprovider/types.go
jigisha620: You're right that there's no structural link ensuring the allocatable that satisfied `fits()` has offerings that are both available and compatible with the pod's requirements.

I think this is safe today because override offerings should be cloned 1:1 from base offerings (same zones, same capacity types, same availability). But this relies on resolvers always producing overrides for the full set of base offerings. Do we expect resolvers that selectively produce override offerings for only a subset of capacity types or zones? If so, we'd need to map each allocatable back to its producing offerings so fits() can verify at least one backing offering is available and compatible. What do you think?

jmdeal: We should address this structurally — this isn't an invariant that's enforced by the type and it's easy to conceptualize use-cases where this won't hold. A trivial example would be an instance type which can support a resource but only in a certain zone. I don't want to only consider our immediate use-case for the AWS provider since this is a shared interface that all providers need to be able to reason about.

pkg/controllers/provisioning/scheduling/nodeclaim.go
```diff
@@ -483,6 +473,18 @@ func compatible(instanceType *cloudprovider.InstanceType, requirements schedulin
 	return instanceType.Requirements.Intersects(requirements) == nil
 }
 
-func fits(instanceType *cloudprovider.InstanceType, requests corev1.ResourceList) bool {
-	return resources.Fits(requests, instanceType.Allocatable())
+func fits(instanceType *cloudprovider.InstanceType, requests corev1.ResourceList, requirements scheduling.Requirements) (itFits bool, hasOffering bool) {
+	for _, group := range instanceType.AllocatableOfferingsList() {
+		resourceFit := resources.Fits(requests, group.Allocatable)
+		for _, of := range group.Offerings {
+			if of.Available && requirements.IsCompatible(of.Requirements, scheduling.AllowUndefinedWellKnownLabels) {
```
DerekFrank: Why include unavailable offerings in the list to begin with?

jigisha620: Good question! I think `precompute()` should be agnostic to availability, its job is purely structural grouping - "what distinct allocatable resource sets can this instance type produce?" Whether an offering is currently available doesn't change what allocatable it would produce, it just can't be used right now.

Availability filtering I think should live in `fits()`, since it's already making the scheduling decision and checking zone/capacity-type requirement compatibility on each offering anyway. Keeping the two concerns separate means precompute() shouldn't need to know about scheduling-time filtering criteria, and if we add new filtering dimensions in `fits()` later, `precompute()` stays untouched.

DerekFrank: I buy the 

> if we add new filtering dimensions in fits() later, precompute() stays untouched.

But we get that either way. I don't know if 

> precompute() should be agnostic to availability

is an invariant we care about. We already made it offering-aware, and availability is one of the core pieces of offerings

jigisha620: I don't feel very strongly about keeping `precompute()` agnostic to availability, updated to address this comment.

DerekFrank (approved):

PR #3110, opened 2026-06-24
feat: dra consumable capacity + partitionable devices support

ryan-mist:
Fixes #N/A 

**Description**
Extends support for DRA to partitionable devices and consumable capacity.

Individual PRs for each feature (if reviewing is easier that way):
- https://github.com/ryan-mist/karpenter/pull/4
- https://github.com/ryan-mist/karpenter/pull/3

For reference, the upstream KEPs:
- [KEP-4815: DRA: Add support for partitionable devices](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4815-dra-partitionable-devices)
- [KEP-5075: DRA Consumable Capacity](https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/5075-dra-consumable-capacity)

The design doc update will be added shortly. 

**Note**: The DistinctAttribute device constraint is introduced in the upstream consumable capacity KEP. This PR does not support DistinctAttribute **yet** and will be done as a fast follow-up.

- unit / integration testing; E2Es will be a follow up


pkg/cloudprovider/dynamicresources.go
```diff
@@ -70,6 +71,15 @@ type Device struct {
...
 	ConsumesCounters []resourcev1.DeviceCounterConsumption
+	// NodeName identifies the node where the device is available.
+	// Only set when the slice uses PerDeviceNodeSelection.
+	NodeName *string
+	// NodeSelector defines the nodes where the device is available.
+	// Only set when the slice uses PerDeviceNodeSelection.
+	NodeSelector *corev1.NodeSelector
+	// AllNodes indicates that all nodes have access to the device.
+	// Only set when the slice uses PerDeviceNodeSelection.
+	AllNodes *bool
```
jmdeal: Structurally , Karpenter's allocator can currently only support node local devices. How are these fields being supported? I think this should likely be dropeed — I don't know if this PR addresses the non-node local device provisioning question for in-flight devices but that poses significant ambiguity that I would not want to scope into this PR.

pkg/cloudprovider/dynamicresources.go
jmdeal, line 52: Can we add some additional documentation here to call out that, if SharedCounters is defined on a slice template, all devices in the slice template are ignored? Currently this fails open. I believe this is the right behavior, but we should provide cloudprovider implementers that are less familiar with DRA some indication the shared counters must be stored in a stand-alone template.

pkg/scheduling/dynamicresources/pool.go
```diff
@@ -53,6 +54,13 @@ type Pool struct {
 	Slices []ResourceSlice
 	// Devices is the flattened list of devices across all slices in this pool.
 	Devices []DeviceWithID
+	// NonTargetingDevices holds devices from slices that don't match the
+	// requirements but have ConsumesCounters. These are not allocation
+	// candidates and solely exist so the allocator can deduct their counter
+	// consumption when they are allocated on other Nodes.
+	NonTargetingDevices []DeviceWithID
```
jmdeal, line 61: I wonder if shared counter consumption should be implicitly tracked by the allocation tracker controller in the same way as consumed capacity? In some ways it's outside of the allocation tracker controller's current scope because the counters are present on the slice rather than the claim, but I think it's consistent with the purpose of the tracker. I think this also better distributes the complexity across the codebase.

If you agree, I think this can be a follow-up. I'm not concerned about this from a correctness standpoint.

pkg/scheduling/dynamicresources/allocationtracker.go
jmdeal, line 227: From an architecture standpoint, I think it makes sense to track all allocation state, including counters and capacity here. The intention was to separate the tree traversal logic from the complex logic that governs if a device is allocatable. It may not be able to retain it's current signature since you would probably want to pass in the allocation state from the current DFS traversal as well to avoid duplicating checks, but that's acceptable. This wasn't done in the original implementation because the in-flight state was as a simple as a set existence lookup.

As far as I can tell at the moment, this is not a correctness issue so this could be a follow-up if you agree.

pkg/scheduling/dynamicresources/allocationtracker.go
```diff
@@ -174,6 +242,15 @@ func (at *AllocationTracker) IsAllocated(deviceID DeviceID, nodeClaim NodeClaim,
 		return false
 	}
 
+	// Multi-allocatable devices are never "fully allocated" from a binary standpoint.
```
jmdeal: These checks are never exercised because of a gate at the call site, right? That being said, if we want to include them to be defensive shouldn't these checks exist for both the template and non-template paths? This may not matter in the context of my other comment: https://github.com/kubernetes-sigs/karpenter/pull/3110/changes#r3469136208

pkg/scheduling/dynamicresources/pool.go
ryan-mist: I agree - it was something I had considered originally, but didn't feel too strongly about either way. 

The reason I went with this originally is to match the behavior upstream (https://github.com/kubernetes/dynamic-resource-allocation/blob/master/structured/internal/incubating/pools_incubating.go) as it was easier for me to ensure correctness since its structurally similar

Will move over to allocation tracker controller as follow-up

pkg/cloudprovider/dynamicresources.go
ryan-mist: Good point, I had added those fields and never followed up or thought about continued integration with this - will drop it

pkg/scheduling/dynamicresources/allocationtracker.go
ryan-mist: Yeah good point, I'll drop these checks for now with the intention to refactor for https://github.com/kubernetes-sigs/karpenter/pull/3110/changes#r3469136208

PR #3112, opened 2026-06-25
chore: make NodeClaim launch timeout configurable

AndrewMitchell25:
Fixes #N/A

**Description**

The NodeClaim launch timeout is currently hard-coded to 5 minutes (`launchTimeout = time.Minute * 5` in `pkg/controllers/nodeclaim/lifecycle/liveness.go`). When a NodeClaim fails to launch, scheduling simulations are blocked until this timeout expires and the NodeClaim is cleaned up, which can cause pending pods to breach pod scheduling/binding latency thresholds.

The expected launch latency is inherently cloud-provider-dependent, so rather than hard-coding (or globally lowering) the timeout, this change exposes it as an exported package variable that cloud providers can override at startup.

`make presubmit`, added two unit tests 

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/liveness.go
+++ b/pkg/controllers/nodeclaim/lifecycle/liveness.go
@@ -46,31 +46,17 @@ type Liveness struct {
 }
 
 // registrationTimeout is a heuristic time that we expect the node to register within
-// launchTimeout is a heuristic time that we expect to be able to launch within
 // If we don't see the node within this time, then we should delete the NodeClaim and try again
 
 const (
 	registrationTimeout       = time.Minute * 15
 	registrationTimeoutReason = "registration_timeout"
-	launchTimeout             = time.Minute * 5
 	launchTimeoutReason       = "launch_timeout"
 )
 
-type NodeClaimTimeout struct {
-	duration time.Duration
-	reason   string
-}
-
-var (
-	RegistrationTimeout = NodeClaimTimeout{
-		duration: registrationTimeout,
-		reason:   registrationTimeoutReason,
-	}
-	LaunchTimeout = NodeClaimTimeout{
-		duration: launchTimeout,
-		reason:   launchTimeoutReason,
-	}
-)
+// LaunchTimeout is a heuristic time that we expect to be able to launch within
+// If we don't launch within this time, then we should delete the NodeClaim and try again
+var LaunchTimeout = time.Minute * 5
 
 //nolint:gocyclo
 func (l *Liveness) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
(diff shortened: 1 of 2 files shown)


pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -57,18 +57,18 @@ const (
 )
 
 type NodeClaimTimeout struct {
-	duration time.Duration
-	reason   string
+	Duration time.Duration
+	Reason   string
```
jigisha620: nit: Does Reason need to be exported here?

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
```diff
@@ -184,6 +185,69 @@ var _ = Describe("Liveness", func() {
 		// expect that the nodeclaim was not deleted
 		ExpectExists(ctx, env.Client, nodeClaim)
 	})
+	It("should delete the NodeClaim past a custom launch timeout configured via the LaunchTimeout var", func() {
```
jigisha620: Can these two tests be combined into one? Step the clock 30s and verify the NodeClaim still exists, then step another 30s (past the 45s timeout) and verify it's deleted. Same coverage with less duplication.

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -57,18 +57,18 @@ const (
...
 }
 
 var (
 	RegistrationTimeout = NodeClaimTimeout{
-		duration: registrationTimeout,
-		reason:   registrationTimeoutReason,
+		Duration: registrationTimeout,
```
jigisha620: The exported RegistrationTimeout with this exported Duration field is a bit confusing, it's settable by cloud providers but the reconciler still reads the registrationTimeout const directly, so changing it wouldn't actually do anything. If only the launch timeout is intended to be configurable, could we avoid exporting RegistrationTimeout?

jigisha620: What do you think about simplifying to a plain exported var instead?

pkg/controllers/nodeclaim/lifecycle/liveness_test.go
AndrewMitchell25: Updated

pkg/controllers/nodeclaim/lifecycle/liveness.go
AndrewMitchell25: Makes sense, updated

jigisha620 (approved):
/lgtm

pkg/controllers/nodeclaim/lifecycle/liveness.go
```diff
@@ -46,31 +46,27 @@ type Liveness struct {
...
 	registrationTimeoutReason = "registration_timeout"
-	launchTimeout             = time.Minute * 5
 	launchTimeoutReason       = "launch_timeout"
 )
 
+// LaunchTimeout is a heuristic time that we expect to be able to launch within
+// If we don't launch within this time, then we should delete the NodeClaim and try again
+var LaunchTimeout = time.Minute * 5
+
 type NodeClaimTimeout struct {
```
jmdeal: nit: as a cleanup exercise, do we want to just inline this struct? We're only constructing one staticlly now and constructing the other dynamically. I also don't think there was ever a purpose to exporting this type — external provider's can't override it since the fields are still private. I would just inline the two fields directly into the `deleteNodeClaimForTimeout` signature.

AndrewMitchell25: Makes sense, updated

PR #3114, opened 2026-06-26
chore: make Buffer expectations provider/env agnostic

sumukha-radhakrishna:
**Description**

make Buffer expectations provider/env agnostic so we can use it in provider tests as well. 

make presubmit and CI


DerekFrank (approved):
/lgtm

PR #3116, opened 2026-06-27
fix: DRA counter tracking mutation race

ryan-mist:
Fixes #N/A 

**Description**
The Allocator struct is shared and, previously, two lazy-init functions mutate the shared AllocationTracker map during Allocate():
  1. InitRemainingCounters — writes to at.RemainingCounters when a pool with SharedCounters is first encountered during DFS
  2. InitTemplateRemainingCounters — writes to at.templateRemainingCounters when template counter budgets are computed per NodeClaim/IT

Fix for (1) is to pre-initialize all pool counter budgets eagerly in NewAllocator() before any parallelism begins. The call in tryDevice becomes a dead no-op (always hits early return). The fix for (2) is compute template counter totals locally during Allocate() (read-only on tracker), store them on the allocation result, and write them to the tracker during Commit() (which is always sequential).

existing unit tests


jmdeal:
Looks good, I'll follow-up with a test that exercises the concurrent add. We should have been able to catch this with some multi-NodePool tests in the provisioning suite.

/lgtm
/approve

PR #3120, opened 2026-06-30
fix: check for nodename and terminal pod for PodSchedulingUndecidedTimeSeconds gauge delete

joshuakguo:
Fixes #N/A 

**Description**

saw an issue with pod scheduling metrics where

1. kube-scheduler marks pod PodScheduled=False/Unschedulable
  2. Provisioning PodController ACKs the pod (sets podAcks[nn] = now)
  3. kube-scheduler binds the pod 2 seconds later via pods/binding
  4. Kubelet's status patch races with scheduler's condition update → PodScheduled condition never reaches True
  5. Metrics controller's first reconcile sees pod already in Failed phase (not Pending):
    - Never inserts into pendingPods → ClearPodSchedulingMappings (which clears podAcks) never runs → ackTime persists
    - Never inserts into unscheduledPods → unbound metric never set (explains its silence)
  6. recordPodSchedulingUndecidedMetric runs every 30s:
    - PodSchedulingDecisionTime = zero (provisioner never decided — pod was bound before it got to it)
    - scheduled = false (PodScheduled condition is still False/Unschedulable)
    - PodAckTime != zero (never cleared)
    - → Sets gauge to time.Since(ackTime) → grows monotonically for 24h
  7. Pod deleted → NotFound path finally clears the gauge


pkg/controllers/metrics/pod/suite_test.go
```diff
@@ -535,6 +535,50 @@ var _ = Describe("Pod Metrics", func() {
...
 	})
+	It("should delete provisioning undecided metrics when the pod has a NodeName but PodScheduled condition is not True", func() {
+		p := test.Pod(test.PodOptions{NodeName: "test-node"})
+		p.Status.Phase = corev1.PodPending
+		p.Status.Conditions = []corev1.PodCondition{{Type: corev1.PodScheduled, Status: corev1.ConditionFalse, Reason: "Unschedulable", LastTransitionTime: metav1.Now()}}
+		ExpectApplied(ctx, env.Client, p)
+
+		cluster.AckPods(p)
+		ExpectReconcileSucceeded(ctx, podController, client.ObjectKeyFromObject(p))
+		env.Clock.Step(1 * time.Hour)
```
jigisha620: Is this needed? We are not reconciling again after this. Or rather should the test first check that the metric is emitted when nodeName is not set and then once it's set and we reconcile again, the metric should get deleted?

pkg/controllers/metrics/pod/suite_test.go
```diff
@@ -535,6 +535,50 @@ var _ = Describe("Pod Metrics", func() {
...
+		Expect(found).To(BeFalse())
+	})
+	It("should delete provisioning undecided metrics when the pod is terminal without PodScheduled condition being True", func() {
+		p := test.Pod()
+		p.Status.Phase = corev1.PodPending
+		ExpectApplied(ctx, env.Client, p)
+
+		cluster.AckPods(p)
+		ExpectReconcileSucceeded(ctx, podController, client.ObjectKeyFromObject(p))
+		env.Clock.Step(1 * time.Hour)
```
jigisha620: Don't think it's needed here either, you can check though. I think the next reconcile should just work on the basis of pod's status condition.

AndrewSirenko (commented):
lgtm

jigisha620 (approved):
/lgtm

DerekFrank (approved):
/lgtm

PR #3122, opened 2026-06-30
fix: ResourceClaim hydration deadlock

ryan-mist:
Fixes #N/A 

**Description**
Provisioning with DRA enabled deadlocks until ResourceClaim CRUD operations (even when no ResourceClaims exist in the cluster). This change fixes that.

#### Currently
In provisioning, if DRA is enabled then we gather the in-cluster allocated devices 
https://github.com/kubernetes-sigs/karpenter/blob/d5e91d67dae32d8c14230c51495fb26f8b59e6ca/pkg/controllers/provisioning/provisioner.go#L334-L339. 
As part of this gathering, we wait for a hydration channel to close to ensure we have synced ResourceClaims in the cluster 
https://github.com/kubernetes-sigs/karpenter/blob/d5e91d67dae32d8c14230c51495fb26f8b59e6ca/pkg/controllers/provisioning/dra.go#L94-L95
https://github.com/kubernetes-sigs/karpenter/blob/d5e91d67dae32d8c14230c51495fb26f8b59e6ca/pkg/controllers/dynamicresources/deviceallocation/controller.go#L244-L255

This hydration channel is only closed after the device allocation controller is reconciled, which only will happen when there is a CRUD to a ResourceClaim object:

https://github.com/kubernetes-sigs/karpenter/blob/d5e91d67dae32d8c14230c51495fb26f8b59e6ca/pkg/controllers/dynamicresources/deviceallocation/controller.go#L120-L123

Without any CRUD to ResourceClaims, then we wait for this channel to close indefinitely.

#### This Change
The fix adds a calls `Hydrate()` after informer caches sync in controller registration, independent of any ResourceClaim reconcile event. `Hydrate()` is wrapped in `sync.Once` so it's safe to call from both the runnable and reconciles (whichever fires first closes the channel and unblocks `AllocatedDevices()`)

* unit tests

### DIFF
--- a/pkg/controllers/dynamicresources/deviceallocation/controller.go
+++ b/pkg/controllers/dynamicresources/deviceallocation/controller.go
@@ -118,9 +118,7 @@ func NewController(kubeClient client.Client) *Controller {
 }
 
 func (c *Controller) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
-	c.hydrationOnce.Do(func() {
-		c.Hydrate(ctx)
-	})
+	c.Hydrate(ctx)
 
 	claim := &resourcev1.ResourceClaim{}
 	if err := c.kubeClient.Get(ctx, req.NamespacedName, claim, client.UnsafeDisableDeepCopy); err != nil {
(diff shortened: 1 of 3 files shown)


DerekFrank (commented):
I think we should re-use the cluster.Synced() style structure instead of re-inventing the wheel here, but @jmdeal and @ryan-mist can take that conversation offline

/lgtm

DerekFrank (approved):

jmdeal:
> I think we should re-use the cluster.Synced() style structure

This serves a different purpose than the `cluster.Synced` mechanism. The purpose of the cluster synced mechanism is to ensure that every NodeClaim we've created (updated via an in-memory process) has had its launch details populated. This is to ensure that we've hydrated all of our data at least once — we're fine with subsequent updates being eventually consistent since updates to the informer cache are eventually consistent anyways. This is a pretty common pattern, the miss here is that an earlier revision invoked hydrate from the `AllocatedDevices` call but it looks like that got dropped in when it was changed to an iterator. @ryan-mist's solution looks good to me as well though.

PR #3126, opened 2026-07-06
fix: defer nodeclaim disruption transition logs until status patch succeeds

### LINKED ISSUE #2002: `marking consolidatable` log messages are duplicated due to read-after-write inconsistency

heyrobertchang:
Fixes #2002

**Description**

**Why this approach**

The stale read itself can't be fixed — that's just how informer caches work. So the question is what we can actually trust.

I went through the other options first:

- Sleeping after the patch (#2018): blocks a worker and the race is still there.
- Event filter on the `NodeClaim` watch: the racing trigger usually comes from the Pod watch, and we need those events, so there's nothing to filter.
- Checking `SetTrue`'s return value: the stale copy doesn't have the condition, so it returns `true` again. Same double log.

The one thing we can trust is the API server. The status patch already uses an optimistic lock, so a stale reconcile always gets a 409. That's a reliable "did my write land" signal, and we weren't using it.

**The idea**

Don't log when we mutate the object — log when the patch goes through. If the patch hits a 409, the buffered lines get thrown away. The log ends up describing what actually happened instead of what a reconcile thought it did.

- Regression test: reconcile a stale copy (old resourceVersion, no condition) right after a successful one, assert the log fires exactly once and the stale one requeues. It fails on `main` without the fix.
- Full `nodeclaim/disruption` suite passes with `-race`.

### DIFF
--- a/pkg/controllers/nodeclaim/disruption/consolidation_test.go
+++ b/pkg/controllers/nodeclaim/disruption/consolidation_test.go
@@ -17,12 +17,16 @@ limitations under the License.
 package disruption_test
 
 import (
+	"strings"
 	"time"
 
+	"github.com/go-logr/logr/funcr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/samber/lo"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	v1 "sigs.k8s.io/karpenter/pkg/apis/v1"
 	"sigs.k8s.io/karpenter/pkg/test"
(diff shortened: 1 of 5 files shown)


DerekFrank:
/assign @AndrewMitchell25 

can you take a look? TY

AndrewMitchell25:
👀

pkg/controllers/nodeclaim/disruption/consolidation_test.go
```diff
@@ -96,6 +100,30 @@ var _ = Describe("Underutilized", func() {
 		nodeClaim = ExpectExists(ctx, env.Client, nodeClaim)
 		Expect(nodeClaim.StatusConditions().Get(v1.ConditionTypeConsolidatable).IsTrue()).To(BeTrue())
 	})
+	It("should not log marking consolidatable again when reconciling against a stale NodeClaim", func() {
```
AndrewMitchell25, line 103: Should we add a similar test in drift_test.go? Other than that I think it LGTM

heyrobertchang: Hi @AndrewMitchell25 

I've added test case for `drift_test.go` and CI is passed.

Thank you!

AndrewMitchell25 (approved):
/lgtm

DerekFrank (commented):
Hey @heyrobertchang! We appreciate the contribution :D 

I talked it over with @jmdeal, and we agree that the complexity is not worth just deduping logs. The logs even say 'removing' not 'removed', indicating that karpenter is trying to take some action, not definitely completing it. We'd rather see a fix that targets the read-after-write consistency, but that is a much larger change and definitely not a /good-first-issue. 

I hope this doesn't dissuade you from more open source contributions!

/close

heyrobertchang:
@DerekFrank No worries! I totally get it, thanks

Issue #3080, opened 2026-06-09
Drift disruption can indefinitely starve NodePools under sustained replacement failures (head-of-line blocking by drift age)


javanthropus:
**What problem are you trying to solve?**

When many NodePools are drifted at once (e.g. a batched AMI/AMIFamily change), a single
NodePool whose replacements repeatedly fail or stay blocked can hold the head of the
disruption line **on every reconcile pass**, so younger NodePools are never reached. This is
not a slowdown that eventually resolves — it is indefinite **starvation** of other NodePools
for as long as the failing pool keeps failing.

*Observed on v1.12.x; the relevant logic is unchanged since v1.11 and on `main` at time of writing.*

**Current behavior** — `Drift.ComputeCommands` (`pkg/controllers/disruption/drift.go`):

1. Rebuilds the candidate set every pass.
2. Sorts candidates strictly by the `Drifted` condition `LastTransitionTime`, oldest first.
3. Walks the sorted list, runs `SimulateScheduling` per candidate, and returns the **first**
   schedulable candidate as a single command (`return []Command{cmd}, nil`).
4. The controller dispatches that command and requeues immediately (`RequeueImmediately`), so
   the loop runs many rapid passes during a rollout.

Failed replacements return the candidate to the pool unchanged. For example, on
InsufficientCapacity (ICE) the replacement NodeClaim is deleted and the node returns to the
candidate set **with its original `LastTransitionTime`** (`pkg/controllers/disruption/queue.go`,
`CompleteCommand`/`UnmarkForDeletion`).

**Why it starves** — because the candidate is restored with its original (old) drift timestamp,
and the list is rebuilt and re-sorted oldest-first every pass, the same failing candidate is
selected first again and again. Nothing about its candidacy has changed: it is still the
oldest, still within its NodePool's budget, and is no longer in the queue. As a result:

- A NodePool whose oldest candidates keep failing fast (ICE is the common trigger, but
  un-evictable pods / blocked scheduling produce the same shape) **monopolizes the single
  command** every pass.
- All younger NodePools sit behind it and are **never serviced**.
- Each pass also burns a `CreateFleet`/launch attempt on the doomed candidate, producing a
  retry storm against the cloud provider.

This is head-of-line blocking created by ordering disruption strictly by drift age across
NodePools.

**Reproduction**

1. Create several NodePools and drift all of them at once (e.g. change a shared AMI selector).
[...]

PR #3128, opened 2026-07-06
docs: add drift-per-nodepool-backoff rfc

cnmcavoy:
Related to #3080

**Description**

This RFC proposes a per-NodePool exponential back-off on unrecoverable drift replacement failures. While a NodePool is backed off, its candidates are skipped during drift candidate selection. After the back-off expires, exactly one candidate is allowed to "probe"; a successful replacement resets the back-off, and a failure grows it exponentially (capped).

Documentation-only change. Reviewed and compared to recent RFC structure in the designs dir.

### DIFF
--- a/designs/drift-per-nodepool-backoff.md
+++ b/designs/drift-per-nodepool-backoff.md
@@ -194,6 +194,7 @@ Introduce a **per-NodePool drift back-off tracker**. Three interaction points:
 type NodePoolBackoff struct {
 	mu    sync.Mutex
 	clock clock.Clock
+	rand  *rand.Rand // injectable so jitter is deterministic in tests
 	state map[string]*backoffEntry // keyed by NodePool name
 }
 
@@ -211,12 +212,14 @@ type backoffEntry struct {
 
 ### Back-off formula
 
-On each unrecoverable failure, grow the window exponentially and clamp it to a single
-absolute ceiling (`maxDelay`):
+On each unrecoverable failure, grow the window exponentially, clamp it to a single
+absolute ceiling (`maxDelay`), and apply randomized **jitter**:
 
 ```
-level := level + 1
-until := now + min(baseDelay * 2^(level-1), maxDelay)
+level  := level + 1
+window := min(baseDelay * 2^(level-1), maxDelay)
+window := window/2 + rand[0, window/2)   // equal jitter: keeps a floor of window/2
+until  := now + window
 ```
 
 Proposed defaults (mirroring the existing retry-duration bounds in `queue.go`, which
@@ -335,13 +338,13 @@ func (b *NodePoolBackoff) Fail(nodePool string) {
 	window := baseDelay << (e.level - 1) // baseDelay * 2^(level-1)
 	if window > maxDelay || window <= 0 {
 		window = maxDelay
+		e.level = maxLevel // saturate level so the shift above can never overflow
 	}
+	// Equal jitter: keep a floor of window/2, spread the remainder randomly. This
+	// de-synchronizes pools that failed from a shared cause (e.g. common-AZ ICE) so
+	// their probe windows don't expire — and re-escalate — in lockstep.
+	window = window/2 + time.Duration(b.rand.Int63n(int64(window/2)))
 	e.until = b.clock.Now().Add(window)
-	// Once the window has saturated at maxDelay, stop growing level so the shift
-	// above can never overflow on a pool that fails indefinitely.
-	if window == maxDelay {
-		e.level = maxLevel
-	}
 }
 
 func (b *NodePoolBackoff) Reset(nodePool string) {
@@ -375,8 +378,9 @@ Sequence for a persistently failing pool (`spark`) alongside a healthy younger p
    and serviced normally.
 4. At `now+1m`: `spark` probes one candidate. If ICE again → `Fail` → `level=2`,
    `until=now+2m`. If it succeeds → `Reset("spark")`, pool healthy again.
-5. Window grows `1m, 2m, 4m, 8m, 10m, 10m…` (clamped at `maxDelay`) until capacity
-   returns.
+5. The window grows `1m, 2m, 4m, 8m, 10m, 10m…` (clamped at `maxDelay`)
+   until capacity returns; each actual window is equal-jittered to `[½w, w)`, so the
+   values above are centers, not exact times.
 
 This removes both symptoms: `spark` no longer holds the head of line every pass, and
 launch attempts drop from "every pass" to "one per back-off window."
@@ -424,7 +428,10 @@ launch attempts drop from "every pass" to "one per back-off window."
 ## Testing plan
 
 - **Unit (tracker):** `Fail` growth/cap, `Reset`, `Selectable` transitions
-  (healthy → backing off → probe → recovered), driven by a fake `clock.Clock`.
+  (healthy → backing off → probe → recovered), driven by a fake `clock.Clock`. Inject a
+  seeded `*rand.Rand` so jitter is deterministic; assert each window lands in `[½w, w)`
+  and that two pools failed at the same instant with the same `level` get *different*
+  `until` values (de-synchronization).
 - **Unit (`Drift.ComputeCommands`):** with a backed-off NodePool, assert its
   candidates are skipped and a younger NodePool's candidate is selected instead;
   assert the probe allows exactly one in-flight replacement after expiry.
--- a/designs/drift-per-nodepool-backoff.md
+++ b/designs/drift-per-nodepool-backoff.md
@@ -26,120 +26,20 @@ the first schedulable candidate.
 
 During a batched drift rollout, such as changing a shared AMI selector so that many
 NodePools drift at once, the drift controller repeatedly picks the globally oldest
-drifted candidate. The relevant selection logic sorts candidates strictly by the
-`Drifted` status condition's `LastTransitionTime`:
-
-```64:108:pkg/controllers/disruption/drift.go
-	sort.Slice(candidates, func(i int, j int) bool {
-		return candidates[i].NodeClaim.StatusConditions().Get(string(d.Reason())).LastTransitionTime.Time.Before(
-			candidates[j].NodeClaim.StatusConditions().Get(string(d.Reason())).LastTransitionTime.Time)
-	})
-
-	emptyCandidates, nonEmptyCandidates := lo.FilterReject(candidates, func(c *Candidate, _ int) bool {
-		return len(c.reschedulablePods) == 0
-	})
-
-	// Prioritize empty candidates since we want them to get priority over non-empty candidates if the budget is constrained.
-	// ...
-	for _, candidate := range slices.Concat(emptyCandidates, nonEmptyCandidates) {
-		if disruptionBudgetMapping[candidate.NodePool.Name] == 0 {
-			continue
-		}
-		results, err := SimulateScheduling(ctx, d.kubeClient, d.cluster, d.provisioner, d.clock, d.recorder, nil, candidate)
-		// ...
-		cmd := Command{
-			Candidates:          []*Candidate{candidate},
-			Replacements:        replacementsFromNodeClaims(results.NewNodeClaims...),
-			Results:             results,
-			PoolDisruptionCosts: computePoolDisruptionCosts([]*Candidate{candidate}),
-		}
-		return []Command{cmd}, nil
-	}
-	return []Command{}, nil
-```
-
-After a command is dispatched, the disruption controller requeues immediately, so
-this selection runs many rapid passes during a rollout:
-
-```166:183:pkg/controllers/disruption/controller.go
-	// Attempt different disruption methods. We'll only let one method perform an action
-	for _, m := range c.methods {
-		c.recordRun(fmt.Sprintf("%T", m))
-		success, err := c.disrupt(ctx, m)
-		if err != nil {
-			// ...
-		}
-		if success {
-			return reconciler.Result{RequeueAfter: singleton.RequeueImmediately}, nil
-		}
-	}
-
-	// All methods did nothing, so return nothing to do
-	return reconciler.Result{RequeueAfter: pollingPeriod}, nil
-```
+drifted candidate. After a command is dispatched, the disruption controller requeues
+immediately.
 
 Replacement orchestration is asynchronous, in the drift/disruption `Queue`. When a
 replacement fails unrecoverably (e.g. the replacement NodeClaim was deleted after an
 ICE, or the command timed out), the queue tears the command down and unmarks the
 candidate for deletion, returning it to the candidate pool unchanged — crucially,
-**with its original drift timestamp**:
-
-```142:181:pkg/controllers/disruption/queue.go
-func (q *Queue) Reconcile(ctx context.Context, nodeClaim *v1.NodeClaim) (reconcile.Result, error) {
-	// ...
-	if err := q.waitOrTerminate(ctx, cmd); err != nil {
-		// If recoverable, re-queue and try again.
-		if !IsUnrecoverableError(err) {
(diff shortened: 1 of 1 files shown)


designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,463 @@
...
+	defer b.mu.Unlock()
+	e, ok := b.state[nodePool]
+	if !ok || e.level == 0 {
+		return true // healthy
+	}
+	if b.clock.Now().Before(e.until) {
+		return false // backing off
+	}
+	return inflight == 0 // probe: at most one in flight
+}
```
jamesmthompson: Is this a potential time-of-check time-of-use bug? Can't b.state[nodePool] update asynchronously from the queue reconciler? Between Selectable returning true, and the command actually entering ProviderIDToCommand, a second pass could see inflight == 0 and select another probe. 

Maybe we could have Selectable flip the entry to "probe in-flight" atomically when it returns true, rather than reading and believing the counts from the map?

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,463 @@
...
+place where a drift command's fate is known. We hook there, gated on the command's
+reason so only drift is affected:
+
+- **Success** (`cmd.Succeeded == true`): `backoff.Reset(nodePool)`.
+- **Unrecoverable failure**: `backoff.Fail(nodePool)`.
+
+Because a drift command always has exactly one candidate (and therefore one
+NodePool), the NodePool key is `cmd.Candidates[0].NodePool.Name`. We guard with
+`cmd.Reason() == v1.DisruptionReasonDrifted` so consolidation/emptiness commands do
+not touch drift back-off state.
```
jamesmthompson: I think this assumes we have each drift command apply to a single pool. I think that's fine, but probably worth specifying as an invariant. I'd really like for us to have more invariants like that so the space of things we can think about is lower. If drift ever has a single command that spans pools, I think we'll need want to think through that.

Could we maybe have a short "Invariants" or "Assumptions" section, similar to "Goals?"

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,463 @@
...
+```
+
+Proposed defaults (mirroring the existing retry-duration bounds in `queue.go`, which
+already use `10m`–`1h`):
+
+| Parameter        | Default | Rationale |
+| ---------------- | ------- | --------- |
+| `baseDelay`      | `1m`    | First retry after a short delay; fast enough to recover quickly if ICE clears, slow enough to stop the storm. |
+| `maxMultiplier`  | `8`     | Caps the window at `8 × baseDelay`. |
+| `maxDelay` (cap) | `10m`   | Absolute ceiling on the window regardless of multiplier. |
```
jamesmthompson: Can we pick one of these? I personally find two caps on the max delay harder to reason about. You can definitely change my mind if you feel strongly about having two.

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,463 @@
...
+
+### Back-off formula
+
+On each unrecoverable failure, grow the window exponentially, matching the model in
+the issue's simulation (`base × 2^(n−1)`, capped):
+
+```
+level      := level + 1
+multiplier := min(2^(level-1), maxMultiplier)
+until      := now + baseDelay * multiplier
```
jamesmthompson: Should we pick a randomly jittered value within the backoff window? Imagine an ICE event that makes multiple pools backoff at the same time. The standard exponential backoff algorithm adds randomness to avoid correlated retries.

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,463 @@
...
+  consumed by NodeClaims that are slower to disrupt (e.g. `do-not-disrupt` with
+  termination-grace-period enabled), effectively trading one head-of-line problem for
+  another. As of this writing the issue thread has not settled whether fairness
+  should become a second, separate RFC/PR — the issue author raised a
+  counter-scenario (a large, healthy NodePool with fast replacements starving a
+  smaller one) suggesting fairness may still be needed for that case. This RFC
+  intentionally stands alone as the back-off half of the two originally proposed
+  fixes; fairness ordering is left to a follow-up RFC pending that discussion, so as
+  not to block back-off (which addresses the reported production issue on its own)
+  on an unresolved design question.
```
jamesmthompson: I think this is good context, but we don't really need it here. I think we'll probably change our minds as the conversation continues over the next few weeks. I think you can just say something like "Backoff in this RFC fixes head-of-line blocking from persistent failures, and that's a nice win. Fairness fixes asymmetric service rates when things are working well enough for us to make progress, and fairness is more nuanced. As such, we'll tackle that later." 

Curious what other folks think.

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,463 @@
...
+```
+
+Sequence for a persistently failing pool (`spark`) alongside a healthy younger pool
+(`ingress`), with defaults `baseDelay = 1m`:
+
+1. Pass N: `spark` oldest candidate selected → replacement launched.
+2. Queue: ICE → unrecoverable → `Fail("spark")` → `level=1`, `until=now+1m`.
+3. Passes N+1..: `spark` skipped (backing off). `ingress` candidates are now reached
+   and serviced normally.
+4. At `now+1m`: `spark` probes one candidate. If ICE again → `Fail` → `level=2`,
```
jamesmthompson: Today's code picks the oldest candidate every time. If that candidate is blocked for persistent reasons the ICE cache can't see today (PDB violations, scheduling constraints, subnet exhaustion), it fails repeatedly and we can't drift any nodes in the pool. Should we rotate or randomize within equivalence classes, and/or have something like the ICE cache inform us when particular candidates can't be replaced?

designs/drift-per-nodepool-backoff.md
cnmcavoy: Yeah, jitter is a good addition.

designs/drift-per-nodepool-backoff.md
cnmcavoy: Sure, I'll simplify this.

designs/drift-per-nodepool-backoff.md
cnmcavoy: Added an invariants section and included some additional ones as well.

designs/drift-per-nodepool-backoff.md
cnmcavoy: You're right that per-pool back-off is coarse and a persistently-blocked oldest candidate is a real gap.

I considered randomizing what candidate is chosen when a NodePool is backed off - so changing the candidate being probed. However, in a scenario where a NodePool is backed off and the oldest candidate is ICE'd, and a younger candidate can be replaced, that would `Reset` the NodePool back to a healthy state - then it would begin backing off the oldest candidate when the oldest-candidate selection criteria resumed. Flip-flopping between states is undesirable, even if the pool is advancing, because of the wasted launches on the oldest each cycle.

As for rotating the candidates, the tradeoff is relaxing strict oldest-first, which within an equivalence class is minor - so I'm open to relaxing this. What do you think?

As for a potential ICE-like cache, I'm not sure how valuable it would be. `SimulateScheduling` already runs every pass and checks the `UnavailableOfferings` cache via the provider interface in `GetInstanceTypes`. Because pre-launch causes (PDB, affinity, scheduling) are already handled because `SimulateScheduling` skips candidates whose pods can't reschedule, a core-side cache would mostly duplicate existing structure, and the only benefit is to post-launch causes like subnet/quota exhaustion.

I'm inclined to keep v1 at per-pool granularity and capture candidate-level viability (rotation + a viability signal) as a scoped follow-up - does that split seem reasonable?

designs/drift-per-nodepool-backoff.md
cnmcavoy: It's not, but only because the disruption controller is currently a singleton and passes are serialized. The async queue reconciler only ever removes entries (frees the slot), which can't cause a double-probe. However, if any of those changes - concurrent reconciles, batched drift commands, or the fairness reordering emitting multiple commands - the current approach can admit two probes.

I updated the tracker implementation to avoid this brittleness, what do you think of this? https://github.com/kubernetes-sigs/karpenter/pull/3128/changes/c6ca3bd148016dbaeb2646ff96dbfb6fc0bdb877

jamesmthompson: Thanks, the atomic claim approach that you made in c6ca3bd is much cleaner. Even if the controller stays single-threaded for a while, encoding the invariant in the data structure rather than relying on external serialization makes future refactors safer. LGTM on this part, thanks for talking through it.

designs/drift-per-nodepool-backoff.md
jamesmthompson: I hadn't fully thought through what happens when Reset brings the stuck oldest candidate right back into contention. That's a compelling reason to keep v1 coarse-grained.

I'm fine with per-pool back-off now, candidate-level viability (rotation within equivalence classes + a post-launch signal) as a scoped follow-up. Could you add a brief "Future Work" or "Deferred" bullet in the RFC capturing the rotation idea and the post-launch viability signal? Might be useful in the future, I'm ok not including it too. 

On the ICE-cache point, agreed. Duplicating SimulateScheduling's pre-launch checks isn't worth it. The interesting gap is post-launch failures (subnet exhaustion, quota), and that's naturally a separate piece of work.

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,540 @@
...
+	recorder    events.Recorder
+	clock       clock.Clock
+}
+```
+
+The `Drift` method instance is created once (via `NewMethods` inside `NewController`)
+and persists for the controller's lifetime, so the shared tracker is stable across
+passes — the same property `PreviouslyUnseenNodePools` relies on.
+
+## Detailed design
```
jigisha620, line 198: The Tryclaim/claim/release pattern is thorough, but I'm wondering if a simpler approach could achieve the same guarantees. The backoff window itself already solves the primary goals, stopping starvation and reducing wasted API calls (from every-pass to once-per-window). The probe cap further reduces it to exactly one attempt per window, but adds the probing flag, the claim struct, and the Release requirement on every non-dispatch path. If Fail simply no-ops when the pool is already backed off (until is in the future), multiple simultaneous failures from the same cycle would only increment the level once, preventing level inflation without needing the probe cap. The cost is a small number of extra API calls per window. Is there a scenario where exactly-one-probe is important beyond the API call reduction? Or would this simpler Ignore repeated failures approach be sufficient?

jigisha620 (commented):
Small format suggestion, some of the pseudocode reads closer to implementation than design (e.g. the full TryClaim/Release logic). The struct definitions and state transitions are great, but the procedural details can be better left to the implementation PR, maybe?

designs/drift-per-nodepool-backoff.md
cnmcavoy: Added a future work section in [67c973a](https://github.com/kubernetes-sigs/karpenter/pull/3128/commits/67c973a2ba1c0371c896b6bd15990da7e98c7ee4).

designs/drift-per-nodepool-backoff.md
cnmcavoy: Yeah, I agree this can be simpler without giving up much. Removing the probe cap changes exactly one thing: how many attempts fire in the brief re-selectable slice at the start of each window — and the disruption budget still bounds that burst.

The main thing exactly-one-probe bought us was protection against slow, timeout-class failures: a burst of those could each hold a replacement in flight for the full command timeout before the first failure re-arms back-off, inflating global disruption concurrency. But per-pool back-off already isolates the pool for the rest of the window, so that's a short transient rather than sustained competition with other NodePools.

And the burst can actually be desirable: if the oldest candidates are ICE'd in one AZ, launching several at once gives other AZs a chance to succeed, and a single success Resets the NodePool to healthy.

I moved the probe-cap implementation details to a Future Work section in [291f486](https://github.com/kubernetes-sigs/karpenter/pull/3128/commits/291f486572ec94076ed7f3f6edaa18211736f24a) if we want to revisit it.

jigisha620: Thanks for making the change. I will take another look soon.

jigisha620 (approved):
/lgtm

jigisha620:
@jmdeal and @DerekFrank, the RFC looks pretty good to me. If you guys want to take a look as well.

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,497 @@
...
+the first schedulable candidate.
+
+## Background
+
+During a batched drift rollout, such as changing a shared AMI selector so that many
+NodePools drift at once, the drift controller repeatedly picks the globally oldest
+drifted candidate. The relevant selection logic sorts candidates strictly by the
+`Drifted` status condition's `LastTransitionTime`:
+
+```64:108:pkg/controllers/disruption/drift.go
```
DerekFrank: I don't know if you need to have all this in the RFC, its useful background but we don't tend to drop this much code into the designs folder

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,497 @@
...
+  is what keeps `level` counting failed *windows* rather than individual attempts, and
+  it means the tracker never needs to inspect queue state.
+
+### Back-off formula
+
+On each unrecoverable failure, grow the window exponentially, clamp it to a single
+absolute ceiling (`maxDelay`), and apply randomized **jitter**:
+
+```
+level  := level + 1
```
DerekFrank, line 128: Can we reuse something like controller runtime's NewItemExponentialFailureRateLimiter? Maybe not given where this has to live.

designs/drift-per-nodepool-backoff.md
```diff
@@ -0,0 +1,497 @@
...
+- **The `Drift` method instance and the shared `NodePoolBackoff` persist for the
+  controller's lifetime.** Karpenter runs a single active disruption reconciler
+  (singleton), so there is exactly one reader and no cross-replica coordination. In-
+  memory state is therefore sufficient and authoritative for the running process. This is a performance/fairness
+  optimization, not correctness-critical; discarding it (e.g. on restart) at worst
+  re-attempts a failing pool once before backing off again.
+
+## Proposal
+
+Introduce a **per-NodePool drift back-off tracker**. Three interaction points:
```
DerekFrank, line 83: Did we consider just having the backoff be if the nodepool is RegistrationHealthy=False?

designs/drift-per-nodepool-backoff.md
cnmcavoy: It doesn't fit cleanly. `NewItemExponentialFailureRateLimiter` is designed to return a delay for the workqueue to sleep on, whereas the existing back-off is read as a boolean check in a different controller.

designs/drift-per-nodepool-backoff.md
cnmcavoy: My understanding is that `NodeRegistrationHealthy=False` is a narrow signal for launch/registration timeouts caused by misconfiguration (bad SG/AMI/userdata) and so it excludes transient/deterministic failures like ICE. In addition, `NodeRegistrationHealthy` is reset to Unknown on any NodeClass/NodePool spec change (e.g an AMI update that triggers drift). So in the starvation scenario it would typically be `Unknown`, not `False`.

In our clusters, the only condition values observed on our NodePools `NodeRegistrationHealthy` is `True` or `Unknown`.

DerekFrank (commented):
Other than the code segments that I don't think this needs, I think it lgtm!

designs/drift-per-nodepool-backoff.md
cnmcavoy: Removed.

DerekFrank (approved):
/lgtm

Excited to see this!

PR #3129, opened 2026-07-07
feat: bump CapacityBuffer to v1beta1 and emit operator_* status metrics

DerekFrank:
Fixes #N/A 

**Description**

Bumps the `CapacityBuffer` API from `v1alpha1` to `v1beta1` and wires it into the operatorpkg status controller so it emits the standard `operator_status_condition_*` metrics like `NodeClaim`, `NodePool`, and `Node`.

- **`v1beta1` is now the sole served + storage version**, matching the upstream Cluster Autoscaler `v1beta1` schema. The serialized CRD schema is unchanged apart from the version name, so stored objects remain wire-compatible with Cluster Autoscaler's CRD.
- **Status conditions now use operatorpkg's `status.Condition`** (a type alias of `metav1.Condition` with identical wire format) instead of the bespoke `SetCondition` helper. This lets `CapacityBuffer` implement the operatorpkg `status.Object` interface (`StatusConditions`/`GetConditions`/`SetConditions`). The root `Ready` condition is derived from `ReadyForProvisioning`, matching the condition the provisioner already gates buffer provisioning on; `Provisioning` and `LimitedByQuotas` remain informational.
- **The status/metrics controller is registered under the existing `CapacityBuffer` feature gate** and additionally gated on `DisableClusterStateObservability`, consistent with the other status controllers.
- Regenerated deepcopy and CRD manifests; updated the controller, provisioner, test helpers, and tests to the new API package and condition API.

- `go build ./...` and `go vet ./pkg/... ./test/...` pass.
- `go test` passes for `pkg/apis/autoscaling/v1beta1`, `pkg/controllers/capacitybuffer`, and `pkg/controllers/provisioning` (including the CapacityBuffer and DRA suites) against a Kubernetes 1.36 envtest apiserver.
- Confirmed the CRD `openAPIV3Schema` for the condition status is byte-identical to the previous `v1alpha1` schema (only the version name and a condition description string change), preserving compatibility with the shared `autoscaling.x-k8s.io` CRD.


pkg/apis/autoscaling/v1beta1/capacitybuffer.go
```diff
@@ -165,23 +171,29 @@ type CapacityBufferStatus struct {
 	// +patchStrategy=merge
 	// +listType=map
 	// +listMapKey=type
-	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,4,rep,name=conditions"` //nolint:kubeapilinter
```
jmdeal, line 168: I'm not sure how I feel about diverging from the upstream type definition, even if it is structurally compatible today. Ideally, we should not be defining this type ourselves and should be pulling it from an upstream types dependency, as we do for all other k8s API types — I don't see why we'd want to treat capacity buffers differently. 

Is there a reason we need to use the upstream type to implement this type? Can't we use a `GenericObjectController` for capacity buffer, just like we do for nodes? This would also address another point of concern I had: the introduction of a `Ready` condition. Upstream likely wouldn't introduce a `Ready` condition in the future, but it's risky to use it and risk one being introduced in the future with semantics that our different than our status semantics. Using a `GenericObjectController` would still result in a `ReadyForProvisioning` metric being emitted without persisting a new, Karpenter specific, condition to the API server.

DerekFrank: Fair point, let me flip it back

pkg/controllers/controllers.go
```diff
@@ -173,6 +174,20 @@ func NewControllers(
 
 	if options.FromContext(ctx).FeatureGates.CapacityBuffer {
 		controllers = append(controllers, capacitybuffer.NewController(kubeClient, p))
+		if !options.FromContext(ctx).DisableClusterStateObservability {
```
sumukha-radhakrishna, line 177: TIL about status observability. Thanks for the change, lgtm

PR #3130, opened 2026-07-07
fix: use max() of replicas and percentage in CapacityBuffer replica calculation

### LINKED ISSUE #3125: computeReplicas uses min() for replicas and percentage, diverging from the CapacityBuffer specification
I think this is a correctness issue, assuming the goal of PR #3028 is to implement the upstream CapacityBuffer API rather than intentionally introduce a semantic divergence. If the divergence is intentional, it should be documented in the API and PR rationale.

https://github.com/kubernetes-sigs/karpenter/pull/3028/changes#r3521842383

> This computes the desired buffer replica count as the minimum of all candidates. Is this intentionally diverging from the upstream CapacityBuffer API? The upstream API docs says when both replicas and percentage are set, the maximum of the two is used. With min(), Karpenter can under-provision the requested buffer.
> 
> Example:
> 
> scalableRef replicas = 100
> percentage = 20%  => 20 buffer pods
> replicas = 50
> 
> Expected by API docs
> 
> max(20, 50) = 50
> 
> Current implementation:
> 
> min(20, 50) = 20
> 
> 
> https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/proposals/buffers.md
> For the initial implementation on buffers the users should use other tools to limit the size of the cluster (max size on the GKE node pool/CCC or karpenter pool, max total size). Once we launch we will reassess the need for other mechanisms for buffers.
> 
> 
> https://docs.cloud.google.com/kubernetes-engine/docs/reference/crds/capacitybuffer
> 
> integerDefines the selected number of buffer units to provision. If you don't set replicas or percentage, GKE creates as many units as fit within defined resource limits. If both are set, GKE uses the maximum of the two values provided they both fit within the limits.
> 
> 
> Here is the full logic:
> ```
> scalableRef replicas = 100
> 
> percentage=30% => percentageResult = 30
> replicas=50    => replicasResult = 50
> limits allows 80 => limitResult = 80
> 
> candidates = [30, 50, 80]
> ```
> 
> Correct upstream-style logic:
> 
> ```
> desired = max(30, 50) = 50
> final   = min(50, 80) = 50
> ```
> 
> Current lo.Min(candidates) logic:
> ```
> min(30, 50, 80) = 30
> ```
> 
> So it under-provisions by 20 buffer pods.
>  

 _Originally posted by @w1t in [#3028](https://github.com/kubernetes-sigs/karpenter/pull/3028/changes#r3521842383)_

DerekFrank:
Fixes #3125

**Description**

`computeReplicas` combined `replicas`, `percentage`, and `limits` into a single `min()`, which under-provisions the buffer when both `replicas` and `percentage` are set. The upstream CapacityBuffer API contract treats `replicas` and `percentage` as sizing signals combined with `max()` (both are floors — the larger wins), with `limits` applied as an upper-bound cap via `min()`.

This changes the calculation to `min( max(replicas, percentage), limits )`:

- `replicas` and `percentage` are combined with `max()` — an explicit `replicas` floor is never silently reduced by a smaller percentage-derived value (or vice versa).
- `limits` remains an upper-bound cap.
- When neither `replicas` nor `percentage` is set, `limits` alone determines the buffer size (unchanged).

Example (`scalableRef` replicas = 100, `percentage` = 20 → 20, `replicas` = 50):
- Before: `min(20, 50) = 20` (under-provisions the requested floor of 50)
- After: `max(20, 50) = 50`

This matches the documented API (`autoscaling.x-k8s.io` `Replicas` field doc-comment: "If both are set, the maximum of the two will be used") and the design proposal (`cluster-autoscaler/proposals/buffers.md`: "if both are set... max from these"), which have specified `max` since the original proposal (kubernetes/autoscaler#8151).

Note: the Cluster Autoscaler reference controller currently implements this with `min()`, contradicting its own API docs. Filed upstream as kubernetes/autoscaler#9948. This PR follows the documented `max()` semantics rather than CAS's current (buggy) implementation.

- Added a unit test covering the previously-uncovered case where both `replicas` and `percentage` are set, asserting `max()` wins.
- Updated the replicas-capped-by-limits test's rationale (result unchanged: `max(10)` capped by a limit of 3 = 3).
- `go test ./pkg/controllers/capacitybuffer/...` and the provisioning buffer suite pass against a Kubernetes 1.36 envtest apiserver.

### DIFF
--- a/pkg/controllers/capacitybuffer/suite_test.go
+++ b/pkg/controllers/capacitybuffer/suite_test.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"sigs.k8s.io/karpenter/pkg/apis"
 	autoscalingv1alpha1 "sigs.k8s.io/karpenter/pkg/apis/autoscaling/v1alpha1"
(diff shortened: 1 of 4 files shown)


pkg/controllers/capacitybuffer/suite_test.go
```diff
@@ -690,6 +556,93 @@ var _ = Describe("CapacityBuffer Controller", func() {
...
+// only valid with a scalableRef.
+type podTemplateBufferSpec struct {
+	replicas int32
+	podCPU   string
+	limitCPU string
+}
+
+// reconcilePodTemplateBufferReplicas applies the PodTemplate + CapacityBuffer described by
+// spec, reconciles once, and returns the resolved status.Replicas.
+func reconcilePodTemplateBufferReplicas(spec podTemplateBufferSpec) int32 {
```
jmdeal: You'll want to add a GinkgoHelper call to the top of each of these helper functions

PR #3131, opened 2026-07-08
fix: buffer computeReplicas must take the max of desired replicas and percentage

### LINKED ISSUE #3125: computeReplicas uses min() for replicas and percentage, diverging from the CapacityBuffer specification

sumukha-radhakrishna:
Fixes #3125 

**Description**

Changes how the CapacityBuffer controller computes the desired buffer replica count to align
with Cluster Autoscaler's buffer semantics.

Before: the controller took the minimum across all configured constraints — min(replicas,
percentage, limits). This meant setting both replicas and percentage produced the smaller of the
two, which is surprising and inconsistent with Cluster Autoscaler.

After: replicas and percentage are combined by taking the maximum, and limits acts as an upper
bound on that value:

min( max(replicas, percentage), limits )

- If both replicas and percentage are set, the larger wins (they express two ways of asking for
capacity; the user gets whichever is bigger).
- limits always caps the result so the buffer never exceeds the configured resource budget.
- If neither replicas nor percentage is set, limits alone determines how many chunks fit.
- If no constraints are set, the count is 0.

This matches Cluster Autoscaler's documented behavior: "If both are set, [the autoscaler] uses
the maximum of the two values provided they both fit within the limits."

This is a behavior change only for buffers that set both replicas and percentage. Buffers using
a single constraint (replicas only, percentage only, or limits only) are unaffected.

Unit tests and make presubmit 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/karpenter
 
-go 1.26.4
+go 1.26.5
 
 require (
 	github.com/Pallinder/go-randomdata v1.2.0
--- a/go.tools.mod
+++ b/go.tools.mod
@@ -3,7 +3,7 @@ module tools
 // Tool dependencies isolated from main module to avoid conflicts.
 // Usage: go tool -modfile=go.tools.mod <toolname>
 
-go 1.26.3
+go 1.26.5
 
 tool (
 	github.com/B1NARY-GR0UP/nwa
--- a/pkg/controllers/capacitybuffer/controller.go
+++ b/pkg/controllers/capacitybuffer/controller.go
@@ -176,19 +176,34 @@ func (c *Controller) resolveAndUpdateStatus(ctx context.Context, cb *autoscaling
 	return true, nil
 }
 
+// computeReplicas derives the desired buffer replica count from the configured
+// constraints, following Cluster Autoscaler's semantics:
+//   - replicas and percentage are combined by taking the MAX of the two.
+//   - limits act as an upper bound (MIN) on that value.
+//   - if neither replicas nor percentage is set, limits alone determine how many
+//     units fit within the resource limits.
+//
+// candidates holds the percentage-derived replica count (if percentage is set);
+// the fixed replicas value is added here.
 func computeReplicas(cb *autoscalingv1alpha1.CapacityBuffer, podSpec *v1.PodSpec, candidates []int32) int32 {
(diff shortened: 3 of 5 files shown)


DerekFrank (approved):
/lgtm

PR #3133, opened 2026-07-09
docs: NodeClaim optimization RFC

mmanders-amzn:
Fixes #N/A

**Description**

Adds the design doc for an **opt-in, post-scheduling NodeClaim optimization pass** to `designs/`.

Karpenter's scheduler packs pods onto NodeClaims in a single greedy pass and never revisits a placement. A pod added late in a batch can push a NodeClaim's cheapest fitting instance type up a tier, leaving the earlier pods on a machine larger than they need. The pass detects this after the queue drains, rolls the claim back to the earlier, cheaper state it recorded during packing, and re-queues the displaced pods onto a fresh, right-sized NodeClaim.

The doc covers:

- The algorithm: transition history, split-point detection (price / efficiency / displaced-pod eligibility / net-savings gates), and revert.
- The correctness invariant: optimized cost ≤ baseline cost, within a float-rounding epsilon.
- Safety: reserved-offering strict-mode short-circuit, topology-count reversal via recorded deltas, and host-port cleanup.
- Rollout: off by default behind the `NodeClaimOptimization` feature gate, and the deliberate carve-out that keeps the pass out of the disruption controller's simulated scheduler.

This is a design-first, docs-only PR. The implementation is in #3008.

Docs-only change; no code is affected.

### DIFF
--- a/designs/nodeclaim-optimization.md
+++ b/designs/nodeclaim-optimization.md
@@ -1,0 +1,156 @@
+# NodeClaim Optimization
+
+## Motivation
+
+Karpenter's scheduler can commit a batch of pods to a single NodeClaim whose instance type is much larger, and much more expensive, than most of those pods actually need. The cause is single-NodeClaim packing: `Solve` processes pods in sort order, grows each new NodeClaim's instance-type set as pods are added, and never revisits a placement within the same call. As pods accumulate, the cheapest fitting instance type steps up through tiers; once it steps up, the claim stays at that larger instance type for the rest of `Solve`, regardless of whether an earlier subset would have packed cleanly on the previous tier. A single NodeClaim often ends up holding a set of pods that would be cheaper split across two claims of different tiers.
+
+For example, 16.5 vCPU of pods land on a single 32 vCPU instance at ~51% efficiency. The same pods split as 16 vCPU on a 16 vCPU instance and 0.5 vCPU on a 2 vCPU instance, assuming this is the smallest vCPU sized instance available, run at ~91% efficiency on cheaper total capacity. The scheduler had enough information to choose the second layout; it simply never reconsidered once the 32 vCPU tier was locked in.
+
+## Proposal
+
+Add an **opt-in post-scheduling optimization pass** that, at the end of the scheduling pass, inspects each newly-built NodeClaim's efficiency and, if a split is cheaper overall, rolls the NodeClaim back to an earlier cheaper state and re-queues the displaced pods. The displaced pods then flow through normal scheduling onto a fresh, right-sized NodeClaim.
+
+The entire feature lives in `pkg/controllers/provisioning/scheduling/` and touches only the scheduler's in-memory simulation: no CRD changes, no persisted state, no user-facing API. The only Go API change is additive and source-compatible: `Topology.Record` now returns a `[]TopologyDelta` (existing callers can silently discard it) and a new `Topology.Unrecord` method replays those deltas. This is discussed in **Topology deltas instead of re-derivation** below.
+
+### Call shape
+
+`Solve`'s outer loop repeats as long as pods remain. When the queue empties, the gated new step runs:
+
+```
+Solve loop:
+  pop pod → place on existing node / buffered NodeClaim / fresh NodeClaim from template
+  ... until queue empty ...
+  if s.OptimizationEnabled && tryOptimize(q):    // NEW, gated
+      for each unlocked new NodeClaim:
+        idx := findSplitPoint()
+        if idx >= 0 { displaced += nc.RevertTo(idx) }
+        nc.locked = true
+      if displaced: *q = NewQueue(displaced)
+      continue outer loop                        // re-enter with refilled queue
+  break                                          // pass disabled or no displacement
+```
+
+Each NodeClaim is visited at most once per `Solve`: once inspected it is *locked* and `CanAdd` rejects further pods. The pass itself runs at most `maxOptimizationPasses` (currently 2) times per `Solve`.
+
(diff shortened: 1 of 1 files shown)


designs/nodeclaim-optimization.md
```diff
@@ -0,0 +1,156 @@
...
+    PodCount         int
+    InstanceTypes    []*cloudprovider.InstanceType
+    Requirements     scheduling.Requirements
+    ResourceRequests corev1.ResourceList
+    CheapestInstance *cloudprovider.InstanceType
+    Price            float64
+}
+```
+
+A SchedulingOption is captured when the addition of a pod pushes the NodeClaim to a larger sized instance; it is not captured at every pod addition. This is enough information to answer "what would this NodeClaim look like if we stopped accepting pods at point N?"
```
jigisha620, line 50: Snapshots are captured only at tier transitions correct? What happens to pods added between two transitions that fit fine at the earlier tier? For example, Pods A, B land on 8 vCPU (snapshot). Pod C is added that still fits on 8 vCPU (no snapshot). Pod D pushes to 32 vCPU (snapshot). If we revert to the 8 vCPU snapshot, both C and D are displaced, even though C didn't cause the bump (unless I have misunderstood this). Gate 4 would reject the split if displacing both makes the math not work out. But could this cause us to miss valid optimizations where displacing only the tier-bumping pod would have been sufficient?

jigisha620 (commented):
The split-point approach is clever and I like that it's gated off by default. One area I'd like to think about is, what are the performance implications of the optimization pass on scheduling latency? With a large batch (e.g 1000 pods creating 50+ NodeClaims), the pass involves walking snapshot histories, estimating displaced pod costs, and maybe re-entering the Solve loop up to 2 times. Are there plans to benchmark this?

designs/nodeclaim-optimization.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Split-point detection
+
+`findSplitPoint` walks the history backwards from the current state looking for the first earlier state that would be worth splitting to. A split is only accepted when four gates all pass:
+
+1. **Price**: the candidate instance type must be at least `minPriceSavingsRatio` (24%) cheaper than current. With cloud instances, pricing is generally by powers of 2, and this decision pushes us back far enough to look back "one size down" in instance families.
+2. **Efficiency**: the candidate must be at least `minEfficiencyGain` (10%) more weighted-utilized than the current state. A split is only worthwhile if the earlier pods actually packed well on the candidate tier.
+3. **Displaced-pod eligibility**: no displaced pod may carry a required hostname-scoped constraint that the displaced-pod cost estimator cannot price correctly. Two cases are rejected (see `displacedCarriesHostFanOut`):
```
jigisha620, line 58: Gate 3 rejects displaced pods whose constraints the estimator can't account for (hostname fan-out, unreachable affinity). Are there other constraints that could lead to inaccurate estimates? For example, if a displaced pod has a PVC bound to a zone-specific volume, the PV's node affinity restricts placement to that zone. Could the estimator pick a cheaper instance in a different zone that the pod can't actually use?

mmanders-amzn:
> The split-point approach is clever and I like that it's gated off by default. One area I'd like to think about is, what are the performance implications of the optimization pass on scheduling latency? With a large batch (e.g 1000 pods creating 50+ NodeClaims), the pass involves walking snapshot histories, estimating displaced pod costs, and maybe re-entering the Solve loop up to 2 times. Are there plans to benchmark this?

This is a good idea. I focused on cost reduction/correctness when testing this across different scales. I'll update the design with relative timing impact through the rapid property-based testing.

designs/nodeclaim-optimization.md
mmanders-amzn: I believe in your example the addition of D is the event that triggers the state snapshot. When we are adding a pod that would push the nodeclaim to a larger instance we capture the state of the current nodeclaim (with A, B, and C on it). We are only taking a snapshot of the nodeclaim when it's full per the ordering of pods placed on it.

In that case, it's only 'full' when A, B, and C are placed on it - it's full because D doesn't exist. So we capture the state with pods A, B, and C.

mmanders-amzn:
> > The split-point approach is clever and I like that it's gated off by default. One area I'd like to think about is, what are the performance implications of the optimization pass on scheduling latency? With a large batch (e.g 1000 pods creating 50+ NodeClaims), the pass involves walking snapshot histories, estimating displaced pod costs, and maybe re-entering the Solve loop up to 2 times. Are there plans to benchmark this?
> 
> This is a good idea. I focused on cost reduction/correctness when testing this across different scales. I'll update the design with relative timing impact through the rapid property-based testing.

@jigisha620 I'll commit the additional tests (disabled) for running the tests to get timing information on nodeclaim optimization's impact to Solve(). 

Here are the results of running this against a set of simple (called uniform - no constraints) and diverse constraints:
image

There is a linear impact when enabling nodeclaim optimization, the base solve() call has a quadratic timing impact. Here is the per-pod cost:
image

How do you recommend we think about timing impacts to updates like these?

designs/nodeclaim-optimization.md
jigisha620: Got it, thanks for the clarification.

jigisha620:
>How do you recommend we think about timing impacts to updates like these?

Looking at the graphs, I think the linear optimization profile wouldn't become the bottleneck when scaling. Given that this feature would also live behind a feature gate which is opt-in, this seems acceptable. Out of curiosity, what constraints were used in the diverse benchmark?

jigisha620 (approved):
/lgtm

PR #3135, opened 2026-07-12
fix!: reset unschedulable_pods_count metric to 0 when no pods are pending

### LINKED ISSUE #1993: `karpenter_scheduler_unschedulable_pods_count` has unnecessary `DeletePartialMatch` in the scheduler

Archong-Liu:
Fixes #1993

**Description**

`karpenter_scheduler_unschedulable_pods_count` is only ever `.Set()` from the provisioning controller (`provisioner.go`), never from disruption, so the `DeletePartialMatch` on it at the top of every `Scheduler.Solve()` call was dead code when invoked from the disruption path — removed. Since the metric is provisioning-only, the now-unnecessary `controller` label is also dropped from the metric definition and both `.Set()` call sites, matching the other zero-label gauges in this file (e.g. `IgnoredPodCount`).

Two prior attempts at this issue (#2049, #2006) were closed without addressing review feedback: in both, @jonathan-innis pointed out that simply removing the delete isn't sufficient on its own, since the metric also needs to be reset to 0 when the provisioning loop has no pending pods to process (`Schedule()` returns early before `Solve()`/`Set()` ever run, so the metric would otherwise keep reporting its last stale value indefinitely). This PR adds that explicit reset in the early-return path.

Added a test that drives 10 unschedulable pods to raise the metric to 10, then deletes them and re-reconciles, asserting the metric returns to 0. Confirmed it fails against the old code (metric stays stuck at 10) and passes with this fix.

`make presubmit`

### DIFF
--- a/pkg/controllers/provisioning/provisioner.go
+++ b/pkg/controllers/provisioning/provisioner.go
@@ -380,6 +380,9 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 	pods := append(pendingPods, deletingNodePods...)
 	// nothing to schedule, so just return success
 	if len(pods) == 0 {
+		// No pods means no unschedulable pods either -- Solve() won't run to report this, so
+		// reset it directly or it will keep reporting its last value indefinitely.
+		scheduler.UnschedulablePodsCount.Set(0, nil)
 		return scheduler.Results{}, nil
 	}
 	deletingPodUIDs := sets.New(lo.Map(deletingNodePods, func(p *corev1.Pod, _ int) types.UID { return p.UID })...)
@@ -432,9 +435,7 @@ func (p *Provisioner) Schedule(ctx context.Context) (scheduler.Results, error) {
 	scheduler.UnschedulablePodsCount.Set(
 		// A reserved offering error doesn't indicate a pod is unschedulable, just that the scheduling decision was deferred.
 		float64(len(results.PodErrors)-len(reservedOfferingErrors)),
-		map[string]string{
-			scheduler.ControllerLabel: injection.GetControllerName(ctx),
-		},
+		nil,
 	)
 	if len(results.NewNodeClaims) > 0 {
 		log.FromContext(ctx).WithValues(
--- a/pkg/controllers/provisioning/scheduling/metrics.go
+++ b/pkg/controllers/provisioning/scheduling/metrics.go
@@ -88,9 +88,7 @@ var (
 			Name:      "unschedulable_pods_count",
 			Help:      "The number of unschedulable Pods.",
 		},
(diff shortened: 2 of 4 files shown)


praveen9354:
Clean fix — the root-cause diagnosis (`Solve()`'s `DeletePartialMatch` was dead code when called from the disruption path, and the early-return-no-pods case in `Schedule()` never reset either) is right, and the explicit `Set(0, nil)` at the top of the early-return is the correct shape. Two things worth thinking about:

**1. Dropping the `controller` label is a schema break for downstream dashboards.** Users with queries like `karpenter_scheduler_unschedulable_pods_count{controller="provisioner"}` will silently break — the matcher will match nothing after upgrade, and the same series appears unlabeled. Since the label only ever had one value in practice, it was uninformative from a query standpoint, but backwards-compat still bites. Two options: (a) keep the label with the fixed `"provisioner"` value for a release cycle, or (b) call it out in a `release-note:` block on the PR body so operators can update their queries during upgrade — I don't see one currently, and Karpenter's release notes tend to flag metric-shape changes explicitly. Curious what the maintainers' precedent is here.

**2. Defensive doc for future callers.** The comment in `scheduler.go` correctly states the metric is "only ever set from the provisioning controller," but that's a runtime invariant, not a compile-time one. If a future PR adds a `Set()` from the disruption `Solve()` call site, the missing reset here becomes a silent counter-inflation bug that's hard to spot in review. A one-liner alongside the metric declaration in `scheduling/metrics.go` (e.g., `// If you Set() this metric from another Solve() caller, add a matching reset there.`) would keep the invariant load-bearing.

Test is precise — 10 unschedulable → clear → assert 0 is exactly the regression shape, and the reference to @jonathan-innis's feedback on the earlier attempts (#2049, #2006) makes clear this is the intended follow-through. Sibling metrics (`PendingPodsByEffectiveZone`, `QueueDepth`) still keep the `ControllerLabel` because they carry per-`Solve()`-run state, which reads correctly.

AndrewMitchell25 (approved):
LGTM

Archong-Liu:
> LGTM

Thanks for reviewing!

Archong-Liu:
@DerekFrank Hi, would appreciate a feedback from you on this pr, thanks!

Archong-Liu:
@DerekFrank Hi, would like to have a little feedback from you, thank you

PR #3137, opened 2026-07-13
perf: cache virtual pod definitions to prevent re-computation

### LINKED ISSUE #3090: Buffer Controller should precompute virtual pods and store in-memory
### Description

Currently, appendVirtualPods in the provisioner hot path performs a List of all CapacityBuffers and a
  typed Get per buffer (PodTemplate or workload) on every scheduling pass. While these are informer-cache
  reads (not API server round-trips) and the cost is negligible at low buffer counts, this could be
  optimized for clusters with many buffers or high scheduling throughput.

Proposed change:

  Have the buffer controller precompute virtual pods at reconcile time and write them into a shared
  in-memory store (similar to pkg/controllers/state/Cluster). The provisioner's appendVirtualPods would
  become a simple cache read with zero API calls.

Not super important as the feature is in v1alpha1. 

zac-nixon:
Fixes #3090 

**Description**

Adds an in-memory cache keyed on the capacity buffer uid, that caches the virtual pod definitions corresponding to that capacity buffer. The buffer is hydrated on pod start up, and continuously updated through the capacity buffer controller reconcile cycle.

Tested using the aws provider. Situations tested ->

1. No capacity buffers => 0 virtual pods
2. 1 cb with 3 pod replicas => 3 virtual pods
3. 3 cbs with [3, 3, 1] replicas => 7 virtual pods
4. Scenario 3, Delete 1 cbs with replicas => 4 virtual pods
5. Remove all cbs => back to 0 virtual pods
6. Define an invalid cb => 0 virtual pods
7. Validate cold start with cbs that have virtual pods


pkg/controllers/capacitybuffer/controller.go
```diff
@@ -51,14 +55,16 @@ type ProvisionerTrigger interface {
 // (from podTemplateRef or scalableRef), computing target replica count, and
 // updating status so the provisioner knows what buffer capacity to maintain.
 type Controller struct {
```
DerekFrank, line 55: If we're going to move state management out of this, what do you think about making this a purely validation controller, and moving the state updater-logic to an informer: https://github.com/kubernetes-sigs/karpenter/tree/main/pkg/controllers/state/informer

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+type VirtualPodCache interface {
+	GetAll() []*corev1.Pod
+	HydrateCache(context.Context, client.Client) error
+	UpdateEntry(context.Context, client.Client, *autoscalingv1beta1.CapacityBuffer)
+	RemoveEntry(string, string)
+}
+
+type virtualPodCacheImpl struct {
```
DerekFrank: Is there a reason to split this into Impl\not? karpenter doesn't typically do this unless we want to make a no-op version

pkg/controllers/capacitybuffer/controller.go
DerekFrank: Looks like this PR is duping some of this status logic. Karpenter's style is typically keep all of the status\validation reconciling in the controller (https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/nodeclaim/lifecycle/controller.go#L133), and then have the actual state caching be somewhere else (/pkg/controllers/state/informer/nodeclaim.go#L78). 

I'm not sure if that can be clearly split in this case given that you have to resolve a lot of the state to do the validation?

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,110 @@
...
+	// capacityBufferToPods maps a capacity buffer identifier (name/namespace) -> to a list of virtual pods
+	capacityBufferToPods map[string][]*corev1.Pod
```
DerekFrank: Is there more we should cache for perf? Maybe the resolved pod template?

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,110 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	autoscalingv1beta1 "sigs.k8s.io/karpenter/pkg/apis/autoscaling/v1beta1"
+)
+
+type VirtualPodCache interface {
+	GetAll() []*corev1.Pod
+	HydrateCache(context.Context, client.Client) error
+	UpdateEntry(context.Context, client.Client, *autoscalingv1beta1.CapacityBuffer)
+	RemoveEntry(string, string)
```
DerekFrank: Maybe?
```suggestion
	RemoveEntry(*autoscalingv1beta1.CapacityBuffer)
```

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,110 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	autoscalingv1beta1 "sigs.k8s.io/karpenter/pkg/apis/autoscaling/v1beta1"
+)
+
+type VirtualPodCache interface {
+	GetAll() []*corev1.Pod
+	HydrateCache(context.Context, client.Client) error
```
DerekFrank: Do we want to just construct the pod cache with a client?

pkg/state/virtualpods/utilities_test.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/samber/lo"
+	corev1 "k8s.io/api/core/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	autoscalingv1beta1 "sigs.k8s.io/karpenter/pkg/apis/autoscaling/v1beta1"
+)
+
+var _ = Describe("buildVirtualPods", func() {
```
DerekFrank, line 32: We don't really do unit tests, we try to only have our tests do end to end on the user expected behavior, that way the implementation is hot swappable

DerekFrank (commented):
Let me know what you think!

pkg/controllers/capacitybuffer/controller.go
zac-nixon: I guess you're talking about resolving the PodTemplate, because it's we need to resolve the PodTemplate once to ensure it's valid and then another time to actually generate the PodSpec. The actual cache is not doing any validation, it's relying on the pattern you mention where the buffer should be validated prior to the cache being updated.

This was tricky because of the Hydration step, we need to hydrate the cache out of band of the controller, otherwise the first couple provisioner runs are running on an incomplete virtual pod state.

pkg/controllers/capacitybuffer/controller.go
zac-nixon: I don't think the controller ever managed state -- it just deferred that state management to the provisioner. I think there's some pros and cons to the approach, notably big con is that we would need to duplicate the status management logic into both this controller and the informer. Naively, the current approach seems simpler.

pkg/state/virtualpods/virtualpods.go
zac-nixon: Sounds good, I can do that.

pkg/state/virtualpods/virtualpods.go
zac-nixon: This was intentional actually, as we don't need the full capacity buffer object to delete the entry this saves us a bunch of a headaches. To achieve getting the capacity buffer correctly type cast in the capacity buffer controller we need to do use something like:
```
    WatchesRawSource(
        source.Kind(
            m.GetCache(),
            &autoscalingv1beta1.CapacityBuffer{},
            handler.TypedFuncs[*autoscalingv1beta1.CapacityBuffer, reconcile.Request]{
                DeleteFunc: func(
                    ctx context.Context,
                    e event.TypedDeleteEvent[*autoscalingv1beta1.CapacityBuffer],
                    _ workqueue.TypedRateLimitingInterface[reconcile.Request],
                ) {
                    c.virtualPodCache.RemoveEntry(e.Object)
                },
            },
        ),
    ).
```

pkg/state/virtualpods/virtualpods.go
zac-nixon: Style choice, because it does allow for easy adaptation of a no-op version. I'll use a concrete type in my next revision

pkg/state/virtualpods/virtualpods.go
zac-nixon: I didn't see where the resolved pod template was being used, caching the virtual pod here seems to cover the duplicated effort that the issue was originally concerned about.

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,142 @@
...
+	v.capacityBufferToPods[bufferKey(cb.GetNamespace(), cb.GetName())] = podCache
+}
+
+func (v *Cache) RemoveEntry(namespace, name string) {
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
+	delete(v.capacityBufferToPods, bufferKey(namespace, name))
+}
+
+func (v *Cache) HydrateCache(ctx context.Context) error {
```
DerekFrank: Okay, final idea. How about this becomes IsSynced(), that checks an atomic bool to see if the cache was warmed, and if not, warms it. The cache has a kubeClient so you only need a ctx, and the provisioner has that
```
type Cache struct {
      kubeClient           client.Client
      mu                   sync.RWMutex
      capacityBufferToPods map[string][]*corev1.Pod
      hasSynced            atomic.Bool
  }
```

```
  if !p.virtualPodCache.Synced(ctx) {
      // signal caller to requeue rather than schedule on a partial buffer view
  }
  pods = append(pods, p.virtualPodCache.GetAll()...)
```

pkg/state/virtualpods/virtualpods_test.go
```diff
@@ -0,0 +1,324 @@
...
+
+var _ = Describe("VirtualPodCache", func() {
+	var ctx context.Context
+
+	BeforeEach(func() {
+		ctx = context.Background()
+	})
+
+	Describe("GetAll", func() {
+		It("should return an empty slice for a fresh cache with no buffers", func() {
```
DerekFrank: nit: Isn't this more: 
```suggestion
		It("should return no virtual pods when there are no buffers", func() {
```

pkg/state/virtualpods/utilities_test.go
```diff
@@ -0,0 +1,340 @@
...
+			Conditions: []metav1.Condition{{
+				Type:   autoscalingv1beta1.ReadyForProvisioningCondition,
+				Status: metav1.ConditionTrue,
+				Reason: "Resolved",
+			}},
+		},
+	}
+}
+
+func readyBuffer(name string, replicas int32) *autoscalingv1beta1.CapacityBuffer {
```
DerekFrank: Can we use these established helpers? https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/test/capacitybuffer.go

pkg/state/virtualpods/utilities_test.go
```diff
@@ -0,0 +1,340 @@
...
+			Expect(p.Labels[autoscalingv1beta1.BufferNameLabel]).To(Equal("scalable-app"))
+			Expect(p.Labels[autoscalingv1beta1.BufferNamespaceLabel]).To(Equal("default"))
+			Expect(*p.Spec.Priority).To(Equal(autoscalingv1beta1.VirtualPodPriority))
+		}
+	})
+})
+
+// filterReadyBuffers mirrors the filtering logic of listBuffersReadyForProvisioning
+// without needing a real kube client.
+func filterReadyBuffers(items []*autoscalingv1beta1.CapacityBuffer) []*autoscalingv1beta1.CapacityBuffer {
```
DerekFrank, line 276: I think this will desync easily if this just replicates existing logic

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,136 @@
...
+
+// GetAll returns a snapshot of every cached virtual pod. The returned pod
+// objects are NOT deep copied, for performance; callers MUST treat them as
+// read-only and never mutate them.
+func (v *Cache) GetAll(ctx context.Context) []*corev1.Pod {
+	v.ensureHydrated(ctx)
+	v.mutex.RLock()
+	defer v.mutex.RUnlock()
+	ans := make([]*corev1.Pod, 0)
+	for _, pods := range v.capacityBufferToPods {
```
DerekFrank, line 109: Do we need to deep copy?

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,136 @@
...
+	if err := v.hydrateCache(ctx); err != nil {
+		log.FromContext(ctx).Error(err, "failed to hydrate virtual pod cache")
+		return
+	}
+	// Only mark warmed after the cache is populated so the fast path above never
+	// hands back an empty cache.
+	v.warmed.Store(true)
+}
+
+func NewVirtualPodCache(kubeClient client.Client) *Cache {
```
DerekFrank: Nit: these are usually right below the struct

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,136 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+
+	autoscalingv1beta1 "sigs.k8s.io/karpenter/pkg/apis/autoscaling/v1beta1"
+)
+
+// Cache holds the virtual (placeholder) pods materialized for each CapacityBuffer
+// so the provisioner can schedule against buffer capacity.
+type Cache struct {
```
DerekFrank, line 32: Where possible, we try to do 
```suggestion
type Cache struct {
     sync.RWMutex
```

then you can just do cache.RLock() instead of cache.mu.RLock()

Especially since I think the hydrate call should just hold the WLock() the whole time. If the cache isn't warm, nobody should be reading from it. GetAll also needs to deepcopy, which means we need to hold a WLock() for that anyways. The only advantage to an RWLock() is if you're going to have a HasEntry() in addition to UpdateEntry()

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	v.hydrateMu.Lock()
+	defer v.hydrateMu.Unlock()
+	// Re-check under the lock: another caller may have hydrated while we blocked.
+	if v.warmed.Load() {
+		return
+	}
+	if err := v.hydrateCache(ctx); err != nil {
+		log.FromContext(ctx).Error(err, "failed to hydrate virtual pod cache")
+		return
```
DerekFrank:
```suggestion
		return err
```
Karpenter always just passes the error up to be handled by the caller. this also prevents us from calling the cache hydrated when it isn't

pkg/state/virtualpods/virtualpods.go
```diff
@@ -0,0 +1,136 @@
...
+
+	newMap := make(map[string][]*corev1.Pod)
+
+	for _, cb := range buffers {
+		spec, err := resolveVirtualPodSpec(ctx, v.kubeClient, cb)
+		if err != nil {
+			log.FromContext(ctx).WithValues("capacitybuffer", client.ObjectKeyFromObject(cb)).V(1).Info("skipping buffer", "reason", err.Error())
+			continue
+		}
+		newMap[bufferKey(cb.GetNamespace(), cb.GetName())] = BuildVirtualPods(cb, spec)
```
DerekFrank:
```suggestion
		newMap[cb.UID] = BuildVirtualPods(cb, spec)
```

this is a pretty large refactor though, because DeleteFunc() calls don't contain the uid, just the namespacedName so you'd need to keep a uid -> namespacedName around. Thats the preferred way to keep track of individual objects

You could also do:

```suggestion
		newMap[object.ObjectKeyFromObject(cb)] = BuildVirtualPods(cb, spec)
```

pkg/state/virtualpods/virtualpods.go
zac-nixon: This changes the API then, 
The initial implementation swallows the error and continues with 0 virtual pods: 
https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/controllers/provisioning/buffers.go#L72

I can change it, but just want to confirm that we want a behavior change.

pkg/state/virtualpods/virtualpods.go
zac-nixon: No, also for performance reasons I don't think we want a deep-copy. Callers do not modify the pod spec. LMK if you think we should take the performance hit here.

pkg/state/virtualpods/utilities_test.go
zac-nixon: Talked offline, these tests were already generated in the first revision of capacity buffers.

pkg/state/virtualpods/virtualpods.go
zac-nixon: Refactored to basically this.

pkg/state/virtualpods/virtualpods.go
zac-nixon: yeah this doesn't to be an issue, will make the change.

pkg/state/virtualpods/virtualpods.go
zac-nixon: I don't think a deep copy is required, as the pods are r/o in their upstreams. The suggestion makes things less complicated but will reduce performance, e.g. we hold a strong mutex through out the read of the objects which means that readers will block on one another. That is probably ok in this scenario for reduced complexity.

pkg/state/virtualpods/utilities_test.go
zac-nixon: fixed

pkg/state/virtualpods/utilities_test.go
zac-nixon: fixed

pkg/state/virtualpods/virtualpods.go
zac-nixon: discussed offline - deep copy invalidates this performance improvement.

pkg/state/virtualpods/virtualpods.go
DerekFrank: I thought I had caught all the error swallowing in that PR, I must have missed one

DerekFrank (approved):
/lgtm

PR #3139, opened 2026-07-14
fix: treat empty feature gate values as unset instead of panicking

### LINKED ISSUE #3073: Helm chart installation fails with version 1.12
### Description

**Observed Behavior**:
Karpenter fails to run using default Helm values.
I noticed that `NodeOverlay` & `StaticCapacity` are empty & dont have their booleans in the `FEATURE_GATES` env variable of the karpenter deployment
```
panic: parsing feature gates, invalid value of NodeOverlay , err: strconv.ParseBool: parsing "": invalid syntax 
```
```
panic: parsing feature gates, invalid value of StaticCapacity: , err: strconv.ParseBool: parsing "": invalid syntax 
```
**Fix**:
I fixed this by adding in the booleans by running `kubectl edit deployment/karpenter KARPENTER_NAMESPACE` & adding the values manually.
Before edit
```
- name: FEATURE_GATES                                                                                                                                                
value: ReservedCapacity=true,SpotToSpotConsolidation=false,NodeRepair=false,NodeOverlay=,StaticCapacity=
```
After edit
```
- name: FEATURE_GATES                                                                                                                                                
value: ReservedCapacity=true,SpotToSpotConsolidation=false,NodeRepair=false,NodeOverlay=false,StaticCapacity=false
```
**Expected Behavior**:
Karpenter runs with default Helm values.

Upgrading from 1.7 to 1.12 using the bellow values.yml
```
USER-SUPPLIED VALUES:
controller:
  resources:
    limits:
      cpu: 1
      memory: 1Gi
    requests:
      cpu: 1
      memory: 1Gi
serviceAccount:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::ACC_NUM:role/KARPENTER_ROLE_NAME
settings:
  clusterName: CLUSTER_NAME
  enableZonalShift: true
  interruptionQueue: CLUSTER_NAME
tolerations:
- effect: NoSchedule
  key: CriticalAddonsOnly
  operator: Exists

```
**Versions**:
- Chart Version: 1.7
```
▶ k version
Client Version: v1.36.1
Kustomize Version: v5.8.1
Server Version: v1.35.4-eks-40737a8
```

heyrobertchang:
Fixes #3073

**Description**

Karpenter panics at startup when `FEATURE_GATES` contains an empty value like `NodeOverlay=`:

```
panic: parsing feature gates, invalid value of NodeOverlay , err: strconv.ParseBool: parsing "": invalid syntax
```

Empty values happen in the wild because Helm renders unset featureGates keys as empty strings. Most common way to hit it is `helm upgrade --reuse-values`: that flag keeps the old release values and drops the new chart defaults, so gates added in newer chart versions (like `nodeOverlay` and `staticCapacity`) don't exist in values and render as empty. This is what happened to the reporter in #3073.

I also found the kwok chart in this repo hits the same panic out of the box, no `--reuse-values` needed.

This PR fixes both layers:

1. **Parser**: `ParseFeatureGates` now treats an empty value as "not set" and falls back to the gate's default. I did not want to silently swallow this, so every ignored gate prints a warning to stderr with a hint about the likely cause. Stderr because the logger is not constructed yet at options parse time. Note empty means "use the default", not "false", since `ReservedCapacity` defaults to `true`. Genuinely malformed input (like `NodeOverlay=banana` or a pair with no `=`) still fails hard, same as before, because that is a real config mistake and should stay loud.
2. **kwok chart**: added the missing `staticCapacity` and `capacityBuffer` defaults to `values.yaml` and regenerated the README with helm-docs.

I considered making the template defensive instead (`{{ ... | default false }}`) but skipped it: sprig's `default` treats `false` as empty, so for a gate that defaults to `true` it would silently flip a user's explicit `false` back to `true`.

- New `DescribeTable` cases in `pkg/operator/options/suite_test.go`: empty values fall back to defaults (including the reporter's exact string from the issue, and `ReservedCapacity=` staying `true`), and malformed strings still error.
- `helm template` on the kwok chart with default values now renders `StaticCapacity=false,...,CapacityBuffer=false` with no empty values.
- Fed both the old broken kwok render and the reporter's exact `FEATURE_GATES` string into `ParseFeatureGates`: no error, defaults applied, one warning line per ignored gate.
- `go build ./...` and `go test ./pkg/operator/...` pass.


jmdeal:
I would lean towards solving this on the Helm side of things rather than the controller. IMO this is a malformed input and we should treat it as such. @heyrobertchang have you taken a look at what it would take to prevent our helm chart from rendering these?

pkg/operator/options/options.go
```diff
@@ -182,6 +183,16 @@ func ParseFeatureGates(gateStr string) (FeatureGates, error) {
 	gateMap := map[string]bool{}
 	gates := DefaultFeatureGates()
 
+	var kept []string
+	for _, pair := range strings.Split(gateStr, ",") {
+		if k, v, found := strings.Cut(pair, "="); found && strings.TrimSpace(v) == "" {
+			fmt.Fprintf(os.Stderr, "WARNING: ignoring feature gate %q with empty value, falling back to its default; this usually means the Helm chart rendered an unset value (e.g. helm upgrade --reuse-values dropping new chart defaults)\n", strings.TrimSpace(k))
```
DerekFrank, line 189: Lets use the standard logger if we can

PR #2970, opened 2026-04-14
chore: upgrade KWOK to 0.6.0

baoalvin1:
Fixes #[2969](https://github.com/kubernetes-sigs/karpenter/issues/2969) 

**Description**
Upgrades KWOK to v0.6.0, latest KWOK 0.7.0 gave errors when upgrading directly so will make a follow up pr

**How was this change tested?**
Tested locally by running the `make install-kwok` command and everything seems to work

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3140, opened 2026-07-14
chore: upgrade KWOK to v0.8.0

### LINKED ISSUE #2969: Upgrade to latest KWOK version
The latest version of KWOK allows you to patch other fields on the Node and not just the status subresource via the patches api https://github.com/kubernetes-sigs/kwok/releases/tag/v0.6.0. It makes it harder to test things like HPAs if you can't modify the stage to add the `metrics.k8s.io/resource-metrics-path: "/metrics/nodes/<nodeName>/metrics/resource"` annotation on the node ([ref](https://kwok.sigs.k8s.io/docs/user/resource-usage-configuration/#out-of-the-box))

musaprg:
Fixes #2969 

**Description**
This PR continues the work started in #2970 and addresses the remaing task of upgrading KWOK version to the latest one ([v0.8.0](https://github.com/kubernetes-sigs/kwok/releases/tag/v0.8.0)). 

Tested by executing `make install-kwok` with my kind cluster locally.

NOTE: This PR was written in part with the assistance of generative AI

### DIFF
--- a/hack/install-kwok.sh
+++ b/hack/install-kwok.sh
@@ -111,13 +111,22 @@ EOF
 
 # Create num_partitions
 # In v0.7.0+, kwok is configured via a kwok.yaml ConfigMap instead of CLI args.
+selector_path=/options/manageNodesWithLabelSelector
 for ((i=0; i<PARTITION_COUNT; i++))
 do
   SUB_LET_DIR=$HOME_DIR/${alphabet[i]}
   mkdir ${SUB_LET_DIR}
 
-  sed "s/manageNodesWithLabelSelector: ''/manageNodesWithLabelSelector: 'kwok-partition=${alphabet[i]}'/" \
-    "${BASE}/kwok.yaml" > "${SUB_LET_DIR}/kwok.yaml"
+  partition_patch=$(cat <<EOF
+[
+  {"op":"test","path":"${selector_path}","value":""},
+  {"op":"replace","path":"${selector_path}","value":"kwok-partition=${alphabet[i]}"}
+]
+EOF
+)
+  kubectl patch --local --type=json -f "${BASE}/kwok.yaml" \
+    -p "${partition_patch}" \
+    -o yaml > "${SUB_LET_DIR}/kwok.yaml"
 
 cat <<EOF > "${SUB_LET_DIR}/kustomization.yaml"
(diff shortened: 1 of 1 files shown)


hack/install-kwok.sh
```diff
@@ -110,33 +110,32 @@ cat <<EOF > "${BASE}/kustomization.yaml"
...
   SUB_LET_DIR=$HOME_DIR/${alphabet[i]}
   mkdir ${SUB_LET_DIR}
 
-  cat <<EOF > "${SUB_LET_DIR}/patch.yaml"
-  - op: replace
-    path: /spec/template/spec/containers/0/args/2
-    value: --manage-nodes-with-label-selector=kwok-partition=${alphabet[i]}
-EOF
+  sed "s/manageNodesWithLabelSelector: ''/manageNodesWithLabelSelector: 'kwok-partition=${alphabet[i]}'/" \
+    "${BASE}/kwok.yaml" > "${SUB_LET_DIR}/kwok.yaml"
```
GaneshBannur, line 130: nit: could we add a `grep` check to make sure `sed` didn't fail. This isn't an issue in single-partition but could cause silent issues in multi-partition. Something like this right after `sed`
`grep -q "kwok-partition=${alphabet[i]}" "${SUB_LET_DIR}/kwok.yaml" || { echo "ERROR: expected 'kwok-partition=${alphabet[i]}' not found in ${SUB_LET_DIR}/kwok.yaml after sed" >&2; exit 1; }`

GaneshBannur (changes requested):

hack/install-kwok.sh
musaprg: Hi. Thank you for your suggestion. In that case, it could be worth replacing it with `kubectl patch --local`, which is more reliable way to replace specific value on `kwok.yaml`. It also supports `test` operation to check the existence of the target key. What are your thoughts on that?

musaprg: fixed in 8c8dd8c3

GaneshBannur (approved):
LGTM

musaprg:
@njtran @tallaxes @engedaam Hi team. Would you check my changes? Thanks.

musaprg:
@DerekFrank Hi. Would you check my changes? Thanks!

DerekFrank (approved):
/lgtm

PR #3142, opened 2026-07-15
feat: dra prioritized alternatives support

ryan-mist:
Fixes #N/A 

**Description**
Extends support for DRA to prioritized alternatives. For reference, the upstream KEP is [KEP-4816: DRA: Prioritized Alternatives in Device Requests](https://github.com/kubernetes/enhancements/blob/master/keps/sig-scheduling/4816-dra-prioritized-list/README.md)

- unit / integration testing; E2Es will be a follow up

### DIFF
--- a/pkg/controllers/provisioning/dra_test.go
+++ b/pkg/controllers/provisioning/dra_test.go
@@ -333,6 +333,58 @@ var _ = Describe("Dynamic Resource Allocation", func() {
 		})
 	})
 
+	Context("Prioritized alternatives / FirstAvailable (H)", func() {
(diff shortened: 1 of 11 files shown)


jmdeal:
Nice work! Test coverage looks super solid overall. I'm still taking a more thorough look through the PR, but some initial callouts:
- Let's add some smoke tests in the provisioning integration suite, there's nothing about this change that **should** impact the integration point with the rest of the scheduler, but we should add smoke tests for each feature to be safe and protect from future regressions that could result from a refactor.
- It doesn't look like we have any test coverage for topology constraints when backtracking from the FA path. Again, this shouldn't be an issue based on the current implementation, but we should add some coverage.

PR #3143, opened 2026-07-16
feat: publish pod event for insufficient capacity error

### LINKED ISSUE #2159: Emit InsufficientCapacityError events for the nominated pods to the nodeclaim
### Description

As a cluster administrator, users with specific node selectors sometimes run into insufficient capacity errors. Unfortunately, they have no way to tell this is preventing their pod from scheduling. Karpenter does not emit an event for the pod, so users have no way to detect this condition. As a result, this ends up being a re-occurring support question internally at Indeed and requires access to the Karpenter logs to debug and confirm that the user experienced insufficient capacity for their workload.

It's a frequent question from our ML users who prefer specific instance types for their workloads.

toVersus:
Fixes https://github.com/kubernetes-sigs/karpenter/issues/2159
Ref: https://github.com/aws/karpenter-provider-aws/issues/4346

**Description**

When a NodeClaim launch fails with an insufficient capacity error, Karpenter deletes the NodeClaim and logs the failure, but nothing is surfaced on the pods that were waiting for that capacity. Users constrained to specific instance types or zones (e.g. via node selectors) just see their pods stuck in Pending with no indication of why, and confirming an `InsufficientCapacityError` requires access to the Karpenter controller logs. This has been a recurring source of support escalations for cluster operators.

This change makes the launch failure visible where users actually look (i.e., on the pod). When cloud provider returns an insufficient capacity error, the nodeclaim lifecycle controller now publishes a warning event with `InsufficientCapacityError` reason on every pod that was nominated to schedule against the failed NodeClaim, alongside the existing NodeClaim event.

```
Warning  InsufficientCapacityError  NodeClaim default-xxxxx, nominated for this pod failed to launch: InsufficientInstanceCapacity: We currently do not have sufficient g5.12xlarge capacity in the Availability Zone you requested (us-west-2a). Our system will be working on provisioning additional capacity. You can currently get g5.12xlarge capacity by not specifying an Availability Zone in your request or choosing us-west-2b, us-west-2c.
```

A few notes on the implementation:

- The launch controller is the only place where the `InsufficientCapacityError` actually surfaces, but a NodeClaim doesn't carry any record of the pods it was created for. Rather than adding a new annotation or status field to the NodeClaim, this reuses the Pod→NodeClaim mapping that cluster state already maintains for scheduling latency metrics and wiring cluster state into the lifecycle controller.
- Pods that have been scheduled elsewhere, are terminal, or are terminating in the meantime are skipped, since the event would be stale or misleading for them. `CapacityBuffer` virtual pods are also skipped naturally, since they never exist in the API server.
- Events are deduped per pod UID by the event recorder, using its default 2-minute window. While insufficient capacity persists, each provisioning loop creates a NodeClaim with a new name and hits `InsufficientCapacityError` again, but since the suppression is keyed on the reason and pod UID only (not the message), a pod gets at most one event every 2 minutes rather than one per retry.

Added a test to the lifecycle suite verifying that on an `InsufficientCapacityError` launch failure the event is emitted for the nominated pending pod and not for an already-scheduled one.

### DIFF
--- a/pkg/controllers/nodeclaim/lifecycle/events.go
+++ b/pkg/controllers/nodeclaim/lifecycle/events.go
@@ -35,6 +35,16 @@ func InsufficientCapacityErrorEvent(nodeClaim *v1.NodeClaim, err error) events.E
 	}
 }
 
+func InsufficientCapacityErrorPodEvent(pod *corev1.Pod, nodeClaim *v1.NodeClaim, err error) events.Event {
+	return events.Event{
+		InvolvedObject: pod,
+		Type:           corev1.EventTypeWarning,
+		Reason:         events.InsufficientCapacityError,
+		Message:        fmt.Sprintf("NodeClaim %s nominated for this pod failed to launch: %s", nodeClaim.Name, truncateMessage(err.Error())),
+		DedupeValues:   []string{string(pod.UID)},
+	}
+}
+
 func NodeClassNotReadyEvent(nodeClaim *v1.NodeClaim, err error) events.Event {
 	return events.Event{
 		InvolvedObject: nodeClaim,
--- a/pkg/controllers/nodeclaim/lifecycle/events.go
+++ b/pkg/controllers/nodeclaim/lifecycle/events.go
@@ -30,7 +30,7 @@ func InsufficientCapacityErrorEvent(nodeClaim *v1.NodeClaim, err error) events.E
 		InvolvedObject: nodeClaim,
 		Type:           corev1.EventTypeWarning,
 		Reason:         events.InsufficientCapacityError,
-		Message:        fmt.Sprintf("NodeClaim %s event: %s", nodeClaim.Name, truncateMessage(err.Error())),
+		Message:        truncateMessage(fmt.Sprintf("NodeClaim %s event: %s", nodeClaim.Name, err.Error())),
 		DedupeValues:   []string{string(nodeClaim.UID)},
 	}
 }
@@ -40,7 +40,7 @@ func InsufficientCapacityErrorPodEvent(pod *corev1.Pod, nodeClaim *v1.NodeClaim,
 		InvolvedObject: pod,
 		Type:           corev1.EventTypeWarning,
 		Reason:         events.InsufficientCapacityError,
-		Message:        fmt.Sprintf("NodeClaim %s nominated for this pod failed to launch: %s", nodeClaim.Name, truncateMessage(err.Error())),
+		Message:        truncateMessage(fmt.Sprintf("NodeClaim %s nominated for this pod failed to launch: %s", nodeClaim.Name, err.Error())),
 		DedupeValues:   []string{string(pod.UID)},
 	}
 }
@@ -50,7 +50,7 @@ func NodeClassNotReadyEvent(nodeClaim *v1.NodeClaim, err error) events.Event {
 		InvolvedObject: nodeClaim,
 		Type:           corev1.EventTypeWarning,
 		Reason:         events.NodeClassNotReady,
-		Message:        fmt.Sprintf("NodeClaim %s event: %s", nodeClaim.Name, truncateMessage(err.Error())),
+		Message:        truncateMessage(fmt.Sprintf("NodeClaim %s event: %s", nodeClaim.Name, err.Error())),
 		DedupeValues:   []string{string(nodeClaim.UID)},
 	}
 }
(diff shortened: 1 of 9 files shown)


pkg/controllers/nodeclaim/lifecycle/events.go
```diff
@@ -35,6 +35,16 @@ func InsufficientCapacityErrorEvent(nodeClaim *v1.NodeClaim, err error) events.E
 	}
 }
 
+func InsufficientCapacityErrorPodEvent(pod *corev1.Pod, nodeClaim *v1.NodeClaim, err error) events.Event {
+	return events.Event{
+		InvolvedObject: pod,
+		Type:           corev1.EventTypeWarning,
+		Reason:         events.InsufficientCapacityError,
+		Message:        fmt.Sprintf("NodeClaim %s nominated for this pod failed to launch: %s", nodeClaim.Name, truncateMessage(err.Error())),
```
cnmcavoy: `truncateMessage` is not necessary here imo. We implemented a similar form of this for our fork of Karpenter and truncate message results in events that look like this:

> InsufficientInstanceCapacity: We currently do not have sufficient g6.2xlarge capacity in the Availability Zone you requested (ca-central-1b). Our system will be working on provisioning additional capacity. You can currently get g6.2xlarg...

The full message could have easily fit, and the truncated message hides important details.

toVersus: > The full message could have easily fit, and the truncated message hides important details.

I agree, but the message field has a 1024-byte length limit, and exceeding it causes an error when the Event is created. This might not be a problem on AWS, but it could cause issues on other cloud providers. So I think `truncateMessage` itself is necessary.
https://github.com/kubernetes/kubernetes/blob/564b0e55c7007745500d579356897848aaacb9dd/pkg/apis/core/validation/events.go#L38

That said, `truncateMessage` currently truncates the message at 300 bytes.
https://github.com/kubernetes-sigs/karpenter/blob/a0d5370441567ce139bb7c3c97fa24704e431037/pkg/controllers/nodeclaim/lifecycle/launch.go#L143-L148

I'm not sure of the rationale behind the 300-byte value, but I think raising this limit to 1024 bytes would be a more practical approach. Either way, I'd like to hear the Karpenter maintainers' opinion.

Ignoramuss: Kubernetes also truncates the message to the 1024-byte length limit, but we can keep a defense-in-depth truncate by updating the 300-byte value in Karpenter's utility as well

 https://github.com/kubernetes/kubernetes/blob/ec9448ceef13adbbcd3fa9f09b4d3c1fa54cc674/pkg/scheduler/schedule_one.go#L1267

toVersus: Thanks for the suggestion! Bumped the truncation limit from 300 to 1024 bytes. I also moved `truncateMessage` to wrap the fully-formatted message instead of just the error string, so the prefix message (e.g., `NodeClaim %s ...`) is counted toward the limit and the final message can never exceed 1024 bytes.

toVersus:
@Ignoramuss
Hi! Just a gentle ping on this one. Please let me know if there's anything else you'd like me to address.

pkg/controllers/nodeclaim/lifecycle/suite_test.go
```diff
@@ -97,6 +100,7 @@ var _ = AfterEach(func() {
 	env.Clock.SetTime(time.Now())
 	ExpectCleanedUp(ctx, env.Client)
 	cloudProvider.Reset()
+	cluster.Reset()
```
Ignoramuss: Doesn't see like Cluster.Reset actually resets the podToNodeClaim https://github.com/kubernetes-sigs/karpenter/blob/da15327e3b062cbea965a00c70df5c287d889421/pkg/controllers/state/cluster.go#L635-L656

pkg/controllers/nodeclaim/lifecycle/events.go
```diff
@@ -30,17 +30,27 @@ func InsufficientCapacityErrorEvent(nodeClaim *v1.NodeClaim, err error) events.E
...
 }
 
+func InsufficientCapacityErrorPodEvent(pod *corev1.Pod, nodeClaim *v1.NodeClaim, err error) events.Event {
+	return events.Event{
+		InvolvedObject: pod,
+		Type:           corev1.EventTypeWarning,
+		Reason:         events.InsufficientCapacityError,
+		Message:        truncateMessage(fmt.Sprintf("NodeClaim %s nominated for this pod failed to launch: %s", nodeClaim.Name, err.Error())),
+		DedupeValues:   []string{string(pod.UID)},
+	}
```
Ignoramuss: Might be a good idea to add a rate limiter here similar to pod nomination https://github.com/kubernetes-sigs/karpenter/blob/da15327e3b062cbea965a00c70df5c287d889421/pkg/controllers/provisioning/scheduling/events.go#L49

Ignoramuss:
@toVersus Thanks for the PR and your patience! I am reviewing this further and should have a full review out by the end of the week

PR #3154, opened 2026-07-18
fix: clear unbound-time metrics for pods stuck at PodScheduled=False

### LINKED ISSUE #3153: `recordPodBoundMetric` leaks unbound-time metrics for terminal pods stuck at PodScheduled=False
**Observed Behavior**:

`karpenter_pods_unbound_time_seconds` and `karpenter_pods_provisioning_unbound_time_seconds` can get stuck growing indefinitely for a pod that has actually already left `Pending` phase, if the `PodScheduled` condition never transitions to `True` before the phase changes.

This is the same root cause as the race described in #3120 (fixed there for `karpenter_pods_provisioning_scheduling_undecided_time_seconds`), but `recordPodBoundMetric` in `pkg/controllers/metrics/pod/controller.go` has an identical structural gap that PR wasn't scoped to touch:

```go
func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.Time) {
	...
	if pod.Status.Phase == corev1.PodPending {
		...
		c.unscheduledPods.Insert(key)
		return
	}
	if c.unscheduledPods.Has(key) && ok && cond.Status == corev1.ConditionTrue {
		// Delete the unbound metric since the pod is now bound
		PodUnboundTimeSeconds.Delete(...)
		PodProvisioningUnboundTimeSeconds.Delete(...)
		...
	}
	// falls through silently if phase != Pending AND cond.Status != True
}
```

If a pod leaves `Pending` (e.g. straight to `Running` or `Failed`) while `PodScheduled` is still stuck at `False` -- the exact race #3120 documents between kubelet's status patch and the scheduler's condition update -- this delete condition never evaluates true. `unscheduledPods` is only ever cleared here or in the top-level `NotFound` branch (confirmed via `grep -n "unscheduledPods" controller.go`), so nothing else resets it. The two metrics would keep reporting stale/growing values until the pod is fully deleted from the cluster, same symptom class #3120 fixes for the sibling metric.

**Expected Behavior**:

`karpenter_pods_unbound_time_seconds` and `karpenter_pods_provisioning_unbound_time_seconds` should stop reporting (or reset) once a pod is bound (`pod.Spec.NodeName != ""`) or terminal (`podutils.IsTerminal(pod)`), regardless of whether the `PodScheduled` condition ever reaches `True` -- mirroring the fix applied to `PodSchedulingUndecidedTimeSeconds` in #3120.

Reproduced against `main` with an envtest case added to `pkg/controllers/metrics/pod/suite_test.go`, modeled on the existing "should update the pod bound and unbound time metrics" test:

```go
p := test.Pod()
p.Status.Phase = corev1.PodPending
env.Clock.Step(1 * time.Hour)
cluster.MarkPodSchedulingDecisions(ctx, map[*corev1.Pod]error{}, map[string][]*corev1.Pod{"n1": {p}}, map[string][]*corev1.Pod{"nc1": {p}})
ExpectApplied(ctx, env.Client, p)
ExpectReconcileSucceeded(ctx, podController, client.ObjectKeyFromObject(p))
// -> karpenter_pods_unbound_time_seconds and karpenter_pods_provisioning_unbound_time_seconds are both emitted here (expected)

// Simulate the kubelet/scheduler race from #3120: the pod goes straight to a terminal
// phase without PodScheduled ever transitioning to True.
p.Status.Phase = corev1.PodFailed
p.Status.Conditions = []corev1.PodCondition{{Type: corev1.PodScheduled, Status: corev1.ConditionFalse, LastTransitionTime: metav1.Now()}}
ExpectApplied(ctx, env.Client, p)
ExpectReconcileSucceeded(ctx, podController, client.ObjectKeyFromObject(p))
// -> karpenter_pods_unbound_time_seconds is still present (should be deleted, per the pod's terminal phase)
```

Result: `karpenter_pods_unbound_time_seconds` is still found after this reconcile, when it should have been deleted -- confirming the metric leaks past the point the pod is terminal. Happy to open a PR with this test + fix if that's useful.

**Versions**:
- Chart Version:

Archong-Liu:
Fixes #3153

**Description**

`karpenter_pods_unbound_time_seconds` and `karpenter_pods_provisioning_unbound_time_seconds` were only ever cleared in `recordPodBoundMetric` when the `PodScheduled` condition transitioned to `True`. A kubelet/scheduler status-patch race (documented in #3120) can leave `PodScheduled` stuck at `False` even after a pod is bound or goes terminal, so these two metrics would keep growing indefinitely until the pod is fully deleted -- the same root cause #3120 fixed for the sibling `karpenter_pods_provisioning_scheduling_undecided_time_seconds` metric.

This adds the same two escape conditions #3120 used (`pod.Spec.NodeName != ""` and `podutils.IsTerminal(pod)`) to the clear path in `recordPodBoundMetric`. The bound duration histograms stay gated on `PodScheduled` actually reaching `True`, since `cond.LastTransitionTime` is the last scheduling failure rather than a bind timestamp otherwise.

Added two envtest cases mirroring the existing "should update the pod bound and unbound time metrics" test:
1. Pod goes straight from `Pending` to `Failed` with `PodScheduled` stuck at `False` -- confirmed both metrics get cleared (previously failed against unmodified code, with the metric still present after the transition).
2. Pod is bound (`NodeName` set at creation, since it's immutable afterward) and transitions to `Running` with `PodScheduled` stuck at `False` -- same result.

`go test ./pkg/controllers/metrics/pod/... -race --ginkgo.randomize-all` -- 15/15 passing.

**AI disclosure**

This PR was written in part with the assistance of generative AI. I have reviewed and verified every change myself.

### DIFF
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -393,7 +393,10 @@ func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.
 		c.unscheduledPods.Insert(key)
 		return
 	}
-	if c.unscheduledPods.Has(key) && ok && cond.Status == corev1.ConditionTrue {
+	if !c.unscheduledPods.Has(key) {
+		return
+	}
+	if ok && cond.Status == corev1.ConditionTrue {
 		// Delete the unbound metric since the pod is now bound
 		PodUnboundTimeSeconds.Delete(map[string]string{
 			podName:      pod.Name,
@@ -408,6 +411,20 @@ func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.
 		if !schedulableTime.IsZero() {
 			PodProvisioningBoundDurationSeconds.Observe(cond.LastTransitionTime.Sub(schedulableTime).Seconds(), nil)
 		}
+		c.unscheduledPods.Delete(key)
+	} else if pod.Spec.NodeName != "" || podutils.IsTerminal(pod) {
+		// The pod moved on (bound or terminal) without PodScheduled ever reaching True -- this can
+		// happen due to a kubelet/scheduler status-patch race (see #3120). Clear the unbound metric
+		// so it doesn't grow indefinitely, but skip the bound-duration observation since there's no
+		// valid PodScheduled transition timestamp to measure from.
+		PodUnboundTimeSeconds.Delete(map[string]string{
+			podName:      pod.Name,
+			podNamespace: pod.Namespace,
+		})
+		PodProvisioningUnboundTimeSeconds.Delete(map[string]string{
+			podName:      pod.Name,
+			podNamespace: pod.Namespace,
+		})
 		c.unscheduledPods.Delete(key)
 	}
 }
--- a/pkg/controllers/metrics/pod/controller.go
+++ b/pkg/controllers/metrics/pod/controller.go
@@ -370,8 +370,9 @@ func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.
 	cond, ok := lo.Find(pod.Status.Conditions, func(c corev1.PodCondition) bool {
 		return c.Type == corev1.PodScheduled
 	})
+	scheduled := ok && cond.Status == corev1.ConditionTrue
 	if pod.Status.Phase == corev1.PodPending {
-		if !ok || cond.Status != corev1.ConditionTrue {
+		if !scheduled {
 			// If the podScheduled condition does not exist, or it exists and is not set to true, we emit pod_current_unbound_time_seconds metric.
 			PodUnboundTimeSeconds.Set(time.Since(pod.CreationTimestamp.Time).Seconds(), map[string]string{
 				podName:      pod.Name,
@@ -393,8 +394,9 @@ func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.
 		c.unscheduledPods.Insert(key)
 		return
 	}
-	if c.unscheduledPods.Has(key) && ok && cond.Status == corev1.ConditionTrue {
-		// Delete the unbound metric since the pod is now bound
+	// A pod can also become bound or terminal without PodScheduled ever reaching True (status-patch race).
+	if c.unscheduledPods.Has(key) && (scheduled || pod.Spec.NodeName != "" || podutils.IsTerminal(pod)) {
+		// The pod is no longer waiting to be bound, so delete the unbound metrics
 		PodUnboundTimeSeconds.Delete(map[string]string{
 			podName:      pod.Name,
 			podNamespace: pod.Namespace,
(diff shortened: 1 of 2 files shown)


pkg/controllers/metrics/pod/controller.go
```diff
@@ -409,6 +412,20 @@ func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.
 			PodProvisioningBoundDurationSeconds.Observe(cond.LastTransitionTime.Sub(schedulableTime).Seconds(), nil)
 		}
 		c.unscheduledPods.Delete(key)
+	} else if pod.Spec.NodeName != "" || podutils.IsTerminal(pod) {
+		// The pod moved on (bound or terminal) without PodScheduled ever reaching True -- this can
```
AndrewSirenko: I liked the more succinct comment #3120 had

```
// If we've made a decision on this pod, the pod is already bound, or the pod is terminal, delete the metric idempotently and return
```

pkg/controllers/metrics/pod/controller.go
```diff
@@ -409,6 +412,20 @@ func (c *Controller) recordPodBoundMetric(pod *corev1.Pod, schedulableTime time.
...
+		// valid PodScheduled transition timestamp to measure from.
+		PodUnboundTimeSeconds.Delete(map[string]string{
+			podName:      pod.Name,
+			podNamespace: pod.Namespace,
+		})
+		PodProvisioningUnboundTimeSeconds.Delete(map[string]string{
+			podName:      pod.Name,
+			podNamespace: pod.Namespace,
+		})
+		c.unscheduledPods.Delete(key)
```
AndrewSirenko: Can we just edit line 396 to include these `pod IsTerminal` and `Node doesn't exist` in the if statement like we did in #3120? Or am I misunderstanding something?

ie 

```
if c.unscheduledPods.Has(key) && ok && (cond.Status == corev1.ConditionTrue || pod.Spec.NodeName != "" || podutils.IsTerminal(pod)) 
```

That way we don't have to repeat this PodUnboundTimeSeconds section.

AndrewSirenko (commented):
Thank you for your contribution!

Left a comment about mimicking what #3120 did instead of repeating logic.

And just in-case: If this PR was made with the assistance of an LLM, please disclose this in your PR description per [Kubernetes policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance). AI usage is allowed but must follow those guidelines. 

Thanks!

AndrewSirenko:
FYI Going to wait for #3120 to merge then review one more time, likely to propose refactoring to use a shared helper function, thank you for your patience @Archong-Liu

Archong-Liu:
> FYI Going to wait for #3120 to merge then review one more time, likely to propose refactoring to use a shared helper function, thank you for your patience @Archong-Liu

Thanks for letting me know, I would like to take more time to evaluate my changes on my end as well. Thanks for the insightful reviews, its a good experience for my first contribution to a Kubernetes project.

PR #3157, opened 2026-07-19
fix(scheduling): prioritize previously failed pods across scheduling loops

### LINKED ISSUE #3152: karpenter not scheduling new nodes for Pending pods
### Description

**Observed Behavior**:

With ~2,170 pending pods and ~1,060 active nodes across 20 NodePools, a subset of pods (light CPU-only workloads requesting 2 CPU / 8Gi memory) remained Pending for 14+ hours.  The 2,170 pending pods were pods waiting for g6.2xlarge nodes to come available (aws insufficient capacity). No karpenter events or "could not schedule" logs were emitted for these pods. NodeClaims were only created for higher-resource ML workflow pods (requesting 6 CPU / 24Gi / 1 GPU). The light CPU-only workload pods were completely invisible to operators - no indication from events or karpenter logs that karpenter that it had ever seen them.

(Note - this is all based on print debugging)

I rebuilt karpenter with additional logging in the `Solve()` loop and determined that the scheduling simulation was timing out at [provisioner.go:415](https://github.com/kubernetes-sigs/karpenter/blob/v1.14.0/pkg/controllers/provisioning/provisioner.go#L415) after 59 seconds, having evaluated only 363 of ~2,170 pods. The light CPU-only workload pods were at queue position ~
2,065 and were never reached before the timeout fired.

Because the queue sort at [queue.go:38](https://github.com/kubernetes-sigs/karpenter/blob/v1.14.0/pkg/controllers/provisioning/scheduling/queue.go#L38) is deterministic (CPU/memory descending) and stateless across loops, the same pods were always evaluated first and the same pods were always starved on every subsequent reconciliation.

I suspect the low throughput (~6 pods/second) is related to:
- karpenter controller CPU requests set to 2 cores (i should probably increase this i guess or setup VPA), giving `numConcurrentReconciles=2` via [provisioner.go:390](https://github.com/kubernetes-sigs/karpenter/blob/v1.14.0/pkg/controllers/provisioning/provisioner.go#L390)
  - this caps parallelism for `addToExistingNode` (scanning ~1,060 nodes), `addToInflightNode`, and `addToNewNodeClaim` (scanning 20 templates)
- All ~1,700 ML workflow pods had `topologySpreadConstraints` with `whenUnsatisfiable: ScheduleAnyway`, which caused each pod to fail on first attempt, get relaxed by `preferences.Relax()`, re-queue, and be evaluated a second time

I'm not certain these are the only contributing factors.

**Expected Behavior**:

All pending pods should eventually be evaluated by the scheduler regardless of queue depth.

Reproduction requires a large active cluster. Key conditions:
- ~2,000+ pending pods with heterogeneous resource requests (mostly gpu requests with aws insufficient capacity, just a few small cpu requests pending later)
- 20 NodePool templates
- ~1,000+ active nodes
- Controller CPU requests set to 2 cores 
- Pending pods with `topologySpreadConstraints` (`whenUnsatisfiable: ScheduleAnyway`)

Pod spec for the starved light CPU-only workload pods (never evaluated):
```yaml
spec:
  containers:
  - resources:
      requests:
        cpu: "2"
        ephemeral-storage: 50Gi
        memory: 8Gi
      limits:
        cpu: "4"
        ephemeral-storage: 115Gi
        memory: 16Gi
  nodeSelector:
    thingy: prod-worker
    kubernetes.io/arch: amd64
  tolerations:
  - effect: NoSchedule
    key: buildkit
    operator: Exists
  - effect: NoSchedule
    key: prod-thingy-worker
    operator: Exists
```

Pod spec for the ML workflow pods that consumed all scheduling budget:
```yaml
spec:
  containers:
  - resources:
      requests:
        cpu: "6"
        memory: "24Gi"
        nvidia.com/gpu: "1"
  nodeSelector:
    thingy: prod-worker
    kubernetes.io/arch: amd64
  tolerations:
  - effect: NoSchedule
    key: nvidia.com/gpu
    operator: Exists
  - effect: NoSchedule
    key: prod-thingy-worker
    operator: Exists
  topologySpreadConstraints:
  - maxSkew: 1
    whenUnsatisfiable: ScheduleAnyway
    labelSelector:
      matchLabels:
        app.kubernetes.io/managed-by: thingy
```

let me know if you need more context on the pod yamls.

**Fix**:

https://github.com/kubernetes-sigs/karpenter/pull/3145 is a proposed fix but I'm unsure if it is the most optimal solution.

**Versions**:
- Chart Version: 1.14.0
- Kubernetes Version: v1.35 (EKS)

Debasish-87:
## Summary

Fixes scheduling starvation caused by deterministic queue ordering across reconciliation loops.

Previously, `Solve()` rebuilt the scheduling queue on every reconciliation and sorted pods deterministically by CPU and memory requests. When scheduling timed out before processing the full queue, the same high-resource pods were evaluated first on the next reconciliation, causing lower-priority pods near the tail of the queue to starve indefinitely.

This change introduces a shared `FairnessState` that tracks unsuccessful scheduling attempts across scheduling loops. Queue ordering is updated to prioritize pods with more previous scheduling failures before falling back to the existing CPU/memory ordering.

## Changes

- Add `FairnessState` to track scheduling retry attempts.
- Share fairness state across scheduler instances through the `Provisioner`.
- Prioritize pods with more previous scheduling failures during queue construction.
- Clear retry history once a pod is successfully scheduled.
- Update scheduler tests and benchmarks for the new constructor.

## Testing

```bash
go test ./pkg/controllers/provisioning/...
go test ./pkg/controllers/provisioning/scheduling/...
go test ./...
```

The package and scheduling tests pass locally.

The remaining failures under `test/suites/*` require an external Kubernetes/EKS integration environment and are unrelated to this change.

Fixes #3152.


Debasish-87:
Hi @DerekFrank and @tallaxes,

I hope you're both doing well.

Whenever you have a chance, could you please review this PR? The latest changes have been pushed, and all presubmit checks are now passing.

This PR addresses #3152 by introducing scheduling fairness across provisioning loops, ensuring that pods which repeatedly fail scheduling are prioritized in subsequent scheduling attempts to prevent starvation.

I'd appreciate any feedback or suggestions. Thank you!

Debasish-87:
Hi @DerekFrank and @tallaxes, just following up on this PR. When you have a chance, could you please take another look? All presubmit checks are passing, and I'd really appreciate any feedback or suggestions on the implementation. Thank you!

nathangeology:
QQ, will this fairness backfire if it ends up prioritizing pods that will just keep failing every time we retry? Maybe we want to inject fairness via randomizing the sort a bit instead?

nathangeology:
Also, looking at the code more and thinking this through, it looks like a pod still has to make it to the 'solve' portion of simulation to get prioritized which could cause the original issue to get worse (i.e. pods that don't make it to solve never even have a chance to prioritized). Also, there may be some DRA things that need to be considered here with scheduling sim 'fairness' that I need to think about some more.

Debasish-87:
Thanks for taking a closer look — these are both good points.

You're right that the current approach is really tracking unsuccessful scheduling attempts, not pods that never make it through scheduling simulation. In the timeout scenario from #3152, pods that remain at the tail of the queue would never accumulate fairness, so this doesn't fully address the original starvation case.

Your point about repeatedly prioritizing pods that are inherently unschedulable is also well taken. I hadn't considered that this retry-based approach could end up biasing the scheduler toward pods that will continue to fail, rather than improving overall fairness.

I'll revisit the design with those concerns in mind. I want to explore an approach that specifically addresses starvation caused by reconciliation timeouts without continuously boosting pods that are simply unschedulable. I'll also take another look at the DRA implications before updating the PR.

Thanks for the thoughtful review.

thc1006:
Late to this, and I would rather not re-tread what @nathangeology already raised, since the two of you landed on the same conclusion and a redesign.

One thing I did not find in the thread that should outlive whichever shape you pick. `FairnessState` only drops an entry on the success branch, `scheduler.go:494`, right after the `Increment` at `:490`, and those are its only two call sites. A pod that fails a round and is then deleted, goes terminal, or gets taken over by another controller never reaches either one, so its UID stays in the map for the life of the Provisioner. Pod UIDs are not reused, so on a cluster with any churn that map only grows.

Whatever the signal ends up being, deferred rounds, age since the last attempt, or a queue cursor, it will still be keyed per pod and will want the same treatment: reconcile it against the pods actually pending at the start of each round, or give entries a TTL. A test that runs many rounds where pods fail once and then disappear, asserting the state does not grow without bound, would hold that regardless of which design you land on.

Happy to take another look when you push the redesign.

This comment was written in part with the assistance of generative AI. Bcuz my English might be sucks.

DerekFrank:
Hey folks!

This kind of change has a lot of potential side-effect interactions with provisioning performance (bin-packing and latency), as well as a couple different viable alternatives. I think this will require a short design doc outlining what the failure modes are, the options for solving them, and the tradeoffs between them.

@Debasish-87 I think you identified in the issue that there are a couple options, the next step is an RFC to discuss them. 

Closing this PR and marking the issue as needs-design

/close

PR #3158, opened 2026-07-20
test: make ExpectApplied retry on conflict

ketanjani21:
`ExpectApplied` does `c.Get()` then `c.Update()` (and separately `c.Status().Update()`) with the fresh resource version set on the passed object. Under heavy parallelism, specifically `pkg/controllers/nodeclaim/garbagecollection/suite_test.go:153` where `workqueue.ParallelizeUntil` spins 100 goroutines each calling `ExpectNodeClaimDeployed` (which calls `ExpectApplied`), the controller-runtime informer can reconcile the same object between our `Get` and `Update`. The `Update` then returns `Operation cannot be fulfilled ... the object has been modified; please apply your changes to the latest version and try again` (409 Conflict), and the test fails.

Wrap both `Update` and `Status().Update()` in `retry.RetryOnConflict(retry.DefaultBackoff, ...)`. Each retry re-fetches the current object to pick up the fresh resource version. Behavior for non-conflict errors (including the "server could not find the requested resource" case for objects without a status subresource) is unchanged.

The initial version of this PR targeted `ExpectFinalizersRemoved`, which uses `client.Patch(ctx, obj, client.MergeFrom(stored))`. Thanks to @GaneshBannur for catching that `MergeFrom` skips the resource version and cannot produce the 409 we saw. Digging into the actual CI log confirmed the failure was in `ExpectNodeClaimDeployed → ExpectApplied → c.Update()`, not the finalizers helper.

- `go build ./...` and `go vet ./pkg/test/...` clean.
- Validation of the fix in CI once this PR is up.

### DIFF
--- a/pkg/test/expectations/expectations.go
+++ b/pkg/test/expectations/expectations.go
@@ -49,6 +49,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/util/retry"
 	"k8s.io/utils/clock"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
@@ -302,11 +303,7 @@ func ExpectFinalizersRemovedFromList(ctx context.Context, c client.Client, objec
 	for _, list := range objectLists {
 		Expect(c.List(ctx, list)).To(Succeed())
 		Expect(meta.EachListItem(list, func(o runtime.Object) error {
-			obj := o.(client.Object)
-			stored := obj.DeepCopyObject().(client.Object)
-			obj.SetFinalizers([]string{})
-			Expect(client.IgnoreNotFound(c.Patch(ctx, obj, client.MergeFrom(stored)))).To(Succeed())
-			return nil
+			return client.IgnoreNotFound(removeFinalizersWithRetry(ctx, c, o.(client.Object)))
 		})).To(Succeed())
 	}
 }
@@ -314,11 +311,22 @@ func ExpectFinalizersRemovedFromList(ctx context.Context, c client.Client, objec
 func ExpectFinalizersRemoved(ctx context.Context, c client.Client, objs ...client.Object) {
 	GinkgoHelper()
 	for _, obj := range objs {
-		Expect(client.IgnoreNotFound(c.Get(ctx, client.ObjectKeyFromObject(obj), obj))).To(Succeed())
+		Expect(client.IgnoreNotFound(removeFinalizersWithRetry(ctx, c, obj))).To(Succeed())
+	}
+}
+
+// removeFinalizersWithRetry clears finalizers on obj, re-fetching and retrying
+// on conflict. This makes the helper safe to call concurrently with controller
+// reconciles that may be updating the same object.
+func removeFinalizersWithRetry(ctx context.Context, c client.Client, obj client.Object) error {
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+		if err := c.Get(ctx, client.ObjectKeyFromObject(obj), obj); err != nil {
+			return err
+		}
 		stored := obj.DeepCopyObject().(client.Object)
 		obj.SetFinalizers([]string{})
-		Expect(client.IgnoreNotFound(c.Patch(ctx, obj, client.MergeFrom(stored)))).To(Succeed())
-	}
+		return c.Patch(ctx, obj, client.MergeFrom(stored))
+	})
 }
 
 func ExpectProvisioned(ctx context.Context, c client.Client, cluster *state.Cluster, cloudProvider cloudprovider.CloudProvider, provisioner *provisioning.Provisioner, pods ...*corev1.Pod) ProvisioningResult {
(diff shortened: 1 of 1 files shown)


ketanjani21:
@GaneshBannur  - gentle check in, did you get a chance to review?

GaneshBannur:
Thanks for the contribution!

> controller reconcile can mutate the same object between the helper's Get and Patch, producing a 409 conflict that the helper propagates as a test failure ("Operation cannot be fulfilled ... the object has been modified"). This surfaces as an intermittent flake in the presubmit matrix

As far as I understand, the `MergeFrom` that we currently use to patch objects shouldn't return a conflict when the object has been modified since it skips this `if` block and doesn't include the resource version in the patch:
https://github.com/kubernetes-sigs/controller-runtime/blob/07d8760faa66433ba3aa64c8929e77665616c256/pkg/client/patch.go#L106-L117

Would you be able to point me to the presubmit run where this was flaking and where you saw the `object has been modified` error. That would help pin down what needs to be fixed.

ketanjani21:
The test was failing (flaky) on my earlier PR: https://github.com/kubernetes-sigs/karpenter/pull/3063



Got the log now: https://github.com/kubernetes-sigs/karpenter/actions/runs/29625923514/job/88030196254

The failure is at `pkg/controllers/nodeclaim/garbagecollection/suite_test.go:153`, not inside `ExpectFinalizersRemoved`. The failing line is:

```go
nodeClaims[i], node, err = ExpectNodeClaimDeployed(ctx, env.Client, cloudProvider, nodeClaims[i])
```

Running inside a `workqueue.ParallelizeUntil` block that spins 100 goroutines. `ExpectNodeClaimDeployed` calls `ExpectApplied`, which does `object.SetResourceVersion(current.GetResourceVersion())` and then `c.Update(ctx, object)`. Under that parallelism the controller informer reconciles the same NodeClaim between the Get and the Update, so the Update hits Conflict/409.

You are correct,  `MergeFrom` in `ExpectFinalizersRemoved` isn't the source, and the previous version of this PR targeted the wrong file. The retryable path is `ExpectApplied`'s Update, plus its `Status().Update()`.

Repushed the branch with the corrected fix. Reverted the `ExpectFinalizersRemoved` changes and wrapped both Update calls in `ExpectApplied` with `retry.RetryOnConflict`. PR title and body updated to match.

Thanks for pushing back on this. Would have shipped a fix that missed the actual bug otherwise.

pkg/test/expectations/expectations.go
```diff
@@ -152,20 +153,30 @@ func ExpectApplied(ctx context.Context, c client.Client, objects ...client.Objec
...
 
-		// Create or Update
-		if err := c.Get(ctx, client.ObjectKeyFromObject(current), current); err != nil {
-			if errors.IsNotFound(err) {
-				Expect(c.Create(ctx, object)).To(Succeed())
-			} else {
-				Expect(err).ToNot(HaveOccurred())
+		// Create or Update, retrying on conflict. Under parallel test workloads
+		// (e.g. workqueue.ParallelizeUntil), the controller informer can reconcile
+		// the same object between our Get and Update, so the Update returns 409.
```
GaneshBannur, line 158:
```suggestion
		// (e.g. workqueue.ParallelizeUntil), the object's resource version in the
		// informer cache can be stale so the Update returns 409.
```
Suggesting this change since I think in the failing line
https://github.com/kubernetes-sigs/karpenter/blob/da15327e3b062cbea965a00c70df5c287d889421/pkg/controllers/nodeclaim/garbagecollection/suite_test.go#L153
each goroutine operates on its own (i<sup>th</sup>) nodeclaim and as far as I can see there isn't a controller reconciling them in the background?
AFAIK what's happening isn't that the object is being updated concurrently but that a previous update hasn't propagated to the informer cache.

pkg/test/expectations/expectations.go
GaneshBannur, line 250: We have the same pattern here, could we make the same fix. `ExpectApplied` calls this function so makes sense to fix it in this PR.
https://github.com/kubernetes-sigs/karpenter/blob/da15327e3b062cbea965a00c70df5c287d889421/pkg/test/expectations/expectations.go#L175

GaneshBannur (changes requested):

GaneshBannur:
Not blocking for this PR, just putting it here for documentation. But we could think about replacing the cache syncing client we have
https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/test/cachesyncingclient.go
with
https://github.com/kubernetes-sigs/controller-runtime/issues/3320
Looks like that would have avoided this flake.

pkg/test/expectations/expectations.go
ketanjani21: updated the PR based on your feedback

pkg/test/expectations/expectations.go
ketanjani21: updated the PR based on your feedback

ketanjani21:
> Not blocking for this PR, just putting it here for documentation. But we could think about replacing the cache syncing client we have https://github.com/kubernetes-sigs/karpenter/blob/main/pkg/test/cachesyncingclient.go with [kubernetes-sigs/controller-runtime#3320](https://github.com/kubernetes-sigs/controller-runtime/issues/3320) Looks like that would have avoided this flake.

Thanks for the pointer. Noted for follow-up. Please help review the PR again.

GaneshBannur (approved):
/lgtm

PR #3159, opened 2026-07-20
fix: tolerate missing PersistentVolume in driverFromVolume to prevent cluster sync paralysis

### LINKED ISSUE #3169: Missing PersistentVolume in driverFromVolume causes permanent cluster sync failure after controller restart
**Observed Behavior**:

When a `PersistentVolume` object is deleted out-of-band while a `PersistentVolumeClaim` still references it (PVC enters `Lost` phase, `spec.volumeName` still points to the now-missing PV), Karpenter's node-state reconciler fails on every reconcile for any node that hosts a pod whose PVC references that PV.

The failure originates in `driverFromVolume` (`pkg/scheduling/volumeusage.go:170-181`), which calls `kubeClient.Get(ctx, {Name: volumeName}, &pv)` and returns the raw error — including `apierrors.IsNotFound` — to its caller. The `NotFound` bubbles up through `ResolveDriver` → `GetVolumes` → `updateForPod` → `NodeController.Reconcile` and is logged every ~30s as:

```
tracking volume usage, PersistentVolume "<name>" not found
```

**Steady-state (before controller restart) — non-fatal:**
`Cluster.Synced()` at `pkg/controllers/state/cluster.go:167` short-circuits when `c.hasSynced.Load() == true`, and `hasSynced` is a one-way atomic latch (`Store(true)` at line 226 on first successful sync, never reset). So the error is logged continuously but Karpenter keeps provisioning, consolidating, and disrupting normally.

**After a controller restart — fatal and permanent:**
The new pod starts with `hasSynced == false`. `Cluster.Synced()` now runs the full superset check at line 224 (`stateNodeClaimNames.IsSuperset(nodeClaimNames) && stateNodeNames.IsSuperset(nodeNames)`). The affected node's `updateForPod` returns the `NotFound` error before the node can be registered in `nodeNameToProviderID`, so the state map is a strict subset of what the API server exposes → superset check fails → `hasSynced` never latches → `karpenter_cluster_state_synced` pins at `0` **forever**.

While the metric is at `0`, all higher-level Karpenter controllers gate on `Cluster.Synced()` and no-op — provisioning stops, consolidation stops, drift response stops, and disruption stops. A single orphaned PV can therefore permanently halt Karpenter cluster-wide until the missing PV is recreated (usually via `kubectl delete pvc` → StatefulSet re-provisions).

**Expected Behavior**:

A missing `PersistentVolume` should be tolerated the same way a missing `PersistentVolumeClaim` already is — logged and skipped, not bubbled up as a hard reconcile failure.

`GetVolumes` at `pkg/scheduling/volumeusage.go:88-94` already tolerates `IsNotFound` for a missing PVC, with an explicit comment:

```go
// If the PVC is not found it was manually deleted and its finalizer removed. We should ignore this volume when
// computing limits, otherwise Karpenter may never be able to update its cluster state.
```

The same reasoning applies to missing PVs: from Karpenter's perspective a deleted PV is not actionable, the volume simply cannot be tracked for limits, and the node/pod should still be reconcilable.

1. Deploy any workload that uses a dynamically-provisioned PVC bound to a PV. Any StatefulSet with `volumeClaimTemplates` works — e.g., a 3-replica ZooKeeper StatefulSet on `gp3`-backed EBS.
2. Wait for the PVC to reach `Bound` and the pod to reach `Running`.
3. From the outside, delete the underlying PV while the pod is still running:
   ```
   kubectl delete pv <bound-pv-name>            # 200 OK, but blocked by kubernetes.io/pv-protection finalizer
   kubectl patch pv <bound-pv-name> --type=json -p='[{"op":"remove","path":"/metadata/finalizers"}]'
   ```
   The PVC transitions to `Lost`. The pod keeps running on its existing volume-attachment.
4. Observe Karpenter controller logs — every ~30 seconds:
   ```
   E<ts> tracking volume usage, PersistentVolume "<pv-name>" not found
   ```
   `karpenter_cluster_state_synced` remains `1` (steady-state, `hasSynced` still `true`).
5. Restart the Karpenter controller pod:
   ```
   kubectl -n karpenter delete pod -l app.kubernetes.io/name=karpenter
   ```
6. Observe: `karpenter_cluster_state_synced` drops to `0` and stays there indefinitely. Logs show:
   ```
   waiting on cluster sync ... duration=30s
   waiting on cluster sync ... duration=1m40s
   waiting on cluster sync ... duration=5m
   ... (grows unbounded)
   ```
   All provisioning / consolidation / disruption halts.
7. Recovery only when someone manually deletes the `Lost` PVC → StatefulSet controller recreates PVC → CSI dynamic-provisions a fresh PV → node registers → sync succeeds → `hasSynced` finally latches.

**Real-world impact:**

This exact scenario caused a **~22-hour production outage** for us in an EKS `1.33` cluster:
- Day 0: an SRE deleted a `Bound` PV as manual cleanup after a node went `NotReady` (the correct fix would have been `kubectl delete node`, to trigger EBS force-detach).
- Days 0-13: latent state. Karpenter logs `NotFound` every 30s but keeps functioning.
- Day 13: the Karpenter controller pod was manually restarted (unrelated investigation of the recurring log spam).
- After restart: `karpenter_cluster_state_synced == 0` for the next 22 hours. All node provisioning halted cluster-wide. The cluster was effectively frozen until the orphaned PVC was manually deleted and the StatefulSet re-provisioned a fresh PV.

The failure mode is disproportionate to the trigger: **one orphaned PV → cluster-wide, indefinite halt of all Karpenter operations after any restart.**

**Versions**:
- Chart Version: karpenter-provider-aws (any version tracking `main` — the code paths involved have been stable since the current `Cluster.Synced()` implementation landed)
- Kubernetes Version (`kubectl version`): observed on EKS 1.33, but the bug is version-independent since the failure is in Karpenter's PV Get call semantics, not any specific K8s API behavior
- Karpenter source verified against `main` at commit `a0d5370` on 2026-07-20 — the affected code (`pkg/scheduling/volumeusage.go:170-181` and `pkg/controllers/state/cluster.go:167`, `:226`) is unchanged relative to that reference.

---

### Proposed fix

Add `apierrors.IsNotFound(err)` tolerance to `driverFromVolume`, matching the existing tolerance for missing PVCs at `GetVolumes` line 88-94:

```go
func driverFromVolume(ctx context.Context, kubeClient client.Client, volumeName string) (string, error) {
    var pv v1.PersistentVolume
    if err := kubeClient.Get(ctx, client.ObjectKey{Name: volumeName}, &pv); err != nil {
        if errors.IsNotFound(err) {
            log.FromContext(ctx).WithValues("PersistentVolume", volumeName).V(1).
                Info("failed tracking CSI volume limits, PersistentVolume not found; ignoring volume")
            return "", nil
        }
        return "", err
    }
    // ... (existing CSI / AWSElasticBlockStore handling unchanged)
}
```

Effect: a missing PV is logged at V(1) and skipped, `driverFromVolume` returns `("", nil)`, `ResolveDriver` returns `""`, `GetVolumes` skips the volume in its `Add` call (existing behavior when `driverName == ""` at line 106), the node registers normally, `Cluster.Synced()` can latch, and Karpenter continues operating.

A PR implementing this fix (with unit tests covering: missing PV tolerated, existing PV correctly resolves the CSI driver, missing PVC tolerated as regression guard) is open at #3159.

/kind bug
/sig autoscaling

omerd2:
Fixes #3169

## Description

Fixes an issue where `driverFromVolume` returns a raw `NotFound` error when querying a missing `PersistentVolume` (e.g., when a PV is deleted out-of-band while a PVC remains in `Lost` phase).

Without this fix, a missing PV causes node state reconciliation to fail repeatedly. Upon a Karpenter controller restart, nodes failing reconciliation are excluded from the internal cluster map, preventing `Cluster.Synced()` from succeeding and pinning `karpenter_cluster_state_synced` at `0` indefinitely — halting all provisioning, consolidation, and disruption operations.

This PR adds `apierrors.IsNotFound(err)` handling to `driverFromVolume` (`pkg/scheduling/volumeusage.go:170`), aligning it with the existing tolerant behavior used for missing PVCs in `GetVolumes()` (line 88-94). That existing code already has a comment explaining why tolerance is needed:

> "If the PVC is not found it was manually deleted and its finalizer removed. We should ignore this volume when computing limits, otherwise Karpenter may never be able to update its cluster state."

The same logic applies to missing PVs — a deleted PV should not permanently block cluster state synchronization.

### Real-world incident

This exact scenario caused a 22-hour outage in a production-grade EKS cluster:
1. An SRE deleted a PV while its PVC was still `Bound` (PVC went to `Lost` phase)
2. Karpenter logged `"tracking volume usage, PersistentVolume not found"` every ~30s but continued operating (because `hasSynced` was already `true`)
3. 13 days later, the Karpenter controller pod was manually restarted
4. After restart, `hasSynced` reset to `false`, the full superset check ran, the affected node couldn't complete reconciliation, and `karpenter_cluster_state_synced` pinned at `0` permanently
5. All provisioning, consolidation, and disruption halted until the Lost PVC was manually deleted

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

- Added unit tests in `pkg/scheduling/volumeusage_test.go` covering:
  - Pod → PVC → missing PV scenario (should return no error, empty volumes)
  - Pod → PVC → existing PV scenario (should return the CSI driver correctly)
  - Pod → missing PVC scenario (existing tolerance, regression test)

## Does this PR introduce a user-facing change?

```release-note
Fixed an issue where a missing PersistentVolume (e.g., manually deleted while PVC was Bound) could permanently prevent Karpenter cluster state from syncing after a controller restart, halting all provisioning operations.
```

### DIFF
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -170,6 +170,10 @@ func driverFromSC(ctx context.Context, kubeClient client.Client, storageClassNam
 func driverFromVolume(ctx context.Context, kubeClient client.Client, volumeName string) (string, error) {
 	var pv v1.PersistentVolume
 	if err := kubeClient.Get(ctx, client.ObjectKey{Name: volumeName}, &pv); err != nil {
+		if errors.IsNotFound(err) {
+			log.FromContext(ctx).WithValues("PersistentVolume", volumeName).V(1).Info("failed tracking CSI volume limits, PersistentVolume not found; ignoring volume")
+			return "", nil
+		}
 		return "", err
 	}
 	if pv.Spec.CSI != nil {
--- a/pkg/scheduling/volumeusage.go
+++ b/pkg/scheduling/volumeusage.go
@@ -176,6 +176,10 @@ func driverFromSC(ctx context.Context, kubeClient client.Client, storageClassNam
 func driverFromVolume(ctx context.Context, kubeClient client.Client, volumeName string) (string, error) {
 	var pv v1.PersistentVolume
 	if err := kubeClient.Get(ctx, client.ObjectKey{Name: volumeName}, &pv); err != nil {
+		if errors.IsNotFound(err) {
+			log.FromContext(ctx).WithValues("PersistentVolume", volumeName).V(1).Info("failed tracking CSI volume limits, PersistentVolume not found; ignoring volume")
+			return "", nil
+		}
 		return "", err
 	}
 	if pv.Spec.CSI != nil {
--- a/pkg/scheduling/volumeusage_test.go
+++ b/pkg/scheduling/volumeusage_test.go
@@ -1,0 +1,113 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduling
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	fakecr "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("VolumeUsage", func() {
+	var (
+		ctx    context.Context
+		scheme *runtime.Scheme
+	)
+
+	BeforeEach(func() {
+		ctx = context.Background()
+		scheme = runtime.NewScheme()
+		Expect(v1.AddToScheme(scheme)).To(Succeed())
+		Expect(storagev1.AddToScheme(scheme)).To(Succeed())
+	})
+
+	Describe("GetVolumes", func() {
+		newPod := func(claimName string) *v1.Pod {
+			return &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-pod", Namespace: "default"},
+				Spec: v1.PodSpec{
+					Volumes: []v1.Volume{{
+						Name: "data",
+						VolumeSource: v1.VolumeSource{
+							PersistentVolumeClaim: &v1.PersistentVolumeClaimVolumeSource{ClaimName: claimName},
+						},
+					}},
+				},
+			}
+		}
+
+		It("should tolerate a missing PersistentVolume and not return an error", func() {
+			// PVC references a PV that has been deleted from the cluster (e.g. manually).
+			// Without this tolerance, one missing PV would fail node-state reconciliation
+			// and could block cluster sync after a controller restart.
+			pvc := &v1.PersistentVolumeClaim{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-pvc", Namespace: "default"},
+				Spec: v1.PersistentVolumeClaimSpec{
+					VolumeName:       "test-pv",
+					StorageClassName: strPtr("test-storage-class"),
+				},
+			}
+			kubeClient := fakecr.NewClientBuilder().WithScheme(scheme).WithObjects(pvc).Build()
+
+			volumes, err := GetVolumes(ctx, kubeClient, newPod("test-pvc"))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(volumes).To(BeEmpty())
+		})
+
+		It("should return volumes correctly when the PersistentVolume exists", func() {
+			pv := &v1.PersistentVolume{
+				ObjectMeta: metav1.ObjectMeta{Name: "test-pv"},
+				Spec: v1.PersistentVolumeSpec{
+					PersistentVolumeSource: v1.PersistentVolumeSource{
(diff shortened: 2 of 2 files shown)


omerd2:
Rebased onto latest `main` (`11c782f`) — the conflict is resolved and CI is green.

The conflict was an `add/add` on `pkg/scheduling/volumeusage_test.go`: #3202 added a file of the same name after this PR was opened. I merged both test sets into one file — the fallback-limit spec from #3202 and the three `GetVolumes` specs here — and switched a local `strPtr` helper to `lo.ToPtr` to match the rest of the repo. The fix itself in `volumeusage.go` is unchanged (4 lines) and merged cleanly.

@jigisha620 you reproduced this in #3169 and confirmed `karpenter_cluster_state_synced` pins at 0 with all operations halted. This is ready for review whenever you have a chance. The issue is also still labeled `needs-priority` if you're able to set one.

semx (commented):
Verified the chain on main and the scope of the change.

- `driverFromVolume` returns the raw NotFound (`volumeusage.go:176-180`); `ResolveDriver` (`:126`) and `GetVolumes` (`:107`) propagate it; `StateNode.updateForPod` wraps it as "tracking volume usage" (`statenode.go:466`); `newStateFromNode` returns `nil, err` (`cluster.go:770-775`), so the node is never inserted and the first-sync superset check in `Synced()` cannot pass. Matches the report.
- The other `ResolveDriver` caller is unaffected: `ValidatePersistentVolumeClaims` rejects `ClaimLost` (`volumetopology.go:243`) and a missing PV via `validateVolume` (`:247, :294-297`) before reaching `ResolveDriver` (`:280`). So this only changes volume-count tracking, and a volume whose PV is gone cannot be attached anyway.
- New spec fails on main (`persistentvolumes "test-pv" not found`) and passes here; `pkg/scheduling` and `pkg/controllers/state` are green with `-race`.

Two nits:

1. Log key: the file uses Kind keys with klog helpers (`klog.KObj(pvc)`, `klog.KRef("", storageClassName)` at `:143, :154`), and AGENTS.md asks for the same. `"PersistentVolume", volumeName` should be `"PersistentVolume", klog.KRef("", volumeName)`.
2. The PVC-not-found sibling at `:98` logs at Error while this logs at V(1) like the StorageClass path at `:154`. Either is defensible; a sentence in the description saying which sibling you followed and why would pre-empt the question.

PR #3165, opened 2026-07-21
test: add per-test-case performance threshold overrides

AndrewMitchell25:
Fixes #N/A

**Description**

Adds a per-test-case override mechanism for the performance suite thresholds so providers whose controller footprint differs from upstream can tune individual thresholds without patching the tests.

Each assertion now declares its base threshold inline (the upstream default) and looks it up through a keyed helper (e.g. `MemoryThreshold("basic/scaleOut", 260)`). Overrides are supplied as JSON in the `KARPENTER_PERF_THRESHOLDS` environment variable, keyed by `"<testCase>/<phase>"`:

```json
{"basic/scaleOut": {"memory_mb": 400, "cpu_cores": 1.2, "total_time_minutes": 3, "cpu_util": 0.30, "memory_util": 0.35}}
```

An override wins for the metric it sets; any metric left unset falls back to the inline base default, so a standard run needs no configuration and keeps full regression sensitivity. Five metrics are configurable: peak memory (MB), average CPU (cores), total time (minutes), reserved CPU utilization, and reserved memory utilization. Correctness invariants (pod counts, net change, protected nodes, test type) are intentionally left fixed.

The override JSON is parsed strictly (unknown metric fields are rejected) and malformed input fails the suite loudly rather than silently falling back to defaults.

This replaces the two blunt global overhead env vars (`KARPENTER_MEMORY_OVERHEAD_MB` / `KARPENTER_CPU_OVERHEAD_CORES`), which are removed along with `overhead.go`.

`go vet ./test/suites/performance/` and `go test -c ./test/suites/performance/` pass. The override parsing logic (no config, per-metric override, base fallback for unset metrics, malformed-JSON panic, and unknown-field rejection) was exercised with a targeted `go test -run` that skips the cluster-backed integration suite.


DerekFrank (approved):
/lgtm

PR #3170, opened 2026-07-23
fix: relax minValues against narrowed NodeClaim requirements

### LINKED ISSUE #2999: BestEffort minValuesPolicy does not relax minValues when pod volume topology narrows zone requirement
### Description

**Observed Behavior:**
When `MIN_VALUES_POLICY=BestEffort` is set and a NodePool has `minValues: 2` on the `topology.kubernetes.io/zone` requirement, NodeClaim creation fails with CRD CEL validation error for pods with PVC zone affinity. The `BestEffort` relaxation does not trigger.

```
NodeClaim.karpenter.sh "monitoring-v8xgc" is invalid: spec.requirements: Invalid value: "array": 
requirements with 'minValues' must have at least that many values specified in the 'values' field
```

**Expected Behavior:**
With `MIN_VALUES_POLICY=BestEffort`, Karpenter should relax `minValues: 2` to `minValues: 1` when the pod's volume topology narrows the zone to a single value, and the NodeClaim should be created successfully.

### Root Cause Analysis

The issue is in `SatisfiesMinValues` (`pkg/cloudprovider/types.go`). This function checks whether the **instance type offerings** satisfy the minValues requirement, but it does not account for the **resolved NodeClaim requirements** which may have been narrowed by pod volume topology.

**Code flow in `CanAdd` (`pkg/controllers/provisioning/scheduling/nodeclaim.go`):**

1. `nodeClaimRequirements` starts with NodePool requirements: `zone=[eu-west-1b, eu-west-1c]`, `minValues=2`
2. `nodeClaimRequirements.Add(podData.Requirements.Values()...)` adds pod requirements including volume topology `zone=[eu-west-1b]`. After intersection: `zone=[eu-west-1b]` (narrowed to 1 value)
3. `filterInstanceTypesByRequirements` is called with the narrowed `nodeClaimRequirements`
4. Inside `filterInstanceTypesByRequirements`, `remaining.SatisfiesMinValues(requirements)` is called
5. `SatisfiesMinValues` iterates instance types and collects `it.Requirements.Get(req.Key).Values()` — this returns the **instance type's own zone values** (e.g., `[eu-west-1b, eu-west-1c]`), NOT the intersected values
6. Since the instance type offers 2 zones, `SatisfiesMinValues` reports "minValues=2 is satisfied" → `incompatibleKeys` is empty → no error returned
7. Back in `CanAdd`, `relaxMinValues` block only runs when `unsatisfiableKeys` is non-empty → relaxation **does not happen**
8. `Add()` stores requirements with `zone=[eu-west-1b]` but `minValues=2`
9. `ToNodeClaim()` creates NodeClaim with `minValues: 2` and `values: [eu-west-1b]`
10. CRD CEL validation rejects: `1 value < minValues 2`

**The fix should be:** `SatisfiesMinValues` should check the **intersection** of instance type values and the NodeClaim requirements (passed as the `requirements` parameter), not just the raw instance type values. Something like:

```go
// Current (broken):
valuesForKey[req.Key] = valuesForKey[req.Key].Insert(it.Requirements.Get(req.Key).Values()...)

// Fixed:
itValues := sets.New[string](it.Requirements.Get(req.Key).Values()...)
reqValues := sets.New[string](req.Values()...)
valuesForKey[req.Key] = valuesForKey[req.Key].Insert(itValues.Intersection(reqValues).UnsortedList()...)
```

### Reproduction Steps

**NodePool configuration:**
```yaml
apiVersion: karpenter.sh/v1
kind: NodePool
metadata:
  name: monitoring
spec:
  template:
    spec:
      requirements:
          operator: In
          values:
            - eu-west-1b
            - eu-west-1c
          minValues: 2
          operator: In
          values:
            - spot
            - on-demand
```

**Karpenter controller settings:**
```
MIN_VALUES_POLICY=BestEffort
```

**Trigger:** Any pod with a PVC bound to a specific zone (e.g., EBS volume in eu-west-1b) that needs to be scheduled on the `monitoring` NodePool.

In our case, this was triggered by:
- Karpenter consolidation trying to replace a monitoring node
- The monitoring namespace has 42 PVC-attached pods (Prometheus, Thanos, Loki StatefulSets)
- PVCs are EBS-backed and zone-locked

### Environment

- **Karpenter AWS Provider:** v1.8.6
- **Karpenter Core:** v1.8.2 (dependency of AWS provider v1.8.6)
- **Kubernetes:** EKS 1.34
- **Region:** eu-west-1 (2 AZs configured: eu-west-1b, eu-west-1c)

### Impact

The `minValues` feature on `topology.kubernetes.io/zone` is unusable for any NodePool that hosts workloads with zone-pinned PVCs, regardless of the `MIN_VALUES_POLICY` setting. This affects common monitoring stacks (Prometheus, Thanos, Loki) and any StatefulSet with persistent storage.

We were using `minValues: 2` to ensure Karpenter spreads nodes across both AZs (as part of a 3→2 AZ migration). We had to revert the change after it blocked monitoring node creation for ~40 minutes in production.

### Related

- #2078 — Feature request for `whenMinValuesUnsatisfiable` (closed, implemented by #2299)
- #2299 — PR that added `BestEffort` minValuesPolicy
- #1034 — minValues invalid with topologySpread
- #1208 — NodeClaim validation fails when requirements are narrowed

karan-vk:
Fixes #2999

**Description**

`SatisfiesMinValues` counts each instance type's raw offered values instead of intersecting them with the NodeClaim's requirements. When pod constraints (PVC volume topology, topology spread) narrow a requirement that carries `minValues` e.g. zone `In [a, b] minValues: 2` narrowed to `In [a]` the check still sees every zone the instance types offer, reports minValues as satisfied, and the BestEffort relaxation never triggers. The NodeClaim is then created with fewer values than `minValues` and rejected by the CRD CEL validation, blocking capacity for affected workloads.

This change filters each instance type's values through `Requirement.Has()` before counting, so only values the narrowed requirement still allows count toward minValues. BestEffort then relaxes correctly (with the relaxed annotation set), and Strict fails at scheduling simulation with a clear error instead of failing at NodeClaim creation.

Note on Strict mode: requirements whose achievable value set is genuinely smaller than `minValues` (e.g. `NotIn`/bounded operators) now fail scheduling simulation rather than producing a NodeClaim that would be rejected downstream a correction, but an observable behavior change.

- Added regression tests reproducing the reported scenario (`WaitForFirstConsumer` StorageClass pinned to one zone + `minValues` on the NodePool zone requirement): without the fix the NodeClaim fails CEL validation and the pod stays pending; with it the pod schedules into the pinned zone with `minValues` relaxed to 1 and the relaxed annotation set. Also covers Strict mode (pod correctly not scheduled).
- Full `provisioning`, `provisioning/scheduling`, `cloudprovider`, `scheduling`, and `disruption` suites pass with `-race` (disruption/consolidation also calls `SatisfiesMinValues`).

### DIFF
--- a/pkg/cloudprovider/types.go
+++ b/pkg/cloudprovider/types.go
@@ -410,7 +410,12 @@ func (its InstanceTypes) SatisfiesMinValues(requirements scheduling.Requirements
 				if _, ok := valuesForKey[req.Key]; !ok {
 					valuesForKey[req.Key] = sets.New[string]()
 				}
-				valuesForKey[req.Key] = valuesForKey[req.Key].Insert(it.Requirements.Get(req.Key).Values()...)
+				// Only count values that the requirement allows. The instance type may offer values (e.g. zones)
+				// that the requirements exclude after being narrowed by pod constraints such as volume topology
+				// or topology spread, and those values can't be satisfied by the resulting NodeClaim.
+				valuesForKey[req.Key] = valuesForKey[req.Key].Insert(lo.Filter(it.Requirements.Get(req.Key).Values(), func(value string, _ int) bool {
+					return req.Has(value)
+				})...)
 			}
 		}
 		for k, v := range valuesForKey {
--- a/pkg/controllers/provisioning/suite_test.go
+++ b/pkg/controllers/provisioning/suite_test.go
@@ -3248,6 +3248,29 @@ var _ = Describe("Provisioning", func() {
 					ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
 					ExpectNotScheduled(ctx, env.Client, pod)
 				})
+
+				It("should not schedule when pod volume topology narrows zones below minValues", func() {
+					// The instance type offers all three zones, but the pod's volume topology narrows the
+					// NodeClaim requirements to a single zone, which can't satisfy minValues=3
+					cloudProvider.InstanceTypes = []*cloudprovider.InstanceType{multiZoneInstanceType("instance-type-1", "test-zone-1", "test-zone-2", "test-zone-3")}
+
+					storageClass := test.StorageClass(test.StorageClassOptions{
+						Zones:             []string{"test-zone-1"},
+						VolumeBindingMode: lo.ToPtr(storagev1.VolumeBindingWaitForFirstConsumer),
+					})
+					persistentVolumeClaim := test.PersistentVolumeClaim(test.PersistentVolumeClaimOptions{StorageClassName: &storageClass.Name})
+					ExpectApplied(ctx, env.Client, nodePool, storageClass, persistentVolumeClaim)
+					pod := test.UnschedulablePod(test.PodOptions{
+						PersistentVolumeClaims: []string{persistentVolumeClaim.Name},
+						ResourceRequirements: corev1.ResourceRequirements{Requests: corev1.ResourceList{
+							corev1.ResourceCPU:    resource.MustParse("0.9"),
+							corev1.ResourceMemory: resource.MustParse("0.9Gi")},
+						},
+					})
+
+					ExpectProvisioned(ctx, env.Client, cluster, cloudProvider, prov, pod)
+					ExpectNotScheduled(ctx, env.Client, pod)
+				})
 			})
 
 			Context("with MinValuesPolicy set to BestEffort", func() {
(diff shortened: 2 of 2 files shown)


ryan-mist (approved):
/lgtm

DerekFrank (approved):
/lgtm

PR #3177, opened 2026-07-27
feat: add nodeclaim condition metrics by nodepool

### LINKED ISSUE #3071: Surface NodeClaim drift/rollout progress in NodePool status

cnmcavoy:
Fixes #3071

**Description**
Adds a new metric `karpenter_nodepools_nodeclaim_condition` which tracks the number of nodeclaims by condition and status for each nodepool. This enables tracking various states in nodepools, like rollout progress of drift for a mass-change (e.g AMI updates). 

`make presubmit` and running locally in our development clusters.

### DIFF
--- a/pkg/controllers/metrics/nodepool/controller.go
+++ b/pkg/controllers/metrics/nodepool/controller.go
@@ -38,7 +38,13 @@ import (
 	"sigs.k8s.io/karpenter/pkg/metrics"
 	"sigs.k8s.io/karpenter/pkg/operator/injection"
 	"sigs.k8s.io/karpenter/pkg/state/cost"
+	nodeclaimutils "sigs.k8s.io/karpenter/pkg/utils/nodeclaim"
 	nodepoolutils "sigs.k8s.io/karpenter/pkg/utils/nodepool"
+)
+
+const (
+	conditionLabel       = "condition"
+	conditionStatusLabel = "status"
 )
 
 var (
@@ -78,6 +84,20 @@ var (
 		},
 		[]string{metrics.NodePoolLabel},
 	)
+	NodeClaimCondition = opmetrics.NewPrometheusGauge(
+		crmetrics.Registry,
+		prometheus.GaugeOpts{
+			Namespace: metrics.Namespace,
+			Subsystem: metrics.NodePoolSubsystem,
+			Name:      "nodeclaim_condition",
+			Help:      "The number of nodeclaims owned by a nodepool broken down by status condition type and status. Labeled by nodepool, condition type, and condition status. Useful for tracking rollout progress, e.g. drift reconciliation via condition=\"Drifted\",status=\"True\".",
+		},
+		[]string{
+			metrics.NodePoolLabel,
+			conditionLabel,
+			conditionStatusLabel,
+		},
+	)
 )
 
 type Controller struct {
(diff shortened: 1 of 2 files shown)


pkg/controllers/metrics/nodepool/controller.go
```diff
@@ -78,6 +84,20 @@ var (
 		},
 		[]string{metrics.NodePoolLabel},
 	)
+	NodeClaimCondition = opmetrics.NewPrometheusGauge(
```
DerekFrank, line 87: We have the operatorpkg status metrics for this, is the main thing missing the nodepool label? https://karpenter.sh/docs/reference/metrics/#operator_nodeclaim_status_condition_count

cnmcavoy: Ah, we have the cluster state observability metrics disabled because the performance in large clusters is terrible. It would be nice to have a way to selectively enable/disable some of the operatorpkg metrics, but I can open a separate issue for that.

DerekFrank: I think we might want to change the cluster state observability metrics disable to be cleaner, if we just remove the nodeclaim dimension from a lot of things we might be able to enable it for almost everyone

cnmcavoy: I opened https://github.com/kubernetes-sigs/karpenter/issues/3319 as follow up

